PR #1979, opened 2020-08-03
bugfix: empty `allow_admin` means that any IP is allowed to access Admin API and Dashboard

### LINKED ISSUE #1939: request help: failed to access Admin API, return 403 always
### Issue description
curl http://127.0.0.1:9080/apisix/admin/routes/10 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "uri": "/index.html",
    "hosts": ["127.0.0.1"],
    "remote_addrs": ["127.0.0.0/8"],
    "methods": ["PUT", "GET"],
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:80": 1
        }
    }
}'

image

### Environment
 本地docker 部署

* apisix version (cmd: `apisix version`):1.4.1
* OS:mac

membphis:
…min API.

fix #1939

### Pre-submission checklist:


spacewander:
Need to update apisix_cli_test.sh  before merge.

spacewander:
The `.travis/apisix_cli_test.sh` is in conflicted now.

.travis/apisix_cli_test.sh
```diff
@@ -211,6 +211,33 @@ fi
...
+
+sed  -i 's/- 127.0.0.9/\# - 127.0.0.9/'  conf/config.yaml
+
+make init
+
+set +ex
+grep "allow 127.0.0.9;" conf/nginx.conf > /dev/null
+if [ ! $? -eq 1 ]; then
+    echo "failed: found allow 127.0.0.9; in conf/nginx.conf"
+    exit 1
```
Yiyiyimu, line 236: I think we should not check if 127.0.0.9 is in nginx.conf, since it's commented out, it would never be in nginx.conf, so current version would still pass this test.

I think we need to test if there is
```
location /apisix/admin {
         deny all;
```
which cause this problem.

membphis: this way is better, updated

.travis/apisix_cli_test.sh
```diff
@@ -211,8 +211,40 @@ fi
...
+fi
+
+echo "
+apisix:
+    allow_admin: ~
+" > conf/config.yaml
+
+make init
+
+count=`grep -c "allow all;" conf/nginx.conf || true`
```
spacewander, line 237: Look like the `|| true` doesn't take effect?

membphis: nice, removed

spacewander (approved):

Firstsawyou:
I think the title is not a good title, I think it’s better like this:  "empty `allow_admin` means that any IP is allowed to access Admin API and Dashboard. " . Because this `pr` does two things: allowed to access Admin API and Dashboard.

.travis/apisix_cli_test.sh
```diff
@@ -211,8 +211,40 @@ fi
 
 echo "passed: worker_shutdown_timeout in nginx.conf is ok"
 
+# empty allow_admin in conf/config.yaml
```
Firstsawyou, line 214: I think the title is not a good title. Because the following example has a situation where ʻallow_admin` is not empty.

Firstsawyou (changes requested):

.travis/apisix_cli_test.sh
membphis: nice catch

Yiyiyimu (approved):

PR #1980, opened 2020-08-03
bugfix: fail to handshake if there is no certificate matched by host.

### LINKED ISSUE #1957: bug: return fail if failed to load SSL certificate by SNI
https://github.com/apache/incubator-apisix/blob/master/apisix/init.lua#L162

only `warn` message is not enough.

membphis:

fixed #1957

### Pre-submission checklist:


nic-chen:
It looks like we need a SSL cert (127.0.0.1)  for test case..

membphis:
> It looks like we need a SSL cert (127.0.0.1) for test case..

How about we use `test.com` for testing?

membphis:
@nic-chen many thx for your help. ^_^

PR #1989, opened 2020-08-04
Fixed  discovery weight bug

frank0417:
### Pre-submission checklist:

### DIFF
--- a/apisix/discovery/eureka.lua
+++ b/apisix/discovery/eureka.lua
@@ -207,7 +207,7 @@ local function fetch_full_registry(premature)
                 end
                 local weight = default_weight
                 if metadata and metadata.weight then
-                    weight = tonumber(metadata.weight)
+                    weight = tonumber(metadata.weight) or default_weight
                 else
                     metadata.weight = nil
                 end
--- a/apisix/discovery/eureka.lua
+++ b/apisix/discovery/eureka.lua
@@ -197,22 +197,26 @@ local function fetch_full_registry(premature)
     for _, app in ipairs(apps) do
         for _, instance in ipairs(app.instance) do
             local ip, port, metadata = parse_instance(instance)
+        
             if ip and port then
                 local nodes = up_apps[app.name]
                 if not nodes then
                     nodes = core.table.new(#app.instance, 0)
                     up_apps[app.name] = nodes
                 end
+                local tmp_weight = default_weight
+                if metadata and metadata.weight then
+                    tmp_weight = tonumber(metadata.weight)
+                else
+                    metadata.weight = nil
+                end
                 core.table.insert(nodes, {
                     host = ip,
                     port = port,
-                    weight = metadata and metadata.weight or default_weight,
+                    weight = tmp_weight,
                     metadata = metadata,
                 })
-                if metadata then
-                    -- remove useless data
-                    metadata.weight = nil
-                end
+               
             end
         end
     end
--- a/apisix/discovery/eureka.lua
+++ b/apisix/discovery/eureka.lua
@@ -36,7 +36,6 @@ local default_weight
 local applications
 
 
-
 local schema = {
     type = "object",
     properties = {
--- a/apisix/discovery/eureka.lua
+++ b/apisix/discovery/eureka.lua
@@ -197,26 +197,24 @@ local function fetch_full_registry(premature)
     for _, app in ipairs(apps) do
         for _, instance in ipairs(app.instance) do
             local ip, port, metadata = parse_instance(instance)
-        
             if ip and port then
                 local nodes = up_apps[app.name]
                 if not nodes then
                     nodes = core.table.new(#app.instance, 0)
                     up_apps[app.name] = nodes
                 end
-                local tmp_weight = default_weight
+                local weight = default_weight
                 if metadata and metadata.weight then
-                    tmp_weight = tonumber(metadata.weight)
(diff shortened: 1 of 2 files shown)


spacewander:
@frank0417 
Please explain why the 'fix' is needed.

frank0417:
> @frank0417
> Please explain why the fix is needed, and add test case for it.

When Eureka is used as a service discovery, the service is configured with weights. For example, the weight of servicea1 is configured as metadata- map.weight:20 ，ServiceA2 The weight configuration of is: metadata- map.weight:80 At present, the weight value of apifix is cached according to the string when parsing, which results in the request can not be handled correctly when the load is loaded, and the load balancing can not be carried out according to the weight set by the service. The default weight is used for the load, for example, the weight is 100

frank0417:
> @frank0417
> Please explain why the fix is needed, and add test case for it.
test case:
ServiceA1:
server:
  port: 9097

eureka:
  instance:
    leaseRenewalIntervalInSeconds: 1
    leaseExpirationDurationInSeconds: 2
    metadata-map:
      weight: 20
  client:
    serviceUrl:
      defaultZone: http://127.0.0.1:8761/eureka/
    healthcheck:
      enabled: true
    lease:
      duration: 5


serviceA2 :

  server:
  port: 9098

eureka:
  instance:
    leaseRenewalIntervalInSeconds: 1
    leaseExpirationDurationInSeconds: 2
    metadata-map:
      weight: 80
  client:
    serviceUrl:
      defaultZone: http://127.0.0.1:8761/eureka/
    healthcheck:
      enabled: true
    lease:
      duration: 5

spring:
  application:
    name: customer-service

membphis:
@qiujiayu please take a look at this PR, need you review

apisix/discovery/eureka.lua
```diff
@@ -197,22 +197,26 @@ local function fetch_full_registry(premature)
...
         for _, instance in ipairs(app.instance) do
             local ip, port, metadata = parse_instance(instance)
+        
             if ip and port then
                 local nodes = up_apps[app.name]
                 if not nodes then
                     nodes = core.table.new(#app.instance, 0)
                     up_apps[app.name] = nodes
                 end
+                local tmp_weight = default_weight
```
qiujiayu, line 207: It would be better to use 'weight'

membphis:
@frank0417 please take a look at the output of CI: https://github.com/apache/apisix/pull/1989/checks?check_run_id=952820522#step:6:125

you need to fix them first

apisix/discovery/eureka.lua
```diff
@@ -34,6 +35,8 @@ local log                = core.log
 local default_weight
 local applications
 
+
+
```
membphis, line 39: two blank lines are enough but got `3`

apisix/discovery/eureka.lua
```diff
@@ -203,16 +205,18 @@ local function fetch_full_registry(premature)
                     nodes = core.table.new(#app.instance, 0)
                     up_apps[app.name] = nodes
                 end
+                local weight = default_weight
+                if metadata and metadata.weight then
+                    weight = tonumber(metadata.weight)
+                else
+                    metadata.weight = nil
+                end
```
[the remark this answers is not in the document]
frank0417: ok，the result of tonumber may be nil

apisix/discovery/eureka.lua
```diff
@@ -203,16 +205,18 @@ local function fetch_full_registry(premature)
                     nodes = core.table.new(#app.instance, 0)
                     up_apps[app.name] = nodes
                 end
+                local weight = default_weight
+                if metadata and metadata.weight then
+                    weight = tonumber(metadata.weight) or default_weight
+                else
+                    metadata.weight = nil
```
membphis, line 212: why we need this line? Can we remove this line?

membphis: and we need test case to confirm this change is correct.

frank0417: > why we need this line? Can we remove this line?

Which line ？

membphis: `metadata.weight = nil`

frank0417: > and we need test case to confirm this change is correct.

We need to learn about the test framework. We haven't touched it before. We will submit the test case later.
haha

frank0417:
> missing test case

test cases have been added

apisix/discovery/eureka.lua
frank0417: > `metadata.weight = nil`
   remove useless data

t/discovery/eureka-weight.t
```diff
@@ -0,0 +1,168 @@
...
+--- request
+GET /t
+--- yaml_config eval: $::yaml_config
+--- response_body
+[{"count":8,"port":"1981"},{"count":2,"port":"1980"}]
+--- no_error_log
+[error]
+
+
+
```
membphis, line 158: only one blank line is need.

frank0417: Test cases can run through the local test case, and fail to test on CI. Can you give me some advice

membphis: > Test cases can run through the local test case, and fail to test on CI. Can you give me some advice

@qiujiayu need your help ^_^

moonming:
@frank0417 @membphis ping

PR #2004, opened 2020-08-06
feat: add '**' for alllow_headers and complete docs

### LINKED ISSUE #1983: bug: cors when the request is a preflight request we shouldn't set the Access-Control-Allow-Headers *
### Issue description
when the request is a preflight request we shouldn't set the Access-Control-Allow-Headers, if we set the Access-Control-Allow-Headers * we will get a error from google chrome.

### Environment

* OS:

1.
2.
3.

### My way to solve this
```lua

local function set_allow_headers_options_method(conf,ctx)

    local allow_headers = conf.allow_headers

    if allow_headers == "*" or allow_headers == "**" then

        local headers = core.request.header(ctx,"Access-Control-Request-Headers")

        if headers then
            ngx.header["Access-Control-Allow-Headers"] = headers
        end

    end

end

        if ctx.var.request_method == "OPTIONS" then
            set_allow_headers_options_method(conf,ctx)
            return 200
        end

```

ShiningRush:
Resolve #1983

### DIFF
--- a/apisix/plugins/cors.lua
+++ b/apisix/plugins/cors.lua
@@ -42,6 +42,7 @@ local schema = {
         allow_headers = {
             description =
                 "you can use '*' to allow all header when no credentials," ..
+                "'**' to allow forcefully(it will bring some security risks, be carefully)," ..
                 "multiple header use ',' to split. default: *.",
             type = "string",
             default = "*"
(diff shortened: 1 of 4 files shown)


spacewander:
Look like we only need to add `Access-Control-Allow-Methods` and so on in preflight request:
https://fetch.spec.whatwg.org/#http-access-control-expose-headers
Since it is too late to ask for permission in normal request.

Anyway, useless headers are fine to me...

PR #2029, opened 2020-08-09
bugfix: only set one response header when enabled `enable_debug=true`

### LINKED ISSUE #1667: bug:  global-rule error when enable_debug=true
### Issue description
I customized a plug-in for global rules,When I set the configuration enable_debug=true, i got the follow error:
```
2020/06/08 11:40:56 [error] 15825#0: *330 failed to run body_filter_by_lua*: /usr/local/share/lua/5.1/apisix/plugin.lua:260: headers have already been sent
stack traceback:
        [C]: in function 'error'
        /usr/local/share/lua/5.1/apisix/core/response.lua:90: in function 'set_header'
        /usr/local/share/lua/5.1/apisix/plugin.lua:260: in function 'filter'
        /usr/local/share/lua/5.1/apisix/init.lua:456: in function 'common_phase'
        /usr/local/share/lua/5.1/apisix/init.lua:470: in function 'http_body_filter_phase'
        body_filter_by_lua:2: in main chunk, client: ::1, server: , request: "GET /config HTTP/1.1", host: "localhost:9080"

```

### Environment

* apisix version (cmd: `apisix version`): 1.2
* OS: Ubuntu 18.04

1. config.yaml
```
apisix:
     enable_debug: true

plugins:
    - my_custom_plugin
```
2. my_custom_plugin.lua
```

local plugin_name = "my_custom_plugin"
local schema = {}

local _M = {
    version = 0.1,
    priority = 0,
    name = plugin_name,
    schema = schema,
}

function _M.access(conf, ctx)
--    core.response.exit(200, "测试！！！")
--    return 200, "测试！！！"
end
return _M
```
3. set global rule:
```
curl -X PUT \
  http://localhost:9080/apisix/admin/global_rules/1 \
  -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' \
  -d '{
        "plugins": {
            "my_custom_plugin": {}
        }
    }'

```

```
➜  ~ curl http://localhost:9080/config
curl: (52) Empty reply from server
```

➜  ~ curl http://localhost:9080/config
{"error_msg":"failed to match any routes"}

membphis:
and `global rules`.

fix #1667

### Pre-submission checklist:

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -228,13 +228,33 @@ function _M.api_routes()
 end
 
 
-function _M.filter(user_route, plugins)
+local function set_response_header_by_debug_flag(plugins, dry_run)
+    if dry_run then
+        return
+    end
+
+    if not local_conf or not local_conf.apisix.enable_debug then
+        return
+    end
+
+    if #plugins == 0 then
+        core.response.set_header("Apisix-Plugins", "no plugin")
+    end
+
+    local t = {}
+    for i = 1, #plugins, 2 do
+        core.table.insert(t, plugins[i].name)
+    end
+    core.response.set_header("Apisix-Plugins", core.table.concat(t, ", "))
+end
+
+
+function _M.filter(user_route, plugins, dry_run)
     local user_plugin_conf = user_route.value.plugins
     if user_plugin_conf == nil or
-       core.table.nkeys(user_plugin_conf) == 0 then
-        if local_conf and local_conf.apisix.enable_debug then
-            core.response.set_header("Apisix-Plugins", "no plugin")
-        end
+       core.table.nkeys(user_plugin_conf) == 0
+    then
+        set_response_header_by_debug_flag(core.empty_tab, dry_run)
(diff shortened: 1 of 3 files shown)


apisix/plugin.lua
```diff
@@ -228,13 +228,34 @@ function _M.api_routes()
 end
 
 
-function _M.filter(user_route, plugins)
+local function set_response_header_by_debug_flag(plugins, dry_run)
+    if dry_run then
```
spacewander, line 232: I think we can use `ngx.headers_sent` instead of passing flag manually.

spacewander: And to avoid the conflict between global rule / route rule, it would be better to use `add_header`.
Maybe we still need to pass the `global` flag to indicate if the current filtering rule is global or not?

membphis: > I think we can use `ngx.headers_sent` instead of passing flag manually.

yes, this is a much better way. I gonna close this PR.

PR #2035, opened 2020-08-10
fix: prometheus metric api bug

### LINKED ISSUE #2025: bug: comparing etcd key's modifiedIndex as a number may cause an error when access to prometheus exporter api
### Issue description

When any upstream node uses domain name as host, modifiedIndex of the etcd key for upstream is modifed to string in the domain name resolution caching logic:
https://github.com/apache/apisix/blob/master/apisix/init.lua#L251

At this time, comparing the etcd key's modifiedIndex as a number will cause an error:
https://github.com/apache/apisix/blob/master/apisix/plugins/prometheus/exporter.lua#L181

### Environment

* apisix version (cmd: `apisix version`): 1.5
* OS: centos 7

1.  use domain name as host for an upstream node.
2. resolve domain name to another IP
3. access prometheus exporter api

cause an error

normal response

nic-chen:
fix #2025 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -178,8 +178,9 @@ local function set_modify_index(key, items, items_ver, global_max_index)
     local max_idx = 0
     if items_ver and items then
         for _, item in ipairs(items) do
-            if type(item) == "table" and item.modifiedIndex > max_idx then
-                max_idx = item.modifiedIndex
+            local modify_index = item.modifiedIndex_org or item.modifiedIndex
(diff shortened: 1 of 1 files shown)


nic-chen:
This is a fault-tolerant handling, and it is not easy to simulate the scenario(changing domain name resolution) to test, so considering test case `TODO`. If you have any suggestions, please let me know, thank you.

membphis:
> This is a fault-tolerant handling, and it is not easy to simulate the scenario(changing domain name resolution) to test, so considering test case `TODO`

we need to add a comment in the source code or create a new issue about this.

PR #2079, opened 2020-08-19
fix: proxy_set_header Host $host may lose port info

halfcrazy:

proxy_set_header Host $host may lose port info, use http_host instead.
### Pre-submission checklist:

### DIFF
--- a/t/node/http_host.t
+++ b/t/node/http_host.t
@@ -1,0 +1,72 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
+log_level('info');
+worker_connections(256);
+no_root_location();
+no_shuffle();
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: set route(id: 1)
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/uri",
+                    "upstream": {
+                        "nodes": {
+                            "127.0.0.1:1980": 1
+                        },
+                        "type": "roundrobin"
+                    }
+                }]]
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 2: hit routes
+--- request
+GET /uri
+--- more_headers
+Host: foo.com:1984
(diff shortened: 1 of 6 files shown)


membphis:
pls add test cases

moonming:
@halfcrazy `$http_host` ONLY get value from request header field, is that ok?
`$host`: in this order of precedence: host name from the request line, or host name from the “Host” request header field, or the server name matching a request 
refer to: http://nginx.org/en/docs/http/ngx_http_core_module.html

moonming:
> proxy_set_header Host $host may lose port info, use http_host instead.

And is this PR anything with `proxy_set_header`?

halfcrazy:
> > proxy_set_header Host $host may lose port info, use http_host instead.
> 
> And is this PR anything with `proxy_set_header`?

Since we use proxy_set_header to set the Host field in header for upstream server. Nginx $host variable doesn't contain a port, as  rfc2616 a request to ports besides 80, 443 should always include a port in the host field.

[1]https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

halfcrazy:
> pls add test cases

Sorry, I have no idea how to add a test for this case.

moonming:
> > pls add test cases
> 
> Sorry, I have no idea how to add a test for this case.

you can add test cases to test if port exists.

membphis:
@halfcrazy https://github.com/apache/apisix/blob/master/.travis/apisix_cli_test.sh

you can take a look at this `shell` script for testing.

membphis:
> Test case added.

"proxy_set_header Host $host may lose port info, use http_host instead."

we need to add test for this case

moonming:
ping @halfcrazy @membphis

halfcrazy:
> Hi, May I ask the conclusion of this question? tks @halfcrazy

There's a lot of change to apisix template, the $host indeed should change to $http_host but I don't familiar with Perl style test cant write a test for this.

membphis:
let us reopen this PR, I think @Firstsawyou can help you, all right? @Firstsawyou

membphis:
I think we need to process the conflict files first.

image

Firstsawyou:
> let us reopen this PR, I think @Firstsawyou can help you, all right? @Firstsawyou

Let me see.

Firstsawyou:
hi @halfcrazy 
The title of the PR should look like this:
 `fix: proxy_set_header Host $host may lose port info`

t/lib/server.lua
```diff
@@ -73,6 +73,7 @@ end
 function _M.plugin_proxy_rewrite()
     ngx.say("uri: ", ngx.var.uri)
     ngx.say("host: ", ngx.var.host)
+    ngx.say("http_host: ", ngx.var.http_host)
```
Firstsawyou, line 76: It is not necessary to add this line of code. Try this `uri`: https://github.com/apache/apisix/blob/master/t/lib/server.lua#L111

t/node/http_host.t
```diff
@@ -0,0 +1,72 @@
...
+
+=== TEST 2: hit routes
+--- request
+GET /uri
+--- more_headers
+Host: foo.com
+--- response_body
+host: foo.com:1984
+--- no_error_log
+[error]
```
Firstsawyou, line 72: Hi, @halfcrazy 
I think this would be better：

```
=== TEST 1: set route(id: 1)
--- config
    location /t {
        content_by_lua_block {
            local code, body = t('/apisix/admin/routes/1',
                ngx.HTTP_PUT,
                [[{
                    "uri": "/uri",
                    "upstream": {
                        "nodes": {
                            "127.0.0.1:1980": 1
                        },
                        "type": "roundrobin"
                    }
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 2: hit routes
--- request
GET /uri
--- more_headers
host: test.com:8080
--- response_body
uri: /uri
host: test.com:8080
x-real-ip: 127.0.0.1
--- no_error_log
[error]

```

spacewander (approved):

lilien1010 (approved):
LGTM

PR #2094, opened 2020-08-21
feature: added to limit req based on the request header and request parameters.

gy09535:
we can use  request parameters and headers to limit req.

### Pre-submission checklist:


moonming:
@gy09535 Apache APISIX already support all nginx variables as limit key: http://nginx.org/en/docs/varindex.html, you just need to add it in the enum list of key: https://github.com/apache/apisix/blob/master/apisix/plugins/limit-req.lua#L29.

moonming:
> @gy09535 Apache APISIX already support all nginx variables as limit key: http://nginx.org/en/docs/varindex.html, you just need to add it in the enum list of key: https://github.com/apache/apisix/blob/master/apisix/plugins/limit-req.lua#L29.

It's a good idea to write in the doc.

gy09535:
Sentinel has  implementation this requirement. according to https://github.com/alibaba/Sentinel/wiki/%E7%BD%91%E5%85%B3%E9%99%90%E6%B5%81#%E7%BD%91%E5%85%B3%E6%B5%81%E6%8E%A7%E6%8E%A7%E5%88%B6%E5%8F%B0, I  think it is very useful.

gy09535:
> > @gy09535 Apache APISIX already support all nginx variables as limit key: http://nginx.org/en/docs/varindex.html, you just need to add it in the enum list of key: https://github.com/apache/apisix/blob/master/apisix/plugins/limit-req.lua#L29.
> 
> It's a good idea to write in the doc.

Sentinel has  implementation this requirement. according to https://github.com/alibaba/Sentinel/wiki/%E7%BD%91%E5%85%B3%E9%99%90%E6%B5%81#%E7%BD%91%E5%85%B3%E6%B5%81%E6%8E%A7%E6%8E%A7%E5%88%B6%E5%8F%B0, I  think it is very useful.

Issue #2111, opened 2020-08-24
bug: plugins should not exit by itself


nic-chen:
### Issue description
plugins should not exit by itself in access and rewrite phase, it can't be reuse in `script`, and the plugin engine has already covered:

https://github.com/apache/apisix/blob/master/apisix/init.lua#L134

### Environment

* apisix version (cmd: `apisix version`): 1.5
* OS:

### Minimal test code / Steps to reproduce the issue
1.
2.
3.

### What's the actual result? (including assertion message & call stack if applicable)

exit by plugin engine

### What's the expected result?

exit by itself

PR #2115, opened 2020-08-24
fix: plugins should not exit by itself

nic-chen:
plugins should not exit by itself in access and rewrite phase, it can't be reuse in script, and the plugin engine has already covered:

https://github.com/apache/apisix/blob/master/apisix/init.lua#L134

#2111 

### Pre-submission checklist:


moonming (approved):

membphis:
nice PR

PR #2120, opened 2020-08-25
[key-auth]fix: skip consumer when config has no key

### LINKED ISSUE #2045: bug: config of plugin `key-auth` for consumer shouldn't be empty
### Issue description
config item `key` of plugin `key-auth` shouldn't be empty

### Environment

* apisix version (cmd: `apisix version`): 1.5
* OS: centos 7

1. 

```
curl http://127.0.0.1:9080/apisix/admin/consumers -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "plugins":{
        "key-auth":{
        }
    },
    "username":"test"
}'
```

2.

```
{
    "uri": "/*",
    "plugins":{
        "key-auth":{
        }
    },
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:80": 1
        }
    }
}'
```

3.

```
curl http://127.0.0.1:9080/test -H 'apikey: 267c6cf0' -d '{"token":"3c3d3af8642e4d4d9439f9f1d317de92"}' -v
```

occur 500

error log:
`2020/08/11 07:29:13 [error] 24#24: *15975 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/plugins/key-auth.lua:49: table index is nil`

normal respond

ShiningRush:
resolve #2045 .


t/plugin/key-auth.t
```diff
@@ -215,3 +215,56 @@ apikey: auth-13
...
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/consumers',
+                ngx.HTTP_PUT,
+                [[{
+                    "username": "error",
+                    "plugins": {
+                        "key-auth": {
+                        }
```
membphis, line 232: > add consumer with empty key

should throw an error message if there is no argument when create a new consumer.

ShiningRush: I think we should not simply throw errors in the consumer, but have a mechanism for the plugin to define its own verification behavior

membphis: Add a new parameter, which is the resource type to which the plugin is bound.

Determine the content of the authentication plug-in according to the type of resource created.

Here is an example of `key-auth`:

```lua
function _M.check_schema(conf, parrent_type)
    local ok, err = core.schema.check(schema, conf)
    if not ok then
        return false, err
    end

    if parrent_type == "consumer" then
        if core.table.nkeys(conf) == 0 then
            return false, "missing properties"
        end
    end

    return true
end
```

ShiningRush: I think it is good, but maybe we should have a discuss in mail list like the changes about core?

membphis: nice, you can create a topic about this ^_^

moonming: any update?

membphis (approved):
we can merge this PR first, fix https://github.com/apache/apisix/pull/2120/files#r477153261 later

moonming (approved):

Issue #2144, opened 2020-09-02
optimization: cache the parsed certs and pkeys into LRU queues to reduce the parsing overheads


tokers:
We can parse the result of `ssl. parse_pem_cert` and `ssl. parse_pem_priv_key` into LRU cache queues, this should help to reduce the CPU utilizations of `ssl_certificate_by_lua*` phase.

membphis:
yes, this way is better. welcome PR

tokers:
Will implement this as soon as possible.

tokers:
The related PR was merged, i'm closing this issue.

PR #2163, opened 2020-09-03
improve: cache parsed SSL certs and pkeys

tokers:

The processes of `ngx_ssl.parse_pem_cert` and `ngx_ssl.parse_pem_priv_key` are relatively expensive, and keypair is inclined to be constant (for a long while), so it's worth to cache the parsed result into LRU cache, which can reduce a part of overheads during `ssl_certificate_by_lua*` phase.

### Pre-submission checklist:

### DIFF
--- a/apisix/http/router/radixtree_sni.lua
+++ b/apisix/http/router/radixtree_sni.lua
@@ -31,11 +31,35 @@ local ssl_certificates
 local radixtree_router
 local radixtree_router_ver
 
+local cert_cache = core.lrucache.new {
+    ttl = 3600, count = 512,
+}
+
+local pkey_cache = core.lrucache.new {
+    ttl = 3600, count = 512,
+}
+
 
 local _M = {
     version = 0.1,
     server_name = ngx_ssl.server_name,
 }
+
+
+local function parse_pem_cert(sni, cert)
+    core.log.debug("parsing cert for sni: ", sni)
+
+    local parsed, err = ngx_ssl.parse_pem_cert(cert)
+    return parsed, err
+end
+
+
+local function parse_pem_priv_key(sni, pkey)
+    core.log.debug("parsing priv key for sni: ", sni)
+
+    local parsed, err = ngx_ssl.parse_pem_priv_key(pkey)
+    return parsed, err
+end
 
 
 local function create_router(ssl_items)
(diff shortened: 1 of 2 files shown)


tokers:
Related issue: https://github.com/apache/apisix/issues/2144

tokers:
Ping @membphis @moonming

spacewander:
Would it be better to store the parsed data in `ssl.value` like https://github.com/apache/apisix/blob/f6228b180946a0c15dd314cdc0c53a9024d04823/apisix/http/router/radixtree_sni.lua#L104?

Therefore we can ensure the parsed data will be reused. If we use LRU cache and cache is not large enough, the newly parsed data will evict an old one.

tokers:
@spacewander The size of LRU cache queue is 512, which is relatively large and is uncommon in most cases, but indeed the way you recommended is better since we can avoid the duplicate parsing totally.

apisix/http/router/radixtree_sni.lua
```diff
@@ -95,6 +134,9 @@ local function create_router(ssl_items)
         end
     end
 
+    parsed_certs = core.table.new(0, idx)
+    parsed_pkeys = core.table.new(0, idx)
```
spacewander, line 138: Maybe we can store the parsed data in `ssl.value` directly, so the data can be reused even after router recreated.

membphis: The current way is wrong. We cannot store them directly in the hash table. This method will cause memory overflow.

We need to cache the parsed data in lrucache, and we need to control the number of cached certificates.

apisix/http/router/radixtree_sni.lua
```diff
@@ -30,14 +30,53 @@ local ngx_decode_base64 = ngx.decode_base64
...
+    end
+
+    core.log.debug("parsing priv key for sni: ", sni)
+
+    local parsed, err = ngx_ssl.parse_pem_priv_key(pkey)
+    if not parsed then
+        return nil, err
+    end
+
+    parsed_pkeys[pkey] = parsed
```
membphis, line 74: ditto

apisix/http/router/radixtree_sni.lua
```diff
@@ -30,14 +30,53 @@ local ngx_decode_base64 = ngx.decode_base64
...
+    end
+
+    core.log.debug("parsing cert for sni: ", sni)
+
+    local parsed, err = ngx_ssl.parse_pem_cert(cert)
+    if not parsed then
+        return nil, err
+    end
+
+    parsed_certs[cert] = parsed
```
membphis, line 55: bad way, need to use `lrucache`

membphis (changes requested):

tokers:
@membphis It's really a problem if the number of certs is large, I will refactor it with the LRU cache.

tokers:
@membphis way changed.

membphis (approved):

membphis:
> @membphis changed.

approved, need one more `approved`, then your PR can be merged ^_^

moonming (approved):

PR #2168, opened 2020-09-04
bugfix: clear all cached object when got `delete` action, and removing stale boolean data in table `self.values` with a safer way

membphis:

If the administrator user creates and deletes the same route multiple times, the new route will not be correctly synchronized to the APISIX gateway node, resulting in some request route matching failures and a 404 response.

bugfix: clear all cached data when get delete action, and removing stale boolean data in table self.values with a safer way

### Pre-submission checklist:

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -14,6 +14,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+
+local table        = require("apisix.core.table")
 local config_local = require("apisix.core.config_local")
 local log          = require("apisix.core.log")
 local json         = require("apisix.core.json")
@@ -338,10 +340,13 @@ local function sync_data(self)
                 res.value.id = key
                 self.values[pre_index] = res
                 res.clean_handlers = {}
+                log.info("update data by key: ", key)
 
             else
                 self.sync_times = self.sync_times + 1
                 self.values[pre_index] = false
+                self.values_hash[key] = nil
+                log.info("delete data by key: ", key)
             end
 
         elseif res.value then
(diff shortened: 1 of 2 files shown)


t/node/route-delete.t
```diff
@@ -0,0 +1,93 @@
...
+--- no_error_log
+[error]
+--- grep_error_log eval
+qr/\w+ (data by key: 126)/
+--- grep_error_log_out
+insert data by key: 126
+delete data by key: 126
+insert data by key: 126
+delete data by key: 126
+--- timeout: 20
```
nic-chen, line 93: has some more test cases  ?

apisix/core/config_etcd.lua
```diff
@@ -314,36 +316,44 @@ local function sync_data(self)
...
     end
 
     -- avoid space waste
     -- todo: need to cover this path, it is important.
     if self.sync_times > 100 then
-        local count = 0
-        for i = 1, #self.values do
-            local val = self.values[i]
-            self.values[i] = nil
+        local values_org = table.clone(self.values)
```
moonming, line 339: What does this suffix mean? `_org`

apisix/core/config_etcd.lua
```diff
@@ -314,36 +316,44 @@ local function sync_data(self)
...
     elseif res.value then
         res.clean_handlers = {}
         insert_tab(self.values, res)
         self.values_hash[key] = #self.values
         res.value.id = key
+        log.info("insert data by key: ", key)
     end
 
     -- avoid space waste
     -- todo: need to cover this path, it is important.
```
moonming, line 337: Is it still a TODO?

t/node/route-delete.t
moonming: not cover https://github.com/apache/apisix/pull/2168/files#diff-f2239cc9d5659a13ba363ebc30959916R319

apisix/core/config_etcd.lua
```diff
@@ -314,36 +316,44 @@ local function sync_data(self)
...
     if self.sync_times > 100 then
-        local count = 0
-        for i = 1, #self.values do
-            local val = self.values[i]
-            self.values[i] = nil
+        local values_org = table.clone(self.values)
+        table.clear(self.values)
+
+        for i = 1, #values_org do
```
moonming, line 342: Why do we need to delete nil elements in the array? We can make judgments in this for loop：https://github.com/apache/apisix/pull/2168/files#diff-f2239cc9d5659a13ba363ebc30959916R352

apisix/core/config_etcd.lua
membphis: we can remove this comment now. we have added test cases about it.

apisix/core/config_etcd.lua
membphis: We don't want the "value" table to grow larger and larger, which will cause memory overflow.

Therefore, we need to delete the useless items in the "value" to keep it at an acceptable size.

apisix/core/config_etcd.lua
membphis: `original`

t/node/route-delete.t
membphis: fixed

apisix/core/config_etcd.lua
moonming: so `values_original`

moonming: `org` means organization

nic-chen (approved):

moonming (approved):

PR #2174, opened 2020-09-07
optimize set nginx conf env

dickens7:

I need to get env variables in the plugin so I need to set nginx conf environment variables

### Pre-submission checklist:

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -116,7 +116,7 @@ nginx_config:                     # config for render the template to genarate n
   event:
     worker_connections: 10620
   envs:
-    - TEST
+    - APISIX_PROFILE
   http:
     access_log: "logs/access.log"
     access_log_format: "$remote_addr - $remote_user [$time_local] $http_host \"$request\" $status $body_bytes_sent $request_time \"$http_referer\" \"$http_user_agent\" $upstream_addr $upstream_status $upstream_response_time"
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -115,6 +115,8 @@ nginx_config:                     # config for render the template to genarate n
   worker_shutdown_timeout: 240s     # timeout for a graceful shutdown of worker processes
   event:
     worker_connections: 10620
+  envs:
+    - TEST
   http:
     access_log: "logs/access.log"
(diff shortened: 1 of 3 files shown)


conf/config-default.yaml
```diff
@@ -115,6 +115,8 @@ nginx_config:                     # config for render the template to genarate n
   worker_shutdown_timeout: 240s     # timeout for a graceful shutdown of worker processes
   event:
     worker_connections: 10620
+  envs:
+    - TEST
```
spacewander, line 119: We can add `APISIX_PROFILE` as the default.

spacewander:
Need to add test like this: https://github.com/apache/apisix/blob/master/.travis/apisix_cli_test.sh#L58

conf/config-default.yaml
spacewander: I changed my idea. We should always include `APISIX_PROFILE`.
Since the `TEST` is useless now, we should comment it out and just let it as an example.

conf/config-default.yaml
```diff
@@ -115,6 +115,8 @@ nginx_config:                     # config for render the template to genarate n
   worker_shutdown_timeout: 240s     # timeout for a graceful shutdown of worker processes
   event:
     worker_connections: 10620
+  envs:
+    - APISIX_PROFILE
```
spacewander, line 119: Since default configure can be overridden by the user, I think we should always include `APISIX_PROFILE` in the nginx.conf template. Otherwise, `APISIX_PROFILE` might be lost by mistake. We can leave a comment here as example.

spacewander (changes requested):

conf/config-default.yaml
dickens7: From the definition of APISIX_PROFILE, you are right. But I think this configuration should be defined in the configuration file instead of environment variables. I think can use the environment variable definition part of the configuration. When there is a corresponding environment variable definition, the environment variable is preferred

dickens7: the above are my thoughts. i modifyd a version first will not cover `APISIX_PROFILE`

spacewander (approved):

membphis (approved):

PR #2179, opened 2020-09-07
bugfix: supported the encryption/decryption for multi priv keys

tokers:

Previously I added the support for setting multiple certs and pkeys for a single domain, but the encryption for private keys  was forgotten, so here i fixed it.

### Pre-submission checklist:

### DIFF
--- a/apisix/http/router/radixtree_sni.lua
+++ b/apisix/http/router/radixtree_sni.lua
@@ -62,6 +62,20 @@ local function parse_pem_priv_key(sni, pkey)
 end
 
 
+local function decrypt_priv_pkey(iv, key)
+    if core.string.has_prefix(key, "---") then
+        return key
+    end
+
+    local decrypted = iv:decrypt(ngx_decode_base64(key))
+    if decrypted then
+        return decrypted
+    end
+
+    core.log.error("decrypt ssl key failed. key[", key, "] ")
+end
+
+
 local function create_router(ssl_items)
     local ssl_items = ssl_items or {}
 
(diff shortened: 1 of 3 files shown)


moonming:
@tokers please rebase master for this PR, thx

tokers:
@moonming OK.

tokers:
@moonming Rebased.

apisix/http/router/radixtree_sni.lua
```diff
@@ -62,6 +62,20 @@ local function parse_pem_priv_key(sni, pkey)
 end
 
 
+local function decrypt_priv_pkey(iv, key)
+    if str_find(key, "---") then
```
membphis, line 66: should we use `core.string.has_prefix` here?

apisix/http/router/radixtree_sni.lua
```diff
@@ -95,14 +109,22 @@ local function create_router(ssl_items)
...
 
             -- decrypt private key
-            if aes_128_cbc_with_iv ~= nil and
-                not core.string.has_prefix(ssl.value.key, "---") then
-                local decrypted = aes_128_cbc_with_iv:decrypt(ngx_decode_base64(ssl.value.key))
-                if decrypted == nil then
-                    core.log.error("decrypt ssl key failed. key[", ssl.value.key, "] ")
-                else
+            if aes_128_cbc_with_iv ~= nil then
+                local decrypted = decrypt_priv_pkey(aes_128_cbc_with_iv,
```
membphis, line 113: the `ssl.value.key` maybe nil. how about this code style?

```lua
if ssl.value.key then
   ...
end

if ssl.value.keys then
    ...
end
```

membphis (changes requested):
do we need to add some new test cases?

apisix/http/router/radixtree_sni.lua
tokers: sure.

apisix/http/router/radixtree_sni.lua
tokers: OK, actually i just copy and paste the original codes.

tokers:
> do we need to add some new test cases?

Will supply some test cases.

membphis (approved):

moonming (approved):

Issue #2182, opened 2020-09-08
bug: error in the pod when deploying apisix-gw-config-cm.yaml in kubernetes


maxluo1992:
### Issue description
根据本项目中kubernetes的部署步骤操作，在部署apisix-gw-config-cm.yaml发现pod内报错
/usr/local/openresty/luajit/bin/luajit: /usr/bin/apisix:701: failed to read `apisix` field from yaml file
stack traceback:
        [C]: in function 'error'
        /usr/bin/apisix:701: in function </usr/bin/apisix:668>
        /usr/bin/apisix:922: in main chunk
        [C]: at 0x00404c00

### Environment

* apisix version (cmd: `apisix version`):
* OS:
Centos7
二进制安装的K8S1.18.3
etcd是部署在K8S集群外的

moonming:
@maxluo1992 Please use English in the public channel, thx
@gxthrj please take a look, thx

maxluo1992:
OK




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Wen Ming"<notifications@github.com&gt;; 
发送时间: 2020年9月8日(星期二) 中午12:04
收件人: "apache/apisix"<apisix@noreply.github.com&gt;; 
抄送: "Max Luo"<963679535@qq.com&gt;; "Mention"<mention@noreply.github.com&gt;; 
主题: Re: [apache/apisix] request help:  (#2182)





 
@ maxluo1992请在公共频道上使用英语，thx 
 @gxthrj请看一下，thx
 
— 
您收到此邮件是因为有人提到您。
直接回复此电子邮件，在GitHub上查看，或取消订阅。

nic-chen:
@maxluo1992  please take a look at your local `apisix-gw-config-cm.yaml`, what is the content in it ?

maxluo1992:
This is config.yaml file content.
apiVersion: v1
data:
  config.yaml: |
kind: ConfigMap
metadata:
  name: apisix-gw-config.yaml
  namespace: default
dns_resolver:
  - 10.244.3.14

maxluo1992:
This is config.yaml file content:
```yaml
apiVersion: v1
data:
  config.yaml: |
kind: ConfigMap
metadata:
  name: apisix-gw-config.yaml
  namespace: default
dns_resolver:
  10.244.3.14

tokers:
@maxluo1992 
[...]

PR #2185, opened 2020-09-08
fix: change the config file when deploying apisix in k8s #2182

gxthrj:
### Pre-submission checklist:


kubernetes/README.md
```diff
@@ -57,11 +48,15 @@ etcd:
 
 #### Create configmap for apache apisix
 
+you will change some config by use `apisix-gw-config-cm.yaml`
```
spacewander, line 51: It seems that we can always create configmap `apisix-gw-config.yaml` via `config.yaml`.
If you use the default one, you can use `config.yaml`.
If you need to change the default, you can change the `config.yaml`, so the behavior is consistent between k8s and normal installation.
By this way, we can ensure only one place hold the customized configure (for instance, only one place to change the etcd address).

gxthrj: yes, you are right

spacewander (approved):

nic-chen (approved):

Issue #2196, opened 2020-09-09
feature(kafka-logger):  need to read the request body data if it was cached to local file


membphis:
https://github.com/apache/apisix/blob/master/apisix/utils/log-util.lua#L67

The file name is not enough.
we need to read the file content.

swayamraina:
Hi @membphis 
I would like to pick this up. Can you describe what exactly is required here?

membphis:
https://github.com/apache/apisix/blob/master/apisix/utils/log-util.lua#L67

```lua
local body_file = ngx.req.get_body_file()
if body_file then
    -- we only readed the file name of request body, it is not enough.
    -- we need to read the file content by file name.
    log.request.body_file = body_file
end
```

for how to read file content by file name, you can take a look at this: https://stackoverflow.com/questions/11201262/how-to-read-data-from-a-file-in-lua

membphis:
@swayamraina I have assigned this issue to you, many thx

swayamraina:
@membphis  I got what exactly is required here and will raise a PR.
I had one question though, why would request body data get cached to a local file and be absent in the request body

sshniro:
@membphis any idea how to create this scenario? Where the body is kept in a local file? Does it occur if the data is too big?

membphis:
> Does it occur if the data is too big?

you are right. for more detail information, you can take a look at: https://github.com/openresty/lua-nginx-module#ngxreqread_body

I think that if the request body is big, we can only read the first 1 MB of the request body data.

spacewander:
@membphis 
Is this feature still required?

membphis:
> @membphis Is this feature still required?

yes, we need this feature.

spacewander:
I don't think it is a good idea. As the disk operation is expensive and it only happens when the body is too large. Reading truncated body is not so helpful in the log.

tzssangglass:
Would it be better to adjust the parameters to avoid the request being logged in the file?

tokers:
> Would it be better to adjust the parameters to avoid the request being logged in the file?

There has a trade-off between the memory usage and the disk operations, but it should be decided by the user.

membphis:
Hope to hear more users' voices.

PR #2221, opened 2020-09-11
send file contents and not file name

swayamraina:
implements https://github.com/apache/apisix/issues/2196

### DIFF
--- a/apisix/utils/log-util.lua
+++ b/apisix/utils/log-util.lua
@@ -16,6 +16,8 @@
 --
 local core     = require("apisix.core")
 
+local open = io.open
+
 local _M = {}
 
 local function get_full_log(ngx, conf)
@@ -64,11 +66,10 @@ local function get_full_log(ngx, conf)
         else
             local body_file = ngx.req.get_body_file()
             if body_file then
-                local lines = {}
-                for line in io.lines(file) do
-                    lines[#lines + 1] = line
-                end
-                log.request.body_file = lines
+                local file = open(body_file, "rb")
+                local content = file:read "*a"
+                file:close()
+                log.request.body_file = content
             end
         end
     end
--- a/apisix/utils/log-util.lua
+++ b/apisix/utils/log-util.lua
@@ -64,7 +64,11 @@ local function get_full_log(ngx, conf)
         else
             local body_file = ngx.req.get_body_file()
             if body_file then
-                log.request.body_file = body_file
+                local lines = {}
+                for line in io.lines(file) do
+                    lines[#lines + 1] = line
+                end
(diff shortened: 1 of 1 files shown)


apisix/utils/log-util.lua
```diff
@@ -64,7 +64,11 @@ local function get_full_log(ngx, conf)
         else
             local body_file = ngx.req.get_body_file()
             if body_file then
-                log.request.body_file = body_file
+                local lines = {}
```
membphis, line 67: https://stackoverflow.com/questions/11201262/how-to-read-data-from-a-file-in-lua

Read the file content in this way, is the right way:

```lua
local open = io.open

local function read_file(path)
    local file = open(path, "rb") -- r read mode and b binary mode
    if not file then return nil end
    local content = file:read "*a" -- *a or *all reads the whole file
    file:close()
    return content
end

local fileContent = read_file("foo.html");
print (fileContent);
```

swayamraina: `if not file then return nil end`
do we need this part of the code? if the body is not present then can we not assume the data will be present in the file?

moonming:
ping @membphis

apisix/utils/log-util.lua
```diff
@@ -64,7 +66,10 @@ local function get_full_log(ngx, conf)
         else
             local body_file = ngx.req.get_body_file()
             if body_file then
-                log.request.body_file = body_file
+                local file = open(body_file, "rb")
+                local content = file:read "*a"
+                file:close()
+                log.request.body_file = content
```
spacewander, line 72: We should use a new field to store the content to avoid break change.

apisix/utils/log-util.lua
```diff
@@ -64,7 +66,10 @@ local function get_full_log(ngx, conf)
         else
             local body_file = ngx.req.get_body_file()
             if body_file then
-                log.request.body_file = body_file
+                local file = open(body_file, "rb")
```
spacewander, line 69: Need to check if `open` is successful to handle something like disk problem.

apisix/utils/log-util.lua
```diff
@@ -64,7 +66,10 @@ local function get_full_log(ngx, conf)
...
+                local content = file:read "*a"
```
spacewander, line 70: Need to check if `read` is successful like `local content, err = file:read "*a"`

spacewander (changes requested):

Yiyiyimu:
Hi @swayamraina could you have a look at the comments

tokers:
Better to add some cases to cover it.

PR #2222, opened 2020-09-14
fix: return ok in kafka-logger

dotSlashLu:

https://github.com/apache/apisix/blob/98da870976a4893a87b31300276c8c548dac14c0/apisix/utils/batch-processor.lua#L59 needs `ok` and `err`, but `func` in `kafka-logger` returns nil if it's ok. This behaviour causes a lot of unnecessary logs though it's actually ok:

```
2020/09/14 07:29:38 [error] 67351#0: *180881201 [lua] batch-processor.lua:61: Batch Processor[kafka logger] failed to process entries: nil, context: ngx.timer, client: 10.100.2.45, server: 0.0.0.0:443
2020/09/14 07:29:38 [error] 67351#0: *180881201 [lua] batch-processor.lua:68: Batch Processor[kafka logger] exceeded the max_retry_count[1] dropping the entries, context: ngx.timer, client: 10.100.2.45, server: 0.0.0.0:443
```

### Pre-submission checklist:


moonming (approved):

ShiningRush:
This bug should result in an error for each request, then test case 5 should fail, proving that the case does not cover the situation, I guess it should be caused by the short waiting time, please modify the case to cover the bug

ShiningRush (changes requested):
This bug should result in an error for each request, then test case 5 should fail, proving that the case does not cover the situation, I guess it should be caused by the short waiting time, please modify the case to cover the bug

dotSlashLu:
@ShiningRush thanks for your guide. You are right about why the error message was missed before, such a short waiting time will cover all kinds of error messages including kafka timeout and topic missing. 

I've increased the waiting time and also the `batch_max_size`, `batch_max_size` in kafka-logger setups to expose errors more quickly if there are any. Now the reproduce is stable.

Before the change:
```
#   Failed test 'TEST 5: access - pattern "[error]" should not match any line in error.log but matches line "2020/09/17
07:59:38 [error] 2386#0: *20 [lua] batch-processor.lua:61: Batch Processor[kafka logger] failed to process entries: nil, context: ngx.timer, client: 127.0.0.1, server: 0.0.0.0:1984" (req 0)
# '
#   at /usr/local/share/perl/5.28.1/Test/Nginx/Socket.pm line 1280.

#   Failed test 'TEST 5: access - pattern "[error]" should not match any line in error.log but matches line "2020/09/17
07:59:38 [error] 2386#0: *20 [lua] batch-processor.lua:68: Batch Processor[kafka logger] exceeded the max_retry_count[1] dropping the entries, context: ngx.timer, client: 127.0.0.1, server: 0.0.0.0:1984" (req 0)
# '
#   at /usr/local/share/perl/5.28.1/Test/Nginx/Socket.pm line 1280.
```

moonming:
@ShiningRush please take a look, thx

ShiningRush (approved):

PR #2092, opened 2020-08-20
bug: removed default access token for Admin API.

membphis:
### What this PR does / why we need it:

as title

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [x] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [x] Is this PR backward compatible?

PR #2230, opened 2020-09-15
bug: removed default access token for Admin API

membphis:

related PR:  #2092

### Pre-submission checklist:


moonming (changes requested):
Users will encounter errors when running `apisix start`, and this experience is bad.

FAQ.md
```diff
@@ -80,7 +80,7 @@ An example, `foo.com/product/index.html?id=204&page=2`, gray release based on `i
 
 here is the way:
 ```shell
-curl -i http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
+curl -i http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: ******' -X PUT -d '
```
moonming, line 83: `******` is not a good hint for user.

moonming (changes requested):

membphis:
> Users will encounter errors when running `apisix start`, and this experience is bad.

I think it is enough for now. We have told the user how to generate the Admin API key.

```shell
./bin/apisix init
ERROR: missing valid apisix.admin_key

You can call `./bin/apisix gen_admin_key` to generate a new Admin API key or
manually update the `conf/config.yaml` file.
```

If you prefer to automatically generate tokens, we can fix it with another new PR maybe.

We need to discuss on the mailing list whether to automatically generate tokens for users when APISIX starts.

The main purpose of this PR is to remove all default tokens. The current PR is already big. I hope this PR will merge as soon as possible instead of being blocked here.

FAQ.md
membphis: what is your advice?

moonming:
YOUR_OWN_API_TOKEN

Thanks,
Ming Wen
Twitter: _WenMing


YuanSheng Wang <notifications@github.com> 于2020年9月15日周二 下午3:09写道：

moonming:
I insist on this, if automatic generation token is not possible, I would
rather not remove the default token

Thanks,
Ming Wen
Twitter: _WenMing


YuanSheng Wang <notifications@github.com> 于2020年9月15日周二 下午3:07写道：

membphis:
> YOUR_OWN_API_TOKEN

that is a bad name, because of the `YOUR_OWN_API_TOKEN` is a valid Admin API key.

the `******`  is a invalid Admin API key.

> valid `config.yaml`

```yaml
apisix:
    admin_key:
        -
            name: "admin"
            key:   YOUR_OWN_API_TOKEN         # set your Admin API Key
            role: admin

```

>  invalid `config.yaml`

```yaml
apisix:
    admin_key:
        -
            name: "admin"
            key:   ******         # set your Admin API Key
            role: admin

```

moonming:
why `******` is invalid? Is this hard-coded in the code?

Thanks,
Ming Wen
Twitter: _WenMing


YuanSheng Wang <notifications@github.com> 于2020年9月15日周二 下午3:15写道：

membphis:
> why `******` is invalid? Is this hard-coded in the code?

you can take a look at this source code in CLI:

https://github.com/apache/apisix/pull/2230/files#diff-62edadffc237f13dc28a694080d293a8R668

bin/apisix
```diff
@@ -643,6 +643,34 @@ local function read_yaml_conf()
...
+        end
+
+        for _, admin in ipairs(default_conf.apisix.admin_key) do
+            if type(admin.key) == "table" then
+                admin.key = ""
+            else
+                admin.key = tostring(admin.key)
+            end
+
+            if admin.key == "" or admin.key:gsub("*", "") == "" then
```
moonming, line 668: `*` is invalid, which is too shadowy

membphis: In APISIX, the Admin API Key containing only `*` is invalid.
It is shadowy, but it works.

If we use `YOUR_OWN_API_TOKEN` as Key, then we have to disable it here too.

membphis:
> I insist on this, if automatic generation token is not possible, I would rather not remove the default token

I think they are two different questions. Let us discuss it in mail list.

moonming:
No, `YOUR_OWN_API_TOKEN` is just hint in doc, not api token itself.

Thanks,
Ming Wen
Twitter: _WenMing


YuanSheng Wang <notifications@github.com> 于2020年9月15日周二 下午4:29写道：

membphis:
> No, `YOUR_OWN_API_TOKEN` is just hint in doc, not api token itself.

How about we use `YOUR_OWN_API_TOKEN` as the default token? 

The APISIX will show a WARNING message when APISIX started, prompt the user to customize a new token if the user does not set a new one.

moonming:
I don't think so, `YOUR_OWN_API_TOKEN` is good as hint, not the real token.
we can keep the current token

membphis:
> I don't think so, `YOUR_OWN_API_TOKEN` is good as hint, not the real token.
> we can keep the current token

you are right.

the current way, if the APISIX starts with the default Admin API Key, we'll show a WARNING message.
and I updated the default token to a different one.

```
$ ./bin/apisix start
WARNING: using the default Key is very dangerous.

You can call `./bin/apisix gen_admin_key` to generate a new Admin API key or
manually update the `conf/config.yaml` file.
```

moonming:
In this pr, we only need to modify one file, and only need to make one change, which is to print a warning when the Admin IP is not 127.0.0.1 and using default token

membphis:
> In this pr, we only need to modify one file, and only need to make one change, which is to print a warning when the Admin IP is not 127.0.0.1 and using default token

will create a new PR in this way. we can close this PR first.

PR #2233, opened 2020-09-15
change: check etcd cluster version when init_etcd

tokers:
Signed-off-by: tokers <zchao1995@gmail.com>

### Pre-submission checklist:

### DIFF
--- a/bin/apisix
+++ b/bin/apisix
@@ -21,9 +21,9 @@ local function trim(s)
     return (s:gsub("^%s*(.-)%s*$", "%1"))
 end
 
--- Note: The `excute_cmd` return value will have a line break at the end,
+-- Note: The `execute_cmd` return value will have a line break at the end,
 -- it is recommended to use the `trim` function to handle the return value.
-local function excute_cmd(cmd)
+local function execute_cmd(cmd)
     local t, err = io.popen(cmd)
     if not t then
         return nil, "failed to execute command: " .. cmd .. ", error info:" .. err
@@ -33,7 +33,7 @@ local function excute_cmd(cmd)
     return data
 end
 
-excute_cmd("install -d -m 777 /tmp/apisix_cores/")
+execute_cmd("install -d -m 777 /tmp/apisix_cores/")
 
 local pkg_cpath_org = package.cpath
 local pkg_path_org = package.path
@@ -42,12 +42,14 @@ local apisix_home = "/usr/local/apisix"
 local pkg_cpath = apisix_home .. "/deps/lib64/lua/5.1/?.so;"
                   .. apisix_home .. "/deps/lib/lua/5.1/?.so;;"
 local pkg_path  = apisix_home ..    "/deps/share/lua/5.1/?.lua;;"
+local min_etcd_version = "3.4.0"
+
 
 -- only for developer, use current folder as working space
 local is_root_path = false
 local script_path = arg[0]
 if script_path:sub(1, 2) == './' then
-    apisix_home = trim(excute_cmd("pwd"))
+    apisix_home = trim(execute_cmd("pwd"))
     if not apisix_home then
         error("failed to fetch current path")
     end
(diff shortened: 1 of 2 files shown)


moonming:
the etcd in CI is 3.2, so failed: https://github.com/apache/apisix/pull/2233/checks?check_run_id=1116987808#step:6:288

tokers:
@moonming It seems the version selection for etcd is not specified in `./bin/apisix` or other stuffs. Is it under the Github Action's control?

moonming:
we used to be the default version installed by github actions, now it is the etcd 3.4 version installed manually in https://github.com/apache/apisix/pull/2036

tokers:
@moonming OK, let me rebase my branch. By the way, i think it's better to check out the CHANGELOG of etcd, AKAIK there are some bugs in 3.x series releases, but i'm not sure the exact version range.

tokers:
@moonming I found PR  #2036 also checked the etcd version, but IMHO the way is wrong.

Firstly, when we ask the version message from etcd, it returns a JSON string like this:

```json
{"etcdserver":"3.5.0-pre","etcdcluster":"3.5.0"}
```

we should check the version of `etcdcluster`, not the endpoint's server version (`etcdserver`) that we requested, the cluster version is the minimal server version of all endpoints in the cluster. So that's the one we need to check.

Secondly, the semantic version check method doesn't consider the `pre release` situation, although it's rare in prod, i think's better to consider it.

moonming:
nice! we can fix it in this PR :)

Thanks,
Ming Wen
Twitter: _WenMing


Alex Zhang <notifications@github.com> 于2020年9月16日周三 上午10:36写道：

bin/apisix
```diff
@@ -885,20 +939,44 @@ local function init_etcd(show_output)
...
-        cmd = "curl " .. host .. "/version 2>&1"
-        local res = excute_cmd(cmd)
-        local op_ver = str_split(res, "\"")[4]
-        local need_ver = "3.4.0"
-        if not check_version(op_ver, need_ver) then
-            io.stderr:write("etcd version must >=", need_ver, " current ", op_ver, "\n")
+        uri = host .. "/version"
+        local cmd = string.format("curl -s -m %d %s | sed -e 's/[{}]/''/g'", timeout * 2, uri)
+        local res = execute_cmd(cmd)
+        local m = split(res, [[%s*,%s*]])
```
membphis, line 949: I think we can use `dkjson` the parse the JSON response body, it was written by pure Lua code.

https://luarocks.org/modules/dhkolf/dkjson

bin/apisix
```diff
@@ -885,20 +939,44 @@ local function init_etcd(show_output)
...
     for index, host in ipairs(yaml_conf.etcd.host) do
-        -- check if etcd version above 3.4
-        cmd = "curl " .. host .. "/version 2>&1"
-        local res = excute_cmd(cmd)
-        local op_ver = str_split(res, "\"")[4]
-        local need_ver = "3.4.0"
-        if not check_version(op_ver, need_ver) then
-            io.stderr:write("etcd version must >=", need_ver, " current ", op_ver, "\n")
+        uri = host .. "/version"
+        local cmd = string.format("curl -s -m %d %s | sed -e 's/[{}]/''/g'", timeout * 2, uri)
```
membphis, line 947: it depends on the tool `sed`. External tools, the less dependency the better.

bin/apisix
```diff
@@ -885,20 +939,44 @@ local function init_etcd(show_output)
...
+            end
+        end
+
+        if not cluster_version then
+            io.stderr:write(errmsg)
             return
         end
 
+        if compare_semantic_version(cluster_version, min_etcd_version) then
+            io.stderr:write(string.format("etcd cluster version \"%s\" is less than the required version \"%s\", please upgrade your etcd cluster", cluster_version, min_etcd_version))
```
membphis, line 971: this line is too long. it is not easy for reading

bin/apisix
tokers: That'd be better.

bin/apisix
tokers: OK, will use jkjson.

bin/apisix
tokers: OK, will split this line.

tokers:
@membphis Fixed.

membphis (approved):

moonming (approved):

Yiyiyimu:
@tokers Thank you for the fix!

PR #2271, opened 2020-09-21
bugfix: avoid overwriting Vary header in cors plugin

spacewander:
### Pre-submission checklist:


membphis (approved):

moonming (approved):

apisix/core/response.lua
```diff
@@ -110,6 +116,31 @@ function _M.set_header(...)
 end
 
 
+function _M.add_header(...)
```
ShiningRush, line 119: `add_header` is almost same with `set_header`, we should extract common logic to avoid repeat code.

t/lib/server.lua
```diff
@@ -269,6 +278,7 @@ function _M.go()
         return ngx.exit(404)
     end
 
+    inject_headers()
```
ShiningRush, line 281: Why we need change header key `resp-x` header to "x"?

ShiningRush (changes requested):
fix point is no problem, but I have some qutions about code.

t/lib/server.lua
spacewander: @ShiningRush 
It's for test case to control the response header. Note that the type of headers are different.

ShiningRush: Well, got it.

apisix/core/response.lua
spacewander: @ShiningRush 
Updated.

membphis: nice ^_^

moonming:
@ShiningRush please take a look again, thx

ShiningRush:
I have already leave comment that the code about `for` can be more simple. Maybe we should discuss it first?
@moonming @spacewander

moonming:
> I have already leave comment that the code about `for` can be more simple. Maybe we should discuss it first?
> @moonming @spacewander

Is this comment https://github.com/apache/apisix/pull/2271#discussion_r493227848? which was fixed

ShiningRush (commented):
My fault, those comment is in pending status, so only me can see it.
I don't know when github's mechanism becomes like this : (.

apisix/core/response.lua
```diff
@@ -97,19 +103,43 @@ function _M.set_header(...)
...
-            ngx_header[k] = v
+        if append then
+            for k, v in pairs(headers) do
+                ngx_add_header(k, v)
+            end
+
+        else
+            for k, v in pairs(headers) do
+                ngx_header[k] = v
+            end
```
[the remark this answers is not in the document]
moonming: I don't think so,  it will run `if append` many times.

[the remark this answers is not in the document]
ShiningRush: In theory, the comparison of boolean values ​​will not cause significant performance loss, but I have not verified this in Lua.
We can get two benefits by extracting the `for`:
-Better readability: less code.
-Better scalability: If there are other conditional parameters to be added in addition to append in the future, the amount of modified code is less

It was mentioned in [Refactoring](https://martinfowler.com/books/refactoring.html) that we should give priority to readability and wait until performance bottlenecks occur before targeted optimization.
However, we can also keep the existing code and modify it until the next modification, in exchange for a slight performance improvement.
@membphis @spacewander How dow you think?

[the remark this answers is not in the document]
spacewander: @ShiningRush 
Updated in the new commit.

ShiningRush (approved):

PR #2273, opened 2020-09-22
test: add integration test

idbeta:
Add some common scenarios‘s test cases,this is to add fuzz tests later
scenario 1： 
1.Add the "/hello" route and forward it to the upstream port 9666, generating a 200 response.
2.Generate multiple requests to the route uri in 10s,should get the 200 response
3.Delete the "/hello" route 
4.Generate multiple requests to the route uri in 10s,should get the 404 response
5.In the meantime, monitor resource usage and error logs

### Pre-submission checklist:


t/pycases/cases/nginx.conf
```diff
@@ -0,0 +1,16 @@
...
+error_log error.log;
+events {
+    worker_connections 1024;
+}
+http {
+    server {
+        listen 9000;
+        location / {
+            default_type text/html;
+            content_by_lua '
```
membphis, line 11: `content_by_lua_block` is better

t/pycases/cases/nginx.conf
```diff
@@ -0,0 +1,16 @@
+worker_processes  1;
+error_log error.log;
+events {
+    worker_connections 1024;
```
membphis, line 4: `4096` is better

membphis (commented):
`t/pycases` is not a good name, how about `t/integration` ?

.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-18.04]
-        os_name: [linux_openresty, linux_tengine, linux_apisix_master_luarocks, linux_apisix_current_luarocks, linux_openresty_mtls]
+        os_name: [linux_apisix_integrationtest]
```
membphis, line 15: we can not use this way, you can add a new item `linux_apisix_integrationtest`

idbeta: got it and fixed

t/pycases/cases/nginx.conf
idbeta: got it and fixed

t/pycases/cases/nginx.conf
idbeta: got it and fixed

idbeta:
> `t/pycases` is not a good name, how about `t/integration` ?

got it and fixed

spacewander:
The `t/integration` is confusing. Actually almost `*.t` files are  for integration test, and only a few of them are unit tests.
Is there any different between the Python test and the existed `*.t` test case?

And you can't say a test which simply hit a route a hundred times is a fuzzy test. It only touches a few area and have a poor coverage.

I think before we accept any codes we should do some research and write down a plan about how to add a fuzzy framework.

spacewander:
The `t/integration` is confusing. Actually almost `*.t` files are  for integration test, and only a few of them are unit tests.
Is there any different between the Python test and the existed `*.t` test case?

And you can't say a test which simply hit a route a hundred times is a fuzzy test. It only touches a few area and have a poor coverage.

I think before we accept any codes we should do some research and write down a plan about how to add a fuzzy framework.

idbeta:
> The `t/integration` is confusing. Actually almost `*.t` files are for integration test, and only a few of them are unit tests.
> Is there any different between the Python test and the existed `*.t` test case?
> 
> And you can't say a test which simply hit a route a hundred times is a fuzzy test. It only touches a few area and have a poor coverage.
> 
> I think before we accept any codes we should do some research and write down a plan about how to add a fuzzy framework.

thank you for your reply,the name is indeed a bit confusing, I will think about what it is better. This PR provides only a test case for a common scenario, not a fuzz test case. It is to test the feasibility of using the pytest framework to test apisix, and future fuzzing tests will be carried out based on this method.

idbeta:
When users use API​​SIX, some problems such as setting too many routes to cause high memory usage after *long time* running, or problems caused by abnormal operation of related components in the APISIX service.
These problems may take *long* testing to discover, and need to obtain related system resource occupation.
Our current test method cannot cover these scenarios well. so we can try another test method to cover these scenarios.

membphis:
> and need to obtain related system resource occupation

Only "relevant system resources" is not enough, we need to confirm the status of the APISIX instance before and after the test. For example, we need to check the used memory, network stats, etc. They are useful.

moonming:
Is this pr solving possible problems that might exist or is it solving real problems?

membphis:
ping @idbeta

idbeta:
There are too many questions about this pr, close it, let me create another pr,thank u all.

PR #2286, opened 2020-09-23
feat(http-logger): support to concat multiple log with separator.

membphis:

as title.

support to report log in this way:

```shell
curl http://****/log -d '
{xxxxxxxxxxxxx}
{yyyyyyyyyyyyy}
'
```

the request logs is concat by '\n' way.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/http-logger.lua
+++ b/apisix/plugins/http-logger.lua
@@ -113,13 +113,6 @@ local function send_http_data(conf, log_message)
             .. "body[" .. httpc_res:read_body() .. "]"
     end
 
-    -- keep the connection alive
-    ok, err = httpc:set_keepalive(conf.keepalive)
-
-    if not ok then
-        core.log.debug("failed to keep the connection alive", err)
-    end
-
     return res, err_msg
 end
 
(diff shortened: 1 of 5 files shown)


nic-chen:
I have a doubt, why not use a array json, but concat by '\n' ?

nic-chen:
> I have a doubt, why not use a array json, but concat by '\n' ?

ignore that, I see it supports two ways..

apisix/plugins/http-logger.lua
```diff
@@ -110,13 +113,6 @@ local function send_http_data(conf, log_message)
             .. "body[" .. httpc_res:read_body() .. "]"
     end
 
-    -- keep the connection alive
-    ok, err = httpc:set_keepalive(conf.keepalive)
```
nic-chen, line 114: is it a default action ? although `keepalive` is not in the plugin's schema.

membphis: the old code is wrong. that is why I delete them.

membphis: if we want to call `set_keepalive`, need to define them first.

gxthrj (approved):

nic-chen (approved):

PR #2290, opened 2020-09-23
fix: increase wait to avoid fail

ShiningRush:
fix: increase wait time to avoid fail

### DIFF
--- a/t/plugin/tcp-logger.t
+++ b/t/plugin/tcp-logger.t
@@ -233,4 +233,4 @@ GET /t
 --- error_log
 failed to connect to TCP server: host[312.0.0.1] port[2000]
 [error]
---- wait: 1.5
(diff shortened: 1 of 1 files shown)


moonming (approved):

membphis (approved):

sshniro:
I'm a bit curious, on which environments/scenarios does this fail? 
Because other loggers also have similar test cases.

ShiningRush:
I have encountered similar problems before( this [PR](https://github.com/apache/apisix/pull/2127/files) ).
My scenarios is that I use a 4c8g server as CI for APISIX, this case will sometimes fail because timeout. the server is owned by company, maybe some proxy process intercept the traffic so increase the delay or just because it's performance is too low : )

PR #2294, opened 2020-09-23
feat(http-logger): support for specified the log formats via admin API

membphis:

as title.

For example:
```
curl http://****/apisix/admin/plugin_metadata/http-logger -d '
{
    "log_format": {
        "host": "$host",
        "@timestamp": "$time_iso8601",
        "client_ip": "$remote_addr"
    }
}'
```

when we enabled plugin `http-logger`, we will get the message body like:

```
{"host":"localhost","@timestamp":"2020-09-23T18:29:07-04:00","client_ip":"127.0.0.1","route_id":"1"}
{"host":"localhost","@timestamp":"2020-09-23T18:29:07-04:00","client_ip":"127.0.0.1","route_id":"1"}
```

### Pre-submission checklist:


nic-chen (approved):

moonming:
why not add `log_format` in http-logger plugin directly?

moonming:
Different routes may have different log formats

bin/apisix
```diff
@@ -1025,7 +1025,7 @@ local function init_etcd(show_output)
         for _, dir_name in ipairs({"/routes", "/upstreams", "/services",
                                    "/plugins", "/consumers", "/node_status",
                                    "/ssl", "/global_rules", "/stream_routes",
-                                   "/proto"}) do
+                                   "/proto", "/plugin_metadata"}) do
```
moonming, line 1028: What is the difference between metadata and the properties of plugin itself?  I didn't see why we need it.

membphis: get plugin schema: 
/apisix/admin/plugins/{plugin-name}, eg: /apisix/admin/plugins/limit-req

get plugin metadata:
/apisix/admin/plugin_metadata/{plugin-name}, eg: /apisix/admin/plugin_metadata/example-plugin

moonming: https://github.com/apache/apisix/pull/2294#issuecomment-698108183

moonming: why we need this PR?

juzhiyuan: I would recommend using `/plugins/{name}/metadat` in a RESTful way if this PR is needed..

ShiningRush: I am also curious about what scenarios need metadata instead of plugin's option.

membphis: this PR has been merged. we can talk in here: https://github.com/apache/apisix/pull/2307#issuecomment-698337991

PR #2299, opened 2020-09-24
bugfix: use etcd:watchcancel to reduce connections in etcdv3

### LINKED ISSUE #2295: request help: Too many connections when using etcdv3
### Issue description
using `etcd v3` find too many connections.
I deployed an apisix and started 4 workers, Up to 300+ connections

image

### Environment

* apisix version : 7beafd345a96b5fc14d9230731f284429ad7168e
* OS: Centos Linux release 7.4.1708 (Core)
image

Yiyiyimu:
fix #2295 

### Pre-submission checklist:


apisix/core/config_etcd.lua
```diff
@@ -123,6 +124,11 @@ local function waitdir(etcd_cli, key, modified_index, timeout)
         res, err = res_func()
     end
 
+    if http_cli then
+        etcd_cli:watchcancel(http_cli)
```
membphis, line 128: need test case.

we can use this way: write a log and confirm it in test case

apisix/core/config_etcd.lua
```diff
@@ -246,24 +248,9 @@ local function sync_data(self)
...
     end
 
-    -- for fetch the etcd index
-    local key_res, _ = getkey(self.etcd_cli, self.key)
-
     local dir_res, err = waitdir(self.etcd_cli, self.key, self.prev_index + 1, self.timeout)
-
     log.info("waitdir key: ", self.key, " prev_index: ", self.prev_index + 1)
     log.info("res: ", json.delay_encode(dir_res, true))
-    if err == "timeout" then
```
membphis, line 256: why we delete this code block? it is useful for etcd `v2`. please confirm this

Yiyiyimu: Because etcd v3 does not have 1000 event history limit anymore, it could support unlimited history.

apisix/core/config_etcd.lua
```diff
@@ -181,10 +181,6 @@ local function sync_data(self)
             return false, err
         end
 
-        if not dir_res.dir then
-            return false, self.key .. " is not a dir"
-        end
-
```
Yiyiyimu, line 181: Currently I still set `dir = true` in [etcd protocol conversion](https://github.com/apache/apisix/blob/9ac02f66b14fa3c1b5a8f9341a82b664ceda17fb/apisix/core/etcd.lua#L61) for consistency between protocols. But I noticed there is only one test case ([global.t TEST 3](https://github.com/apache/apisix/blob/9ac02f66b14fa3c1b5a8f9341a82b664ceda17fb/t/admin/global-rules.t#L125-L127)) needs this parameter, so maybe we could remove it

apisix/core/config_etcd.lua
Yiyiyimu: But we also need to consider history compaction, or the accumulation of history would affect performance in a long term. Reference: https://etcd.io/docs/v3.4.0/op-guide/maintenance/#history-compaction

apisix/core/config_etcd.lua
Yiyiyimu: thx, test added

apisix/core/config_etcd.lua
nic-chen: @membphis  I think we don't need to consider compatibility with v2

membphis: got it. many thanks for explain

Yiyiyimu:
Hi @membphis is there something else we should do before merging

Yiyiyimu:
ping @moonming it seems good to go

idbeta (approved):
I passed in my local environment

apisix/core/etcd.lua
```diff
@@ -36,6 +36,7 @@ local function new()
     etcd_conf.host = nil
     etcd_conf.prefix = nil
     etcd_conf.protocol = "v3"
+    etcd_conf.api_prefix = "/v3"
```
moonming, line 39: Where will it be used?

Yiyiyimu: In lua-resty-etcd. If api_prefix is not set, it would check the version of etcd to find the correct api_prefix to use in uri. Since we support etcd v3.4+ only, we could directly set it to avoid useless version check each time.

moonming (approved):

membphis (approved):

apisix/core/config_etcd.lua
```diff
@@ -123,6 +124,15 @@ local function waitdir(etcd_cli, key, modified_index, timeout)
         res, err = res_func()
     end
 
+    if http_cli then
+        local res_cancel, err_cancel = etcd_cli:watchcancel(http_cli)
+        if res_cancel and res_cancel == 1 then
```
membphis, line 129: `if res_cancel == 1 then` is enough

membphis (commented):
code style

apisix/core/config_etcd.lua
Yiyiyimu: Thx! Should I create a PR for it or leave it to the next PR

moonming: sure,welcome

PR #2305, opened 2020-09-24
fix: kafka-logger plugin sent to the same partition everytime

uptree:
1. fix: kafka-logger plugin sent to the same partition everytime
2. random select partition_id for init kafka producer
3. we can use "TEST 4" and "TEST 5" of kafka-logger.t test it，and number of partitions for a kafka topic >= 2. 

### Pre-submission checklist:


apisix/plugins/kafka-logger.lua
```diff
@@ -82,6 +84,11 @@ local function send_kafka_data(conf, log_message)
     end
 
     broker_config["request_timeout"] = conf.timeout * 1000
+    broker_config["partitioner"] = function(key, num, correlation_id)
```
moonming, line 87: It will be better if you can add test cases to cover.

moonming (changes requested):

apisix/plugins/kafka-logger.lua
uptree: we can use "TEST 4" and "TEST 5" of kafka-logger.t test it，and number of partitions for a kafka topic >= 2.

uptree: > It will be better if you can add test cases to cover.

we can use "TEST 4" and "TEST 5" of kafka-logger.t test it，and number of partitions for a kafka topic >= 2.

PR #2306, opened 2020-09-24
1.fix generation of uuid;

zlm0125:
Change-Id: Iff73ad95f87a81bc66ecc26e9ab250a8ca41ebf6

### Pre-submission checklist:

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -107,6 +107,10 @@ function _M.http_init_worker()
     require("apisix.debug").init_worker()
     require("apisix.upstream").init_worker()
 
+    -- need call seed
+    local uuid = require("resty.jit-uuid")
+    uuid.seed()
+
     local_conf = core.config.local_conf()
     local dns_resolver_valid = local_conf and local_conf.apisix and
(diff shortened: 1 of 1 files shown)


apisix/init.lua
```diff
@@ -107,6 +107,10 @@ function _M.http_init_worker()
     require("apisix.debug").init_worker()
     require("apisix.upstream").init_worker()
 
+    -- need call seed
+    local uuid = require("resty.jit-uuid")
+    uuid.seed()
```
membphis, line 112: https://github.com/apache/apisix/blob/master/apisix/init.lua#L72

We have set the seed with a better way. So we do not need to set it agian.

zlm0125: But in k8s, it not work very well.

membphis: you can print more log for confirming why it works badly. we need to know the reason.

moonming: @ zlm0125  ping

membphis:
resolve this bug by another PR: https://github.com/apache/apisix/pull/2357
I think we can close this PR. @zlm0125

Issue #2257, opened 2020-09-18
bug: http_to_https ret_code 307 not working


TigerGxl:
### Environment

* apisix version (cmd: `apisix version`):1.5
* OS:ubuntu16.04

i use plugin http_to_https and set ret_code 307, like below.
```
{
  "priority": 0,
  "plugins": {
    "redirect": {
      "ret_code": 307,
      "http_to_https": true
    }
  },
  "uris": [
    "/*"
  ],
  "id": "1a5f4337d",
  "hosts": [
    "demo-apisix.xxxxxxx.com"
  ],
  "upstream": {
    "nodes": {
      "172.16.3.2:80": 100
    },
    "hash_on": "vars",
    "type": "roundrobin"
  }
}
```
when client sends **POST** reqeust with **http**.it's not working, apisix should  return 307 and client sends a new request with https again, in fact, apisix return 301 and client sends a new request with **https** and the method is **GET**

![image](https://user-images.githubusercontent.com/32395903/93565351-cce64680-f9bd-11ea-9349-80aaaacda6f3.png)
![image](https://user-images.githubusercontent.com/32395903/93565479-074fe380-f9be-11ea-953d-c884d6647f9b.png)

TigerGxl:
启用http跳转https后,设置状态码为307不生效,etcd中存储的是301,但是实际跳转的时候是301,导致原本POST的method在301跳转后变为GET

membphis:
that is a bug. here is the test case. welcome PR to fix it

```
=== TEST 1: add plugin with new uri: /test/add
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/routes/1',
                ngx.HTTP_PUT,
                [[{
                    "plugins": {
                        "redirect": {
                            "http_to_https": true,
                            "ret_code": 307
                        }
                    },
                    "uri": "/hello"
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 2: redirect
--- request
GET /hello
--- response_headers
Location: /test/add
--- error_code: 307
--- no_error_log
[error]

membphis:
the patch

```patch
diff --git a/apisix/plugins/redirect.lua b/apisix/plugins/redirect.lua
index 64819735..eb386dea 100644
--- a/apisix/plugins/redirect.lua
[...]

PR #2177, opened 2020-09-07
feat: Support plugin for "aliyun" log service

gy09535:
### What this PR does / why we need it:
Fix:https://github.com/apache/apisix/issues/2169

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [x] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [x] Is this PR backward compatible?

PR #2311, opened 2020-09-25
bugfix: return right response code when `http_to_https` enabled in redirect plugin.

gy09535:
fix: https://github.com/apache/apisix/issues/2257

### Pre-submission checklist:


apisix/plugins/redirect.lua
```diff
@@ -133,7 +135,11 @@ function _M.rewrite(conf, ctx)
         -- TODO： add test case
         -- PR: https://github.com/apache/apisix/pull/1958
         uri = "https://$host$request_uri"
-        ret_code = 301
+        if ngx.req.get_method() == http_post then
+            ret_code = 308
+        else
+            ret_code = 301
+        end
```
liuhengloveyou, line 142: User's configuration priority

`
if not ret_code or ret_code == 302 then
            if ngx.req.get_method() == "POST" then
                ret_code = 308
            else
                ret_code = 301
            end
        end
`

t/plugin/redirect.t
```diff
@@ -676,3 +676,51 @@ close: 1 nil}
...
+=== TEST 27: redirect
+--- request
+POST /hello-https
+--- more_headers
+Host: test.com
+--- response_headers
+Location: https://test.com/hello-https
+--- error_code: 308
+--- no_error_log
+[error]
```
liuhengloveyou, line 726: When the user has a configuration, the value of the user configuration should be returned

apisix/plugins/redirect.lua
```diff
@@ -133,7 +135,11 @@ function _M.rewrite(conf, ctx)
         -- TODO： add test case
         -- PR: https://github.com/apache/apisix/pull/1958
         uri = "https://$host$request_uri"
-        ret_code = 301
+        if ngx.req.get_method() == http_post then
```
spacewander, line 138: I think it would be better to use whitelist (`GET` OR `HEAD`) instead.

apisix/plugins/redirect.lua
gy09535: I think when you choice http to https, you can not use user ret_code ,because http to https is an features, it means you set the redirect 301 but how handle post method , you can set 308, when the http is become https ,the user code is effect. what do you think @liuhengloveyou ?

apisix/plugins/redirect.lua
gy09535: sure

gy09535:
If have times, please help review another PR, https://github.com/apache/apisix/pull/2177 @spacewander @liuhengloveyou

apisix/plugins/redirect.lua
liuhengloveyou: I think,,,user can choose not to configure it。If the user's configuration is wrong, should get the wrong result，
not be ignored。
The role of the user's configuration should be very definite.

spacewander (approved):

apisix/plugins/redirect.lua
gy09535: The http to https is a feature and the  ret_code is other feature, when apisix receive http request should change it to https, when apisix receive https we should redirect it to user ret_code, they are not conflict.

apisix/plugins/redirect.lua
```diff
@@ -133,7 +134,12 @@ function _M.rewrite(conf, ctx)
         -- TODO： add test case
         -- PR: https://github.com/apache/apisix/pull/1958
         uri = "https://$host$request_uri"
-        ret_code = 301
+        local method_name = ngx.req.get_method()
+        if method_name == "GET" or method_name == "HEAD" then
+            ret_code = 301
```
moonming, line 139: need add test cases for these codes.

moonming (changes requested):

moonming (approved):

apisix/plugins/redirect.lua
```diff
@@ -133,7 +134,12 @@ function _M.rewrite(conf, ctx)
         -- TODO： add test case
         -- PR: https://github.com/apache/apisix/pull/1958
         uri = "https://$host$request_uri"
-        ret_code = 301
+        local method_name = ngx.req.get_method()
```
membphis, line 137: It is recommended to put the following link in the comments for easy reading:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308

membphis (approved):

Issue #2310, opened 2020-09-25
bug: failed to match any routes


Firstsawyou:
### Issue description
Before the `etcd` service is closed, the `apisix` request is normal. After the `etcd` service is turned off and turned on again, the `apisix` request reports 404 ({"error_msg":"failed to match any routes"}).
### Environment

* apisix version (cmd: `apisix version`): 1.5
* OS: centos 7

### Minimal test code / Steps to reproduce the issue
1.Add route and configure upstream information

```shell
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "methods": ["GET"],
    "uri": "/index.html",
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:9081": 1
        }
    },
    "desc": "upstream node"
}'
```

2. Access to 'route' is normal

3.Turn off the etcd service and turn it on again

4.Exception in accessing 'route' again

5.After restarting `apisix`, the request returns to normal.

### What's the actual result? (including assertion message & call stack if applicable)

```shell
curl -i http://127.0.0.1:9080/index.html
HTTP/1.1 404 Not Found
Date: Fri, 25 Sep 2020 03:36:10 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Server: APISIX/1.5
{"error_msg":"failed to match any routes"}
```

error log:
```
2020/09/25 10:37:07 [error] 16527#16527: *28 lua entry thread aborted: runtime error: ...r_service-id_route-id/apisix/apisix/core/config_etcd.lua:421: attempt to concatenate local 'err' (a nil value)
stack traceback:
coroutine 0:
        ...r_service-id_route-id/apisix/apisix/core/config_etcd.lua: in function <...r_service-id_route-id/apisix/apisix/core/config_etcd.lua:414>, context: ngx.timer
2020/09/25 10:37:07 [error] 16527#16527: *30 lua entry thread aborted: runtime error: ...r_service-id_route-id/apisix/apisix/core/config_etcd.lua:421: attempt to concatenate local 'err' (a nil value)
stack traceback:
coroutine 0:
        ...r_service-id_route-id/apisix/apisix/core/config_etcd.lua: in function <...r_service-id_route-id/apisix/apisix/core/config_etcd.lua:414>, context: ngx.timer
```

### What's the expected result?
```shell
[...]

PR #2312, opened 2020-09-25
bugfix: create etcd object in `xpcall`, this step may fail

membphis:
fix issue: #2310

1. The old process, if creating etcd fails, etcd data will no longer be synchronized.  We need to create the etcd object in `xpcall`. 
2. the return value should be `res, err` of `etcd.new`.

### Pre-submission checklist:


membphis:
@Yiyiyimu please take a look at this issue, it is a bugfix

Yiyiyimu:
@membphis Sure that's my mistake. Thank you for the fix!!

nic-chen:
I checked the old version. we created a etcd cli only once here: 
https://github.com/apache/apisix/blob/7f64ba810ef4ea9a105f7c3e8a83ac54e47ad58e/apisix/core/config_etcd.lua#L434-L446
and then we could reuse it.

I think we could use that mode.

do you remember that change it? @Yiyiyimu

membphis:
> do you remember that change it? @Yiyiyimu

I know the reason. 

When we create a new etcd v3 object, it depends on `cosocket` feature. so we have to move this code block into `timer` land.

membphis:
I just updated the initialization process and now it only initializes once. https://github.com/apache/apisix/pull/2312/commits/bd2d95038d29593c1b205fc4516768ea94bd4976

nic-chen:
> > do you remember that change it? @Yiyiyimu
> 
> 
> 
> I know the reason. 
> 
> 
> 
> When we create a new etcd v3 object, it depends on `cosocket` feature. so we have to move this code block into `timer` land. it is not supports `cosocket` feature in `init` and `init_worker` phase.

OK

nic-chen (approved):

moonming:
@membphis please choose a better title of this PR

membphis:
> @membphis please choose a better title of this PR

updated

apisix/core/config_etcd.lua
```diff
@@ -416,16 +416,20 @@ local function _automatic_fetch(premature, self)
...
         local ok, err = xpcall(function()
+            if not self.etcd_cli then
+                local etcd_cli, err = etcd.new(self.etcd_conf)
+                if not etcd_cli then
+                    error("failed to create etcd instance for key ["
+                          .. self.key .. "]: " .. (err or "unknown"))
+                end
+                self.etcd_cli = etcd_cli
+            end
+
```
liuhengloveyou, line 432: Why has the number of returned values changed?

membphis: the old code is wrong. the return value is `res, err` for latest version of `resty-etcd`

Yiyiyimu (approved):

PR #2313, opened 2020-09-25
fix: Update redirect state code logic

liuhengloveyou:

#2257 

As I said above.
If the user has explicitly configured ret_code, we should not change it.
Unless the configuration is wrong, we should indicate the range in the documentation.

### Pre-submission checklist:


moonming:
the same as https://github.com/apache/apisix/pull/2311

Issue #2320, opened 2020-09-25
bug: limit-req plugin can not work properly for this case.


membphis:
It seems a bit complicated, we need a smaller test case to find the problem easily.

```
=== TEST 1: create consumer
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/consumers',
                ngx.HTTP_PUT,
                [[{
                    "username": "jack",
                    "plugins": {
                        "key-auth": {
                            "key": "auth-one"
                        }
                    }
                }]]
            )

            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 2: set service
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/services/1',
                ngx.HTTP_PUT,
                [[{
                    "plugins": {
                        "key-auth": {}
                    },
                    "desc": "new service"
                }]]
            )
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 3: create route with plugin `limit-req`(upstream node contain domain)
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/routes/1',
                ngx.HTTP_PUT,
                [[{
                    "plugins": {
                        "limit-req": {
                            "rate": 1,
                            "burst": 0,
                            "rejected_code": 503,
                            "key": "remote_addr"
                        }
                    },
                    "upstream": {
                        "nodes": {
                            "www.baidu.com:80": 1
                        },
                        "pass_host": "node",
                        "type": "roundrobin"
                    },
                    "service_id": 1,
                    "uri": "/index.html"
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 4: make a test
--- config
location /t {
    content_by_lua_block {
        local core = require("apisix.core")
        local t = require("lib.test_admin")
        local headers = {
            ["User-Agent"] = "curl/7.68.0",
            ["apikey"] = "auth-one",
        }

        for i = 1, 3 do
            local code, body = t.test('/index.html',
                ngx.HTTP_GET,
                "",
                nil,
                headers
            )
            ngx.say("return: ", code)
        end
    }
}
[...]

PR #2322, opened 2020-09-25
bugfix: when service works with upstream that contains a domain name, the merged configuration should not always changing.

membphis:

when service works with upstream that contains a domain name, the merged configuration is always changing.
if the configuration is always changing,  it won't hit the LRU item, which causes some plugins to work abnormally, such as `limit-req`.

fix: https://github.com/apache/apisix/issues/2320

### Pre-submission checklist:


t/plugin/key-auth-upstream-domain-node.t
```diff
@@ -0,0 +1,155 @@
...
+                        "limit-req": {
+                            "rate": 1,
+                            "burst": 0,
+                            "rejected_code": 503,
+                            "key": "remote_addr"
+                        }
+                    },
+                    "upstream": {
+                        "nodes": {
+                            "www.baidu.com:80": 1
```
nic-chen, line 100: seems visit `baidu.com` unstable in github actions before?

apisix/init.lua
```diff
@@ -461,23 +455,15 @@ function _M.http_access_phase()
 
     else
         if route.has_domain then
-            local parsed_route, err = lru_resolved_domain(route, api_ctx.conf_version,
-                                        return_direct, nil)
+            local err
+            route, err = lru_resolved_domain(route, api_ctx.conf_version,
+                            parse_domain_in_route, route, api_ctx)
```
gxthrj, line 460: `api_ctx` need to be removed ？

gxthrj (approved):

apisix/init.lua
membphis: nice, will fix it later

membphis:
@moonming this is an old bug for long time, and it seems that I need more time to fix it. 
Therefore, I think it is not necessary for the 2.0 milestone.

moonming:
ok

nic-chen (approved):

spacewander (approved):

PR #2331, opened 2020-09-27
fix(basic-auth): split schema to adapt different occasions

spacewander:
### Pre-submission checklist:


liuhengloveyou:
fix which issues?

spacewander:
The original schema check is too floppy. It will let the invalid schema in `TEST 10: invalid schema, not field given` pass, but when you hit the rule later, you will be in trouble.

membphis:
@ShiningRush I remember we discussed this method before, you can look at this PR

membphis:
this way is much better than the old way ^_^

apisix/core/schema.lua
```diff
@@ -20,7 +20,11 @@ local lrucache = require("apisix.core.lrucache")
 local cached_validator = lrucache.new({count = 1000, ttl = 0})
 local pcall = pcall
 
-local _M = {version = 0.3}
+local _M = {
+    version = 0.3,
+    TYPE_ANY = 1,
```
ShiningRush, line 25: - `TYPE_ANY` is confused, I think `TYPE_CONF` or `TYPE_OPT` would be better, how do you think?
- It seems like the code of checking plugin's config does not using this parameter, are there any special considerations?

ShiningRush (changes requested):

apisix/admin/plugins.lua
```diff
@@ -29,10 +30,16 @@ local table_insert = table.insert
 local _M = {}
 
 
-function _M.check_schema(plugins_conf)
+function _M.check_schema(plugins_conf, schema_type)
     for name, plugin_conf in pairs(plugins_conf) do
         core.log.info("check plugin scheme, name: ", name, ", configurations: ",
                       core.json.delay_encode(plugin_conf, true))
+        if type(plugin_conf) ~= "table" then
```
ShiningRush, line 37: No test case cover this logic.

ShiningRush (changes requested):

apisix/admin/plugins.lua
spacewander: @ShiningRush 
Updated.

ShiningRush (approved):

moonming (approved):

Issue #2297, opened 2020-09-24
feat: support to fetch `consumer_name` or `consumer_id` via `ctx.var`


Firstsawyou:
### Issue description
Hide the problem of obtaining `consumer_id` in the "limit-req" plugin without changing the original plugin code. #2275 

### Environment

* apisix version (cmd: `apisix version`):1.5
* OS:

Firstsawyou:
assigned to me.

PR #2333, opened 2020-09-27
feat: support to fetch `consumer_name` or `consumer_id` via `ctx.var`

Firstsawyou:

 fix: #2297 
### Pre-submission checklist:

### DIFF
--- a/apisix/core/ctx.lua
+++ b/apisix/core/ctx.lua
@@ -28,6 +28,7 @@ local ngx_var      = ngx.var
 local re_gsub      = ngx.re.gsub
 local type         = type
 local error        = error
+local ngx          = ngx
 
 
 ffi.cdef[[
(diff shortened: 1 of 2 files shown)


liuhengloveyou (commented):
good

apisix/core/ctx.lua
```diff
@@ -96,6 +96,9 @@ do
             elseif key == "service_id" then
                 val = ngx.ctx.api_ctx and ngx.ctx.api_ctx.service_id
 
+            elseif key == "consumer_name" then
+                val = ngx.ctx.api_ctx and ngx.ctx.api_ctx.consumer_id
```
moonming, line 100: It always confuses me, `name` or `id`? we should choose a good name

moonming (changes requested):

apisix/core/ctx.lua
Firstsawyou: `consumer_id` and `consumer_name` are the same string type value. After processing the "consumer_id" in the core code "ctx.lua" file here, all users will see the "consumer_name".

moonming (approved):

apisix/core/ctx.lua
```diff
@@ -96,6 +96,9 @@ do
             elseif key == "service_id" then
                 val = ngx.ctx.api_ctx and ngx.ctx.api_ctx.service_id
 
+            elseif key == "consumer_name" then
```
membphis, line 99: The same value can also be obtained when using `consumer_id`.

Firstsawyou: ok, I will add it later.

membphis (approved):

PR #2345, opened 2020-09-29
feat: Add lables for route/consumer/service/ssl objects

imjoey:
Signed-off-by: imjoey <majunjiev@gmail.com>

Following with #2279 , this PR is going to add `labels` property support for `Route`/`Consumer`/`Service`/`SSL` objects, as well as fix some typos introduced by #2279 .

### Pre-submission checklist:

### DIFF
--- a/apisix/schema_def.lua
+++ b/apisix/schema_def.lua
@@ -452,6 +452,15 @@ _M.route = {
         plugins = plugins_schema,
         upstream = upstream_schema,
 
+        labels = {
+            description = "key/value pairs to specify attributes",
+            type = "object",
+            patternProperties = {
+                [".*"] = label_value_def
+            },
+            maxProperties = 16
+        },
+
         service_id = id_schema,
         upstream_id = id_schema,
         service_protocol = {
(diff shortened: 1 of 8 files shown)


imjoey:
ping @membphis @nic-chen @gxthrj for reviewing, much appreciated.

apisix/schema_def.lua
```diff
@@ -555,6 +580,14 @@ _M.ssl = {
             type = "integer",
             minimum = 1588262400,  -- 2020/5/1 0:0:0
         },
+        labels = {
```
nic-chen, line 583: I think ssl may not need labels

membphis:
nice PR. LGTM

t/admin/consumers.t
```diff
@@ -223,3 +223,78 @@ GET /t
...
+                         "env":"production",
+                         "version":"v2"
+                     }
+                }]],
+                [[{
+                    "node": {
+                        "value": {
+                            "username": "jack",
+                            "desc": "new consumer",
+                            "labels": {
```
liuxiran, line 250: node does not support `labels`, so here may need to remove `labels`?

apisix/schema_def.lua
imjoey: @nic-chen  IMHO, we could use label to distinguish among dozens of SSLs according to their usage.

imjoey: @membphis  @liuxiran  any thoughts on this? 😄

membphis: > we could use label to distinguish among dozens of SSLs according to their usage.

Agree this. Searching by category will be helpful.

liuxiran: > we could use label to distinguish among dozens of SSLs according to their usage

agree +1

nic-chen: OK, I think it could work with SNI.But for management, labels is helpful for it.

membphis (approved):

nic-chen (approved):

PR #2357, opened 2020-10-05
bugfix: set random seed for each worker process at `init_worker` phase, only `init` phase is not enough.

membphis:

related PR: https://github.com/apache/apisix/pull/2306

### Pre-submission checklist:

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -82,6 +82,14 @@ end
 
 
 function _M.http_init_worker()
+    local seed, err = core.utils.get_seed_from_urandom()
+    if not seed then
+        core.log.warn('failed to get seed from urandom: ', err)
+        seed = ngx_now() * 1000 + ngx.worker.pid()
+    end
+    math.randomseed(seed)
+    core.log.info("random test in [1, 10000]: ", math.random(1, 1000000))
+
     local we = require("resty.worker.events")
     local ok, err = we.configure({shm = "worker-events", interval = 0.1})
     if not ok then
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -81,6 +81,7 @@ function _M.http_init_worker()
         seed = ngx_now() * 1000 + ngx.worker.pid()
     end
     math.randomseed(seed)
+    -- for testing only
     core.log.info("random test in [1, 10000]: ", math.random(1, 1000000))
 
     local we = require("resty.worker.events")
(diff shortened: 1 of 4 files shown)


membphis:
@zlm0125 welcome to review and make a test with this PR. I think it should work fine for your case

moonming:
why only `init` phase is not enough?

membphis:
> why only `init` phase is not enough?

`init` works in `master` process if the [lua_code_cache](https://github.com/openresty/lua-nginx-module#lua_code_cache) is `on`.

We want to set different `seed` for different work processes, so we should use `init_worker` phase.

we can delete those code https://github.com/apache/apisix/pull/2357/files#diff-d982d52466e7c93c7b604358339b2a29R85-R90 . 
And run the test case, then we will get the same random number which is wrong.

image

Here is the right one:

image


image

moonming:
> related PR: #2306

I don't understand what it has to do with this PR

moonming:
And we should remove random in `init` phase if add random in `init_worker` phase.

membphis:
> And we should remove random in `init` phase if add random in `init_worker` phase.

you are right. let me do this later.

apisix/init.lua
```diff
@@ -82,6 +75,14 @@ end
 
 
 function _M.http_init_worker()
+    local seed, err = core.utils.get_seed_from_urandom()
+    if not seed then
+        core.log.warn('failed to get seed from urandom: ', err)
+        seed = ngx_now() * 1000 + ngx.worker.pid()
+    end
+    math.randomseed(seed)
+    core.log.info("random test in [1, 10000]: ", math.random(1, 1000000))
```
moonming, line 84: should be removed

apisix/init.lua
```diff
@@ -759,6 +760,14 @@ end
 
 function _M.stream_init_worker()
     core.log.info("enter stream_init_worker")
+    local seed, err = core.utils.get_seed_from_urandom()
+    if not seed then
+        core.log.warn('failed to get seed from urandom: ', err)
+        seed = ngx_now() * 1000 + ngx.worker.pid()
+    end
+    math.randomseed(seed)
+    core.log.info("random stream test in [1, 10000]: ", math.random(1, 1000000))
```
moonming, line 769: ditto

apisix/init.lua
membphis: why do we need to remove this line?
It is necessary for debugging.

moonming: Can it help user to find any problems?

membphis: it not for the user, the developer need it for testing.

here is the code link of the test case: https://github.com/apache/apisix/pull/2357/files#diff-a46ca98b2569d841ad4742ef90427a1fR39

moonming (approved):

apisix/init.lua
spacewander: @membphis 
I think we should add a comment for it.

spacewander (approved):

PR #2390, opened 2020-10-12
fix: it does'nt make sence to compare etcd_version, since it's nil.

RocFang:

1. the result below will always be false ,since etcd_version is nil:

```lua
if (etcd_version == "v3" and not res:find("OK", 1, true)) then
                is_success = false
                if (index == host_count) then
                    error(cmd .. "\n" .. res)
                end

                break
            end
``` 
and since etcd_version has been checked before, I think maybe there is no need to check it again here.

2. cluster_version has been defined previously, no need to define a new local one.

### Pre-submission checklist:

### DIFF
--- a/bin/apisix
+++ b/bin/apisix
@@ -984,7 +984,6 @@ local function init_etcd(show_output)
         yaml_conf.etcd.host = {yaml_conf.etcd.host}
     end
 
-    local cluster_version
     local host_count = #(yaml_conf.etcd.host)
     local dkjson = require("dkjson")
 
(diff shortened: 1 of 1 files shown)


membphis (approved):

bin/apisix
```diff
@@ -1000,7 +1000,7 @@ local function init_etcd(show_output)
             os.exit(1)
         end
 
-        local cluster_version = body["etcdcluster"]
+        cluster_version = body["etcdcluster"]
```
Yiyiyimu, line 1003: Maybe delete the outside definition is better, since we use the new one in each cycle.

membphis: agree +1

RocFang:
@membphis  BTW, I think bin/apisix should also be checked by luacheck, and then the little mistake like this undifinend var in comparasion expression will not happen. Only problem is the constraint of line length may be different now.

membphis:
@RocFang many thx

PR #2397, opened 2020-10-12
feature(core): implement `core.sleep`

### LINKED ISSUE #2170: feature(core): implement `core.sleep`
Use a small time interval (such as 1s) to complete a relatively long sleep behavior, so that the worker process can be shut down gracefully.

Here is the fake code(we may):

```lua
function sleep(sec)
    if sec <= 1 then
        ngx_sleep(sec)
        return
    end

    while true do
        if sec > 1 then
            ngx_sleep(1)
        end

        if exiting() then
            return
        end

        sec = sec - 1
        if sec < 1 then
            ngx_sleep(sec)
            return
        end
    end
end
```

dabue:
fix #2170 

Use a small time interval (such as 1s) to complete a relatively long sleep behavior, so that the worker process can be shut down gracefully.

```
local max_sleep_interval = 1

local function sleep(sec)
    if sec <= max_sleep_interval then
        ngx_sleep(sec)
        return
    end
    ngx_sleep(max_sleep_interval)
    if exiting() then
        return
    end
    sec = sec - max_sleep_interval
    sleep(sec)
end

### CODE REVIEW

**Comment by membphis** [comment]
> local function sleep(sec)
>     if sec <= max_sleep_interval then
>         ngx_sleep(sec)
>         return
>     end
>     ngx_sleep(max_sleep_interval)
>     if exiting() then
>         return
>     end
>     sec = sec - max_sleep_interval
>     sleep(sec)
> end

missing`while true do ... end`, so your current way is wrong @dabue

File: apisix/core/utils.lua [hunk 1]
```diff
@@ -194,4 +196,22 @@ function _M.validate_header_value(value)
 end
 
 
+local function sleep(sec)
+    if sec <= max_sleep_interval then
+        ngx_sleep(sec)
+        return
```
membphis, line 202: `return ngx_sleep(sec)`
use proper tail recursion is good here

apisix/core/utils.lua
```diff
@@ -194,4 +196,22 @@ function _M.validate_header_value(value)
...
+    ngx_sleep(max_sleep_interval)
+    if exiting() then
+        return
+    end
+    sec = sec - max_sleep_interval
+    sleep(sec)
+end
+
+
+function _M.sleep(sec)
```
membphis, line 213: `_M.sleep = sleep` is enough

membphis (commented):
LGTM, except for minor code style

apisix/core/utils.lua
dabue: I‘ve updated it.’

membphis (approved):

moonming:
Is there a long sleep in the project?

moonming:
This change seems to be of little use to the current code

dabue:
> This change seems to be of little use to the current code
Among the current code, it really only work better when the  worker process being shut down at the time of  triggering request limiting and delaying processing.  And we can keep it as a better way to sleep for the future usage scenarios.

membphis:
I think this feature is useful. 

When the APISIX instance restarts, we need to exit `sleep` as soon as possible. 
Then the Nginx process can be gracefully shut down.

spacewander (approved):

membphis:
@dabue many thx

PR #2430, opened 2020-10-15
fix: plugin hot reload should work on node

spacewander:
### Pre-submission checklist:


membphis:
that is a big feature, I'll take a look at this PR when I time ^_^

many thx for your contribution @spacewander

apisix/core/set.lua
```diff
@@ -0,0 +1,40 @@
...
+--
+local nkeys        = require("table.nkeys")
+local pairs        = pairs
+
+
+local _M = {}
+
+
+-- Compare two tables as if they are sets (only compare the key part)
+function _M.eq(a, b)
```
membphis, line 25: how about move the `eq` into `apisix.core.table`?

membphis: if yes, we need a better function name.

apisix/plugins/prometheus.lua
```diff
@@ -29,7 +29,6 @@ local _M = {
     version = 0.2,
     priority = 500,
     name = plugin_name,
-    init = exporter.init,
```
membphis, line 32: why do we need to delete this line? I think we need to keep it.

membphis, line 32: why do we need to delete this line? I think we need to keep it.

spacewander: @membphis 
Because we don't call the prometheus init function when the plugin reload anymore.

apisix/schema_def.lua
```diff
@@ -665,6 +665,23 @@ _M.stream_route = {
...
+        type = "object",
+        properties = {
+            name = {
+                type = "string"
+            },
+            stream = {
+                type = "boolean"
+            },
+        },
+        required = {"name"}
```
membphis, line 680: add `additionalProperties = false`

apisix/schema_def.lua
```diff
@@ -665,6 +665,23 @@ _M.stream_route = {
 }
 
 
+_M.plugins = {
+    type = "array",
+    items = {
+        type = "object",
+        properties = {
+            name = {
+                type = "string"
```
membphis, line 674: need to limit the `max`, `min` length.

membphis (approved):
LGTM

juzhiyuan (approved):
LGTM from codes style

membphis:
@spacewander many thx, merged already

PR #2435, opened 2020-10-16
CI: turn off the Lua code coverage check in CI.

membphis:

We don't enable Lua code override now, its output is unstable and this is unacceptable.

Disabling the Lua code override first will make CI run faster.

### Pre-submission checklist:

### DIFF
--- a/.travis/linux_openresty_runner.sh
+++ b/.travis/linux_openresty_runner.sh
@@ -169,7 +169,8 @@ script() {
     sudo bash ./utils/check-plugins-code.sh
 
     make lint && make license-check || exit 1
-    APISIX_ENABLE_LUACOV=1 PERL5LIB=.:$PERL5LIB prove -Itest-nginx/lib -r t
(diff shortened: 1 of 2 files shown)


.travis/linux_openresty_runner.sh
```diff
@@ -169,7 +169,8 @@ script() {
     sudo bash ./utils/check-plugins-code.sh
 
     make lint && make license-check || exit 1
-    APISIX_ENABLE_LUACOV=1 PERL5LIB=.:$PERL5LIB prove -Itest-nginx/lib -r t
+    # APISIX_ENABLE_LUACOV=1 PERL5LIB=.:$PERL5LIB prove -Itest-nginx/lib -r t
```
juzhiyuan, line 172: Could we remove this comment line?

juzhiyuan (approved):
Looks good to me from codes style

spacewander (approved):

.travis/linux_openresty_runner.sh
membphis: we may enable it later, so I think we should keep the original version.

Issue #2402, opened 2020-09-25
plugin(consumer-restriction): update the schema


liuxiran:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

I'm confused about the option select item in the consumer-restriction plugin
![image](https://user-images.githubusercontent.com/2561857/94222544-3adfc000-ff20-11ea-8c4e-b4448e045b34.png)
 what does it mean?

___
### Bug
- Which version of Apache APISIX Dashboard, OS, and Browser?

- What happened?
If possible, provide a way to reproduce the error.

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

juzhiyuan:
Just transferred the issue to the apisix repository, cc @membphis

Jaycean:
assign to me

Issue #2439, opened 2020-09-23
plugin(basic-auth): update the schema


liuxiran:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [x] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

What is the purpose of SELECT item？
![image](https://user-images.githubusercontent.com/2561857/93974652-ba4a8380-fda8-11ea-927c-66918099437a.png)
 
___
### Bug
- Which version of Apache APISIX Dashboard, OS, and Browser?

- What happened?
If possible, provide a way to reproduce the error.

1. config basic-auth plugin for a route

1. select `work with consumer object`, and then click [ok]button

1. action failed , and got an validate error in the console
![image](https://user-images.githubusercontent.com/2561857/93974925-2927dc80-fda9-11ea-9b1f-d0ee3d44feae.png)
 

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

liuxiran:
The solution is the same with https://github.com/apache/apisix/issues/2441, it has already worked well with the apisix master branch and dashboard v2.0 branch.

close is now :)

PR #2440, opened 2020-10-16
plugin(consumer-restriction): use draft7 way to rewrite the JSON Schema.

Jaycean:
- fix: #2402 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -27,7 +27,7 @@ local schema = {
                     type = "string",
                     enum = {"consumer_name", "service_id"},
                     default = "consumer_name"
-               },               
+               },
 		blacklist = {
                     type = "array",
                     minItems = 1,
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -19,27 +19,41 @@ local core      = require("apisix.core")
 
 local schema = {
     type = "object",
-    properties = {
-        type = {
-            type = "string",
-            enum = {"consumer_name", "service_id"},
-            default = "consumer_name"
-        },
-        whitelist = {
-            type = "array",
-            items = {type = "string"},
-            minItems = 1
-        },
-        blacklist = {
-            type = "array",
-            items = {type = "string"},
-            minItems = 1
-        },
-        rejected_code = {type = "integer", minimum = 200, default = 403}
-    },
     oneOf = {
-        {required = {"whitelist"}},
-        {required = {"blacklist"}}
+        {
+            title = "blacklist",
+            properties = {
+                type = {
+                    type = "string",
+                    enum = {"consumer_name", "service_id"},
+                    default = "consumer_name"
+               },               
+		blacklist = {
+                    type = "array",
+                    minItems = 1,
+                    items = {type = "string"}
+                },
+                rejected_code = {type = "integer", minimum = 200, default = 403}
+            },
+            required = {"blacklist"},
+        },
+        {
+            title = "whitelist",
+            properties = {
+	        type = {
+                    type = "string",
+                    enum = {"consumer_name", "service_id"},
+                    default = "consumer_name"
+               },
+                whitelist = {
+                    type = "array",
+                    minItems = 1,
+                    items = {type = "string"}
+	       },
+               rejected_code = {type = "integer", minimum = 200, default = 403}
+            },
+            required = {"whitelist"},
+        }
     }
 }
 
--- a/t/plugin/consumer-restriction.t
+++ b/t/plugin/consumer-restriction.t
@@ -30,8 +30,9 @@ __DATA__
     location /t {
         content_by_lua_block {
             local plugin = require("apisix.plugins.consumer-restriction")
-            local conf = {
-                whitelist = {
+            local conf = {                
+		title = "whitelist",
+		whitelist = {
                     "jack1",
                     "jack2"
                 }
@@ -47,7 +48,7 @@ __DATA__
 --- request
 GET /t
 --- response_body
-{"rejected_code":403,"type":"consumer_name","whitelist":["jack1","jack2"]}
+{"type":"consumer_name","title":"whitelist","rejected_code":403,"whitelist":["jack1","jack2"]}
 --- no_error_log
 [error]
 


juzhiyuan:
@Jaycean Please have a look at CI.

Jaycean:
> @Jaycean Please have a look at CI.

OK, I'll take a look at the reason. It passed the local test

Jaycean:
According to the previous JSON schema data format, without changing the logic, the current writing method is not the same as the data obtained by the front end. I need to discuss with @liuxiran  again tomorrow

membphis (approved):

juzhiyuan:
> > @Jaycean Please have a look at CI.
> 
> OK, I'll take a look at the reason. It passed the local test

Well, I will convert this PR to draft.

liuxiran:
> According to the previous JSON schema data format, without changing the logic, the current writing method is not the same as the data obtained by apisix-dashboard. I need to discuss with @liuxiran again tomorrow

@Jaycean 
with the newest json schema
```json
{
	"oneOf": [{
		"properties": {
			"blacklist": {
				"items": {
					"type": "string"
				},
				"minItems": 1,
				"type": "array"
			},
			"rejected_code": {
				"default": 403,
				"minimum": 200,
				"type": "integer"
			},
			"type": {
				"default": "consumer_name",
				"enum": ["consumer_name", "service_id"],
				"type": "string"
			}
		},
		"required": ["blacklist"],
		"title": "blacklist"
	}, {
		"properties": {
			"rejected_code": {
				"default": 403,
				"minimum": 200,
				"type": "integer"
			},
			"type": {
				"default": "consumer_name",
				"enum": ["consumer_name", "service_id"],
				"type": "string"
			},
			"whitelist": {
				"items": {
					"type": "string"
				},
				"minItems": 1,
				"type": "array"
			}
		},
		"required": ["whitelist"],
		"title": "whitelist"
	}],
	"properties": {
		"disable": {
			"type": "boolean"
		}
	},
	"type": "object"
}
```
dashboard can create a route with `key-auth` plugin and `consumer-restriction` plugin successfully. and the `consumer-restriction` plugin can work correctly with the configuration.

If you can change the keys **order** in properties one from current:
`blacklist, rejected_code, type`  to expected `rejected_code, type, blacklist`, that would be better :)

juzhiyuan:
> > According to the previous JSON schema data format, without changing the logic, the current writing method is not the same as the data obtained by apisix-dashboard. I need to discuss with @liuxiran again tomorrow
> 
> @Jaycean
> with your newest json schema
> 
> ```json
> {
> 	"oneOf": [{
> 		"properties": {
> 			"blacklist": {
> 				"items": {
> 					"type": "string"
> 				},
> 				"minItems": 1,
> 				"type": "array"
> 			},
> 			"rejected_code": {
> 				"default": 403,
> 				"minimum": 200,
> 				"type": "integer"
> 			},
> 			"type": {
> 				"default": "consumer_name",
> 				"enum": ["consumer_name", "service_id"],
> 				"type": "string"
> 			}
> 		},
> 		"required": ["blacklist"],
> 		"title": "blacklist"
> 	}, {
> 		"properties": {
> 			"rejected_code": {
> 				"default": 403,
> 				"minimum": 200,
> 				"type": "integer"
> 			},
> 			"type": {
> 				"default": "consumer_name",
> 				"enum": ["consumer_name", "service_id"],
> 				"type": "string"
> 			},
> 			"whitelist": {
> 				"items": {
> 					"type": "string"
> 				},
> 				"minItems": 1,
> 				"type": "array"
> 			}
> 		},
> 		"required": ["whitelist"],
> 		"title": "whitelist"
> 	}],
> 	"properties": {
> 		"disable": {
> 			"type": "boolean"
> 		}
> 	},
> 	"type": "object"
> }
> ```
> 
> dashboard can create a route with `key-auth` plugin and `consumer-restriction` plugin successfully. and the `consumer-restriction` plugin can work correctly with the configuration.
> 
> And what is the purpose of the disable property below, which is not mentioned both in the [doc](https://github.com/apache/apisix/blob/master/doc/zh-cn/plugins/consumer-restriction.md) and in [test cases](https://github.com/apache/apisix/blob/master/t/plugin/consumer-restriction.t). it would be better to remove if it is no use:)
> cc @juzhiyuan @membphis
> 
> ```json
> "properties": {
> 		"disable": {
> 			"type": "boolean"
> 		}
> 	}
> ```

hi please see here https://github.com/apache/apisix-dashboard/issues/495

liuxiran:
> hi please see here apache/apisix-dashboard#495

so you means  apisix should  not return disable filed until get request when FE needed? @juzhiyuan 

it seems that when the schema of the plugin does not have a `properties` property in the first level of the json schema, the interface returns the data with `disable` field.
related code: https://github.com/membphis/apisix/blob/133b52ec78665b7292af62bab6af6bf981363a42/apisix/plugin.lua#L81
which doesn't seem reasonable, especially after json schema  changed from draft5 to draft7. 
in json schema draft7, the `oneof` `anyof` and `allof` type of schema e.g the schema in pr https://github.com/apache/apisix/pull/2440,whice does not contain `properties` in first level 
cc @Jaycean 

the https://github.com/apache/apisix/issues/2439  may also have something to do with it.

juzhiyuan:
> > hi please see here [apache/apisix-dashboard#495](https://github.com/apache/apisix-dashboard/issues/495)
> 
> so you means apisix should not return disable filed until get request when FE needed? @juzhiyuan
> 
> it seems that when the schema of the plugin does not have a `properties` property in the first level of the json schema, the interface returns the data with `disable` field.
> related code: https://github.com/membphis/apisix/blob/133b52ec78665b7292af62bab6af6bf981363a42/apisix/plugin.lua#L81
> which doesn't seem reasonable, especially after json schema changed from draft5 to draft7.
> in json schema draft7, the `oneof` `anyof` and `allof` type of schema e.g the schema in pr #2440 does not contain `properties` in first level
> cc @Jaycean
> 
> the #2439 may also have something to do with it.

hi @membphis, What do you think?

Jaycean:
> > > hi please see here [apache/apisix-dashboard#495](https://github.com/apache/apisix-dashboard/issues/495)
> > 
> > 
> > so you means apisix should not return disable filed until get request when FE needed? @juzhiyuan
> > it seems that when the schema of the plugin does not have a `properties` property in the first level of the json schema, the interface returns the data with `disable` field.
> > related code: https://github.com/membphis/apisix/blob/133b52ec78665b7292af62bab6af6bf981363a42/apisix/plugin.lua#L81
> > which doesn't seem reasonable, especially after json schema changed from draft5 to draft7.
> > in json schema draft7, the `oneof` `anyof` and `allof` type of schema e.g the schema in pr #2440 does not contain `properties` in first level
> > cc @Jaycean
> > the #2439 may also have something to do with it.
> 
> hi @membphis, What do you think?

Yes, Lua code is in `apisix/apisix/plugin.lua`
I think it's the code's judgment on the current schema structure that needs to be changed
However, I may not have a deep understanding of the whole project and the role of the disable field, so I have not thought about how to change the code logic
```
local function load_plugin(name, plugins_list, is_stream_plugin)
    # line82-88
    if plugin.schema and plugin.schema.type == "object" then
        if not plugin.schema.properties or
           core.table.nkeys(plugin.schema.properties) == 0
        then
            plugin.schema.properties = core.schema.plugin_disable_schema
        end
    end
```

Because in the JSON schema Draft 7, some structures do not contain `properties` in the first layer, but there are `properties` in the second layer, but the logic fails to judge. As a result, some schemas do not need to add the` disable` field

juzhiyuan:
the `disable` filed is an injection for every plugin's jsonschema, just waiting for @membphis's reply :)

membphis:
> the `disable` filed is an injection for every plugin's jsonschema, just waiting for @membphis's reply :)

we do not need to add this `disable` field in schema.

Jaycean:
> > the `disable` filed is an injection for every plugin's jsonschema, just waiting for @membphis's reply :)
> 
> we do not need to add this `disable` field in schema.

OK, this field has no effect on the current pr. it affects all schemas. Do you mean that the current function can directly remove this field? If it can be removed, I will submit a new PR later.

membphis:
> Do you mean that the current function can directly remove this field? 

You haven't added this field yet? I do not find it in the changed files: https://github.com/apache/apisix/pull/2440/files

Jaycean:
> > Do you mean that the current function can directly remove this field?
> 
> You haven't added this field yet? I do not find it in the changed files: https://github.com/apache/apisix/pull/2440/files

My current PR does not modify the logic of this disable field. I intend to submit a new PR for modification

membphis:
thx for your explain

liuxiran:
> My current PR does not modify the logic of this disable field. I intend to submit a new PR for modification
> 
> After all, the logic related to the disable field has nothing to do with the current PR, which first deals with the schema format of the plugin(consumer-restriction)

Eagerly awaiting for a change in logic of this disable field😍, which will allow the dashboard better to use.

liuxiran (approved):

juzhiyuan (approved):
Looks good to me then

Issue #2432, opened 2020-10-15
bug: incorrect json schema definition for plugin request-validation


imjoey:
Part of #2082 .

Current request-validation schema definition is shown as below:

```json
{
    "anyOf": [
        {
            "required": [
                "body_schema"
            ]
        }, 
        {
            "required": [
                "header_schema"
            ]
        }
    ], 
    "properties": {
        "body_schema": {
            "type": "object"
        }, 
        "header_schema": {
            "type": "object"
        }
    }, 
    "type": "object"
}
```

The definition should be changed to:

```json
{
    "anyOf":[
        {
	    "title": "Body schema",
            "properties":{
                "body_schema":{
                    "type":"object"
                }
            },
            "required":[
                "body_schema"
            ]
        },
        {
	    "title": "Header schema",
            "properties":{
                "header_schema":{
                    "type":"object"
                }
            },
            "required":[
                "header_schema"
            ]
        }
    ],
    "type": "object"
}
```

**BTW**:
Honestly, after the schema has been updated, we are still not be able to input anything for `header_schema` and `body_schema` fields. See the snapshot below:

![image](https://user-images.githubusercontent.com/2542401/96113795-43bc2400-0f17-11eb-8a01-0dc1aeea7455.png)

The reason is that both `body_schema` and `header_schema` are of `object` type without any `properties` defined, while the react-jsonschema-form library will not show any input fields for them. I tried to refer to https://react-jsonschema-form.readthedocs.io/en/v1.8.1/form-customization/ , unfortunately there seems no way to make use of `uiSchema` to solve this problem.

After discussed with @liuxiran , in order to make it happen that we could define the schema on the web, we also need to customize the dashboard UI, as well as well design the `header_schema` and `body_schema` data structures submitted to APISIX. That could be committed to dashboard project following this issue got fixed.

imjoey:
@juzhiyuan @liuxiran  Please see https://bjdash.github.io/JSON-Schema-Builder/ that may be a good choice for online json schema designers. 😄

juzhiyuan:
ok then, @membphis Please take a look at this new schema

fukiki:
@juzhiyuan  @membphis @liuxiran Please assign to me.

PR #2451, opened 2020-10-17
Plugin(request-validation): use draft7 way to rewrite the JSON Schema

fukiki:
- fix bug #2432

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/request-validation.lua
+++ b/apisix/plugins/request-validation.lua
@@ -21,13 +21,21 @@ local io           = io
 
 local schema = {
     type = "object",
-    properties = {
-        body_schema = {type = "object"},
-        header_schema = {type = "object"}
-    },
     anyOf = {
-        {required = {"body_schema"}},
-        {required = {"header_schema"}}
+        {
+            title = "Body schema",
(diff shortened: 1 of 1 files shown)


t/plugin/request-validation.t
```diff
@@ -271,46 +231,6 @@ hello1 world
                             "type": "roundrobin"
                         },
                         "uri": "/opentracing"
-                }]],
```
imjoey, line 274: @fukiki may I ask the reason to to remove these contents? Thanks.

fukiki: > @fukiki may I ask the reason to to remove these contents? Thanks.

In my test environment, the parameter string is larger than the remain space of the fixed buffer, which causes nginx to report an error. Therefore, the optional parameter was deleted.
The code in `t/plugin/request-validation.t` has been restored.

imjoey: @fukiki Thanks and good to see it's back. 😄

membphis (approved):

juzhiyuan (approved):

PR #2456, opened 2020-10-18
CI: remove `luacov`, maybe reopen it later.

membphis:

as title

### Pre-submission checklist:


moonming (changes requested):
Why remove Luacov?

membphis:
> Why remove Luacov?

the result is unstable, and it causes the test case to time out and fails.

and we have no way to view the result now.

I think we can reopen it after we resolve those problems as I commented in the CI scripts.

moonming (approved):

juzhiyuan (approved):

PR #2465, opened 2020-10-19
fix(limit-conn): always save the data of the limit object, and release it in log phase.

### LINKED ISSUE #2450: bug: limit-conn
### Issue description
I had test  the plugin of limit-conn.I found the question that the same ip only request the router by five times when conn is 5 and key is remote_addr, and i don't request the  router next time.Why so that?

### Environment

* OS:

Plugin limit-conn Edit
 conn=5
rejected_code=503
 burst=0
 key=remote_addr
 default_conn_delay=0

same user only visit the router with the config 

i hope that the plugin only limit concurrency,and different users also continue to request the router when user request the service after first .

membphis:

fix #2450

### Pre-submission checklist:

### DIFF
--- a/t/plugin/limit-conn.t
+++ b/t/plugin/limit-conn.t
@@ -989,7 +989,7 @@ done
 
 
 === TEST 26: create consumer and bind key-auth plugin
---- config 
+--- config
     location /t {
         content_by_lua_block {
             local t = require("lib.test_admin").test
@@ -1019,7 +1019,7 @@ passed
 
 
 
-=== TEST 27: create route and enable plugin 'key-auth' 
+=== TEST 27: create route and enable plugin 'key-auth'
 --- config
     location /t {
         content_by_lua_block {
@@ -1207,9 +1207,9 @@ qr/limit key: consumer_jackroute&consumer\d+/
         content_by_lua_block {
             local plugin = require("apisix.plugins.limit-conn")
             local ok, err = plugin.check_schema({
-                conn = 1, 
-                default_conn_delay = 0.1, 
-                rejected_code = 503, 
+                conn = 1,
+                default_conn_delay = 0.1,
+                rejected_code = 503,
                 key = 'consumer_name'
                     })
             if not ok then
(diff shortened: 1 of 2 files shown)


membphis:
@LY-GO you can make a try with this PR

LY-GO:
> @LY-GO you can make a try with this PR
two question:
1.now i have a requirement.iI need to add a key that it belonds to config of limit-conn-plugin .I just modify  the define of key?
key
2.i modify the code of limit-conn according to your repair.But the response code is 500

LY-GO:
> > @LY-GO you can make a try with this PR
> 2.i modify the code of limit-conn according to your repair.But the response code is 500
code

spacewander (approved):

spacewander:
@LY-GO 
I think the 500 may be related with #2472

LY-GO:
> @LY-GO you can make a try with this PR

I test the plugin,i find it don't  work;now every request can success,even if number of request over concurrency limit.I just copy your code,overwrite a new plugin and reload the plugin.

LY-GO:
> > > @LY-GO you can make a try with this PR
> > > 2.i modify the code of limit-conn according to your repair.But the response code is 500
> > > code



> @LY-GO you can make a try with this PR

I test the plugin,i find it don't work;now every request can success,even if number of request over concurrency limit.I just copy your code,overwrite a new plugin and reload the plugin

LY-GO:
> > > > @LY-GO you can make a try with this PR
> > > > 2.i modify the code of limit-conn according to your repair.But the response code is 500
> > > > code
> 
> > @LY-GO you can make a try with this PR
> 
> I test the plugin,i find it don't work;now every request can success,even if number of request over concurrency limit.I just copy your code,overwrite a new plugin and reload the plugin

two suggesstions:
1.When you use limit-conn plugin,you must set default_conn_delay > 0, otherwise you will find assertion failed.
2.When you test limit-conn plugin,your upstream must set sleep time.Because concurrency is required at the same time .

spacewander:
@LY-GO 
> I test the plugin,i find it don't work;now every request can success,even if number of request over concurrency limit.I just copy your code,overwrite a new plugin and reload the plugin

> 2.When you test limit-conn plugin,your upstream must set sleep time.Because concurrency is required at the same time .

Is the note 2 caused the plugin didn't work in your test? Does it work after you set sleep time?

membphis:
@LY-GO please provide your plugin configuration. it maybe invalid.

image

LY-GO:
> @LY-GO
> 
> > I test the plugin,i find it don't work;now every request can success,even if number of request over concurrency limit.I just copy your code,overwrite a new plugin and reload the plugin
> 
> > 2.When you test limit-conn plugin,your upstream must set sleep time.Because concurrency is required at the same time .
> 
> Is the note 2 caused the plugin didn't work in your test? Does it work after you set sleep time?

When you set sleep time in your procedure,you can use jmeter to test plugin and the plugin can work successfully.

spacewander:
I think the problem is solved. What made the fix won't work is just the way of test (for example, the upstream doesn't hold the connection).

LY-GO:
> I think the problem is solved. What made the fix won't work is just the way of test (for example, the upstream doesn't hold the connection).

I don't understand word "the upstream doesn't hold the connection"

membphis:
@tokers @moonming @spacewander  I have fixed the conflict, you can take a look

t/plugin/limit-conn.t
```diff
@@ -1207,9 +1207,9 @@ qr/limit key: consumer_jackroute&consumer\d+/
         content_by_lua_block {
             local plugin = require("apisix.plugins.limit-conn")
             local ok, err = plugin.check_schema({
-                conn = 1, 
-                default_conn_delay = 0.1, 
-                rejected_code = 503, 
+                conn = 1,
+                default_conn_delay = 0.1,
+                rejected_code = 503,
```
moonming, line 1212: why modify this test case？

t/plugin/limit-conn.t
```diff
@@ -1225,3 +1225,61 @@ property "burst" is required
...
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 33: hit route and should not be limited
+--- pipelined_requests eval
```
moonming, line 1273: Shouldn't the request be rejected here?  I did not understand

t/plugin/limit-conn.t
membphis: old test cases have `space` at the end of line. 

my editor helps me do the `slim`, this is fine.

image

t/plugin/limit-conn.t
membphis: All requests are sent one after another, they are not concurrent.

So the request should never be limited for `"conn": 1,`.

In the old code, it will be limited due to this bug.

membphis:
@moonming @tokers @nic-chen please take a look at this PR

moonming (approved):

PR #2488, opened 2020-10-21
feat: add error-log-logger plugin

### LINKED ISSUE #2410: feat: plugin `error-log-logger`, collect error log and report to remote server by `cosocket`
all of the error log will be written into local files now.

```
$ ls logs
access.log  error.log
```

we can implement a new plugin which named `error-log-logger`, it'll collect the error log from API: https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/errlog.md

we can report those error log to `syslog-ng` server or HTTP server.

dabue:
add a plugin to collect error log and report to remote TCP server .

fix #2410

you can set the plugin's config in config-default.yaml,  likes below:

plugin_attr:
    error-log-logger:
      host: "127.0.0.1"  # the remote tcp server
      port: 33333         # the port
      loglevel: "warn"  # the filter log level


conf/config-default.yaml
```diff
@@ -193,6 +193,7 @@ plugins:                          # plugin list
   - proxy-mirror
   - request-id
   - hmac-auth
+  - error-log-logger
```
membphis, line 196: Do not need to be turned on by default.

`# - error-log-logger`, should be fine.

conf/config-default.yaml
```diff
@@ -201,3 +202,9 @@ plugin_attr:
   log-rotate:
     interval: 3600    # rotate interval (unit: second)
     max_kept: 168     # max number of log files will be kept
+  error-log-logger:
+    host: "127.0.0.1"
```
membphis, line 206: need some comment for each field

conf/config-default.yaml
```diff
@@ -201,3 +202,9 @@ plugin_attr:
...
+    port: 33333
+    loglevel: "warn"
+
```
membphis, line 209: one blank line is enough here

conf/config-default.yaml
```diff
@@ -201,3 +202,9 @@ plugin_attr:
   log-rotate:
     interval: 3600    # rotate interval (unit: second)
     max_kept: 168     # max number of log files will be kept
+  error-log-logger:
+    host: "127.0.0.1"
+    port: 33333
+    loglevel: "warn"
```
membphis, line 208: need a better name: `log_level` or `level`

t/plugin/error-log-logger.t
```diff
@@ -0,0 +1,122 @@
...
+    my $stream_single_server = <<_EOC_;
+    # fake server, only for test
+    server {
+        listen 33333;
+
+        content_by_lua_block {
+            local exiting = ngx.worker.exiting
+            local sock, err = ngx.req.socket(true)
+            if not sock then
+                ngx.log(ngx.WARN, "socket error:" .. err)
```
membphis, line 35: `ngx.log(ngx.WARN, "socket error:" .. err)`

to

`ngx.log(ngx.WARN, "socket error:", err)`

t/plugin/error-log-logger.t
```diff
@@ -0,0 +1,122 @@
...
+            end
+        }
+    }
+_EOC_
+
+    $block->set_value("stream_config", $stream_single_server);
+});
+
+
+add_block_preprocessor(sub {
```
membphis, line 59: why we need two `add_block_preprocessor`? I think one is enough

t/plugin/error-log-logger.t
```diff
@@ -0,0 +1,122 @@
...
+--- request
+GET /tg
+--- response_body
+--- error_log eval
+qr/\[Server\] receive data:.*this is an error message for test./
+--- wait: 1
+
+
+
+=== TEST 3: log an info level message
```
membphis, line 109: more test case about when user set different address or wrong address

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,140 @@
...
+    local host, port
+    local timeout = 3
+    local keepalive = 3
+    local level = "warn"
+    if try_attr(local_conf, "plugin_attr", plugin_name) then
+        local attr = local_conf.plugin_attr[plugin_name]
+        host = attr.host
+        port = attr.port
+        level = attr.loglevel or level
+        timeout = attr.timeout or timeout
```
membphis, line 85: need to add this field in `config-default.yaml`

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,140 @@
...
+    local ok, err = sock:connect(host, port)
+    if not ok then
+        core.log.warn("connect to the server failed for " .. err)
+        return
+    end
+    local logs = errlog.get_logs(10)
+    while ( logs and #logs>0 ) do
+        for i = 1, #logs, 3 do
+            if logs[i] <= level then --ommit the lower log producted at the initial
+                local bytes, err = sock:send(logs[i + 2])
```
membphis, line 111: https://github.com/apache/apisix/blob/master/apisix/utils/batch-processor.lua

I think we can use this way to send the error log.

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,140 @@
...
+    sock:settimeout(timeout*1000)
+    local ok, err = sock:connect(host, port)
+    if not ok then
+        core.log.warn("connect to the server failed for " .. err)
+        return
+    end
+    local logs = errlog.get_logs(10)
+    while ( logs and #logs>0 ) do
+        for i = 1, #logs, 3 do
+            if logs[i] <= level then --ommit the lower log producted at the initial
```
membphis, line 110: I think we can remove this line, it seems useless

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,140 @@
...
+                local bytes, err = sock:send(logs[i + 2])
+                if not bytes then
+                    core.log.info("send data  failed for " , err, ", the data:", logs[i + 2] )
```
membphis, line 113: `] )` remove this space

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,140 @@
...
+                    return
+                end
+            end
+        end
+        logs = errlog.get_logs(10)
+	end
+    sock:setkeepalive(keepalive*1000)
```
membphis, line 120: code style: need a space, `keepalive * 1000`

membphis:
need to update the doc ^_^

t/plugin/error-log-logger.t
dabue: In the test case ,  I need to add a new TCP server to receive the error log data.  At the beginnging, I want to use a special port instead of a built-in one, so I  aad the first add_block_preprocessor, but there is a  built-in server which has no handler , It'll cases error if I don't add_block_preprocessor for it.

t/plugin/error-log-logger.t
dabue: This plugin is global, and it's attributes are in the config.yaml which should be created before the test cases  are executed. I've not found a way to include the right address case and the wrong address case in a single '.t' file .

dabue: or I create a new 't' file to add the wrong address case?

dabue:
@membphis  how to resolve the conflicts. there was no notice when I push my update.

dabue:
@membphis  please have a review, thx.

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,235 @@
...
+    end
+
+    return true
+end
+
+local function load_attr()
+    local local_conf = core.config.local_conf()
+    if try_attr(local_conf, "plugin_attr", plugin_name) then
+        local attr = local_conf.plugin_attr[plugin_name]
+        config.host = attr.host
```
membphis, line 106: this is not a good way to set the default value. 

we should use JSONSchema to check if it is valid and set the default value.

you can take a look at this code: 

https://github.com/apache/apisix/blob/master/apisix/plugins/http-logger.lua#L58-L71

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,235 @@
...
+    local res = false
+    local err_msg
+    local sock, soc_err = tcp()
+
+    if not sock then
+        err_msg = "failed to init the socket " .. soc_err
+        return res, err_msg
+    end
+
+    sock:settimeout(config.timeout*1000)
```
membphis, line 131: style: need a space between the arguments

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,235 @@
...
+    for i = 1, #entries do
+        log_buffer:push(entries[i])
+    end
+
+end
+function _M.init()
+    if ngx.get_phase() ~= "init" and ngx.get_phase() ~= "init_worker"  then
+        return
+    end
+    buffers = {}
```
membphis, line 209: style: add a blank line before this line

bin/apisix
```diff
@@ -224,6 +224,8 @@ http {
     lua_ssl_verify_depth 5;
     ssl_session_timeout 86400;
 
+    lua_capture_error_log     128k; # cache for  capture the error log
```
membphis, line 227: only set this directive when the user enabled `error-log-logger` plugin

conf/config-default.yaml
```diff
@@ -208,3 +209,9 @@ plugin_attr:
   log-rotate:
     interval: 3600    # rotate interval (unit: second)
     max_kept: 168     # max number of log files will be kept
+#  error-log-logger:
```
membphis, line 212: we can set those by `plugin_metadata` via Admin API.

https://github.com/apache/apisix/blob/master/apisix/plugins/http-logger.lua#L58-L71

t/plugin/error-log-logger.t
membphis: one `add_block_preprocessor ` is enough for your case.

t/plugin/error-log-logger.t
membphis: nice ^_^

membphis (changes requested):

apisix/plugins/error-log-logger.lua
dabue: has updated it

dabue:
@membphis  please review ：）

apisix/plugins/error-log-logger.lua
```diff
@@ -62,14 +63,18 @@ local log_level = {
...
 local config = {
 }
 
+
 local _M = {
     version = 0.1,
     priority = 1091,
     name = plugin_name,
     schema = schema,
+    metadata_schema = schema,
```
membphis, line 76: that is wrong. they are different between `schema` and `metadata_schema`.

membphis (changes requested):

apisix/plugins/error-log-logger.lua
dabue: I think the metadata_schema define the config used by the plug-in, Is there sth wrong?

dabue:
@membphis please review ：）

apisix/plugins/error-log-logger.lua
```diff
@@ -0,0 +1,224 @@
...
+        name = config.name,
+        retry_delay = config.retry_delay,
+        batch_max_size = config.batch_max_size,
+        max_retry_count = config.max_retry_count,
+        buffer_duration = config.buffer_duration,
+        inactive_timeout = config.inactive_timeout,
+    }
+
+    local err
+    log_buffer, err = batch_processor:new(func, config_bat)
```
membphis, line 188: we can use `send_to_server` directly, all right? then we can remove https://github.com/apache/apisix/pull/2488/files#diff-f33917797f0ad1ad27d82b581c7ba6afbd62abed8f6fc3038d9663ad0bc0519cR175-R177

dabue (commented):
got it.

.travis/linux_apisix_master_luarocks_runner.sh
```diff
@@ -59,6 +59,9 @@ script() {
     sudo mkdir -p /usr/local/apisix/deps
     sudo PATH=$PATH ./utils/install-apisix.sh install > build.log 2>&1 || (cat build.log && exit 1)
 
+    which apisix
```
membphis, line 62: You used the wrong way to rebase the branch.

This wrong result is usually due to the use of `merge` and `rebase` at the same time.

We have to close this PR first, then you can submit your correct code via new PR.

dabue: ok, I will submit code via  new PR.

dabue:
I will submit code via a new PR.

.travis/linux_apisix_master_luarocks_runner.sh
dabue: the new pr: https://github.com/apache/apisix/pull/2592

PR #2493, opened 2020-10-22
optimize: read trailing bytes only one time in stand-alone mode.

RocFang:

IMHO, there is no need to loop 10 times for read trailing bytes in stand-alone mode,
or maybe there is some special reason i didn't get.
If so, @membphis  would you please explain it a little? thanks.

### Pre-submission checklist:

### DIFF
--- a/apisix/core/config_yaml.lua
+++ b/apisix/core/config_yaml.lua
@@ -56,8 +56,8 @@ local mt = {
 }
 
 
-    local apisix_yaml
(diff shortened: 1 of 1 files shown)


membphis (approved):

membphis:
@RocFang your way is correct and better. I have approved, many thx for your contribution.

spacewander (approved):

Issue #2441, opened 2020-10-16
plugin(hmac-auth): update the schema


liuxiran:
Part of https://github.com/apache/apisix/issues/2082
### Current schema
```json
{
	"additionalProperties": false,
	"properties": {
		"disable": {
			"type": "boolean"
		}
	},
	"title": "work with route or service object",
	"type": "object"
}
```
the plugin page looks:
![2020-10-16 20-39-49屏幕截图](https://user-images.githubusercontent.com/2561857/96259158-c3b5bd00-0fef-11eb-8375-b631eecead37.png)

---

### Expected schema
according to the [doc](https://github.com/apache/apisix/blob/master/doc/zh-cn/plugins/hmac-auth.md#%E5%B1%9E%E6%80%A7) and [test case](https://github.com/apache/apisix/blob/1afa8f77272e6099d2c54e4f703f6a70adb16f72/t/plugin/hmac-auth.t#L27)

```json
{
  "type": "object",
  "required": [
    "access_key",
    "secret_key"
  ],
  "properties": {
    "access_key": {
      "type": "string",
      "title": "Accesskey"
    },
    "secret_key": {
      "type": "string",
      "title": "Secretkey"
    },
    "algorithm": {
      "type": "string",
      "enum": [
        "hmac-sha1",
        "hmac-sha256",
        "hmac-sha512"
      ],
      "default": "hmac-sha256",
      "title": "Algorithm"
    },
    "clock_skew": {
      "type": "number",
      "title": "clockskew",
      "default": 0
    },
    "signed_headers": {
      "type": "array",
      "title": "A list of strings",
      "items": {
        "type": "string",
        "default": "bazinga"
      }
    }
  }
}
```
the form looks:

![2020-10-16 21-06-48屏幕截图](https://user-images.githubusercontent.com/2561857/96261849-8b17e280-0ff3-11eb-8243-de6d4fca3691.png)

```

liuxiran:
`hmac-auth` define two schemas:

- one is : https://github.com/apache/apisix/blob/d1729c0af3bc416a88904568b4d2718c99a1982d/apisix/plugins/hmac-auth.lua#L40

- the other one is: https://github.com/apache/apisix/blob/d1729c0af3bc416a88904568b4d2718c99a1982d/apisix/plugins/hmac-auth.lua#L47

[...]

PR #2509, opened 2020-10-24
feat: get schema by schema_type

Jaycean:
- part fix apache/apisix#2441

### Pre-submission checklist:


apisix/admin/plugins.lua
```diff
@@ -25,7 +25,7 @@ local type      = type
 local table_remove = table.remove
 local table_sort = table.sort
 local table_insert = table.insert
-
+local json_schema = {}
```
membphis, line 28: `json_schema` should be a temporary variable in function.

membphis (changes requested):

apisix/admin/plugins.lua
Jaycean: Yes, the previous wording is indeed wrong, it has been revised

Jaycean:
@membphis PTAL,Thx

Jaycean:
@membphis I have some questions I want to discuss with you.

The method I currently implement is to directly determine whether to return `consumer_schema` through whether the query field has `schema_type=consumer` in `apisix/admin/plugins.lua`. 
```
    local arg = get_uri_args()
    local json_schema = plugin.schema	    local json_schema = plugin.schema
    if arg and arg["schema_type"] == "consumer" then
        json_schema = plugin.consumer_schema
    end
```
https://github.com/apache/apisix/issues/2441#issuecomment-714506729
I thought of another implementation method to obtain the corresponding schema directly through the value of `schema_type`, but if Obtain directly through the value of `schema_type`, then you need to pass the corresponding complete schema_type:`/apisix/admin/schema/plugins/{plugin_name}?schema_type=consumer_schema`
Instead of: `/apisix/admin/schema/plugins/{plugin_name}?schema_type=consumer`

membphis:
> /apisix/admin/schema/plugins/{plugin_name}?schema_type=consumer

I think this way is simpler and enough.

membphis (approved):

juzhiyuan (approved):
LGTM

liuxiran:
LGTM

tips:
after this pr merged, jwt-auth also needs to add `consumer_schema` to the `_M` @Jaycean 
https://github.com/apache/apisix/blob/1dfc4cdacd620e6d59c3d4c75506ff80581612eb/apisix/plugins/jwt-auth.lua#L59

liuxiran (approved):

Jaycean:
> LGTM
> 
> tips:
> after this pr merged, jwt-auth also needs to add `consumer_schema` to the `_M` @Jaycean
> https://github.com/apache/apisix/blob/1dfc4cdacd620e6d59c3d4c75506ff80581612eb/apisix/plugins/jwt-auth.lua#L59

OK, I will submit a new PR unified addition to consumer_schema support

juzhiyuan:
> LGTM
> 
> tips:
> after this pr merged, jwt-auth also needs to add `consumer_schema` to the `_M` @Jaycean
> https://github.com/apache/apisix/blob/1dfc4cdacd620e6d59c3d4c75506ff80581612eb/apisix/plugins/jwt-auth.lua#L59

- [ ] TODO this issue

PR #2516, opened 2020-10-26
feature: support percentage for fault injection

tokers:

Support percentage for the finer granularity fault injection.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/fault-injection.lua
+++ b/apisix/plugins/fault-injection.lua
@@ -14,8 +14,10 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
-local core          = require("apisix.core")
-local sleep         = core.sleep
+local core = require("apisix.core")
+
+local sleep = core.sleep
+local random = math.random
 
 local plugin_name   = "fault-injection"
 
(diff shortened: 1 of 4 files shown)


apisix/plugins/fault-injection.lua
```diff
@@ -28,13 +30,15 @@ local schema = {
             properties = {
                 http_status = {type = "integer", minimum = 200},
                 body = {type = "string", minLength = 0},
+                percentage = {type = "integer", minimum = 0, maximum = 100}
```
membphis, line 33: I am not sure if this way is power enough.

I am thinking, how about support `vars` for this feature?

`vars`, here is an example: `["random_100", ">", 70]`.

then we can support to use any request header or query string as condition, eg: `["arg_age", ">", 18]`

tokers: Actually the percentage is inspired from Istio. I think the purpose of fault injection is to observe whether the whole request link is still normal when some parts of them don't work so perfectly,which may need the randomness to mimic the production, instead of the deliberate order from administrator.

membphis: thx for your explain

membphis (approved):

gxthrj (approved):
LGTM

membphis:
@tokers merged already, many thx for your contribution

PR #2519, opened 2020-10-26
feature: support injecting lua_ssl_trusted_certificate.

tokers:

Support to configure `lua_ssl_trusted_certificate` in configuration, so that one can use the certs signed by their own CA as an external dependency of APISIX, like ETCD.

This PR is necessary for the subsequent TLS for ETCD connection feature.

### Pre-submission checklist:

### DIFF
--- a/bin/apisix
+++ b/bin/apisix
@@ -389,6 +389,10 @@ http {
         ssl_session_cache    shared:SSL:20m;
         ssl_session_timeout 10m;
 
+        {% if ssl.ssl_trusted_certificate ~= nil then %}
+        lua_ssl_trusted_certificate {* ssl.ssl_trusted_certificate *};
+        {% end %}
+
         ssl_protocols {* ssl.ssl_protocols *};
         ssl_ciphers {* ssl.ssl_ciphers *};
         ssl_prefer_server_ciphers on;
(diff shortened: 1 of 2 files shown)


moonming:
Is this also work for vault?

conf/config-default.yaml
```diff
@@ -101,6 +101,9 @@ apisix:
     enable: true
     enable_http2: true
     listen_port: 9443
+    ssl_trusted_certificate: /path/to/ca-cert # Specifies a file path with trusted CA certificates in the PEM format
+                                              # used to verify the certificate when APISIX needs to do SSL/TLS handshaking
+                                              # with external services (e.g. ETCD)
```
spacewander, line 106: ETCD => etcd

bin/apisix
```diff
@@ -389,6 +389,10 @@ http {
         ssl_session_cache    shared:SSL:20m;
         ssl_session_timeout 10m;
 
+        {% if ssl.ssl_trusted_certificate ~= nil then %}
+        lua_ssl_trusted_certificate {* ssl.ssl_trusted_certificate *};
```
spacewander, line 393: Better to check if the certificate exists before generating the configuration.

conf/config-default.yaml
tokers: Is this the conventional way in APISIX?

bin/apisix
tokers: OK

tokers:
> Is this also work for vault?

Yes, actually for all the external services.

bin/apisix
tokers: Added.

spacewander (approved):

PR #2596, opened 2020-11-02
fix(log-rotate): avoid creating multiple timers

spacewander:
### Pre-submission checklist:

### DIFF
--- a/apisix/timers.lua
+++ b/apisix/timers.lua
@@ -25,8 +25,7 @@ local thread_wait = ngx.thread.wait
 local timers = {}
 
 
-local _M = {
-}
+local _M = {}
 
 
 local function background_timer()
--- a/apisix/timers.lua
+++ b/apisix/timers.lua
@@ -1,0 +1,90 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core = require("apisix.core")
+local process = require("ngx.process")
+local pairs = pairs
+local unpack = unpack
+local thread_spawn = ngx.thread.spawn
+local thread_wait = ngx.thread.wait
+
+
+local timers = {}
+
+
+local _M = {
+}
+
+
+local function background_timer()
+    local threads = {}
+    for name, timer in pairs(timers) do
+        core.log.info("run timer[", name, "]")
+
+        local th, err = thread_spawn(timer)
+        if not th then
+            core.log.error("failed to spawn thread for timer [", name, "]: ", err)
+        end
+
+        core.table.insert(threads, th)
+    end
+
+    local ok, err = thread_wait(unpack(threads))
+    if not ok then
+        core.log.error("failed to wait threads: ", err)
+    end
+end
+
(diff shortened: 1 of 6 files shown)


spacewander:
This PR should be draft but I hit the wrong button by accident.

apisix/timers.lua
```diff
@@ -0,0 +1,90 @@
...
+local unpack = unpack
+local thread_spawn = ngx.thread.spawn
+local thread_wait = ngx.thread.wait
+
+
+local timers = {}
+
+
+local _M = {
+}
```
Firstsawyou, line 29: I think this `local _M = {}` format will look better, what do you think?

t/plugin/log-rotate.t
```diff
@@ -95,11 +95,89 @@ GET /t
...
+
+
+=== TEST 3: fix: ensure only one timer is running
+--- config
+    location /t {
+        content_by_lua_block {
+            ngx.sleep(0.5)
+            local t = require("lib.test_admin").test
+            local code, _, org_body = t('/apisix/admin/plugins/reload',
+                                        ngx.HTTP_PUT)
```
Firstsawyou, line 114: For me, I think this line of code is not very long, we can write it on the same line, so it looks clearer. What do you think？

t/plugin/log-rotate.t
```diff
@@ -95,11 +95,89 @@ GET /t
...
+apisix:
+  node_listen: 1984
+  admin_key: null
+plugins:
+  - prometheus
+            ]]
+            require("lib.test_admin").set_config_yaml(data)
+            local t = require("lib.test_admin").test
+            local code, _, org_body = t('/apisix/admin/plugins/reload',
+                                        ngx.HTTP_PUT)
```
Firstsawyou, line 161: ditto.

t/plugin/log-rotate.t
spacewander: @Firstsawyou 
No. I like the indentation now, which I can see it is a PUT request at the first glance.
We make a new line not just because the old one is too long, but also make the important part of code easier to notice.

Firstsawyou: Okay, got it.

apisix/timers.lua
spacewander: @Firstsawyou 
Updated in the new commit.

apisix/timers.lua
```diff
@@ -0,0 +1,89 @@
...
+end
+
+
+local function is_privileged()
+    return process.type() == "privileged agent" or process.type() == "single"
+end
+
+
+function _M.init_worker()
+    -- currently only log-rotate needs the timer
```
membphis, line 57: only for plugin `log-rotate`, the current way is enough.

but I think this library can be used for different plugins, it is basic library.

so I think we should remove https://github.com/apache/apisix/pull/2596/files#diff-6ff817ed7efbdfcf2da85cb1a68a48aa2ef9e18422b80b2f255fc41032ee773eR58-R60

dabue (approved):

dabue:
I've used this new backgroud timer in my local plug-in, it worked well.

membphis (approved):

dabue (approved):
I've used the 'apisix/timers' in my local plug-in, it worked well.

apisix/timers.lua
```diff
@@ -0,0 +1,84 @@
...
+local function is_privileged()
+    return process.type() == "privileged agent" or process.type() == "single"
+end
+
+
+function _M.init_worker()
+    local timer, err = core.timer.new("background", background_timer, {check_interval = 0.5})
+    if not timer then
+        core.log.error("failed to create background timer: ", err)
+    else
```
nic-chen, line 60: we could return here，and remove 'else'

apisix/timers.lua
```diff
@@ -0,0 +1,84 @@
...
+
+
+local function background_timer()
+    local threads = {}
+    for name, timer in pairs(timers) do
+        core.log.info("run timer[", name, "]")
+
+        local th, err = thread_spawn(timer)
+        if not th then
+            core.log.error("failed to spawn thread for timer [", name, "]: ", err)
```
nic-chen, line 38: Should we use 'continue' here,since it's empty?

spacewander: @nic-chen 
Updated.

nic-chen (approved):

PR #2629, opened 2020-11-05
fix: after proxy-mirror plugin mirroring request, http version change…

### LINKED ISSUE #2618: bug(proxy-mirror): after proxy-mirror plugin mirroring request, http version changed from 1.1 to 1.0
### Issue description

The http version changed from 1.1 to 1.0 after the proxy-mirror plug-in mirroring request. I think the http version should not be changed. Problem recurrence is the same as [#2617](https://github.com/apache/apisix/issues/2617).

Firstsawyou:
…d from 1.1 to 1.0.

fix #2618

### Pre-submission checklist:

### DIFF
--- a/t/plugin/proxy-mirror.t
+++ b/t/plugin/proxy-mirror.t
@@ -32,10 +32,11 @@ add_block_preprocessor(sub {
         server_tokens off;
 
         location / {
-            content_by_lua_block {            
+            content_by_lua_block {
                 local core = require("apisix.core")
 
-                local headers_tab = ngx.req.get_headers()               
+                core.log.info("upstream_http_version: ", ngx.req.http_version())
+                local headers_tab = ngx.req.get_headers()
                 for k, v in pairs(headers_tab) do
                     core.log.info(k, ": ", v)
                 end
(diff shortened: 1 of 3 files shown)


spacewander:
Now we force the proxy HTTP request to be version 1.1.
But it is fine... Since there is no way to dynamically change proxy HTTP version without modifying Nginx core and the majority HTTP request is 1.1.

t/plugin/proxy-mirror.t
```diff
@@ -32,10 +32,11 @@ add_block_preprocessor(sub {
...
 
         location / {
-            content_by_lua_block {            
+            content_by_lua_block {
                 local core = require("apisix.core")
 
-                local headers_tab = ngx.req.get_headers()               
+                core.log.info("upstream_http_version: ", ngx.req.http_version())
+                local headers_tab = ngx.req.get_headers()
                 for k, v in pairs(headers_tab) do
```
spacewander, line 40: I run the tests locally. It seems the order of `headers_tab` is unstable?

Firstsawyou:
> Now we force the proxy HTTP request to be version 1.1.
> But it is fine... Since there is no way to dynamically change proxy HTTP version without modifying Nginx core and the majority HTTP request is 1.1.

Yes, I tried to dynamically change the proxy HTTP version, but without success.

t/plugin/proxy-mirror.t
Firstsawyou: Thank you very much for your suggestions, I will update later.

Firstsawyou: hi, @spacewander , I tried it locally. The `headers_tab` printed to the log does not need to be sorted. We need to sort the `_M.uri()` function in `t/lib/server.lua`. I think we can solve this by creating a new pr problem. What do you think?

gxthrj (approved):

membphis (approved):

membphis:
@spacewander do you have time to take a look at this PR?

Firstsawyou:
@spacewander ping

spacewander (approved):

PR #2657, opened 2020-11-08
fix: the jwt plugin does not handle the case where `exp` is empty. (#…

### LINKED ISSUE #2649: bug: The jwt plugin does not handle the case where `exp` is empty
https://github.com/apache/apisix/commit/27875e57901ff1bbb1161bbd55364ecef2990e27#r43958477

tzssangglass:
…2649)

fix #2649

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/jwt-auth.lua
+++ b/apisix/plugins/jwt-auth.lua
@@ -53,7 +53,7 @@ local consumer_schema = {
             enum = {"HS256", "HS512", "RS256"},
             default = "HS256"
         },
-        exp = {type = "integer", minimum = 1},
+        exp = {type = "integer", minimum = 1, default = 86400},
         base64_secret = {
             type = "boolean",
             default = false
@@ -121,10 +121,6 @@ function _M.check_schema(conf, schema_type)
             if not conf.private_key then
                 return false, "missing valid private key"
             end
-        end
-
-        if not conf.exp then
-            conf.exp = 60 * 60 * 24
         end
     end
 
--- a/t/plugin/jwt-auth.t
+++ b/t/plugin/jwt-auth.t
@@ -526,7 +526,7 @@ GET /t
 --- request
 GET /apisix/admin/schema/plugins/jwt-auth?schema_type=consumer
 --- response_body
-{"required":["key"],"properties":{"exp":{"minimum":1,"type":"integer"},"private_key":{"type":"string"},"public_key":{"type":"string"},"algorithm":{"type":"string","default":"HS256","enum":["HS256","HS512","RS256"]},"base64_secret":{"default":false,"type":"boolean"},"secret":{"type":"string"},"key":{"type":"string"}},"additionalProperties":false,"type":"object"}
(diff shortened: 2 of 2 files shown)


membphis (commented):
missing test case

membphis:
please update the test case

tzssangglass:
ok

apisix/plugins/jwt-auth.lua
```diff
@@ -53,7 +53,7 @@ local consumer_schema = {
             enum = {"HS256", "HS512", "RS256"},
             default = "HS256"
         },
-        exp = {type = "integer", minimum = 1},
+        exp = {type = "integer", minimum = 1, default = 86400},
```
membphis, line 56: and we need to add a test case to check the default value, we need to sure it works.

and we need a test case for this case: https://github.com/apache/apisix/issues/2649

we need to sure we fixed this bug.

tzssangglass: Do you mean that when jwt expires (`exp` = `default` value), need to verify that jwt is invalid?

t/plugin/jwt-auth.t
```diff
@@ -1140,3 +1140,48 @@ base64_secret required but the secret is not in base64 format
 [error]
 --- request
 GET /t
+
```
tzssangglass, line 1143: @membphis does this test case cover the situation you mentioned
https://github.com/apache/apisix/pull/2657#discussion_r519373662

t/plugin/jwt-auth.t
```diff
@@ -1140,3 +1140,48 @@ base64_secret required but the secret is not in base64 format
...
+
+
+=== TEST 50: when the exp value is not set,make sure the default value(86400) works
```
tzssangglass, line 1146: @membphis does this test case cover the situation you mentioned
https://github.com/apache/apisix/pull/2657#discussion_r519373662

tzssangglass, line 1146: @membphis does this test case cover the situation you mentioned
https://github.com/apache/apisix/pull/2657#discussion_r519373662

t/plugin/jwt-auth.t
```diff
@@ -1140,3 +1140,73 @@ base64_secret required but the secret is not in base64 format
 [error]
 --- request
 GET /t
+
+
+
+=== TEST 50: when the exp value is not set, make sure the default value(86400) works
```
tzssangglass, line 1146: @membphis do these two cases cover the scenarios you mention?
I'm not sure if the TEST51 is elegant enough.

t/plugin/jwt-auth.t
```diff
@@ -1140,3 +1140,73 @@ base64_secret required but the secret is not in base64 format
...
+
+
+
+=== TEST 51: when the exp value is not set, sign jwt use the default value(86400)
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body, res_data = t('/apisix/plugin/jwt/sign?key=exp-not-set',
+                ngx.HTTP_GET, nil, nil)
```
Firstsawyou, line 1197: I think it would be better not to need nil , for example: `local code, body, res_data = t('/apisix/plugin/jwt/sign?key=exp-not-set', ngx.HTTP_GET)` . What do you think？

tzssangglass: let me try

nic-chen (approved):

membphis (approved):

PR #2676, opened 2020-11-09
fix: make set_more_retries() work when upstream_type is chash

zfs123:
chash will pick next if last doesn't work to enable set_more_retries().

### Pre-submission checklist:

### DIFF
--- a/apisix/balancer/chash.lua
+++ b/apisix/balancer/chash.lua
@@ -69,11 +69,11 @@ function _M.new(up_nodes, upstream)
         upstream = upstream,
         get = function (ctx)
             local id
-            if ctx.balancer_try_count > 1 and ctx.last_server_index then
-                id, ctx.last_server_index = picker:next(ctx.last_server_index)
+            if ctx.balancer_try_count > 1 and ctx.chash_last_server_index then
+                id, ctx.chash_last_server_index = picker:next(ctx.chash_last_server_index)
             else
                 local chash_key = fetch_chash_hash_key(ctx, upstream)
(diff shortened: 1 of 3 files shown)


apisix/balancer/chash.lua
```diff
@@ -55,6 +55,7 @@ end
 
 function _M.new(up_nodes, upstream)
     local str_null = str_char(0)
+    local last_server_index
```
spacewander, line 58: The `last_server_index` should be in ctx scope, not the upstream one. We should not share the same index across different requests to an upstream.

spacewander (changes requested):

apisix/balancer/chash.lua
```diff
@@ -68,8 +68,13 @@ function _M.new(up_nodes, upstream)
     return {
         upstream = upstream,
         get = function (ctx)
-            local chash_key = fetch_chash_hash_key(ctx, upstream)
-            local id = picker:find(chash_key)
+            local id
+            if ctx.balancer_try_count > 1 and ctx.last_server_index then
+                id, ctx.last_server_index = picker:next(ctx.last_server_index)
```
spacewander, line 73: Better to use name like `ctx.chash_last_server_index`

spacewander (changes requested):

spacewander:
Please write new test to verify the fix and make sure the tests are passed.

moonming:
ping @zfs123

spacewander (approved):

tokers (approved):

Issue #2670, opened 2020-11-09
request help: while etcd auth enable，apisix cannot get etcd settings after restart 


jxhecong:
### Issue description
enable etcd auth, add right user and password in config.yaml, add a right routes, curl the service is OK
apisix restart, curl the service result 404
when remove user and password in config.yaml and disable etcd auth, the service is fine again

### Environment

* apisix version (cmd: `apisix version`): 2.0
* OS: centos7.6

spacewander:
Have you seen any error message after apisix restart?

jxhecong:
just "signal process started" in error.log

jxhecong:
someting strange:
etcd --user=***:*** get / --prefix
the command's just output /apisix/routes/1 which is configured
such as
"
/apisix/consumers/
init_dir
/apisix/global_rules/
init_dir
/apisix/node_status/
init_dir
/apisix/plugin_metadata/
init_dir
/apisix/plugins/
init_dir
/apisix/proto/
init_dir
..."
is not found

jxhecong:
1. reproduce the failure.
1.1 etcd --user=: delete / --prefix
      etcd --user=: get / --prefix, to make sure nothing left
1.2 enable etcd auth, add right user and password in config.yaml, apisix start, add a right routes, curl the service is OK
      etcd --user=: get / --prefix, just output the /apisix/routes/ which is configured
1.3 apisix restart, curl the service result 404
      etcd --user=: get / --prefix, just output the /apisix/routes/ which is configured

2. to avoid the problem
2.1 disable etcd auth, apisix restart
      etcd get / --prefix, output all KVs that apisix configured
2.2 then enable etcd auth, it's OK to restart apisix

nic-chen:
> someting strange:
> 
> etcd --user=***:*** get / --prefix
> 
> the command just output /apisix/routes/1 which is configured
> 
> such as
> 
> "
> 
> /apisix/consumers/
> 
> init_dir
> 
> /apisix/global_rules/
> 
> init_dir
> 
> /apisix/node_status/
> 
> init_dir
> 
> /apisix/plugin_metadata/
> 
> init_dir
> 
> /apisix/plugins/
> 
> init_dir
> 
> /apisix/proto/
> 
> init_dir
> 
> ..."
> 
> is not found

Thanks for feedback，that is why.
'init_etcd' doesn't support auth yet.

jxhecong:
[...]

PR #2690, opened 2020-11-10
fix(CLI): add Authorization header while enable etcd auth

jxhecong:
close https://github.com/apache/apisix/issues/2670

### Pre-submission checklist:

### DIFF
--- a/.travis/apisix_cli_test.sh
+++ b/.travis/apisix_cli_test.sh
@@ -494,3 +494,68 @@ fi
 make stop
 
 echo "passed: access log with JSON format"
+
+# check etcd while enable auth
+git checkout conf/config.yaml
+
+export ETCDCTL_API=3
+etcdctl version
+etcdctl --endpoints=127.0.0.1:2379 user add "root:apache-api6"
+etcdctl --endpoints=127.0.0.1:2379 role add root
+etcdctl --endpoints=127.0.0.1:2379 user grant-role root root
+etcdctl --endpoints=127.0.0.1:2379 user get root
+etcdctl --endpoints=127.0.0.1:2379 auth enable
+etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 del /apisix --prefix
+
+echo '
+etcd:
+  host:
+    - "http://127.0.0.1:2379"
+  prefix: "/apisix"
+  timeout: 30
+  user: root
+  password: apache-api6
+' > conf/config.yaml
+
+make init
+cmd_res=`etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 get /apisix --prefix`
+etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 auth disable
+etcdctl --endpoints=127.0.0.1:2379 role delete root
+etcdctl --endpoints=127.0.0.1:2379 user delete root
+
+init_kv=(
+/apisix/consumers/
+init_dir
+/apisix/global_rules/
+init_dir
+/apisix/node_status/
+init_dir
+/apisix/plugin_metadata/
+init_dir
+/apisix/plugins/
+init_dir
+/apisix/proto/
+init_dir
+/apisix/routes/
+init_dir
+/apisix/services/
+init_dir
+/apisix/ssl/
+init_dir
+/apisix/stream_routes/
+init_dir
+/apisix/upstreams/
+init_dir
+)
+i=0
+
+for kv in $cmd_res
+do
+    if [ "${init_kv[$i]}" != "$kv" ]; then
+        echo "failed: index=$i, $kv is not equal to ${init_kv[$i]}"
+        exit 1
+    fi
+    let i=$i+1
+done
+
+echo "passed: etcd auth enabled and init kv has been set up correctly"
--- a/bin/apisix
+++ b/bin/apisix
@@ -560,6 +560,26 @@ local function init_etcd(show_output)
     for index, host in ipairs(yaml_conf.etcd.host) do
         local is_success = true
 
+        local token_head = ""
+        if etcd_conf.user and etcd_conf.password then
+            local uri_auth = host .. "/v3/auth/authenticate"
+            local json_auth = {
+                name =  etcd_conf.user,
+                password = etcd_conf.password
+            }
+            local post_json_auth = dkjson.encode(json_auth)
+            local cmd_auth = "curl -s " .. uri_auth .. " -X POST -d '" .. post_json_auth
+                             .. "' --connect-timeout " .. timeout
+                             .. " --max-time " .. timeout * 2 .. " --retry 1 2>&1"
+
+            local res_auth = util.execute_cmd(cmd_auth)
+            local body_auth, _, err_auth = dkjson.decode(res_auth)
+            if err_auth then
+                error(cmd_auth .. "\n" .. res_auth)
+            end
+            token_head = " -H 'Authorization: " .. body_auth.token .. "'"
+        end
+
         for _, dir_name in ipairs({"/routes", "/upstreams", "/services",
                                    "/plugins", "/consumers", "/node_status",
                                    "/ssl", "/global_rules", "/stream_routes",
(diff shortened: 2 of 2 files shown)


spacewander:
Can we calculate the token from user and password?

jxhecong:
> Can we calculate the token from user and password?

you mean calculate the token in local, no need to request etcd server?

bin/apisix
```diff
@@ -560,6 +560,21 @@ local function init_etcd(show_output)
...
+            local uri_auth = host .. "/v3/auth/authenticate"
+            local post_json_auth = '{"name": "' .. etcd_conf.user .. '", "password": "' .. etcd_conf.password .. '"}'
+            local cmd_auth = "curl -s " .. uri_auth .. " -X POST -d '" .. post_json_auth
+                        .. "' --connect-timeout " .. timeout
+                        .. " --max-time " .. timeout * 2 .. " --retry 1 2>&1"
+
+            local res_auth = execute_cmd(cmd_auth)
+            local body_auth, _, err_auth = dkjson.decode(res_auth)
+            if not err_auth then
+                token_head = " -H 'Authorization: " .. body_auth.token .. "'"
```
spacewander, line 574: Need to show error message and stop the command when failed.

spacewander:
> 
> 
> > Can we calculate the token from user and password?
> 
> you mean calculate the token in local, no need to request etcd server?

I read the doc of etcd. It is impossible. There can be better than your way.

spacewander:
You need to provide a test in https://github.com/apache/apisix/blob/master/.travis/apisix_cli_test.sh like https://github.com/apache/apisix/blob/1ea912d75268e9a244aeea82397db011c766124d/t/core/etcd-auth.t#L29

bin/apisix
jxhecong: Error message shoud be show, but I think it's not necessary to stop when the command failed

jxhecong:
> You need to provide a test in https://github.com/apache/apisix/blob/master/.travis/apisix_cli_test.sh like
> 
> https://github.com/apache/apisix/blob/1ea912d75268e9a244aeea82397db011c766124d/t/core/etcd-auth.t#L29

sorry, i have no idea how to do this and why is it needed.

bin/apisix
spacewander: @jxhecong
We should stop the command when the authorization is failed. Otherwise, we won't init the etcd correctly, just like what happen now.

bin/apisix
```diff
@@ -560,6 +560,23 @@ local function init_etcd(show_output)
     for index, host in ipairs(yaml_conf.etcd.host) do
         local is_success = true
 
+        local token_head = ""
+        if etcd_conf.user and etcd_conf.password then
+            local uri_auth = host .. "/v3/auth/authenticate"
+            local post_json_auth = '{"name": "' .. etcd_conf.user .. '", "password": "' .. etcd_conf.password .. '"}'
```
juzhiyuan, line 566: Could we use some libraries or functions to stringify that object instead of combining string?

https://stackoverflow.com/questions/24908199/convert-json-string-to-lua-table

juzhiyuan: According to https://github.com/apache/apisix/pull/2690/files#diff-0cfc3cc4673fc5c2de7be20b9f8763a8955212292f08a8a7b5d1f459533f5fcdR572 , could we use `dkjson.encode`?

bin/apisix
juzhiyuan: Agree, once the authentication failed, we have to terminate the command.

bin/apisix
membphis: yes, `dkjson` is more safe.

bin/apisix
jxhecong: Done

bin/apisix
jxhecong: Done

bin/apisix
```diff
@@ -560,6 +560,27 @@ local function init_etcd(show_output)
...
+        local token_head = ""
+        if etcd_conf.user and etcd_conf.password then
+            local uri_auth = host .. "/v3/auth/authenticate"
+            local json_auth = {
+                name =  etcd_conf.user,
+                password = etcd_conf.password
+            }
+            local post_json_auth = dkjson.encode(json_auth)
+            local cmd_auth = "curl -s " .. uri_auth .. " -X POST -d '" .. post_json_auth
+                        .. "' --connect-timeout " .. timeout
```
membphis, line 572: bad indentation

bin/apisix
```diff
@@ -560,6 +560,27 @@ local function init_etcd(show_output)
...
+                        .. " --max-time " .. timeout * 2 .. " --retry 1 2>&1"
+
+            local res_auth = execute_cmd(cmd_auth)
+            local body_auth, _, err_auth = dkjson.decode(res_auth)
+            if not err_auth then
```
membphis, line 577: bad style, here is a good style, please take a look at:

```lua
if err_auth then
    error(....)
end

...
```

membphis (approved):
LGTM, except minor style things

membphis (changes requested):
missing CLI test, this is important.
we need to confirm new CLI can work fine for this case

jxhecong:
> missing CLI test, this is important.
> we need to confirm new CLI can work fine for this case

How far should be checked?
In the check environment, what is etcd's address and passwd of root user?

membphis:
> In the check environment, what is etcd's address and passwd of root user?

you can start an etcd instance with as you want by shell.

bin/apisix
jxhecong: Done

bin/apisix
jxhecong: Done

bin/apisix
```diff
@@ -560,6 +560,26 @@ local function init_etcd(show_output)
...
+            local json_auth = {
+                name =  etcd_conf.user,
+                password = etcd_conf.password
+            }
+            local post_json_auth = dkjson.encode(json_auth)
+            local cmd_auth = "curl -s " .. uri_auth .. " -X POST -d '" .. post_json_auth
+                             .. "' --connect-timeout " .. timeout
+                             .. " --max-time " .. timeout * 2 .. " --retry 1 2>&1"
+
+            local res_auth = execute_cmd(cmd_auth)
```
spacewander, line 575: Should be `util.execute_cmd`

spacewander (changes requested):

bin/apisix
jxhecong: anything else? 1 failing checks

spacewander: @jxhecong
The check failed because `execute_cmd` doesn't exist (it should be `util.execute_cmd`). Let's fix it and see what's next.

spacewander (approved):

moonming (approved):

moonming:
@membphis please take a look

.travis/apisix_cli_test.sh
```diff
@@ -494,3 +494,68 @@ fi
...
+' > conf/config.yaml
+
+make init
+cmd_res=`etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 get /apisix --prefix`
+etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 auth disable
+etcdctl --endpoints=127.0.0.1:2379 role delete root
+etcdctl --endpoints=127.0.0.1:2379 user delete root
+
+init_kv=(
+/apisix/consumers/
```
membphis, line 527: I think we only need to confirm part of it. This list may change in the future.

we can do this in a new PR.

membphis (approved):
LGTM

fix this comment later:  https://github.com/apache/apisix/pull/2690/files#r527367999

.travis/apisix_cli_test.sh
wfnuser: So the problem is which part should we confirm?

wfnuser: I see... You mean we should only check part of this key exists? But there might be more keys, and not in the same order?

PR #2696, opened 2020-11-10
chore: sort the request headers output by the _M.uri and _M.old_uri  function in t/lib/server.lua

### LINKED ISSUE #2636: chore: sort the request headers output by the _M.uri and _M.old_uri function in 't/lib/server.lua'
### Issue description

In writing test cases, when the `_M.uri()` method in `t/lib/server.lua` is used, the order of output request headers is unstable, resulting in test cases sometimes failing. We need to sort the headers obtained by `local headers = ngx.req.get_headers()` to make it stable.

Need to modify the code: https://github.com/apache/apisix/blob/master/t/lib/server.lua#L103-L106
Modification method reference:: https://github.com/apache/apisix/blob/master/t/plugin/proxy-mirror.t#L40-L49

Firstsawyou:
fix #2636

### Pre-submission checklist:

### DIFF
--- a/t/lib/server.lua
+++ b/t/lib/server.lua
@@ -101,8 +101,15 @@ function _M.uri()
     -- ngx.sleep(1)
     ngx.say("uri: ", ngx.var.uri)
     local headers = ngx.req.get_headers()
-    for k, v in pairs(headers) do
-        ngx.say(k, ": ", v)
+
+    local headers_key = {}
+    for k, _ in pairs(headers) do
+        table.insert(headers_key, k)
+    end
+    table.sort(headers_key)
+
+    for _, v in pairs(headers_key) do
+        ngx.say(v, ": ", headers[v])
     end
 end
 _M.uri_plugin_proxy_rewrite = _M.uri
@@ -112,8 +119,15 @@ function _M.old_uri()
     -- ngx.sleep(1)
     ngx.say("uri: ", ngx.var.uri)
     local headers = ngx.req.get_headers()
-    for k, v in pairs(headers) do
-        ngx.say(k, ": ", v)
+
+    local headers_key = {}
+    for k, _ in pairs(headers) do
+        table.insert(headers_key, k)
+    end
+    table.sort(headers_key)
+
+    for _, v in pairs(headers_key) do
+        ngx.say(v, ": ", headers[v])
     end
 end
 
--- a/t/lib/server.lua
+++ b/t/lib/server.lua
@@ -103,12 +103,12 @@ function _M.uri()
     local headers = ngx.req.get_headers()
 
     local keys = {}
-    for k, _ in pairs(headers) do
+    for k in pairs(headers) do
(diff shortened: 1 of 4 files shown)


t/lib/server.lua
```diff
@@ -101,8 +101,15 @@ function _M.uri()
     -- ngx.sleep(1)
     ngx.say("uri: ", ngx.var.uri)
     local headers = ngx.req.get_headers()
-    for k, v in pairs(headers) do
-        ngx.say(k, ": ", v)
+
+    local keys = {}
+    for k, _ in pairs(headers) do
```
tokers, line 106: Just ignore the second dumb variable if you don't need it.

```lua
for k in pairs(headers) do
      ......
end
```

t/lib/server.lua
```diff
@@ -101,8 +101,15 @@ function _M.uri()
...
+        table.insert(keys, k)
+    end
+    table.sort(keys)
+
+    for _, key in pairs(keys) do
```
tokers, line 111: Use `ipairs` here since `keys` is an array-like table.

t/lib/server.lua
Firstsawyou: nice.

Firstsawyou: fixed.

t/lib/server.lua
Firstsawyou: fixed.

t/plugin/hmac-auth.t
```diff
@@ -1246,16 +1246,16 @@ location /t {
 GET /t
 --- response_body
 uri: /uri
-host: 127.0.0.1
+content-length: 52
 content-type: application/x-www-form-urlencoded
-x-real-ip: 127.0.0.1
+host: 127.0.0.1
 user-agent: lua-resty-http/0.14 (Lua) ngx_lua/10017
```
nic-chen, line 1252: a little confuse, why `user-agent` is in this place ?

Firstsawyou: When requesting the URL of `/uri`, all request headers will be output. So it is normal to have the request header `user-agent`.

spacewander: @Firstsawyou 
Can we omit the `user-agent`? Since this field will change across different versions.

nic-chen: > When requesting the URL of `/uri`, all request headers will be output. So it is normal to have the request header `user-agent`.

I mean it is not sorted in order.

Firstsawyou: > > When requesting the URL of `/uri`, all request headers will be output. So it is normal to have the request header `user-agent`.
> 
> I mean it is not sorted in order.

It has been sorted, `user-agent` is after `host`.  ^ _ ^

spacewander (approved):

nic-chen (approved):

PR #2724, opened 2020-11-12
docs: modify batch-requests.md

idbeta:
`--d` should be `--data`


juzhiyuan (approved):
Good catch and looks good to me!

https://ec.haxx.se/http/http-post

spacewander (approved):

Issue #2713, opened 2020-11-12
request help: How to turn off access log?


fregie:
### Issue description
My apisix run in docker and It's may be dangerous if access.log keep logging.
I want to turn off the access log.

idbeta:
I think turn the access log off is not a good idea.

fregie:
I have written a plugin to upload access log to my log service,so I don't need nginx save a access.log in container.
I think it's ncecssary providing a switch of access log.

spacewander:
@fregie 
Pull request is welcome!

PR #2766, opened 2020-11-16
feat: add a switch of access log

HelloBug0:
satify the need of [#2713](https://github.com/apache/apisix/issues/2713)

### Pre-submission checklist:

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -128,7 +128,7 @@ nginx_config:                     # config for render the template to genarate n
   #envs:                            # allow to get a list of environment variables
   #  - TEST_ENV
   http:
-    access_log_switch: on          # on or off, default on
+    enable_access_log: true
     access_log: "logs/access.log"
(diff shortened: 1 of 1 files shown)


conf/config-default.yaml
```diff
@@ -128,6 +128,7 @@ nginx_config:                     # config for render the template to genarate n
   #envs:                            # allow to get a list of environment variables
   #  - TEST_ENV
   http:
+    access_log_switch: on          # on or off, default on
```
spacewander, line 131: Better to use `enable_access_log: true` to ensure consistency.

spacewander:
#2765 should be put into this one.

conf/config-default.yaml
HelloBug0: yeah, right

HelloBug0:
> #2765 should be put into this one.

It occurs to me, but it's kind of awkward that I still don't know how to PR once with two changed files in windows even after I searched that in the Internet. I realize that I need to learn more about git operations.

PR #2769, opened 2020-11-17
feat: enable access log or not

HelloBug0:
satify the need of [#2713](https://github.com/apache/apisix/issues/2713)

### Pre-submission checklist:

* [X] Is this PR backward compatible?


conf/config-default.yaml
```diff
@@ -128,6 +128,7 @@ nginx_config:                     # config for render the template to genarate n
   #envs:                            # allow to get a list of environment variables
   #  - TEST_ENV
   http:
+    enable_access_log: true
```
tokers, line 131: For the compatibilities' sake, the access_log should be enabled even if this field absent.

membphis: agree

membphis:
missing test case

conf/config-default.yaml
HelloBug0: OK, I will update this PR later.

HelloBug0:
> missing test case

OK, I will add it. Would you like to tell me that is it OK if I add a new test file in the directory https://github.com/apache/apisix/tree/master/t/config-center-yaml? Or should I add some test cases in an existing file?

spacewander:
@HelloBug0 
You can add it in https://github.com/apache/apisix/blob/master/.travis/apisix_cli_test.sh

.travis/apisix_cli_test.sh
```diff
@@ -391,6 +391,52 @@ fi
 
 git checkout conf/config.yaml
```
spacewander, line 392: We don't need to checkout in the middle of a group of test, as we will change it later.

.travis/apisix_cli_test.sh
```diff
@@ -391,6 +391,52 @@ fi
...
+    exit 1
+fi
+
+grep "access_log off;" conf/nginx.conf > /dev/null
+if [ $? -eq 0 ]; then
+    echo "failed: nginx.conf file find access_log off; when enable access log"
+    exit 1
+fi
+
+git checkout conf/config.yaml
```
spacewander, line 415: ditto

membphis:
@HelloBug0 CI failed, please take a look at the output of CI

.travis/apisix_cli_test.sh
```diff
@@ -389,12 +389,86 @@ if [ ! $? -eq 0 ]; then
...
+echo '
+nginx_config:
+  http:
+    enable_access_log: false
+    access_log_format: "$remote_addr - $remote_user [$time_local] $http_host test_enable_access_log"
+' > conf/config.yaml
+
+make init
+
+grep "test_enable_access_log" conf/nginx.conf > /dev/null
```
spacewander, line 440: Use `grep -c xx || true` for this line?

HelloBug0: Do you know the reason why this test case is failed? It is a little bit confusing~

spacewander: Because there is `set -e` at the top of the file.

HelloBug0:
Finally, all checks have passed, whoopee~

.travis/apisix_cli_test.sh
HelloBug0: Oh, I see. Thank you:smile:

membphis (approved):
LGTM

spacewander (approved):

.travis/apisix_cli_test.sh
```diff
@@ -389,12 +389,86 @@ if [ ! $? -eq 0 ]; then
...
+    exit 1
+fi
+
+make run
+sleep 0.1
+curl http://127.0.0.1:9080/hi
+sleep 4
+tail -n 1 logs/access.log > output.log
+
+count_grep=`grep -c "test_enable_access_log_false" output.log || true`
```
tokers, line 458: We should remove the output.log after testing.

PR #2789, opened 2020-11-18
fix: log `index` function should use `self` instead of `m` or `_M` 

juzipeek:

 It is more reasonable to use `self` variables than to use `_M` or `m` directly.
### Pre-submission checklist:

### DIFF
--- a/apisix/core/log.lua
+++ b/apisix/core/log.lua
@@ -59,7 +59,7 @@ function _M.new(prefix)
 
         -- cache the lazily generated method in our
         -- module table
-        m[cmd] = method
+        self[cmd] = method
         return method
     end})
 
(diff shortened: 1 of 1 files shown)


apisix/core/log.lua
```diff
@@ -82,7 +82,7 @@ setmetatable(_M, {__index = function(self, cmd)
 
     -- cache the lazily generated method in our
     -- module table
-    _M[cmd] = method
+    self[cmd] = method
```
spacewander, line 85: Ditto

apisix/core/log.lua
```diff
@@ -59,7 +59,7 @@ function _M.new(prefix)
 
         -- cache the lazily generated method in our
         -- module table
-        m[cmd] = method
+        self[cmd] = method
```
spacewander, line 62: The `m` and `self` are the same in this situation.

spacewander (changes requested):

spacewander:
I don't think we need to waste too much time on such "readable" since everyone has her difference standard.

Issue #1838, opened 2020-07-13
bug: discovery Eureka cannot automatically proxy to a new node


Hmemories:
### Issue description
register a new instance to Eureka,but can't proxy to this instance

### Environment

* apisix version (cmd: `apisix version`): 1.4
* OS:mac os

### Minimal test code / Steps to reproduce the issue
1.
2.
3.

### What's the actual result? (including assertion message & call stack if applicable)
oldnode    host":"172.16.50.93","port":8888
newnode  host":"172.16.50.93","port":8081
apisix balance proxy to oldnode not to newnode
after make reload it‘s OK

membphis:
please make a try with this PR: https://github.com/apache/incubator-apisix/pull/1835/files

I think this PR should fix this bug.

Hmemories:
![image](https://user-images.githubusercontent.com/30172398/87372843-1e87e480-c5bb-11ea-9cb1-e0fa4dd7c80a.png)
I tried with this PR and copied the code that different from main to my own branch ,but  it always proxy to old

Hmemories:
2020/07/14 10:37:02 [info] 60274#3150962: *173669 [lua] balancer.lua:193: pick_server(): [{"host":"172.16.50.93","port":8889,"metadata":{"management.port":"8889","commit-time":"2019-08-27T14:48:42+0800","git-short-commitid":"11efbdd","commit-branch":"dev","jar-version":"0.10.2","msv-version":"1.7.2","commit-message":"Merge branch 'apollo-generator' into 'dev'","git-commitid":"11efbdd805134d9e5ff7a0cd41aea0f6241c9602"},"weight":100},{"host":"172.16.50.93","port":8081,"metadata":{"management.port":"8081"},"weight":100}] while connecting to upstream, client: ::1, server: , request: "GET /atel/all/cmdb/apps HTTP/1.1", host: "localhost:9080"
[...]

PR #2793, opened 2020-11-19
modify balancer.lua to solve issue that discovery Eureka cannot autom…

Hmemories:
…atically proxy to a new node

to solve issue that discovery Eureka cannot automatically proxy to a new node #1838
### Pre-submission checklist:

When we are getting the cached data, if we match "key" with the data"md5" of up_conf.nodes, we can find the changes of the service.

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -19,6 +19,7 @@ local require     = require
 local discovery   = require("apisix.discovery.init").discovery
 local balancer    = require("ngx.balancer")
 local core        = require("apisix.core")
+local ngx = ngx
 local ipairs      = ipairs
 local tostring    = tostring
 local set_more_tries   = balancer.set_more_tries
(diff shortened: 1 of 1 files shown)


apisix/balancer.lua
```diff
@@ -218,8 +219,9 @@ local function pick_server(route, ctx)
     if checker then
         version = version .. "#" .. checker.status_ver
     end
+    local cache_key = key..ngx.md5(up_conf.nodes)
```
tokers, line 222: style: need space on both left and right side of `..`.

tokers:
@Hmemories CI run is failed, please take a look :).

tokers:
@Hmemories Missing test cases.

apisix/balancer.lua
tokers: The argument type of `ngx.md5` is string while `up_conf.nodes` is an array-like table.

BTW, i don't think calculating the md5sum of nodes is a good way to differentiate the version, it's heavy, what about introducing the version mechanism for `up_conf.nodes`, either version number or last modify time is OK.

tokers: Or settling for the second best, just use the up_conf.version.

Hmemories: 嗯嗯，我准备在看看，这部分代码已经被我是用在了生产环境我在考虑有没有好的方式优化它

Hmemories: 如果md5比较重的话，hash是不是也可以

moonming: md5 is good to me

moonming:
@Hmemories it will be perfect if you can add test cases

membphis:
ping @Hmemories

PR #2802, opened 2020-11-20
refactor: separate admin and proxy port in default config

ShiningRush:
Recently, we have a serious security vulnerability caused by `adminapi`.
I synchronize  scenarios here:
- Some external business team forget to modify admin port and key after debugging(the ip whitelist is removed), so that anyone can access the `adminapi`
- Some people using the default admin key to scan `adminapi` and found the command execution vulnerability, they reported it to the our company's security platform

I think there are two points we can optimize:
- the default configuration is to separate proxy and admin port
- `adminapi` command execution vulnerability.I will push a PR to apisix after I produce and fix the vulnerability


spacewander (approved):

tokers (approved):

spacewander:
Note: merge it after 2.1 is released and works well.

juzhiyuan (commented):
🤔 Just to make sure, I only see there have docs changed, does it will resolve the vulnerable?

spacewander:
@juzhiyuan 
` conf/config-default.yaml` is changed.

juzhiyuan:
> @juzhiyuan 
> ` conf/config-default.yaml` is changed.

Oh yes!

juzhiyuan (approved):

PR #2824, opened 2020-11-23
chore: sort default plugin list for easier search

spacewander:
### Pre-submission checklist:

### DIFF
--- a/apisix/admin/plugins.lua
+++ b/apisix/admin/plugins.lua
@@ -22,7 +22,6 @@ local pairs     = pairs
 local ipairs    = ipairs
 local pcall     = pcall
 local type      = type
-local table_remove = table.remove
 local table_sort = table.sort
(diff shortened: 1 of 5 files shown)


membphis (approved):

apisix/admin/plugins.lua
```diff
@@ -125,10 +124,6 @@ end
 
 function _M.get_plugins_list()
     local plugins = core.config.local_conf().plugins
-    if plugins[1] == 'example-plugin' then
```
moonming, line 128: @juzhiyuan I’m not sure if delete it will affect the display of the front-end, for example, this plugin will be displayed on the interface

juzhiyuan: No, it will not effect the frontend any more.

moonming (approved):

Issue #2800, opened 2020-11-20
test(CLI): only need to confirm part of result, returned list may change in the future.


membphis:
original comment: https://github.com/apache/apisix/pull/2690/files#r527367999

membphis:
@ jxhecong would like to this this issue?

wfnuser:
I would like to have a try.😄

membphis:
many thx, good luck. if you need any help, pls let us know

wfnuser:
I'm going through the whole project background. I'm not so sure what 'init_kv=(' and 'apisix_cli_test.sh ' for, yet. Where can I figure it out quickly. Would you give me some clues, that will be a big help.

membphis:
> I'm going through the whole project background. I'm not so sure what 'init_kv=(' and 'apisix_cli_test.sh ' for, yet. Where can I figure it out quickly. Would you give me some clues, that will be a big help.

@jxhecong need your help ^_^

wfnuser:
I think what @membphis means in the original comment: https://github.com/apache/apisix/pull/2690/files#r527367999 is, that there might be more routes in $cmd_res, we should only check part of the key-value pairs are correct.
So I would like to use a hash map in shell supported by shell 4. Is that ok?

membphis:
> So I would like to use a hash map in shell supported by shell 4. Is that ok?

I think a regular matching is enough and simpler.

wfnuser:
> > So I would like to use a hash map in shell supported by shell 4. Is that ok?
> 
> I think a regular matching is enough and simpler.

Not so sure how to use regular matching here. 😂

spacewander:
Solved.

PR #2846, opened 2020-11-25
feat: confirm only part of the etcd kvs

wfnuser:

we only need to confirm part of result, returned list may change in the future.
#2800 

### Pre-submission checklist:


membphis (changes requested):
you can make a try with this way, it is simpler way.

```sh
make init
cmd_res=`etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 get /apisix --prefix`
etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 auth disable
etcdctl --endpoints=127.0.0.1:2379 role delete root
etcdctl --endpoints=127.0.0.1:2379 user delete root

count=`echo $cmd_res | grep -c "/apisix/global_rules" || true`
if [ $count -ne 1 ]; then
    echo "failed: failed to match '/apisix/global_rules'"
    exit 1
fi

wfnuser:
> you can make a try with this way, it is simpler way.
> 
> ```shell
> make init
> cmd_res=`etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 get /apisix --prefix`
> etcdctl --endpoints=127.0.0.1:2379 --user=root:apache-api6 auth disable
> etcdctl --endpoints=127.0.0.1:2379 role delete root
> etcdctl --endpoints=127.0.0.1:2379 user delete root
> 
> count=`echo $cmd_res | grep -c "/apisix/global_rules" || true`
> if [ $count -ne 1 ]; then
>     echo "failed: failed to match '/apisix/global_rules'"
>     exit 1
> fi
> ```

Thx. This is really the simpler way. 
And I had some misunderstanding before. Actually we should check that there is some part of the ETCD keys must exists? If the particular key-value not exists in ETCD we should fail it.

membphis:
Yes, you are right.

wfnuser:
> Yes, you are right.

The pr is ready now.😄

membphis (approved):

wfnuser:
Can we merge this pull request now? It seems that we need another approving review. @moonming

membphis:
> Can we merge this pull request now? It seems that we need another approving review. @moonming

only 1 approval is not enough. we should wait for more time.

moonming (approved):

Issue #2857, opened 2020-11-26
bug: authz-keycloak plugin should run in access phase.


jenskeiner:
### Issue description
The `authz-keycloak` plugin currently runs in the Openresty *rewrite* lifecycle phase. This is in line with the plugin development guidelines: "In APISIX, the authentication logic is implemented in the rewrite phase." However, this means it will run before most, if not all, other authentication plugins, e.g. `openid-connect`.

I also think that authentication and authorization should both be done in the *access * phase which seems the most appropriate.

I stumbled upon this when trying to set up authentication via OIDC (provided by a Keycloak instance) together with additional authorization (again via the same Keycloak instance). It seems appropriate to handle both in separate plugins, first authentication via `openid-connect` and then authorization via `authz-keycloak`. But since the used plugins run in different phases, they run in the wrong order.

Wrong order here means that before the `openid-connect` plugin gets a chance to obtain an access token on behalf of the user via the Relying Party flow -- in a situation where no token is provided in the original incoming request -- the `authz-keycloak` plugin will already reject the request due to the missing token.

A related issue, but one that should be fixed separately in the `openid-connect` plugin, is that the access token, if it is contained in a session cookie header, is not added in an "Authorization" header. So even if `authz-keycloak` would be running after `openid-connect`, it would not be able to extract the token if it was contained in a session cookie in the original request. A separate issue will be created to look into this.

### Environment

* apisix version (cmd: `apisix version`): 2.0
* OS: Alpine Linux Docker image

### Minimal test code / Steps to reproduce the issue
1. Create a route with plugins `openid-connect` and `authz-keycloak` appropriately configured.
2. Verify that `authz-keycloak` runs before `openid-connect`.

### What's the actual result? (including assertion message & call stack if applicable)
The result is explained above. A route that is protected only through `openid-connect` may work correctly with the plugin  automatically redirecting to the authentication endpoint if no access token is provided in the request. Adding `authz-keycloak` to the route breaks that process by rejecting immediately when no token is found in the request.

### What's the expected result?
The `authz-keycloak` plugin should run in the *access* phase so that it always runs after `openid-connect`. The respective plugin priorities are already ok, it's just that `authz-keycloak` needs to run in a different phase.

PR #2859, opened 2020-11-26
fix: authz-keycloak plugin should not be marked as auth-type and should run in access phase, basic-auth plugin should run in rewrite phase.

jenskeiner:
See #2857.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/basic-auth.lua
+++ b/apisix/plugins/basic-auth.lua
@@ -125,7 +125,7 @@ do
     end
 end
 
-function _M.access(conf, ctx)
+function _M.rewrite(conf, ctx)
     core.log.info("plugin access phase, conf: ", core.json.delay_encode(conf))
 
     -- 1. extract authorization from header
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -49,6 +49,7 @@ local schema = {
 local _M = {
     version = 0.1,
     priority = 2599,
+    type = 'auth',
     name = plugin_name,
     schema = schema,
 }
--- a/doc/plugin-develop.md
+++ b/doc/plugin-develop.md
@@ -168,8 +168,8 @@ local _M = {
 Determine which phase to run, generally access or rewrite. If you don't know the [Openresty life cycle](https://openresty-reference.readthedocs.io/en/latest/Directives/), it's
 recommended to know it in advance. For example key-auth is an authentication plugin, thus the authentication should be completed
 before forwarding the request to any upstream service. Therefore, the plugin can be executed in the rewrite and access phases.
-In APISIX, the authentication logic is implemented in the access phase. Generally, IP access and interface
-permission are also completed in the access phase.
+In APISIX, the authentication logic is implemented in the rewrite phase. Generally, IP access and interface
+permission are completed in the access phase.
 
 The following code snippet shows how to implement any logic relevant to the plugin in the Openresty log phase.
 
--- a/doc/plugin-develop.md
+++ b/doc/plugin-develop.md
@@ -168,8 +168,8 @@ local _M = {
 Determine which phase to run, generally access or rewrite. If you don't know the [Openresty life cycle](https://openresty-reference.readthedocs.io/en/latest/Directives/), it's
 recommended to know it in advance. For example key-auth is an authentication plugin, thus the authentication should be completed
 before forwarding the request to any upstream service. Therefore, the plugin can be executed in the rewrite and access phases.
-In APISIX, the authentication logic is implemented in the rewrite phase. Generally, IP access and interface
-permission are completed in the access phase.
+In APISIX, the authentication logic is implemented in the access phase. Generally, IP access and interface
+permission are also completed in the access phase.
 
 The following code snippet shows how to implement any logic relevant to the plugin in the Openresty log phase.
 
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -513,7 +513,7 @@ function _M.http_access_phase()
                 api_ctx.consumer,
                 api_ctx
             )
-            core.log.debug("find consumer ", api_ctx.consumer.username,
+            core.log.info("find consumer ", api_ctx.consumer.username,
                            ", config changed: ", changed)
             if changed then
                 core.table.clear(api_ctx.plugins)
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -513,6 +513,7 @@ function _M.http_access_phase()
                 api_ctx.consumer,
                 api_ctx
             )
+            core.log.debug("find consumer ", api_ctx.consumer.username, ", config changed: ", changed)            
             if changed then
                 core.table.clear(api_ctx.plugins)
                 api_ctx.plugins = plugin.filter(route, api_ctx.plugins)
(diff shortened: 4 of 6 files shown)


membphis:
Authentication type plug-ins should run in the rewrite phase

jenskeiner:
> Authentication type plug-ins should run in the rewrite phase

How do you suggest this get's fixed then when `openid-connect` and other authentication plugins currently run in the access phase?

The plugin development guidelines are currently not in line with the actual code base.

Also, what's the rationale for running authentication/authorization in the *rewrite* phase when the *access* phase is the one that specifically caters for that purpose? It's not obvious why you would run in a different phase.

membphis:
reply you later

need more time to read your issue and pr first

spacewander:
@membphis @jenskeiner 
`authz-keycloak` is for authorization but not authentication. It doesn't interact with consumers, so I think this change is fine.

doc/plugin-develop.md
```diff
@@ -168,8 +168,8 @@ local _M = {
 Determine which phase to run, generally access or rewrite. If you don't know the [Openresty life cycle](https://openresty-reference.readthedocs.io/en/latest/Directives/), it's
 recommended to know it in advance. For example key-auth is an authentication plugin, thus the authentication should be completed
 before forwarding the request to any upstream service. Therefore, the plugin can be executed in the rewrite and access phases.
-In APISIX, the authentication logic is implemented in the rewrite phase. Generally, IP access and interface
-permission are completed in the access phase.
+In APISIX, the authentication logic is implemented in the access phase. Generally, IP access and interface
```
spacewander, line 171: We don't need to change it as authz-keycloak is for authorization.

membphis:
> @membphis @jenskeiner
> `authz-keycloak` is for authorization but not authentication. It doesn't interact with consumers, so I think this change is fine.

many thx for your explain. @spacewander 
we can continue this PR

doc/plugin-develop.md
jenskeiner: Ok, I can revert this one. However, note that some authentication plugins currently run in the access phase, not the rewrite one. So the document doesn’t reflect the current state correctly, I believe.

spacewander: @jenskeiner
We can fix it later, but in those plugins, not the documentation.

spacewander: @jenskeiner
Thank you for the problem you point out. I just search for all plugins which has `type = 'auth'` (excluding the authz-keycloak) and find out `basic-auth` doesn't run the phase correctly.

spacewander: The reason why we need to run authentication plugins in the rewrite phases:
https://github.com/apache/apisix/blob/f4161d39a495879edffbc75e2099f6e487bae81d/apisix/init.lua#L508-L520
And:
https://github.com/apache/apisix/blob/f4161d39a495879edffbc75e2099f6e487bae81d/apisix/admin/consumers.lua#L50

Plugins has `type = 'auth'` should run in rewrite phase and set up the consumer for current api_ctx.

jenskeiner: Ok, understood why you need to run in the rewrite phase. Then I would propose to change the phase for all authentication-type plugins to *rewrite*, instead of adjusting the phase for `authz-keycloak` from *rewrite* to *access*. Also, I think it would make the plugin development guidelines clearer if the reason they need to run in the rewrite phase was pointed out more clearly.

If you're fine with this, I can update this PR accordingly. Just let me know what you think.

apisix/plugins/basic-auth.lua
```diff
@@ -125,7 +125,7 @@ do
     end
 end
 
-function _M.access(conf, ctx)
+function _M.rewrite(conf, ctx)
```
spacewander, line 128: Need a test case to check if the `ctx.consumer` set by this plugin can be used by other plugin runs in access phase.

doc/plugin-develop.md
spacewander, line 171: Remember to revert this change.

jenskeiner: > We don't need to change it as authz-keycloak is for authorization.

Reverted.

apisix/plugins/basic-auth.lua
jenskeiner: Happy to add a test case. I don't have much experience writing them yet, so if you could point to a similar example, that would be helpful.

doc/plugin-develop.md
jenskeiner: Done.

apisix/plugins/basic-auth.lua
spacewander: You can add a stub:
```diff
diff --git apisix/init.lua apisix/init.lua
index 7333dd1..411cdd4 100644
--- apisix/init.lua
+++ apisix/init.lua
@@ -514,6 +514,10 @@ function _M.http_access_phase()
                 api_ctx.consumer,
                 api_ctx
             )
+
+            core.log.info("find consumer ", api_ctx.consumer.username,
+                          ", config changed: ", changed)
+
             if changed then
                 core.table.clear(api_ctx.plugins)
                 api_ctx.plugins = plugin.filter(route, api_ctx.plugins)
```

And then check it in:
https://github.com/apache/apisix/blob/e66dd9a314e0b545e412f806b8b8d98f3b1a8f2d/t/plugin/basic-auth.t#L192

Add something like:
```
--- error_log
find consumer foo
```

jenskeiner: Ok, I think I've seen an example in the test cases for `key-auth`. I'll try to add test cases acceding for `basic-auth`.

membphis, line 128: why do we need to change another plugin?

one PR for one thing, we should only update something about plugin `authz-keycloak`

please revert this code

apisix/plugins/openid-connect.lua
```diff
@@ -49,6 +49,7 @@ local schema = {
 local _M = {
     version = 0.1,
     priority = 2599,
+    type = 'auth',
```
membphis, line 52: ditto

you can create a new PR to fix this

membphis (changes requested):

apisix/plugins/openid-connect.lua
jenskeiner: I'm sorry but I currently can't afford the time to split this into separate PRs. I think the change set is reasonably small to be considered in one PR. If you prefer more fine-grained PRs, then please consider creating these yourself, e.g. based on the changes in this branch.

membphis: I found you have submitted new PR: https://github.com/apache/apisix/pull/2905/files

Many thanks for your nice job.

Small PR is important, it easy to find problems and we can merge it asap when it is small.

By the way, it is a clearer way to show what we are doing.

Many thx for nice contribution again @jenskeiner

jenskeiner: Sure, you're very welcome. Thanks for the info. I'll try to put the remaining outstanding changes from this one into a separate new PR and then we take it from there.

jenskeiner: Have created #2910 which should contain the outstanding changes from this PR.

jenskeiner:
Closing this one as it has since been superseded by other PRs addressing the same issues.

PR #2869, opened 2020-11-27
fix: start server failure

willmafh:
When you shutdown the machine directly, the logs/nginx.pid won't be deleted.
After rebooting and run `make run`, it reports “APISIX is running...” but actually it's not.


Makefile
```diff
@@ -86,6 +86,8 @@ run: default
 ifeq ("$(wildcard logs/nginx.pid)", "")
 	mkdir -p logs
 	$(OR_EXEC) -p $$PWD/ -c $$PWD/conf/nginx.conf
+else ifeq ("$(shell lsof -p $(shell cat logs/nginx.pid))", "")
+	$(OR_EXEC) -p $$PWD/ -c $$PWD/conf/nginx.conf
```
spacewander, line 90: Would be better if we can output the reason to start APISIX even the `nginx.pid` exists.

Firstsawyou: agree +1.

willmafh: Actually we can't...The only thing we know for sure is nginx.pid exists but there's no corresponding process...Also when a user execute `make run`, **the implication** here is that he believes the apisix server is not running, but if it does running, telling him that the server is running, just like the way we do now...But if it doesn't, then we should just start the new server process since the user doesn't care about it...

spacewander: @willmafh 
Yes, our suggestion is to output some message that `nginx.pid exists but there's no corresponding process`, and then start the new one.

willmafh: @spacewander I still don't think it's necessary to output some message here...We can see [bin/apisix](https://github.com/apache/apisix/blob/ce1d40b42f056f46b20386e1937fc81ba69e84c6/bin/apisix#L348) start function, their logic are actually the same...If we really want to output some message here, then the start function of  bin/apisix should also be changed, but it's really unnecessary.

willmafh: @membphis @moonming

membphis: `nginx.pid exists but there's no corresponding process`

We should add this tip in both `Makefile` and `bin/apisix`, it is more user-friendly.

I created a new issue https://github.com/apache/apisix/issues/2948 , welcome to fix this issue.

membphis:
we need some test cases about this change

willmafh:
> we need some test cases about this change

I don't know how to add test cases in this condition. But I do test all situation on my local machine.

membphis:
We mainly use the shell scripts for this case

membphis:
You can take a look at the existing test cases

membphis:
@willmafh need to resolve Conflicting files

spacewander:
The issue is already solved by others.

PR #2886, opened 2020-11-28
feat: provide the ability to view plug-in types

### LINKED ISSUE #2308: request help: Provide the ability to view plug-in types
### Issue description
some auth plugins has a type named "auth", Identifies that this plugin is used for authentication

image

I need to know the type of plugin, so can make some judgments based on the type.

### Environment

* apisix version: 1.5
* OS:

tzssangglass:

provides an API to get the plug-in's `name`, `priority`, `priority`, `schema`, `consumer_schema` and `version`. 
all plugins are applicable, including custom plugins

fix #2308 
### Pre-submission checklist:


tzssangglass:
@membphis The error is reported in CI, but not when I run it locally, what conditions do I need to add to my local test environment to reproduce this?
https://github.com/apache/apisix/pull/2886/checks?check_run_id=1468996023#step:11:163

membphis:
> @membphis The error is reported in CI, but not when I run it locally, what conditions do I need to add to my local test environment to reproduce this?
> https://github.com/apache/apisix/pull/2886/checks?check_run_id=1468996023#step:11:163

it seems you have fixed this issue, all right?

tzssangglass:
yes, i get the CI tip.

doc/zh-cn/admin-api.md
```diff
@@ -723,3 +723,15 @@ Content-Type: text/plain
...
+
+*地址*：/apisix/admin/plugins/meta_attributes
+
+*说明*: 插件元属性，包括插件的 `name`, `priority`, `priority`, `schema`, `consumer_schema` and `version`。
+
+> 请求方法:
+
+|Method   |请求 URI|请求 body|说明        |
+|---------|-------------------------|--|------|
+|GET      |/apisix/admin/plugins/meta_attributes/{plugin_name}|无|获取资源|
```
membphis, line 735: I do not like this way(use a new API style)

We can add a new argument to the old API for this feature.

`/apisix/admin/schema/plugins/limit-count?type=priority`
`/apisix/admin/schema/plugins/limit-count?type=schema`
`/apisix/admin/schema/plugins/limit-count?type=consumer_schema `
`/apisix/admin/schema/plugins/limit-count?type=version`

we talked about this at Github issue or email list, but I can not find them :(

membphis (changes requested):

doc/zh-cn/admin-api.md
tzssangglass: i think to adopt a new API to return the meta attributes of all plugins, avoiding dashboard calls many times.
but perhaps the dashboard doesn't have to load all the plugin's meta properties at once, but only performs the loading when needed.

on the basis of the above example, if there is no type parameter, the default is to return the plugin's `type`,  `priority`, `schema`, `consumer_schema` and `version`. 
```
/apisix/admin/schema/plugins/limit-count

{
    "priority":
    "schema":
    "consumer_schema":
    "version":
    "type":
}
```
how ablot this?

membphis: We cannot change the style of old API.  

 So we add new argument and return the different value.

juzhiyuan:
```
/apisix/admin/schema/plugins/limit-count?type=priority
```

When I see this API, I will think it is used to fetch the schema for the target plugin, but what is the type? Oh! it's used to fetch the metadata according to type 🤔 confused for me. We could use this way to compatible with the old API style now, but do we have any plan to improve it? more like a standard RESTful API?

juzhiyuan: BTW,

```
/apisix/admin/plugins/{plugin_name}/meta_attributes
```

would be more RESTful.

We could use the old way to keep the style currently if possiable.

tzssangglass: > BTW,
> 
> ```
> /apisix/admin/plugins/{plugin_name}/meta_attributes
> ```
> 
> would be more RESTful.
> 
> We could use the old way to keep the style currently if possiable.

this style looks great!

juzhiyuan: 🤔 Emma then let's use the old way in 2.x? We could use the more RESTful API in breaking changes IMO, what do you think? @moonming @liuxiran @ShiningRush

moonming: > ```
> 
> /apisix/admin/schema/plugins/limit-count?type=priority
> 
> ```
> 
> 
> 
> When I see this API, I will think it is used to fetch the schema for the target plugin, but what is the type? Oh! it's used to fetch the metadata according to type 🤔 confused for me. We could use this way to compatible with the old API style now, but do we have any plan to improve it? more like a standard RESTful API.

I don't like this style too.
We need restful API

moonming: > 🤔 Emma then let's use the old way in 2.x? We could use the more RESTful API in breaking changes IMO, what do you think? @moonming @liuxiran @ShiningRush 

Metadata and schema are two different things, and should not be mixed in the same API

spacewander: Why not `/apisix/admin/plugins/{plugin_name}`? We don't need `meta_attributes` since we return everything from the plugin.

liuxiran: Agree with metadata and schema are two different things, so for the schema, the old API could just remain the same.

and for the `meta_attributes`, IMHO, `GET /apisix/admin/plugins/{plugin_name}` is more simple😊

`GET /apisix/admin/plugins/{plugin_name}` for get {plugin_name}'s meta_attrs
`GET /apisix/admin/plugins` for get all plugins' meta_attrs

juzhiyuan: > Why not `/apisix/admin/plugins/{plugin_name}`? We don't need `meta_attributes` since we return everything from the plugin.

This is also a good choice 👏

juzhiyuan: > Why not `/apisix/admin/plugins/{plugin_name}`? We don't need `meta_attributes` since we return everything from the plugin.

This is also a good choice 👏

juzhiyuan: 🤔

```
GET /apisix/admin/plugins/{plugin_name}
```

This API aims to fetch that entity's data from my view point, could we return fields like schema, meta, etc?

tzssangglass: > > thinking Emma then let's use the old way in 2.x? We could use the more RESTful API in breaking changes IMO, what do you think? @moonming @liuxiran @ShiningRush
> 
> Metadata and schema are two different things, and should not be mixed in the same API

you are right, I was confusing metadate and schema.

membphis: https://github.com/apache/apisix/blob/master/t/admin/plugins.t#L128
https://github.com/apache/apisix/blob/master/t/admin/plugins.t#L149

the APISIX API has done in this way for .

`/apisix/admin/plugins/basic-auth`
`/apisix/admin/plugins/basic-auth?schema_type=consumer`

and `/apisix/admin/plugins/{plugin-name}` is an old API for months. If we want to change it, it will not be backward compatible.

that is the original PR link: https://github.com/apache/apisix/pull/2509

membphis: > thinking Emma then let's use the old way in 2.x? We could use the more RESTful API in breaking changes IMO, what do you think? @moonming @liuxiran @ShiningRush

If we want to do in this way, need to discuss in mail-list at first:

image

membphis:
I think we have discussed this at mail-list: https://lists.apache.org/thread.html/r33b8d142ba57cfda467b4912ec94d5ae9c77b1eec4982f39f2e6bf0e%40%3Cdev.apisix.apache.org%3E

@tzssangglass welcome to join in and discuss it.

tzssangglass:
> I think we have discussed this at mail-list: https://lists.apache.org/thread.html/r33b8d142ba57cfda467b4912ec94d5ae9c77b1eec4982f39f2e6bf0e%40%3Cdev.apisix.apache.org%3E
got it. have joined and discussed it. 王三 is me. I will revise this PR based on the results of the discussion soon.

t/admin/plugins.t
```diff
@@ -166,9 +166,27 @@ plugins:
 
 
 === TEST 8: get all the attributes of all plugins
```
membphis, line 168: add a test to confirm the plugin of `auth` type

tzssangglass: add

t/admin/plugins.t
```diff
@@ -162,3 +162,65 @@ plugins:
...
+        }
+    }
+--- response_body eval
+qr/\{"name":"example-plugin","priority":0,"schema":\{"properties":\{"i":\{"minimum":0,"type":"number"\},"ip":\{"type":"string"\},"port":\{"type":"integer"\},"s":\{"type":"string"\},"t":\{"minItems":1,"type":"array"\}\},"required":\["i"\],"type":"object"\},"type":"other","version":0.1\}/
+--- no_error_log
+[error]
+
+
+
+=== TEST 9: confirm the plugin of auth type
```
membphis, line 196: more test cases: `priority`, `schema`, `consumer_schema`

doc/zh-cn/admin-api.md
```diff
@@ -754,3 +754,15 @@ Content-Type: text/plain
 ```
 
 [Back to TOC](#目录)
+
+*地址*：/apisix/admin/plugins/?all=true
+
+*说明*: 所有插件的所有属性，每个插件包括 `name`, `priority`, `priority`, `schema`, `consumer_schema` and `version`。
```
membphis, line 760: duplicated `priority`

doc/admin-api.md
```diff
@@ -744,3 +744,15 @@ Content-Type: text/plain
 ```
 
 [Back to TOC](#Table-of-Contents)
+
+*API*：/apisix/admin/plugins/?all=true
+
+*Description*: all the attributes of all plugins, each plug-in includes `name`, `priority`, `priority`, `schema`, `consumer_schema` and `version`.
```
membphis, line 750: duplicated `priority`

doc/zh-cn/admin-api.md
tzssangglass: update

t/admin/plugins.t
tzssangglass: add a test case for `consumer_schema`.
update the title of test case no.8 in `t/admin/plugins.t`, the `name`, `priority`, `schema`,`type` and `version` of plugin has covered by test case no.8.

tzssangglass: or as you say more test cases means?

membphis (approved):
LGTM

doc/admin-api.md
```diff
@@ -744,3 +744,15 @@ Content-Type: text/plain
 ```
 
 [Back to TOC](#Table-of-Contents)
+
+*API*：/apisix/admin/plugins/?all=true
+
+*Description*: all the attributes of all plugins, each plug-in includes `name`, `priority`, `type`, `schema`, `consumer_schema` and `version`.
```
spacewander, line 750: `plug-in` should be plugin, since `plugins` is used in the same sentence.

apisix/admin/plugins.lua
```diff
@@ -37,7 +37,42 @@ function _M.stream_check_schema(plugins_conf, schema_type)
 end
 
 
+local function get_plugins_all_attributes()
+    local plugins = core.config.local_conf().plugins
+    local all_attributes = core.table.new(#plugins, 0)
+    for _, name in ipairs(plugins) do
+        local plugin_name = "apisix.plugins." .. name
+        local ok, plugin = pcall(require, plugin_name)
```
spacewander, line 45: There is not need to reload all the plugins. You can directly get them like:
https://github.com/apache/apisix/blob/0802c95faf4c5615c3b9fbe56ef913be7af08287/apisix/control/v1.lua#L37-L45

spacewander (changes requested):

apisix/admin/plugins.lua
spacewander: This method is not provided when you wrote this PR. But I think now we can make it simpler.

tzssangglass: Is this pr still necessary?  I see https://github.com/apache/apisix/pull/3092 already contains the functions provided by this pr.

spacewander: It is different thing. #3092 is in control API.

tzssangglass: update

apisix/admin/plugins.lua
```diff
@@ -23,6 +23,7 @@ local pcall     = pcall
 local table_sort = table.sort
 local table_insert = table.insert
 local get_uri_args = ngx.req.get_uri_args
+local v1 = require("apisix.control.v1")
```
spacewander, line 26: It would be better to use `plugin.get_all` directly, instead of let admin API depends on control API. And the fields returned by this one is different from the control API one?

spacewander (changes requested):

apisix/admin/plugins.lua
tzssangglass: yes, different, at first, this api was planned to return only the schema of the plugins, not included as `route`, `service`

spacewander (approved):

doc/admin-api.md
```diff
@@ -710,3 +710,15 @@ Content-Type: text/plain
 ```
 
 [Back to TOC](#Table-of-Contents)
+
+*API*：/apisix/admin/plugins/?all=true
```
juzhiyuan, line 714: valid to use this format?

cc @membphis @spacewander @gxthrj

tzssangglass: image
I saw that after this format was proposed on the mail-list: https://lists.apache.org/thread.html/r33b8d142ba57cfda467b4912ec94d5ae9c77b1eec4982f39f2e6bf0e%40%3Cdev.apisix.apache.org%3E, no one ended up having a problem with it.

juzhiyuan: image

But there has a `/` in the API

tzssangglass: `/apisix/admin/plugins/?all=true` and `/apisix/admin/plugins?all=true`  both can be accessed.

tzssangglass: should I update this md?

spacewander: Would be better to use `/apisix/admin/plugins?all=true`

tzssangglass: can reopen this PR?

tzssangglass: resolve by #3153

Issue #2870, opened 2020-11-27
bug: update plugin config for route does not take effect


nic-chen:
### Issue description

Update the plugin config for route, and then access to the route, the plugin config is not updated.

### Environment

* apisix version (cmd: `apisix version`):  master source
* OS: macOS

### Minimal test code / Steps to reproduce the issue
1.  create route
```sh
$ curl http://127.0.0.1:9180/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "uri": "/opentracing",
    "plugins": {
        "http-logger": {
            "uri": "http://172.16.238.20:1982/hello",
            "batch_max_size": 1,
            "max_retry_count": 1,
            "retry_delay": 2,
            "buffer_duration": 2,
            "inactive_timeout": 2,
            "name": "http logger",
            "timeout": 3,
            "concat_method": "json"
        }
    },
    "upstream": {
        "type": "roundrobin",
        "nodes": [{
            "host": "172.16.238.20",
            "port": 1981,
            "weight": 1
        }]
    }
}'
```

2. access the route
```sh
$ curl http://127.0.0.1:9080/opentracing
```

3. update the route
```sh
$ curl http://127.0.0.1:9180/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "uri": "/opentracing",
    "plugins": {
        "http-logger": {
            "uri": "http://172.16.238.20:8888/hello",
            "batch_max_size": 1,
            "max_retry_count": 1,
            "retry_delay": 2,
            "buffer_duration": 2,
            "inactive_timeout": 2,
            "name": "http logger",
            "timeout": 3,
            "concat_method": "json"
        }
    },
    "upstream": {
        "type": "roundrobin",
        "nodes": [{
            "host": "172.16.238.20",
            "port": 1980,
            "weight": 1
        }]
    }
}'
```

4. access the route again
```sh
$ curl http://127.0.0.1:9080/opentracing
```

### What's the actual result? (including assertion message & call stack if applicable)

upstream has changed to `172.16.238.20:1980`

but log is still pushed to `http://172.16.238.20:1982/hello`

### What's the expected result?

log should push to `http://172.16.238.20:8888/hello`

tokers:
The old log_buffer object was still in use after the route was updated.

tokers:
Assigned to me.

PR #2901, opened 2020-11-30
fix: fixed the non effective config update in http-logger

tokers:
### Pre-submission checklist:

Close https://github.com/apache/apisix/issues/2870

### DIFF
--- a/apisix/plugins/http-logger.lua
+++ b/apisix/plugins/http-logger.lua
@@ -180,7 +180,7 @@ local function remove_stale_objects(premature)
 
     for key, batch in ipairs(buffers) do
         if #batch.entry_buffer.entries == 0 and #batch.batch_to_process == 0 then
-            core.log.debug("removing batch processor stale object, route id:", tostring(key))
+            core.log.warn("removing batch processor stale object, route id:", tostring(key))
             buffers[key] = nil
         end
     end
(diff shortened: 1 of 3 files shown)


t/plugin/http-logger.t
```diff
@@ -637,3 +637,149 @@ GET /t
 done
 --- no_error_log
 [error]
+
+
```
Firstsawyou, line 641: Need to add a blank line.

apisix/plugins/http-logger.lua
```diff
@@ -169,6 +174,23 @@ local function gen_log_format(metadata)
...
 
+-- remove stale objects from the memory after timer expires
+local function remove_stale_objects(premature)
+    if premature then
+        return
+    end
+
+    for key, batch in ipairs(buffers) do
+        if #batch.entry_buffer.entries == 0 and #batch.batch_to_process == 0 then
+            core.log.debug("removing batch processor stale object, route id:", tostring(key))
```
spacewander, line 185: Better to use a high level log so that we can notice user. What about using `warn`?

t/plugin/http-logger.t
tokers: Fixed.

apisix/plugins/http-logger.lua
tokers: OK, this logic was copied from other x-logger.lua, i'd like to fix all of them in another PR.

tokers: @spacewander Fixed it in http-logger firstly.

spacewander (approved):

membphis (approved):

membphis:
@tokers you should fork APISIX repository and checkout new branch for developing, this is the right way.

image

Issue #2880, opened 2020-11-27
feat: Make HTTP headers set by openid-connect plugin more configurable.


jenskeiner:
## Background
The `openid-connect` plugin, when configured accordingly, may obtain an _ID token_, a _user info_ JSON object, and an _access token_, respectively, on behalf of the requestor by automatically redirecting to the ID provider. These tokens are then stored in a session cookie.

Additionally, the plugin sets the HTTP header `X-ID-Token` to the value of the ID token, the header `X-Access-Token` to the value of the access token, and the header `X-Userinfo` to the user info data, for downstream handlers.

Above code path requires that the respective tokens are already stored in a session (cookie). There is an alternative code path where an incoming request already contains an access (bearer) token in the `Authorization` header. In this case, the token is validated and the plugin then sets the header `X-Userinfo` to either the decoded token value (when validating directly against a public key) or the results of the introspection endpoint of the ID provider.

Importantly, in this code path, I believe it is possible to have different information being put into the header, depending on how the validation is performed. This does not necessarily break since access token and introspection result may contain overlapping and entirely identical information, but an access token is not guaranteed to be identical to the introspection result, so breaks could occur where they differ.

## Issues with the current solution

1. As noted above, in one code path, the `X-Userinfo` header contains the user info data, and in another one, the same header contains the opaque access token. This looks inconsistent and I would expect downstream handlers to break if they depend on this being a legitimate introspection endpoint response data and the access tokens happen to use an incompatible  format.

2. The code path where the tokens are already stored in a session cookie sets three additional headers (with entirely redundant information from the session cookie) and may increase the request size substantially. This may be inefficient where downstream handlers actually don't need all three tokens. This can also lead to breaks where request header size is constrained.

3. Downstream handlers, e.g. the `authz-keycloak` plugin or actual backend services typically expect an access token to be sent in the `Authorization` header like `Authorization: bearer <token>`, but not the non-standard `X-Access-Token`. As a result, they may fail to extract the token from the incoming request, causing breaks.

## Suggested changes

1. Make setting the HTTP headers configurable, so that one can choose which tokens (if available) should actually be passed to downstream handlers in the respective headers.

2. Add a configuration option that allows the access token to be set in the `Authorization` header instead of `X-Access-Token`.

3. When the incoming request already contains a bearer token (in the `Authorization` header), additionally set the `X-Access-Token` header to the same value if that's the configured format. Otherwise, don't add the header.

[...]

PR #2903, opened 2020-11-30
feat: Make headers to add to request in openid-connect plugin configurable.

jenskeiner:
See #2880. The current implementation may lead to very large headers, thus bloating the request size to downstream substantially. This can be an issue by itself but is also not desired if the different tokens (access, ID, user info) largely contain the same information. Also, instead of using the `X-Access-Token` header it may be more appropriate to use the `Authorization` header for the access token, so that downstream plugins or services can extract it correctly; see e.g. `authz-keycloak` plugin. Finally, the plugin was so far only expecting an incoming access token in the `Authorization` header, but not the `X-Access-Token` header, which seemed somewhat inconsistent.

I've added options that can be configured for the `openid-connect` plugin that allow to control which headers get added and which header specifically shall be used for the access token. The defaults reproduce the current behavior. The plugin will also look in the `X-Access-Token` header, in addition to the `Authorization` header, for an access token in an incoming request.

### Pre-submission checklist:

Haven't added any new cases for now. The headers set by the plugin were not tested, as far as I can tell. It may be a good idea to think about some new tests around here. That would likely require a bit of legwork to set up a dummy OIDC Identity Provider with all needed endpoints to go through all possible code paths though.
I understand that the plugin documentation should be updated, but I want to get some feedback on the code changes first. Happy to adjust the documentation if the code changes get signed off.


apisix/plugins/openid-connect.lua
```diff
@@ -82,63 +87,131 @@ function _M.check_schema(conf)
...
+end
+
+
 local function introspect(ctx, conf)
-    if has_bearer_access_token(ctx) or conf.bearer_only then
+    -- Extract token, maybe. Ignore errors.
+    local has_token, token, _ = check_bearer_access_token(ctx)
+
+    -- Check if token was extracted or if we always require a token in the request.
+    if has_token(ctx) or conf.bearer_only then
```
spacewander, line 169: Why you call the boolean value `has_token`?

apisix/plugins/openid-connect.lua
```diff
@@ -152,26 +225,28 @@ function _M.access(plugin_conf, ctx)
             core.log.error("failed to introspect in openidc: ", err)
             return response
         end
-        if response then
```
spacewander, line 155: Why remove this part?

apisix/plugins/openid-connect.lua
```diff
@@ -82,63 +87,131 @@ function _M.check_schema(conf)
...
             else
+                -- Token is valid and res contains the response from the introspection endpoint.
+
+                if conf.set_userinfo_token_header then
+                    -- Set X-Userinfo header to introspection endpoint response.
+                    add_user_header(res)
+                end
+
+                -- Add configured access token header, maybe.
+                add_access_token_header(token, conf)
```
spacewander, line 196: `add_access_token_header(ctx, conf, token)`
Argument mismatch.

apisix/plugins/openid-connect.lua
```diff
@@ -152,26 +225,28 @@ function _M.access(plugin_conf, ctx)
...
+            if conf.set_userinfo_token_header and response.user then
                 add_user_header(response.user)
             end
-            if response.access_token then
-                ngx.req.set_header("X-Access-Token", response.access_token)
-            end
-            if response.id_token then
+
+            -- Add configured access token header, maybe.
+            add_access_token_header(ctx, conf, response.access_token)
```
spacewander, line 246: Should check if `response.access_token` is nil

apisix/plugins/openid-connect.lua
```diff
@@ -14,6 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+local string = require("string")
```
spacewander, line 17: `local string = string`
is enough.

apisix/plugins/openid-connect.lua
```diff
@@ -40,7 +41,11 @@ local schema = {
         logout_path = {type = "string"}, -- default is /logout
         redirect_uri = {type = "string"}, -- default is ngx.var.request_uri
         public_key = {type = "string"},
-        token_signing_alg_values_expected = {type = "string"}
+        token_signing_alg_values_expected = {type = "string"},
+        set_access_token_header = {type = "boolean"}, -- default is true
```
spacewander, line 45: Let's use the new way to set the default value:
https://github.com/apache/apisix/blob/9dfe697b644691509b55cd7ec58188b33dd3e792/apisix/plugins/cors.lua#L37

spacewander (changes requested):

spacewander:
Please write new tests to cover your change, and make sure they are passed.

apisix/plugins/openid-connect.lua
jenskeiner: It's a bug, have pushed a fix.

apisix/plugins/openid-connect.lua
jenskeiner: ok

apisix/plugins/openid-connect.lua
jenskeiner: Sure, hadn't seen it before.

jenskeiner: Sure, hadn't seen it before. Will change.

apisix/plugins/openid-connect.lua
jenskeiner: Correct, good spot. Will fix.

apisix/plugins/openid-connect.lua
jenskeiner: The reason is that adding the user info header is now handled in the `introspect` method. The header will still be added, if so configured, and introspection is done via the actual endpoint as opposed to just checking against the public key. Let me know if you see any problem with this approach.

apisix/plugins/openid-connect.lua
jenskeiner: Agreed. Will change.

apisix/plugins/openid-connect.lua
jenskeiner: I've changed the scheme for the options that I've added. But have left the existing options untouched to not make this PR any larger.

jenskeiner:
> Please write new tests to cover your change, and make sure they are passed.

I'm on it. I had a look at the existing structure and I'm relatively confident I can add more test cases to verify the setting of the respective headers.

However, one thing is still unclear to me: There are cases already where token introspection is performed against a configured identity provider endpoint. The cases were added by @sshniro here: https://github.com/apache/apisix/commit/14f979edf84c5a76a96902cc2ab8e5697a905fe9

Can anyone point me to the place where this introspection endpoint is set up for the tests? I can't find the corresponding code inside the project.

I may need to add more dummy endpoints to verify the authorization flow in case a bearer token is not present in the incoming request and I want to make that consistent with the existing dummy introspection endpoint.

apisix/plugins/openid-connect.lua
```diff
@@ -93,59 +118,124 @@ function _M.check_schema(conf)
...
         local res, err
 
         if conf.public_key then
+            -- Validate token against public key.
             res, err = openidc.bearer_jwt_verify(conf)
             if res then
+                -- Token is valid.
+
+                -- Add configured access token header, maybe.
+                add_access_token_header(ctx, conf, token)
```
spacewander, line 203: Token is invalid if has_token is false but conf.bearer_only is true

apisix/plugins/openid-connect.lua
```diff
@@ -93,59 +118,124 @@ function _M.check_schema(conf)
...
             else
+                -- Token is valid and res contains the response from the introspection endpoint.
+
+                if conf.set_userinfo_token_header then
+                    -- Set X-Userinfo header to introspection endpoint response.
+                    add_user_header(ctx, res)
+                end
+
+                -- Add configured access token header, maybe.
+                add_access_token_header(ctx, conf, token)
```
spacewander, line 220: Ditto

spacewander (changes requested):

sshniro:
Hi @jenskeiner for your question on the introspection endpoint test case, I have used the Keycloak to test this scenario.

The code link to setup Keycloak:
https://github.com/apache/apisix/blob/f84a44e409826cfbeebf89ec7a8ebe47dcf9e387/.travis/linux_openresty_runner.sh#L27
Test cases written for the introspection with Keyclaok:
https://github.com/apache/apisix/blob/f84a44e409826cfbeebf89ec7a8ebe47dcf9e387/t/plugin/openid-connect.t#L424

@membphis the docker image is using my personal docker repo, if needed I can migrate it to an official APISIX docker repository.

membphis:
@membphis the docker image is using my personal docker repo, if needed I can migrate it to an official APISIX docker repository.

seems good for me.

what do you think? @moonming

moonming:
> @membphis the docker image is using my personal docker repo, if needed I can migrate it to an official APISIX docker repository.
> 
> seems good for me.
> 
> what do you think? @moonming

we can add `keycloak-apisix` docker to CI, and I am not sure I am not sure if it is appropriate to migrate it to an official APISIX docker repository, most of APISIX users don't need keycloak.

t/plugin/openid-connect.t
```diff
@@ -112,7 +112,7 @@ done
                         },
                         "upstream": {
                             "nodes": {
-                                "127.0.0.1:1980": 1
+                                "127.0.0.1:8888": 1
```
membphis, line 115: `8888`, Is it a valid "openid" port?

jenskeiner: Hey, I'm trying to add tests to check if the plugin sets the headers correctly as configured. I don't have a local development setup yet, so have to run the checks on this PR on GitHub to see if it's working.

I switched the upstream port to `8888` since that's where a `mendhak/http-https-echo` container is running. This one returns in the response body the full content of the incoming request as a JSON string. So I think I can check if the plugin has set the headers correctly by checking the response body. If the tests still work with the new port, I will try to add a case to do some actual verification of the headers.

spacewander: @jenskeiner 
You can setup the test environment via https://github.com/apache/apisix/blob/master/.travis/linux_openresty_runner.sh

jenskeiner: Yes, thanks for the link. It's more that I have to set up a fresh Linux VM or similar since I want that separate from the rest.

spacewander: And you can write your echo-back here: https://github.com/apache/apisix/blob/master/t/lib/server.lua

jenskeiner: Understood, seems to be the better option. Will adjust the code.

t/plugin/openid-connect.t
```diff
@@ -626,6 +628,86 @@ Authorization: Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJkYXRhMSI6IkRhdGEgM
 
 
 
+=== TEST 9z: update plugin public key, so tokens can be validated locally
```
tokers, line 631: Why using `9z`?

jenskeiner: I'm in the middle of the process of adding cases, so will fix the numbering when done. No point doing it now as I don't know how many cases I will add.

Also, why number the cases in the first place? If you insert just one somewhere you have to adjust the numbers of all cases below.

jenskeiner: Also, if anyone could help me setting up a proper dev environment so I don't have to go through the entire suite of GitHub checks all the time just to run the tests on the oidc module, I would appreciate that very much.

I tried setting up an Ubuntu 18.04 VM according to Travis configuration, but it's frankly a nightmare to get everything up and running because most scripts need tiny little adjustments to actually work as expected in that environment. In the end, I seemingly managed to get all containers and everything up and running, but the tests still wouldn't run because the APISIX process was still throwing up a lot in the error log. Not a single test passed as a result.

I need to find a better development workflow or I may have to abandon this whole ticket. At this stage, I'm wasting too much time on stuff not really related to the actual dev work. Can't afford that any longer.

Maybe it might be a good idea to set up a Vagrant box or similar so new developers don't have to go through all this trial and error.

Any hints appreciated.

spacewander: It is amazing that you finally get all the test passed! Thanks a lot!

spacewander: @jenskeiner 
We can fulfill the tests for you later. Before we can merge, the only thing you need to do is to update the English doc: https://github.com/apache/apisix/blob/master/doc/plugins/openid-connect.md#attributes

jenskeiner: Hang on, I disabled most of the tests, so currently just running the OIDC ones. Need to revert that before we proceed.

Also, was thinking how to test the path where the initial request doesn't have a token and is redirected to the identity provider. When the user has authenticated, the identity provider will call the redirect URL with a code so that APISIX (the relying party) can obtain access token, ID token, and userinfo from the provider's endpoints.

We need to test this because the plugin wil set headers to the upstream service accordingly in this case.

Probably best to mock up the endpoints instead of interacting with the Keycloak container. I'll see if I can get this done as well.

In any case, I will post here when I believe the PR is actually ready for final review round.

jenskeiner:
@spacewander @membphis @moonming Sorry for the noise, but took a while to set up test cases that go through the full OIDC Relying Party flow which can be used when the incoming request doesn't contain a bearer token yet.

Please give this another good review and let me know any issues. Thanks.

I know, I still need to update the documentation, but will wait on feedback regarding code changes first.

jenskeiner:
The code changes in this PR are basically just to solve some issues I was seeing when combining this plugin with the `auth-keycloak` one. I wanted to make sure that the latter can extract the access token from the request, even if it is contained in a session cookie or potentially some other server-side storage as supported by the `openidc` module. The changes make sure that the token can always be added in a header where `authz-keycloak` expects it. I also wanted to reduce some noise from the ID token and the user info that were getting added in separate headers because you probably don't need those always and they make the HTTP header even larger.

Generally, I think the interplay between this plugin and `auth-keycloak` should be re-considered. Messing with the request headers could be avoided completely if `auth-keycloak` could leverage the `openidc` module to extract the access token from the request (from a header or cookie or...) itself using a method that is currently internal to that module. Then, it wouldn't need to rely on the token being stored in a header and code to extract the token could be removed in favor of using the `openidc` module's code for that.

I'm just wondering what anyone's thoughts on this are.

The current solution with the headers works for me and I'll look at `authz-keycloak` next to start the work on lazy path loading from Keycloak. But longer term, would be better to clean up the code some more and not rely on the `openid-connect` plugin to fix up the headers for the `auth-keycloak` plugin. Thoughts and feedback welcome.

@sshniro Maybe you have thoughts on this as well...

apisix/plugins/openid-connect.lua
```diff
@@ -95,56 +120,107 @@ function _M.check_schema(conf)
...
     return false
 end
 
 
+
 local function introspect(ctx, conf)
-    if has_bearer_access_token(ctx) or conf.bearer_only then
-        local res, err
+    -- Extract token, maybe.
+    local has_token, token, _ = get_bearer_access_token(ctx)
```
spacewander, line 156: Better to check the err: `_`.

spacewander (changes requested):

t/plugin/openid-connect.t
```diff
@@ -263,20 +1053,105 @@ passed
 
 
 
-=== TEST 7: access
---- timeout: 10s
+
+=== TEST 14: Update route URI to '/uri' where upstream endpoint returns request headers in response body.
```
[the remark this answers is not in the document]
jenskeiner: Fixed. Thanks.

apisix/plugins/openid-connect.lua
jenskeiner: Done. Added a test case for a malformed Authorization header as well.

apisix/plugins/openid-connect.lua
```diff
@@ -95,56 +120,107 @@ function _M.check_schema(conf)
...
+        -- endpoint contains the userinfo.
+        return res, err, token, res
+    end
 end
 
 
-local function add_user_header(user)
-    local userinfo = core.json.encode(user)
-    ngx.req.set_header("X-Userinfo", ngx_encode_base64(userinfo))
+local function add_access_token_header(ctx, conf, token)
```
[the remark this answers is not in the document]
jenskeiner: @spacewander Can you specify what you suggest should be checked? The `add_access_oken_header` method is only called when an access token is actually available in the request.

sshniro (approved):

apisix/plugins/openid-connect.lua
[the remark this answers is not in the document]
spacewander: @jenskeiner 
There is two checks outside this function. I think we can move the check inside so that we don't need to repeat it.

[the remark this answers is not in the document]
jenskeiner: Understood and done.

spacewander (approved):

spacewander:
Fantasy! Before we can merge it, @jenskeiner please write some English docs.

jenskeiner:
> Fantasy! Before we can merge it, @jenskeiner please write some English docs.

Yep, will do, may take until next Monday though. In the meantime, let me know any worries.

jenskeiner:
@spacewander Updated the documentation now as well. Can you give it another good review to ensure it's consistent and clear enough?

spacewander:
@jenskeiner 
Merged. Thanks!

PR #2905, opened 2020-11-30
fix: Basic auth plugin should run in rewrite phase.

jenskeiner:
The `basic-auth` plugin, as an authentication plugin, should run in the rewrite phase, not the access phase.

### Pre-submission checklist:


jenskeiner:
As per comment from @spacewander, added a check for the consumer handling; see https://github.com/apache/apisix/pull/2859#discussion_r531816904.

apisix/init.lua
```diff
@@ -513,6 +513,8 @@ function _M.http_access_phase()
                 api_ctx.consumer,
                 api_ctx
             )
+            core.log.debug("find consumer ", api_ctx.consumer.username,
```
spacewander, line 516: IMHO, we can use `info` level error log, therefore we can check it without `log_level('debug')`. `debug` is too low to check it in the production if needed.

t/plugin/basic-auth.t
```diff
@@ -189,6 +190,7 @@ Authorization: Basic Zm9vOmZvbwo=
 
 
 === TEST 8: verify
+--- yaml_config eval: $::yaml_config
```
spacewander, line 193: We don't need this line?

apisix/init.lua
jenskeiner: Sure, just thought we don't want to clutter the log too much in a prod setting where you might be running with `info` level.

t/plugin/basic-auth.t
jenskeiner: Yep, will remove it. Was left over from my first attempt at changing the log level.

spacewander (approved):

moonming (approved):

membphis:
@jenskeiner many thx, merged right now

PR #2910, opened 2020-12-01
fix: authz-keycloak plugin should run in access phase and should not be designated as of type auth anymore.

jenskeiner:
The `authz-keycloak` plugin is concerned with authorization against Keycloak, not authentication. As such, it should not be designated as of type `auth`. Also, I suggest to run it in the access phase which seems more appropriate.

### Pre-submission checklist:


spacewander (approved):

membphis (approved):

membphis:
@jenskeiner many thx

PR #2911, opened 2020-12-01
feat: validate the operator in vars

### LINKED ISSUE #2878: request help: should strengthen the verification of operators
### Issue description
create with vars which operator was wrong, but APISIX still returns 200 and insert to etcd.

### Environment

* apisix version (cmd: `apisix version`): master
* OS:

```
{
			caseDesc: "add route with invalid vars (error operators)",
			Object:   MangerApiExpect(t),
			Method:   http.MethodPut,
			Path:     "/apisix/admin/routes/r1",
			Body: `{
					"uri": "/hello",
					"vars": [
						["arg_name","=","aaa"]
					],
					"upstream": {
						"type": "roundrobin",
						"nodes": [{
							"host": "172.16.238.20",
							"port": 1980,
							"weight": 1
						}]
					}
				}`,
			Headers:      map[string]string{"Authorization": token},
			ExpectStatus: http.StatusNotFound,
		},
```
return 200, and insert the route to etcd, but it is useless.

return 400

spacewander:
Fix #2878

### Pre-submission checklist:

### DIFF
--- a/apisix/http/route.lua
+++ b/apisix/http/route.lua
@@ -1,0 +1,42 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local expr = require("resty.expr.v1")
+local plugin_checker = require("apisix.plugin").plugin_checker
+
+
+local _M = {}
+
+
+-- additional check for synced route configuration, run after schema check
+function _M.check_route(route)
+    local ok, err = plugin_checker(route)
+    if not ok then
+        return nil, err
+    end
+
+    if route.vars then
+        ok, err = expr.new(route.vars)
+        if not ok then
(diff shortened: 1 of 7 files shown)


membphis (approved):

apisix/http/route.lua
```diff
@@ -0,0 +1,42 @@
...
+-- limitations under the License.
+--
+local expr = require("resty.expr.v1")
+local plugin_checker = require("apisix.plugin").plugin_checker
+
+
+local _M = {}
+
+
+-- additional check for synced route configuration, run after schema check
```
Firstsawyou, line 24: If the title is like this：`additional "vars" rule check for synced route configuration, run after schema check`,  I think it would be better, right?

apisix/http/route.lua
```diff
@@ -0,0 +1,42 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local expr = require("resty.expr.v1")
+local plugin_checker = require("apisix.plugin").plugin_checker
+
+
+local _M = {}
+
+
```
Firstsawyou, line 23: If there is only one blank line above and below `local _M = {}`, I think it would be better.

spacewander: @Firstsawyou 
I think there is not rule about the number of blank lines around `_M`.
For example, `http/router/radixtree_host_uri.lua` and `stream/plugins/mqtt-proxy.lua` and more other files have two blank lines above.

Actually if APISIX wants to follow the style of OpenResty's standard library, two blank lines around `_M` is required. You can take a look at `lua-resty-core`.

apisix/http/route.lua
spacewander: @Firstsawyou 
There are other checks here (like plugin checker), so I don't think adding `"vars"` is a good idea.

apisix/http/route.lua
Firstsawyou: Ok, I get it.

apisix/http/route.lua
Firstsawyou: ok.

tokers (approved):

PR #2913, opened 2020-12-01
fix: Use correct name when accessing Authorization header in auth-keycloak plugin.

jenskeiner:
The `auth-keycloak` plugin expects to find a bearer token in the standard `Authorization` header. However, it tries to obtain a token from a header called `authorization`, i.e. not properly capitalized. This PR fixes the capitalization so the correct header is used; cf. https://en.wikipedia.org/wiki/List_of_HTTP_header_fields.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/authz-keycloak.lua
+++ b/apisix/plugins/authz-keycloak.lua
@@ -136,7 +136,7 @@ end
 
 
 local function fetch_jwt_token(ctx)
-    local token = core.request.header(ctx, "authorization")
+    local token = core.request.header(ctx, "Authorization")
     if not token then
         return nil, "authorization header not available"
     end


jenskeiner:
To be honest, I currently fail to see why unit tests in the non-fixed version didn't blow up before. Need to investigate that.

But I was able to confirm that the header was not found when the incoming request used `authorization` instead of `Authorization`.

membphis (approved):

moonming:
Is http header case sensitive? @membphis

membphis:
> Is http header case sensitive? @membphis

Case insensitive, confirm.  `Authorization` is more readable.

moonming (approved):

membphis:
@jenskeiner many thx, merged

PR #2917, opened 2020-12-02
fix: refresh cached header when we set request header

spacewander:
### Pre-submission checklist:

### DIFF
--- a/apisix/core/request.lua
+++ b/apisix/core/request.lua
@@ -68,12 +68,25 @@ function _M.header(ctx, name)
 end
 
 
-function _M.set_header(header_name, header_value)
+function _M.set_header(ctx, header_name, header_value)
+    if type(ctx) == "string" then
+        -- It would be simpler to keep compatibility if we put 'ctx'
+        -- after 'header_value', but the style is too ugly!
+        header_value = header_name
+        header_name = ctx
+        ctx = nil
+    end
+
     local err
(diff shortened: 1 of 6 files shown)


apisix/core/request.lua
```diff
@@ -68,12 +68,25 @@ function _M.header(ctx, name)
 end
 
 
-function _M.set_header(header_name, header_value)
+function _M.set_header(ctx, header_name, header_value)
+    if type(ctx) == "string" then
+        -- It would be simpler to keep compatibility if we put 'ctx'
+        -- after 'header_value', but the style is too ugly!
```
membphis, line 74: how about print a warning message to log?

jenskeiner:
@spacewander I reckon this would obsolete a similar change I made here: bfa4aa7297c14c181a7403daea4afdce2684040f Correct?

I had noticed that setting a request header in `openid-connect` plugin would not be visible to the downstream `auth-keycloak` plugin since the headers there are retrieved from a cache which hadn't been updated.

Better to fix in the core method anyway, so +1 for this change.

spacewander:
@jenskeiner 
Yes, you are right!

apisix/core/request.lua
spacewander: @membphis 
Added.

moonming (approved):

membphis (approved):

PR #2926, opened 2020-12-02
feat: server info

### LINKED ISSUE #2821: Implement status API
We have already some disscussion about this topic: https://lists.apache.org/thread.html/re3097f7e88d636e8137231c36462c32773b2b4817d91f7997e91358f%40%3Cdev.apisix.apache.org%3E.

This thread is used to track the development progress about APISIX part.

The related issues:

* https://github.com/apache/apisix-dashboard/issues/849
* https://github.com/apache/apisix-dashboard/issues/850

tokers:

Close #2821.

### Pre-submission checklist:


doc/plugins/server-info.md
```diff
@@ -0,0 +1,115 @@
...
+```yaml
+plugin_attr:
+  server-info:
+    report_interval: 10
+```
+
+## Test Plugin
+
+```bash
+curl http://127.0.0.1:9080/apisix/admin/server_info -s | jq
```
membphis, line 81: the URI should be `/apisix/server_info`

doc/zh-cn/plugins/server-info.md
```diff
@@ -0,0 +1,116 @@
...
+plugin_attr:
+  server-info:
+    report_interval: 10
+```
+
+
+## 测试插件
+
+```bash
+curl http://127.0.0.1:9080/apisix/admin/server_info -s | jq
```
membphis, line 81: ditto

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,274 @@
...
+    additionalProperties = false,
+}
+local attr_schema = {
+    type = "object",
+    properties = {
+        report_interval = {
+            type = "integer",
+            description = "server info reporting interval (unit: second)",
+            default = 60,
+            minimum = 5,
```
membphis, line 39: need to specify a maximum value

doc/plugins/server-info.md
tokers: Fixed.

doc/zh-cn/plugins/server-info.md
tokers: @membphis FIxed.

apisix/plugins/server-info.lua
tokers: Fixed.

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,274 @@
...
+    server_info.last_report_time = ngx_time()
+
+    local data, err = core.json.encode(server_info)
+    if not data then
+        core.log.error("failed to encode server_info: ", err)
+        return
+    end
+
+    local key = "/data_plane/server_info/" .. server_info.id
+    local ok, err = core.etcd.set(key, data, 180)
```
spacewander, line 161: I think it would be better to let the CP clean up the data. For example, if the DP is just shutdowned for maintain, it would be better to keep the info data in etcd.

BTW, `180` is rather short compared with the 60 second report internal. And the number is undocumented. People will shout at us when they find their data is disappeared.

spacewander (changes requested):

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,274 @@
...
+    end
+end
+
+
+function _M.api()
+    return {
+        {
+            methods = {"GET"},
+            uri = "/apisix/server_info",
+            handler = get_server_info,
```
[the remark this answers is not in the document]
tokers: That is to say we haven't a way to expose these data in data plane, at least for now.

apisix/plugins/server-info.lua
tokers: > I think it would be better to let the CP clean up the data. For example, if the DP is just shutdowned for maintain, it would be better to keep the info data in etcd.

That can be solved by a long TTL.

> BTW, 180 is rather short compared with the 60 second report internal. And the number is undocumented. People will shout at us when they find their data is disappeared.

OK, we can add another item like `report_ttl` in the plugin_attr.

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,274 @@
...
+        return
+    end
+
+    local report_interval = attr.report_interval
+
+    -- we don't use core.timer and the apisix.timers here for:
+    -- 1. core.timer is not cancalable, timers will be leaked if plugin
+    -- reloading happens.
+    -- 2. the background timer in apisix.timers fires per 500 milliseconds, if
+    -- report_interval is not multiple of 0.5, the real report interval will be
```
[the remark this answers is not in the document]
tokers: I forgot to remove the second comment.

But the backend timer is not suitable, we use backend timer without caring the fire period, if we use it here, we have to do some calculations according to the 500 milliseconds, which must be updated once the backend timer fire period changed.

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,275 @@
...
+    additionalProperties = false,
+}
+local attr_schema = {
+    type = "object",
+    properties = {
+        report_interval = {
+            type = "integer",
+            description = "server info reporting interval (unit: second)",
+            default = 60,
+            minimum = 60,
```
starsz, line 39: Hi, I think we can turn off the report by setting the report_interval zero.

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
spacewander: Yes, except we can expose it via etcd.

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
spacewander: > we have to do some calculations according to the 500 milliseconds, which must be updated once the backend timer fire period changed

We can expose the `0.5` as a constant, and used in this plugin. It would be better than inventing a new cycle.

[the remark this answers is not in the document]
tokers: Well, if we do so, the interval must be a multiple of 0.2, 0.5. I think it should be done in another PR.

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,275 @@
...
+
+
+local function report()
+    local server_info, err = get()
+    if not server_info then
+        core.log.error("failed to get server_info: ", err)
+        return
+    end
+
+    if server_info.etcd_version == "unknown" then
```
spacewander, line 137: What if APISIX runs under standalone mode?

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
tokers: > Well, if we do so, the interval must be a multiple of 0.2, 0.5. I think it should be done in another PR.

All right, i changed my mind, i will also export the check interval in this PR.

tokers:
@membphis @moonming Please take a look when you have time~

conf/config-default.yaml
```diff
@@ -238,6 +238,7 @@ plugins:                          # plugin list (sorted in alphabetical order)
   - uri-blocker
   - wolf-rbac
   - zipkin
+  - server-info
```
membphis, line 241: we should disable this plugin by default

doc/zh-cn/plugins/server-info.md
```diff
@@ -0,0 +1,110 @@
...
+## 插件简介
+
+`server-info` 是一款能够定期将服务基本信息上报至 etcd 的插件。
+
+服务信息中每一项的含义如下：
+
+| 名称    | 类型 | 描述 |
+|---------|------|-------------|
+| up_time | integer | APISIX 服务实例当前的运行时间, 如果对 APSIX
+进行热更新操作，该值将被重置；普通的 reload 操作不会影响该值。 |
```
membphis, line 41: need a test case to confirm this

apisix/core/utils.lua
```diff
@@ -203,6 +209,26 @@ function _M.validate_header_value(value)
...
+    end
+
+    local buf = get_string_buf(max_hostname_len)
+
+    if C.gethostname(buf, max_hostname_len) == 0 then
+        buf[max_hostname_len - 1] = str_byte('\0')
+        hostname = ffi_string(buf, ffi.C.strlen(buf))
+
+    else
+        log.alert("ffi.C.gethostname() failed")
```
membphis, line 224: make a try:

use `io.popen("/bin/hostname")` to fetch the host name

apisix/plugins/server-info.lua
tokers: Since now server-info plugin only reports the basic info to etcd, when APISIX runs under standalone mode, we don't need to launch the timer.

conf/config-default.yaml
tokers: What's the reason for disabling this plugin by default? Users are no feeling about the behavior of server-info.

doc/zh-cn/plugins/server-info.md
tokers: OK.

apisix/core/utils.lua
tokers: That'd be concise, i will try it.

apisix/plugins/server-info.lua
tokers: Since for now this plugin only reports data to etcd, we can turn off the report just by disabling the plugin.

doc/zh-cn/plugins/server-info.md
tokers: Added.

apisix/core/utils.lua
tokers: Changed.

membphis:
@tokers CI is broken. You need to fix it: 

https://github.com/apache/apisix/pull/2926/checks?check_run_id=1509266841#step:11:286

```
#   Failed test 'TEST 3: fix: ensure only one timer is running - response_body - response is expected (repeated req 0, req 0)'
#   at /home/runner/work/apisix/apisix/test-nginx/lib/Test/Nginx/Socket.pm line 1589.
#          got: 'done
# not a single rotater run at the same time: 2020-12-07_08-05-09__error.log
# '
#     expected: 'done
# '
# Looks like you failed 1 test of 12.
t/plugin/log-rotate.t .....................

File: doc/plugins/server-info.md [hunk 12]
```diff
@@ -0,0 +1,104 @@
...
+## Name
+
+`server-info` is a plugin that reports basic server information to etcd periodically.
+
+The meaning of each item in server information is following:
+
+| Name    | Type | Description |
+|---------|------|-------------|
+| up_time | integer | Elapsed time since APISIX instance was launched, value will be reset when you hot updating APISIX but is kept for intact if you just reloading APISIX. |
+| boot_time | integer | Bootstrap timestamp of the APISIX instance, value will be reset when you hot updating APISIX but is kept for intact if you just reloading APISIX. |
```
spacewander, line 40: I think the value is different, as different workers write data to the same shdict.

spacewander (changes requested):

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,238 @@
...
+    end
+
+    local threshold = ceil(attr.report_interval / timers.check_interval())
+    local report_ttl = attr.report_ttl
+    local count = 0
+
+    local fn = function()
+        count = count + 1
+        if count == threshold then
+            report(nil, report_ttl)
```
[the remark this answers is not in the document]
tokers: Only the privilege agent process will do the report.

doc/plugins/server-info.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Name
+
+`server-info` is a plugin that reports basic server information to etcd periodically.
+
+The meaning of each item in server information is following:
+
+| Name    | Type | Description |
+|---------|------|-------------|
+| up_time | integer | Elapsed time since APISIX instance was launched, value will be reset when you hot updating APISIX but is kept for intact if you just reloading APISIX. |
```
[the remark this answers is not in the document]
tokers: Added. @spacewander

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,238 @@
...
+    core.log.info("server info: ", core.json.delay_encode(get()))
+
+    if core.config ~= require("apisix.core.config_etcd") then
+        -- we don't need to report server info if etcd is not in use.
+        return
+    end
+
+    local local_conf = core.config.local_conf()
+    local attr = core.table.try_read_attr(local_conf, "plugin_attr",
+                                          plugin_name)
```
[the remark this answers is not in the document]
tokers: Fixed.

apisix/plugins/server-info.lua
```diff
@@ -0,0 +1,238 @@
...
+    if not ok then
+        return false, err
+    end
+
+    return true
+end
+
+
+function _M.init()
+    local ok, err = encode_and_save(uninitialized_server_info(), true)
```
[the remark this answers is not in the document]
tokers: Yes, fixed.

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
tokers: Yes, fixed.

doc/plugins/server-info.md
tokers: Sorry i didn't capture your meaning.

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
spacewander: Miss the `apisix.timers` part? Better to do it by extending the background timer.

doc/plugins/server-info.md
spacewander: When reloading Nginx, there will be two workers A and B. B is just created and has a later boot time, but both workers share the same shdict.

tokers: Previously, workers will try to write server info to shdict in the init_worker phase, only when the key "server_info" doesn't exist, so server info written by old worker will not be overwritten by new workers, that is to say, the boot time will not be changed.

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
tokers: Do you mean the `premature` argument?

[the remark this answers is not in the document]
spacewander: Need to only run in worker 0 in the timer created by `apisix.timers`.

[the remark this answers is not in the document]
tokers: That's already handled since the timer only created by the privilege agent process.

doc/plugins/server-info.md
spacewander: > only when the key "server_info" doesn't exist

So where is the check? Do I miss something?

tokers: Now the logic about writing the uninitialized data was removed.

since after APISIX launches, timer will be created immediately which updates the server info according to existing one in shdict, the boot_time will be inherited (or the uninitialized data if no data in shdict).

apisix/plugins/server-info.lua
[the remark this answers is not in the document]
spacewander: @tokers 
My bad!

spacewander (approved):

membphis (approved):
LGTM

PR #2935, opened 2020-12-02
feat: Implement traffic splitting plugin

### LINKED ISSUE #2303: Proposal: inptroduce traffic split plugin
### Background

I observed the Pull Request https://github.com/apache/apisix/pull/2279 adds labels for `upstream` object, which inspires me that if we attach labels for each node in `upstream`,  then we can implement fancy traffic split feature.

### Design

Traffic split, which means to setup several route rules and schedule requests by respecting these rules to specified upstream (or a subset nodes in a single upstream). With this feature, we can support the [Canary Release](https://martinfowler.com/bliki/CanaryRelease.html#:~:text=Canary%20release%20is%20a%20technique,making%20it%20available%20to%20everybody.), [Blue Green Release](https://docs.cloudfoundry.org/devguide/deploy-apps/blue-green.html#:~:text=Blue%2Dgreen%20deployment%20is%20a,live%20and%20Green%20is%20idle.) for backend applications when they are releasing, it's useful to reduce the downtime when something fault happens.

But it's not a good idea to introduce these concepts into APISIX directly, since features like Blue Green Release is closer to business, not the infrastructure, so what we can do is introducing the concept traffic split , to abstract all the business logics into the traffic split rules (route rules), and it can be implemented as a plugin.

Both the Istio and [SMI](https://github.com/servicemeshinterface/smi-spec) support traffic split by `VirtualService` and `TrafficSplit` CRD respectively, although the latter is not perfect (at lease for now), we still can be aware of the similarity between them.

In general, we need two parts to define a traffic split rule, match and route, the `match` defines the conditions that needed to judge wether a request is eligible to apply the `route`, for instance, a `match` might be "the method of HTTP request must be `GET`, and the parameter `id` must be equal to `1006`. From APISIX's point of view, match conditions can be scoped to HTTP, TLS, so conditions can vary, like URI, method, headers, SNI and even other APISIX instance-scoped variables (hostname, env and etc). 

The APISIX Route already defines a fantastic match mechanism which is similar with the traffic split. But that not means we don't need to embed the match part in traffic split plugin, instead, we can reuse the match mechanism in Route as it's own match mechanism to further split requests that hit the same Route. See https://github.com/api7/lua-resty-radixtree#new for more details about Route match.

The route, decides the ultimate upstream that a request will go, either specifying the upstream name or with a label selector to filter an eligible subset from that upstream. What's more, multiple upstreams can be specified, each with a non-negative weight to support probabilistic selection. For now, node in APISIX Upstream doesn't contain labels attribute,  so we can implement it by the metadata map in node indirectly.

```json
# route requests to the nodes that has label release=canary in fake_upstream,
{
    "route": [
        {
            "upstream_id": "1",
            "labels": {
                "release": "canary"
            }
        }
    ]
}

# route 75% requests to upstream 2 and other 25% to upstream 3.
{
    "route": [
        {
            "upstream_id": "2",
            "weight": 75
        },
        {
            "upstream_id": "3",
            "weight": 25
        }
    ]
}
```

The Route already contains an upstream, after introducing the traffic split plugin, this upstream will be treated as the default upstream and will be used when requests can not hit the route rules.

#### Weighted Blue Green Release

Say we have two upstreams `app-blue` (id: 1) and `app-green` (id: 2), which represent the old and new releases for `app` respectively. Now we want to route 10% requests which UA is Android to `app-green`.

```json
[
    {
        "match": [
            {
                "vars": [
                    [ "http_user_agent", "~~", "Android"]
                ]
            }
        ],
        "route": [
            {
                "upstream_id": 2,
                "weight": 10
            }
        ]
    },
    {
        "route": [
            {
                "upstream_id": 1,
                "weight": 90
            }
        ]
    }
]
```

#### Canary Release in a single upstream

Say we updated a node in upstream `app` (id: 3), and this node have a unique label `release=canary`, other nodes in `app` has label `release=stable`, now we want to route requests which parameter `user_id` is between [13, 133] to this node.

```json
[
    {
        "match": [
            {
                "vars": [
                    [ "arg_user_id", ">=", 13],
                    [ "arg_user_id", "<=", 133]
                ]
            }
        ],
        "route": [
            {
                "upstream_id": 3,
                "labels": "release=canary"
            }
        ]
    },
    {
        "route": [
            {
                "upstream_id": 3,
                "labels": "release=stable"
            }
        ]
    }
]
```

### References

* VirtualService: https://istio.io/latest/docs/reference/config/networking/virtual-service
* TrafficSplit: https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-split/v1alpha3/traffic-split.md

### LINKED ISSUE #2603: request help: dynamic-upstream 插件如何与1.5版本的apisix融合及2.0版本融合，后期考虑将这个插件直接集成到apisix中吗
### Environment

* OS:

Firstsawyou:
close #2303  close #2603

Implement a traffic splitting plugin, the plugin is named `traffic-split`. The main function of the plug-in is to divide the request traffic according to the specified ratio and divert it to the corresponding "upstream". The functions of `Gray release`, `Blue-green release` and `Custom release` can be realized through this plugin.

Related issue：
[https://github.com/apache/apisix/issues/2303 ](https://github.com/apache/apisix/issues/2303)

[https://github.com/apache/apisix/issues/2603 ](https://github.com/apache/apisix/issues/2603)

Mailing list discussion address:
https://lists.apache.org/thread.html/rf02dc53a4af5d98d2513d89256b47466934d129af06d0bdcdb49cc8e%40%3Cdev.apisix.apache.org%3E 

### Pre-submission checklist:


moonming:
Please write a better title and desc

tokers:
@Firstsawyou CI failed, please fix it.

Firstsawyou:
@spacewander  @membphis 
Help me take a look at this test case: https://github.com/apache/apisix/pull/2935/files#diff-1badbb10ecfbb255409af122651c06a0b308ba0d51e672180d4e78719dce7e88R128. When executing this test case, the following error message is prompted:

```
nginx: [emerg] Lua code block missing the closing long bracket "]=]" in /home/runner/work/apisix/apisix/t/servroot/conf/nginx.conf:148
nginx: [emerg] Lua code block missing the closing long bracket "]=]" in /home/runner/work/apisix/apisix/t/servroot/conf/nginx.conf:148
nginx: [emerg] Lua code block missing the closing long bracket "]=]" in /home/runner/work/apisix/apisix/t/servroot/conf/nginx.conf:148
```

I don't know what caused this.

moonming:
@Firstsawyou What is the name of this plugin? the PR title and code file are not the same.

Firstsawyou:
> @Firstsawyou What is the name of this plugin? the PR title and code file are not the same.

updated.

moonming:
> > @Firstsawyou What is the name of this plugin? the PR title and code file are not the same.
> 
> updated.

Not yet , please read your desc again

t/plugin/dynamic-upstream.t
```diff
@@ -0,0 +1,213 @@
...
+1980, 1980, 1981, 1981, 1981, 1981
+--- no_error_log
+[error]
+
+
+
+=== TEST 5: Multiple vars rules and multiple plugin upstream
+--- config
+    location /t {
+        content_by_lua_block {
```
membphis, line 131: Reduce the total code length in content_by_lua_block

tokers:
We should use traffic split instead dynamic-upstream, in the future, the traffic split plugin will provide the labels based selector to choose nodes inside a upstream, this function is not covered by the name dynamic-upstream, which gives us a illusion this plugin is used for choosing one upstream from multiple upstreams.

t/plugin/dynamic-upstream.t
Firstsawyou: Thank you very much, I will try.

Firstsawyou:
> We should use traffic split instead dynamic-upstream, in the future, the traffic split plugin will provide the labels based selector to choose nodes inside a upstream, this function is not covered by the name dynamic-upstream, which gives us an illusion this plugin is used for choosing one upstream from multiple upstreams.

Ok i will update later.

Firstsawyou:
There was a problem when this `pr` handled the conflict. A related `pr` has been newly submitted.

moonming:
No, please fix it in this PR

spacewander (changes requested):

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Attributes
+
+| Name             | Type    | Requirement | Default | Valid   | Description                                                                              |
+| ---------------- | ------- | ----------- | ------- | ------- | ---------------------------------------------------------------------------------------- |
+| rules.match      | array[object]  | optional    |         |  | List of matching rules.                                                                    |
+| rules.match.vars | array[array] | optional    |     |  | A list consisting of one or more {var, operator, val} elements, like this: {{var, operator, val}, {var, operator, val}, ...}}. For example: {"arg_name", "==", "json"}, which means that the current request parameter name is json. The var here is consistent with the naming of Nginx internal variables, so request_uri, host, etc. can also be used; for the operator part, the currently supported operators are ==, ~=, ~~, >, <, in, has and !. For the specific usage of operators, please see the section `Operator List` below. |
```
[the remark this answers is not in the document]
Firstsawyou: Ok i will update later.

tokers:
@Firstsawyou Could you paste some sample traffic split configurations here to help us understand the schema better.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+
+function _M.access(conf, ctx)
+    if not conf or not conf.rules then
+        return
+    end
+
+    local upstreams, match_flag
+    for _, rule in pairs(conf.rules) do
+        match_flag = true
+        for _, single_match in pairs(rule.match) do
```
tokers, line 254: `match` is an array, so using `ipairs` here.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+            local expr, err = expr.new(single_match.vars)
```
tokers, line 255: Please assess the overheads of initilaizing the expr object in each request, maybe we should cache it.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+            if err then
+                return 500, err
+            end
+
+            match_flag = expr:eval()
+            if match_flag then
```
tokers, line 261: Shouldn't the logic relationship between all conditions is AND rather than OR?

@membphis @moonming @spacewander What do you think?

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+        return
+    end
+
+    local rr_up, err = lrucache(upstreams, nil, new_rr_obj, upstreams)
+    if not rr_up then
+        core.log.error("lrucache roundrobin failed: ", err)
+        return 500
+    end
+
+    local upstream = rr_up:find()
```
tokers, line 284: Why we need weighted round robin here? Just use weighted random is enough.

tokers (approved):

tokers:
@Firstsawyou Please resolve the conflicts.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+local upstreams_schema = {
+    type = "array",
+    items = {
+        type = "object",
+        properties = {
+            upstream_id = schema_def.id_schema,    -- todo: support upstream_id method
+            upstream = schema_def.upstream,
+            weight = {
+                description = "used to split traffic between different" ..
+                               "upstreams for plugin configuration",
```
tokers, line 103: Style: keep the double quote sign aligned.

apisix/plugins/traffic-split.lua
spacewander: Yes. It is expected to use `and`.

spacewander (approved):

Firstsawyou:
> @Firstsawyou Could you paste some sample traffic split configurations here to help us understand the schema better.

@tokers These are two cases, which will help you understand：
### Grayscale Release

Traffic is split according to the weight value configured by upstreams in the plugin (the rule of `match` is not configured, and `match` is passed by default). The request traffic is divided into 4:2, 2/3 of the traffic reaches the upstream of the `1981` port in the plugin, and 1/3 of the traffic reaches the upstream of the default `1980` port on the route.

```json
{
    "weight": 2
}
```

There is only a `weight` value in the plugin upstreams, which represents the weight value of the upstream traffic arriving on the route.

```shell
{
    "uri": "/index.html",
    "plugins": {
        "traffic-split": {
            "rules": [
                {
                    "upstreams": [
                        {
                            "upstream": {
                                "name": "upstream_A",
                                "type": "roundrobin",
                                "nodes": {
                                    "127.0.0.1:1981":10
                                },
                                "timeout": {
                                    "connect": 15,
                                    "send": 15,
                                    "read": 15
                                }
                            },
                            "weight": 4
                        },
                        {
                            "weight": 2
                        }
                    ]
                }
            ]
        }
    },
    "upstream": {
            "type": "roundrobin",
            "nodes": {
                "127.0.0.1:1980": 1
            }
    }
}'
```

### Custom Release

Multiple matching rules can be set in `match` (multiple conditions in `vars` are the relationship of `add`, and the relationship between multiple `vars` rules is the relationship of `or`; as long as one of the vars rules passes, it means `match` passed), only one is configured here, and the traffic is divided into 4:2 according to the value of `weight`. Among them, only the `weight` part represents the proportion of upstream on the route. When `match` fails to match, all traffic will only hit upstream on the route.

```shell
{
    "uri": "/index.html",
    "plugins": {
        "traffic-split": {
            "rules": [
                {
                    "match": [
                        {
                            "vars": [
                                ["arg_name","==","jack"],
                                ["http_user-id",">","23"],
                                ["http_apisix-key","~~","[a-z]+"]
                            ]
                        }
                    ],
                    "upstreams": [
                        {
                            "upstream": {
                                "name": "upstream_A",
                                "type": "roundrobin",
                                "nodes": {
                                    "127.0.0.1:1981":10
                                }
                            },
                            "weight": 4
                        },
                        {
                            "weight": 2
                        }
                    ]
                }
            ]
        }
    },
    "upstream": {
            "type": "roundrobin",
            "nodes": {
                "127.0.0.1:1980": 1
            }
    }
}'
```

apisix/plugins/traffic-split.lua
Firstsawyou: Yes, the expectation is `and`. Multiple matching rules can be set in `match` ,multiple conditions in `vars` are the relationship of `add`, and the relationship between multiple `vars` rules is the relationship of `or`; as long as one of the vars rules passes, it means `match` passed.

apisix/plugins/traffic-split.lua
Firstsawyou: The method of `random` is inaccurate in dividing traffic, it is only close to the division ratio. And `roundrobin` will be more precise.

moonming (changes requested):
This plugin needs to be mentioned on the project README and the readme of the plugins.

apisix/plugins/traffic-split.lua
tokers: Then the logic here should be:

```
if not expr:eval() then
    return
end
```

apisix/plugins/traffic-split.lua
tokers: The weighted random is adopted by Envoy, Is there any docs show the precise difference between them?

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+    end
+
+    return node
+end
+
+
+local function set_upstream(upstream_info, ctx)
+    local nodes = upstream_info.nodes
+    local new_nodes = {}
+    for addr, weight in pairs(nodes) do
```
tokers, line 177: We use array to arrange nodes, you should also support it.

tokers:
> > @Firstsawyou Could you paste some sample traffic split configurations here to help us understand the schema better.
> 
> @tokers These are two cases, which will help you understand：
> 
> ### Grayscale Release
> Traffic is split according to the weight value configured by upstreams in the plugin (the rule of `match` is not configured, and `match` is passed by default). The request traffic is divided into 4:2, 2/3 of the traffic reaches the upstream of the `1981` port in the plugin, and 1/3 of the traffic reaches the upstream of the default `1980` port on the route.
> 
> ```json
> {
>     "weight": 2
> }
> ```
> 
> There is only a `weight` value in the plugin upstreams, which represents the weight value of the upstream traffic arriving on the route.
> 
> ```shell
> curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
> {
>     "uri": "/index.html",
>     "plugins": {
>         "traffic-split": {
>             "rules": [
>                 {
>                     "upstreams": [
>                         {
>                             "upstream": {
>                                 "name": "upstream_A",
>                                 "type": "roundrobin",
>                                 "nodes": {
>                                     "127.0.0.1:1981":10
>                                 },
>                                 "timeout": {
>                                     "connect": 15,
>                                     "send": 15,
>                                     "read": 15
>                                 }
>                             },
>                             "weight": 4
>                         },
>                         {
>                             "weight": 2
>                         }
>                     ]
>                 }
>             ]
>         }
>     },
>     "upstream": {
>             "type": "roundrobin",
>             "nodes": {
>                 "127.0.0.1:1980": 1
>             }
>     }
> }'
> ```
> 
> ### Custom Release
> Multiple matching rules can be set in `match` (multiple conditions in `vars` are the relationship of `add`, and the relationship between multiple `vars` rules is the relationship of `or`; as long as one of the vars rules passes, it means `match` passed), only one is configured here, and the traffic is divided into 4:2 according to the value of `weight`. Among them, only the `weight` part represents the proportion of upstream on the route. When `match` fails to match, all traffic will only hit upstream on the route.
> 
> ```shell
> curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
> {
>     "uri": "/index.html",
>     "plugins": {
>         "traffic-split": {
>             "rules": [
>                 {
>                     "match": [
>                         {
>                             "vars": [
>                                 ["arg_name","==","jack"],
>                                 ["http_user-id",">","23"],
>                                 ["http_apisix-key","~~","[a-z]+"]
>                             ]
>                         }
>                     ],
>                     "upstreams": [
>                         {
>                             "upstream": {
>                                 "name": "upstream_A",
>                                 "type": "roundrobin",
>                                 "nodes": {
>                                     "127.0.0.1:1981":10
>                                 }
>                             },
>                             "weight": 4
>                         },
>                         {
>                             "weight": 2
>                         }
>                     ]
>                 }
>             ]
>         }
>     },
>     "upstream": {
>             "type": "roundrobin",
>             "nodes": {
>                 "127.0.0.1:1980": 1
>             }
>     }
> }'
> ```

I have several questions:

1. Why the upstream defined in traffic split has name?
2. The block which only contains weight is confused, the configuration should be self-contained as soon as possible, you nedd some mark to tell the admin that it points to the default upstream.
3. The upstream embedded inside traffic split brings some troubles such as health check, people wishes unhealthy nodes should be excluded, which is not implemented in your PR, a better way to achieve this is using upstream_id to reference already defined upstreams.
4. the `upstreams` should highlight its function, i.e. weighted selection, so maybe "weighted_upstreams" or "weighted_clusters" is more suitable, since we will have label based selector in the future, we should distinguish them.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+        if err then
+            return nil, err
+        end
+    end
+
+    return node
+end
+
+
+local function set_upstream(upstream_info, ctx)
```
tokers, line 174: logic in this function is highly duplicated with the partial logic in `http_access_phase`, we may abstract them so we can reuse it.

Firstsawyou:
> This plugin needs to be mentioned on the project README and the readme of the plugins.

Yes i will update later.

Firstsawyou:
> > > @Firstsawyou Could you paste some sample traffic split configurations here to help us understand the schema better.
> > 
> > 
> > @tokers These are two cases, which will help you understand：
> > ### Grayscale Release
> > Traffic is split according to the weight value configured by upstreams in the plugin (the rule of `match` is not configured, and `match` is passed by default). The request traffic is divided into 4:2, 2/3 of the traffic reaches the upstream of the `1981` port in the plugin, and 1/3 of the traffic reaches the upstream of the default `1980` port on the route.
> > ```json
> > {
> >     "weight": 2
> > }
> > ```
> > 
> > 
> > There is only a `weight` value in the plugin upstreams, which represents the weight value of the upstream traffic arriving on the route.
> > ```shell
> > curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
> > {
> >     "uri": "/index.html",
> >     "plugins": {
> >         "traffic-split": {
> >             "rules": [
> >                 {
> >                     "upstreams": [
> >                         {
> >                             "upstream": {
> >                                 "name": "upstream_A",
> >                                 "type": "roundrobin",
> >                                 "nodes": {
> >                                     "127.0.0.1:1981":10
> >                                 },
> >                                 "timeout": {
> >                                     "connect": 15,
> >                                     "send": 15,
> >                                     "read": 15
> >                                 }
> >                             },
> >                             "weight": 4
> >                         },
> >                         {
> >                             "weight": 2
> >                         }
> >                     ]
> >                 }
> >             ]
> >         }
> >     },
> >     "upstream": {
> >             "type": "roundrobin",
> >             "nodes": {
> >                 "127.0.0.1:1980": 1
> >             }
> >     }
> > }'
> > ```
> > 
> > 
> > ### Custom Release
> > Multiple matching rules can be set in `match` (multiple conditions in `vars` are the relationship of `add`, and the relationship between multiple `vars` rules is the relationship of `or`; as long as one of the vars rules passes, it means `match` passed), only one is configured here, and the traffic is divided into 4:2 according to the value of `weight`. Among them, only the `weight` part represents the proportion of upstream on the route. When `match` fails to match, all traffic will only hit upstream on the route.
> > ```shell
> > curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
> > {
> >     "uri": "/index.html",
> >     "plugins": {
> >         "traffic-split": {
> >             "rules": [
> >                 {
> >                     "match": [
> >                         {
> >                             "vars": [
> >                                 ["arg_name","==","jack"],
> >                                 ["http_user-id",">","23"],
> >                                 ["http_apisix-key","~~","[a-z]+"]
> >                             ]
> >                         }
> >                     ],
> >                     "upstreams": [
> >                         {
> >                             "upstream": {
> >                                 "name": "upstream_A",
> >                                 "type": "roundrobin",
> >                                 "nodes": {
> >                                     "127.0.0.1:1981":10
> >                                 }
> >                             },
> >                             "weight": 4
> >                         },
> >                         {
> >                             "weight": 2
> >                         }
> >                     ]
> >                 }
> >             ]
> >         }
> >     },
> >     "upstream": {
> >             "type": "roundrobin",
> >             "nodes": {
> >                 "127.0.0.1:1980": 1
> >             }
> >     }
> > }'
> > ```
> 
> I have several questions:
> 
> 1. Why the upstream defined in traffic split has name?
> 2. The block which only contains weight is confused, the configuration should be self-contained as soon as possible, you nedd some mark to tell the admin that it points to the default upstream.
> 3. The upstream embedded inside traffic split brings some troubles such as health check, people wishes unhealthy nodes should be excluded, which is not implemented in your PR, a better way to achieve this is using upstream_id to reference already defined upstreams.
> 4. the `upstreams` should highlight its function, i.e. weighted selection, so maybe "weighted_upstreams" or "weighted_clusters" is more suitable, since we will have label based selector in the future, we should distinguish them.

1. The name is to identify the upstream service name, usage scenario, etc.
2. I think it is not appropriate to add the default upstream configuration of `route` to only include the `weight`, because the configuration is repeated, and currently this plugin does not support some fields of the default upstream.
3. Yes, this is a version, we need to support it later.
4. I will update it later.

apisix/plugins/traffic-split.lua
Firstsawyou: We need `match_flag` to record the result of `expr:eval()`, because we need the value of `match_flag` to judge whether the `match` rule passes.

tokers: I mean, the relationship between each rule in `match` should be `and`, so the judgement can be short circuit, we can quit the loop once one of them is `false` .

membphis:
> 4\. the `upstreams` should highlight its function, i.e. weighted selection, so maybe "weighted_upstreams" or "weighted_clusters" is more suitable, since we will have label based selector in the future, we should distinguish them.

`weighted_upstreams` is better than before, but I think we should use the singular form here, this field only works for one upstream.

I prefer the name `weighted_upstream`.

apisix/plugins/traffic-split.lua
Firstsawyou: There is no relevant docs to show their differences, but there are obvious differences in writing test cases.

apisix/plugins/traffic-split.lua
Firstsawyou: I think there is no need for caching for the time being, the expr object here is limited. Secondly, they need to be cached, which is not an easy task for me.

Firstsawyou:
> > 4. the `upstreams` should highlight its function, i.e. weighted selection, so maybe "weighted_upstreams" or "weighted_clusters" is more suitable, since we will have label based selector in the future, we should distinguish them.
> 
> `weighted_upstreams` is better than before, but I think we should use the singular form here, this field only works for one upstream.
> 
> I prefer the name `weighted_upstream`.

agree, updated.

wfgydbu:
Hi all, thansk for this great feature! One quick question, In this plugin, how do you deal with healthcheck and
retry mechanism? are you planning to support it?

tokers:
> Hi all, thansk for this great feature! One quick question, In this plugin, how do you deal with healthcheck and
> retry mechanism? are you planning to support it?

This feature will be supported, please pay attention to this PR continuously.

wfgydbu:
> > Hi all, thansk for this great feature! One quick question, In this plugin, how do you deal with healthcheck and
> > retry mechanism? are you planning to support it?
> 
> This feature will be supported, please pay attention to this PR continuously.

Great! Looking forward to it !

Firstsawyou:
> Hi all, thansk for this great feature! One quick question, In this plugin, how do you deal with healthcheck and
> retry mechanism? are you planning to support it?

will support soon.

moonming:
`traffic shaping` is a better ame

moonming:
is this PR still active？

README.md
```diff
@@ -99,6 +99,7 @@ A/B testing, canary release, blue-green deployment, limit rate, defense against
   - [Health Checks](doc/health-check.md): Enable health check on the upstream node, and will automatically filter unhealthy nodes during load balancing to ensure system stability.
   - Circuit-Breaker: Intelligent tracking of unhealthy upstream services.
   - [Proxy Mirror](doc/plugins/proxy-mirror.md): Provides the ability to mirror client requests.
+  - [Traffic Split](doc/plugins/traffic-split.md): Supports dividing the request traffic according to the specified proportional relationship.
```
moonming, line 102: Boring description.  Why we need this plugin?

moonming (changes requested):

Firstsawyou:
> is this PR still active？

Not active anymore.

README.md
Firstsawyou: `Support gray-scale release, blue-green release and custom release function`. Is this description feasible?

tokers (changes requested):
I accidently approved this PR.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    local ok, err = upstream.check_schema(up_conf)
+    if not ok then
+        return 500, err
+    end
+
+    local matched_route = ctx.matched_route
+    upstream.set(ctx, up_conf.type .. "#route_" .. matched_route.value.id,
```
wfgydbu, line 224: `up_conf.type .. "#route_" .. matched_route.value.id` would produce something like "roundrobin#route_1", which will further to be used as the upstream key in https://github.com/apache/apisix/blob/1f72f6726055e81cc245409d498084740cb3e1af/apisix/balancer.lua#L242.
Will it be a problem? In my opinion, it will cause different upstreams to use the same key when picking node.

spacewander: Yes, you are right. We should provide an unique key for the generated upstream.

spacewander (changes requested):

apisix/plugins/traffic-split.lua
Firstsawyou: Yes, thank you very much for your comments.

Firstsawyou: @spacewander 
I tried it, but couldn't find a good way to generate unique keys, because in the plug-in, except for the nodes information of upstream, the other configuration information is the same.
Do you have any good suggestions for this problem?

spacewander: What about generating a UUID as a suffix? We don't need to reuse in this occasion.

Firstsawyou: I have tried the uuid method, but this will result in a situation where all requests go to only one upstream.

```
got: '1982, 1982, 1982, 1982, 1982
# '
#     expected: '1980, 1981, 1981, 1982, 1982
```

apisix/plugins/traffic-split.lua
Firstsawyou: updated.

Firstsawyou:
> Hi all, thansk for this great feature! One quick question, In this plugin, how do you deal with healthcheck and
> retry mechanism? are you planning to support it?

Hi @wfgydbu ,
I am very sorry. Regarding the health check and retry mechanism, I plan to implement it in the next version, which is currently not supported. I have established a related issue to follow up on this. You can continue to follow the progress of this issue.
issue:  [https://github.com/apache/apisix/issues/3094 ](https://github.com/apache/apisix/issues/3094)

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,310 @@
...
+    end
+
+    return node
+end
+
+
+local function set_pass_host(ctx, upstream_info, host)
+    -- Currently only supports a single upstream of the domain name.
+    -- When the upstream is `IP`, do not do any `pass_host` operation.
+    if not core.utils.parse_ipv4(host) and not core.utils.parse_ipv6(host) then
```
membphis, line 177: bad style too

Firstsawyou: updated.

README.md
```diff
@@ -99,6 +99,7 @@ A/B testing, canary release, blue-green deployment, limit rate, defense against
   - [Health Checks](doc/health-check.md): Enable health check on the upstream node, and will automatically filter unhealthy nodes during load balancing to ensure system stability.
   - Circuit-Breaker: Intelligent tracking of unhealthy upstream services.
   - [Proxy Mirror](doc/plugins/proxy-mirror.md): Provides the ability to mirror client requests.
+  - [Traffic Split](doc/plugins/traffic-split.md): Support the functions of gray release, blue-green release and custom release.
```
tokers, line 102: Not so intuitive.

Use: allows users to incrementally direct percentages of traffic between various upstreams.

README_CN.md
```diff
@@ -97,6 +97,7 @@ A/B 测试、金丝雀发布(灰度发布)、蓝绿部署、限流限速、抵
   - [健康检查](doc/zh-cn/health-check.md)：启用上游节点的健康检查，将在负载均衡期间自动过滤不健康的节点，以确保系统稳定性。
   - 熔断器: 智能跟踪不健康上游服务。
   - [代理镜像](doc/zh-cn/plugins/proxy-mirror.md): 提供镜像客户端请求的能力。
+  - [流量拆分](doc/zh-cn/plugins/traffic-split.md): 支持灰度发布、蓝绿发布和自定义发布的功能。
```
tokers, line 100: Ditto

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,314 @@
...
+end
+
+
+function _M.access(conf, ctx)
+    if not conf or not conf.rules then
+        return
+    end
+
+    local upstreams, match_flag
+    for _, rule in pairs(conf.rules) do
```
tokers, line 272: For array-like table, use ipairs.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,314 @@
...
+        match_flag = true
+        for _, single_match in ipairs(rule.match) do
+            local expr, err = expr.new(single_match.vars)
+            if err then
```
tokers, line 276: We should log this error.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,314 @@
...
+                return 500, err
+            end
+
+            match_flag = expr:eval()
+            if match_flag then
```
tokers, line 281: Why we can break when just the current match is true, we should check all matchs.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,314 @@
...
+        return
+    end
+
+    local rr_up, err = lrucache(upstreams, nil, new_rr_obj, upstreams)
+    if not rr_up then
+        core.log.error("lrucache roundrobin failed: ", err)
+        return 500
+    end
+
+    local upstream = rr_up:find()
```
tokers, line 304: Do not use wrr here, instead, please use weighted random algorithm.

Say we have two upstreams and the weight of them are same,  when rr state is just reset, the first selected upstream is fixed, depending on their order, this is the inherent drawback of wrr. it might overwhelm the first selected upstream after the wrr state is reset.

Also, wrr state will be reset once config is changd, the percentage curve is not so smooth.

doc/README.md
```diff
@@ -81,6 +81,7 @@
 * [request-validation](plugins/request-validation.md): Validates requests before forwarding to upstream.
 * [proxy-mirror](plugins/proxy-mirror.md): Provides the ability to mirror client requests.
 * [api-breaker](plugins/api-breaker.md): Circuit Breaker for API that stops requests forwarding to upstream in case of unhealthy state.
+* [traffic-split](plugins/traffic-split.md): The traffic division plug-in divides the request traffic according to the specified ratio and diverts it to the corresponding upstream; through this plug-in, gray-scale publishing, blue-green publishing and custom publishing functions can be realized.
```
tokers, line 84: Why traffic division?

apisix/plugins/traffic-split.lua
Firstsawyou: The design logic here is: the relationship of `or` between multiple `vars` in `match`. Therefore, when there is a `vars` match that is `true`, it can no longer match.
For example, in the following example: When the first `vars` matches `true`, then the second `vars` does not need to match.

```lua
"match": [
      {
          "vars": [
              ["arg_name","==","jack"],
              ["http_user-id",">","23"],
              ["http_apisix-key","~~","[a-z]+"]
          ],
          "vars": [
              ["arg_name2","==","rose"],
              ["http_user-id2","!",">","33"],
              ["http_apisix-key2","~~","[a-z]+"]
          ]
      }
  ]
```

doc/README.md
Firstsawyou: There is a mistake here, I will fix it later.

apisix/plugins/traffic-split.lua
spacewander: Look like it is because the wrr is used as the picker? If we use random, this isn't a problem. Since the weight and node are changed per request, it is not safe to reuse the picker.

Firstsawyou: I think this does not affect the results of our choice of different upstream. Here wrr is used as the picker the same as the picker in upstream nodes. They are able to work normally.

spacewander: I think it again. Look like we can assign a virtual id to each upstream of the `upstreams` field, then use the vid as the suffix, so that we can reuse the upstream without messing up.

README.md
Firstsawyou: updated.

apisix/plugins/traffic-split.lua
Firstsawyou: I set up a related `issue` to record this issue, let’s look at the community’s feedback first. Then make adjustments.

t/plugin/traffic-split.t
```diff
@@ -1150,3 +1144,75 @@ GET /t
 1980, 1981, 1981, 1982, 1982
 --- no_error_log
 [error]
+
+
+
+=== TEST 34: multiple upstream
```
spacewander, line 1150: Better to add a test for "multiple upstream + empty_upstream is used"

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,322 @@
...
+            anyOf = {
+                {type = "string"},
+                {type = "number"},
+                {type = "boolean"},
+                {
+                    type = "array",
+                    items = {
+                        anyOf = {
+                            {
+                            type = "string",
```
tokers, line 60: Bad indent.

apisix/plugins/traffic-split.lua
```diff
@@ -0,0 +1,322 @@
...
+                            minLength = 1, maxLength = 100
```
tokers, line 61: Ditto

apisix/plugins/traffic-split.lua
tokers: OK

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+  - [**Custom Release**](#custom-release)
+- [**Test Plugin**](#test-plugin)
+  - [**Grayscale Test**](#grayscale-test)
+  - [**Blue-green Test**](#blue-green-test)
+  - [**Custom Test**](#custom-test)
+- [**Disable Plugin**](#disable-plugin)
+
+## Name
+
+The traffic division plugin divides the request traffic according to the specified ratio and diverts it to the corresponding upstream; through this plugin, gray-scale publishing, blue-green publishing and custom publishing functions can be realized.
```
tokers, line 38: traffic division => traffic split.

conf/config-default.yaml
```diff
@@ -246,6 +246,7 @@ plugins:                          # plugin list (sorted in alphabetical order)
   - wolf-rbac
   - zipkin
   # - server-info
+  - traffic-split
```
tokers, line 249: Better to disable this plugin by default.

Firstsawyou: This plugin does not involve changes to the core code, I think we can enable.

t/plugin/traffic-split.t
Firstsawyou: updated.

apisix/plugins/traffic-split.lua
Firstsawyou: updated.

t/plugin/traffic-split.t
```diff
@@ -0,0 +1,1289 @@
...
+--- response_body
+1981, 1982
+--- grep_error_log eval
+qr/upstream_key: roundrobin#route_1_\d/
+--- grep_error_log_out
+upstream_key: roundrobin#route_1_1
+upstream_key: roundrobin#route_1_2
+--- no_error_log
+[error]
+--- LAST
```
spacewander, line 1219: Should remove the mark

Firstsawyou: fixed.

spacewander (approved):

doc/README.md
```diff
@@ -81,6 +81,7 @@
 * [request-validation](plugins/request-validation.md): Validates requests before forwarding to upstream.
 * [proxy-mirror](plugins/proxy-mirror.md): Provides the ability to mirror client requests.
 * [api-breaker](plugins/api-breaker.md): Circuit Breaker for API that stops requests forwarding to upstream in case of unhealthy state.
+* [traffic-split](plugins/traffic-split.md): The traffic split plugin divides the request traffic according to the specified ratio and diverts it to the corresponding upstream; through this plug-in, gray-scale publishing, blue-green publishing and custom publishing functions can be realized.
```
moonming, line 84: I think you should keep the desc as readme.md

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+    - [**Custom Release**](#custom-release)
+  - [**Test Plugin**](#test-plugin)
+    - [**Grayscale Test**](#grayscale-test)
+    - [**Blue-green Test**](#blue-green-test)
+    - [**Custom Test**](#custom-test)
+  - [**Disable Plugin**](#disable-plugin)
+
+## Name
+
+The traffic split plugin divides the request traffic according to the specified ratio and diverts it to the corresponding upstream; through this plugin, gray-scale publishing, blue-green publishing and custom publishing functions can be realized.
```
moonming, line 38: ditto

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+## 如何启用
+
+在插件的 weighted_upstreams 中只有 `weight` 值，表示到达默认 `route` 上的 upstream 流量权重值。
+
+```json
+{
+    "weight": 2
+}
+```
+
```
moonming, line 68: I still don't how to enable the plugin

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+### 灰度发布
+
+根据插件中 weighted_upstreams 配置的 `weight` 值做流量分流（不配置 `match` 的规则，已经默认 `match` 通过）。将请求流量按 4:2 划分，2/3 的流量到达插件中的 `1981` 端口上游， 1/3 的流量到达 route 上默认的 `1980` 端口上游。
```
moonming, line 71: What is "4:2" and 1/3?

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+            }
+    }
+}'
+```
+
+插件设置了请求的匹配规则并设置端口为`1981`的 upstream，route 上具有端口为`1980`的upstream。
+
+## 测试插件
+
+### 灰度测试
```
moonming, line 266: Why are configuration and testing so far apart?  Hard to understand

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+
+$ curl http://127.0.0.1:9080/index.html -i
+HTTP/1.1 200 OK
+Content-Type: text/html; charset=utf-8
+......
+
+world 1981
+```
+
+### 蓝绿测试
```
moonming, line 286: bad desc

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+```shell
+$ curl 'http://127.0.0.1:9080/index.html?name=jack' -H 'new-release: blue' -i
+HTTP/1.1 200 OK
+Content-Type: text/html; charset=utf-8
+......
+
+world 1981
+```
+
+当 `match` 匹配通过后，所有请求都命中到插件配置的 `upstream`，否则命中 `route` 上配置的 upstream 。
```
moonming, line 297: where is “match”？ I don't know how to use this plugin

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,413 @@
...
+
+### 自定义测试
+
+**示例1：**
+
+**在`match` 规则校验通过后, 2/3 的请求命中到1981端口的upstream, 1/3 命中到1980端口的upstream。**
```
moonming, line 303: Ditto

moonming (changes requested):

doc/README.md
Firstsawyou: ok.

doc/zh-cn/plugins/traffic-split.md
Firstsawyou: Let me update the document again, it is a bit confusing to understand

doc/zh-cn/plugins/traffic-split.md
Firstsawyou: updated.

doc/zh-cn/plugins/traffic-split.md
Firstsawyou: updated.

doc/zh-cn/plugins/traffic-split.md
Firstsawyou: `match` In the example configuration, the configuration example and the test have been put together.

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,409 @@
...
+  - [**Grayscale Release**](#grayscale-release)
+  - [**Blue-green Release**](#blue-green-release)
+  - [**Custom Release**](#custom-release)
+- [**Disable Plugin**](#disable-plugin)
+
+## Name
+
+The traffic split plugin allows users to incrementally direct percentages of traffic between various upstreams.
+
+Note: Since the selection of different upstream in the plugin is based on the roundrobin algorithm, the ratio of traffic distribution is not completely accurate when the algorithm state is reset.
```
moonming, line 36: Can not understand

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,409 @@
...
+| rules.weighted_upstreams.upstream.timeout  | object | optional    |  15     |   | Set the timeout period for connecting, sending and receiving messages (time unit: second, all default to 15 seconds).  |
+| rules.weighted_upstreams.upstream.pass_host | enum | optional    | "pass"  | ["pass", "node", "rewrite"]  | pass: pass the host requested by the client, node: pass the host requested by the client; use the host configured with the upstream node, rewrite: rewrite the host with the value configured by the upstream_host. |
+| rules.weighted_upstreams.upstream.name      | string | optional    |        |   | Identify the upstream service name, usage scenario, etc.  |
+| rules.weighted_upstreams.upstream.upstream_host | string | optional    |    |   | Only valid when pass_host is configured as rewrite.    |
+| rules.weighted_upstreams.weight | integer | optional    | weight = 1   |  | The traffic is divided according to the `weight` value, and the roundrobin algorithm is used to divide multiple `weight`. |
+
+## How To Enable
```
moonming, line 55: The content is nothing with how to enable

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,420 @@
...
+| rules.weighted_upstreams.upstream.timeout | object | 可选   |  15     |        | 设置连接、发送消息、接收消息的超时时间(时间单位：秒，都默认为 15 秒)。 |
+| rules.weighted_upstreams.upstream.pass_host  | enum | 可选   | "pass"   | ["pass", "node", "rewrite"]  | pass: 透传客户端请求的 host, node: 不透传客户端请求的 host; 使用 upstream node 配置的 host, rewrite: 使用 upstream_host 配置的值重写 host 。 |
+| rules.weighted_upstreams.upstream.name  | string | 可选   |        |  | 标识上游服务名称、使⽤场景等。 |
+| rules.weighted_upstreams.upstream.upstream_host | string | 可选   |        |        | 只在 pass_host 配置为 rewrite 时有效。 |
+| rules.weighted_upstreams.weight       | integer | 可选   |   weight = 1     |        | 根据 `weight` 值做流量划分，多个 weight 之间使用 roundrobin 算法划分。|
+
+## 如何启用
+
+traffic-split 插件主要由 `match` 和 `weighted_upstreams` 两部分组成，`match` 是自定义的条件规则，`weighted_upstreams` 是 upstream 的信息。在使用插件时，至少需要配置 `weighted_upstreams` 部分，这样将默认 `match` 规则通过，会根据 `weighted_upstreams` 中的 `weight` 值，逐步引导各个 upstream 之间的流量比例。你也可以同时配置 `match` 和 `weighted_upstreams`，这样只有 `match` 规则匹配通过后，才会对 `weighted_upstreams` 中的流量进行划分。
```
moonming, line 57: The grammar is not clear

doc/zh-cn/plugins/traffic-split.md
```diff
@@ -0,0 +1,420 @@
...
+{
+    "weight": 2
+}
+```
+
+下面提供插件的使用示例，这将有助于你对插件使用上的理解。
+
+### 灰度发布
+
+不配置 `match` 规则部分（已经默认 `match` 通过），根据插件中 weighted_upstreams 配置的 `weight` 值做流量分流。将 `插件 upstream` 与 `route 的 upstream` 请求流量按 3:2 进行划分，其中 60% 的流量到达插件中的 `1981` 端口的 upstream， 40% 的流量到达 route 上默认 `1980` 端口的 upstream。
```
moonming, line 71: ditto

moonming (changes requested):

doc/plugins/traffic-split.md
Firstsawyou: Currently, the `roundrobin` algorithm is used when selecting the upstream, and the traffic distribution ratio may not be completely consistent. So I made a note in the document.

doc/plugins/traffic-split.md
Firstsawyou: updated.

doc/zh-cn/plugins/traffic-split.md
Firstsawyou: updated.

doc/plugins/traffic-split.md
tokers: May change to:

```
Note the ratio between each upstream may not so accurate since the drawback of weighted round robin algorithm (especially when the wrr state is reset).
```

Firstsawyou: updated.

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,409 @@
...
+
+```json
+{
+    "weight": 2
+}
+```
+
+## How To Enable
+
+The following provides examples of plugin usage, which will help you understand the use of plugin.
```
tokers, line 67: Here you just need to add the necessary steps to enable this plugin.

The `Grayscale`, `BlueGreenVersion` is some features we can implement resorting to this plugin, it should't be written inside `How to Enable`, they are Best Practices.

tokers, line 67: Here you just need to add the necessary steps to enable this plugin.

The `Grayscale`, `BlueGreenVersion` is some features we can implement resorting to this plugin, it should't be written inside `How to Enable`, they are Best Practices.

Firstsawyou: I think it is appropriate to write the grayscale version and the blue-green version of the example here, because it includes the basic usage scenarios of this plugin. This shows the usage details of the plugin.

tokers: Note the title is "How to enable", not "Example".

Firstsawyou: "How to enable" has been changed to "Example". ^ _ ^

tokers: No! See other plugin documents, we need How to enable section.

Firstsawyou: added.

tokers (approved):

moonming (approved):

doc/plugins/traffic-split.md
```diff
@@ -0,0 +1,453 @@
...
+    "plugins": {
+        "traffic-split": {
+            "rules": [
+                {
+                    "match": [
+                        {
+                            "vars": [
+                                ["arg_name","==","jack"],
+                                ["http_user-id",">","23"],
+                                ["http_apisix-key","~~","[a-z]+"]
```
juzhiyuan, line 264: This should not work :(

Issue #2818, opened 2020-11-22
[DISCUSS]: Use luasocket lib instead of `curl` in "bin/apisix" 


starsz:
Hi, I advise using luasocket lib instead of `curl` in "bin/apisix".

Although, it's very convenient to get HTTP responses by using "io.open" and "curl". 
But, I think it's not an elegant way and has some limitations. Such as can't distinguish error info between `Connection refused` and empty response.

So I think we can use luasocket lib and create a common function in "cli/util.lua", so that we can deal with some HTTP request better in "bin/apisix".

releated: #2718

spacewander:
Do you mean writing a new HTTP client based on luasocket? Since lua-resty-http can't be used without `ngx`.

tokers:
What about just use luasocket?

tokers:
> Do you mean writing a new HTTP client based on luasocket? Since lua-resty-http can't be used without `ngx`.

The luasocket lib already has a HTTP client implementation.

starsz:
Hi, @spacewander @tokers, Yeah, luasocket had an HTTP client implementation already. 
See: http://w3.impa.br/~diego/software/luasocket/http.html
So, I want to use luasocket lib instead of curl in "bin/apisix".

tokers:
But i'm not sure whether it's built in? Or users have to install it before they can use `bin/apisix`.

starsz:
> But i'm not sure whether it's built in? Or users have to install it before they can use `bin/apisix`.

Users need to do `make deps` before they use `bin/apisix`. So I think it can be solved.

tokers:
> > But i'm not sure whether it's built in? Or users have to install it before they can use `bin/apisix`.
> 
> Users need to do `make deps` before they use `bin/apisix`. So I think it can be solved.

OK, that's good.

membphis:
We can make a try.

Need to pay attention to the use of different systems(ubuntu, centos and mac os), whether there are compatibility issues.

moonming:
Can we specify the path of openresty to solve this problem?  So we can use lua-resty-http.

membphis:
Do you mean to run the CLI in fake `openresty` ENV? Is it the same as the "resty" command?

moonming:
yes

Thanks,
Ming Wen
Twitter: _WenMing


YuanSheng Wang <notifications@github.com> 于2020年11月23日周一 下午11:42写道：

spacewander:
[...]

Issue #2718, opened 2020-11-12
[DISCUSS] The error message "got malformed version message from etcd" is ambiguous


tokers:
When the etcd initialization fails, the output error message:

```
got malformed version message: "" from etcd
```

Which is ambiguous, i think the error details should be exposed. What do you think?

Firstsawyou:
I agree with you. I have also encountered such a situation, this prompt message is not clear.

nic-chen:
agree +1

membphis:
agree +1

BTW, how can we get this error message? is it a bug?

tokers:
> agree +1
> 
> BTW, how can we get this error message? is it a bug?

It's note a bug but a not so friendly hint when we cannot get the etcd version when the etcd cluster is abnormal, like cluster without leader and we are timed out, or even the cluster is totally offline.

tokers:
@starsz Would you like to optimize this? Thanks!

starsz:
> @starsz Would you like to optimize this? Thanks!

No problem. My pleasure!

PR #2965, opened 2020-12-04
feat: use luasocket instead of curl  in etcd.lua

starsz:

fix: #2818
fix: #2718 

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -227,6 +227,7 @@ function _M.init(env, show_output)
             if res_put:find("404 page not found", 1, true) then
                 errmsg = str_format("gRPC gateway is not enabled in etcd cluster \"%s\",",
                                     "which is required by Apache APISIX\n")
+                util.die(errmsg)
             end
 
             if res_put:find("error", 1, true) then
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -19,15 +19,20 @@ local base64_encode = require("base64").encode
 local dkjson = require("dkjson")
 local util = require("apisix.cli.util")
 local file = require("apisix.cli.file")
+local http = require("socket.http")
+local ltn12 = require("ltn12")
 
 local type = type
 local ipairs = ipairs
 local print = print
 local tonumber = tonumber
 local str_format = string.format
+local table_concat = table.concat
 
 local _M = {}
 
+-- Timeout for all I/O operations
+http.TIMEOUT = 3
 
 local function parse_semantic_version(ver)
     local errmsg = "invalid semantic version: " .. ver
@@ -106,9 +111,6 @@ function _M.init(env, show_output)
 
     local etcd_conf = yaml_conf.etcd
 
-    local timeout = etcd_conf.timeout or 3
-    local uri
-
     -- convert old single etcd config to multiple etcd config
     if type(yaml_conf.etcd.host) == "string" then
         yaml_conf.etcd.host = {yaml_conf.etcd.host}
@@ -132,22 +134,23 @@ function _M.init(env, show_output)
 
     -- check the etcd cluster version
     for index, host in ipairs(yaml_conf.etcd.host) do
-        uri = host .. "/version"
-        local cmd = str_format("curl -s -m %d %s", timeout * 2, uri)
-        local res = util.execute_cmd(cmd)
-        local errmsg = str_format("got malformed version message: \"%s\" from etcd\n",
-                                  res)
+        local version_url = host .. "/version"
+        local errmsg
+
+        local res, err = http.request(version_url)
+        if err and type(err) == "string" then
+            errmsg = str_format("request etcd endpoint \'%s\' error, %s\n", version_url, err)
+            util.die(errmsg)
+        end
 
         local body, _, err = dkjson.decode(res)
-        if err then
+        if err or (body and not body["etcdcluster"]) then
+            errmsg = str_format("got malformed version message: \"%s\" from etcd \"%s\"\n", res,
+                                version_url)
             util.die(errmsg)
         end
 
         local cluster_version = body["etcdcluster"]
-        if not cluster_version then
-            util.die(errmsg)
-        end
-
         if compare_semantic_version(cluster_version, env.min_etcd_version) then
             util.die("etcd cluster version ", cluster_version,
                      " is less than the required version ",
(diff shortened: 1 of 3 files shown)


apisix/cli/etcd.lua
```diff
@@ -132,22 +134,23 @@ function _M.init(env, show_output)
...
-        local errmsg = str_format("got malformed version message: \"%s\" from etcd\n",
-                                  res)
+        local version_url = host .. "/version"
+        local errmsg
 
-        local body, _, err = dkjson.decode(res)
-        if err then
+        local res, err = http.request(version_url)
+        if err and type(err) == "string" then
+            errmsg = str_format("request etcd endpoint \'%s\' error, %s\n", host, err)
```
spacewander, line 142: Better to log the full uri

apisix/cli/etcd.lua
```diff
@@ -160,27 +163,39 @@ function _M.init(env, show_output)
...
-                             .. " --max-time " .. timeout * 2 .. " --retry 1 2>&1"
+            local response_body = {}
+            local _, err = http.request{url = auth_url, method = "POST",
+                                        source = ltn12.source.string(post_json_auth),
+                                        sink = ltn12.sink.table(response_body),
+                                        headers = {["Content-Length"] = #post_json_auth}}
+
+            -- err is string type
+            if err and type(err) == "string" then
+                errmsg = str_format("request etcd endpoint \"%s\" error, %s\n", host, err)
```
spacewander, line 186: Better to log the full uri

apisix/cli/etcd.lua
```diff
@@ -191,26 +206,36 @@ function _M.init(env, show_output)
...
+            end
 
-            local res = util.execute_cmd(cmd)
-            if res:find("error", 1, true) then
+            local _, err = http.request{url = put_url, method = "POST",
+                                        source = ltn12.source.string(post_json),
+                                        sink = ltn12.sink.table(response_body),
+                                        headers = headers}
+            if err and type(err) == "string" then
+                errmsg = str_format("request etcd endpoint \"%s\" error, %s\n", host, err)
```
spacewander, line 223: Better to log the full uri

spacewander (changes requested):

.travis/apisix_cli_test.sh
```diff
@@ -817,3 +817,53 @@ fi
...
+git checkout conf/config.yaml
+
+echo '
+etcd:
+  host:
+    - "http://127.0.0.1:2389"
+  prefix: "/apisix"
+' > conf/config.yaml
+
+make init &>/tmp/apisix_temp &
```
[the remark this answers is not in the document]
starsz: Because the exit code of `make init` is 1. It will terminate the shell.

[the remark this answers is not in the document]
spacewander: @starsz 
You can try this way:
https://github.com/apache/apisix/blob/c41aa8534a18f43ef0c3cd6780f106595f9f9b88/.travis/apisix_cli_test_in_ci.sh#L43

apisix/cli/etcd.lua
```diff
@@ -132,22 +134,23 @@ function _M.init(env, show_output)
...
-        local cmd = str_format("curl -s -m %d %s", timeout * 2, uri)
-        local res = util.execute_cmd(cmd)
-        local errmsg = str_format("got malformed version message: \"%s\" from etcd\n",
-                                  res)
+        local version_url = host .. "/version"
+        local errmsg
 
-        local body, _, err = dkjson.decode(res)
-        if err then
+        local res, err = http.request(version_url)
```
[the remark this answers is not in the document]
starsz: Right. The library doesn't support max-timeout.

starsz:
Hi, @spacewander, please have a look when you are free.

apisix/cli/etcd.lua
```diff
@@ -160,31 +163,38 @@ function _M.init(env, show_output)
...
-            if err_auth then
-                util.die(cmd_auth, "\n", res_auth)
+            local post_json_auth = dkjson.encode(json_auth)
+            local response_body = {}
+            local _, err = http.request{url = auth_url, method = "POST",
+                                        source = ltn12.source.string(post_json_auth),
+                                        sink = ltn12.sink.table(response_body),
+                                        headers = {["Content-Length"] = #post_json_auth}}
+            -- err is string type
+            if err and type(err) == "string" then
```
spacewander, line 184: Better to add more info about the `err` handle, so that people can know what you are doing without looking into luasocket's doc.

starsz: @spacewander.Sorry, I can't get your point.
That we need to distinguish the `connection refused` and `timeout`

spacewander: I mean, why we need to check the type of `err`? We need to clarify it in the comment, so that people can know it without searching luasocket's doc.

starsz: OK. I got it now.

starsz: Added.

spacewander: I think it would be better to use `if not res then handle err`? Since the first argument is nil when error happened.

starsz: Well. Agree with you.

apisix/cli/etcd.lua
```diff
@@ -179,14 +179,14 @@ function _M.init(env, show_output)
 
             local post_json_auth = dkjson.encode(json_auth)
             local response_body = {}
-            local _, err = http.request{url = auth_url, method = "POST",
+            local res, err = http.request{url = auth_url, method = "POST",
                                         source = ltn12.source.string(post_json_auth),
                                         sink = ltn12.sink.table(response_body),
                                         headers = {["Content-Length"] = #post_json_auth}}
             -- In case of failure, request returns nil followed by an error message.
             -- Else the first return value is just the number 1
```
spacewander, line 187: The first return value isn't the response body?

starsz: Yep. It has two forms.
You can see more details about it.

>In case of failure, the function returns nil followed by an error message. If successful, the simple form returns the response body as a string, followed by the response status code, the response headers, and the response status line. The generic function returns the same information, except the first return value is just the number 1 (the body goes to the sink).

http://w3.impa.br/~diego/software/luasocket/http.html

spacewander: Got it

spacewander (approved):

apisix/cli/etcd.lua
```diff
@@ -195,31 +210,44 @@ function _M.init(env, show_output)
...
-            if res:find("404 page not found", 1, true) then
-                util.die("gRPC gateway is not enabled in your etcd cluster, ",
-                         "which is required by Apache APISIX.", "\n")
+            local response_body = {}
+            local headers = {["Content-Length"] = #post_json}
+            if auth_token then
+                headers["Authorization"] = auth_token
+            end
+
+            local res, err = http.request{url = put_url, method = "POST",
```
membphis, line 222: style: I prefer `http.request({...})`, it is easy to read

membphis (approved):

PR #2980, opened 2020-12-07
fix: we should use [1, 100] for percentage instead of [0, 100]

spacewander:
Previously, we got n+1/101 as the percentage.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/fault-injection.lua
+++ b/apisix/plugins/fault-injection.lua
@@ -60,7 +60,7 @@ local function sample_hit(percentage)
         return true
     end
 
-    return random(0, 100) <= percentage
+    return random(1, 100) <= percentage
 end
 
 


tokers (approved):

membphis (approved):

apisix/plugins/fault-injection.lua
```diff
@@ -60,7 +60,7 @@ local function sample_hit(percentage)
         return true
     end
 
-    return random(0, 100) <= percentage
+    return random(1, 100) <= percentage
```
membphis, line 63: do we need to update the `doc`?

spacewander: Current behavior is the correct one according to the doc. Previously, fault-injection will happen even percentage is zero.

membphis: got it

PR #3013, opened 2020-12-10
fix: enable ssl as default, using placeholder certificate

### LINKED ISSUE #3005: feat: enable the SSL feature in DP by default
APISIX version: `2.1`

It not a good idea to disable the SSL feature by default, and it is not easy to enable this feature.
we should enable it by default.

Then we can use the placeholder certificate, and then APISIX can start normally. 

image

Yiyiyimu:

Fix #3005 

We still need to enable SSL as default, thus using Random certificate as the placeholder

### Pre-submission checklist:


conf/config-default.yaml
```diff
@@ -64,8 +64,8 @@ apisix:
   # https_admin: true             # enable HTTPS when use a separate port for Admin API.
                                 # Admin API will use conf/apisix_admin_api.crt and conf/apisix_admin_api.key as certificate.
   admin_api_mtls:               # Depends on `port_admin` and `https_admin`.
-    admin_ssl_cert: ""             # Path of your self-signed server side cert.
-    admin_ssl_cert_key: ""         # Path of your self-signed server side key.
+    admin_ssl_cert: "cert/mtls_PLACE_HOLDER.crt"             # Path of your self-signed server side cert.
```
spacewander, line 67: The certificate for admin can't be a placeholder.

spacewander (changes requested):

conf/config-default.yaml
Yiyiyimu: already removed. thx!

conf/config-default.yaml
```diff
@@ -101,15 +101,14 @@ apisix:
...
-                                  # enable it to use your own cert and key
+    enable: true
     enable_http2: true
     listen_port: 9443
     # ssl_trusted_certificate: /path/to/ca-cert # Specifies a file path with trusted CA certificates in the PEM format
                                                 # used to verify the certificate when APISIX needs to do SSL/TLS handshaking
                                                 # with external services (e.g. etcd)
-    # ssl_cert: /path/to/ssl_cert
-    # ssl_cert_key: /path/to/ssl_cert_key
+    ssl_cert: "cert/ssl_PLACE_HOLDER.crt"
```
membphis, line 110: I think we can remove this option, the user does not need to care about it

conf/config-default.yaml
```diff
@@ -101,15 +101,14 @@ apisix:
...
+    ssl_cert_key: "cert/ssl_PLACE_HOLDER.key"
```
membphis, line 111: ditto

membphis (changes requested):

moonming:
nice

spacewander (approved):

moonming (approved):

conf/config-default.yaml
tokers: It's good to save this, since sometimes users may need to customize the default certs.

tokers (approved):

conf/config-default.yaml
membphis: this default cert is useless. and the APISIX never uses those certs.

Yiyiyimu: Thanks! PTAL

membphis (approved):

PR #3015, opened 2020-12-10
style: local var [http_statuses] redefined in apisix/init.lua

diaosj:

Just an issue about code style. Local variable `http_statuses` was defined at line 686 (https://github.com/apache/apisix/blob/master/apisix/init.lua#L686) and renamed at line (https://github.com/apache/apisix/blob/master/apisix/init.lua#L701). 

### Pre-submission checklist:


membphis (approved):

apisix/init.lua
```diff
@@ -698,7 +698,7 @@ local function healcheck_passive(api_ctx)
         end
     end
 
-    local http_statuses = passive and passive.unhealthy and
+    http_statuses = passive and passive.unhealthy and
                           passive.unhealthy.http_statuses
```
spacewander, line 702: Please update the indentation.

spacewander (changes requested):

apisix/init.lua
```diff
@@ -698,8 +698,8 @@ local function healcheck_passive(api_ctx)
         end
     end
 
-    local http_statuses = passive and passive.unhealthy and
-                          passive.unhealthy.http_statuses
+    http_statuses = passive and passive.unhealthy and
+            passive.unhealthy.http_statuses
```
spacewander, line 702: Err. It would be better if the `passive` can be aligned with the `= passive` in the previous line, just like the state before your change.

spacewander (changes requested):

apisix/init.lua
diaosj: ok. Got it.

spacewander (approved):

PR #3021, opened 2020-12-11
feat: rewrite handwriting validation with jsonschema

### LINKED ISSUE #2967: request help: In some plugins, directly use the code for parameter verification, we need to replace it with JSONSchema verification
## Background

In Apache APISIX Dashboard, we use a JSONSchema synchronized from APISIX for parameter verification.

However, during the development and testing process, we found that there are still some plugins that directly use code for parameter verification, instead of using JSONSchema verification. 

Here is an example:
https://github.com/apache/apisix/blob/master/apisix/plugins/jwt-auth.lua#L109-L124

## Issues with the current solution

1. Not clear enough, many people may not know that there are special codes for parameter verification besides JSONSchema verification, and subsequent maintenance is also troublesome.

2. In the Dashboard, both the FE and `Manager API` need to synchronize these code changes, which becomes very cumbersome, and it is likely to cause problems due to missing or untimely synchronization.

## Suggested changes

1. All parameter verification should be verified with JSONSchema as much as possible, except for those that JSONSchema cannot achieve

2. Newly added plug-ins in the future should also follow this specification.

## Related plugins
- [x] jwt-auth
- [x] ip-restriction
- [x] openid-connect
- [x] redirect
- [ ] response-rewrite

- [ ] request-validation
- [ ] cors
- [ ] proxy-rewrite
- [ ] proxy-cache (need to read `conf/config.yaml` in APISIX to check cache zone )
- [ ] uri-blocker (regular expression in Lua and `ngx.re`)

spacewander:
Fix #2967. Some plugins can't be rewritten, so I have to keep them.

### Pre-submission checklist:


apisix/plugins/jwt-auth.lua
```diff
@@ -42,12 +42,11 @@ local schema = {
 
 local consumer_schema = {
     type = "object",
-    additionalProperties = false,
+    -- can't use additionalProperties with dependencies
```
membphis, line 45: Is it a bug of JSONSchema?

apisix/schema_def.lua
```diff
@@ -43,14 +45,19 @@ local host_def = {
...
+end
+local ipv4_def = table_concat(ipv4_def_buf, [[\.]])
 -- There is false negative for ipv6/cidr. For instance, `:/8` will be valid.
 -- It is fine as the correct regex will be too complex.
 local ipv6_def = "([a-fA-F0-9]{0,4}:){1,8}(:[a-fA-F0-9]{0,4}){0,8}"
                  .. "([a-fA-F0-9]{0,4})?"
 local ip_def = {
     {title = "IPv4", type = "string", format = "ipv4"},
-    {title = "IPv4/CIDR", type = "string", pattern = "^" .. ipv4_def .. "/[0-9]{1,2}$"},
+    {title = "IPv4/CIDR", type = "string", pattern = "^" .. ipv4_def .. "/([12]?[0-9]|3[0-2])$"},
```
membphis, line 60: I am not sure if the current regex is good enough, I have listed the example of IPs, we can confirm:

* 127.0.0.1/1
* 127.0.0.1/10
* 127.0.0.1/11
* 127.0.0.1/20
* 127.0.0.1/21
* 127.0.0.1/30
* 127.0.0.1/32
* 127.0.0.1/33

apisix/plugins/jwt-auth.lua
spacewander: When `additionalProperties  = false`, you can't use `dependencies` to add new field. This is not explicitly shown in the spec, but both Lua and Go implementation have this limitation.

membphis: ok, got it. many thx for your explain

membphis (approved):

apisix/plugins/openid-connect.lua
```diff
@@ -29,16 +29,43 @@ local schema = {
...
+        },
+        ssl_verify = {
+            type = "boolean",
+            default = false,
+        },
+        timeout = {
+            type = "integer",
+            minimum = 1,
+            default = 3,
+            description = "timeout in second",
```
jbampton, line 44:
```suggestion
            description = "timeout in seconds",
```

apisix/plugins/openid-connect.lua
```diff
@@ -148,6 +150,13 @@ end
 
 function _M.rewrite(plugin_conf, ctx)
     local conf = core.table.clone(plugin_conf)
+
+    -- Previously, we multiple conf.timeout before storing it in etcd.
```
jbampton, line 154:
```suggestion
    -- Previously, we multiply conf.timeout before storing it in etcd.
```

apisix/plugins/openid-connect.lua
```diff
@@ -148,6 +150,13 @@ end
...
+    -- If the timeout is too large, we should not multiple it again.
```
jbampton, line 155:
```suggestion
    -- If the timeout is too large, we should not multiply it again.
```

doc/plugins/openid-connect.md
```diff
@@ -41,7 +41,7 @@ The OAuth 2 / Open ID Connect(OIDC) plugin provides authentication and introspec
 | bearer_only                        | boolean | optional    | false                 |         | Setting this `true` will check for the authorization header in the request with a bearer token |
 | logout_path                        | string  | optional    | "/logout"             |         |                                                                                                |
 | redirect_uri                       | string  | optional    | "ngx.var.request_uri" |         |                                                                                                |
-| timeout                            | integer | optional    | 3                     | [1,...] |                                                                                                |
+| timeout                            | integer | optional    | 3                     | [1,...] | Timeout in second                                                                              |
```
jbampton, line 44:
```suggestion
| timeout                            | integer | optional    | 3                     | [1,...] | Timeout in seconds                                                                              |
```

nic-chen (approved):

apisix/plugins/openid-connect.lua
```diff
@@ -148,6 +150,13 @@ end
 
 function _M.rewrite(plugin_conf, ctx)
     local conf = core.table.clone(plugin_conf)
+
+    -- Previously, we multiply conf.timeout before storing it in etcd.
+    -- If the timeout is too large, we should not multiply it again.
+    if not (conf.timeout >= 1000 and conf.timeout % 1000 == 0) then
```
juzhiyuan, line 156: Hi @spacewander 👋 May I know why we need to use `multiply` here? 🤔 because both our document[1] and code comments (`schema.description`) says it's in `second`.

[1] https://apisix.apache.org/docs/apisix/next/plugins/openid-connect

PR #3036, opened 2020-12-14
feat: support ctx.var to get service_name and route_name

### LINKED ISSUE #2982: [DISSCUS]:  support `ctx.var` to get service_name and route_name
### Issue description

When using the http-logger plugin, I need to obtain the name value corresponding to the service (for example: `"service_name:$service_name"`). But it is currently not available.
Is it necessary for us to support `ctx.var` to get the value of `service_name`?
By the way, sometimes `route_name` is also obtained, but it is not currently supported.

### Environment

* apisix version (cmd: `apisix version`): 2.0
* OS:

Firstsawyou:
close #2982

### Pre-submission checklist:


membphis (commented):
nice PR, need to update the doc

t/plugin/http-logger-log-format.t
```diff
@@ -319,7 +319,245 @@ qr/request log: \[\{"\@timestamp":".*","client_ip":"127.0.0.1","host":"127.0.0.1
...
+--- request
+GET /hello
+--- response_body
+hello world
+--- error_log eval
+qr/request log: \{"route_id":"1","route_name":"ctx_var-support-route_name"\}/
+
+
+
+=== TEST 17: remove service
```
spacewander, line 537: Need a test to check when there is not matched service.

spacewander (changes requested):

t/plugin/http-logger-log-format.t
Firstsawyou: The test case of this PR should not be written in `http-logger.t`, I will update the test case of this PR to `ctx.t` later. So this `delete service` operation will be deleted.

Firstsawyou:
> nice PR, need to update the doc

updated.

t/plugin/http-logger-log-format.t
spacewander: What if `service_name` is configured but don't have a matched service?

Firstsawyou: > What if `service_name` is configured but don't have a matched service?

For the `http-logger` plugin, the obtained `service_name` is `nil`, and it will not be displayed in the `http-logger` log.

spacewander: I think we need to add test case for it.

Firstsawyou: agree, updated.

https://github.com/apache/apisix/pull/3036/files#diff-fbf74d528978b104d8618c1cee462ad94674894666cddee60f41318c687d290dR761-R767

spacewander (approved):

membphis (approved):

PR #3088, opened 2020-12-21
feat: add control api for plugin server-info

tokers:

Previously we added plugin server-info without exposing API since the same port problem of admin api and proxy. Now we have Control API support so we expose server-info API through Control API. People now can insight the server info about APISIX instance easily.

### Pre-submission checklist:


apisix/plugins/server-info.lua
```diff
@@ -99,6 +99,17 @@ local function get()
 end
 
 
+local function get_server_info()
+    local info, err = get()
```
spacewander, line 103: Look like we need to way to store the `boot_time` in shdict when `etcd` is not used. Otherwise the `boot_time` will change when plugin reloads.

tokers: That's really, i will change it.

doc/control-api.md
```diff
@@ -49,3 +49,9 @@ Here is the supported API:
 Introduced since `v2.2`.
 
 Return the jsonschema used by this APISIX instance.
+
+### GET /v1/server_info
```
spacewander, line 53: I think it would be better to put this to plugin's own section. So that the API in this page is builtin and will be provided whether you use the plugin or not.

t/plugin/server-info.t
```diff
@@ -133,3 +133,54 @@ integral
...
 --- error_log
 timer created to report server info, interval: 60
+
+
+
+=== TEST 3: get server_info from plugin control API
+--- yaml_config
+apisix:
+    id: 123456
+    enable_control: true
```
spacewander, line 143: This configuration may confuses beginner, because the control server in the test is hardcoded in APISIX.pm, instead of generating from config.yaml.

spacewander (changes requested):

t/plugin/server-info.t
tokers: Fixed.

doc/control-api.md
tokers: Fixed.

apisix/plugins/server-info.lua
tokers: Fixed.

spacewander (approved):

t/plugin/server-info.t
```diff
@@ -133,3 +133,51 @@ integral
...
+=== TEST 3: get server_info from plugin control API
+--- yaml_config
+apisix:
+    id: 123456
+plugins:
+    - server-info
+--- config
+location /t {
+    content_by_lua_block {
+        local json_decode = require("apisix.core").json.decode
```
membphis, line 148: https://github.com/api7/test-toolkit

we can use this `json` to print the sorted data. we can do this in a new PR

membphis (approved):

PR #3092, opened 2020-12-21
feat: expose schema via control API

### LINKED ISSUE #3022: feat: support to export the JSON schema of APISIX via `CLI`
discuss mail:
https://lists.apache.org/thread.html/rb1c19fe9b6e51ede3b5aead122e2bcd4de6249f42b971eabfb8a79ac%40%3Cdev.apisix.apache.org%3E

Because the dashboard project requires schema verification for plug-ins,
upstreams, etc., and the definition of the schema is written in the Lua
code of Apache APISIX. Therefore, we need a tool to generate a JSON schema
file based on the Lua code of Apache APISIX so that it can be used by the
Go and front-end code of the dashboard project.

If the developer want to sync the JSON schema, he/she should follow this step:

1. export JSON schema file: call `apisix export-jsonschema`, we will get the file `schema.json`
2. copy the file `schema.json` from apisix to the dashboard folder
3. start the dashboard

those steps are much safer. and the dashboard does not need any update.

spacewander:
Close #3022

### Pre-submission checklist:

### DIFF
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -24,7 +24,10 @@ local _M = {}
 function _M.schema()
     local schema = {
         main = {
+            consumer = core.schema.consumer,
             global_rule = core.schema.global_rule,
+            plugins = core.schema.plugins,
+            proto = core.schema.proto,
             route = core.schema.route,
             service = core.schema.service,
             ssl = core.schema.ssl,
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -14,11 +14,33 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+local core = require("apisix.core")
+local plugin = require("apisix.plugin")
+
+
 local _M = {}
 
 
 function _M.schema()
-    return 200, {}
+    local schema = {
+        main = {
+            global_rule = core.schema.global_rule,
+            route = core.schema.route,
+            service = core.schema.service,
+            ssl = core.schema.ssl,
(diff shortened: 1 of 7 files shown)


apisix/control/v1.lua
```diff
@@ -14,11 +14,33 @@
...
-    return 200, {}
+    local schema = {
+        main = {
+            global_rule = core.schema.global_rule,
+            route = core.schema.route,
+            service = core.schema.service,
+            ssl = core.schema.ssl,
+            stream_route = core.schema.stream_route,
+            upstream = core.schema.upstream,
+        },
```
nic-chen, line 33: how about could use `core.schema` as main ?

There will be redundant data, but no data will not be missed, and there is no need to modify the code here after adding a new schema.

spacewander: There is some methods in `core.schema` can't be json encoded, like `core.schema.check`.

nic-chen: ok

nic-chen (approved):

apisix/control/v1.lua
nic-chen: lack of consumer

nic-chen (changes requested):

apisix/control/v1.lua
spacewander: @nic-chen 
Solved.

nic-chen (approved):

membphis (approved):

Issue #2532, opened 2020-10-27
feat:   Apisix upstream domain resolver can not support "search  xxxx, options  xxxxx" in k8s environment


gy09535:
### Issue description
for example  in k8s environment , we can use  service name as backend nodes, sometime  for core DNS query we can combine the service name  and "svc.cluster.local" as the domain, the  resolve config is 
```
nameserver 10.96.0.10
search svc.cluster.local cluster.local
options ndots:5
```
apisix just support "nameserver",ref code:  https://github.com/apache/apisix/blob/master/apisix/core/utils.lua#L75,so we must define the full domain "xxxx.svc.cluster.local", But  we can not us other suffix such as "cluster.local". Should we support the full dns resolve rule?
### Environment

* apisix version (cmd: `apisix version`):
* OS:

### Minimal test code / Steps to reproduce the issue
1.
2.
3.

membphis:
> Should we support the full dns resolve rule?

do you know how it works? I am not familiar with this.

gy09535:
I know some from k8s core dns resolver, Now I read this page:https://linux.die.net/man/5/resolv.conf , full support is difficult, But support search and options can work in k8s cluster.

membphis:
welcome PR for this feature.

if you need any help, please let us know.

gy09535:
> welcome PR for this feature.
> 
> if you need any help, please let us know.

Sure, thanks.

dickens7:
To solve this problem, you can refer to or use this library directly

https://github.com/Kong/lua-resty-dns-client

@gy09535

gy09535:
> To solve this problem, you can refer to or use this library directly
> 
> https://github.com/Kong/lua-resty-dns-client
> 
> @gy09535

See it, It it great.

ShaoZeMing:
> > 要解决此问题，您可以直接参考或使用此库
> > https://github.com/Kong/lua-resty-dns-client
> > @ gy09535
> 
> 看到它，太好了 💯

+1

ShaoZeMing:
> > 要解决此问题，您可以直接参考或使用此库
> > https://github.com/Kong/lua-resty-dns-client
> > @ gy09535
> 
> 看到它，太好了 💯

hello 我们目前特别迫切需要这个功能，想知道这个功能开发情况如何了？

gy09535:
> > > 要解决此问题，您可以直接参考或使用此库
> > > https://github.com/Kong/lua-resty-dns-client
[...]

PR #3129, opened 2020-12-25
feat:  support  linux  "resolv.conf" file  resolve process

gy09535:

fix:https://github.com/apache/apisix/issues/2532

### Pre-submission checklist:


tokers:
Good feature!

apisix/core/utils.lua
```diff
@@ -112,13 +104,17 @@ local function dns_parse(domain, resolvers)
         return nil, "unsupport DNS answer"
     end
 
-    return dns_parse(answer.cname, resolvers)
+    return dns_parse(answer.cname)
 end
 _M.dns_parse = dns_parse
 
+function _M.init_dns_proxy (options)
```
tokers, line 111: style: redundant space before `(`.

apisix/init.lua
```diff
@@ -43,15 +43,18 @@ if ngx.config.subsystem == "http" then
...
 
 
-local function parse_args(args)
-    dns_resolver = args and args["dns_resolver"]
-    core.utils.set_resolver(dns_resolver)
-    core.log.info("dns resolver", core.json.delay_encode(dns_resolver, true))
+local function parse_args()
+    local options = {}
+    core.utils.init_dns_proxy(options)
+    if options.nameservers then
```
tokers, line 53: It's so strange here, you defined an empty `options` at line 51, in the meanwhile you check its contents here after calling `core.utils.init_dns_proxy`, it doesn't do any initializations for `options`.

You should fill the `options` with the config in `config.yaml`, like `dns_resolver`.

apisix/init.lua
```diff
@@ -69,7 +72,7 @@ function _M.http_init(args)
                              "maxrecord=8000", "sizemcode=64",
                              "maxmcode=4000", "maxirconst=1000")
 
-    parse_args(args)
```
tokers, line 72: We must keep the compatibility here, otherwise people use this way to initialize DNS resolvers will be surprised after they upgrade APISIX.

apisix/init.lua
```diff
@@ -201,7 +204,6 @@ local function parse_domain(host)
     end
 
     core.log.info("parse addr: ", core.json.delay_encode(ip_info))
-    core.log.info("resolver: ", core.json.delay_encode(dns_resolver))
```
tokers, line 204: Could we still output it?

apisix/init.lua
gy09535: I am not recognize that , their is a config  in yaml, thanks for your remind it

gy09535: fixed.

apisix/init.lua
gy09535: fixed.

apisix/init.lua
gy09535: fixed

apisix/core/utils.lua
gy09535: fixed.

membphis (approved):

t/core/utils.t
```diff
@@ -239,3 +236,32 @@ res:tell David to
...
+            local core = require("apisix.core")
+            local resolvers = {"223.5.5.5", "223.6.6.6"}
+            local search = {"google.com"}
+            local ndots = 5
+            core.utils.init_dns_proxy({
+                ndots = ndots,
+                search = search,
+                resolvers = resolvers,
+            })
+            local ip_info, err = core.utils.dns_parse("news.google.com")
```
tokers, line 255: Try to resolve domain like `news`, since we have the custom `search` option, `news.google.com` will be resolved.

Also, it's not good to depend on external suite, it might cause the CI unstable.

tokers (changes requested):

t/core/utils.t
gy09535: > Try to resolve domain like `news`, since we have the custom `search` option, `news.google.com` will be resolved.
> 
> Also, it's not good to depend on external suite, it might cause the CI unstable.

It should the dns serve support, such as core dns, we can not mock it. We can change it to github.com to avoid the CI unstable.

gy09535: I try to use test with https://github.com/Kong/lua-resty-dns-client/blob/master/spec/client_spec.lua#L219 , but the method "_search_iter"  is not export.So I use the special domain to do the test.

tokers: Then how to proof the code can be used correctly in k8s pod? Say we need to resolve the etcd service with its short name like "myetcd", while actually the record in dns is "myetcd.ns.svc.cluster.local" as far as I know.

gy09535: > Then how to proof the code can be used correctly in k8s pod? Say we need to resolve the etcd service with its short name like "myetcd", while actually the record in dns is "myetcd.ns.svc.cluster.local" as far as I know.

It is my fault, It can work now.

gy09535:
So stranger why I got kafka and other errors.

apisix/core/utils.lua
```diff
@@ -80,26 +82,16 @@ function _M.split_uri(uri)
...
-        return nil, "failed to query the DNS server: " .. err
-    end
+    local answers, err
+    answers, err = dns_proxy.client.resolve(domain)
 
-    if answers.errcode then
-        return nil, "server returned error code: " .. answers.errcode
-                    .. ": " .. answers.errstr
+    if not answers then
+      return nil, "failed to query the DNS server: " .. err
```
tokers, line 94: Style: bad indent, should keep 4 spaces.

tokers:
Only a style problem, others are LGTM :).

tokers (approved):

apisix/core/utils.lua
```diff
@@ -80,28 +82,18 @@ function _M.split_uri(uri)
 end
 
 
-local function dns_parse(domain, resolvers)
-    resolvers = resolvers or _M.resolvers
-    local r, err = resolver:new{
-        nameservers = table.clone(resolvers),
-        retrans = 5,  -- 5 retransmissions on receive timeout
-        timeout = 2000,  -- 2 sec
```
spacewander, line 88: We should keep the options for the new lua-resty-dns-client way. It would be better if user can configure them.

rockspec/apisix-master-0.rockspec
```diff
@@ -58,6 +58,7 @@ dependencies = {
     "resty-redis-cluster = 1.02-4",
     "lua-resty-expr = 1.0.0",
     "graphql = 0.0.2",
+    "lua-resty-dns-client = 5.1.1-2",
```
spacewander, line 61: I think it would be safer if we keep two dns client libraries, and use lua-resty-dns-client by default but provide an option to switch to the old one.

spacewander (changes requested):

apisix/core/utils.lua
gy09535: Added it in default value for keep same with old logical.

spacewander:
It seems the new DNS client is not 100% compatibility with the old one, and causes the test to failed.

gy09535:
> It seems the new DNS client is not 100% compatibility with the old one, and causes the test to failed.

It can work fine in my side.

gy09535:
> It seems the new DNS client is not 100% compatibility with the old one, and causes the test to failed.

After I merge the master to this branch, it can not pass in CI, but can work fine in my side.

ShaoZeMing:
[抱大腿]

apisix/core/utils.lua
```diff
@@ -112,13 +104,13 @@ local function dns_parse(domain, resolvers)
...
     end
 
-    return dns_parse(answer.cname, resolvers)
+    return dns_parse(answer.cname)
 end
 _M.dns_parse = dns_parse
 
-
-function _M.set_resolver(resolvers)
-    _M.resolvers = resolvers
```
spacewander, line 121: To maintain the backward compatibility, we still need to set `_M.resolvers`. The API in `core` is expected unchanged.

spacewander (changes requested):

apisix/core/utils.lua
gy09535: I add this method and to do init logical to keep when set resolver the dns server should init again.

gy09535: After check their is no place use this method, so we can remove it.

apisix/init.lua
```diff
@@ -50,7 +50,12 @@ local ver_header    = "APISIX/" .. core.version.VERSION
 
 local function parse_args(args)
     dns_resolver = args and args["dns_resolver"]
-    core.utils.set_resolver(dns_resolver)
+    local options = {
+        nameservers = dns_resolver,
+        timeout = 5000,
```
spacewander, line 55: Why change timeout to 5000?

gy09535: Just for check test case, changed.

spacewander:
The test is quite unstable. Maybe we can put `lua-resty-dns-client` as an option and use the old one by default?

gy09535:
> The test is quite unstable. Maybe we can put `lua-resty-dns-client` as an option and use the old one by default?

Strange things is  when I remove  resolvers in start it can pass.

apisix/init.lua
```diff
@@ -52,7 +52,7 @@ local function parse_args(args)
     dns_resolver = args and args["dns_resolver"]
     local options = {
         nameservers = dns_resolver,
-        timeout = 5000,
+        timeout = 2000,
```
membphis, line 55: why do we need to change this?

apisix/cli/ops.lua
```diff
@@ -105,6 +105,7 @@ local function local_dns_resolver(file_path)
     for line in file:lines() do
         local addr, n = line:gsub("^nameserver%s+(%d+%.%d+%.%d+%.%d+)%s*$", "%1")
         if n == 1 then
+            io.stderr:write("DNS: ",addr,"\n")
```
membphis: why do we use `io.stderr` here? is it an error case?

gy09535: already removed 😁

membphis:
@ gy09535 CI failed

gy09535:
> @ gy09535 CI failed

I am consider to refactor it, make the dns resolve as an backend resolve.

PR #3139, opened 2020-12-27
feat: support var in upstream_uri on proxy-rewrite plugin

### LINKED ISSUE #3133: Combining basic-auth & proxy-rewrite
I am trying to use two plugins to create a sort of password-protected area in my application. I was hoping to use the basic-auth plugin for username/password functionality and proxy-rewrite to automatically add the username to the route. For example, dynamically rerouting requests to `host:3000` to `host:3000/{username}`.

Is this something that falls within the functionalities APISIX aims to cover? Is this something I can achieve with a custom plugin?

BFergerson:
Adds the ability to use `ctx.var` in `upstream_uri` on the proxy-rewrite plugin.
Fixes #3133


Firstsawyou:
Need to add semantic "PR" title, for example: `feat: ctx.var in upstream_uri on proxy-rewrite plugin`

Firstsawyou:
We need test cases to cover this feature.

spacewander:
Please merge master to make CI pass.

spacewander:
You can write test like
https://github.com/apache/apisix/blob/45e909a0b122b8cda5a75e95adcfaf62378f7a30/t/plugin/proxy-rewrite.t#L477-L526

t/plugin/proxy-rewrite.t
```diff
@@ -1249,3 +1249,41 @@ version: nginx_var_does_not_exist
 x-real-ip: 127.0.0.1
 --- no_error_log
 [error]
+
+
+
+=== TEST 42: set route(rewrite uri based on ctx.var)
```
membphis, line 1255: only set route is not enough. we need to make a test

membphis (changes requested):

t/plugin/proxy-rewrite.t
BFergerson: Unfortunately, I think this is the end of my understanding of how to contribute to APISIX. I never heard of Lua before this project so I'm relying on the documentation and contributors of this project to help me understand what to do.

Can you point to other tests (or documentation) which does everything I need to do in my test?

membphis: I write a full test case example, you can make a try.

^_^

```
=== TEST 42: set route(rewrite uri based on ctx.var)
--- config
    location /t {
        content_by_lua_block {
            local code, body = t('/apisix/admin/routes/1',
                 ngx.HTTP_PUT,
                 [[{
                        "plugins": {
                            "proxy-rewrite": {
                                "uri": "/$arg_new_uri"
                            }
                        },
                        "upstream": {
                            "nodes": {
                                "127.0.0.1:1980": 1
                            },
                            "type": "roundrobin"
                        },
                        "uri": "/test"
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 43: hit route(upstream uri: should be /hello)
--- request
GET /test?new_uri=hello
--- response_body
hello world
--- no_error_log
[error]

```

BFergerson: Thanks for the help. I'm still not sure how your tests cover anything more than what my test covers though.

Based on your tests it appears tests are executed sequentially and share a context between them. The first test creates a route with the `proxy-rewrite` plugin rewriting the uri to `$arg_new_uri`. This appears to be the same thing my test covers with just a different variable to subsitute.

Your second test then appears to continue work from the first test and now hits `/test?new_uri=hello` and expects `hello world` as a response. That doesn't make sense to me. Where does `hello world` come from and how does hitting `/test?new_uri=hello` verify that `proxy-rewrite` has worked correctly? There should be some way to set the value of `$arg_new_uri` and verify the resulting value of uri after `proxy-rewrite` is equiavlent. Also, if `$arg_new_uri` is really the value of `new_uri` that is a bit confusing. I wouldn't have guessed adding `arg_` makes a variable come from url get params.

Apologies if I'm making this more difficult than it needs to be. I've tried reading the [documention](https://github.com/apache/apisix/blob/master/doc/plugin-develop.md#write-test-case) but it doesn't really explain to me how testing works.

membphis: 'arg_****': argument name in the request line

nginx variable link: http://nginx.org/en/docs/http/ngx_http_core_module.html#var_arg_

`$arg_new_uri`, it means to fetch the value from query string which name is `new_uri`.

1.  the current request is `/test?new_uri=hello`, so the `$arg_new_uri` should equal `hello`
2. `"uri": "/$arg_new_uri"` -> `"uri": "/hello"`
3.  response body is 'hello world' if the request is `/hello`: https://github.com/apache/apisix/blob/master/t/lib/server.lua#L36
4. check the response body 

```
--- response_body 
hello world
```

BFergerson: Apologies for the delay. Thanks for the help with understanding how tests work.

I've added some information to the documentation which I think would have been enough for me to know how to setup this test myself. Let me know if it's acceptable.

doc/plugin-develop.md
```diff
@@ -285,6 +285,10 @@ When we request __/t__, which config in the configuration file, the Nginx will c
 "__no_error_log__" means to check the "__error.log__" of Nginx. There must be no ERROR level record. The log files for the unit test
 are located in the following folder: 't/servroot/logs'.
 
+The above test case represents a simple scenario. Most scenarios will require multiple tests to validate. To do this, create multiple tests `=== TEST 1`, `=== TEST 2`, and so on. These tests will be executed sequentially, allowing you to break down scenarios into a sequence of atomic steps.
```
spacewander, line 288: "require multiple tests" should be "require multiple steps"

BFergerson, line 288:
```suggestion
The above test case represents a simple scenario. Most scenarios will require multiple steps to validate. To do this, create multiple tests `=== TEST 1`, `=== TEST 2`, and so on. These tests will be executed sequentially, allowing you to break down scenarios into a sequence of atomic steps.
```

spacewander (approved):

tokers (approved):

membphis (approved):

moonming:
ping @membphis

PR #3140, opened 2020-12-27
test: limit-count add more test cases to cover the case where the upstream is a domain name

### LINKED ISSUE #3108: bug: limit-count plugin does not take effect
### Issue description

When the `limit-count` plugin is enabled on `service` and `upstream` is the domain name. This will cause the `local key = (ctx.var[conf.key] or "") .. ctx.conf_type .. ctx.conf_version` generated in the limit-count plugin to be inconsistent. When `ctx.conf_version` is deleted (`local key = (ctx.var[conf.key] or "") .. ctx.conf_type`), the plug-in function returns to normal.

Steps to reproduce, will be added later.

### Environment

* apisix version (cmd: `apisix version`): 2.1
* OS: (cmd: `uname -a`)：centos7
* OpenResty / Nginx version: (cmd: `nginx -V` or `openresty -V`): openresty/1.19.3.1

1.
2.
3.

Firstsawyou:

close #3108

 `limit-count` plugin add more test cases to cover the case where the upstream is a domain name.

### Pre-submission checklist:

### DIFF
--- a/t/plugin/limit-count.t
+++ b/t/plugin/limit-count.t
@@ -1178,7 +1178,7 @@ passed
 
 
 
-=== TEST 37: plugin is bound to the service and upstream is the domain name
+=== TEST 37: add service and route, upstream is the domain name
 --- config
     location /t {
         content_by_lua_block {
@@ -1205,25 +1205,10 @@ passed
 
             if code >= 300 then
                 ngx.status = code
+                ngx.say(body)
             end
-            ngx.say(body)
-        }
-    }
---- request
-GET /t
---- response_body
-passed
---- no_error_log
-[error]
 
-
-
-=== TEST 38: add route and bind service_id
---- config
-    location /t {
-        content_by_lua_block {
-            local t = require("lib.test_admin").test
-            local code, body = t('/apisix/admin/routes/1',
+            code, body = t('/apisix/admin/routes/1',
                  ngx.HTTP_PUT,
                  [[{
                     "methods": ["GET"],
@@ -1247,7 +1232,7 @@ passed
 
 
 
-=== TEST 39: normal, the result is as expected
+=== TEST 38: normal, the result is as expected
 --- init_by_lua_block
     require "resty.core"
     apisix = require("apisix")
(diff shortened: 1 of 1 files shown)


t/plugin/limit-count.t
```diff
@@ -1175,3 +1175,95 @@ passed
...
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 38: add route and bind service
```
spacewander, line 1220: I think we should merge TEST 38 into TEST 37, so that it is clear that they are both for TEST 39.

Firstsawyou: updated.

t/plugin/limit-count.t
```diff
@@ -1175,3 +1175,73 @@ passed
 [200, 200, 503, 503]
 --- no_error_log
 [error]
+
+
+
+=== TEST 37: add route and upstream is the domain name
```
spacewander, line 1181: The #3108 is limit-count on the `service`, not a route?

spacewander (changes requested):

t/plugin/limit-count.t
Firstsawyou: The main test is the case where the upstream is a domain name, which is also feasible.
But I am adding some tests that limit the service.

t/plugin/limit-count.t
```diff
@@ -1175,3 +1175,170 @@ passed
...
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 38: add route and upstream is the domain name
```
spacewander, line 1214: TEST 38 should be merged into TEST 37.

spacewander (changes requested):

t/plugin/limit-count.t
Firstsawyou: updated.

t/plugin/limit-count.t
```diff
@@ -1178,7 +1178,7 @@ passed
 
 
 
-=== TEST 37: plugin is bound to the service
+=== TEST 37: plugin is bound to the service and upstream is the domain name
```
spacewander, line 1181: I think we can also add route and bind service_id in this one.

Firstsawyou: updated.

spacewander (approved):

gxthrj (approved):

PR #3172, opened 2021-01-01
docs: update the title of ci

### LINKED ISSUE #3167: ci: need a better title
here is the current list, it is not easy to read.

image

dabue:
update the title of ci

fix #3167 
### Pre-submission checklist:


.github/workflows/test-e2e.yml
```diff
@@ -7,7 +7,7 @@ on:
     branches: [master]
 
 jobs:
-  build:
+  e2e:
```
spacewander, line 10: Since our tests are not only e2e test, would it be better to use name "ci/test"?

dabue: agree +1

spacewander (approved):

spacewander:
I found a way to let the CI run: keep the original files and then we can submit a new one to remove them after this one is merged.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,101 @@
+name: ci
```
spacewander, line 1: Also need to change the badge and its Chinese version:
https://github.com/apache/apisix/blob/151d74fd9c835d9d1dc5cc7b25fb9bdbf9fc56cf/README.md#L22

dabue:
> d then we can submit a new one to remove t

I will have a try , thanks

membphis (approved):

PR #3204, opened 2021-01-06
feat: support more customize configure in nginx.conf

shenshaoye:
在使用apisix过程中由于要使用第三方模块，需要用到load_module和include这两个指令，目前关于nginx的自定义配置对这两个指令的支持有问题，因此这个pull request丰富了用户自定义配置的支持
具体见https://github.com/apache/apisix/issues/3203

### DIFF
--- a/.travis/apisix_cli_test/test_main.sh
+++ b/.travis/apisix_cli_test/test_main.sh
@@ -675,6 +675,8 @@ nginx_config:
         set $my "var";
     http_admin_configuration_snippet: |
         log_format admin "$request_time $pipe";
+    http_end_configuration_snippet: |
+        server_names_hash_bucket_size 128;
     stream_configuration_snippet: |
         tcp_nodelay off;
 ' > conf/config.yaml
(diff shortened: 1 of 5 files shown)


apisix/cli/ngx_tpl.lua
```diff
@@ -26,6 +26,12 @@ worker_processes {* worker_processes *};
 worker_cpu_affinity auto;
 {% end %}
 
+# main start configuration snippet starts
```
spacewander, line 29: I think we can directly move `main_configuration_snippet` here instead of creating new one?

.travis/apisix_cli_test/test_main.sh
```diff
@@ -675,6 +677,8 @@ nginx_config:
         set $my "var";
     http_admin_configuration_snippet: |
         log_format admin "$request_time $pipe";
+    http_end_configuration_snippet: |
```
spacewander, line 680: Need to check `http_end_configuration_snippet` like other snippet in the test.

spacewander (changes requested):

.travis/apisix_cli_test/test_main.sh
```diff
@@ -705,6 +707,12 @@ if [ ! $? -eq 0 ]; then
     exit 1
 fi
 
+grep 'server_names_hash_bucket_size 128;' -A 2 conf/nginx.conf | grep "configuration snippet ends" > /dev/null
```
spacewander, line 710: Would be better if we also check there is a '}' in the output, like:
grep 'server_names_hash_bucket_size 128;' -A 3 conf/nginx.conf | grep "}" > /dev/null

spacewander (approved):

tokers:
@shenshaoye Should add some docs about the new adding fields.

tokers (approved):

PR #3213, opened 2021-01-07
chore: catch error while new `healthcheck` instance

### LINKED ISSUE #3169: chore: cache the error message when create new `healthcheck` instance
`healthcheck.new` may fail, we need to check the return value first.

https://github.com/apache/apisix/blob/4d316bc0681c80435ab1a3fc113c3c6ea3c82831/apisix/balancer.lua#L90

here is the doc:

image

dabue:
catch error while new `healthcheck` instance and return the checker with nil.

the function pick_server will treat the nodes as normal  with the 'checker == nil', and get a correct one by rules.

fix #3169 
### Pre-submission checklist:

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -87,11 +87,16 @@ local function create_checker(upstream, healthcheck_parent)
     if healthcheck == nil then
         healthcheck = require("resty.healthcheck")
     end
-    local checker = healthcheck.new({
+    local checker, err = healthcheck.new({
         name = "upstream#" .. healthcheck_parent.key,
         shm_name = "upstream-healthcheck",
         checks = upstream.checks,
     })
+
+    if not checker then
+        core.log.error("fail to create healthcheck instance: ", err)
+        return
+    end
 
     local host = upstream.checks and upstream.checks.active and upstream.checks.active.host
     local port = upstream.checks and upstream.checks.active and upstream.checks.active.port


apisix/balancer.lua
```diff
@@ -87,12 +87,17 @@ local function create_checker(upstream, healthcheck_parent)
     if healthcheck == nil then
         healthcheck = require("resty.healthcheck")
     end
-    local checker = healthcheck.new({
+    local checker, err = healthcheck.new({
```
spacewander, line 90: Is it possible to write a test for it?

dabue: I remember that error happened when more workers tried to create checkers at the same moment. In the PROD, I have encountered such problems. I will consider how to add test cases， but I think it's a bit difficult :(

spacewander: OK

spacewander (approved):

apisix/balancer.lua
dabue: I can't find a way to reproduce it by using Test::Nginx.  In my test case, I've set master_on,workers(10) and added a server on different port, I create many threads to connect that server when got the initial request, and wanted the different workers would handled that connections, but  only the same worker  which got  the initial request dealt with the connection requests.

@membphis  @spacewander  Do you have a way to get test requests on different workers?

spacewander: AFAIK, there is no such a way.

membphis (approved):

PR #3232, opened 2021-01-09
apisix should check config file when apisix restart or apisix reload 

Yangzcy423911:

apisix configfile error service down

nginx -t check nginx config 

apisix check config   aviod restart or reload fail


tokers:
@Yangzcy423911 It seems your base branch is too old, please rebase it with the latest master branch.

membphis:
@Yangzcy423911  you can create a new PR with your commit.

your commit log is wrong now.

image

PR #3250, opened 2021-01-12
feat(CI): check for trailing whitespace

jbampton:

Standardizes the code by checking for trailing whitespace.

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,3 +31,10 @@ jobs:
         run: |
           # return non-zero exit code on warnings
           yamllint --strict .
+  trailing-whitespace:
+    name: 🌌 Trailing whitespace
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: 🧹 Check for trailing whitespace
+        run: "! git grep -EIn $'[ \t]+$' | grep -v 'received: '"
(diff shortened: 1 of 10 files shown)


t/node/remote-addr-ipv6.t
```diff
@@ -116,7 +116,7 @@ received: Content-Type: text/plain
 received: Content-Length: 12
 received: Connection: close
 received: Server: APISIX/\d\.\d+(\.\d+)?
-received:
```
spacewander, line 119: The space is required as we use `received: ` as the prefix.

utils/check-for-trailing-whitespace.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+cd .. || exit
+# print first
+grep -EHInr '( +)$' ./* | grep -v 'received: '
```
spacewander, line 22: It would be better to change the test code to avoid magic word `received: `.
Or is it acceptable to skip the check in `t/` file? @membphis

membphis: > It would be better to change the test code to avoid the magic word `received: `.

agree with this. it seems that we have to update the test cases first.

we can create new issues about these to-do things, fix them later after this PR merged.

membphis (approved):

tokers (approved):

PR #3263, opened 2021-01-12
feat(authz-keycloak): Add Keycloak Authorization Services endpoint discovery.

jenskeiner:
The `authz-keycloak` plugin needs to know the Keycloak token endpoint to use. A configuration option to point to the token endpoint is already there.

Additionally, Keycloak supports discovery of Authorization Services-related endpoints, like the token endpoint, through a discovery document served under a well-known path relative to the realm, e.g. `https://keycloak-host/auth/realms/foo/.well-known/uma2-configuration`.

It can be convenient to use the discovery document to get the token endpoint URL, instead of specifying the token endpoint explicitly, since the URL is typically shorter. More importantly, future build out of the `authz-keycloak` may require access to additional related endpoints. In this case, just providing the single path to the discovery document would be more convenient than specifying each endpoint URL explicitly.

This PR adds an attribute that allows to specify the discovery document URL and makes the token endpoint attribute optional. However, it is checked that at least one, discovery or token endpoint, is given.

The plugin now fetches the discovery document lazily and caches it in shared storage already used by the `openid-connect` plugin (better: the `openidc` module used therein) for similar purposes in the context of OIDC endpoint discovery.

If both, discovery and token endpoint are given, the plugin prefers the explicit token endpoint value over the one from discovery.

### Pre-submission checklist:


jenskeiner:
@sshniro: Penny for your in your thoughts. I have a POC in a separate branch that dynamically resolves URIs to resources defined in Keycloak, so this is just a first change in that direction.

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +66,164 @@ local _M = {
...
+  -- Check the response from the OP.
+  if response.status ~= 200 then
+    err = "response indicates failure, status=" .. response.status .. ", body=" .. response.body
+  else
+    if ignore_body_on_success then
+      return nil, nil
+    end
+
+    -- Decode the response and extract the JSON object.
+    res = cjson_s.decode(response.body)
```
spacewander, line 139: Better to use `core.json.decode`.

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +66,164 @@ local _M = {
...
+
+    if not res then
+      err = "JSON decoding failed"
```
spacewander, line 142: Better to concat the original decode err returned by `decode`

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +66,164 @@ local _M = {
...
+      keepalive = (keepalive ~= "no")
+    }))
+    if not res then
+      err = "accessing discovery url (" .. url .. ") failed: " .. error
+      log.error(err)
+    else
+      log.debug("response data: " .. res.body)
+      json, err = authz_keycloak_parse_json_response(res)
+      if json then
+        authz_keycloak_cache_set("discovery", url, cjson.encode(json), exptime or 24 * 60 * 60)
```
spacewander, line 180: Better to use `core.json.encode`.

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +66,164 @@ local _M = {
...
+      else
+        err = "could not decode JSON from Discovery data" .. (err and (": " .. err) or '')
+        log.error(err)
+      end
+    end
+
+  else
+    json = cjson.decode(v)
```
spacewander, line 188: ditto

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +66,164 @@ local _M = {
...
+  end
+  return value
+end
+
+
+-- Set value in server-wide cache, if available.
+local function authz_keycloak_cache_set(type, key, value, exp)
+  local dict = ngx.shared[type]
+  if dict and (exp > 0) then
+    local success, err, forcible = dict:set(key, value, exp)
```
spacewander, line 94: Better to `log.error` if set dict failed.

apisix/plugins/authz-keycloak.lua
```diff
@@ -52,8 +55,7 @@ local schema = {
         keepalive_timeout = {type = "integer", minimum = 1000, default = 60000},
         keepalive_pool = {type = "integer", minimum = 1, default = 5},
         ssl_verify = {type = "boolean", default = true},
-    },
-    required = {"token_endpoint"}
```
spacewander, line 56: We can do it in jsonschema: https://github.com/apache/apisix/blob/3e19b06293463f52c8d26f0958542fc475ff0fcd/apisix/plugins/echo.lua#L48-L51
And need to a test for it.

spacewander (changes requested):

apisix/plugins/authz-keycloak.lua
```diff
@@ -17,15 +17,18 @@
 local core      = require("apisix.core")
 local http      = require "resty.http"
 local sub_str   = string.sub
-local url       = require "net.url"
-local tostring  = tostring
+local type      = type
 local ngx       = ngx
+local cjson     = require("cjson")
+local cjson_s   = require("cjson.safe")
```
Firstsawyou, line 23: It is better to use the public methods provided here.[https://github.com/apache/apisix/blob/master/apisix/core/json.lua ](https://github.com/apache/apisix/blob/master/apisix/core/json.lua)

Firstsawyou (approved):

apisix/plugins/authz-keycloak.lua
jenskeiner: Agreed. Will fix.

apisix/plugins/authz-keycloak.lua
jenskeiner: Ok.

apisix/plugins/authz-keycloak.lua
jenskeiner: Yep.

apisix/plugins/authz-keycloak.lua
jenskeiner: Right, I'm not yet a JSON schema pro.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

spacewander: @jenskeiner 
Sorry for my mistake! It should be `oneOf` instead of `anyOf`. 😢

apisix/plugins/authz-keycloak.lua
jenskeiner: Yep, using that one now. We're incurring a slight performance penalty by now using `cjson.safe.decode` through `apisix.core.json.decode` to decode values from the shared memory which are known to have been encoded correctly. But not sure how much of a hit that actually is. Just this line here: https://github.com/jenskeiner/apisix/blob/26277f1ea98366ac852ee6673667e1be1320001c/apisix/plugins/authz-keycloak.lua#L190

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixed.

apisix/plugins/authz-keycloak.lua
jenskeiner: Right, true. Will change.

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +68,165 @@ local _M = {
...
+  return http_request_decorator and http_request_decorator(req) or req
+end
+
+
+-- Get the Discovery metadata from the specified URL.
+local function authz_keycloak_discover(url, ssl_verify, keepalive, timeout,
+                                       exptime, proxy_opts, http_request_decorator)
+  log.debug("authz_keycloak_discover: URL is: " .. url)
+
+  local json, err
```
spacewander, line 163: Better to use different name than `json`, since we already have `json` defined as a module.

jenskeiner: Ok. I'm using e.g. `core.json.decode` now, then `json` can be used as a variable name w/o any confusion.

apisix/plugins/authz-keycloak.lua
jenskeiner: Hmm, I think we should actually use `anyOf` since the explicit endpoint URL will be preferred if both options are given. The full schema unit tests already uses both options so would have to touch that again as well if we make this XOR.

I have future dev work in mind that would require using an additional endpoint. In that case, you may want to use discovery for one endpoint and an explicit value for the other. But, admittedly, that's a rather strange case. In any sane environment, you would always use only discovery or only explicit endpoints, I think.

Let me know your thoughts @spacewander.

t/plugin/authz-keycloak.t
```diff
@@ -420,7 +587,7 @@ passed
 
 
 
-=== TEST 10: TEST with fake token and https endpoint
+=== TEST 14: TEST with fake token and https endpoint
```
spacewander, line 590: Should update the expected message to
```
error while sending authz request to https://127.0.0.1:8443/auth/realms/University/protocol/openid-connect/token: 18: self signed certificate
```

jenskeiner: done

apisix/plugins/authz-keycloak.lua
```diff
@@ -64,10 +67,169 @@ local _M = {
...
+    log.debug("authz_keycloak_configure_proxy : use http proxy")
+    httpc:set_proxy_options(proxy_opts)
+  else
+    log.debug("authz_keycloak_configure_proxy : don't use http proxy")
+  end
+end
+
+
+-- Parse the JSON result from a call to the OP.
+local function authz_keycloak_parse_json_response(response, ignore_body_on_success)
```
spacewander, line 130: What is the `ignore_body_on_success` for? It seems there is nowhere to pass it.

doc/plugins/authz-keycloak.md
```diff
@@ -40,14 +40,23 @@ For more information on Keycloak, refer to [Keycloak Authorization Docs](https:/
...
 | ----------------------- | ------------- | ----------- | --------------------------------------------- | ----------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| token_endpoint          | string        | required    |                                               | [1, 4096]                                       | A OAuth2-compliant Token Endpoint that supports the `urn:ietf:params:oauth:grant-type:uma-ticket` grant type.                                               |
+| discovery               | string        | optional    |                                               | [1, 4096]                                       | URL to discovery document for Keycloak Authorization Services.                                                                                              |
```
spacewander, line 43: We should change the `[1, 4096]` to a valid example. It is the problem of original doc.

apisix/plugins/authz-keycloak.lua
spacewander: OK

apisix/plugins/authz-keycloak.lua
jenskeiner: I've taken some of these helper functions largely from the `openidc` module; see https://github.com/zmartzone/lua-resty-openidc

Agree that we don't need that parameter here. Will remove it.

doc/plugins/authz-keycloak.md
jenskeiner: done

t/plugin/authz-keycloak.t
```diff
@@ -49,12 +49,38 @@ done
...
 
-=== TEST 2: full schema check
+=== TEST 2: sanity (using discovery endpoint)
 --- config
     location /t {
         content_by_lua_block {
             local plugin = require("apisix.plugins.authz-keycloak")
-            local ok, err = plugin.check_schema({token_endpoint = "https://efactory-security-portal.salzburgresearch.at/",
+            local ok, err = plugin.check_schema({
+                                   discovery = "https://efactory-security-portal.salzburgresearch.at/.well-known/uma2-configuration",
```
sshniro, line 58: Aw, bummer, it seems I have used my previous dev configurations, please do change it to a generic URL if possible.

jenskeiner: sure

sshniro (approved):

spacewander (approved):

Issue #3261, opened 2021-01-12
request help: support chash key splicing


nanamikon:
### Issue description
Nginx support hash key splicing, just as follow
```
hash ${request_uri}${http_token}${http_sign}  consistent
```

I can't find how to replace it with chash in apisix，  and I think it’s an importance function in prd env

Thanks for helping

### Environment

* apisix version (2.1):
* OS: (ubuntu 16)
* OpenResty / Nginx version: (1.19`)

Firstsawyou:
Take a look at this document: [https://github.com/apache/apisix/blob/master/doc/admin-api.md#upstream ](https://github.com/apache/apisix/blob/master/doc/admin-api.md#upstream)

spacewander:
@nanamikon 
We can add a new `hash_on` type to support it.
There are already a helper function to eval Nginx variable:  https://github.com/apache/apisix/blob/3e19b06293463f52c8d26f0958542fc475ff0fcd/apisix/core/utils.lua#L268
You need to modify https://github.com/apache/apisix/blob/3e19b06293463f52c8d26f0958542fc475ff0fcd/apisix/balancer/chash.lua#L33-L41

PR is welcome!

nanamikon:
@spacewander 
I am new to apisix,  and I test this function follow your tips using test-nginx。
I have following points to discuss

1. the key of hash_on can be defined as  “vars_combinations”,     refers to http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html
```
The key can contain text, variables, and their combinations (1.11.2)
```

2. In upstreams.lua，  should I check the schema in this situation？  I think it is no need to do this， because it is an advance feature
```
local function get_chash_key_schema(hash_on)
    if not hash_on then
        return nil, "hash_on is nil"
    end

    if hash_on == "vars" then
        return core.schema.upstream_hash_vars_schema
    end

    if hash_on == "header" or hash_on == "cookie" then
        return core.schema.upstream_hash_header_schema
    end

    if hash_on == "consumer" then
        return nil, nil
    end
    return nil, "invalid hash_on type " .. hash_on
end
```
3.  If all vals are not exist， for example
```
$request_uri-$host
``` 
the hash on should be "-",  and will not use default chash_key "remote_addr"，  is it acceptable？
Refer to https://trac.nginx.org/nginx/ticket/765，  any plan to support map？

spacewander:
1. good name!
2. we can simply check if it is string type:
```
local xxx_schema = {
[...]

PR #3278, opened 2021-01-14
feat: support hash on vars_combinations

nanamikon:
Support hash on vars_combinations, refer to #3261

### Pre-submission checklist:

* [Y] Did you explain what problem does this PR solve? Or what new features have been added?
* [Y] Have you added corresponding test cases?
* [Y] Have you modified the corresponding document?
* [Y] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

### DIFF
--- a/apisix/admin/upstreams.lua
+++ b/apisix/admin/upstreams.lua
@@ -45,6 +45,10 @@ local function get_chash_key_schema(hash_on)
         return nil, nil
     end
 
+    if hash_on == "vars_combinations" then
+        return core.schema.upstream_hash_vars_combinations_schema
+    end
+
     return nil, "invalid hash_on type " .. hash_on
 end
 
--- a/apisix/balancer/chash.lua
+++ b/apisix/balancer/chash.lua
@@ -39,7 +39,7 @@ local function fetch_chash_hash_key(ctx, upstream)
     elseif hash_on == "cookie" then
         chash_key = ctx.var["cookie_" .. key]
     elseif hash_on == "vars_combinations" then
-        chash_key, err= core.utils.resolve_var(key, ctx.var);
+        chash_key, err = core.utils.resolve_var(key, ctx.var);
         if err then
             core.log.error("could not resolve vars in ", key, " error: ", err)
         end
--- a/apisix/balancer/chash.lua
+++ b/apisix/balancer/chash.lua
@@ -38,6 +38,9 @@ local function fetch_chash_hash_key(ctx, upstream)
         chash_key = ctx.var["http_" .. key]
     elseif hash_on == "cookie" then
         chash_key = ctx.var["cookie_" .. key]
+    elseif hash_on == "vars_combinations" then
+        chash_key = core.utils.resolve_var(key, ctx.var);
+        core.log.info("vars_combinations: " .. chash_key)
     end
 
     if not chash_key then
--- a/apisix/balancer/chash.lua
+++ b/apisix/balancer/chash.lua
@@ -39,8 +39,10 @@ local function fetch_chash_hash_key(ctx, upstream)
     elseif hash_on == "cookie" then
         chash_key = ctx.var["cookie_" .. key]
     elseif hash_on == "vars_combinations" then
-        chash_key = core.utils.resolve_var(key, ctx.var);
-        core.log.info("vars_combinations: " .. chash_key)
+        chash_key, err= core.utils.resolve_var(key, ctx.var);
+        if err then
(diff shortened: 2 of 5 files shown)


t/node/chash-hashon.t
```diff
@@ -625,3 +625,90 @@ chash_key: "chash_val_3"
 chash_key: "chash_val_4"
 chash_key: "chash_val_5"
 chash_key: "chash_val_6"
+
+=== TEST 13: set route(two upstream node, type chash), hash_on vars_combinations
```
spacewander, line 629: Style: need to run reindex:
https://github.com/apache/apisix/blob/9361dc8182053d224aeedb9c2edbf88365a95fc7/CONTRIBUTING.md#L123-L133

juzhiyuan, line 629:
```suggestion
=== TEST 13: set route(two upstream nodes, type chash), hash_on vars_combinations
```

apisix/balancer/chash.lua
```diff
@@ -38,6 +38,9 @@ local function fetch_chash_hash_key(ctx, upstream)
         chash_key = ctx.var["http_" .. key]
     elseif hash_on == "cookie" then
         chash_key = ctx.var["cookie_" .. key]
+    elseif hash_on == "vars_combinations" then
+        chash_key = core.utils.resolve_var(key, ctx.var);
+        core.log.info("vars_combinations: " .. chash_key)
```
membphis, line 43: `val, err = core.utils.resolve_var(key, ctx.var)`

this step may return `nil, err` case, we need to check the return value.

code style, here is a better one:

```lua
        chash_key = core.utils.resolve_var(key, ctx.var)
        core.log.info("vars_combinations: ", chash_key)
```

membphis (changes requested):

apisix/balancer/chash.lua
membphis: and we need an error test case

spacewander: The error check is needed. But the test case can be omitted. Since there is no backtrack in the regex which used to match variable.

t/node/chash-hashon.t
```diff
@@ -625,3 +625,90 @@ chash_key: "chash_val_3"
...
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
```
Firstsawyou, line 662: Style: need three blank lines.

membphis: @Firstsawyou here is the correct way: https://github.com/apache/apisix/pull/3278#discussion_r557028364

t/node/chash-hashon.t
nanamikon: get it

nanamikon: get it

apisix/balancer/chash.lua
nanamikon: @membphis @spacewander I will check the err , and just log the err message ,  is it ok?   I think I have nothing to do with it here,  refer to other codes using "resolve_var"

spacewander: Yes.

apisix/balancer/chash.lua
```diff
@@ -39,8 +39,10 @@ local function fetch_chash_hash_key(ctx, upstream)
     elseif hash_on == "cookie" then
         chash_key = ctx.var["cookie_" .. key]
     elseif hash_on == "vars_combinations" then
-        chash_key = core.utils.resolve_var(key, ctx.var);
-        core.log.info("vars_combinations: " .. chash_key)
+        chash_key, err= core.utils.resolve_var(key, ctx.var);
```
spacewander, line 42: Need a space before `=`.

spacewander (approved):

apisix/admin/upstreams.lua
```diff
@@ -45,6 +45,10 @@ local function get_chash_key_schema(hash_on)
         return nil, nil
     end
 
+    if hash_on == "vars_combinations" then
```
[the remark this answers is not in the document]
spacewander: I think `vars_combinations` is good enough for ordinary people. `complex value` looks like the opposite of `simple value`.

moonming (approved):

tokers (approved):

membphis (approved):

PR #3295, opened 2021-01-14
fix: avoid caching outdated discovery upstream nodes

### LINKED ISSUE #1838: bug: discovery Eureka cannot automatically proxy to a new node

### LINKED ISSUE #2369: bug: Unable to get the latest upstream node list by service discovery

spacewander:
Fix #2369
Fix #1838

### Pre-submission checklist:

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -67,7 +67,7 @@ local function fetch_health_nodes(upstream, checker)
     end
 
     if core.table.nkeys(up_nodes) == 0 then
-        core.log.warn("all upstream nodes is unhealth, use default")
+        core.log.warn("all upstream nodes is unhealthy, use default")
         for _, node in ipairs(nodes) do
             up_nodes[node.host .. ":" .. node.port] = node.weight
         end
--- a/apisix/http/service.lua
+++ b/apisix/http/service.lua
@@ -47,7 +47,13 @@ local function filter(service)
         return
     end
 
-    if not service.value.upstream or not service.value.upstream.nodes then
+    if not service.value.upstream then
+        return
+    end
+
+    service.value.upstream.parent = service
+
+    if not service.value.upstream.nodes then
         return
     end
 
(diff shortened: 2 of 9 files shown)


tokers (approved):

lilien1010 (approved):
LGTM

moonming (approved):

Issue #3211, opened 2021-01-07
Feature: implement p2c balancer with random, ewma, leastloaded


hnlq715:
P2c implemention should have better performance than the implemention with for loop when the upstream contains lots of backend servers. [apisix/balancer/ewma.lua](https://github.com/apache/apisix/blob/master/apisix/balancer/ewma.lua#L82)

I would like to send a PR for this, if you guys interested.

Firstsawyou:
If the performance is better, then welcome to submit PR.

tokers:
@hnlq715 A benchmark result is also desired to let us know the earnings :).

hnlq715:
> @hnlq715 A benchmark result is also desired to let us know the earnings :).

Absolutely 😄

hnlq715:
Simple benchmark result with 20 servers in upstream:
* roundrobin: 10685.03 qps
* ewma: 9207.02 qps
* p2cewma: 10101.92 qps

If the number of servers in upstream grows, the difference between p2cewma and ewma should be significant.
```
apisix: 1 worker + 1 upstream + no plugin
+ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "uri": "/hello",
    "plugins": {
    },
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:1980": 1,
            "127.0.0.1:1981": 1,
            "127.0.0.1:1982": 1,
            "127.0.0.1:1983": 1,
            "127.0.0.1:1984": 1,
            "127.0.0.1:1985": 1,
            "127.0.0.1:1986": 1,
            "127.0.0.1:1987": 1,
            "127.0.0.1:1988": 1,
            "127.0.0.1:1989": 1,
            "127.0.0.1:1990": 1,
            "127.0.0.1:1991": 1,
            "127.0.0.1:1992": 1,
            "127.0.0.1:1993": 1,
            "127.0.0.1:1994": 1,
            "127.0.0.1:1995": 1,
            "127.0.0.1:1996": 1,
            "127.0.0.1:1997": 1,
            "127.0.0.1:1998": 1,
            "127.0.0.1:1999": 1
        }
    }
}'
[...]

PR #3300, opened 2021-01-15
feat: ewma use p2c to improve performance

hnlq715:

Details can be found in #3211 

### Pre-submission checklist:

* [] Have you added corresponding test cases?
* [] Have you modified the corresponding document?

Before
===
apisix: 1 worker + 200 upstream + no plugin
---
```
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.65ms    1.53ms  32.40ms   97.65%
    Req/Sec     5.26k   663.77     8.41k    79.21%
  52974 requests in 5.11s, 211.22MB read
Requests/sec:  10369.65
Transfer/sec:     41.35MB
+ sleep 1
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.82ms    1.99ms  30.06ms   96.26%
    Req/Sec     5.10k   845.05     6.21k    81.00%
  50759 requests in 5.00s, 202.39MB read
Requests/sec:  10149.88
Transfer/sec:     40.47MB
```

apisix: 1 worker + 200 upstream + no plugin + ewma
---

```
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     2.95ms    1.55ms  27.52ms   93.33%
    Req/Sec     2.84k   490.54     6.22k    88.12%
  28488 requests in 5.10s, 113.60MB read
Requests/sec:   5585.89
Transfer/sec:     22.27MB
+ sleep 1
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     3.01ms    0.86ms  23.33ms   85.91%
    Req/Sec     2.68k   228.62     3.11k    68.00%
  26713 requests in 5.00s, 106.52MB read
Requests/sec:   5338.53
Transfer/sec:     21.29MB
```
After
===

apisix: 1 worker + 200 upstream + no plugin + ewma
---
```
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.85ms    0.88ms  21.68ms   94.33%
    Req/Sec     4.48k   562.97     5.48k    68.00%
  44599 requests in 5.00s, 177.83MB read
Requests/sec:   8912.87
Transfer/sec:     35.54MB
+ sleep 1
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.73ms  693.40us  16.92ms   95.53%
    Req/Sec     4.73k   434.96     5.49k    74.51%
  48052 requests in 5.09s, 191.60MB read
Requests/sec:   9440.28
Transfer/sec:     37.64MB
```

### DIFF
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -187,8 +187,6 @@ local function _ewma_find(ctx, up_nodes)
 end
 
 local function _ewma_after_balance(ctx, before_retry)
-    ngx.log(ngx.WARN,"--------",core.json.encode(ctx.balancer_tried_servers), ctx.balancer_tried_servers_count)
-
     if before_retry then
         if not ctx.balancer_tried_servers then
             ctx.balancer_tried_servers = core.tablepool.fetch("balancer_tried_servers", 0, 2)
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -140,9 +140,12 @@ end
 local function _ewma_find(ctx, up_nodes)
     local peers
 
-    if not up_nodes or nkeys(up_nodes) == 0 or
-        (ctx.balancer_tried_servers and ctx.balancer_tried_servers_count == nkeys(up_nodes)) then
+    if not up_nodes or nkeys(up_nodes) == 0 then
         return nil, 'up_nodes empty'
+    end
+
+    if ctx.balancer_tried_servers and ctx.balancer_tried_servers_count == nkeys(up_nodes) then
+        return nil, "all upstream servers tried"
     end
 
     peers = lrucache_trans_format(ctx.upstream_key, ctx.upstream_version, _trans_format, up_nodes)
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -14,6 +14,7 @@ local ngx_shared = ngx.shared
 local ngx_now = ngx.now
 local math = math
 local pairs = pairs
+local ipairs = ipairs
 local next = next
 local error = error
 
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -31,10 +31,8 @@ local _M = {name = "ewma"}
 
 local function lock(upstream)
     local _, err = ewma_lock:lock(upstream .. LOCK_KEY)
-    if err then
-        if err ~= "timeout" then
-            core.log.error("EWMA Balancer failed to lock: ", err)
-        end
+    if err and err ~= "timeout" then
+        core.log.error("EWMA Balancer failed to lock: ", err)
     end
 
     return err
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -3,7 +3,6 @@
 -- Inspiration drawn from:
 -- https://github.com/twitter/finagle/blob/1bc837c4feafc0096e43c0e98516a8e1c50c4421
 --   /finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/PeakEwma.scala
-
 local core = require("apisix.core")
 local resty_lock = require("resty.lock")
 
@@ -12,111 +11,108 @@ local ngx_shared = ngx.shared
 local ngx_now = ngx.now
 local math = math
 local pairs = pairs
-local tostring = tostring
-local string = string
 
 local DECAY_TIME = 10 -- this value is in seconds
 local LOCK_KEY = ":ewma_key"
 
 local shm_ewma = ngx_shared.balancer_ewma
-local shm_last_touched_at= ngx_shared.balancer_ewma_last_touched_at
+local shm_last_touched_at = ngx_shared.balancer_ewma_last_touched_at
 
-local lrucache_addr = core.lrucache.new({
-    ttl = 300, count = 1024
-})
-local lrucache_trans_format = core.lrucache.new({
-    ttl = 300, count = 256
-})
+local lrucache_addr = core.lrucache.new({ttl = 300, count = 1024})
+local lrucache_trans_format = core.lrucache.new({ttl = 300, count = 256})
 
-local ewma_lock, ewma_lock_err = resty_lock:new("balancer_ewma_locks", {timeout = 0, exptime = 0.1})
+local ewma_lock, ewma_lock_err = resty_lock:new("balancer_ewma_locks",
+                                                {timeout = 0, exptime = 0.1})
 if not ewma_lock then
-  error(ewma_lock_err)
+    error(ewma_lock_err)
 end
 
-local _M = { name = "ewma" }
+local _M = {name = "ewma"}
 
 local function lock(upstream)
-  local _, err = ewma_lock:lock(upstream .. LOCK_KEY)
-  if err then
-    if err ~= "timeout" then
-      core.log.error("EWMA Balancer failed to lock: ", err)
+    local _, err = ewma_lock:lock(upstream .. LOCK_KEY)
+    if err then
+        if err ~= "timeout" then
+            core.log.error("EWMA Balancer failed to lock: ", err)
+        end
     end
-  end
 
-  return err
+    return err
 end
 
 local function unlock()
-  local ok, err = ewma_lock:unlock()
-  if not ok then
-    core.log.error("EWMA Balancer failed to unlock: ", err)
-  end
+    local ok, err = ewma_lock:unlock()
+    if not ok then
+        core.log.error("EWMA Balancer failed to unlock: ", err)
+    end
 
-  return err
+    return err
 end
 
 local function decay_ewma(ewma, last_touched_at, rtt, now)
-  local td = now - last_touched_at
-  td = (td > 0) and td or 0
-  local weight = math.exp(-td/DECAY_TIME)
+    local td = now - last_touched_at
+    td = (td > 0) and td or 0
+    local weight = math.exp(-td / DECAY_TIME)
 
-  ewma = ewma * weight + rtt * (1.0 - weight)
-  return ewma
+    ewma = ewma * weight + rtt * (1.0 - weight)
+    return ewma
 end
 
 local function store_stats(upstream, ewma, now)
-  local success, err, forcible = shm_last_touched_at:set(upstream, now)
-  if not success then
-    core.log.warn("shm_last_touched_at:set failed: ", err)
-  end
-  if forcible then
-    core.log.warn("shm_last_touched_at:set valid items forcibly overwritten")
-  end
+    local success, err, forcible = shm_last_touched_at:set(upstream, now)
+    if not success then
+        core.log.warn("shm_last_touched_at:set failed: ", err)
+    end
+    if forcible then
+        core.log
+            .warn("shm_last_touched_at:set valid items forcibly overwritten")
+    end
 
-  success, err, forcible = shm_ewma:set(upstream, ewma)
-  if not success then
-    core.log.warn("shm_ewma:set failed: ", err)
-  end
-  if forcible then
-    core.log.warn("shm_ewma:set valid items forcibly overwritten")
-  end
+    success, err, forcible = shm_ewma:set(upstream, ewma)
+    if not success then
+        core.log.warn("shm_ewma:set failed: ", err)
+    end
+    if forcible then
+        core.log.warn("shm_ewma:set valid items forcibly overwritten")
+    end
 end
 
 local function get_or_update_ewma(upstream, rtt, update)
-  local lock_err = nil
-  if update then
-    lock_err = lock(upstream)
-  end
+    local lock_err = nil
+    if update then
+        lock_err = lock(upstream)
+    end
 
-  local ewma = shm_ewma:get(upstream) or 0
-  if lock_err ~= nil then
-    return ewma, lock_err
-  end
+    local ewma = shm_ewma:get(upstream) or 0
+    if lock_err ~= nil then
+        return ewma, lock_err
+    end
 
-  local now = ngx_now()
-  local last_touched_at = ngx.shared.balancer_ewma_last_touched_at:get(upstream) or 0
-  ewma = decay_ewma(ewma, last_touched_at, rtt, now)
+    local now = ngx_now()
+    local last_touched_at = ngx.shared.balancer_ewma_last_touched_at:get(
+                                upstream) or 0
+    ewma = decay_ewma(ewma, last_touched_at, rtt, now)
 
-  if not update then
+    if not update then
+        return ewma, nil
+    end
+
+    store_stats(upstream, ewma, now)
+
+    unlock()
+
     return ewma, nil
-  end
-
-  store_stats(upstream, ewma, now)
-
-  unlock()
-
-  return ewma, nil
 end
 
 local function get_upstream_name(upstream)
-   return upstream.host .. ":" .. upstream.port
+    return upstream.host .. ":" .. upstream.port
 end
 
 local function score(upstream)
-  -- Original implementation used names
-  -- Endpoints don't have names, so passing in IP:Port as key instead
-  local upstream_name = get_upstream_name(upstream)
-  return get_or_update_ewma(upstream_name, 0, false)
+    -- Original implementation used names
+    -- Endpoints don't have names, so passing in IP:Port as key instead
+    local upstream_name = get_upstream_name(upstream)
+    return get_or_update_ewma(upstream_name, 0, false)
 end
 
 local function parse_addr(addr)
@@ -126,7 +122,7 @@ end
 
 local function _trans_format(up_nodes)
     -- trans
-    --{"1.2.3.4:80":100,"5.6.7.8:8080":100}
+    -- {"1.2.3.4:80":100,"5.6.7.8:8080":100}
     -- into
     -- [{"host":"1.2.3.4","port":"80"},{"host":"5.6.7.8","port":"8080"}]
     local peers = {}
@@ -147,8 +143,7 @@ end
 local function _ewma_find(ctx, up_nodes)
     local peers
 
-    if not up_nodes
-       or core.table.nkeys(up_nodes) == 0 then
+    if not up_nodes or core.table.nkeys(up_nodes) == 0 then
         return nil, 'up_nodes empty'
     end
 
@@ -161,9 +156,9 @@ local function _ewma_find(ctx, up_nodes)
     local endpoint, backendpoint = peers[1], nil
 
     if #peers > 1 then
-        local a, b = math.random(1, #peers), math.random(1, #peers-1)
+        local a, b = math.random(1, #peers), math.random(1, #peers - 1)
         if b >= a then
-            b = b+1
+            b = b + 1
         end
 
         endpoint, backendpoint = peers[a], peers[b]
@@ -174,7 +169,6 @@ local function _ewma_find(ctx, up_nodes)
 
     return get_upstream_name(endpoint)
 end
-
 
 local function _ewma_after_balance(ctx, before_retry)
     if before_retry then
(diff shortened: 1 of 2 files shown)


membphis:
nice PR @hnlq715

apisix/balancer/ewma.lua
```diff
@@ -25,80 +28,102 @@ local lrucache_trans_format = core.lrucache.new({
...
-    if not success then
-        core.log.error("balancer_ewma_last_touched_at:set failed ", err)
-    end
-    if forcible then
-        core.log.warn("balancer_ewma_last_touched_at:set valid items forcibly overwritten")
+local function lock(upstream)
+  local _, err = ewma_lock:lock(upstream .. LOCK_KEY)
+  if err then
+    if err ~= "timeout" then
+      ngx.log(ngx.ERR, string.format("EWMA Balancer failed to lock: %s", tostring(err)))
```
membphis, line 42: better style(better performance):

```suggestion
    core.log.error("EWMA Balancer failed to lock: ", err)
```

apisix/balancer/ewma.lua
```diff
@@ -25,80 +28,102 @@ local lrucache_trans_format = core.lrucache.new({
...
-    if forcible then
-        core.log.warn("balancer_ewma:set valid items forcibly overwritten")
-    end
+  return err
 end
 
+local function unlock()
+  local ok, err = ewma_lock:unlock()
+  if not ok then
+    ngx.log(ngx.ERR, string.format("EWMA Balancer failed to unlock: %s", tostring(err)))
```
membphis, line 52: ditto

please fix similar points

apisix/balancer/ewma.lua
```diff
@@ -25,80 +28,102 @@ local lrucache_trans_format = core.lrucache.new({
...
+  end
 
-local function score(upstream)
-    -- Original implementation used names
-    -- Endpoints don't have names, so passing in host:Port as key instead
-    local upstream_name = upstream.host .. ":" .. upstream.port
-    return get_or_update_ewma(upstream_name, 0, false)
-end
+  local ewma = ngx.shared.balancer_ewma:get(upstream) or 0
+  if lock_err ~= nil then
```
membphis, line 92: it seems wrong, please confirm this code

apisix/balancer/ewma.lua
hnlq715: fixed

apisix/balancer/ewma.lua
hnlq715: fixed

apisix/balancer/ewma.lua
hnlq715: fixed
```
ngx.shared.balancer_ewma => shm_ewma
```

apisix/balancer/ewma.lua
```diff
@@ -135,16 +153,23 @@ local function _ewma_find(ctx, up_nodes)
...
-        endpoint = peers[1]
+        local a, b = math.random(1, #peers), math.random(1, #peers - 1)
+        if b >= a then
+            b = b + 1
+        end
+
+        endpoint, backendpoint = peers[a], peers[b]
+        if score(endpoint) > score(backendpoint) then
+            endpoint, backendpoint = backendpoint, endpoint
+        end
```
spacewander, line 167: We need to sync the `tried_endpoint` check from https://github.com/kubernetes/ingress-nginx/blob/a2e77185cc2e91278962f4f1267246c8fefc6e73/rootfs/etc/nginx/lua/balancer/ewma.lua#L180 to our new implementation.
You can take a look at:
https://github.com/apache/apisix/blob/bbbdf58d555ee89144b6923e751deabf7e0bbf15/apisix/balancer/roundrobin.lua#L54-L74

hnlq715: Maybe this is another topic we could improve the stability, like passive/active healthcheck, tried record & tries count?

spacewander: It would be better to fix this known issue in this PR.

hnlq715: Got it

apisix/balancer/ewma.lua
```diff
@@ -3,105 +3,125 @@
...
     if forcible then
-        core.log.warn("balancer_ewma:set valid items forcibly overwritten")
+        core.log.warn("shm_ewma:set valid items forcibly overwritten")
     end
 end
 
-
 local function get_or_update_ewma(upstream, rtt, update)
+    local lock_err = nil
+    if update then
```
ElvinEfendi, line 82: Not locking for `get` without update will result in potentially incorrect behaviour. Imagine at line 86 you fetch current `ewma` value, then another worker updates it and its `last_touched_at` value before you retrieve `last_touched_at` at line 92. Then you will end up treating the old ewma as a new one. I'm not sure if in practice it would make big difference, but it is definitely not a correct behaviour.

ElvinEfendi: I think the only way of getting rid of locking for "get" operations is to combine ewma and timestamp in the same value and store under single key. But then you would need to do encoding and decoding every time you set and fetch it. It can be interesting to try that and see the performance.

apisix/balancer/ewma.lua
```diff
@@ -3,105 +3,125 @@
...
+local lrucache_trans_format = core.lrucache.new({ttl = 300, count = 256})
+
+local ewma_lock, ewma_lock_err = resty_lock:new("balancer_ewma_locks",
+                                                {timeout = 0, exptime = 0.1})
+
+local _M = {name = "ewma"}
+
+local function lock(upstream)
+    local _, err = ewma_lock:lock(upstream .. LOCK_KEY)
+    if err then
```
tokers, line 34: We can merge these two if conditions to one:

```lua
if err and err ~= "timeout" then
end
```

apisix/balancer/ewma.lua
```diff
@@ -3,105 +3,125 @@
...
-
 local function store_stats(upstream, ewma, now)
     local success, err, forcible = shm_last_touched_at:set(upstream, now)
     if not success then
-        core.log.error("balancer_ewma_last_touched_at:set failed ", err)
+        core.log.warn("shm_last_touched_at:set failed: ", err)
     end
     if forcible then
-        core.log.warn("balancer_ewma_last_touched_at:set valid items forcibly overwritten")
+        core.log
```
tokers, line 67: Why breaking line here?

apisix/balancer/ewma.lua
```diff
@@ -3,105 +3,125 @@
...
     ewma = ewma * weight + rtt * (1.0 - weight)
     return ewma
 end
 
-
 local function store_stats(upstream, ewma, now)
     local success, err, forcible = shm_last_touched_at:set(upstream, now)
     if not success then
-        core.log.error("balancer_ewma_last_touched_at:set failed ", err)
+        core.log.warn("shm_last_touched_at:set failed: ", err)
```
tokers, line 64: I think the `error` level is reasonable.

hnlq715: This error often came out when the dict is full, but everything goes fine because we can evict the old items in the dict.

Refer to ingress-nginx
https://github.com/kubernetes/ingress-nginx/blob/master/rootfs/etc/nginx/lua/balancer/ewma.lua#L68

apisix/balancer/ewma.lua
hnlq715: Because  the max column number is set to `80` so this line break into two lines by format tool, I fixed it by setting it to `82`.

apisix/balancer/ewma.lua
hnlq715: done

apisix/balancer/ewma.lua
tokers: Just following the codeing style of APISIX, 100 characters are allowed in one line.

apisix/balancer/ewma.lua
tokers: Most of time, `error` level is used in production clusters, the dict full is not a trivial problem, if the level is `warn`, i'm afraid the problem will be covered up and the troubleshooting might be difficult.

apisix/balancer/ewma.lua
membphis: ping @hnlq715

hnlq715: > Not locking for `get` without update will result in potentially incorrect behaviour. Imagine at line 86 you fetch current `ewma` value, then another worker updates it and its `last_touched_at` value before you retrieve `last_touched_at` at line 92. Then you will end up treating the old ewma as a new one. I'm not sure if in practice it would make big difference, but it is definitely not a correct behaviour.

@ElvinEfendi You're right, this is not a correct behavior, but maybe a proper solution. In this situation you mentioned, `last_touched_at` value should have little difference (almost the same time), and we do not need to lock all `get` operation, which is quite heavy.

> I think the only way of getting rid of locking for "get" operations is to combine ewma and timestamp in the same value and store under single key. But then you would need to do encoding and decoding every time you set and fetch it. It can be interesting to try that and see the performance.

My first implementation is using worker process cache to store this, and we can simply avoid locking, without frequently encoding and decoding, which have a better performance. But we still have other concerns, details can be found in #3211 

And with shared memory, we have to sacrifice between performance and correctness.

I did some benchmark between these two situation:

lock get and update
---
```
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.62ms  282.29us   7.64ms   90.45%
    Req/Sec     4.98k   511.87     9.62k    97.03%
  50013 requests in 5.10s, 199.42MB read
Requests/sec:   9806.24
Transfer/sec:     39.10MB
+ sleep 1
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.69ms  288.94us   7.25ms   89.40%
    Req/Sec     4.77k   285.81     5.28k    63.73%
  48370 requests in 5.10s, 192.87MB read
Requests/sec:   9484.82
Transfer/sec:     37.82MB
```

lock update
---
```
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.57ms  289.91us   7.23ms   89.61%
    Req/Sec     5.14k   584.14    10.43k    96.04%
  51652 requests in 5.10s, 205.95MB read
Requests/sec:  10128.09
Transfer/sec:     40.38MB
+ sleep 1
+ wrk -d 5 -c 16 http://127.0.0.1:9080/hello
Running 5s test @ http://127.0.0.1:9080/hello
  2 threads and 16 connections
    Latency     1.55ms  255.99us   6.55ms   89.96%
    Req/Sec     5.18k   539.62     9.77k    95.05%
  52008 requests in 5.10s, 207.37MB read
Requests/sec:  10198.48
Transfer/sec:     40.66MB
```

apisix/balancer/ewma.lua
hnlq715: done

apisix/balancer/ewma.lua
spacewander: It seems this problem is still not addressed?

spacewander (changes requested):

apisix/balancer/ewma.lua
hnlq715: Sorry, filter logic is added, I'm focusing something else in these days.

apisix/balancer/ewma.lua
```diff
@@ -119,38 +137,76 @@ local function _trans_format(up_nodes)
...
+        ctx.balancer_tried_servers = tried_endpoints
     else
-        endpoint = peers[1]
+        tried_endpoints = ctx.balancer_tried_servers
     end
 
-    return endpoint.host .. ":" .. endpoint.port
-end
+    local filtered_peers
+    for _, peer in ipairs(peers) do
```
spacewander, line 161: ERROR: apisix/balancer/ewma.lua: line 161: getting the Lua global "ipairs"

spacewander:
And we need a test for "fiter tried servers".

apisix/balancer/ewma.lua
hnlq715: fixed

hnlq715:
> And we need a test for "fiter tried servers".

@spacewander test case added.

apisix/balancer/ewma.lua
```diff
@@ -35,73 +58,69 @@ local function decay_ewma(ewma, last_touched_at, rtt, now)
...
 
-
 local function get_or_update_ewma(upstream, rtt, update)
+    local lock_err = nil
+    if update then
+        lock_err = lock(upstream)
+    end
+
     local ewma = shm_ewma:get(upstream) or 0
+    if lock_err ~= nil then
```
tokers, line 86: Why check `lock_err` after fetching `ewma`? It's so strange we don't check the return value after `lock` return. Or you should get `ewma` before fetching the lock.

t/node/ewma.t
```diff
@@ -215,3 +215,74 @@ GET /t
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+[{"count":12,"port":"1980"}]
+--- error_code: 200
+--- no_error_log
+[error]
+
```
tokers, line 287: Redundant empty lines.

hnlq715: done.

apisix/balancer/ewma.lua
hnlq715: We need lock when update, and do not lock when get.

spacewander:
The CI was broken. You might need to submit a new commit to trigger it.

hnlq715:
> The CI was broken. You might need to submit a new commit to trigger it.

Seems like the CI is still broken...

t/node/ewma.t
```diff
@@ -215,3 +215,79 @@ GET /t
...
+            if code ~= 200 then
+                ngx.say("update route failed")
+                return
+            end
+
+            local http = require "resty.http"
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port
+                        .. "/ewma"
+
+            --should select the 1980 node, because 1984 is invalid
```
spacewander, line 259: The 1984 is not invalid, you can confirm it via `less t/servroot/conf/nginx.conf`.

t/node/ewma.t
```diff
@@ -215,3 +215,79 @@ GET /t
...
+            local ports_count = {}
+            for i = 1, 12 do
+                local httpc = http.new()
```
spacewander, line 262: You can speed up the test like this:
https://github.com/apache/apisix/blob/b78223c38cfe19529c0d96882dc85eb4c9e2bb54/t/node/chash-balance.t#L309

apisix/balancer/ewma.lua
```diff
@@ -119,38 +138,76 @@ local function _trans_format(up_nodes)
...
+    peers = lrucache_trans_format(ctx.upstream_key, ctx.upstream_version, _trans_format, up_nodes)
     if not peers then
         return nil, 'up_nodes trans error'
     end
 
-    if #peers > 1 then
-        endpoint = pick_and_score(peers)
+    local tried_endpoints
+    if not ctx.balancer_tried_servers then
+        tried_endpoints = {}
```
spacewander, line 155: It seems we can save the table allocation of `tried_endpoints` and `filtered_peers` when `not ctx.balancer_tried_servers`?

apisix/balancer/ewma.lua
```diff
@@ -119,38 +138,76 @@ local function _trans_format(up_nodes)
...
+            if not filtered_peers then
+                filtered_peers = {}
+            end
+            table_insert(filtered_peers, peer)
+        end
+    end
 
+    if not filtered_peers then
+        core.log.warn("all endpoints have been retried")
+        filtered_peers = table_deepcopy(peers)
```
spacewander, line 173: Need a test to cover this branch. And since we don't need to copy the peers?

apisix/balancer/ewma.lua
```diff
@@ -119,38 +138,76 @@ local function _trans_format(up_nodes)
...
     if not peers then
         return nil, 'up_nodes trans error'
     end
 
-    if #peers > 1 then
-        endpoint = pick_and_score(peers)
+    local tried_endpoints
+    if not ctx.balancer_tried_servers then
+        tried_endpoints = {}
+        ctx.balancer_tried_servers = tried_endpoints
```
spacewander, line 156: We already set `ctx.balancer_tried_servers` in the `after_balancer`?

apisix/balancer/ewma.lua
hnlq715: Got it, we called `after_balance` in the first time.

apisix/balancer/ewma.lua
hnlq715: Yep, I got it :-)

t/node/ewma.t
hnlq715: To avoid misunderstand, I changed it to 8888

apisix/balancer/ewma.lua
hnlq715: done

t/node/ewma.t
hnlq715: done

apisix/balancer/ewma.lua
```diff
@@ -119,38 +137,75 @@ local function _trans_format(up_nodes)
...
         return nil, 'up_nodes trans error'
     end
 
-    if #peers > 1 then
-        endpoint = pick_and_score(peers)
-    else
-        endpoint = peers[1]
+    local filtered_peers
+    for _, peer in ipairs(peers) do
+        if ctx.balancer_tried_servers then
```
spacewander, line 158: This branch should move outside the for loop. And set the filtered_peers to peers if not in retry.

apisix/balancer/ewma.lua
```diff
@@ -119,38 +137,75 @@ local function _trans_format(up_nodes)
...
+
+                table_insert(filtered_peers, peer)
+            end
+        end
     end
 
-    return endpoint.host .. ":" .. endpoint.port
-end
+    if not filtered_peers then
+        core.log.warn("all endpoints have been retried")
```
spacewander, line 170: We don't need this branch as we already reject this case above:
```lua
 if ctx.balancer_tried_servers and ctx.balancer_tried_servers_count == nkeys(up_nodes) then
        return nil, "all upstream servers tried"
    end
```

spacewander (changes requested):

apisix/balancer/ewma.lua
```diff
@@ -35,73 +57,69 @@ local function decay_ewma(ewma, last_touched_at, rtt, now)
...
 
     if not update then
-        return ewma
+        return ewma, nil
     end
 
     store_stats(upstream, ewma, now)
 
-    return ewma
+    unlock()
```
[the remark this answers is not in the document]
hnlq715: We have record the error, when unlock() failed.

apisix/balancer/ewma.lua
hnlq715: done

apisix/balancer/ewma.lua
hnlq715: done

apisix/balancer/ewma.lua
hnlq715: We need to return ewma value for get operation.

spacewander (approved):

membphis (approved):

PR #3308, opened 2021-01-15
feat(authz-keycloak): dynamic scope and resource mapping.

### LINKED ISSUE #3274: feat: auth-keycloak plugin - Dynamically resolve resources and scopes from incoming request.
I'm opening this issue to get some feedback on some proposed changes to the `auth-keycloak` plugin before I shoot off and start working on them.

The current incarnation of the plugin has some limitations that limit its usefulness. The most visible limitation, to me at least, is that you can only configure a static set of resource/scope pairs.

Without delving into too many details of the Keycloak Authorization Services API (described here in all its glory: https://www.keycloak.org/docs/latest/authorization_services/index.html), it's worth mentioning that permissions are defined in terms of resource/scope pairs. If you think of some REST service that you want to protect, then a resource would typically be a subset of the available endpoints, defined by some URI pattern, e.g. `/foo/bar/*`. A scope typically describes the type of action to perform on the resource, e.g. read, write, delete, but can also really be something completely different and abstract, depending on the use case.

To check if an incoming request should be allowed to proceed to the upstream, or if it should be rejected, the plugin currently uses the token endpoint (which is also used for OIDC authentication) with a special grant type. The requested permissions, specified by resource/scope pairs are sent as a parameter to that endpoint together with the bearer token from the request. Keycloak evaluates the permissions and answers with the decision, i.e. whether the permissions are granted or not. If yes, then the request can proceed, otherwise, it is rejected.

As I said above, the set of resource/scope pairs is currently fixed in the plugin configuration. But often, what you want to do is to determine the resource and scope from features of the incoming request like HTTP method and URI. Actually, the documentation mentions this as future work, but that has so far not materialized.

Just upfront: The changes I want to propose would need to be split into smaller PRs that build onto each other.

Basically, I want to propose to implement (optional) dynamic resource and scope resolution as follows:
- The scope is determined by the HTTP method of the request (i.e. a GET request maps to scope GET).
- The resource is determined by using Keycloak's resource registration endpoint which can return the resources that match a given URI.

I have a working example of this in a branch. There are some points to observe though:
- Using the resource registration endpoint requires another request to Keycloak on top of the one to the token endpoint that returns the decision. Longer term, but maybe not in the first iteration, caching can be implemented to reduce the number of necessary requests.
- To be able to invoke the resource registration endpoint, the plugin first needs to obtain a separate access token for itself from the token endpoint via client ID and secret. This is a special token that has the `uma_protection` scope which allows access to the Protection API and hence the resource registration endpoint. It's relative easy to obtain the token, but some management needs to be implemented because the token needs to be refreshed regularly as it will typically be short-lived. This will make the code base a bit more complex, but cannot be avoided.
- Longer term, it may be better to load all resources from Keycloak once (maybe with periodic updates), and match the request URI to the resources in the plugin, e.g. with the help of the radix tree library already used for routes. This would avoid the requests that are needed just to determine the resource.

To sum up, I would propose above changes, but with the focus on first getting the scope and resource discovery working properly, before focusing on further optimizations like caching.

Would love to hear anyone's thoughts. Maybe @sshniro can chime in because you've worked a lot with the plugin already.

jenskeiner:
Add code to `authz-keycloak` plugin to optionally resolve requests URI to resource(s) and map HTTP method to scope; fixes #3274.

### Pre-submission checklist:

Open this PR to get some feedback on the changes. Still need to add tests to cover the new functionality. May still need to polish the JSON schema slightly as well. However, already tested changes in a real deployment to confirm this is actually working.

Also, strictly speaking, this plugin is not backwards compatible. E.g. we now require static permissions to be set when not using the new dynamic resource resolution. We also require `client_id` to be set because it will be used in any sensible configuration scenario. This mostly affects corner cases, so I don't think it represents a real problem.

In any case, will likely polish everything a bit more, but would appreciate any feedback already.

@sshniro You should definitely have a look.


apisix/cli/ngx_tpl.lua
```diff
@@ -371,16 +374,16 @@ http {
         {% end %}
         {% end %} {% -- if enable_ipv6 %}
 
+        {% if ssl.ssl_trusted_certificate ~= nil then %}
```
jenskeiner, line 377: Moved this here since it was previously only used when APISIX itself was accepting HTTPS from the outside. But even if TLS is disabled, internally, plugins that send requests may still need to be able to use TLS and may have custom CA certs configured.

apisix/plugins/authz-keycloak.lua
```diff
@@ -42,7 +43,6 @@ local schema = {
             enum = {"urn:ietf:params:oauth:grant-type:uma-ticket"},
             minLength = 1, maxLength = 100
         },
-        audience = {type = "string", minLength = 1, maxLength = 100},
```
jenskeiner, line 45: This is now called `client_id` which is more consistent with the `opendid-connect` plugin as well.

spacewander: Maybe we can keep the old field and add a deprecated check?
https://github.com/apache/apisix/blob/bbbdf58d555ee89144b6923e751deabf7e0bbf15/apisix/init.lua#L463-L470

jenskeiner: @spacewander Possible, it's just a name. So we keep `client_id` as well and prefer it, if given?

spacewander: Yes. Maybe we can add a description on them:
https://github.com/apache/apisix/blob/bbbdf58d555ee89144b6923e751deabf7e0bbf15/apisix/schema_def.lua#L75

jenskeiner: Ok, will change accordingly. Will also add/fix more test cases soon.

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +239,332 @@ local function authz_keycloak_get_token_endpoint(conf)
...
+
+                        -- Calculate and save refresh token expiry time.
+                        session.refresh_token_expiration = current_time
+                                + authz_keycloak_refresh_token_expires_in(conf,
+                                                                          json.refresh_expires_in)
+                    end
+
+                    authz_keycloak_cache_set("access_tokens",
+                                             token_endpoint .. ":" .. conf.client_id,
+                                             core.json.encode(session), 24 * 60 * 60)
```
spacewander, line 361: Better to allow customizing cache time.

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +239,332 @@ local function authz_keycloak_get_token_endpoint(conf)
...
+                        client_secret = conf.client_secret,
+                        refresh_token = session.refresh_token,
+                    }),
+                    ssl_verify = conf.ssl_verify,
+                    headers = {
+                        ["Content-Type"] = "application/x-www-form-urlencoded"
+                    }
+                }
+
+                local current_time = ngx.time()
```
spacewander, line 314: We already set `current_time` above?

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +239,332 @@ local function authz_keycloak_get_token_endpoint(conf)
...
+        if json.refresh_token ~= nil then
+            session.refresh_token = json.refresh_token
+
+            -- Calculate and save refresh token expiry time.
+            session.refresh_token_expiration = current_time
+                    + authz_keycloak_refresh_token_expires_in(conf, json.refresh_expires_in)
+        end
+
+        authz_keycloak_cache_set("access_tokens", token_endpoint .. ":" .. conf.client_id,
+                                 core.json.encode(session), 24 * 60 * 60)
```
spacewander, line 435: Ditto

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +239,332 @@ local function authz_keycloak_get_token_endpoint(conf)
...
-        return false
+local function authz_keycloak_get_resource_registration_endpoint(conf)
+    return authz_keycloak_get_endpoint(conf, "resource_registration_endpoint")
+end
+
+
+-- computes access_token expires_in value (in seconds)
+local function authz_keycloak_access_token_expires_in(opts, expires_in)
+  return (expires_in or opts.access_token_expires_in or 300)
+         - 1 - (opts.access_token_expires_leeway or 0)
```
spacewander, line 250: Where do we set `access_token_expires_in` and `access_token_expires_leeway`?

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +239,332 @@ local function authz_keycloak_get_token_endpoint(conf)
...
+end
+
+
+-- computes refresh_token expires_in value (in seconds)
+local function authz_keycloak_refresh_token_expires_in(opts, expires_in)
+  return (expires_in or opts.refresh_token_expires_in or 3600)
+         - 1 - (opts.refresh_token_expires_leeway or 0)
```
spacewander, line 257: Ditto

apisix/cli/ngx_tpl.lua
```diff
@@ -146,6 +146,9 @@ http {
     lua_shared_dict jwks                  1m; # cache for JWKs
     lua_shared_dict introspection        10m; # cache for JWT verification results
 
+    # for authz-keycloak
+    lua_shared_dict access_tokens         1m; # cache for service account access tokens
```
spacewander, line 150: Need to add it after:
https://github.com/apache/apisix/blob/bbbdf58d555ee89144b6923e751deabf7e0bbf15/t/APISIX.pm#L248

apisix/plugins/authz-keycloak.lua
```diff
@@ -53,10 +54,24 @@ local schema = {
         keepalive_timeout = {type = "integer", minimum = 1000, default = 60000},
         keepalive_pool = {type = "integer", minimum = 1, default = 5},
         ssl_verify = {type = "boolean", default = true},
+        client_id = {type = "string", minLength = 1, maxLength = 100},
+        client_secret = {type = "string", minLength = 1, maxLength = 100},
+        lazy_load_paths = {type = "boolean", default = false},
+        http_method_as_scope = {type = "boolean", default = false},
     },
+    required = {"client_id"},
```
spacewander, line 62: To avoid break change, we should not make `client_id` required. If `client_id` missed, we should continue the old logic.

spacewander (changes requested):

apisix/plugins/authz-keycloak.lua
jenskeiner: This has been copied over from the `openidc` module. As far as I know, you cannot assume that the token endpoint will always return an `expires_in` and/or `refresh_expires_in` field as well. If it's not returned, this code allows to configure an expiry time assumption via `opts.refresh_token_expires_in`, falling back to a hard-coded default if that's not available neither.

Generally, I think having a fallback in place is useful. If needed, we could expose it through the schema as well. Same holds for the `leeway` parameter which can be used to make the plugin refresh tokens earlier. On the other hand, it's true that the mentioned options are currently not set anywhere.

Please advise whether this should be removed for now, or whether keeping it makes sense.

apisix/cli/ngx_tpl.lua
jenskeiner: Will do. Can you quickly explain the reason?

spacewander: The configuration used in test is generated from `apisix/t/APISIX.pm` instead of `apisix/cli/ngx_tpl.lua`.

apisix/plugins/authz-keycloak.lua
spacewander: We could expose it through the schema. If you think it will make this PR more complex, you can remove it too.
It's all up to you.

apisix/plugins/authz-keycloak.lua
```diff
@@ -53,10 +54,26 @@ local schema = {
...
+        http_method_as_scope = {type = "boolean", default = false},
+        cache_ttl_seconds = {type = "integer", minimum = 1, default = 24 * 60 * 60},
     },
     anyOf = {
         {required = {"discovery"}},
-        {required = {"token_endpoint"}}}
+        {required = {"token_endpoint"}}
+    },
+    dependencies = {
+        lazy_load_paths = {
```
spacewander, line 70: Should we not give a default value to `lazy_load_paths`?
Currently either discovery or resource_registration_endpoint is required as there is always a value for lazy_load_paths.

apisix/plugins/authz-keycloak.lua
jenskeiner: Ok, will add it to the schema then. Better to have some levers to configure this in case of issues with token expiry.

apisix/cli/ngx_tpl.lua
jenskeiner: Ah, ok.

apisix/plugins/authz-keycloak.lua
jenskeiner: Fixing. Defaults to `false`. Should require one of `discovery` or `resource_registration_endpoint` if set to `true`.

apisix/plugins/authz-keycloak.lua
jenskeiner: Ok, done.

apisix/plugins/authz-keycloak.lua
jenskeiner: done

apisix/plugins/authz-keycloak.lua
jenskeiner: Removed this invocation.

apisix/plugins/authz-keycloak.lua
jenskeiner: We don't set these at the moment; see next item below.

apisix/cli/ngx_tpl.lua
jenskeiner: @spacewander Quick question: We also need a shared dict for the discovery documents. I have added that to `apisix/t/APISIX.pm` now as well. But I can't see other dicts that e.g. the `openid-connect` plugin needs in that file. How do the tests then run successfully w/o the dict definitions?

apisix/plugins/authz-keycloak.lua
jenskeiner: done

apisix/plugins/authz-keycloak.lua
jenskeiner: Should be fixed now.

jenskeiner:
I reckon the new tests added are sufficient to check the basic validity of the updated schema wrt. to (conditionally) required attributes like `resource_registration_endpoint`.

However, it would make sense to add more cases to test the actual functionality of `lazy_load_paths` and `http_method_as_scope`. The problem is that this would very likely require some modifications to the Keycloak container image since we need resources linked to scopes that correspond to HTTP method names. Also would probably need to verify resource definitions' URI settings so we can test the resource resolution via the resource registration endpoint.

Feedback?

The Docker image is in @sshniro's private repo. Can you advise @sshniro how we should go about making modifications?

jenskeiner:
Also, I could use some help on test case #7. I want to conditionally require either `discovery` or `resource_registration_endpoint` when `lazy_load_paths` is set to `true`. But there must be something wrong with my schema definition where I'm currently trying to use `if` and `then` (part of newer schema draft standards). The invalid configuration in case #7 seems to be accepted -- which is wrong. Can anybody help? Maybe @spacewander?

spacewander:
The `if` and `then` schema is not supported by the Lua implementation of JSONSchema yet.

jenskeiner:
> The `if` and `then` schema is not supported by the Lua implementation of JSONSchema yet.

Ok, I think I'm on top of it now. Just used `anyOf` to the same effect. Minor point is that the error message that the schema validator returns  -- see test case 7 -- is not really detailed. But it works as expected.

jenskeiner:
I've switched to an updated Keycloak image in my own DockerHub repo. This one has some slight configuration changes to be able to test the URI-to-resource and HTTP-method-to-scope mapping.

Also, other tests currently disabled to make the checks shorter. Will re-enable when I've added more test cases.

jenskeiner:
When @sshniro has given me access to his Docker repo, I will update the Keycloak image there and revert the change in the test runner script to pull the image from a different repo.

sshniro:
LGTM, also @jenskeiner I have granted access to the repo, would be great if we can tag this image, rather than replacing the latest.

jenskeiner:
> LGTM, also @jenskeiner I have granted access to the repo, would be great if we can tag this image, rather than replacing the latest.

Ok, great. Any preference as to which tag name I should use?

doc/plugins/authz-keycloak.md
```diff
@@ -38,24 +38,37 @@ For more information on Keycloak, refer to [Keycloak Authorization Docs](https:/
...
+| resource_registration_endpoint | string        | optional    |                                               | https://host.domain/auth/realms/foo/authz/protection/resource_set  | A Keycloak Protection API-compliant resource registration endpoint. Overrides value from discovery, if given.                                               |
+| grant_type                     | string        | optional    | "urn:ietf:params:oauth:grant-type:uma-ticket" | ["urn:ietf:params:oauth:grant-type:uma-ticket"]                    |                                                                                                                                                             |
+| client_id                      | string        | required    |                                               |                                                                    | The client identifier of the resource server to which the client is seeking access. <br>This parameter is mandatory when parameter permission is defined.   |
```
spacewander, line 47: > This parameter is mandatory when parameter permission is defined.

Look like it should be `when parameter lazy_load_paths is defined`? Should require it in the schema and add a test for it.

apisix/plugins/authz-keycloak.lua
```diff
@@ -224,31 +275,326 @@ local function authz_keycloak_get_token_endpoint(conf)
...
+    local ttl = conf.cache_ttl_seconds
+    local token_endpoint = authz_keycloak_get_token_endpoint(conf)
+
+    if not token_endpoint then
+      log.error("Unable to determine token endpoint.")
+      return 500, "Unable to determine token endpoint."
+    end
+
+    local session = authz_keycloak_cache_get("access_tokens", token_endpoint .. ":"
+                                             .. client_id)
```
spacewander, line 308: The `client_id` could be nil if you don't force when lazy_load_paths is used.

t/plugin/authz-keycloak.t
```diff
@@ -137,7 +224,7 @@ done
                             "authz-keycloak": {
                                 "token_endpoint": "http://127.0.0.1:8090/auth/realms/University/protocol/openid-connect/token",
                                 "permissions": ["course_resource#view"],
-                                "audience": "course_management",
```
spacewander, line 140: Maybe we can leave `audience` unchanged in one or two cases so that we know the old configuration is still compatible?

spacewander:
BTW, you can split the test into two files, like the https://github.com/apache/apisix/blob/master/t/plugin/limit-conn2.t. Since it is too big now.

doc/plugins/authz-keycloak.md
```diff
@@ -38,24 +38,37 @@ For more information on Keycloak, refer to [Keycloak Authorization Docs](https:/
...
+| permissions                    | array[string] | optional    |                                               |                                                                    | Static permission to request, an array of strings each representing a resources and optionally one or more scopes the client is seeking access.             |
+| lazy_load_paths                | boolean       | optional    | false                                         |                                                                    | Dynamically resolve the request URI to resource(s) using the resource registration endpoint instead of using the static permission.                         |
+| http_method_as_scope           | boolean       | optional    | false                                         |                                                                    | Map HTTP request type to scope of same name and add to all permissions requested.                                                                           |
```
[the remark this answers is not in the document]
jenskeiner: Correct. Will add.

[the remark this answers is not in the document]
jenskeiner: Added.

apisix/plugins/authz-keycloak.lua
```diff
@@ -171,13 +222,13 @@ local function authz_keycloak_discover(url, ssl_verify, keepalive, timeout,
       keepalive = (keepalive ~= "no")
```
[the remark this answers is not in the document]
jenskeiner: My bad, was copied over from `openidc` module. Fixed it and consolidated the HTTP client setup -- which is done in various places -- to apply options like `timeout`, `keepalive`, `ssl_verify`and others consistently.

doc/plugins/authz-keycloak.md
jenskeiner: Ok, so I'm now again requiring one of `client_id` or `audience`, the latter for better backwards compatibility. Rationale: The plugin always uses the token endpoint to get a decision from Keycloak on the requested permissions. As per https://www.keycloak.org/docs/4.8/authorization_services/#_service_obtaining_permissions the `audience` parameter (what is either in `client_id` or `audience` for the plugin) is required when the `permission` parameter is set (which the plugin always does).

If `lazy_load_paths` is `true`, the the plugin also needs to obtain an access token for itself from Keycloak. To that end, it most likely also needs `client_secret` to be set, but that depends on how Keycloak is configured. Hence, this is still optional.

Also added test cases to check if one of `client_id` or `audience` is correctly required and schema verification fails if both are absent.

apisix/plugins/authz-keycloak.lua
jenskeiner: See above, we're now requiring `client_id` (or `audience`).

t/plugin/authz-keycloak.t
jenskeiner: Sure, I can add some cases that use it.

doc/plugins/authz-keycloak.md
spacewander: I see.

jenskeiner:
> BTW, you can split the test into two files, like the https://github.com/apache/apisix/blob/master/t/plugin/limit-conn2.t. Since it is too big now.

done

jenskeiner:
@spacewander I think I've addressed all of the feedback so far. Thanks for reviewing. Let me know any further issues.

apisix/plugins/authz-keycloak.lua
```diff
@@ -69,225 +109,566 @@ local _M = {
...
+-- Return the resource registration endpoint from the configuration.
+local function authz_keycloak_get_resource_registration_endpoint(conf)
+    return authz_keycloak_get_endpoint(conf, "resource_registration_endpoint")
+end
+
+
+-- Return access_token expires_in value (in seconds).
+local function authz_keycloak_access_token_expires_in(conf, expires_in)
+    return (expires_in or conf.access_token_expires_in or 300)
+           - 1 - (conf.access_token_expires_leeway or 0)
```
spacewander, line 318: Will you add this configurations in this PR, or in the future one?

apisix/plugins/authz-keycloak.lua
```diff
@@ -69,225 +109,566 @@ local _M = {
...
+end
+
+
+-- Return refresh_token expires_in value (in seconds).
+local function authz_keycloak_refresh_token_expires_in(conf, expires_in)
+    return (expires_in or conf.refresh_token_expires_in or 3600)
+           - 1 - (conf.refresh_token_expires_leeway or 0)
```
spacewander, line 325: Ditto

spacewander (approved):
We can merge this PR first and solve the remain in the next one.

jenskeiner:
@spacewander Thanks, I will surely continue the work on this plugin and I can add the config options in a future PR.

membphis:
I have been busy recently, I will review this PR later.

Welcome everyone to review.  many thx

moonming (approved):

Issue #3362, opened 2021-01-20
discuss: should we support the etcd missing some special keys case


yongboy:
Sometimes, we may not need some keys below existing in ETCD, or maybe deleted by some OPS.

| Key | Value    |
|-----|----------|
| /apisix/services | init_dir |
| /apisix/node_status | init_dir |
| /apisix/ssl | init_dir |
| /apisix/global_rules | init_dir |
| /apisix/stream_routes | init_dir |
| /apisix/proto | init_dir |
| /apisix/plugin_metadata | init_dir |
| ... | ... |

In that case, when the nginx's worker process got reload signal, which will not load all the data from ETCD failure.

En, should I to submit a PR to fix the BUG ?

PS:

The PR code below is working well with Apisix 2.2  in my online env now ~

The `apisix.core.config-etcd.lua` ori code:

```lua
            if not dir_res.nodes then
                dir_res.nodes = {}
            end
```

My way:

```lua
            if not dir_res.nodes then
                if dir_res.key then
                    dir_res.nodes = { clone_tab(dir_res) }
                else
                    dir_res.nodes = {}
                end
            end
```

nic-chen:
I remember that if these keys (`/apisix/routes`, `/apisix/services`, etc.) do not exist, the ETCD watch will fail. 
@Yiyiyimu  please have a look.

yongboy:
Here is the PR: #3372

Yiyiyimu:
> I remember that if these keys (`/apisix/routes`, `/apisix/services`, etc.) do not exist, the ETCD watch will fail.
> @Yiyiyimu please have a look.

Yeah I remember the same. But lots of things got optimized after my attempt, and it would be great if we could get rid of them. I'll help to track the PR

PR #3372, opened 2021-01-21
bugfix: add compatible for ETCD missing some special keys lead to watch failure

yongboy:

In some case,  the ETCD maybe missing some special keys, such as `/apisix/services`, then the APISIX will fail in init existing etcd data.

Just fix the discuss: #3362.

### Pre-submission checklist:

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -241,7 +241,11 @@ local function sync_data(self)
 
         else
             if not dir_res.nodes then
-                dir_res.nodes = {}
+                if dir_res.key then
+                    dir_res.nodes = { clone_tab(dir_res) }
+                else
+                    dir_res.nodes = {}
+                end
             end
 
             self.values = new_tab(#dir_res.nodes, 0)


apisix/core/config_etcd.lua
```diff
@@ -241,7 +241,11 @@ local function sync_data(self)
 
         else
             if not dir_res.nodes then
-                dir_res.nodes = {}
+                if dir_res.key then
+                    dir_res.nodes = { clone_tab(dir_res) }
+                else
+                    dir_res.nodes = {}
+                end
```
spacewander, line 248: Should remove the keys: https://github.com/apache/apisix/blob/aff1a907a3671bd6610b304a54ebef47022ca6cf/apisix/cli/etcd.lua#L206
to see if this idea works.

spacewander (changes requested):

Yiyiyimu:
Hi @yongboy it would be better if you write some test case to prove it's correctness

apisix/core/config_etcd.lua
```diff
@@ -241,7 +241,11 @@ local function sync_data(self)
 
         else
             if not dir_res.nodes then
-                dir_res.nodes = {}
+                if dir_res.key then
+                    dir_res.nodes = { clone_tab(dir_res) }
```
Yiyiyimu, line 245: Could you explain it a bit? I didn't get why could it solve the problem

Yiyiyimu:
ping @yongboy

yongboy:
Need I submit the PR again ?

Yiyiyimu:
@yongboy no need to resubmit the PR, but add more commits to the same branch. And you could notice that the CI hasn't passed yet so it could not be merged in any aspects. 

BTW, I don't think it's a good idea to click **resolve conversation** button when you actually didn't. If you found the conversation does not matter to your PR, at least leave some comments about why it's not important.

yongboy:
Maybe I did something wrong on the github workflow, and maybe I did not copy your meaning  :((

Yiyiyimu:
> Maybe I did something wrong on the github workflow, and maybe I did not copy your meaning :((

Hi @yongboy you should make the CI pass. See
image

PR #3382, opened 2021-01-21
chore: the 500 response code in the upstream.lua file was changed to 503 response code

### LINKED ISSUE #3374: Modify the 500 response code in the upstream.lua file to 503 response code
### Issue description

The `500` response code should be changed to a `503` response code. Because this is a situation where upstream cannot work.
Code link: [https://github.com/apache/apisix/blob/master/apisix/upstream.lua#L125-L168 ](https://github.com/apache/apisix/blob/master/apisix/upstream.lua#L125-L168)

### Environment

* OS: (cmd: `uname -a`)
* OpenResty / Nginx version: (cmd: `nginx -V` or `openresty -V`)

Firstsawyou:

fix #3374

### Pre-submission checklist:

### DIFF
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -131,21 +131,21 @@ function _M.set_by_route(route, api_ctx)
 
     local up_conf = api_ctx.matched_upstream
     if not up_conf then
-        return 500, "missing upstream configuration in Route or Service"
+        return 503, "missing upstream configuration in Route or Service"
     end
     -- core.log.info("up_conf: ", core.json.delay_encode(up_conf, true))
 
     if up_conf.service_name then
         if not discovery then
-            return 500, "discovery is uninitialized"
+            return 503, "discovery is uninitialized"
         end
         if not up_conf.discovery_type then
-            return 500, "discovery server need appoint"
+            return 503, "discovery server need appoint"
         end
 
         local dis = discovery[up_conf.discovery_type]
         if not dis then
-            return 500, "discovery " .. up_conf.discovery_type .. " is uninitialized"
+            return 503, "discovery " .. up_conf.discovery_type .. " is uninitialized"
         end
         local new_nodes = dis.nodes(up_conf.service_name)
         local same = upstream_util.compare_upstream_node(up_conf.nodes, new_nodes)
@@ -172,7 +172,7 @@ function _M.set_by_route(route, api_ctx)
 
     local nodes_count = up_conf.nodes and #up_conf.nodes or 0
     if nodes_count == 0 then
-        return 502, "no valid upstream node"
+        return 503, "no valid upstream node"
     end
 
     if nodes_count > 1 then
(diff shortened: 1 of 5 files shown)


membphis (approved):
LGTM

tokers:
Does the broken compatibility really  no matter?

moonming (changes requested):
why change this？

apisix/upstream.lua
```diff
@@ -131,21 +131,21 @@ function _M.set_by_route(route, api_ctx)
 
     local up_conf = api_ctx.matched_upstream
     if not up_conf then
-        return 500, "missing upstream configuration in Route or Service"
+        return 503, "missing upstream configuration in Route or Service"
```
moonming, line 134: no test case cover

moonming (changes requested):

Firstsawyou:
> why change this？

I think the use of the 500 response code in the code is unreasonable, because this is not caused by a code program error, but because there is no upstream available. So it is more reasonable to use the 503 response code.

Firstsawyou:
> Does the broken compatibility really no matter?

This will only affect some test cases.

moonming:
> > why change this？
> 
> 
> 
> I think the use of the 500 response code in the code is unreasonable, because this is not caused by a code program error, but because there is no upstream available. So it is more reasonable to use the 503 response code.

why will be no upstream available？

moonming:
> > Does the broken compatibility really no matter?
> 
> 
> 
> This will only affect some test cases.

of course not.  This is too dangerous, you don’t know the scope of the impact

tokers:
You don't know whether people rely on the status code. @Firstsawyou

Firstsawyou:
> You don't know whether people rely on the status code. @Firstsawyou

got it, this is something to consider.
But these 500 response codes are inappropriate for use here, we need to fix them.
This problem needs to be solved sooner or later, and it would be better for us to solve these problems earlier.

apisix/upstream.lua
Firstsawyou: updated

tokers:
> > You don't know whether people rely on the status code. @Firstsawyou
> 
> got it, this is something to consider.
> But these 500 response codes are inappropriate for use here, we need to fix them.
> This problem needs to be solved sooner or later, and it would be better for us to solve these problems earlier.

we call them historical problems, we have to be compatible with them. Even if you'd like to change it, sending a statement and claiming that we'll change these status codes from `500` to `503` once a.b version is released, which gives some time to user to do the change.

moonming:
@Firstsawyou  Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

Firstsawyou:
> @Firstsawyou Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

Okay.

tokers:
So i think this PR can be closed now, just reopen it when necessary.

PR #3396, opened 2021-01-22
change: global rules should not be executed on the internal api

nic-chen:

global rules should not be executed on the internal api, currently it will run `access` and `rewrite` of plugins in global rule, so we need to change it.

### Pre-submission checklist:


apisix/init.lua
```diff
@@ -324,6 +324,28 @@ function _M.http_access_phase()
...
+    local uri = api_ctx.var.uri
+    if local_conf.apisix and local_conf.apisix.delete_uri_tail_slash then
+        if str_byte(uri, #uri) == str_byte("/") then
+            api_ctx.var.uri = str_sub(api_ctx.var.uri, 1, #uri - 1)
+            core.log.info("remove the end of uri '/', current uri: ",
+                          api_ctx.var.uri)
+        end
+    end
+
+    if router.api.has_route_not_under_apisix() or
```
spacewander, line 336: Maybe we can add a configuration in the `config.yaml` for it?

spacewander:
We need a test case for it.

nic-chen:
> We need a test case for it.

we have [test case](https://github.com/apache/apisix/pull/3396/files#diff-bdc9989c1ae307a5194244a61cec93ef98b83b41c0e0ae0180564cfb4bfb1981R123-R146) for it, : )

apisix/init.lua
nic-chen: @spacewander  do you mean a switch for it ?

spacewander: Yes

t/node/global-rule.t
```diff
@@ -120,7 +120,34 @@ GET /hello
...
+=== TEST 6: not limited
+--- request
+GET /apisix/nginx_status
+--- error_code: 200
+--- no_error_log
+[error]
+
+
+
+=== TEST 7: not limited
```
tokers, line 132: Why add a same test case?

apisix/init.lua
nic-chen: @spacewander 
I thought about it again. There shouldn't have a switch here. If skip the `if` branch and continue execution, a 404 will occur because the route does not exist.

And I think if need to configure the plugin for the internal api, it is more appropriate to configure a route and then configure the plugin instead of using the global rule. 

What do you think ? Looking forward to your opinion. Thanks.

t/node/global-rule.t
nic-chen: @tokers  because it's testing the `limit-count` plugin here.

tokers (approved):

apisix/init.lua
spacewander: Better to provide a switch to execute the global rule before matching internal API or don't execute the global rule. It is not skipped the internal API matched.

> And I think if need to configure the plugin for the internal api, it is more appropriate to configure a route and then configure the plugin instead of using the global rule.

We can't configure a plugin for internal API, this is the real problem.

nic-chen: @spacewander  Thanks for for the explanation. I know that.., emm, will update later.

spacewander: OK. We can wait for it.

spacewander (changes requested):

apisix/init.lua
```diff
@@ -325,8 +325,29 @@ function _M.http_access_phase()
...
+            core.log.info("remove the end of uri '/', current uri: ",
+                          api_ctx.var.uri)
+        end
+    end
+
+    local matched_internal_api = false
+    if router.api.has_route_not_under_apisix() or
+        core.string.has_prefix(uri, "/apisix/")
+    then
+        matched_internal_api = router.api.match(api_ctx)
```
spacewander, line 341: The internal API will call `exit` inside so there is not change to execute the global rule.
Maybe we can refactor and exact the global rule code, then run it before https://github.com/apache/apisix/blob/c895cdf4ff008c13dec8197e9f0a0407be144517/apisix/api_router.lua#L124

conf/config-default.yaml
```diff
@@ -89,6 +89,8 @@ apisix:
       role: viewer
 
   delete_uri_tail_slash: false    # delete the '/' at the end of the URI
+  global_rule_skip_internal_api: true    # does not run global rule in internal apis
```
spacewander, line 92: Can we add test when global_rule_skip_internal_api is false?

apisix/init.lua
nic-chen: got it, thanks!

conf/config-default.yaml
spacewander: We need to add test when global_rule_skip_internal_api is false.

nic-chen: > We need to add test when global_rule_skip_internal_api is false.

sure, test failed now, submit it after solving it.

nic-chen: @spacewander  teat case added. please help review when you have time, thanks.

spacewander (approved):

Issue #2757, opened 2020-11-16
feat: deploy ChaosMesh on APISIX, to simulate more faults


Yiyiyimu:
# Proposal

### Background
Nowadays, we have unit tests, integration tests, and e2e tests, to ensure the fault tolerance of APISIX. But there are still some problems, like network delay and CPU stress, that have not covered by the above tests. Thus, it would be a better idea to introduce chaos engineering, to simulate different types of faults, and test the performance of APISIX in these circumstances.

To deploy chaos engineering, [ChaosMesh](https://chaos-mesh.org/) could be a good choice for us. There are several benefits above other chaos engineering tools:

1. ChaosMesh is a CNCF sandbox project and has quite an active community, which ensures the project would be better and we could get help when needed.
2. ChaosMesh support Github Actions, so when we set up the workflow of this integration, it would be easy to do the test in our daily working
3. ChaosMesh currently supports most types of different chaos for now and is supporting more. Although we might not need that much for now, it is a good point when we decide to test more with it.
BTW, chaos types ChaosMesh [supports for now](https://chaos-mesh.org/docs/chaos_experiments)(Nov.16, 2020) includes pod chaos, network chaos, stress chaos, io chaos, time chaos, kernel chaos, HTTP chaos, and DNS chaos.

### TODO

Following the principles of chaos engineering, there are two main parts we need to care about: 1. what should we test and 2. how to prove the correctness after chaos injection.

As for what we got for now, the current problems we encounter and need to simulating are:

1. the connection with etcd is unstable
2. etcd failure
3. problems when cpu/memory/disk stressed out

And the method to test correctness including:

~1. error log of Nginx and APISIX
2. whether cpu/memory use of APISIX is abnormally high
3. whether wrk benchmarking would fail~

After some more investigation of how people make use of chaos engineering, to get how things going after certain chaos takes effect, it would be better to use Prometheus/Grafana to plot the metrics of APISIX performance, rather than only focusing on nginx logs. Also, since chaos is more about mocking problems facing in production, directly using monitoring tools could let us get what users are facing.

To use Prometheus, we need a demo to run basic functions of APISIX, like a certain amount of traffic, and new rules set by a certain time interval. It seems we do not have that kind of demo, so maybe I plan to write a simple script to implement these features. 

[...]

PR #3404, opened 2021-01-24
feat: chaos test on route could still works when etcd is down

Yiyiyimu:
Related to #2757

Use chaos mesh to generate the situation, when etcd is all killed, route could still works.

**TODO**
- [ ] add test on when etcd restart, how's it going

### Pre-submission checklist:


.github/workflows/chaos.yml
```diff
@@ -0,0 +1,61 @@
...
+        kubectl create -f example/deployment.yaml
+        while [[ $(kubectl get pods -l app=etcd-operator -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do echo "Waiting for pod running" && sleep 10; done
+        kubectl create -f example/example-etcd-cluster.yaml
+        while [[ $(kubectl get pods -l app=etcd -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True True True" ]]; do echo "Waiting for pod running" && sleep 10; done
+
+    - name: Deploy APISIX
+      run: |
+        echo '
+        dns_resolver:
+          - 10.96.0.10
```
spacewander, line 44: Why chose this IP?

kubernetes/service.yaml
```diff
@@ -36,8 +36,10 @@ spec:
...
-      type: NodePort
-      externalTrafficPolicy: Local
-      # sessionAffinity: None
+  selector:
+    app: apisix-gw
+  type: NodePort
+  externalTrafficPolicy: Local
+  # sessionAffinity: None
+  externalIPs:
+    - 10.253.0.11
```
spacewander, line 45: Why chose this IP?

t/chaos-test/go.mod
```diff
@@ -0,0 +1,5 @@
+module chaos
```
spacewander, line 1: Better to use a uri contains apisix, like
github.com/apache/apisix/t/chaos.

And `t/chaos/` might be a better name, as all files under `t` are tests.

t/chaos-test/kill-etcd_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"net/http"
+	"os/exec"
+	"testing"
+
+	"github.com/gavv/httpexpect/v2"
+)
+
+var (
+	token = "edd1c9f034335f136f87ad84b625c8f1"
+	host  = "http://10.253.0.11:9080"
```
spacewander, line 30: Need a comment to explain why this IP is chosen.

.github/workflows/chaos.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: setup go
+      uses: actions/setup-go@v2.1.3
+      with:
+        go-version: '1.13'
+
+    - name: Creating kind cluster
+      uses: helm/kind-action@v1.0.0-rc.1
```
tokers, line 17: Does these non-official and non-Apache actions can be used now?

t/chaos-test/kill-etcd.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: kill-etcd
+  namespace: chaos-testing
+spec:
+  action: pod-kill
+  mode: all
+  value: ''
+  selector:
+    labelSelectors:
+      'app': 'etcd'
```
tokers, line 29: It's strange why the key needs to be wrapped by the single quotes.

t/chaos-test/kill-etcd_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestGetSuccessWhenEtcdKilled(t *testing.T) {
+	e := httpexpect.New(t, host)
+
+	// check if everything works
+	setRoute(e, http.StatusOK)
+	getRoute(e, http.StatusOK)
+
+	// TODO: use client-go
+	// apply chaos to kill all etcd pods
+	_, err := exec.Command("kubectl apply kill-etcd.yaml").CombinedOutput()
```
tokers, line 106: We'd better to use https://github.com/gruntwork-io/terratest

t/chaos-test/kill-etcd_test.go
```diff
@@ -0,0 +1,115 @@
...
+		ExpectStatus: expectStatus,
+	})
+}
+
+func TestGetSuccessWhenEtcdKilled(t *testing.T) {
+	e := httpexpect.New(t, host)
+
+	// check if everything works
+	setRoute(e, http.StatusOK)
+	getRoute(e, http.StatusOK)
```
moonming, line 102: We should add check such as nginx error log and Prometheus

t/chaos-test/kill-etcd_test.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"net/http"
+	"os/exec"
+	"testing"
+
+	"github.com/gavv/httpexpect/v2"
+)
+
+var (
+	token = "edd1c9f034335f136f87ad84b625c8f1"
```
juzhiyuan, line 29: Could we use a public file to store those constants?

.github/workflows/chaos.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Deploy APISIX
+      run: |
+        echo '
+        dns_resolver:
+          - 10.96.0.10
+        etcd:
+          host:
+            - "http://etcd-cluster-client.default.svc.cluster.local:2379"
+        ' > conf/config.yaml
+        example/rbac/create_role.sh
```
idbeta, line 49: This line shouldn't be needed here.

.github/workflows/chaos.yml
```diff
@@ -0,0 +1,61 @@
...
+        kubectl create configmap apisix-gw-config.yaml --from-file=../conf/config.yaml
+        kubectl apply -f deployment.yaml
+        kubectl apply -f service.yaml
+        while [[ $(kubectl get pods -l app=apisix-gw -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True True" ]]; do echo "Waiting for pod running" && sleep 10; done
```
gxthrj, line 53: If APISIX can’t start up for some reason, here’s an endless loop.

Yiyiyimu: Thanks! Changed to exit when couldn't start up in 300s

.github/workflows/chaos.yml
Yiyiyimu: Thanks! Changed to `bash example/rbac/create_role.sh`

t/chaos-test/kill-etcd_test.go
Yiyiyimu: Maybe later when we have more test files, we could make them public? Since there are not too many constants here

t/chaos-test/kill-etcd.yaml
Yiyiyimu: Thanks! Lint tool helped me changing it to doubel quotes

.github/workflows/chaos.yml
Yiyiyimu: Thanks for the remind! Changed to a minikube startup script

t/chaos-test/go.mod
Yiyiyimu: Thanks! Fixed

.github/workflows/chaos.yml
```diff
@@ -0,0 +1,61 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: setup go
+      uses: actions/setup-go@v2.1.3
+      with:
+        go-version: '1.13'
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks! My bad..

.github/workflows/chaos.yml
Yiyiyimu: Thanks! Changed to dynamically get dns ip

kubernetes/service.yaml
Yiyiyimu: I could not find method to randomly choose external IP, so I just randomly choose one. Is there any accustomed choice?

tokers: Well, Do we really need the external IP?

tokers: Why not use the NodePort service?

Yiyiyimu: Thanks for the suggestions! Changed to port forward

t/chaos-test/kill-etcd_test.go
Yiyiyimu: Changed to port forward so it is changed to localhost

t/chaos-test/kill-etcd_test.go
Yiyiyimu: Thanks for the suggestion! But the shell module seems a bit redundant (need to create a struct each time run a command), so I write a function for error check.

t/chaos-test/kill-etcd_test.go
Yiyiyimu: Thanks! Added

t/chaos/kill-etcd_test.go
```diff
@@ -51,6 +52,7 @@ func caseCheck(tc httpTestCase) {
 	}
 
 	if req == nil {
+		fmt.Printf("%#v", tc)
```
tokers, line 59: Is this for debugging?

Yiyiyimu: yes

t/chaos/kill-etcd_test.go
```diff
@@ -0,0 +1,173 @@
...
+	t.Run("error log not contains etcd error", func(t *testing.T) {
+		errorLog := runCommand(t, fmt.Sprintf("kubectl exec -it %s -- cat logs/error.log", podName))
+		g.Expect(strings.Contains(errorLog, "failed to fetch data from etcd")).To(BeFalse())
+	})
+
+	// TODO: use client-go
+	// apply chaos to kill all etcd pods
+	t.Run("kill all etcd pods", func(t *testing.T) {
+		_ = runCommand(t, "kubectl apply -f kill-etcd.yaml")
+		time.Sleep(30)
```
idbeta, line 158: have to sleep so long?

t/chaos-test/kill-etcd_test.go
tokers: OK.

t/chaos/kill-etcd_test.go
tokers: We'd better to use log.

t/chaos/kill-etcd_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// check if everything works
+	setRoute(e, http.StatusCreated)
+	getRoute(e, http.StatusOK)
+	testPrometheusEtcdMetric(e, 1)
+
+	// run in background
+	go func() {
+		for i := 1; ; i++ {
+			go getWithoutTest(g, "/hello", map[string]string{"Host": "foo.com"})
```
Yiyiyimu, line 213: **HELP NEEDED!!**

Here I try to visit the route in the same frequency, so I could check if it keep the same before and after etcd shutdown.

For some reason, using bare `net/http` pkg to access the route would always return "404 Route not found", although the route is created, and accessing the same URL with `httpexpect` could work(in line 207).

While the reason I don't want to use `httpexpect` this place, is due to it would print lots of log each time a request sent and it would make test log into a mess. So a bare HTTP get is needed.

Yiyiyimu: Hi @juzhiyuan could you offer some help on this

Yiyiyimu:
Currently when etcd is killed, apisix would keep print out `config_etcd.lua:530: failed to fetch data from etcd: connection refused`, and the speed is 2 log per second per core. It could be fixed after https://github.com/api7/lua-resty-etcd/issues/111 (return error code when no healthy etcd available) got fixed.

t/chaos/kill-etcd_test.go
Yiyiyimu: Since using `httpexpect` would print log each time a request sent, I turn the route visit frequency from 10 times/s to 1 times/s to reduce the log printing. The reason not to reduce visit duration is that, when frequency is larger than 1 time/s, some visit would return `"503 Service Unavailable"`, although I don't think that's reasonable since the frequency is still very low compared to the real usage.

t/chaos/kill-etcd_test.go
```diff
@@ -0,0 +1,257 @@
...
+		E:            e,
+		Method:       http.MethodDelete,
+		Path:         "/apisix/admin/routes/1",
+		Headers:      map[string]string{"X-API-KEY": token},
+		ExpectStatus: expectStatus,
+	})
+}
+
+/*
+// not working, don't know why
```
spacewander, line 126: Can we remove the commented code?

spacewander (approved):

t/chaos/kill-etcd_test.go
Yiyiyimu: Sure I'll leave that to a new PR

t/chaos/kill-etcd_test.go
Yiyiyimu: Thanks I directly delete it 🤣

t/chaos/kill-etcd_test.go
Yiyiyimu: Commented code removed, and the question is raised in https://github.com/apache/apisix-dashboard/issues/1388

t/chaos/kill-etcd_test.go
Yiyiyimu: The question is raised in https://github.com/apache/apisix-dashboard/issues/1388

tokers:
@Yiyiyimu It seems in this case you kill all etcd members. What about just killing the leader (will recover overtime), killing the follower (no issue), kill majority members (cluster will not recover)?

Yiyiyimu:
> @Yiyiyimu It seems in this case you kill all etcd members. What about just killing the leader (will recover overtime), killing the follower (no issue), kill majority members (cluster will not recover)?

Hi @tokers, I think in that case, that's more about test etcd itself, but not apisix. Besides, I don't think we should test etcd in this way, since that's what we make sure how raft and etcd works.

If we test how apisix performs, like when we kill the leader, in a rare case (because the sync is so fast), we could find maybe one new route is missing. But the test case is so unstable, so it's maybe not so reasonable

t/chaos/go.sum
```diff
@@ -0,0 +1,97 @@
+github.com/ajg/form v1.5.1 h1:t9c7v8JUKu/XxOGBU0yjNpaMloxGEJhUkqFRq0ibGeU=
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks for remind! License all checked and eligible

t/chaos/kill-etcd_test.go
```diff
@@ -0,0 +1,240 @@
...
+	setRoute(e, http.StatusCreated)
+
+	// to avoid route haven't been set yet
+	time.Sleep(1 * time.Second)
+	getRoute(e, http.StatusOK)
+	testPrometheusEtcdMetric(e, 1)
+
+	// run in background
+	go func() {
+		for i := 1; ; i++ {
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks! fixed

membphis (approved):

PR #3420, opened 2021-01-26
feat: allow to pass SNI in HTTPS proxy

### LINKED ISSUE #2988: Carry SNI while doing SSL handshaking with upstream
So far we don't enable the [SNI extension](https://en.wikipedia.org/wiki/Server_Name_Indication) when communicating with upstream. If upstream server depends on SNI to select proper certificate, then the TLS/SSL handshaking will fail. So we should carry the SNI extension.

Nginx has a directive named `proxy_ssl_server_name`.

```
Syntax: proxy_ssl_server_name on | off;
proxy_ssl_server_name off;
http, server, location
```
Enables or disables passing of the server name through TLS Server Name Indication extension (SNI, RFC 6066) when establishing a connection with the proxied HTTPS server.

So we can add this directive into `apisix/cli/ngx_tpl.lua`.

spacewander:
Fix #2988
### Pre-submission checklist:


Firstsawyou:
LGTM.

tokers (approved):

membphis (approved):

PR #3415, opened 2021-01-25
feat: support to communicate with etcd by TLS in command line

tokers:
### What this PR does / why we need it:

This PR implements the SSL handshaking to etcd cluster in the APISIX command line tool, and it will obey the TLS configuration in `config.yaml`.

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [x] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [x] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #3467, opened 2021-02-01
docs: sync install-dependencies.md & polish it

spacewander:
Signed-off-by: spacewander <spacewanderlzx@gmail.com>

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #3466, opened 2021-02-01
feat: release 2.3

spacewander:
Signed-off-by: spacewander <spacewanderlzx@gmail.com>

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #3472, opened 2021-02-01
chore: fix 'make deps'

Yiyiyimu:

related to #3415 

### Pre-submission checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,7 @@ help: default
 .PHONY: deps
 deps: default
 ifeq ($(LUAROCKS_VER),luarocks 3.)
-ifeq ("$(wildcard ~/.luarocks)", "")
-	mkdir ~/.luarocks || true
-endif
+	mkdir -p ~/.luarocks || true
 	luarocks config variables.OPENSSL_LIBDIR $(addprefix $(OR_PREFIX), openssl/lib)
 	luarocks config variables.OPENSSL_INCDIR $(addprefix $(OR_PREFIX), openssl/include)
 	luarocks install rockspec/apisix-master-0.rockspec --tree=deps --only-deps --local
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,9 @@ help: default
 .PHONY: deps
 deps: default
 ifeq ($(LUAROCKS_VER),luarocks 3.)
+ifeq ("$(wildcard ~/.luarocks)", "")
 	mkdir ~/.luarocks || true
+endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -55,7 +55,9 @@ help: default
 .PHONY: deps
 deps: default
 ifeq ($(LUAROCKS_VER),luarocks 3.)
+ifeq ("$(wildcard ~/.luarocks)", "")
```
spacewander, line 58: I think calling `mkdir ~/.luarocks` even it exists doesn't do any harm?

spacewander (changes requested):

Makefile
Yiyiyimu: It would leave an error message, which confused me at first before I realize it does no harm

spacewander: @Yiyiyimu 
I think you are making something wrong. The `make deps` only install dependencies locally. There is no need to call `sudo`.

Yiyiyimu: Got it I installed luarocks with `apt`, so when it changes luarocks config, it needs sudo privilege. I'll change to the method mentioned in #3467 and see if it works.

spacewander: Use `mkdir -p ~/.luarocks` would be better.

Yiyiyimu: Yeah that would be better! Thanks!

And I made a mistake yesterday. No matter how we install luarocks, the config would always write to `/usr/local/etc/luarocks`, so we all need sudo permission. Is that the same on your side?

spacewander: What about:
```diff
 deps: default
 ifeq ($(LUAROCKS_VER),luarocks 3.)
        mkdir ~/.luarocks || true
-       luarocks config variables.OPENSSL_LIBDIR $(addprefix $(OR_PREFIX), openssl/lib)
+       luarocks config --local variables.OPENSSL_LIBDIR $(addprefix $(OR_PREFIX), openssl/lib)
        luarocks config variables.OPENSSL_INCDIR $(addprefix $(OR_PREFIX), openssl/include)
        luarocks install rockspec/apisix-master-0.rockspec --tree=deps --only-deps --local
 else
```

Yiyiyimu: It's weird... I installed lua5.3 and configure luarocks to it this morning to do some stuff, and found out it changes some config  that even though I re-configure luarocks to luajit from openresty, it still tries to find lua5.3.

So I remove everything about luarocks, and re-install it with the script. Now this time, it would write config to `~/.luarocks` without the `--local` flag.

I'm not so sure right now. If you could not reproduce this problem, I'll close it

Yiyiyimu:
Fixed in v2.3 #3466. Closed.

PR #3493, opened 2021-02-02
feat: smooth weighted round-robin

redynasc:

 When using the now roundrobin balancer, when the greatest common divisor of the node 's weight is small, the requests is not 
 smooth
for example,the upstream like below:
{
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:1980": 21,
            "127.0.0.1:1981": 11
}
The 11th request can hit  "127.0.0.1:1981"

### Pre-submission checklist:


apisix/schema_def.lua
```diff
@@ -323,7 +323,7 @@ local upstream_schema = {
         type = {
             description = "algorithms of load balancing",
             type = "string",
-            enum = {"chash", "roundrobin", "ewma", "least_conn"}
+            enum = {"chash", "roundrobin", "ewma", "least_conn", "swrr"}
```
tokers, line 326: I'm inclined to optimize the current roundrobin loadbalancer, not creating a new one, the `roundrobin` and `swrr` are same in essence.

spacewander: This implementation is more fair when `(max weight - second max weight) / gcd` is large. 

@redynasc 
Maybe we need to put this as part of the original rr algorithm?
We can choice this one if `(max weight - second max weight) / gcd > 5`.
Thanks!

apisix/balancer/swrr.lua
```diff
@@ -0,0 +1,109 @@
...
+--
+local core = require("apisix.core")
+local nkeys = core.table.nkeys
+local pairs = pairs
+local next = next
+local _M = {}
+
+local function find(self, tried)
+    local number = #self.nodes
+    if number == 1 then
```
spacewander, line 25: We don't need this as we only create balancer when #nodes > 1

apisix/schema_def.lua
spacewander: The magical number `5` can be configured, so if we choice 0 the `swrr` will be used instead.

PR #3530, opened 2021-02-05
change: default to cache DNS record according to the TTL

spacewander:
Since lua-resty-dns-client provides an internal cache, we don't need to
cache it twice.

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -279,6 +279,12 @@ Please modify "admin_key" in conf/config.yaml .
         end
     end
 
+    if yaml_conf.apisix.dns_resolver_valid then
+        if tonumber(yaml_conf.apisix.dns_resolver_valid) == nil then
+            util.die("apisix->dns_resolver_valid should be a number")
+        end
+    end
+
     -- Using template.render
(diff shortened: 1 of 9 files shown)


apisix/cli/ops.lua
```diff
@@ -279,6 +279,12 @@ Please modify "admin_key" in conf/config.yaml .
         end
     end
 
+    if yaml_conf.apisix.dns_resolver_valid then
+        if tonumber(yaml_conf.apisix.dns_resolver_valid) == nil then
```
membphis, line 283: Can we use `JSON Schema` to check configuration values?

membphis (approved):
lgtm

apisix/cli/ops.lua
tokers: Seems that the whole `config.yaml` doesn't have the json schema.

membphis: I think it is time to do this in this way now

tokers: It's more proper to do it in another PR.

lilien1010 (approved):

PR #3542, opened 2021-02-07
fix: make deps error: file changed as we read it

Yiyiyimu:

There are error `tar: .: file changed as we read it` when do `make deps`, and the reason is the argument `--exclude $(RELEASE_SRC).tgz` seems not working, so that the tarball is changing as it's being built. 

So move it to a separate path to avoid this problem.

Some other problems would be fixed with https://github.com/apache/apisix/commit/63f779aeae7a4cfdc3baa5c6dad955ef139719f8

### Pre-submission checklist:


Makefile
```diff
@@ -238,13 +239,7 @@ release-src:
 	--exclude t \
 	--exclude utils \
 	--exclude release \
-	--exclude $(RELEASE_SRC).tgz \
```
spacewander, line 241: We already switch to a whitelist solution: https://github.com/apache/apisix/pull/3466/files

spacewander (changes requested):

Makefile
Yiyiyimu: Cool I didn't realize whitelist could solve this problem.

PR #3551, opened 2021-02-07
feat(zipkin): support b3 req header

### LINKED ISSUE #1125: feature: parse b3 single header for zipkin plugin
implement https://github.com/openzipkin/b3-propagation/issues/21

spacewander:
Fix #1125

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/zipkin.lua
+++ b/apisix/plugins/zipkin.lua
@@ -127,8 +127,8 @@ function _M.rewrite(plugin_conf, ctx)
         err, trace_id, request_span_id, sampled, parent_span_id = parse_b3(b3)
 
         if err then
-            core.log.warn("invalid b3 header: ", b3, ", ignored: ", err)
-            return
+            core.log.error("invalid b3 header: ", b3, ", ignored: ", err)
+            return 400
         end
 
         if sampled == "d" then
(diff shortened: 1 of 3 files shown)


tokers (approved):

apisix/plugins/zipkin.lua
```diff
@@ -88,29 +110,58 @@ function _M.rewrite(plugin_conf, ctx)
...
+    local b3 = headers["b3"]
+    if b3 then
+        -- don't pass b3 header by default
+        core.request.set_header(ctx, "b3", nil)
+
+        local err
+        err, trace_id, request_span_id, sampled, parent_span_id = parse_b3(b3)
+
+        if err then
+            core.log.warn("invalid b3 header: ", b3, ", ignored: ", err)
```
membphis, line 130: use `error` is better here

spacewander: @membphis 
Raise `error` in a request may be considered a bug of APISIX (as it will provide 500 and backtrack).
Maybe we can return 400 to reject it?

tokers: I think `400` is better, since an invalid header is always caused by client.

membphis: > I think `400` is better, since an invalid header is always caused by client.

agree this too. 

> Raise error in a request may be considered a bug of APISIX (as it will provide 500 and backtrack).

I means `core.log.error(****)` ^_^

membphis (approved):

membphis:
@Yiyiyimu Chaos Test failed, do you have time to look at this PR?

PR #3595, opened 2021-02-15
fix: should assign value to `api_ctx.global_rules` before running global rules

nic-chen:

this bug introduced by #3396.

should assign value to `api_ctx.global_rules` before running global rules, otherwise if a plugin exits in the access phase, global rules will not be able to run in the subsequent phases.

### Pre-submission checklist:

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -297,7 +297,6 @@ function _M.http_access_phase()
     router.router_http.match(api_ctx)
 
     -- run global rule
-    api_ctx.global_rules = router.global_rules
     plugin.run_global_rules(api_ctx, router.global_rules, "access")
 
     local route = api_ctx.matched_route
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -297,8 +297,8 @@ function _M.http_access_phase()
     router.router_http.match(api_ctx)
 
     -- run global rule
-    plugin.run_global_rules(api_ctx, router.global_rules, "access")
     api_ctx.global_rules = router.global_rules
+    plugin.run_global_rules(api_ctx, router.global_rules, "access")
 
     local route = api_ctx.matched_route
     if not route then
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -689,6 +689,10 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
         local orig_conf_version = api_ctx.conf_version
         local orig_conf_id = api_ctx.conf_id
 
+        if phase_name == "access" then
+            api_ctx.global_rules = global_rules
+        end
+
         local plugins = core.tablepool.fetch("plugins", 32, 0)
         local values = global_rules.values
         for _, global_rule in config_util.iterate_values(values) do
--- a/apisix/api_router.lua
+++ b/apisix/api_router.lua
@@ -125,7 +125,6 @@ function fetch_api_router()
                             if not skip_global_rule then
                                 plugin_mod.run_global_rules(api_ctx,
                                     apisix_router.global_rules, "access")
-                                api_ctx.global_rules = apisix_router.global_rules
                             end
 
                             code, body = route.handler(api_ctx)
--- a/t/node/global-rule.t
+++ b/t/node/global-rule.t
@@ -262,7 +262,7 @@ apisix:
 plugins:
   - response-rewrite
   - uri-blocker
-  - node-status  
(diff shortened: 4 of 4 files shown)


moonming:
Is this a bug?  Which PR introduced it?

apisix/init.lua
```diff
@@ -297,8 +297,8 @@ function _M.http_access_phase()
     router.router_http.match(api_ctx)
 
     -- run global rule
-    plugin.run_global_rules(api_ctx, router.global_rules, "access")
     api_ctx.global_rules = router.global_rules
```
spacewander, line 300: Do it in run_global_rules?
Need to update the api_router.lua too

nic-chen:
> Is this a bug? Which PR introduced it?

yes, it's a bug. it's introduced by #3396

apisix/init.lua
nic-chen: @spacewander  
Thanks for reviewing.
I still have some doubts. The original code also assign value to `api_ctx.global_rules` after `run_plugin`, but it does not cause a bug like this. .

nic-chen: > @spacewander
> Thanks for reviewing.
> I still have some doubts. The original code also assign value to `api_ctx.global_rules` after `run_plugin`, but it does not cause a bug like this. .

I re-checked the original code, it filtered plugins of global rule and assigned them to `api_ctx.plugins`, that is why..

nic-chen:
@membphis @tokers @moonming   please have a look when you have time, thanks.

tokers:
@nic-chen Could you tell me where we use `api_ctx.global_rules`?

nic-chen:
> @nic-chen Could you tell me where we use `api_ctx.global_rules`? I saw the assignment but not found the usage.

use in:
https://github.com/apache/apisix/blob/master/apisix/init.lua#L481

and define in:
https://github.com/apache/apisix/blob/master/apisix/plugin.lua#L685-L714

you could also have a look at the prev PR #3396 for more detail.

nic-chen:
> yes, it's a bug. it's introduced by #3396

Actually #3396 is not the real cause of the bug, but it exposes the bug

tokers:
> > @nic-chen Could you tell me where we use `api_ctx.global_rules`? I saw the assignment but not found the usage.
> 
> use in:
> https://github.com/apache/apisix/blob/master/apisix/init.lua#L481
> 
> and define in:
> https://github.com/apache/apisix/blob/master/apisix/plugin.lua#L685-L714
> 
> you could also have a look at the prev PR #3396 for more detail.

Got it.

apisix/plugin.lua
```diff
@@ -689,6 +689,10 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
         local orig_conf_version = api_ctx.conf_version
         local orig_conf_id = api_ctx.conf_id
 
+        if phase_name == "access" then
```
tokers, line 692: Why not assign it before you call `run_global_rules` in the access phase, instead of check the phase every time the `run_global_rules` are called.

membphis (approved):

apisix/plugin.lua
```diff
@@ -689,6 +689,10 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
         local orig_conf_version = api_ctx.conf_version
```
membphis, line 689: better style, we can fix it in a new PR

```lua
function _M.run_global_rules(api_ctx, global_rules, phase_name)
    if not global_rules or not global_rules.values
       and #global_rules.values == 0 then
        return
    end

    -- your code
end

apisix/plugin.lua
nic-chen: @tokers 
Putting it inside could make it clearer, because there is more than one entry. you could have look at @spacewander's suggestion: https://github.com/apache/apisix/pull/3595#discussion_r576153115
thanks.

tokers: OK.

apisix/plugin.lua
nic-chen: > better style, we can fix it in a new PR
> 
> ```lua
> function _M.run_global_rules(api_ctx, global_rules, phase_name)
>     if not global_rules or not global_rules.values
>        and #global_rules.values == 0 then
>         return
>     end
> 
>     -- your code
> end
> ```

OK

tokers (approved):

PR #3603, opened 2021-02-19
fix: use openssl111 in openresty dir in precedence

tokers:

We use luasec since 5399a31b, however, it uses the openssl libs and headers to compile itself, currently, we set two variables `$OPENSSL_LIBDIR` (e.g. `/usr/local/openresty/openssl/lib`) and `$OPENSSL_INCDIR` (e.g. `/usr/local/openresty/openssl/include`).

However, OpenResty 1.17.8 or higher version actually uses `openssl111` as the dir name of openssl. So in this PR, we detect the existence of dir `openssl111` and use it in precedence.

### Pre-submission checklist:

### DIFF
--- a/doc/install-dependencies.md
+++ b/doc/install-dependencies.md
@@ -39,6 +39,8 @@
 
 - On some platforms, installing LuaRocks via the package manager will cause Lua to be upgraded to Lua 5.3, so we recommend installing LuaRocks via source code. if you install OpenResty and its OpenSSL develop library (openresty-openssl-devel for rpm and openresty-openssl-dev for deb) via the official repository, then [we provide a script for automatic installation](../utils/linux-install-luarocks.sh). If you compile OpenResty yourself, you can refer to the above script and change the path in it. If you don't specify the OpenSSL library path when you compile, you don't need to configure the OpenSSL variables in LuaRocks, because the system's OpenSSL is used by default. If the OpenSSL library is specified at compile time, then you need to ensure that LuaRocks' OpenSSL configuration is consistent with OpenResty's.
 
+- If you are using OpenResty/1.17.8 or higher version, please installing openresty-openssl111-devel instead of openresty-openssl-devel.
+
 # CentOS 7
 
 ```shell
@@ -52,7 +54,6 @@ tar -xvf etcd-v3.4.13-linux-amd64.tar.gz && \
 sudo yum install yum-utils
 sudo yum-config-manager --add-repo https://openresty.org/package/centos/openresty.repo
 
-# install OpenResty and some compilation tools
 sudo yum install -y openresty curl git gcc openresty-openssl-devel unzip
 
 # install LuaRocks
--- a/doc/install-dependencies.md
+++ b/doc/install-dependencies.md
@@ -54,6 +54,7 @@ tar -xvf etcd-v3.4.13-linux-amd64.tar.gz && \
 sudo yum install yum-utils
 sudo yum-config-manager --add-repo https://openresty.org/package/centos/openresty.repo
 
+# install OpenResty and some compilation tools
 sudo yum install -y openresty curl git gcc openresty-openssl-devel unzip
 
 # install LuaRocks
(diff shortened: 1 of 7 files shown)


doc/install-dependencies.md
```diff
@@ -39,6 +39,8 @@
 
 - On some platforms, installing LuaRocks via the package manager will cause Lua to be upgraded to Lua 5.3, so we recommend installing LuaRocks via source code. if you install OpenResty and its OpenSSL develop library (openresty-openssl-devel for rpm and openresty-openssl-dev for deb) via the official repository, then [we provide a script for automatic installation](../utils/linux-install-luarocks.sh). If you compile OpenResty yourself, you can refer to the above script and change the path in it. If you don't specify the OpenSSL library path when you compile, you don't need to configure the OpenSSL variables in LuaRocks, because the system's OpenSSL is used by default. If the OpenSSL library is specified at compile time, then you need to ensure that LuaRocks' OpenSSL configuration is consistent with OpenResty's.
 
+- If you are using OpenResty/1.17.8 or higher version, please installing openresty-openssl111-devel instead of openresty-openssl-devel.
```
spacewander, line 42: We should warn for low version instead.
And also update the CI scripts.

tokers: You mean the low version of OpenResty?

Also, Do we still need to update the Chinese?

spacewander: We should install openresty-openssl111-devel by default and install openresty-openssl-devel for OpenResty 1.15.8.
And we need to update the Chinese docs.

membphis: if we only support `> openresty 1.17.8.1`, can it be easier?

membphis:
more docs needs to be updated:

https://github.com/apache/apisix/blob/master/doc/install-dependencies.md#centos-7

image

tokers:
> more docs need to be updated:
> 
> https://github.com/apache/apisix/blob/master/doc/install-dependencies.md#centos-7
> 
> image

Do you view the file on my branch? It was updated already.

membphis:
> Do you view the file on my branch? It was updated already.

https://github.com/apache/apisix/blob/22827880f46e50cc657371ed36a5ae76f412fa57/README.md#configure-and-installation

We need to update `README.md`. 

image

doc/install-dependencies.md
tokers: Yes, but we have a transition time, for now, we still need this.

tokers:
> > Do you view the file on my branch? It was updated already.
> 
> https://github.com/apache/apisix/blob/22827880f46e50cc657371ed36a5ae76f412fa57/README.md#configure-and-installation
> 
> We need to update `README.md`.
> 
> image

The change belongs to the dependencies installation, which is covered in the install-dependencies.md, Is there anything should also be change there?

utils/linux-install-openresty.sh
```diff
@@ -80,4 +81,23 @@ else
...
+verlt() {
+    [ "$1" = "$2" ] && return 1 || verlte $1 $2
+}
+
+sudo apt-get install "$openresty" lua5.1 liblua5.1-0-dev
+
+if [ "$OPENRESTY_VERSION" == "default" ]; then
+    sudo apt-get install openresty-openssl111-debug-dev
+else
+    verlt $OPENRESTY_VERSION 1.17.8.1
```
spacewander, line 97: We can simply check if the version is `1.15.8.2` so there is no need to comment out `set -euo pipefail`

tokers: Fixed.

spacewander (approved):

utils/linux-install-openresty.sh
```diff
@@ -80,4 +80,10 @@ else
     openresty="openresty-debug=$OPENRESTY_VERSION*"
 fi
 
-sudo apt-get install "$openresty" lua5.1 liblua5.1-0-dev openresty-openssl-debug-dev
+sudo apt-get install "$openresty" lua5.1 liblua5.1-0-dev
+
+if [ "$OPENRESTY_VERSION" == "1.15.8.2" ]; then
```
membphis, line 85: We will not support version 1.15.8.2 soon.

membphis (approved):

membphis:
> The change belongs to the dependencies installation, which is covered in the install-dependencies.md, Is there anything should also be change there?

all fine now. thanks for your explaining

PR #3617, opened 2021-02-20
fix: add `local require = require` for every lua file

Miss-you:
fix: add `local require = require` for every lua file
### Pre-submission checklist:

### DIFF
--- a/apisix/timers.lua
+++ b/apisix/timers.lua
@@ -14,6 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+local require = require
 local core = require("apisix.core")
 local process = require("ngx.process")
 local pairs = pairs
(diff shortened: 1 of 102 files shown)


spacewander:
The performance gain is trivial.
How much performance we can benefit from this change?

Miss-you:
> The performance gain is trivial.
> How much performance we can benefit from this change?

In all honesty, there is little performance improvement, just the customary change of some global variables to local variables

apisix/timers.lua
```diff
@@ -14,6 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+local require = require
```
starsz, line 17: Should keep the same if we want blank line at the beginning.

Firstsawyou:
Hi, @Miss-you 
The file conflicts, you need to resolve the conflict.

PR #3614, opened 2021-02-19
ci: use httpbin.org in chaos test to make CI works

Yiyiyimu:
Signed-off-by: yiyiyimu <wosoyoung@gmail.com>

### What this PR does / why we need it:
Temporary trial to make CI works. The problem would be fixed in #3613

### Pre-submission checklist:

* [ ] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #3613, opened 2021-02-19
fix: use localhost in chaos test

Yiyiyimu:
Signed-off-by: yiyiyimu <wosoyoung@gmail.com>

### What this PR does / why we need it:
Currently chaos test use `foo.com` and `bar.org` as host and upstream when creating routes. Changed to localhost to avoid unstable (actually failed) situations.

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [x] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #3623, opened 2021-02-21
ci: use httpbin deployed in kubernetes to avoid unstability

Yiyiyimu:

After #3614 got merged, chaos test is using `httpbin.org` as upstream, which could be unstable. Struggled some time with #3613 and suddenly found apisix inside kubernetes could not directly access localhost openresty...

So changed to deploy httpbin inside kubernetes, to make it stable

### Pre-submission checklist:

### DIFF
--- a/t/chaos/kill-etcd_test.go
+++ b/t/chaos/kill-etcd_test.go
@@ -68,9 +68,10 @@ func TestGetSuccessWhenEtcdKilled(t *testing.T) {
 
 	// check if everything works
 	setRoute(e, http.StatusCreated)
+	getRouteList(e, http.StatusOK)
 
 	// to avoid route haven't been set yet
-	time.Sleep(3 * time.Second)
+	time.Sleep(10 * time.Second)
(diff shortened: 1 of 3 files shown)


Yiyiyimu:
It seems httpbin deployed unstable...

t/chaos/kill-etcd_test.go
```diff
@@ -80,7 +83,7 @@ func TestGetSuccessWhenEtcdKilled(t *testing.T) {
 	}()
 
 	// wait 5 second to let first route access returns
```
juzhiyuan, line 85:
```suggestion
	// wait 3 seconds to let first route access returns
```

spacewander (approved):

Yiyiyimu:
@juzhiyuan PTAL. Rerun the CI three times and it all passed. I think it's stable for now.

Still don't know the reason why it failed, but it seems to prolong the waiting time could avoid errors. Would try to find the reason later.

moonming (approved):

PR #3648, opened 2021-02-23
[WIP] APISIX-2291 - AMQP Logger Initial Implementation

tashiya-menuri:
Work in progress PR. Added to get feedback on luarocks dependency and plugin design. Documentation and Unit test cases will be added once manual testing is fully completed.


dickens7:
I have implemented a version before but the performance is relatively poor, so I gave up using amqp in apisix and forwarded it to amqp proxy through http.

PR #3660, opened 2021-02-24
chore: update makefile install to install everything in conf

### LINKED ISSUE #3657: bug: luarocks not pack `conf/debug.yaml` in
Currently, `luarocks` would not pack `conf/debug.yaml` in, so if users install apisix with luarocks (and rpm/docker which rely on luarocks), they could not run debug in apisix.

Considering right now tarball would pack all `conf` directory in, we could do the same to luarocks.

Yiyiyimu:

fix #3657 

### Pre-submission checklist:


Makefile
```diff
@@ -143,9 +143,7 @@ install: default
 	$(INSTALL) -d /usr/local/apisix/
 	$(INSTALL) -d /usr/local/apisix/logs/
 	$(INSTALL) -d /usr/local/apisix/conf/cert
-	$(INSTALL) conf/mime.types /usr/local/apisix/conf/mime.types
-	$(INSTALL) conf/config.yaml /usr/local/apisix/conf/config.yaml
-	$(INSTALL) conf/config-default.yaml /usr/local/apisix/conf/config-default.yaml
+	$(INSTALL) conf/* /usr/local/apisix/conf/
```
tokers, line 146: The `cert` dir inside `conf/` is not necessary in the production environment, I think it should be excluded.

Yiyiyimu: We need ssl ca placeholders to start APISIX. 
See #3013

Makefile
```diff
@@ -143,9 +143,8 @@ install: default
 	$(INSTALL) -d /usr/local/apisix/
 	$(INSTALL) -d /usr/local/apisix/logs/
 	$(INSTALL) -d /usr/local/apisix/conf/cert
-	$(INSTALL) conf/mime.types /usr/local/apisix/conf/mime.types
-	$(INSTALL) conf/config.yaml /usr/local/apisix/conf/config.yaml
-	$(INSTALL) conf/config-default.yaml /usr/local/apisix/conf/config-default.yaml
+	$(INSTALL) conf/*.types /usr/local/apisix/conf/
+	$(INSTALL) conf/*.yaml /usr/local/apisix/conf/
```
spacewander, line 147: IMHO, it is better to specify the files explicitly to avoid mistake.

spacewander (changes requested):

Makefile
Yiyiyimu: Thanks for the suggestion! Since we need everything under `/conf`, as what we did when creating tarball, I think it's fine to use `*`. 
Besides, in `make install`, it seems people intend to this kind of style.

Makefile
tokers: OK, got it.

Makefile
spacewander: Unlike the Lua code, the configuration file are independent and should be list explicitly.

Yiyiyimu: Changed

spacewander (approved):

tokers (approved):

Issue #3340, opened 2021-01-19
How to do fuzz testing on Apache APISIX


idbeta:
### Why
The current test method of Apache APISIX pays more attention to normal input/output, in order to test the robustness of the program, fuzz testing can be used. There are many types of fuzzing, you can fuzzing the source code, you can also fuzzing the interface service. 
### Investigation process
I have investigated some popular fuzzing tools, the results are in the table below, hoping to help Apache APISIX improve its ability in this area.
| Tool name  | Program language | Actual combat | Conclusion |
| ------------- | ------------- | ------------- | ------------- |
| go-fuzz  | go  | To fuzz the go source code, you need to write a corresponding fuzz function for each function  | not applicable  |
| peach | java  | Crash itself Unhandled Exception: ... ...for parameter of type 'System.Object' of method 'Void SetValue(System.Object, System.Object)'  | not applicable  |
| SSRFmap | python  | Use Burp request to perform fuzz test on the specified module, there is no apisix, but you can customize the apisix module, which is difficult to get started  | not applicable  |
| PyJFuzz | python  | Support fuzzing the admin api, the principle is Construct Payload by automatically fuzzing the provided request body (json) and then sending the request to the target, like guided fuzzing  | **applicable**  |
| wfuzz  | python  | A lot of dictionary files are provided to traverse the content of the dictionary and send it to the target, but these dictionaries are not json, like blind fuzzing | **applicable**  |
| restler-fuzzer | python  | The support for linux is not good, dotnet is needed, I failed to install it  | not applicable  |
| boofuzz  | python  | Supports fuzzing of admin api, it uses the http request data example set in the case code to automatically fuzz all the request parameters to construct the payload, and then send the request to the target, like guided fuzzing  | **applicable**  |
| ffuf  | go  | Need to find a way to generate the payload first, and then use it to send. so you need to use another tool to generate the fuzzing data first  | not applicable  |
### How
From the table above, we can see that there are three tools suitable for APISIX, and their fuzzing focus is different,
**PyJFuzz** can fuzzing the request body, belongs to guided fuzzing,
 **wfuzz** provides many dictionaries for fuzzing the request body, belonging to blind fuzzing,
 **boofuzz** can fuzzing all request fields, belonging to blind fuzzing,
so my thoughts is to combine them to fuzzing APISIX, the whole process is as follows: 
[...]

PR #3667, opened 2021-02-25
test: add fuzzing case for the route with vars

idbeta:
relate #3340
fuzzing for the route with vars

### Pre-submission checklist:


idbeta:
ping @spacewander can you check the CI fail? It seems that the case is unstable.

spacewander:
It is the solved problem I said at the noon.

t/fuzzing/vars_route_test.py
```diff
@@ -0,0 +1,95 @@
...
+    session.fuzz()
+
+if __name__ == "__main__":
+    # before test
+    create_route()
+    r1 = check_process()
+    main()
+    # after test
+    boofuzz_log = cur_dir() + "/test.log"
+    apisix_errorlog = "~/work/apisix/apisix/logs/error.log"
```
spacewander, line 89: All the fuzzy scripts are similar. Can we share the common part of them?

spacewander (changes requested):

t/fuzzing/vars_route_test.py
```diff
@@ -0,0 +1,95 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import subprocess
+from public import *
+from boofuzz import *
```
moonming, line 22: “import * ” is not good style in python

moonming (changes requested):

t/fuzzing/vars_route_test.py
idbeta: yes, let me fixed it.

idbeta:
> It is the solved problem I said at the noon.
> Please merge master to make CI pass.

Thanks for remind.

t/fuzzing/serverless_route_test.py
```diff
@@ -100,12 +92,10 @@ def main():
     # before test
```
spacewander, line 92: Can we refactor this part as function and use `create_route` & `main` as arguments?

t/fuzzing/vars_route_test.py
idbeta: `from public import *` —— I have changed.
`from boofuzz import *` ——This is the official [usage](https://github.com/jtpereyda/boofuzz/blob/master/examples/http_simple.py),  I think it's ok.

t/fuzzing/serverless_route_test.py
idbeta: I got it.

t/fuzzing/public.py
```diff
@@ -43,3 +44,27 @@ def check_process():
...
+    session = Session(
+        target=Target(
+            connection=TCPSocketConnection("127.0.0.1", 9080, send_timeout=5.0, recv_timeout=5.0, server=False)
+        ),
+        fuzz_loggers=fuzz_loggers,
+        keep_web_open=False,
+    )
+    return session
+
+def runtest(create_route, run):
```
spacewander, line 60: Should be `run_test`?

spacewander (approved):

t/fuzzing/public.py
idbeta: sure

idbeta:
ping @moonming

t/fuzzing/public.py
```diff
@@ -20,14 +20,15 @@
 import subprocess
 import os
 import psutil
+from boofuzz import *
```
moonming, line 23: still not fixed

idbeta: This is the official usage. If it is changed to `import boofuzz`, the code will be very complicated. @moonming

moonming (approved):

Issue #3663, opened 2021-02-24
bug: script does not work


tzssangglass:
### Issue description
I used the script, but the script did not execute, see the following test case
```
use t::APISIX 'no_plan';

repeat_each(1);
log_level('info');
no_long_string();
no_root_location();

run_tests;

__DATA__

=== TEST 1: add services
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/services/1',
                 ngx.HTTP_PUT,
                 [[{
                    "name": "script_test",
                    "upstream": {
                        "nodes": {
                            "127.0.0.1:1980": 1
                        },
                        "type": "roundrobin"
                    }
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 2: route binding service and route has script
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/routes/1',
                 ngx.HTTP_PUT,
                 [[{
                    "service_id": 1,
                    "script": "local _M = {} \n function _M.access(api_ctx) \n ngx.log(ngx.INFO,\"hit access phase\") \n end \nreturn _M",
                    "uri": "/hello"
                }]]
                )

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]



=== TEST 3: hit route and trigger script
--- request
GET /hello
--- response_body
hello world
--- error_log eval
qr/loaded script_obj: \{"access":"function: 0x[\w]+"\}/

```

based on the documentation for the script, I think the script should be executed because it is bound to the route and neither the route nor the service has plugins configured

bugs may appear in https://github.com/apache/apisix/blob/16f017dfedebf8bf85ff8be9d48c05d088d21d0e/apisix/plugin.lua#L385

maybe this function should be preceded by code like this before return

```
    if route_conf.value.script then
        new_conf.value.script = route_conf.value.script
    end
```

### Environment

* apisix version (cmd: `apisix version`): greater than 2.2, less than 2.3
* OS (cmd: `uname -a`): Centos 7
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): openresty/1.19.3.1
[...]

PR #3678, opened 2021-02-25
fix: script does not work when the route is bound to a service.

tzssangglass:

fix: #3663
### Pre-submission checklist:


tzssangglass:
@spacewander @membphis CI is ok, but I don't koow why PR title is solecistic

Firstsawyou:
Your PR title needs to look like this: `fix: script does not work when the route is bound to a service`

t/admin/script.t
```diff
@@ -0,0 +1,103 @@
...
+repeat_each(1);
+log_level('info');
+no_long_string();
+no_root_location();
+
+run_tests;
+
+__DATA__
+
+=== TEST 1: add service and service has plugin
```
tokers, line 28: Title should be changed since you just add exactly one service.

spacewander: The title is correct, the service has plugin.

t/admin/script.t
```diff
@@ -0,0 +1,103 @@
+#
```
spacewander, line 1: Can we add this under `t/script` instead?

tzssangglass: got

spacewander (approved):

t/admin/script.t
tokers: I misunderstood it when I first saw it... Maybe `add service which has plugins` is better.

tzssangglass: update

membphis (approved):

Issue #3652, opened 2021-02-24
建议：是不是后续版本可以考虑把插件的优先级集成到统一的配置里去


StretchV:
因为插件一多，当需要新增一个自定义插件的时候需要挨个查看其他的插件文件中的priority以避免重复，体验不是很好

StretchV:
https://github.com/apache/apisix-dashboard/blob/v2.4/docs/FAQ.md#4-after-modifying-the-plugin-schema-or-creating-a-custom-plugin-in-apache-apisix-why-cant-i-find-it-on-the-dashboard

moonming:
Maybe we can automatically generate a priority list of plugins
@Firstsawyou

Firstsawyou:
> Maybe we can automatically generate a priority list of plugins
> @Firstsawyou

Yes, I will try it.

PR #3680, opened 2021-02-25
docs: plugin-develop.md add to view all plugin priorities

Firstsawyou:

Related issues: https://github.com/apache/apisix/issues/3652

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/plugin-develop.md
+++ b/docs/en/latest/plugin-develop.md
@@ -103,7 +103,7 @@ local _M = {
 }
 ```
 
-Note : The priority of the new plugin cannot be the same as the priority of any existing plugin, you can view the priority of all plugins in the [t/debug/debug-mode.t](../../../t/debug/debug-mode.t) file. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
+Note : The priority of the new plugin cannot be the same as the priority of any existing plugin, you can use the `/v1/schema` method of [control API](./control-api.md) to view the priority of all plugins. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
 
 in the "__conf/config-default.yaml__" configuration file, the enabled plugins (all specified by plugin name) are listed.
 
--- a/docs/en/latest/plugin-develop.md
+++ b/docs/en/latest/plugin-develop.md
@@ -103,7 +103,7 @@ local _M = {
 }
 ```
 
-Note : The priority of the new plugin cannot be the same as the priority of any existing plugin. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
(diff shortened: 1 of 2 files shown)


docs/en/latest/plugin-develop.md
```diff
@@ -103,7 +103,7 @@ local _M = {
 }
 ```
 
-Note : The priority of the new plugin cannot be the same as the priority of any existing plugin. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
+Note : The priority of the new plugin cannot be the same as the priority of any existing plugin, you can view the priority of all plugins in the [t/debug/debug-mode.t](../../../t/debug/debug-mode.t) file. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
```
spacewander, line 106: I think it would be better to use control API: https://github.com/apache/apisix/blob/master/docs/en/latest/control-api.md#get-v1schema

spacewander (changes requested):

docs/en/latest/plugin-develop.md
Firstsawyou: I think that `Control API` has obtained the schema data of all plugins, and the content information is too much, and it is not intuitive to check the priority of all plugins.

spacewander: No, it is the official way to check the priority since users may add their own plugins and those plugins can't be reflected in our test file.

Firstsawyou: Okay, but I think there is too much data obtained through `/v1/schema`. We should add a new API interface, such as `/v1/plugins/priority`, this API only returns the priority value of all plugins, and they are ordered. What do you think?

spacewander: No. Don't make thing complex. The method is called rarely.

Firstsawyou: got it.

docs/en/latest/plugin-develop.md
```diff
@@ -103,7 +103,7 @@ local _M = {
 }
 ```
 
-Note : The priority of the new plugin cannot be the same as the priority of any existing plugin. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
+Note : The priority of the new plugin cannot be the same as the priority of any existing plugin, you can use the `/v1/schema` method of [control API](./control-api.md) to view the priority of all plugins. In addition, plugins with a high priority value will be executed first in a given phase (see the definition of `phase` in [choose-phase-to-run](#choose-phase-to-run)). For example, the priority of example-plugin is 0 and the priority of ip-restriction is 3000. Therefore, the ip-restriction plugin will be executed first, then the example-plugin plugin.
```
juzhiyuan, line 106: `./control-api.md#get-v1schema`

juzhiyuan: would better use a more clear & directly way, what do you think @spacewander @Firstsawyou ?

moonming:
how about add an official plugin priority list？

docs/en/latest/plugin-develop.md
tokers, line 106: Some typos.

> The priority of the new plugin cannot be the same as the priority of any existing plugin

Change to "The priority of the new plugin cannot be same to any existing ones".

> plugins with a high priority

Change to "plugins with higher priority".

Firstsawyou: > would better use a more clear & direct way, what do you think @spacewander @Firstsawyou?

agree.

Firstsawyou:
> how about add an official plugin priority list？

I think adding a new API interface through `control API`, such as `/v1/plugins/priority`, the API only returns the priority value of all plugins, and they are in order.
Have discussed with @spacewander  in the above review.  I will add an issue to discuss it.

docs/en/latest/plugin-develop.md
Firstsawyou: updated.

spacewander (approved):

juzhiyuan (approved):

Issue #3635, opened 2021-02-22
feat: Add method filter on consummer-restriction plugin


jp-gouin:
Hi,
I think it could be interesting to add a filter on the method in the consummer-restriction plugin;

With something like : 
```
"consumer-restriction": {
            "whitelist": [
                {
                    "jack1" : ["OPTIONS","GET","POST","DELETE","PATCH","*"]
                }
            ]
        }
```
This feature could be optional and fallback on the current processing.

spacewander:
PR is welcome!

jp-gouin:
Ok sure , i'll work on that

PR #3691, opened 2021-02-26
feat: Add filter on HTTP methods for consumer-restriction plugin

jp-gouin:
Fix for #3635 
Add the possibility to add `allowed_methods` for a user when `whitelist` is set.
This will restrict the user to only performed the HTTP action matching the list specified in `allowed_method.methods`

This is the format : 
```
"plugins": {
        "key-auth": {},
        "consumer-restriction": {
            "whitelist": [
                "jack1"
            ],
            "allowed_methods":[{
              "user": "jack1",
              "methods": ["POST"]
            }]
        }
    }
```
I choose to add a dedicated section instead of modifying the existing `whitelist` one because i think it's more readable. 
When nothing is set , then only the whitelist is applied .
And in order to set `allowed_methods` , `whitelist` is required.

### Pre-submission checklist:
Test case updated with 2 basics tests 
1. only allow `post` on resources and try to `get` the resource -> unauthorized
2. Add `get`capability to the user and try to `get` the resource -> authorized

Add 2 test cases for testing the plugin scheme 

1. Only `allowed_methods` is set -> failed
2. `allowed_methods` is set with `blacklist` -> failed

English documentation updated , any help for the Chineese one would be appreciated.

* [X] Have you added corresponding test cases?
* [X] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

### DIFF
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -19,74 +19,49 @@ local core      = require("apisix.core")
 local ngx       = ngx
 local schema = {
     type = "object",
-    oneOf = {
-        {
-            title = "blacklist",
-            properties = {
-                type = {
-                    type = "string",
-                    enum = {"consumer_name", "service_id"},
-                    default = "consumer_name"
-                },
-                blacklist = {
-                    type = "array",
-                    minItems = 1,
-                    items = {type = "string"}
-                },
-                rejected_code = {type = "integer", minimum = 200, default = 403}
-            },
-            required = {"blacklist"},
+    properties = {
+        type = {
+            type = "string",
+            enum = {"consumer_name", "service_id"},
+            default = "consumer_name"
         },
-        {
-            title = "whitelist",
-            properties = {
-                type = {
-                    type = "string",
-                    enum = {"consumer_name", "service_id"},
-                    default = "consumer_name"
-                },
-                whitelist = {
-                    type = "array",
-                    minItems = 1,
-                    items = {type = "string"}
-                },
-                rejected_code = {type = "integer", minimum = 200, default = 403}
-            },
-            required = {"whitelist"},
+        blacklist = {
+            type = "array",
+            minItems = 1,
+            items = {type = "string"}
         },
-        {
-            title = "allowed_by_methods",
-            properties = {
-                type = {
-                    type = "string",
-                    enum = {"consumer_name", "service_id"},
-                    default = "consumer_name"
-                },
-                whitelist = {
-                    type = "array",
-                    items = {
-                        type = "object",
-                        properties = {
-                            user = {
-                                type = "string"
-                            },
-                            methods = {
-                                type = "array",
-                                minItems = 1,
-                                items = {
-                                    type = "string",
-                                    enum = {"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD",
-                                            "OPTIONS", "CONNECT", "TRACE"},
-                                }
-                            }
+        whitelist = {
+            type = "array",
+            minItems = 1,
+            items = {type = "string"}
+        },
+        allowed_by_methods = {
+            type = "array",
+            items = {
+                type = "object",
+                properties = {
+                    user = {
+                        type = "string"
+                    },
+                    methods = {
+                        type = "array",
+                        minItems = 1,
+                        items = {
+                            type = "string",
+                            enum = {"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD",
+                                    "OPTIONS", "CONNECT", "TRACE"},
                         }
                     }
-                },
-                rejected_code = {type = "integer", minimum = 200, default = 403}
-            },
-            required = {"allowed_by_methods"},
-        }
-    }
+                }
+            }
+        },
+        rejected_code = {type = "integer", minimum = 200, default = 403}
+    },
+    anyOf = {
+        {required = {"blacklist"}},
+        {required = {"whitelist"}},
+        {required = {"allowed_by_methods"}}
+    },
 }
 
 local plugin_name = "consumer-restriction"
@@ -130,13 +105,14 @@ local function is_method_allowed(allowed_methods, method, user)
     return true
 end
 
+local function reject(conf)
+    return conf.rejected_code, { message = "The " .. conf.type .. " is forbidden." }
+end
+
 function _M.check_schema(conf)
     local ok, err = core.schema.check(schema, conf)
    if not ok then
         return false, err
-   end
-   if ((conf.allowed_by_methods and #conf.allowed_by_methods > 0) and not conf.whitelist ) then
-    return false, "allowed_by_methods set but no whitelist provided"
    end
    return true
 end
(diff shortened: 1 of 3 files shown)


apisix/plugins/consumer-restriction.lua
```diff
@@ -114,12 +161,16 @@ function _M.access(conf, ctx)
         if not is_include(value, conf.whitelist) then
             block = true
         end
+        if conf.allowed_methods and #conf.allowed_methods > 0 then
+            if not is_method_allow(conf.allowed_methods, method, value) then
```
spacewander, line 165: "is_method_allowed"

apisix/plugins/consumer-restriction.lua
```diff
@@ -53,6 +53,38 @@ local schema = {
                 rejected_code = {type = "integer", minimum = 200, default = 403}
             },
             required = {"whitelist"},
+        },
+        {
+            title = "allowed_methods",
+            properties = {
+                type = {
```
spacewander, line 60: Can we extra this part and share across the components with different titles?

apisix/plugins/consumer-restriction.lua
```diff
@@ -53,6 +53,38 @@ local schema = {
                 rejected_code = {type = "integer", minimum = 200, default = 403}
             },
             required = {"whitelist"},
+        },
+        {
+            title = "allowed_methods",
```
spacewander, line 58: Would it be better to use `allowed_by_methods`?

docs/en/latest/plugins/consumer-restriction.md
```diff
@@ -116,6 +117,80 @@ HTTP/1.1 403 Forbidden
...
+    "upstream": {
+        "type": "roundrobin",
+        "nodes": {
+            "127.0.0.1:1980": 1
+        }
+    },
+    "plugins": {
+        "basic-auth": {},
+        "consumer-restriction": {
+            "whitelist": [
```
spacewander, line 137:
```
        "consumer-restriction": {
            "allowed_by_methods":[{
                "user": "jack1",
                "methods": ["POST"]
            }]
        }
```
is enough. There is no need to configure `jack1` in two places. And the `allow` already has the meaning of `whitelist`.

spacewander (changes requested):

apisix/plugins/consumer-restriction.lua
jp-gouin: Sorry, I'm not sure to understand what you mean by extra

apisix/plugins/consumer-restriction.lua
jp-gouin: agree

apisix/plugins/consumer-restriction.lua
jp-gouin: agree

docs/en/latest/plugins/consumer-restriction.md
jp-gouin: Ok so i completely drop the whitelist part in favor of allowed_by_methods ?

jp-gouin: If so that would mean some rework of the whole plugin.
Right now the `allowed_by_methods` required to set the list of methods , so if we drop `whitelist`we need a quick setup . Like if there is no method , then full access ? or the user still ha to set the list of method but we take into account a `*` wildcard ?

apisix/plugins/consumer-restriction.lua
spacewander: Sorry, it is my typo. It should be `extract`.

docs/en/latest/plugins/consumer-restriction.md
spacewander: OK. I made a mistake, the `whitelist` is not required with `allowed_by_methods`. I think we should show them separately in the doc and test.

spacewander: I think we can use the schema below to allow both `whitelist` and `allowed_by_methods`:
```lua
local schema = {
    type = "object",
    properties = {
        type = {
            type = "string",
            enum = {"consumer_name", "service_id"},
            default = "consumer_name"
        },
        blacklist = {
            type = "array",
            minItems = 1,
            items = {type = "string"}
        },
        whitelist = {
            type = "array",
            minItems = 1,
            items = {type = "string"}
        },
        allowed_by_methods = {
            type = "array",
            items = {
                type = "object",
                properties = {
                    user = {
                        type = "string"
                    },
                    methods = {
                        type = "array",
                        minItems = 1,
                        items = {
                            type = "string",
                            enum = {"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD",
                                    "OPTIONS", "CONNECT", "TRACE"},
                        }
                    }
                }
            }
        },
        rejected_code = {type = "integer", minimum = 200, default = 403}
    },
    anyOf = {
        {required = {"blacklist"}},
        {required = {"whitelist"}},
        {required = {"allowed_by_methods"}}
    },
}

```

spacewander: We can change the title of https://github.com/apache/apisix/blob/29b1e2dfa4ae27cb8fff0511d30106803224d30c/t/plugin/consumer-restriction.t#L57 and say that the priority is blacklist > whitelist > allowed_by_methods.

spacewander: It is a large change but I think it is flexible for the users.

jp-gouin: Ok , no problem.
So `allowed_by_methods` could exist without `whitelist`.
If whitelist is set it will override what could be specified in the `allowed_by_methods` section

jp-gouin:
I made the following update : 
`blacklist` , `whitelist` and `allowed_by_methods` can be set at the same time.
`blacklist` has the priority over `whitelist` and  `whitelist` has the priority over `allowed_by_methods`.

There is now a direct reject of the connection if the user is blacklisted
```
if conf.blacklist and #conf.blacklist > 0 then
        if is_include(value, conf.blacklist) then
            return reject(conf)
        end
    end
```

`allowed_by_methods` is only performed if the user is not whitelisted 
```
 if conf.allowed_by_methods and #conf.allowed_by_methods > 0 and not whitelisted then
        if not is_method_allowed(conf.allowed_by_methods, method, value) then
            block = true
        end
    end
```
The overall behavior stay the same is you set only `blacklist` and or `whitelist`

t/plugin/consumer-restriction.t
```diff
@@ -515,14 +521,254 @@ passed
...
+--- more_headers
+Authorization: Basic amFjazIwMjA6MTIzNDU2
+--- response_body
+hello world
+--- no_error_log
+[error]
+
+
+
+=== TEST 28: test blacklist priority
```
spacewander, line 656: test whitelist priority

t/plugin/consumer-restriction.t
```diff
@@ -475,7 +474,7 @@ Authorization: Basic amFjazIwMjA6MTIzNDU2
 
 
 
-=== TEST 21: remove consumer-restriction
```
spacewander, line 478: Why change this test?

jp-gouin: I bump the remove restriction to test 31 due to the tests i added and to remains consistent .

t/plugin/consumer-restriction.t
jp-gouin: Ah yes my bad

spacewander (approved):

moonming (approved):

PR #3705, opened 2021-03-01
docs: update the plugin-develop.md doc

Firstsawyou:

Fix errors in the `check dependencies` section of the plugin-develop.md document.

### Pre-submission checklist:


docs/en/latest/plugin-develop.md
```diff
@@ -85,8 +85,8 @@ The plugin itself provides the init method. It is convenient for plugins to perf
  the plugin is loaded.
 
 Note : if the dependency of some plugin needs to be initialized when Nginx start, you may need to add logic to the initialization
-       method "http_init" in the file __apisix.lua__, And you may need to add some processing on generated part of Nginx
-       configuration file in __bin/apisix__ file. but it is easy to have an impact on the overall situation according to the
+       method "http_init" in the file __apisix/init.lua__, and you may need to add some processing on generated part of Nginx
+       configuration file in __apisix/cli/ngx_tpl.lua__ file. But it is easy to have an impact on the overall situation according to the
```
spacewander, line 89: Can we use https://github.com/apache/apisix/blob/master/docs/en/latest/customize-nginx-configuration.md instead?

spacewander (commented):
Can we use https://github.com/apache/apisix/blob/master/docs/en/latest/customize-nginx-configuration.md instead?

docs/en/latest/plugin-develop.md
spacewander, line 89: Can we use https://github.com/apache/apisix/blob/master/docs/en/latest/customize-nginx-configuration.md instead?

Firstsawyou: I think the dependency of plugin development is better declared in the `ngx_tpl.lua` file, because it is not easy to be changed. The declaration in the `config.yaml` file is more suitable for user-defined configuration. And we used `ngx_tpl.lua`  above (https://github.com/apache/apisix/pull/3705/files#diff-6c0909835441e968e645d3d914cdc3471303a072f5ff26c8314a2d23f0a05593R68),  we should keep the same. What do you think?

spacewander: The solution is imperfect to me. But I have to accept it because there is no better way.

spacewander (approved):

tokers (approved):

Issue #3740, opened 2021-03-03
bug: the traffic-split plugin is invalid to bind upstream via upstream_id


Firstsawyou:
### Issue description

When both the routing and traffic-split plugins use upstream_id to bind upstream, the request distribution is not accurate.

1、add route and bind plugin

The upstream_id is 1, the response data is 1980, and the upstream_id is 2 the response data is 1981.

```shell
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "create_time":1614743731,
    "update_time":1614745342,
    "uris":[
        "/*"
    ],
    "name":"luarocks_cn",
    "methods":[
        "GET",
        "HEAD",
        "POST",
        "PUT",
        "DELETE",
        "OPTIONS",
        "PATCH"
    ],   
    "vars":[

    ],
    "plugins":{
        "proxy-rewrite":{
            "disable":false,
            "scheme":"http"
        },
        "traffic-split":{
            "disable":false,
            "rules":[
                {
                    "match":[
                        {
                            "vars":[
                                [
                                    "uri",
                                    "==",
                                    "/hello"
                                ]
                            ]
                        }
                    ],
                    "weighted_upstreams":[
                        {
                            "upstream_id":"2"
                        }
                    ]
                }
            ]
        }
    },
    "upstream_id":"1",
    "status":1
}
'
```

2、Test plugin

The vars rule passes:
```shell
$ curl http://127.0.0.1:9080/hello
1981
```

The vars rule failed：

```shell
$ curl http://127.0.0.1:9080/hello12
1980
```

The vars rule passes:
```shell
$ curl http://127.0.0.1:9080/hello
1981
```

The vars rule failed (an exception occurred)：

```shell
$ curl http://127.0.0.1:9080/hello12
1981
```

tzssangglass:
```json
"weighted_upstreams":[
                        {
                            "upstream_id":"2"
                        }
                    ]
```
should be 
```json
"weighted_upstreams":[
                        {
                            "upstream_id":2
                        }
                    ]
```

the type of `upstream_id` is number but not string.

by the way, it is an interesting bug
I added a debug log
```lua
core.log.warn("route.value.upstream_id: ", core.json.encode(route.value.upstream_id))
```
[...]

PR #3758, opened 2021-03-04
fix: the traffic-split plugin is invalid to bind upstream via upstream_id

tzssangglass:

fix: #3740 
### Pre-submission checklist:
the cause of the bug reference: https://github.com/apache/apisix/issues/3740#issuecomment-789852741
~~this change temporarily stores the route's `upstream_id` in the `matched_route` table. If the current request `match` fails and the temporarily stored `upstream_id` exists in the `matched_route` table, the route's `upstream_id` is restored to the temporarily stored `upstream_id`.~~
when the `route` is configured with an `upstream_id`, the `upstream_id` is passed into the `server_list`.

### DIFF
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -314,7 +314,8 @@ function _M.access(conf, ctx)
         return
     end
 
-    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams, ctx.matched_route.value.upstream_id)
+    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
+                                ctx.matched_route.value.upstream_id)
     if not rr_up then
         core.log.error("lrucache roundrobin failed: ", err)
         return 500
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -306,8 +306,8 @@ function _M.access(conf, ctx)
     core.log.info("match_flag: ", match_flag)
 
     if not match_flag then
-        if ctx.matched_route.value["original_uid"] then
-            ctx.matched_route.value.upstream_id = ctx.matched_route.value["original_uid"]
+        if ctx.matched_route.value.original_upstream_id then
+            ctx.matched_route.value.upstream_id = ctx.matched_route.value.original_upstream_id
             core.log.info("original_uid: ", ctx.matched_route.value.upstream_id)
         end
         return
@@ -324,7 +324,7 @@ function _M.access(conf, ctx)
         core.log.info("upstream: ", core.json.encode(upstream))
         return set_upstream(upstream, ctx)
     elseif upstream and upstream ~= "plugin#upstream#is#empty" then
-        ctx.matched_route.value["original_uid"] = ctx.matched_route.value.upstream_id
+        ctx.matched_route.value.original_upstream_id = ctx.matched_route.value.upstream_id
         ctx.matched_route.value.upstream_id = upstream
         core.log.info("upstream_id: ", upstream)
         return
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -306,6 +306,10 @@ function _M.access(conf, ctx)
     core.log.info("match_flag: ", match_flag)
 
     if not match_flag then
+        if ctx.matched_route.value["original_uid"] then
+            ctx.matched_route.value.upstream_id = ctx.matched_route.value["original_uid"]
+            core.log.info("original_uid: ", ctx.matched_route.value.upstream_id)
+        end
         return
     end
 
@@ -320,6 +324,7 @@ function _M.access(conf, ctx)
         core.log.info("upstream: ", core.json.encode(upstream))
         return set_upstream(upstream, ctx)
     elseif upstream and upstream ~= "plugin#upstream#is#empty" then
+        ctx.matched_route.value["original_uid"] = ctx.matched_route.value.upstream_id
         ctx.matched_route.value.upstream_id = upstream
         core.log.info("upstream_id: ", upstream)
         return
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -314,8 +314,9 @@ function _M.access(conf, ctx)
         return
     end
 
+    local route_upstream_id = ctx.matched_route.value.upstream_id
     local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
-                                ctx.matched_route.value.upstream_id)
+                                route_upstream_id)
     if not rr_up then
         core.log.error("lrucache roundrobin failed: ", err)
         return 500
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -315,8 +315,8 @@ function _M.access(conf, ctx)
     end
 
     local route_upstream_id = ctx.matched_route.value.upstream_id
-    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
-                                route_upstream_id)
+    local rr_up, err = core.lrucache.plugin_ctx(lrucache, ctx, nil, new_rr_obj,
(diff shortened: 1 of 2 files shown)


apisix/plugins/traffic-split.lua
```diff
@@ -320,6 +324,7 @@ function _M.access(conf, ctx)
         core.log.info("upstream: ", core.json.encode(upstream))
         return set_upstream(upstream, ctx)
     elseif upstream and upstream ~= "plugin#upstream#is#empty" then
+        ctx.matched_route.value["original_uid"] = ctx.matched_route.value.upstream_id
```
spacewander, line 327: Should do the same thing with the code below?
```
    core.log.info("route_up: ", upstream)
    ctx.matched_route.value.upstream_id = nil
```

tzssangglass: I don't think so, because here it will directly return
https://github.com/apache/apisix/blob/4820ad93e5fc0550addbd7029a20eab6b7ea800e/apisix/plugins/traffic-split.lua#L322-L325

apisix/plugins/traffic-split.lua
```diff
@@ -306,6 +306,10 @@ function _M.access(conf, ctx)
     core.log.info("match_flag: ", match_flag)
 
     if not match_flag then
+        if ctx.matched_route.value["original_uid"] then
```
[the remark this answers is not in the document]
tzssangglass: got

apisix/plugins/traffic-split.lua
spacewander: I mean, if upstream == "plugin#upstream#is#empty", will the original upstream_id be overridden?

tzssangglass: yes, it will be overridden, I continue to fix it.

apisix/plugins/traffic-split.lua
```diff
@@ -309,7 +314,8 @@ function _M.access(conf, ctx)
         return
     end
 
-    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams)
+    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
```
spacewander, line 317: Need to use `ctx.matched_route.value.upstream_id` as the extra key.

tzssangglass: you mean
```lua
local route_upstream_id = ctx.matched_route.value.upstream_id
local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams, route_upstream_id)
```
?

tzssangglass: update

spacewander: `lrucache` accepts an extra key as the second argument.

t/plugin/traffic-split.t
```diff
@@ -1615,7 +1615,134 @@ passed
...
+
+
+
+=== TEST 49: all requests `match` rule passed, proxy requests to the upstream of route based on the structure with only `weight` in `weighted_upstreams`
+--- config
+location /t {
+    content_by_lua_block {
+        local t = require("lib.test_admin").test
+        local bodys = {}
+        local headers = {}
```
Firstsawyou, line 1727: I think it is not used by `headers` and we should remove it.

tzssangglass: update

apisix/plugins/traffic-split.lua
spacewander: Need to solve it.

spacewander:
After apply this change:
```
diff --git apisix/plugins/traffic-split.lua apisix/plugins/traffic-split.lua
index eb7a2028..3d5e7d8f 100644
--- apisix/plugins/traffic-split.lua
+++ apisix/plugins/traffic-split.lua
@@ -315,8 +315,7 @@ function _M.access(conf, ctx)
     end

     local route_upstream_id = ctx.matched_route.value.upstream_id
-    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
-                                route_upstream_id)
+    local rr_up, err = new_rr_obj(weighted_upstreams, route_upstream_id)
     if not rr_up then
         core.log.error("lrucache roundrobin failed: ", err)
         return 500
```
many test cases are failed. Look like `new_rr_obj` corrupts the route's configuration. This may happen once the cache is evicted.

tzssangglass:
>  Look like new_rr_obj corrupts the route's configuration. This may happen once the cache is evicted.

yes, according to my understanding, I think the design idea of `traffic-split` itself is to achieve the goal by corrupts the route's configuration.

take a look at 
https://github.com/apache/apisix/blob/6cbaf55aab7c572410adfc4981f42f725d058ccb/apisix/plugins/traffic-split.lua#L322-L323

spacewander:
There should be a way to do the same thing without corrupting the data. Otherwise once the cache is evicted we will be in trouble.

tzssangglass:
This PR does not do this work. We should open another issue to discuss it if needed, because this time it involves refactoring the plugin's implementation logic.

> There should be a way to do the same thing without corrupting the data. Otherwise once the cache is evicted we will be in trouble.

This PR does not do this work. If needed, we should open another issue to discuss it, as it would involve refactoring the plugin's implementation logic.

apisix/plugins/traffic-split.lua
spacewander: @tzssangglass 
Personally, we need to add `route_upstream_id` to the key as `route_upstream_id` and `weighted_upstreams` may change differently. As using `route_upstream_id` in the second argument will hit the bug I mention, maybe we can switch to use https://github.com/apache/apisix/blob/c530135c5a179c99637d52d99e0e8a45ce4cc8b8/apisix/plugins/limit-req.lua#L69?

tzssangglass: I tried it and it still doesn't work, because this plugin still need to control the weight by modifying the `upstream_id` in the route,  even if `core.lrucache.plugin_ctx` is used, since `route_upstream_id` is in the key and `route_upstream_id` will be overwritten by the `plugin_upstream_id`, it will result in two keys and inaccurate data obtained.

use `weighted_upstreams` as key because `weighted_upstreams` do not change.

spacewander: Try?
```diff
diff --git apisix/plugins/traffic-split.lua apisix/plugins/traffic-split.lua
index eb7a2028..966275f9 100644
--- apisix/plugins/traffic-split.lua
+++ apisix/plugins/traffic-split.lua
@@ -315,8 +315,8 @@ function _M.access(conf, ctx)
     end

     local route_upstream_id = ctx.matched_route.value.upstream_id
-    local rr_up, err = lrucache(weighted_upstreams, nil, new_rr_obj, weighted_upstreams,
-                                route_upstream_id)
+    local rr_up, err = core.lrucache.plugin_ctx(lrucache, ctx, nil, new_rr_obj, weighted_upstreams,
+                                                route_upstream_id)
     if not rr_up then
         core.log.error("lrucache roundrobin failed: ", err)
         return 500
```

tzssangglass: try
```lua   
 local rr_up, err = core.lrucache.plugin_ctx(lrucache, ctx, route_upstream_id, new_rr_obj, weighted_upstreams,
                                                route_upstream_id)
```

tzssangglass: update.
I know that using `core.lrucache.plugin_ctx` is more in line with the common way, but I don't understand the benefit of using `api_ctx.conf_type` and `api_ctx.conf_id` as lrucache keys over `weighted_upstreams`.

spacewander: The default key should ensure if the configuration of the route changes, it will change too. So we can ensure either route_upstream_id or weighted_upstreams changes, the version will change.

The corruption of the route configuration is another story (which should be fixed)...

spacewander (approved):

apisix/plugins/traffic-split.lua
tzssangglass: > So we can ensure either route_upstream_id or weighted_upstreams changes, the version will change.

got!

membphis (approved):

PR #3820, opened 2021-03-13
feat: add nacos support

benx203:
feat:add nacos support


spacewander (changes requested):
Please add test under t/discovery.

spacewander:
And you need to add docs under https://github.com/apache/apisix/tree/master/docs/en/latest/discovery

spacewander:
Please make sure it can pass the lint.

docs/en/latest/discovery/nacos.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Configuration for Nacos
+
+Add following configuration in `conf/config.yaml` :
+
+```yaml
+discovery:
+  nacos:
+    host:
+      - "http://192.168.33.1:8848"
```
spacewander, line 32: Why don't mention the user / password feature?

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+        host = {
+            type = "array",
+            minItems = 1,
+            items = {
+                type = "string",
+            },
+        },
+        fetch_interval = {type = "integer", minimum = 1, default = 30},
+        prefix = {type = "string"},
+        weight = {type = "integer", minimum = 0},
```
spacewander, line 50: Missing default for weight?
And the minimum weight should be 1.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+        if not body then
+            return nil, 'invalid body : ' .. err
+        end
+        -- log.warn(method, url, body)
+        headers['Content-Type'] = 'application/json'
+    end
+
+    local httpc = http.new()
+    local timeout = local_conf.discovery.nacos.timeout
+    local connect_timeout = timeout and timeout.connect or 2000
```
spacewander, line 123: The default value will be injected via schema check. We don't need to assign it.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+    local ok, err = core.schema.check(schema, local_conf.discovery.nacos)
+    if not ok then
+        error("invalid nacos configuration: " .. err)
+        return
+    end
+    default_weight = local_conf.discovery.nacos.weight or 100
+    log.info("default_weight:", default_weight, ".")
+    local fetch_interval = local_conf.discovery.nacos.fetch_interval or 30
+    log.info("fetch_interval:", fetch_interval, ".")
+    service_list_path = local_conf.discovery.nacos.service_list_path or
```
spacewander, line 228: Where should I configure `service_list_path`? There is neither schema nor doc for it.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+            'ns/service/list?pageNo=1&pageSize=20'
+    instance_list_path = local_conf.discovery.nacos.instance_list_path or
```
spacewander, line 230: ditto

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+        return
+    end
+
+    local base_uri, basic_auth = service_info()
+    if not base_uri then
+        return
+    end
+
+    local up_apps = core.table.new(0, 0)
+    local data = get_url(base_uri,basic_auth,service_list_path)
```
spacewander, line 173: Need space after the comma.
And please check the returned value of `get_url`.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+
+    return url, basic_auth
+end
+
+
+local function request(request_uri, basic_auth, method, path, query, body)
+    log.info("nacos uri:", request_uri, ".")
+    local url = request_uri .. path
+    local headers = core.table.new(0, 5)
+    headers['Connection'] = 'Keep-Alive'
```
spacewander, line 104: We don't need it for HTTP 1.1

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    return url, basic_auth
+end
+
+
+local function request(request_uri, basic_auth, method, path, query, body)
+    log.info("nacos uri:", request_uri, ".")
+    local url = request_uri .. path
+    local headers = core.table.new(0, 5)
```
spacewander, line 103: Why choose 5?

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+        local protocol = string_sub(url, 1, protocol_idx + 2)
+        local user_and_password = string_sub(url, protocol_idx + 3, auth_idx - 1)
+        local other = string_sub(url, auth_idx + 1)
+        url = protocol .. other
+        basic_auth = "Basic " .. ngx.encode_base64(user_and_password)
+    end
+    if local_conf.discovery.nacos.prefix then
+        url = url .. local_conf.discovery.nacos.prefix
+    end
+    if string_sub(url, #url) ~= "/" then
```
spacewander, line 92: Better to use `str_byte(uri, #uri) == str_byte("/")`

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,237 @@
...
+        ssl_verify = false,
+    })
+end
+
+
+local function get_url(request_url,basic_auth,path)
+    local res, err = request(request_url, basic_auth, "GET", path)
+    if not res then
+        log.error("failed to fetch registry", err)
+        return
```
spacewander, line 144: Please pass the error out so the caller can check it.

docs/en/latest/discovery/nacos.md
benx203: Because Nacos not need login yet.

apisix/discovery/nacos.lua
benx203: OK,I copy from eureka.

apisix/discovery/nacos.lua
benx203: OK.

apisix/discovery/nacos.lua
benx203: I will fix it.

apisix/discovery/nacos.lua
benx203: OK.
By the way,when i add nacos test script in centos7-ci.yml,can't submit,how can i  fix it?
error tips:
[remote rejected] (refusing to allow an OAuth App to create or update workflow `.github/workflows/centos7-ci.yml` without `workflow` scope)

benx203: content:
```
 rm -rf nacos-docker
        git clone --depth 1 https://github.com/benx203/nacos-docker.git
        cd nacos-docker
        docker-compose -f example/standalone-test.yaml up
        ps aux|grep nacos-gateway-provider-example-2.2.6-SNAPSHOT |grep -v grep|awk '{print $2}'|xargs kill -9
        nohup java -jar example/nacos-gateway-provider-example-2.2.6-SNAPSHOT.jar --spring.application.name=APISIX-NACOS  > /dev/null 2>&1 &
```

spacewander: > refusing to allow an OAuth App

Try to update it locally via a regular editor and submit it with Git?

benx203: Thanks,Submit it with TortoiseGit success.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,285 @@
...
+    end
+
+    if tostring(data.count) == "0" then
+        applications = up_apps
+        return
+    end
+
+    for _, service_name in ipairs(data.doms) do
+        data, err = get_url(base_uri, instance_list_path .. service_name .. token_param)
```
spacewander, line 224: We can add `healthyOnly` to the argument to reduce the response

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,285 @@
...
+            type = "array",
+            minItems = 1,
+            items = {
+                type = "string",
+            },
+        },
+        fetch_interval = {type = "integer", minimum = 1, default = 30},
+        prefix = {type = "string", default = "/nacos/v1/"},
+        auth_path = {type = "string", default = "auth/login"},
+        service_list_path = {type = "string", default = "ns/service/list?pageNo=1&pageSize=20"},
```
spacewander, line 55: Only return the first 20 services is not a good idea.
We need to iterate the upstreams and build a service list from those that have `nacos` discovery_type.
Like this:
https://github.com/apache/apisix/blob/e9a59ac5e8618737f69eddbaf05097b7ffab46ff/apisix/control/v1.lua#L144-L149

You might need some lazy load trick to break the dependent cycle.

t/discovery/nacos.t
```diff
@@ -0,0 +1,125 @@
...
+=== TEST 1: prepare nacos register nodes
+--- config
+location /nacos {
+    proxy_pass http://127.0.0.1:8848;
+}
+
+--- pipelined_requests eval
+[
+    "POST /nacos/v1/ns/instance?port=18001&healthy=true&ip=127.0.0.1&weight=1.0&serviceName=APISIX-NACOS&encoding=GBK&enabled=true",
+    "DELETE /nacos/v1/ns/service?serviceName=APISIX-NACOS",
```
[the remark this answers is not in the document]
benx203: Delete directly will return 400.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,285 @@
...
+        instance_list_path = {type = "string", default = "ns/instance/list?serviceName="},
```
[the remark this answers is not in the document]
benx203: When Nacos api change.

t/discovery/nacos.t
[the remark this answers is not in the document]
spacewander: But it is not the feature of APISIX. We don't need to test nacos itself.

apisix/discovery/nacos.lua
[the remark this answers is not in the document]
spacewander: @benx203
I think we can remove it now and add it if needed.
We need to add a doc and test for this feature if you insist to keep it.
(I don't recommend to do that as this PR is big enough)

.travis/linux_tengine_runner.sh
```diff
@@ -40,6 +40,9 @@ before_install() {
     # start consul servers
     docker run --rm --name consul_1 -d -p 8500:8500 consul:1.7 consul agent -server -bootstrap-expect=1 -client 0.0.0.0 -log-level info -data-dir=/consul/data
     docker run --rm --name consul_2 -d -p 8600:8500 consul:1.7 consul agent -server -bootstrap-expect=1 -client 0.0.0.0 -log-level info -data-dir=/consul/data
+
+    # start nacos server
+    docker run --rm --name nacos_1 -d --env PREFER_HOST_MODE=hostname --env MODE=standalone --env JVM_XMS=512m --env JVM_XMX=512m --env JVM_XMN=256m -p8848:8848 nacos/nacos-server:latest
```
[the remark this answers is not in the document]
benx203: If add NACOS_AUTH_ENABLE=true, file nacos.t need login too.
Could you give me a t file login demo?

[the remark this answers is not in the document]
spacewander: You can use different conf in different tests, one for auth success and another for auth fail.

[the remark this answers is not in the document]
benx203: How to check nacos in discovery plugin whether authed?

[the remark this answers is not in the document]
spacewander: If the auth enable nacos can be used, the auth is successful.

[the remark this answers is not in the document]
benx203: But if nacos start with auth mode,in nacos.t add services to nacos need send auth token.
How can i save auth token in nacos.t?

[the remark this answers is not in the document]
spacewander: You can export the auth function: https://github.com/apache/apisix/pull/3820/files#diff-46fd008790f9ea29612a92ca5c94212ae43c81ab6b902ef3014d3d6b4bdfc5b3R202
and call it in the test file.

apisix/discovery/nacos.lua
benx203: fixed.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,324 @@
...
+            local_conf.discovery.nacos and local_conf.discovery.nacos.host
+    if not host then
+        log.error("do not set nacos.host")
+        return
+    end
+
+    local username, password
+    -- TODO Add health check to get healthy nodes.
+    local url = host[math_random(#host)]
+    local auth_idx = str_find(url, "#")
```
spacewander, line 94: Is there any need to change `@` to `#`? Maybe I miss something?
It is the standard to use `@` to separate the authority part from the rest:
https://en.wikipedia.org/wiki/URL#Syntax

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,324 @@
...
+    end
+
+    for _, service_name in ipairs(data.doms) do
+        core.table.insert(infos, service_name)
+    end
+    return data, err, path
+end
+
+
+local function iter_and_add_service_info(infos, base_uri, token_param)
```
spacewander, line 195: Emm... I mean to iterate the upstream configurations and collect the service list from those that have `discovery_type` set to `nacos`. There is no need to ask nacos for the service list, we already know it via the upstream configuration.

spacewander: You can take a look at https://github.com/apache/apisix/blob/49e4fd307c65dc125e62987b3634a9e7358179bf/apisix/control/v1.lua#L115-L122
for how we iterate the upstream configurations and collect the health check info.
It is similar to collect the service_name.

apisix/discovery/nacos.lua
benx203: If use @,when run test file,will throw error:
Possible unintended interpolation of @127 in string at t/discovery/nacos.t line 70

spacewander: You need to escape it as `\@` in the template string.

spacewander:
Please merge master to resolve conflicts.

apisix/discovery/nacos.lua
benx203: OK.

apisix/discovery/nacos.lua
benx203: If import these in nacos.lua:
local upstream_mod = require("apisix.upstream")
local get_upstreams = upstream_mod.upstreams

When run test file,will throw error:
attempt to index local 'upstream_mod' (a number value)

It's not right import?

spacewander: I have mentioned it before: https://github.com/apache/apisix/pull/3820#discussion_r599481973

> You might need some lazy load trick to break the dependent cycle.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,323 @@
...
+
+
+local _M = {
+    version = 0.1,
+}
+
+
+local function service_info()
+    local host = local_conf.discovery and
+            local_conf.discovery.nacos and local_conf.discovery.nacos.host
```
Firstsawyou, line 84: I think this style is better：
```suggestion
    local host = local_conf.discovery and local_conf.discovery.nacos
                 and local_conf.discovery.nacos.host
```

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,323 @@
...
+        headers['Content-Type'] = 'application/json'
+    end
+
+    local httpc = http.new()
+    local timeout = local_conf.discovery.nacos.timeout
+    local connect_timeout = timeout.connect
+    local send_timeout = timeout.send
+    local read_timeout = timeout.read
+    log.info("connect_timeout:", connect_timeout, ", send_timeout:", send_timeout,
+        ", read_timeout:", read_timeout, ".")
```
Firstsawyou, line 144: Bad style, this would be better:
```suggestion
    log.info("connect_timeout:", connect_timeout, ", send_timeout:", send_timeout,
             ", read_timeout:", read_timeout, ".")
```

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,323 @@
...
+        end
+        local other = string_sub(url, auth_idx + 1)
+        url = protocol .. other
+    end
+    if local_conf.discovery.nacos.prefix then
+        url = url .. local_conf.discovery.nacos.prefix
+    end
+    if str_byte(url, #url) ~= str_byte("/") then
+        url = url .. "/"
+    end
```
Firstsawyou, line 111: It is better to keep a blank line between if conditional statements.
```suggestion
    if local_conf.discovery.nacos.prefix then
        url = url .. local_conf.discovery.nacos.prefix
    end
    
    if str_byte(url, #url) ~= str_byte("/") then
        url = url .. "/"
    end
```

apisix/discovery/nacos.lua
benx203: Could you give me some demo code?

spacewander: Don't worry, I can update it for you once you have solved the other issues.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,326 @@
...
+    else
+        token_param = ""
+    end
+end
+
+
+local function service_info()
+    local host = local_conf.discovery and local_conf.discovery.nacos
+                 and local_conf.discovery.nacos.host
+    if not host then
```
spacewander, line 164: There is no need to check the host repeatedly. We can check it in the schema.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,326 @@
...
+        applications = up_apps
+        return
+    end
+
+    local data, err
+    for _, service_name in ipairs(infos) do
+        data, err = get_url(base_uri, instance_list_path .. service_name .. token_param)
+        if err then
+            log.error("get_url:" .. instance_list_path .. " err:" .. err)
+            applications = up_apps
```
spacewander, line 263: IMHO, when something bad happened, we should keep the `application` unchanged. Stale configuration is better than no configuration.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+local ngx_timer_every    = ngx.timer.every
+local string             = string
+local string_sub         = string.sub
+local str_byte           = string.byte
+local str_find           = core.string.find
+local str_format         = string.format
+local log                = core.log
+
+local default_weight
+local applications
```
spacewander, line 39: We can initialize `applications = {}` so that we can save many codes like:

```
        if not applications then
            applications = up_apps
        end
```

benx203: If use applications = {},run test could fail cause maybe call _M.nodes method before finish init.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+        end
+    end
+    applications = up_apps
+end
+
+
+function _M.nodes(service_name)
+    while( not applications )
+    do
+        log.info('wait init')
```
spacewander, line 283: Better to use a `WARN` level log and only log for the first sleep.
`info` level is invisible by default.

benx203: OK.
BTW,when i run test file in local CentOS7,most time success,sometime will throw error:
 [error] 23302#23302: *10 [lua] init.lua:462: http_access_phase(): failed to set upstream: no valid upstream node: nil, client: 127.0.0.1, server: localhost, request: "GET /hello HTTP/1.1", host: "localhost"
Is it a bug?

apisix/discovery/nacos.lua
```diff
@@ -278,9 +278,13 @@ end
 
 
 function _M.nodes(service_name)
-    while( not applications )
+    local loged = false
```
spacewander, line 281: Should be `logged`

apisix/discovery/nacos.lua
```diff
@@ -278,9 +278,13 @@ end
...
+    while(not applications)
```
spacewander, line 282: Style: better to write `while not applications do`.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,319 @@
...
+        core.table.insert(infos, service_name)
+    end
+    return data, err, path
+end
+
+
+local function iter_and_add_service_info(infos, base_uri, token_param)
+    local data, err, path = get_page_service(infos, base_uri, token_param, 1)
+    if err then
+        log.error("get_url:" .. path .. " err:" .. err)
```
membphis, line 203: bad code style

this is the right style: `log.error("get_url:", path, " err:", err)`

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,319 @@
...
+        return
+    end
+
+    local maxPage = math.ceil(data.count / 100)
+    if maxPage == 0 then
```
membphis, line 208: magic number, need some comments

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,319 @@
...
+        return
+    end
+
+    if maxPage > 1 then
```
membphis, line 212: ditto

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,319 @@
...
+
+local function fetch_full_registry(premature)
+    if premature then
+        return
+    end
+
+    local up_apps = core.table.new(0, 0)
+    local base_uri, username, password = get_base_uri()
+    local token_param, err = get_token_param(base_uri, username, password)
+    if err then
```
membphis, line 235: we need to writhe this error message to error log

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,319 @@
...
+    if #infos == 0 then
+        applications = up_apps
+        return
+    end
+
+    local data, err
+    for _, service_name in ipairs(infos) do
+        data, err = get_url(base_uri, instance_list_path .. service_name .. token_param)
+        if err then
+            log.error("get_url:" .. instance_list_path .. " err:" .. err)
```
membphis, line 252: bad code style

t/discovery/nacos.t
```diff
@@ -0,0 +1,152 @@
...
+routes:
+  -
+    uri: /hello
+    upstream:
+      service_name: APISIX-NACOS
+      discovery_type: nacos
+      type: roundrobin
+
+#END
+--- pipelined_requests eval
```
spacewander, line 78: Maybe we can use sleep before sending HTTP requests?
Like this: https://github.com/apache/apisix/blob/252af4143d527b5a644c0a4959b96504099237e2/t/discovery/consul_kv.t#L307
Don't forget to set a long "--- timeout".

It look like syncing configuration from Nacos is time-consuming job.

t/discovery/nacos.t
```diff
@@ -0,0 +1,152 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
+log_level('info');
+no_root_location();
+no_shuffle();
+
```
spacewander, line 23: We can add this block to check the error log:
```
add_block_preprocessor(sub {
    my ($block) = @_;

    if (!$block->no_error_log) {
    }
});
```

It would be helpful for debugging the 503.

benx203: After add this,it throws error when run test file in my local pc:
pattern "[error]" should not match any line in error.log but matches line "2021/04/08 05:39:35 [error] 25739#25739: *3 [lua] init.lua:462: http_access_phase(): failed to set upstream: no valid upstream node: nil, client: 127.0.0.1, server: localhost, request: \"GET /hello HTTP/1.1\", host: \"localhost\"" (req 0)

t/discovery/nacos.t
benx203: Should be useful,thanks!

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,322 @@
...
+        if err then
+            log.error("get_url:", instance_list_path, " err:", err)
+            if not applications then
+                applications = up_apps
+            end
+            return
+        end
+
+        for _, host in ipairs(data.hosts) do
+            if tostring(host.valid) == 'true' and
```
spacewander, line 263: I set up the environment and debug it myself.
Here is the `host`: {\"enabled\":true,\"instanceHeartBeatTimeOut\":15000,\"ipDeleteTimeout\":30000,\"instanceHeartBeatInterval\":5000,\"metadata\":{\"preserved.register.source\":\"SPRING_CLOUD\"},\"port\":18001,\"serviceName\":\"DEFAULT_GROUP@@APISIX-NACOS\",\"instanceId\":\"192.168.65.3#18001#DEFAULT#DEFAULT_GROUP@@APISIX-NACOS\",\"healthy\":true,\"ip\":\"192.168.65.3\",\"ephemeral\":true,\"clusterName\":\"DEFAULT\",\"weight\":1}

There is not `valid` field. The response doesn't match their API doc: https://nacos.io/en-us/docs/open-api.html

I am not sure if it is because the latest Nacos has changed the API. Maybe we need to use the tagged nacos image `2.0.0` instead of the latest one?

.github/workflows/centos7-ci.yml
```diff
@@ -101,6 +101,37 @@ jobs:
         docker run --rm --name consul_1 -d -p 8500:8500 consul:1.7 consul agent -server -bootstrap-expect=1 -client 0.0.0.0 -log-level info -data-dir=/consul/data
         docker run --rm --name consul_2 -d -p 8600:8500 consul:1.7 consul agent -server -bootstrap-expect=1 -client 0.0.0.0 -log-level info -data-dir=/consul/data
 
+        # start nacos server
+        nohup docker network rm nacos_net > /dev/null 2>&1 &
+        nohup docker network create nacos_net > /dev/null 2>&1 &
+        # nacos no auth server - for test no auth
+        docker run --rm -d --name nacos_no_auth --network nacos_net --hostname nacos2 --env NACOS_SERVERS="nacos1:8848 nacos2:8848" --env PREFER_HOST_MODE=hostname --env MODE=cluster --env EMBEDDED_STORAGE=embedded  --env JVM_XMS=512m --env JVM_XMX=512m --env JVM_XMN=256m -p8858:8848 nacos/nacos-server:latest
```
spacewander, line 108: Would it be better to use `nacos/nacos-server:2.0.0`?

.github/workflows/centos7-ci.yml
```diff
@@ -101,6 +101,37 @@ jobs:
...
+        url="127.0.0.1:8858/nacos/v1/ns/service/list?pageNo=1&pageSize=2"
+        until  [[ "$(curl -s -o /dev/null -w ''%{http_code}'' $url)"  == "200" ]]; do
+          echo 'wait nacos server...'
+          sleep 1;
+        done
+        # register nacos service
+        rm -rf tmp
+        mkdir tmp
+        cd tmp
+        wget https://raw.githubusercontent.com/benx203/nacos-test-service/main/spring-nacos-1.0-SNAPSHOT.jar
```
spacewander, line 120: @benx203 
Would you also submit the Java code to https://github.com/benx203/nacos-test-service? Thanks!

benx203: OK.

.github/workflows/centos7-ci.yml
benx203: The nacos version should be 1.4.1,2.0.0 just released few days ago.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,322 @@
...
+                    tostring(host.healthy) == 'true' and
+                    tostring(host.enabled) == 'true' then
```
spacewander, line 265: One more question: why we need to filter the result according to those fields?

.github/workflows/centos7-ci.yml
spacewander: @benx203
Why we can't support 2.0.0 directly? Is `2.0.0` incompatible with `1.4.1`?

apisix/discovery/nacos.lua
benx203: If query all records,need to filter.
But now should not need any more.

spacewander:
@benx203
I will work on this PR and fulfill my promise in a few days. Please be patient.

benx203:
> @benx203
> I will work on this PR and fulfill my promise in a few days. Please be patient.

OK.Take your time.

.github/workflows/centos7-ci.yml
benx203: I think better way is support both 2.0.0 and 1.4.1.
Maybe support 2.0.0 later?

spacewander (approved):

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+local instance_list_path = "ns/instance/list?healthyOnly=true&serviceName="
+
+local schema = {
+    type = "object",
+    properties = {
+        host = {
+            type = "array",
+            minItems = 1,
+            items = {
+                type = "string",
```
membphis, line 49: add a pattern to limit the string to the host format

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+            },
+        },
+        fetch_interval = {type = "integer", minimum = 1, default = 30},
+        prefix = {type = "string", default = "/nacos/v1/"},
```
membphis, line 53: ditto

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+                read = 5000,
+            }
+        },
+    },
+    required = {"host"}
+}
+
+
+local _M = {
+    version = 0.1,
```
membphis, line 74: this version is useless, we can drop it

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+}
+
+
+local function request(request_uri, path, body, method, basic_auth)
+    local url = request_uri .. path
+    log.info("request url:", url)
+    local headers = core.table.new(0, 0)
+    headers['Accept'] = 'application/json'
```
membphis, line 82: for the string object, I think we use `"`, it is easier for reading.

`headers["Accept"] = "application/json"`

membphis: and please fix the similar points

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+end
+
+
+local function post_url(request_uri, path, body)
+    return request(request_uri, path, body, "POST", nil)
+end
+
+
+local function get_token_param(base_uri, username, password)
+    if username and password then
```
membphis, line 139: better style:

```lua
if not username or not password then
    return ""
end

...
```

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+        local data, err = post_url(base_uri, auth_path .. "?username=" .. username
```
membphis, line 140: I think we should call [ngx.encode_args](https://github.com/openresty/lua-nginx-module#ngxencode_args) for `username` and `password`

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+    if str_byte(url, #url) ~= str_byte("/") then
+        url = url .. "/"
+    end
+
+    return url, username, password
+end
+
+
+local function iter_and_add_service(services, values)
+	if not values then
```
membphis, line 184: bad indentation

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+    end
+    return ""
+end
+
+
+local function get_base_uri()
+    local host = local_conf.discovery.nacos.host
+    -- TODO Add health check to get healthy nodes.
+    local url = host[math_random(#host)]
+    local auth_idx = str_find(url, "@")
```
membphis, line 156: we'd better use the plain text model, better performance.

https://stackoverflow.com/questions/18973019/is-there-a-lua-string-find-without-pattern

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,316 @@
...
+            })
+        end
+    end
+    applications = up_apps
+end
+
+
+function _M.nodes(service_name)
+    local logged = false
+    while not applications do
```
membphis, line 279: I think we need to set a maximum waiting time , eg: 5 seconds

moonming (approved):

moonming:
@benx203 thanks for your contribution.
You can create a new PR to fix the code style issue.
And it will be better if you can add `support Nacos` in the README.md to let more people know about this feature.

benx203:
> @benx203 thanks for your contribution.
> You can create a new PR to fix the code style issue.
> And it will be better if you can add `support Nacos` in the README.md to let more people know about this feature.

OK,i will try,thank you all.

apisix/discovery/nacos.lua
```diff
@@ -275,41 +273,33 @@ end
 
 
 function _M.nodes(service_name)
-    local logged = false
```
spacewander: The original code is good. We just need to set a 5s limitation.

apisix/discovery/nacos.lua
```diff
@@ -0,0 +1,306 @@
...
+    end
+    return '&accessToken=' .. data.accessToken
+end
+
+
+local function get_base_uri()
+    local host = local_conf.discovery.nacos.host
+    -- TODO Add health check to get healthy nodes.
+    local url = host[math_random(#host)]
+    local auth_idx = str_find(url, '@', 1, true)
```
spacewander, line 160: The `core.string.find` is already the plain string find... 😅 
Not all the suggestion from the reviewer is correct.

apisix/discovery/nacos.lua
benx203: Just keep simple.

apisix/discovery/nacos.lua
```diff
@@ -182,47 +179,47 @@ local function get_base_uri()
...
+
+    -- here we use lazy load to work around circle dependency
+    local get_upstreams = require("apisix.upstream").upstreams
+    local get_routes = require("apisix.router").http_routes
+    local get_services = require("apisix.http.service").services
+
+    local values = get_upstreams()
+    iter_and_add_service(services, values)
+    values = get_routes()
+    iter_and_add_service(services, values)
```
benx203, line 224: By the way,Are there 'discovery_type' config in  service and route?
@spacewander

spacewander: Those can have embedded upstream field.

apisix/discovery/nacos.lua
spacewander: The lazy init is important as the sync is slow.

benx203: OK,i rollback the code.

apisix/discovery/nacos.lua
benx203: I see,thanks.

PR #3831, opened 2021-03-15
chore: the route_name and service_name in ctx should remain different

### LINKED ISSUE #3826: chore:  the route_name and service_name in ctx should remain different
### Issue description

Prerequisite: `route_name`(my-route) is configured in `route`, `service_name`(my-service) is configured in `service`, route_name is different from service_name, and `service` is bound to `route` .

Bug: In the plug-in, output `ctx` through core.log.warn(core.json.encode(ctx, true)), and found that ctx.route_name is the same as ctx.service_name, and the value is `my-service`.

Reason: During the `merge_service_route` operation, the `route_name` was not merged, resulting in the route_name and service_name being the same. For details, please see the following code snippet:
https://github.com/apache/apisix/blob/master/apisix/plugin.lua#L385-L421
https://github.com/apache/apisix/blob/master/apisix/init.lua#L342-L360

I think route_name and service_name should keep their respective values, they should be different.

Firstsawyou:

close #3826

### Pre-submission checklist:

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -416,6 +416,10 @@ local function merge_service_route(service_conf, route_conf)
         new_conf.value.script = route_conf.value.script
     end
 
+    if route_conf.value.name then
+        new_conf.value.name = route_conf.value.name
+    end
+
     -- core.log.info("merged conf : ", core.json.delay_encode(new_conf))
     return new_conf
 end
--- a/t/core/ctx.t
+++ b/t/core/ctx.t
@@ -909,3 +909,109 @@ GET /hello
 hello world
 --- error_log eval
 qr/request log: \{"route_id":"1"\}/
+
+
+
+=== TEST 34: add plugin metadata "route_name" and "service_name"
+--- config
+    location /t {
+        content_by_lua_block {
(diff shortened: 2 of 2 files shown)


apisix/plugin.lua
```diff
@@ -416,6 +416,10 @@ local function merge_service_route(service_conf, route_conf)
         new_conf.value.script = route_conf.value.script
     end
 
+    if route_conf.value.name then
```
spacewander, line 419: Can we use `new_conf.value.name = route_conf.value.name` directly?

t/core/ctx.t
```diff
@@ -909,3 +909,109 @@ GET /hello
...
+[error]
+
+
+
+=== TEST 35: add service and route, the service is bound to the route
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/services/1',
```
spacewander, line 959: We can put TEST 35 into TEST 34.

Firstsawyou: updated.

spacewander (approved):

tokers (approved):

Issue #3787, opened 2021-03-09
[discuss]: does the traffic-split plugin has a way to implement the original capabilities without corrupting the data?


tzssangglass:
### Issue description
prerequisite: the traffic-split plugin now controls the request proxy to a specific upstream by corrupting the upstream configuration in the route.

topic: is there a way to do the same thing without corrupting the upstream configuration in the route?

if there are problems with the prerequisite, please correct me

reference #3758

moonming:
@Firstsawyou

membphis:
ping @Firstsawyou

Firstsawyou:
@tzssangglass  uses the method of `"when the route is configured with an upstream_id, the upstream_id is passed into the server_list"` to solve it.
Thank you very much @tzssangglass .

Firstsawyou:
I think we can create a temporary variable to cache the upstream_id configured in the plug-in, and delete it after the end of use, so as to avoid the modification of `ctx.matched_route.value.upstream_id`. For example, create a temporary variable upstream_id_tmp: `ctx.matched_route.value.upstream_id_tmp`

PR #3842, opened 2021-03-16
fix(traffic-split): binding upstream via upstream_id is invalid

### LINKED ISSUE #3740: bug: the traffic-split plugin is invalid to bind upstream via upstream_id

Firstsawyou:

fix #3740
Related issues: #3787 

Because the current traffic-split plug-in supports `upstream_id` to refer to the upstream by rewriting the value of `ctx.matched_route.value.upstream_id`. However, when the upstream_id is used by the route and the plugin at the same time, the rewriting of `ctx` will cause the `upstream_id` on the original cached route to be lost.
Therefore, this PR creates a temporary variable of `ctx.upstream_id` to cache the upstream_id configured in the plug-in, avoiding the modification of `ctx.matched_route.value.upstream_id`.

### Pre-submission checklist:

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -391,7 +391,6 @@ function _M.http_access_phase()
     -- used for the traffic-split plugin
     if api_ctx.upstream_id then
         up_id = api_ctx.upstream_id
-        api_ctx.upstream_id = nil
     end
 
     if up_id then
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -387,6 +387,13 @@ function _M.http_access_phase()
     end
 
     local up_id = route.value.upstream_id
+
+    -- used for the traffic-split plugin
+    if route.value.upstream_id_tmp then
+        up_id = route.value.upstream_id_tmp
+        route.value.upstream_id_tmp = nil
+    end
+
     if up_id then
         local upstreams = core.config.fetch_created_obj("/upstreams")
         if upstreams then
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -389,9 +389,9 @@ function _M.http_access_phase()
     local up_id = route.value.upstream_id
 
     -- used for the traffic-split plugin
-    if route.value.upstream_id_tmp then
-        up_id = route.value.upstream_id_tmp
-        route.value.upstream_id_tmp = nil
+    if api_ctx.upstream_id then
+        up_id = api_ctx.upstream_id
+        api_ctx.upstream_id = nil
     end
 
     if up_id then
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -226,7 +226,7 @@ local function set_upstream(upstream_info, ctx)
 end
 
 
-local function new_rr_obj(weighted_upstreams, route_upstream_id)
+local function new_rr_obj(weighted_upstreams)
     local server_list = {}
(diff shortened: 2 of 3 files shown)


apisix/init.lua
```diff
@@ -387,6 +387,13 @@ function _M.http_access_phase()
     end
 
     local up_id = route.value.upstream_id
+
+    -- used for the traffic-split plugin
+    if route.value.upstream_id_tmp then
```
spacewander, line 392: Maybe we can set & use `ctx.upstream_id` directly. No more modification to the route.

spacewander (changes requested):

apisix/init.lua
Firstsawyou: Do you mean to directly set the variable of `ctx.upstream_id` in the traffic-split plugin to cache the upstream_id in the plugin?

spacewander: Instead of setting `ctx.matched_route.value.upstream_id_tmp`, can we set `ctx.upstream_id` directly?

Firstsawyou: This can directly set `ctx.upstream_id`, I will update the code.

Firstsawyou: updated.

apisix/init.lua
```diff
@@ -387,6 +387,13 @@ function _M.http_access_phase()
     end
 
     local up_id = route.value.upstream_id
+
+    -- used for the traffic-split plugin
+    if api_ctx.upstream_id then
+        up_id = api_ctx.upstream_id
+        api_ctx.upstream_id = nil
```
spacewander, line 394: We don't need to remove it?

Firstsawyou: Yes, we can choose not to delete it.

Firstsawyou: updated.

spacewander (approved):

tokers (approved):

PR #3983, opened 2021-04-04
docs: add ip support into route host field

starsz:

Now, we can fill the IP in the route host field to match.
Add doc to introduce it.

### Pre-submission checklist:


docs/en/latest/admin-api.md
```diff
@@ -74,7 +74,7 @@ Note: When the `Admin API` is enabled, it will occupy the API prefixed with `/ap
...
-| host             | False, can't be used with `hosts`        | Match Rules | Currently requesting a domain name, such as `foo.com`; PAN domain names such as `*.foo.com` are also supported.                                                                                                                                                                                                                                                                                                                    | "foo.com"                                            |
+| host             | False, can't be used with `hosts`        | Match Rules | Currently requesting a domain name or IP address, such as `foo.com`, `127.0.0.1` ; PAN domain names such as `*.foo.com` are also supported, `*.0.0.1` also.                                                                                                                                                                                                                                                                                                                    | "foo.com"                                            |
```
juzhiyuan, line 77: CIDR?

starsz: Not CIDR.
CIDR is like: "a.b.c.0/24"
But this is like: "*.0.0.1"
😀

spacewander: Better not mention this "feature" as it is supported unexpectedly. It just happens to work.

spacewander (changes requested):

spacewander:
The IP address should not be support as the host. If this does something harmful, we should consider rejecting it via validation.

docs/en/latest/admin-api.md
starsz: OK.Got it.

juzhiyuan:
cc @moonming to know this.

moonming:
> The IP address should not be supported as the host. If this does something harmful, we should consider rejecting it via validation.

agreed.  cc @membphis

PR #4004, opened 2021-04-08
feat: use route/service name instead of id in prometheus metric

Yiyiyimu:

So users could get better idea about specific route/services.

Also this PR delete the situation that when no route is found, it would use host as service id.

### Pre-submission checklist:


spacewander (changes requested):
We should provide an option to enable this feature.
Additionally, people can configure multiple routes with the same name.

Yiyiyimu:
@spacewander PTAL

t/plugin/prometheus.t
```diff
@@ -809,3 +809,90 @@ qr/apisix_/
 qr/etcd/
 --- no_error_log
 [error]
+
+
+
+=== TEST 45: create service with name
```
spacewander, line 815: 1. New test should be added to prometheus2.t 
2. Test 45 & Test 46 should be merged into one
3. We need another two tests: one for route with name & one for name preferred but name missed.

apisix/plugins/prometheus.lua
```diff
@@ -23,6 +23,13 @@ local plugin_name = "prometheus"
 local default_export_uri = "/apisix/prometheus/metrics"
 local schema = {
     type = "object",
+    properties = {
+        route_type = {
```
spacewander, line 27: `prefer_name` would be better?

t/plugin/prometheus2.t
```diff
@@ -78,6 +78,7 @@ passed
 [401, 401, 401, 401]
 --- no_error_log
 [error]
+-- LAST
```
Firstsawyou, line 81: We should remove `-- LAST`.

Firstsawyou (changes requested):

apisix/plugins/prometheus/exporter.lua
```diff
@@ -133,9 +139,10 @@ function _M.log(conf, ctx)
         matched_host = ctx.curr_req_matched._host or ""
     end
 
+
     metrics.status:inc(1,
         gen_arr(vars.status, route_id, matched_uri, matched_host,
-                service_id, consumer_name, balancer_ip))
+        service_id, consumer_name, balancer_ip))
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks! Fixed

t/plugin/prometheus.t
Yiyiyimu: Fixed. PTAL

apisix/plugins/prometheus.lua
Yiyiyimu: Thanks for the suggestion and I finally choose `prefer_type` since it could include `id` and `name`

t/plugin/prometheus2.t
Yiyiyimu: Thanks for remind! Fixed

apisix/plugins/prometheus.lua
```diff
@@ -23,6 +23,13 @@ local plugin_name = "prometheus"
 local default_export_uri = "/apisix/prometheus/metrics"
 local schema = {
     type = "object",
+    properties = {
+        prefer_type = {
```
spacewander, line 27: I mean using a boolean `prefer_name`. The type is confusing, why `id` & `name` are types?

Yiyiyimu: Ah boolean I didn't realize that. Yes that would be better.

docs/en/latest/plugins/prometheus.md
```diff
@@ -25,7 +25,9 @@ This plugin exposes metrics in Prometheus Exposition format.
 
 ## Attributes
 
-none.
+| Name         | Type   | Requirement | Default  | Valid          | Description                                            |
+| ------------ | ------ | ----------- | -------- | -------------- | ------------------------------------------------------ |
+| prefer_type   | string | optional    | "id"     | ["id", "name"] | output type of route/service in prometheus metric.     |
```
spacewander, line 30: Should update the doc too

apisix/plugins/prometheus/exporter.lua
```diff
@@ -115,15 +116,20 @@ function _M.log(conf, ctx)
 
     local route_id = ""
     local balancer_ip = ctx.balancer_ip or ""
-    local service_id
+    local service_id = ""
```
spacewander, line 119: Better to use `service_id = vars.host` to be compatible

docs/en/latest/plugins/prometheus.md
Yiyiyimu: Sorry I forgot. Updated.

apisix/plugins/prometheus/exporter.lua
Yiyiyimu: Resolved since host is not related to service

spacewander (approved):

starsz (approved):

Yiyiyimu:
@spacewander PTAL thanks!

Yiyiyimu:
@spacewander found two other related problems that need some help
1. One situation that the test didn't cover is when set service name, but not route name. The actual output is we make the route use service name where the change happens [here](https://github.com/apache/apisix/blob/baf843403461883c1334e63d15a6bb3622c31940/apisix/plugin.lua#L419-L421). Since I notice there's a recent change about this part(#3831), maybe we need to further change the logic, so route name would not be interfered with by service name.
2. Found another [place](https://github.com/apache/apisix/blob/baf843403461883c1334e63d15a6bb3622c31940/apisix/utils/log-util.lua#L38) that set `service_id` to `host` when route is nil. Do we need to also remove this part.

PR #4020, opened 2021-04-10
fix: when route name is not set, it would be given the service name

Yiyiyimu:

Related to #4004, when only set service name but not route name, both of them would use service name in prometheus metric. 

The problem is partially fixed in #3831, so when route name is set it could be fetched. But it does not cover the above situation.

### Pre-submission checklist:


spacewander:
@Yiyiyimu 
Need to fix the lint

spacewander (approved):

nic-chen (approved):

PR #4034, opened 2021-04-12
feat: support client certificate verification

spacewander:

### Pre-submission checklist:


membphis (approved):

apisix/ssl/router/radixtree_sni.lua
```diff
@@ -194,6 +195,24 @@ function _M.match_and_set(api_ctx)
         end
     end
 
+    if matched_ssl.value.client then
+        local client_cert = matched_ssl.value.client.ca
```
tokers, line 199: The name `client_cert` is confusing, actually, it's the CA cert(s) to verify the client cert. What about `ca_cert`.

apisix/schema_def.lua
```diff
@@ -642,6 +642,18 @@ _M.ssl = {
             type = "array",
             items = private_key_schema,
         },
+        client = {
+            type = "object",
+            properties = {
+                ca = certificate_scheme,
```
tokers, line 648: We may support CA cert chains in the future.

apisix/admin/ssl.lua
```diff
@@ -70,6 +70,17 @@ local function check_conf(id, conf, need_id)
         end
     end
 
+    if conf.client then
+        if not apisix_ssl.support_client_verification() then
```
tokers, line 74: This check is difficult when we using manager-api, unless we have a mechanism to advertise this capability to it (like rolling feature sets in the server-info plugin).

apisix/schema_def.lua
spacewander: Currently, OpenResty only supports a single CA cert chain. We can use `ca_list` or anything else if multiple chains are supported. However, using such a name now is confusing as we only support one.

apisix/admin/ssl.lua
spacewander: There is an additional check in the DP side. The `conf.client` will be ignored if the DP doesn't support it. The check here is not critical as the DP may be different with the Admin API.

docs/en/latest/admin-api.md
```diff
@@ -784,6 +784,8 @@ Return response from etcd currently.
...
 | certs       | False    | An array of certificate | when you need to configure multiple certificate for the same domain, you can pass extra https certificates (excluding the one given as cert) in this field |                                                  |
 | keys        | False    | An array of private key | https private keys. The keys should be paired with certs above                                                                                             |                                                  |
+| client.ca | False | Certificate| set the CA certificate which will use to verify client. This feature requires OpenResty 1.19+.                                                            |                                                  |
+| client.depth | False | Certificate| set the verification depth in the client certificates chain, default to 1. This feature requires OpenResty 1.19+.                                                            |                                                  |
```
Firstsawyou, line 788: The first letter should be capitalized.  set --> Set
```
set the CA certificate which will use to verify client.
```

t/node/client-mtls.t
```diff
@@ -0,0 +1,310 @@
...
+
+=== TEST 5: no client certificate
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin")
+            local json = require("toolkit.json")
+            local ssl_ca_cert = t.read_file("t/certs/mtls_ca.crt")
+            local ssl_cert = t.read_file("t/certs/mtls_client.crt")
+            local ssl_key = t.read_file("t/certs/mtls_client.key")
```
Firstsawyou, line 210: These certificates were not used in this test, I think they can be removed.

apisix/schema_def.lua
tokers: OK, just an advice and won't block this PR.

apisix/admin/ssl.lua
tokers: Got it.

docs/en/latest/admin-api.md
spacewander: None of the description in this table start with uppercase letter.

t/node/client-mtls.t
spacewander: @Firstsawyou 
Updated.

docs/en/latest/admin-api.md
Firstsawyou: Got it.

tokers (approved):

PR #4050, opened 2021-04-14
ci: cancel duplicate workflow to reduce CI queue time

Yiyiyimu:

We always wait for CI queue time, and we always trying to reduce waiting time. Here is another method to achieve it.

The method is that whenever a new commit added to PR, the previous running CI could be canceled since it is of no use. This could be achieved by using Github API, find the workflow on the same branch but with different SHA. And we could achieve this with [potiuk/cancel-workflow-runs](https://github.com/potiuk/cancel-workflow-runs) which is also [used in Apache Airflow](https://github.com/apache/airflow/blob/master/.github/workflows/build-images-workflow-run.yml).

To bypass the limit of third party Actions, [the recommendation provided by Apache Infra](https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status#GitHubActionsstatus-Security) is to use submodules.

Since the "cancel" workflow could only be triggered after merged, so I could not test it in this PR. Thus I duplicate the same situation in my own repo, and it works as expect. See https://github.com/Yiyiyimu/GithubActionsTest/pull/5 you could find previous running workflow got cancelled when a new one coming in. Also see https://github.com/Yiyiyimu/GithubActionsTest/runs/2344066993?check_suite_focus=true for cancellation detail.

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
         with:
+          persist-credentials: false
           submodules: recursive
 
       - name: Cache deps
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,6 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
         with:
-          persist-credentials: false
           submodules: recursive
 
       - name: Cache deps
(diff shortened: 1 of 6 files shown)


spacewander (approved):

tokers (approved):

Yiyiyimu:
Work as expect:
https://github.com/apache/apisix/actions/runs/754861826
image

PR #4069, opened 2021-04-16
feat: optimize skywalking instancename

### LINKED ISSUE #4040: skywalking plugin attribute service_instance_name hard-coded
### Issue description

Whether its attribute “service_instance_name” can be obtained automatically instead of hard-coded in the configuration

dickens7:
Fix #4040

### DIFF
--- a/apisix/plugins/skywalking.lua
+++ b/apisix/plugins/skywalking.lua
@@ -126,6 +126,11 @@ function _M.init()
 
     -- TODO: maybe need to fetch them from plugin-metadata
     local metadata_shdict = ngx.shared.tracing_buffer
+
+    if local_plugin_info.service_instance_name == "__hostname__" then
+        local_plugin_info.service_instance_name = core.utils.gethostname()
+    end
+
     metadata_shdict:set('serviceName', local_plugin_info.service_name)
     metadata_shdict:set('serviceInstanceName', local_plugin_info.service_instance_name)
 
--- a/apisix/plugins/skywalking.lua
+++ b/apisix/plugins/skywalking.lua
@@ -33,7 +33,7 @@ local metadata_schema = {
         service_instance_name = {
             type = "string",
             description = "User Service Instance Name",
-            default = "APISIX Instance Name",
+            default = core.utils.gethostname(),
         },
         endpoint_addr = {
             type = "string",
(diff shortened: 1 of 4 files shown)


apisix/plugins/skywalking.lua
```diff
@@ -33,7 +33,7 @@ local metadata_schema = {
         service_instance_name = {
             type = "string",
             description = "User Service Instance Name",
-            default = "APISIX Instance Name",
+            default = core.utils.gethostname(),
```
spacewander, line 36: We can't use this way as the default value need to be constant

spacewander (changes requested):

apisix/plugins/skywalking.lua
dickens7: let me see if there are other solutions

dickens7: Use placeholders to implement functionality

apisix/plugins/skywalking.lua
```diff
@@ -126,6 +126,11 @@ function _M.init()
 
     -- TODO: maybe need to fetch them from plugin-metadata
     local metadata_shdict = ngx.shared.tracing_buffer
+
+    if local_plugin_info.service_instance_name == "__hostname__" then
```
tokers, line 130: The variable style that we use in APISIX always with the prefix `$`. What about `$hostname`.

docs/en/latest/plugins/skywalking.md
```diff
@@ -90,9 +90,9 @@ We can set the endpoint by specified the configuration in `conf/config.yaml`.
 | Name         | Type   | Default  | Description                                                          |
 | ------------ | ------ | -------- | -------------------------------------------------------------------- |
 | service_name | string | "APISIX" | service name for skywalking reporter                                 |
-|service_instance_name|string|"APISIX Instance Name" | service instance name for skywalking reporter |
-| endpoint_addr| string | "http://127.0.0.1:12800" | the http endpoint of Skywalking, for example: http://127.0.0.1:12800 |
-| report_interval| integer | use the value in the skywalking client library | the report interval, in seconds |
+| service_instance_name | string |"APISIX Instance Name" | service instance name for skywalking reporter，  set it to `$hostname` to get local hostname directly.|
```
spacewander, line 93:
```suggestion
| service_instance_name | string |"APISIX Instance Name" | service instance name for skywalking reporter. If you want to use the machine's hostname directly, set it to `$hostname`.|
```

A little tweak for the doc

spacewander (approved):

tokers (approved):

lilien1010 (approved):

PR #4099, opened 2021-04-21
feat: jwt-auth support extension payload

### LINKED ISSUE #3594: how to set the JWT content as header values?
### Issue description
Hi,
I need apisix to validate a JWT token in the request header that was issued by my own auth-service and set the content of the JWT token (like user_id) as header arguments. I've read through the JWT doc section, but I can't find how to set the JWT content as header values.
### Environment

* apisix version (cmd: `apisix version`): latest
* OS (cmd: `uname -a`): ubuntu
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): latest
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
* apisix-dashboard version, if have:

benx203:

Make jwt-auth support extension payload.
Fix #3594 

### Pre-submission checklist:

* [X] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

### DIFF
--- a/apisix/plugins/jwt-auth.lua
+++ b/apisix/plugins/jwt-auth.lua
@@ -28,7 +28,6 @@ local ngx_time = ngx.time
 local sub_str  = string.sub
 local plugin_name = "jwt-auth"
 local pcall = pcall
-local log      = core.log
 
 
 local lrucache = core.lrucache.new({
@@ -105,7 +104,7 @@ do
         core.table.clear(consumer_names)
 
         for _, consumer in ipairs(consumers.nodes) do
-            log.info("consumer node: ", core.json.delay_encode(consumer))
+            core.log.info("consumer node: ", core.json.delay_encode(consumer))
             consumer_names[consumer.auth_conf.key] = consumer
         end
 
@@ -116,7 +115,7 @@ end -- do
 
 
 function _M.check_schema(conf, schema_type)
-    log.info("input conf: ", core.json.delay_encode(conf))
+    core.log.info("input conf: ", core.json.delay_encode(conf))
 
     local ok, err
     if schema_type == core.schema.TYPE_CONSUMER then
@@ -193,8 +192,8 @@ local function get_real_payload(key, auth_conf, payload)
         exp = ngx_time() + auth_conf.exp
     }
     if payload then
-        local payloadEx = core.json.decode(payload)
-        core.table.merge(real_payload, payloadEx)
+        local extra_payload = core.json.decode(payload)
+        core.table.merge(real_payload, extra_payload)
     end
     return real_payload
 end
@@ -213,7 +212,7 @@ local function sign_jwt_with_HS(key, auth_conf, payload)
         }
     )
     if not ok then
-        log.warn("failed to sign jwt, err: ", jwt_token.reason)
+        core.log.warn("failed to sign jwt, err: ", jwt_token.reason)
         core.response.exit(500, "failed to sign jwt")
     end
     return jwt_token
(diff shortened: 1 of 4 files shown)


apisix/plugins/jwt-auth.lua
```diff
@@ -104,7 +105,7 @@ do
         core.table.clear(consumer_names)
 
         for _, consumer in ipairs(consumers.nodes) do
-            core.log.info("consumer node: ", core.json.delay_encode(consumer))
+            log.info("consumer node: ", core.json.delay_encode(consumer))
```
spacewander, line 108: The original `core.log` works well. Please don't touch unrelative code.

apisix/plugins/jwt-auth.lua
```diff
@@ -186,48 +187,55 @@ local function get_secret(conf)
...
 
 
-local function sign_jwt_with_HS(key, auth_conf)
+local function get_real_payload(key, auth_conf, payload)
+    local real_payload = {
+        key = key,
+        exp = ngx_time() + auth_conf.exp
+    }
+    if payload then
+        local payloadEx = core.json.decode(payload)
```
spacewander, line 196: Better to use `extra_payload`

apisix/plugins/jwt-auth.lua
```diff
@@ -277,18 +285,44 @@ function _M.rewrite(conf, ctx)
...
         return 401, {message = jwt_obj.reason}
     end
 
     consumer_mod.attach_consumer(ctx, consumer, consumer_conf)
-    core.log.info("hit jwt-auth rewrite")
+    log.info("hit jwt-auth rewrite")
+end
+
+
+local function user_info()
```
spacewander, line 303: Is this function necessary for the core feature? We should only accept the necessary API to avoid bloating.

spacewander (changes requested):

apisix/plugins/jwt-auth.lua
benx203: Yes,the function provide ability to check up the payload content.

apisix/plugins/jwt-auth.lua
benx203: OK.

apisix/plugins/jwt-auth.lua
spacewander: This feature can be implemented as a local script to check.
There is no need to expose it as a public API of the gateway.

benx203: I get it,thanks.

spacewander:
Would you add two test cases with payload, one for default algorithm and another for RS256?

Like https://github.com/apache/apisix/blob/86e49837d20147471eeb852d60d4c5e2f15a2bd4/t/plugin/jwt-auth.t#L1180

Thanks!

t/plugin/jwt-auth.t
```diff
@@ -1209,7 +1241,39 @@ hello world
 
 
 
-=== TEST 43: test for unsupported algorithm
+=== TEST 44: sign / verify (algorithm = HS512,with extra payload)
```
spacewander, line 1244: Maybe you are misled by my example. We need two tests, one for default and another for RS256 (but not for HS512).

benx203: Test 19 is for RS256.

spacewander (approved):
Next time, please add new tests **at the bottom**.

benx203:
> Next time, please add new tests **at the bottom**.

@spacewander OK,get it.

tokers (approved):

PR #4106, opened 2021-04-22
fix: remove unexpected blocking in background timer

### LINKED ISSUE #3923: request help: timer enhancement
Currently, apisix provides a background timer for logger related plugins, whitch is not universal.

See this demand: a plugin may register a timer in `init` function, the timer will be recurring to pull data, and unregister it in `destroy` function, the phase functions may use these pulled data. But the current timer implemention is not possible to do that.

How can i do this, any suggestions?

spacewander:
Previously we actually run background timer each second.
And the backgroud timer will be blocked if there is not timer registered
To be compatible, now we increase the interval to 1 second.

Fix #3923

### Pre-submission checklist:

### DIFF
--- a/t/misc/timers.t
+++ b/t/misc/timers.t
@@ -41,8 +41,6 @@ __DATA__
     location /t {
         content_by_lua_block {
             local timers = require("apisix.timers")
-            local ngx_now = ngx.now
-            local start_at = ngx_now()
             timers.register_timer("t", function()
                 ngx.log(ngx.WARN, "fire")
             end)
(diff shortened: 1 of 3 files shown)


tokers (approved):

t/misc/timers.t
```diff
@@ -0,0 +1,60 @@
...
+
+__DATA__
+
+=== TEST 1: always start a new timer even the previous one is blocked
+--- config
+    location /t {
+        content_by_lua_block {
+            local timers = require("apisix.timers")
+            local ngx_now = ngx.now
+            local start_at = ngx_now()
```
Firstsawyou, line 45: These two variables do not seem to be used. How about removing them?

membphis (approved):

PR #4109, opened 2021-04-22
feat: support to fetch balancer ip/port from ctx.var

### LINKED ISSUE #4091: feature: support to fetch balancer `ip`/`port` from `ctx.vars`
we have stored the upstream ip/port in `ctx` object, here is the link: https://github.com/apache/apisix/blob/master/apisix/balancer.lua#L204-L205

we should support to fetch from `ctx.vars`, here is the example:

```lua
local up_ip = ctx.vars["balancer_ip"]
local up_port = ctx.vars["balancer_port"]
```

Then we can use this value in plugins(eg: SkyWalking).

I think we can use the same way as "consumer_name", which is the code link:

https://github.com/apache/apisix/blob/master/apisix/core/ctx.lua#L176

Firstsawyou:

Signed-off-by: Firstsawyou yuelinz99@gmail.com
close #4091

### Pre-submission checklist:

### DIFF
--- a/t/core/ctx2.t
+++ b/t/core/ctx2.t
@@ -123,12 +123,6 @@ post uri: /server_port
             ngx.say(body)
         }
     }
---- request
-GET /t
---- response_body
-passed
---- no_error_log
-[error]
 
 
 
--- a/t/core/ctx2.t
+++ b/t/core/ctx2.t
@@ -87,3 +87,58 @@ GET /xxx
 --- error_log
 pre uri: /hello
 post uri: /server_port
+
+
+
+=== TEST 3: get balancer_ip and balancer_port through ctx.var
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/hello",
+                    "plugins": {
+                        "serverless-post-function": {
+                            "phase": "log",
+                            "functions" : ["return function(conf, ctx)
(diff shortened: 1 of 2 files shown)


t/core/ctx2.t
```diff
@@ -87,3 +87,58 @@ GET /xxx
...
+                }]]
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- request
```
spacewander, line 126: We don't need that stuff. They are already set in https://github.com/Firstsawyou/apisix/blob/a78bd10f4f1efb3547cae880d7cdea7d7ffb08e1/t/core/ctx2.t#L23

Firstsawyou: Got it, updated.

t/core/ctx2.t
```diff
@@ -87,3 +87,52 @@ GET /xxx
...
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/hello",
+                    "plugins": {
+                        "serverless-post-function": {
+                            "phase": "log",
+                            "functions" : ["return function(conf, ctx)
+                                        ngx.log(ngx.WARN, 'balancer_ip: ', ctx.var.balancer_ip);
```
tokers, line 106: Redundant semicolons.

spacewander (approved):

t/core/ctx2.t
Firstsawyou: Got it, updated.

tokers (approved):

Issue #4150, opened 2021-04-28
bug: Inconsistent behavior of schema check for response-rewrite plugin between AdminAPI and ManagerAPI


imjoey:
### Issue description

Regarding the `response-rewrite` plugin, the Admin API does additional checks for schema definition, such as checking whether the `body` is a valid base64 string if `body_base64` is set to `true`. While the Manager API only use `schema.json` for checking the type of properties. This would lead to incorrect data could be successfully submitted via the Dashboard, but errors occur when APISIX is trying to load them from etcd.

As said in https://github.com/openresty/lua-nginx-module/blob/master/t/013-base64.t, `ngx.decode_base64` will generates `500 Internal Server Error` for `nil` and `null` string for `""`.

### Environment

* apisix version (cmd: `apisix version`): 2.4 and master
* OS (cmd: `uname -a`): MacOS
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): nginx version: openresty/1.19.3.1
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
* apisix-dashboard version, if have: 2.6

### Minimal test code / Steps to reproduce the issue

1. In the dashboard, create a route with `response-rewrite` plugin enabled;
2. Configure the `response-rewrite` plugin to only consist of the `body_base64` equals to `true`
```json
{
     "body_base64" : true,
}
```
3. Submit the  configurations

### What's the actual result? (including assertion message & call stack if applicable)

1. The route creation could be successfully submitted.
2. APISIX(v2.4) continuously generate the following logs:
```
2021/04/28 14:45:09 [error] 348311#348311: *30 [lua] config_etcd.lua:552: failed to fetch data from etcd: ...xiran/product/apisix/apisix/plugins/response-rewrite.lua:84: string argument only
stack traceback:
  [C]: in function 'error'
  /usr/local/openresty/lualib/resty/core/base64.lua:98: in function 'decode_base64'
  ...xiran/product/apisix/apisix/plugins/response-rewrite.lua:84: in function 'check_schema'
  /home/liuxiran/product/apisix/apisix/plugin.lua:572: in function 'plugin_checker'
  /home/liuxiran/product/apisix/apisix/http/route.lua:106: in function 'checker'
  /home/liuxiran/product/apisix/apisix/core/config_etcd.lua:357: in function 'sync_data'
[...]

PR #4151, opened 2021-04-28
fix(response-rewrite): properly base64 process the nil and empty body

imjoey:

Related to #4150, this PR is going to generate a proper error response when the `body` is a `nil` or an `empty` string. This would prevent APISIX from getting uncaught errors, as the error log mentioned in #4150.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/response-rewrite.lua
+++ b/apisix/plugins/response-rewrite.lua
@@ -127,7 +127,7 @@ function _M.body_filter(conf, ctx)
         return
     end
 
-    if conf.body and #conf.body > 0 then
+    if conf.body then
 
         if conf.body_base64 then
             ngx.arg[1] = ngx.decode_base64(conf.body)
--- a/apisix/plugins/response-rewrite.lua
+++ b/apisix/plugins/response-rewrite.lua
@@ -100,6 +100,9 @@ function _M.check_schema(conf)
     end
 
     if conf.body_base64 then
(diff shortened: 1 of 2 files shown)


t/plugin/response-rewrite.t
```diff
@@ -629,3 +629,50 @@ hello world
...
+
+
+=== TEST 22: set an empty body with setting body_base64 to true
+--- config
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.response-rewrite")
+            local ok, err = plugin.check_schema({
+                            body = "",
+                            body_base64 =  true
```
tokers, line 642:
```suggestion
                            body_base64 = true
```

imjoey: @tokers thank you very much. 😄

tokers (approved):

apisix/plugins/response-rewrite.lua
```diff
@@ -124,7 +127,7 @@ function _M.body_filter(conf, ctx)
         return
     end
 
-    if conf.body then
+    if conf.body and #conf.body > 0 then
```
spacewander, line 130: Do we need the check here? We already check it in `check_schema`.

imjoey: @spacewander yes, this additional check is indeed redundant. Done. Thank you.

spacewander (approved):

PR #4180, opened 2021-05-04
fix: fix peer missed in skywalking

dmsolr:

So far, `peer` of the exit span is a hard code, "upstream service". This is PR intends to fix it by getting `ip:port` from dynamic balancer.

I tried to get the information at the `rewrite` phase, but it seems like cannot get it. Because of that, I move it into the `body_filter` phase.

### Pre-submission checklist:


dmsolr:
This PR is not ready, and I need to take much time to resolve how to get the exactly upstream in rewrite phase. For this reason, I close it before I get the way.

PR #4189, opened 2021-05-06
docs: avoid the misleading comment

### LINKED ISSUE #3962: [discuss]: how to make the annotations in the config-default.yaml file clearer
# Description:

Currently, the annotations style in the config-default.yaml file is very misleading, for example:

```yaml
router:
  http: 'radixtree_uri'         # radixtree_uri: match route by uri(base on radixtree)
                                  # radixtree_host_uri: match route by host + uri(base on radixtree)
                                  # radixtree_uri_with_parameter: like radixtree_uri but match uri with parameters,
                                  #   see https://github.com/api7/lua-resty-radixtree/#parameters-in-path for
                                  #   more details.
  ssl: 'radixtree_sni'          # radixtree_sni: match route by SNI(base on radixtree)
# stream_proxy:                 # TCP/UDP proxy
#   tcp:             
```
This annotations style is easy to understand that `stream_proxy` is at the same level as `ssl`, but `stream_proxy` is at the same level as `router`.
I think we need a better way of annotation to make it easier for users to understand the hierarchical relationship before them.
what do you think?

spacewander:
Also remove helpless yaml lint.

Fix #3962

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,23 +14,6 @@ jobs:
           node-version: '12.x'
       - run: npm install -g markdownlint-cli@0.25.0
       - run: markdownlint '**/*.md'
-  yamllint:
-    name: 🍺 YAML
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '3.x'  # Version range or exact version of a Python version to use, using SemVer's version range syntax
-          architecture: 'x64'  # optional x64 or x86. Defaults to x64 if not specified
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install yamllint
-      - name: 🧹 YAML Lint
-        run: |
-          # return non-zero exit code on warnings
-          yamllint --strict .
   trailing-whitespace:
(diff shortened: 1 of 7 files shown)


jbampton (commented):
Why are you removing this test ?  
Without the YAML lint test on the CI we will probably go backwards.
You can always just ignore YAML files or folders with yamllint or ignore tests in whole files.

The YAML comment syntax is configurable:

Screen Shot 2021-05-06 at 8 48 46 pm

spacewander:
If we ignore it, there is almost nothing left that can be checked.

I quite regard about accepting yamllint. Several months have been passed, it has been proved that the linter is helpless but causes misleading and inconvenient.

spacewander:
Don't try to sell us linter anymore...
We are not an experimental lab for such "modern" technologies.

jbampton:
You are wrong again since the comments disable rule is only one yamllint rule and there are many to check.

There are about 20 yamllint rules :)

Also most of the worlds top projects use yamllint like Apache Airflow and they know it's not helpless.

https://yamllint.readthedocs.io/en/latest/rules.html

Screen Shot 2021-05-06 at 9 15 08 pm

juzhiyuan (approved):

Yiyiyimu (approved):

Issue #4209, opened 2021-05-10
New feature suggestion:  request-id plugin supports "snowflake" algorithm


dickens7:
Add support for snowflake algorithm to Request-ID plugin, or provide a separate snowflake algorithm plugin.

tzssangglass:
This is a feature to be implemented, and as far as I know there is no suitable implementation of the snowflake algorithm in the lua language, see also this comment:
https://github.com/apache/apisix/blob/ac21146c356cc9f9702b79b55dc42bbb3dec6877/apisix/core/etcd.lua#L296-L299

dickens7:
I have my own request-id plugin implements a snowflake algorithm using https://github.com/stuartcarnie/lua-snowflake

spacewander:
@dickens7 
We can add more options in the request-id to control the id generation way.
PR is welcome!

spacewander:
And there is a question, how can it handle more than 1024 workers?

dickens7:
> @dickens7
> Thanks for your PR! I have just taken a quick look at it. Would you submit it as the part of request-id plugin before I can review it?

no problem
But now there are two problems
1. Is the storage location of worker number reasonable?
```lua
        local etcd_cli, prefix = core.etcd.new()
        local res, err = etcd_cli:grant(30)
        local data, err = etcd_cli:keepalive(res.body.ID)

        local prefix = prefix .. "/plugins/snowflake/{id}"
```
2. What to do when workers is greater than 1024

spacewander:
What about forking that library and modify it?
Currently,

machine bits: 10
per ms bits: 12
timestamp bits: 42

We can change it to

machine bits: 16 (65535 workers)
per ms bits: 10 (512 per ms, which is 512000 per second and is enough as a per worker QPS)
timestamp bits: 38 (8 years, we can change the start point to 2021-01-01)

Yiyiyimu:
> timestamp bits: 38 (8 years, we can change the start point to 2021-01-01)

Or we could implement it like [sonyflake](https://github.com/sony/sonyflake) to change the time interval from 1ms to 10ms, and then we would have ~80 years. But not sure if 512 per 10ms is enough for us

tokers:
> > timestamp bits: 38 (8 years, we can change the start point to 2021-01-01)
> 
[...]

PR #4212, opened 2021-05-11
[WIP] feat snowflake plugin

dickens7:
#4209 

This PR is a draft version, used to discuss implementation


moonming:
I am not sure that do we should add a plugin for request id?

Yiyiyimu:
> I am not sure that do we should add a plugin for request id?

Agreed. Since it's an alternative algorithm for request-id, it might be better to add an enum argument to request-id, instead of using it as another plugin

dickens7:
> > I am not sure that do we should add a plugin for request id?
> 
> Agreed. Since it's an alternative algorithm for request-id, it might be better to add an enum argument to request-id, instead of using it as another plugin

It is planned like this. Now we are solving the problem that the snowflake algorithm can only have 1024 workers

Yiyiyimu:
> It is planned like this. Now we are solving the problem that the snowflake algorithm can only have 1024 workers

Do you mind sharing the reason for "planned like this"

dickens7:
Refactor

PR #4332, opened 2021-05-28
feat: add real_ip_recursive to nginx.conf

keilon:
Enable to set real_ip_recursive to nginx.conf. When apisix is behind multiple layer network, the real client ip may not be recognized. See also [http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_recursive](http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_recursive)

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -229,6 +229,10 @@ http {
     real_ip_header {* http.real_ip_header *};
     {% end %}
 
+    {% if http.real_ip_recursive then %}
+    real_ip_recursive {* http.real_ip_recursive *};
+    {% end %}
+
     {% if real_ip_from then %}
     {% print("\nDeprecated: apisix.real_ip_from has been moved to nginx_config.http.real_ip_from. apisix.real_ip_from will be removed in the future version. Please use nginx_config.http.real_ip_from first.\n\n") %}
     {% for _, real_ip in ipairs(real_ip_from) do %}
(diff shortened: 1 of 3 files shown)


Firstsawyou (commented):
We should add corresponding test cases.  https://github.com/apache/apisix/blob/master/t/cli/test_main.sh

apisix/cli/ngx_tpl.lua
```diff
@@ -229,6 +229,10 @@ http {
     real_ip_header {* http.real_ip_header *};
     {% end %}
 
+    {% if http.real_ip_recursive then %}
+    real_ip_recursive {* http.real_ip_recursive *};
```
spacewander, line 233: Need to add a test like: https://github.com/apache/apisix/blob/9a40fd7cb497974f9d7bd75f128eed5dd238cdc6/t/cli/test_main.sh#L382-L399

t/cli/test_main.sh
```diff
@@ -621,3 +621,10 @@ if ! grep "keepalive_timeout 6s;" conf/nginx.conf > /dev/null; then
 fi
 
 echo "passed: found the keepalive related parameter in nginx.conf"
+
+if ! grep "real_ip_recursive off;" conf/nginx.conf > /dev/null; then
+    echo "failed: 'real_ip_recursive off;' not in nginx.conf"
+    exit 1
+fi
+
+echo "passed: disable real_ip_recursive by default in nginx.conf"
```
spacewander, line 630: Need a test to change its default value and check, like `client_max_body_size: 512m ` in my example.

keilon: Fixed.

tokers (approved):

spacewander (approved):

Issue #4157, opened 2021-04-29
bug: After installing the Apisix on ARM-based System,but it can't works properly.


Molio-tan:
### Issue description
I'm sorry to ask,have you deploy the Apisix successfully on ARM-based hardware,and it works properly?No, I found the Apisix installed successfully but it cannot works properly.

### Environment

* apisix version (cmd: `apisix version`): v2.4 or v2.5
* OS (cmd: `uname -a`): Linux test123 4.19.90-17.5.ky10.aarch64 #1 SMP Fri Aug 7 13:35:33 CST 2020 aarch64 aarch64 aarch64 GNU/Linux
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): v1.19.3.1
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API): v3.4.13
* apisix-dashboard version, if have:
* luarocks version, if the issue is about installation (cmd: `luarocks --version`): v3.6.0

### Minimal test code / Steps to reproduce the issue
I tried some method to install the Apisix,such as,`luarocks install apisix` that apisix v2.4 was installed successfully or succeed in using source code to install apisix v2.5. When I run apisix with `./bin/apisix start`,there are normal for the terminal but errors in `logs/error.log` like
[error.log](https://github.com/apache/apisix/files/6396102/error.log)

I use this [blog](https://zhuanlan.zhihu.com/p/84467919) to install apisix on 鲲鹏系统.

### What's the expected result?
I found the 'resty.worker.events' in direction` deps`
![image](https://user-images.githubusercontent.com/31587573/116501151-afc60e80-a8e2-11eb-91a6-06104ef072fa.png)
So what's wrong with it?
Thanks anyway.

imjoey:
Hi @Molio-tan, the error logs indicate that perhaps you've installed the OpenResty and APISIX to the incorrect path. Could you please see if the path `/usr/local/apisix/` and `/usr/local/openresty` both exist? Thanks.

tokers:
@Molio-tan How did you start APISIX? As your APISIX was installed under /root`, actually when you launch APISIX in such a path, APISIX will reject to start.

See https://github.com/apache/apisix/issues/3207 for more details.

Firstsawyou:
Judging from the log information, there are many lua files that were not found. It is likely that the installation is abnormal. You should provide specific installation steps.
```
[...]

PR #4362, opened 2021-06-02
feat: support customzing worker's user

ysc972:

fix apisix cannot workes on kylin v10 Chinese os
#4157 
### Pre-submission checklist:

* [ x ] Did you explain what problem does this PR solve? Or what new features have been added?

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -19,6 +19,8 @@ return [=[
 # Configuration File - Nginx Server Configs
 # This is a read-only file, do not try to modify it.
 
+user root;
+
 master_process on;
 
 worker_processes {* worker_processes *};
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -18,9 +18,9 @@
 return [=[
 # Configuration File - Nginx Server Configs
 # This is a read-only file, do not try to modify it.
-
-user root;
-
+{% if user and user ~= '' then %}
+user {* user *};
+{% end %}
 master_process on;
 
 worker_processes {* worker_processes *};
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -130,6 +130,9 @@ apisix:
   disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
 
 nginx_config:                     # config for render the template to generate nginx.conf
+  #user: "root"                   # specifies the execution user of worker process.
+                                  # the "user" directive makes sense only if the master process runs with super-user privileges.
+                                  # if you is not root user,the default is current user.
   error_log: "logs/error.log"
   error_log_level: "warn"         # warn,error
   worker_processes: auto          # one worker will get best performance, you can use "auto", but remember it is just work well only on physical machine
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -132,7 +132,7 @@ apisix:
 nginx_config:                     # config for render the template to generate nginx.conf
   #user: "root"                   # specifies the execution user of worker process.
                                   # the "user" directive makes sense only if the master process runs with super-user privileges.
-                                  # if you is not root user,the default is current user.
+                                  # if you're not root user,the default is current user.
   error_log: "logs/error.log"
   error_log_level: "warn"         # warn,error
   worker_processes: auto          # one worker will get best performance, you can use "auto", but remember it is just work well only on physical machine
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -130,7 +130,7 @@ apisix:
   disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
 
 nginx_config:                     # config for render the template to generate nginx.conf
-  #user: "root"                   # specifies the execution user of worker process.
+  #user: "root"                   # specifies the execution user of the worker process.
                                   # the "user" directive makes sense only if the master process runs with super-user privileges.
                                   # if you're not root user,the default is current user.
   error_log: "logs/error.log"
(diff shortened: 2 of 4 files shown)


Yiyiyimu:
Hi @ysc972 thanks for your contribution! I think it would be better if you could
1. Disable the root access by default. Make it as a configuration field in `config-default.yaml` and add a comment on who should enable this
2. Also add an entry in maybe FAQ to inform people how to do with Chinese OS, since it people have larger chances to view this file when they got problems

apisix/cli/ngx_tpl.lua
```diff
@@ -19,6 +19,8 @@ return [=[
 # Configuration File - Nginx Server Configs
 # This is a read-only file, do not try to modify it.
 
+user root;
```
tokers, line 22: It's dangerous to use `root` to launch APISIX. Exposing this item to `config.yaml` so people can decide to use which user.

spacewander: This can be archived without modifying APISIX: https://github.com/apache/apisix/blob/master/docs/en/latest/customize-nginx-configuration.md

ysc972: I have add  a configuration field in `config-default.yaml`,and add `How to run APISIX on the Chinese OS` to FAQ.

docs/zh/latest/FAQ.md
```diff
@@ -501,3 +501,13 @@ $ acme.sh --renew --domain demo.domain
 ```
 
 详细步骤，可以参考博客 https://juejin.cn/post/6965778290619449351
+
+## 如何在kylin v10 国产操作系统运行apisix
```
spacewander, line 505: Is this problem confirmed by the official kylin v10 developer?
We don't want to be in trouble for spreading incorrect message about their OS.

spacewander (changes requested):

docs/zh/latest/FAQ.md
ysc972: Officials don't know about this problem, it's just my adaptation to the kylin v10 OS.

spacewander: Maybe we can submit a PR that supporting specifying the `user`, but don't mention a specific OS name as it's not confirmed by the official.

conf/config-default.yaml
```diff
@@ -130,6 +130,9 @@ apisix:
   disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
 
 nginx_config:                     # config for render the template to generate nginx.conf
+  #user: "root"                   # specifies the execution user of worker process.
+                                  # the "user" directive makes sense only if the master process runs with super-user privileges.
+                                  # if you is not root user,the default is current user.
```
tokers, line 135:
```suggestion
                                  # if you're not root user,the default is current user.
```

tokers (approved):

conf/config-default.yaml
```diff
@@ -130,6 +130,9 @@ apisix:
   disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
 
 nginx_config:                     # config for render the template to generate nginx.conf
+  #user: "root"                   # specifies the execution user of worker process.
```
Yiyiyimu, line 133:
```suggestion
  #user: "root"                   # specifies the execution user of the worker process.
```

conf/config-default.yaml
Yiyiyimu, line 135:
```suggestion
                                  # if you is not root user, by default it would be the current user.
```

docs/en/latest/FAQ.md
```diff
@@ -502,3 +502,14 @@ $ acme.sh --renew --domain demo.domain
 ```
 
 Blog https://juejin.cn/post/6965778290619449351 has detail setup.
+
+
+## How to run APISIX on the Chinese OS
+
+When the APISIX is run as the root user on the Chinese OS, the worker process of nginx is started as the nobody user by default. In the Chinese OS, luajit will have a ·`require` failure problem. You can start APISIX by adding the following configuration in `conf/conf-default.yaml` and executing `make run` as the root user.
```
Yiyiyimu, line 509:
```suggestion
When APISIX runs as the root user on Chinese OS, the worker process of Nginx is started as the nobody user by default. In Chinese OS, luajit will have a "require" failure problem. You can start APISIX by adding the following configuration to `conf/conf-default.yaml` and executing `make run` as the root user.
```

docs/zh/latest/FAQ.md
ysc972: Need I submit a new PR  or  change FAQ file in this PR ?

spacewander: I think we can remove the FAQ.

ysc972: But in the Chinese arm OS , if we do not change this config, the apisix cannot run . It will throw "require faild" error like issue #4157.  So should we write something in the FAQ to tell someone who uses apisix in Chinese arm OS how to resolve the problem.

spacewander: We need an official confirm for the FAQ.

ysc972: I have removed this FAQ and committed it.

docs/zh/latest/FAQ.md
```diff
@@ -501,3 +501,4 @@ $ acme.sh --renew --domain demo.domain
 ```
 
 详细步骤，可以参考博客 https://juejin.cn/post/6965778290619449351
+
```
spacewander, line 504: Please remove the blank line which causes the failure: https://github.com/apache/apisix/pull/4362/checks?check_run_id=2751014883

ysc972: I am sorry.I want to know if there is a way to let me know my mistakes like this blank in my native environment before I commit.Thanks

Yiyiyimu: @ysc972 we haven't provided a one-line method, but you could follow the actions step (for this is using npm to install `markdownlint-cli`) to test locally

ysc972: @Yiyiyimu Thank you . I wii try it .

spacewander (approved):

PR #4368, opened 2021-06-03
feat: upgrade protobuf to 0.3.2

yuhongyu879:
protobufjs use custom wrapper for google.protobuf.any and other types, if we want decode message like this, we can use pb.hook which introduced in 0.3.2, but after upgrade lua-protobuf to that version, the [grpc-transcode plugin](https://github.com/apache/apisix/blob/62ceab5710859d39828c6a88e3602ab6df1a589f/apisix/plugins/grpc-transcode/proto.lua#L48) won't work because of this [mr](https://github.com/starwing/lua-protobuf/commit/89faac205585360d800472c62752d5b532d7ce9e), so as a quick fix, we can pass a filename to load function, or if needed we can write a function does the same as the lib function do but return the resolve message info additionally.

thank you.


apisix/plugins/grpc-transcode/proto.lua
```diff
@@ -43,7 +43,7 @@ local function create_proto_obj(proto_id)
     end
 
     local _p  = protoc.new()
-    local res = _p:load(content)
+    local res = _p:load(content, "filename for loaded") -- https://github.com/starwing/lua-protobuf/commit/89faac205585360d800472c62752d5b532d7ce9e
```
Firstsawyou, line 46:
```suggestion
    -- https://github.com/starwing/lua-protobuf/commit/89faac205585360d800472c62752d5b532d7ce9e
    local res = _p:load(content, "filename for loaded") 
```

spacewander, line 46: We need to explain it clearly in the comment instead of just a commit link.

spacewander (changes requested):

apisix/plugins/grpc-transcode/proto.lua
yuhongyu879: how about the explanation now?

apisix/plugins/grpc-transcode/proto.lua
```diff
@@ -43,7 +43,11 @@ local function create_proto_obj(proto_id)
     end
 
     local _p  = protoc.new()
-    local res = _p:load(content)
+    -- the loaded proto won't appears in _p.loaded without a file name after lua-protobuf=0.3.2, which means _p.loaded 
+    -- after _p:load(content) is always empty, so we can pass a default to keep the codes below unchanged, or we can create
```
spacewander, line 47:
```suggestion
    -- after _p:load(content) is always empty, so we can pass a fake file name to keep the code below unchanged, or we can create 
```

spacewander:
Please fix the lint

apisix/plugins/grpc-transcode/proto.lua
yuhongyu879: thx

yuhongyu879: thx

yuhongyu879: oh should i use ur suggestion?

spacewander (approved):

Firstsawyou (approved):

yuhongyu879:
@spacewander Whom could i ask for approving? It says that 'Only reviews by reviewers with write access count toward mergeability' on @Firstsawyou 's approving

tokers (approved):

Issue #4460, opened 2021-06-22
bug:  The request-id plugin may have a bug


zuiyangqingzhou:
### Issue description
When I enable the global request-id plugin and the request-id plugin on the route at the same time, there will be a BUG

Bug report without environment information will be ignored or closed.

* apisix version (cmd: `apisix version`):  2.6
* OS (cmd: `uname -a`):  Darwin Kernel Version 20.5.0
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):  nginx version: openresty/1.19.3.2 
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API): etcd Version: 3.4.16
* apisix-dashboard version, if have:
* luarocks version, if the issue is about installation (cmd: `luarocks --version`):

### Minimal test code / Steps to reproduce the issue

Bug report without steps to reproduce will be ignored or closed.

1. I enable the request-id plugin on a specific route, the default header_name is X-Request-Id, and I get the following result, which looks normal
```
$ curl -i http://127.0.0.1:9080/hello -H 'Authorization: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJrZXkiOiJ0ZXN0LWp3dC1kZW1vLTEiLCJleHAiOjE2MjQzNTk3MjZ9.TF2o2KngjIHMYO2CqPurPr1f2UnZVry76kN2XOyKEIU'

HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Content-Length: 408
Connection: keep-alive
Date: Tue, 22 Jun 2021 02:25:37 GMT
Server: APISIX/2.6
X-Request-Id: f02ac2c2-5333-4c03-a4ba-9b81fafdcdab

# my upstream get headers from apisix
[...]

PR #4479, opened 2021-06-24
fix(request-id): we can use different ids with the same request

zuiyangqingzhou:
fix: [4460](https://github.com/apache/apisix/issues/4460)

### Pre-submission checklist:


tzssangglass:
@zuiyangqingzhou hi, need to add test cases

apisix/plugins/request-id.lua
```diff
@@ -49,7 +49,7 @@ function _M.rewrite(conf, ctx)
     end
 
     if conf.include_in_response then
-        ctx.x_request_id = uuid_val
+        ctx[conf.header_name] = uuid_val
```
spacewander, line 52: I think we need to add a prefix like `request-id` as `conf.header_name` may override other value by accident.

tokers: Agree +1

zuiyangqingzhou: I don’t really understand the intention of adding a prefix, because header_name is originally defined by the user. He should know what the definition of header_name looks like.

spacewander: But he doesn't know all the fields in the ctx and the field will be added to ctx in the future.

zuiyangqingzhou: Do you mean like this？
```
if conf.include_in_response then
        ctx["request-id" .. conf.header_name] = uuid_val
end
```

spacewander: `ctx["request-id-" .. conf.header_name] = uuid_val` would be better.

zuiyangqingzhou: All right！

In addition, I found that unit testing is not easy to do. Is there any related documentation? Thank you

spacewander: You can configure global rule and route like this one, with `/echo` uri: https://github.com/apache/apisix/blob/b76f20eddfdbd0bd546afbc515c5948a29e86600/t/node/global-rule.t#L312
https://github.com/apache/apisix/blob/b76f20eddfdbd0bd546afbc515c5948a29e86600/t/lib/server.lua#L394

And then hit it and check the header sent to upstream via :
https://github.com/apache/apisix/blob/b76f20eddfdbd0bd546afbc515c5948a29e86600/t/plugin/zipkin.t#L458-L462

t/plugin/request-id.t
```diff
@@ -387,3 +387,86 @@ GET /t
...
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/opentracing"
+            local res, err = httpc:request_uri(uri,
+                {
+                    method = "GET",
+                    headers = {
+                        ["Content-Type"] = "application/json",
+                    }
+                })
+
+            if res.headers["X-Request-Id"] and res.headers["Custom-Header-Name"] and res.headers["X-Request-Id"] ~= res.headers["Custom-Header-Name"] then
```
tokers, line 459: We should not verify these two values when they are present, actually they must present.

t/plugin/request-id.t
```diff
@@ -387,3 +387,86 @@ GET /t
...
+                ngx.say("X-Request-Id and Custom-Header-Name is different")
```
tokers, line 460:
```suggestion
                ngx.say("X-Request-Id and Custom-Header-Name are different")
```

t/plugin/request-id.t
zuiyangqingzhou: Okay, got it

tokers (approved):

t/plugin/request-id.t
```diff
@@ -387,3 +387,86 @@ GET /t
...
+            end
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
```
spacewander, line 441: Please fix the linter: https://github.com/apache/apisix/pull/4479/checks?check_run_id=2929454713

tzssangglass: there should be 3 blank lines between test cases, or you can use the reindex to do. the CI error is about this.

zuiyangqingzhou:
What cause the error?  https://github.com/apache/apisix/pull/4479/checks?check_run_id=2948088762

spacewander:
Don't be afraid, known issue: https://github.com/apache/apisix/pull/4503

spacewander (approved):

PR #4486, opened 2021-06-25
feat(stream): support stream subsystem discovery

adugeek:

     int function init.lua:stream_init_worker() not call discovery.init_worker(), 
     so, we can't use discovery in stream subsystem.
     by this pr, we can slove it 
     of course, the current and future discovery components need to ensure that they can work in the stream subsystem 
  

### Pre-submission checklist:

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -823,6 +823,11 @@ function _M.stream_init_worker()
         core.config.init_worker()
     end
 
+    local discovery = require("apisix.discovery.init").discovery
+    if discovery and discovery.init_worker then
+        discovery.init_worker()
+    end
+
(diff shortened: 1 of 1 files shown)


tokers:
@adugeek Please change the git commit log so that it can be detected as semantic.

See https://github.com/zeke/semantic-pull-requests for the details.

adugeek:
> @adugeek Please change the git commit log so that it can be detected as semantic.
> 
> See https://github.com/zeke/semantic-pull-requests for the details.

@tokers ,TKS,  I had change the log to "feat(stream): support stream subsystem discovery" ,  suitable ?

Issue #4421, opened 2021-06-13
request help: key auth plugin didn't support get key from querystring


forgaoqiang:
### Issue description
I think it's a basic demand that the apikey should be check from HTTP header or HTTP Querystring, could you guys add that feature, It's very convenient to support both HTTP header and query string 

### Environment

Request help without environment information will be ignored or closed.

* apisix version (cmd: `apisix version`):
* OS (cmd: `uname -a`):
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
* apisix-dashboard version, if have:
* luarocks version, if the issue is about installation (cmd: `luarocks --version`):

tokers:
@forgaoqiang PR's welcome to support it!

forgaoqiang:
thanks for the feature

PR #4490, opened 2021-06-26
feat(key-auth): supporting key-auth plugin to get key from query string

wisdom-yzh:

fixes issue #4421 

Adding a new `query` property to key-auth plugin, which can get key value from http query string

### Pre-submission checklist:


apisix/plugins/key-auth.lua
```diff
@@ -32,6 +32,10 @@ local schema = {
             type = "string",
             default = "apikey",
         },
+        query = {
```
tokers, line 35: Should add some constraints to indicate that `query` and `header` are exclusive.

apisix/plugins/key-auth.lua
```diff
@@ -84,6 +88,12 @@ end
 
 function _M.rewrite(conf, ctx)
     local key = core.request.header(ctx, conf.header)
+
+    if not key then
```
tokers, line 92: We should not introduce using order implicitly.

tokers:
@wisdom-yzh From the changes I guess you tried to introduce an using order between the `header` and `query`, it will be somewhat obscure, we may treat them equally but set them as exclusive in jsonschema (e.g. by introducing the `oneOf` primitive).

wisdom-yzh:
@tokers Accord to description of #4421, I guess @forgaoqiang wants to get key from both query string and headers, not exclusive. Maybe I can define one field instead of two fields, and check key from both query and headers?

spacewander:
I think "header and query string" is OK, as jwt-auth plugin does it: https://github.com/apache/apisix/blob/07bfe5b9c03ad361cb71541eb02fd661f1323213/apisix/plugins/jwt-auth.lua#L154

t/plugin/key-auth.t
```diff
@@ -300,7 +300,45 @@ passed
 
 
 
-=== TEST 11: valid consumer
+=== TEST 11: customize querystring
```
spacewander, line 303: Please don't add the test in the middle of the other test. And we need to verify it via `GET /hello?xxx` in its own version of "valid consumer"

wisdom-yzh:
@spacewander hi~I have changed strategy to "header and query string", and then rename field from "header" to "key_name“.

spacewander:
@wisdom-yzh 
Err... We can't change the existing fields which will break the user's configuration. Your original solution is good enough, please bring it back.

wisdom-yzh:
> @wisdom-yzh
> Err... We can't change the existing fields which will break the user's configuration. Your original solution is good enough, please bring it back.

ok, i'll restore the origin commit and fix the test case issue soon

t/plugin/key-auth.t
```diff
@@ -309,3 +309,51 @@ Authorization: auth-one
...
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST13: valid consumer
```
spacewander, line 353: need a space after TEST

spacewander (approved):

t/plugin/key-auth.t
wisdom-yzh: ok resolved

tokers (approved):

forgaoqiang:
Great, this feature is what I needed, thanks

Issue #4428, opened 2021-06-16
ip-restriction plugin support DIY message


belongwqz:
# Improve Docs

## Please describe which part of docs should be improved or typo fixed

ip-restriction plugin support DIY message

## Describe the solution you'd like

current message is a hard code: Your IP address is not allowed.
if we can DIY, it should be better.

tokers:
@belongwqz PR's welcome. BTW, it's not a doc-related issue.

PR #4493, opened 2021-06-27
feat(ip-restriction): support user-defined configuration message

Applenice:
fix: https://github.com/apache/apisix/issues/4428

Added custom message to ip-restriction plugin and no problem in my local tests. It looks like I need to learn how to write test cases, right?

### Pre-submission checklist:


apisix/plugins/ip-restriction.lua
```diff
@@ -36,6 +36,10 @@ local schema = {
                     items = {anyOf = core.schema.ip_def},
                     minItems = 1
                 },
+                message = {
```
spacewander, line 39: Better to extract this field to avoid repeating it

docs/en/latest/plugins/ip-restriction.md
```diff
@@ -70,6 +70,20 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
...
+If you want to use a custom message, you can configure it in the plugin section.
+
+```shell
+"plugins": {
+    "ip-restriction": {
+        "whitelist": [
+            "127.0.0.1",
+            "113.74.26.106/24"
+        ],
+        "message": "Do you want to do something bad?"
```
spacewander, line 82: Need to update `Attributes`

docs/en/latest/plugins/ip-restriction.md
```diff
@@ -70,6 +70,20 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
 }'
 ```
 
+If you want to use a custom message, you can configure it in the plugin section.
+
+```shell
```
spacewander, line 75: Should be json

docs/en/latest/plugins/ip-restriction.md
```diff
@@ -70,6 +70,20 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
 }'
 ```
 
+If you want to use a custom message, you can configure it in the plugin section.
```
spacewander, line 73: Use a custom message for what?

apisix/plugins/ip-restriction.lua
tokers: And add some limitations for the max length?

docs/en/latest/plugins/ip-restriction.md
Applenice: Thanks for the review, I saw https://github.com/apache/apisix/issues/4428, which mentioned ip-restriction plugin support DIY message, was marked `good first issue`, so wanted to try to implement it.  I'll make the changes mentioned above later.

apisix/plugins/ip-restriction.lua
Applenice: I'll figure out how to modify it, thanks

docs/en/latest/plugins/ip-restriction.md
Applenice: Sorry, I didn't understand😥, what does this attribute refer to?

docs/en/latest/plugins/ip-restriction.md
spacewander: Err. Actually I mean we should write it more clear in the documentation, as the user won't have the same context as us.

docs/en/latest/plugins/ip-restriction.md
spacewander: https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/ip-restriction.md#attributes
Need to update the table.

apisix/plugins/ip-restriction.lua
spacewander: See https://github.com/apache/apisix/blob/07bfe5b9c03ad361cb71541eb02fd661f1323213/apisix/schema_def.lua#L33

docs/en/latest/plugins/ip-restriction.md
Applenice: Thanks, now I understand😊,previously ignored this place

docs/en/latest/plugins/ip-restriction.md
Applenice: I understand, I will write more clearly later

apisix/plugins/ip-restriction.lua
Applenice: ok👌

apisix/plugins/ip-restriction.lua
```diff
@@ -27,40 +27,29 @@ local lrucache  = core.lrucache.new({
...
-                    default = "Your IP address is not allowed"
-                },
-            },
-            required = {"whitelist"},
-            additionalProperties = false,
+    properties = {
+        message = {
+            type = "string",
+            minLength = 1,
+            maxLength = 64,
```
spacewander, line 34: 64 looks too short to me. What about 4096?

docs/en/latest/plugins/ip-restriction.md
```diff
@@ -41,9 +41,10 @@ in CIDR notation like 10.10.10.0/24 can be used.
 | --------- | ------------- | ----------- | ------- | ----- | ---------------------------------------- |
 | whitelist | array[string] | optional    |         |       | List of IPs or CIDR ranges to whitelist. |
 | blacklist | array[string] | optional    |         |       | List of IPs or CIDR ranges to blacklist. |
+| message | string | optional    | Your IP address is not allowed. | [1, 64] | Message returned in case IP access is not allowed. |
 
-One of `whitelist` or `blacklist` must be specified, and they can not work
-together.
+One of `whitelist` or `blacklist` must be specified, and they can not work together.
+The message supports user-defined configuration.
```
spacewander, line 47: The message can be user-defined.

spacewander:
@Applenice 
Need to add test in https://github.com/apache/apisix/blob/master/t/plugin/ip-restriction.t

apisix/plugins/ip-restriction.lua
Applenice: Would 4096 be too long?What about 1024?😃

docs/en/latest/plugins/ip-restriction.md
Applenice: ok,later modify

Applenice:
> @Applenice
> Need to add test in https://github.com/apache/apisix/blob/master/t/plugin/ip-restriction.t

Okay😊

Applenice:
I changed the maxLength value from 64 to 1024, updated the documentation and added test cases, but there is a problem.
For test cases that are out of maxLength range, it is possible to pass the test using a fixed 1025 length string, but it should not be a graceful operation, how to call it with code generation? I searched some sources and didn't find an answer, so I didn't add a test case for when maxLength is exceeded.

spacewander:
You can use `string.rep`, like this one: https://github.com/apache/apisix/blob/d4c7fbbe831bc12c50ba63f00757f0eee14eae15/t/node/upstream-mtls.t#L100

tokers (approved):

tokers:
@Applenice Please take a look, the CI failed.

Applenice:
> You can use `string.rep`, like this one:
> 
> https://github.com/apache/apisix/blob/d4c7fbbe831bc12c50ba63f00757f0eee14eae15/t/node/upstream-mtls.t#L100

Thanks, I'll try it later

Applenice:
> @Applenice Please take a look, the CI failed.

Yes, I saw `Failed tests:  2, 5, 8, 14, 17, 72`, I observed it when testing locally with `prove -Itest-nginx/lib -r t/plugin/ip-restriction.t`, I'll check the reason later

spacewander (approved):

PR #4504, opened 2021-06-30
feat(cli): check if control port conflicts with the node port

arthur-zhang:
batch-requests plugin is working based on routing all the the request to 127.0.0.1:9080, if enable-control is on and port is the same with node_listen port, it will add configuration file like this：

```
  server {
      listen 127.0.0.1:9080;
      access_log off;
      location / {
          content_by_lua_block {
              local prometheus = require("apisix.plugins.prometheus")
              prometheus.export_metrics()
          }
      }
  }
```

this will cause all the 127.0.0.1:9080/* request route to this control api server, and all the expected  batch-requests  url return 404.

I add some force  check when batch-requests  is enabled.

### DIFF
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -24,6 +24,7 @@ local profile = require("apisix.core.profile")
 local template = require("resty.template")
 local argparse = require("argparse")
 local pl_path = require("pl.path")
+local pl_stringx = require("pl.stringx")
 local jsonschema = require("jsonschema")
 
 local stderr = io.stderr
@@ -539,6 +540,29 @@ Please modify "admin_key" in conf/config.yaml .
             end
 
             sys_conf.control_server_addr = ip .. ":" .. port
+        end
+    end
+
+    -- check control port conflict when batch-requests is enabled
+    if enabled_plugins["batch-requests"] and yaml_conf.apisix.enable_control then
+        local split = pl_stringx.split
+        local ip_port = split(sys_conf.control_server_addr, ":")
+        local port = tonumber(ip_port[#ip_port])
+        if port == nil then
+            util.die("port not number, should not happen\n")
+        end
+
+        if type(yaml_conf.apisix.node_listen) == "number" then
+            if port == yaml_conf.apisix.node_listen then
+                util.die("control port conflict node_listen port when batch-requests is on\n")
+            end
+        elseif type(yaml_conf.apisix.node_listen) == "table" then
+            for _, value in ipairs(yaml_conf.apisix.node_listen) do
+                if type(value) == "table" and type(value.port) == "number"
+                        and value.port == port then
+                    util.die("control port conflict node_listen port when batch-requests is on\n")
+                end
+            end
         end
     end
 
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -433,7 +433,7 @@ Please modify "admin_key" in conf/config.yaml .
     if type(yaml_conf.apisix.node_listen) == "number" then
 
         if yaml_conf.apisix.node_listen == control_port then
-            util.die("control port conflict with node_listen port\n")
+            util.die("control port conflicts with node_listen port\n")
         end
 
         local node_listen = {{port = yaml_conf.apisix.node_listen}}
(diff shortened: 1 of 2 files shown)


apisix/cli/ops.lua
```diff
@@ -542,6 +543,29 @@ Please modify "admin_key" in conf/config.yaml .
         end
     end
 
+    -- check control port conflict when batch-requests is enabled
+    if enabled_plugins["batch-requests"] and yaml_conf.apisix.enable_control then
```
spacewander, line 547: Although this only happens when batch-requests is enabled now, I think we can always check if the port conflicts even batch-requests is not enabled.

apisix/cli/ops.lua
```diff
@@ -542,6 +543,29 @@ Please modify "admin_key" in conf/config.yaml .
...
+        local split = pl_stringx.split
+        local ip_port = split(sys_conf.control_server_addr, ":")
+        local port = tonumber(ip_port[#ip_port])
+        if port == nil then
+            util.die("port not number, should not happen\n")
+        end
+
+        if type(yaml_conf.apisix.node_listen) == "number" then
```
spacewander, line 555: Maybe we can move the control API stuff above: https://github.com/apache/apisix/blob/d0bc7238c912259e325f45ad6caea039d24e4292/apisix/cli/ops.lua#L410

And use a port denylist to check it when iterating the node_listen? So that we don't need to iterate it twice.

spacewander:
Could you add a test in https://github.com/apache/apisix/blob/master/t/cli/test_control.sh?

apisix/cli/ops.lua
arthur-zhang: I think so. I've commit the code.

arthur-zhang:
> Could you add a test in https://github.com/apache/apisix/blob/master/t/cli/test_control.sh?

I read all the shell code, but till  have no idea how to make a test case in this scenario

spacewander:
You can set node_listen to 9090: https://github.com/apache/apisix/blob/047f5e7d503ea62eba28bae7386efb3e6d5b8268/t/cli/test_main.sh#L85
And then check the output of `make init`.

apisix/cli/ops.lua
```diff
@@ -407,16 +407,53 @@ Please modify "admin_key" in conf/config.yaml .
         util.die("missing apisix.proxy_cache for plugin proxy-cache\n")
     end
 
+    local control_port = 9090
```
spacewander, line 410:
```suggestion
    local control_port
```

arthur-zhang: > You can set node_listen to 9090:
> 
> https://github.com/apache/apisix/blob/047f5e7d503ea62eba28bae7386efb3e6d5b8268/t/cli/test_main.sh#L85
> 
> 
> And then check the output of `make init`.

Many thanks, test case added.

apisix/cli/ops.lua
```diff
@@ -407,16 +407,53 @@ Please modify "admin_key" in conf/config.yaml .
...
+            control_server_addr = ip .. ":" .. port
+            control_port = port
+        end
+    end
+
     -- support multiple ports listen, compatible with the original style
     if type(yaml_conf.apisix.node_listen) == "number" then
+
+        if yaml_conf.apisix.node_listen == control_port then
+            util.die("control port conflict with node_listen port\n")
```
spacewander, line 436:
```suggestion
            util.die("control port conflicts with node_listen port\n")
```

apisix/cli/ops.lua
```diff
@@ -407,16 +407,53 @@ Please modify "admin_key" in conf/config.yaml .
...
+
         local node_listen = {{port = yaml_conf.apisix.node_listen}}
         yaml_conf.apisix.node_listen = node_listen
     elseif type(yaml_conf.apisix.node_listen) == "table" then
         local node_listen = {}
         for index, value in ipairs(yaml_conf.apisix.node_listen) do
             if type(value) == "number" then
+
+                if value == control_port then
+                    util.die("control port conflict with node_listen port\n")
```
spacewander, line 447: ditto

arthur-zhang: done

tokers (approved):

spacewander (approved):

spacewander:
@arthur-zhang 
Would you also submit a PR to check port conflict with prometheus one? https://github.com/apache/apisix/blob/cd8afe6ec4e043a188718c41fd5937d2cd88c59d/conf/config-default.yaml#L313

arthur-zhang:
> @arthur-zhang
> Would you also submit a PR to check port conflict with prometheus one?
> 
> https://github.com/apache/apisix/blob/cd8afe6ec4e043a188718c41fd5937d2cd88c59d/conf/config-default.yaml#L313

OK，I will submit later

PR #4515, opened 2021-07-01
feat(stream): add limit-conn

spacewander:

### Pre-submission checklist:


docs/zh/latest/plugins/limit-conn.md
```diff
@@ -35,6 +35,9 @@ title: limit-conn
 
 **注：key 是可以被用户自定义的，只需要修改插件的一行代码即可完成。并没有在插件中放开是处于安全的考虑。**
 
+: when used in the stream proxy, only `remote_addr` and `server_addr` can be used as `key`. And `rejected_code` is meaningless.
```
tokers, line 38: What is the meaning of the leading colon symbol?

tzssangglass (approved):

tokers (approved):

t/stream-plugin/limit-conn.t
```diff
@@ -0,0 +1,191 @@
...
+                ngx.HTTP_PUT,
+                [[{
+                    "nodes": {
+                        "127.0.0.1:1995": 1
+                    },
+                    "type": "roundrobin"
+                }]]
+                )
+
+            if code >= 300 then
```
tao12345666333, line 100: Should it only be restricted to not equal to 200 or 201? Although currently its value will only be 200, 201 or 500

tao12345666333 (approved):
Approved.
Leave a comment, but it doesn’t affect.

t/stream-plugin/limit-conn.t
spacewander: `code >= 300` acts as `err != nil`, since it is so widely used, we can keep using it.

PR #4546, opened 2021-07-06
docs: update CN and EN getting-started documents

yzeng25:

This change fixes the problems on 'getting-started' page both in Chinese and in English:

- Fix issue https://github.com/apache/apisix/issues/4554
- Added a summary section on the top of this page.
- Listed out 3-step procedure in summary section.
- To avoid calling one thing with multiple names, we decided to replace "Upstream" with "Upstream service", and replace "上游" and "后端服务" with "上游服务" (fixed after review comments), and call 'Route' "路由".
- Removed '$' in front of each code block, because it creates problem when users try to copy the codes or commands down, '$' is not a part of the command, but it gets copied down, it is annoying.
- Fixed some mix uses of punctuations，replaced (,./;'') with （，。、；’）.
- Fixed sections' titles. Titles usually should follow 'verb + noun(s)' pattern.
- Removed unnecessary sentences.
- Replaced "注意" with a pattern supported by docusaurus, to make these "注意" more notable. Reference: https://docusaurus.io/docs/next/markdown-features/admonitions

### Pre-submission checklist:

### DIFF
--- a/docs/zh/latest/getting-started.md
+++ b/docs/zh/latest/getting-started.md
@@ -122,9 +122,9 @@ curl "http://127.0.0.1:9080/apisix/admin/services/" -H 'X-API-KEY: edd1c9f034335
 
 Apache APISIX 提供了强大的 [Admin API](./admin-api.md) 和 [Dashboard](https://github.com/apache/apisix-dashboard) 可供我们使用。在本文中，我们使用 Admin API 来做演示。
 
-我们可以创建一个 [Route](./architecture-design/route.md) 并与上游（通常也被称为[Upstream](./architecture-design/upstream.md)或后端服务）绑定，当一个 `请求（Request）` 到达 Apache APISIX 时，Apache APISIX 就会明白这个请求应该转发到哪个上游中。
+我们可以创建一个 [Route](./architecture-design/route.md) 并与上游服务（通常也被称为[Upstream](./architecture-design/upstream.md)或后端服务）绑定，当一个 `请求（Request）` 到达 Apache APISIX 时，Apache APISIX 就会明白这个请求应该转发到哪个上游服务中。
 
-因为我们为 Route 对象配置了匹配规则，所以 Apache APISIX 可以将请求转发到对应的上游。以下代码是一个 Route 配置示例：
+因为我们为 Route 对象配置了匹配规则，所以 Apache APISIX 可以将请求转发到对应的上游服务。以下代码是一个 Route 配置示例：
 
 ```json
 {
@@ -140,13 +140,13 @@ Apache APISIX 提供了强大的 [Admin API](./admin-api.md) 和 [Dashboard](htt
 }
 ```
 
-这条路由配置意味着，当它们满足下述的 **所有** 规则时，所有匹配的入站请求都将被转发到 `httpbin.org:80` 这个上游：
+这条路由配置意味着，当它们满足下述的 **所有** 规则时，所有匹配的入站请求都将被转发到 `httpbin.org:80` 这个上游服务：
 
 - 请求的 HTTP 方法为 `GET`。
 - 请求头包含 `host` 字段，且它的值为 `example.com`。
 - 请求路径匹配 `/services/users/*`，`*` 意味着任意的子路径，例如 `/services/users/getAll?limit=10`。
 
-当这条路由创建后，我们可以使用 Apache APISIX 对外暴露的地址去访问上游：
+当这条路由创建后，我们可以使用 Apache APISIX 对外暴露的地址去访问上游服务：
 
 ```bash
 curl -i -X GET "http://{APISIX_BASE_URL}/services/users/getAll?limit=10" -H "Host: example.com"
@@ -154,9 +154,9 @@ curl -i -X GET "http://{APISIX_BASE_URL}/services/users/getAll?limit=10" -H "Hos
 
 这将会被 Apache APISIX 转发到 `http://httpbin.org:80/services/users/getAll?limit=10`。
 
-### 创建上游（Upstream）
+### 创建上游服务（Upstream）
 
-读完上一节，我们知道必须为 `Route` 设置 `Upstream`。只需执行下面的命令即可创建一个上游：
+读完上一节，我们知道必须为 `Route` 设置 `Upstream`。只需执行下面的命令即可创建一个上游服务：
 
 ```bash
 curl "http://127.0.0.1:9080/apisix/admin/upstreams/1" -H "X-API-KEY: edd1c9f034335f136f87ad84b625c8f1" -X PUT -d '
@@ -168,15 +168,15 @@ curl "http://127.0.0.1:9080/apisix/admin/upstreams/1" -H "X-API-KEY: edd1c9f0343
 }'
 ```
 
-我们使用 `roundrobin` 作为负载均衡机制，并将 `httpbin.org:80` 设置为我们的上游目标（后端服务），其 ID 为 `1`。更多字段信息，请参阅 [Admin API](./admin-api.md)。
+我们使用 `roundrobin` 作为负载均衡机制，并将 `httpbin.org:80` 设置为我们的上游服务，其 ID 为 `1`。更多字段信息，请参阅 [Admin API](./admin-api.md)。
 
 :::note 注意
-创建上游实际上并不是必需的，因为我们可以使用 [插件](./architecture-design/plugin.md) 拦截请求，然后直接响应。但在本指南中，我们假设需要设置至少一个上游。
+创建上游服务实际上并不是必需的，因为我们可以使用 [插件](./architecture-design/plugin.md) 拦截请求，然后直接响应。但在本指南中，我们假设需要设置至少一个上游服务。
 :::
 
-### 绑定路由与上游
+### 绑定路由与上游服务
 
-我们刚刚创建了一个上游(引用我们的后端服务)，现在让我们为它绑定一个路由！
+我们刚刚创建了一个上游服务，现在让我们为它绑定一个路由！
 
 ```bash
 curl "http://127.0.0.1:9080/apisix/admin/routes/1" -H "X-API-KEY: edd1c9f034335f136f87ad84b625c8f1" -X PUT -d '
@@ -189,13 +189,13 @@ curl "http://127.0.0.1:9080/apisix/admin/routes/1" -H "X-API-KEY: edd1c9f034335f
 
 ## 第三步：验证
 
-我们已创建了路由与上游，并将它们进行了绑定。现在让我们访问 Apache APISIX 来测试这条路由：
+我们已创建了路由与上游服务，并将它们进行了绑定。现在让我们访问 Apache APISIX 来测试这条路由：
 
 ```bash
 curl -i -X GET "http://127.0.0.1:9080/get?foo1=bar1&foo2=bar2" -H "Host: httpbin.org"
 ```
 
-它从我们的上游（实际是 `httpbin.org`）返回数据，并且结果符合预期。
+它从我们的上游服务（实际是 `httpbin.org`）返回数据，并且结果符合预期。
 
 ## 进阶操作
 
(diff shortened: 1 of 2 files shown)


docs/zh/latest/getting-started.md
```diff
@@ -243,27 +256,31 @@ $ curl "http://127.0.0.1:9080/apisix/admin/routes/1" -H "X-API-KEY: edd1c9f03433
...
 
     下面是终止正在侦听特定端口（基于 unix 的系统）的进程的命令。
 
     ```bash
-    $ sudo fuser -k 9443/tcp
+    sudo fuser -k 9443/tcp
     ```
 
-- 如果 Docker 容器持续不断地重启/失败，请登录容器并观察日志以诊断问题。
+- 如果 Docker 容器持续不断地重启失败，请登录容器并观察日志以诊断问题。
```
tao12345666333, line 282:
```suggestion
- 如果 Docker 容器持续不断地重启或失败，请登录容器并观察日志以诊断问题。
```

yzeng25: fixed

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,29 @@ title: 快速入门指南
 #
 -->
 
-本指南旨在让大家入门 Apache APISIX，我们将配置一个对外提供公共 API 的服务，并由 API key 进行访问保护。
+## 概述
 
-另外，我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+本文是 Apache APISIX 的快速入门指南。快速入门分为三个步骤：首先通过[Docker](https://www.docker.com/) 和 [Docker Compose](https://docs.docker.com/compose/) 安装 Apache APISIX，然后创建路由并绑定后端服务，最后使用命令行语句`curl`验证绑定之后返回的结果是否符合预期。
```
juzhiyuan, line 26: How about using a numbered list? e.g

1. xx
2. xx
3. xx

juzhiyuan: added spaces around `curl`

```suggestion
本文是 Apache APISIX 的快速入门指南。快速入门分为三个步骤：首先通过[Docker](https://www.docker.com/) 和 [Docker Compose](https://docs.docker.com/compose/) 安装 Apache APISIX，然后创建路由并绑定后端服务，最后使用命令行语句 `curl` 验证绑定之后返回的结果是否符合预期。
```

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,29 @@ title: 快速入门指南
...
 
-**Request**
+除此之外，本文也提供了 Apache APISIX 的一些进阶操作技巧，包括：添加身份验证、为路由添加前缀、使用 APISIX Dashboard以及常见问题排查。
```
juzhiyuan, line 28:
```suggestion
除此之外，本文也提供了 Apache APISIX 的一些进阶操作技巧，包括：添加身份验证、为路由添加前缀、使用 APISIX Dashboard 以及常见问题排查。
```

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,29 @@ title: 快速入门指南
...
+
+我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+
+**请求内容**
+
+请求 URL 由以下这些参数构成：
+
+- Protocol：即网络传输协议，这里使用的是最常见的HTTP协议
+- Port：即端口，这里使用的`80`端口。
```
juzhiyuan, line 37: Please add spaces around `number` and `URL`

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,29 @@ title: 快速入门指南
...
+- Host：即宿主机，这里的主机是`httpbin.org`
+- URI/Path：`/get`
```
juzhiyuan, line 39: path? uri?

docs/zh/latest/getting-started.md
yzeng25: > How about using a numbered list? e.g
> 
> 1. xx
> 2. xx
> 3. xx

Good point, will do

docs/zh/latest/getting-started.md
yzeng25: Is it a path or a URI? It is written this way before the change. Please clarify.

yzeng25: Fixed, Path.

juzhiyuan (commented):
Chinese LGTM, and do we need to update the English version in the same PR?

cc @spacewander

yzeng25:
> Chinese LGTM, and do we need to update the English version in the same PR?
> 
> cc @spacewander

Yeah sure.

spacewander:
Yes, change EN doc at the same time can avoid content mismatch.

yzeng25:
> Yes, change EN doc at the same time can avoid content mismatch.

OK, will deliver.

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,33 @@ title: 快速入门指南
...
 -->
 
-本指南旨在让大家入门 Apache APISIX，我们将配置一个对外提供公共 API 的服务，并由 API key 进行访问保护。
+## 概述
 
-另外，我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+本文是 Apache APISIX 的快速入门指南。快速入门分为三个步骤：
 
-**Request**
+1. 通过[Docker](https://www.docker.com/) 和 [Docker Compose](https://docs.docker.com/compose/) 安装 Apache APISIX。
```
Yiyiyimu, line 28: Maybe `docker compose` is enough here, since it seems impossible to only install docker-compose but not docker

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,33 @@ title: 快速入门指南
...
+我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+
+**请求内容**
+
+请求 URL 由以下这些参数构成：
+
+- Protocol：即网络传输协议，这里使用的是最常见的HTTP协议。
+- Port：即端口，这里使用的 `80` 端口。
+- Host：即宿主机，这里的主机是 `httpbin.org`。
+- Path：`/get`。
```
Yiyiyimu, line 43:
```suggestion
- Path：即路径，这里是`/get`。
```

docs/zh/latest/getting-started.md
```diff
@@ -50,39 +66,39 @@ $ curl --location --request GET "http://httpbin.org/get?foo1=bar1&foo2=bar2"
...
-
-- Protocol: HTTP
-- Port: 80
-- Host: `httpbin.org`
-- URI/Path: `/get`
-- Query Parameters: foo1, foo2
+## 前提条件
 
-## 前提
+- 已安装[Docker](https://www.docker.com/) 和 [Docker Compose 组件](https://docs.docker.com/compose/)。
```
Yiyiyimu, line 71: ditto

docs/zh/latest/getting-started.md
```diff
@@ -50,39 +66,39 @@ $ curl --location --request GET "http://httpbin.org/get?foo1=bar1&foo2=bar2"
...
 
-> 如果您已经安装了 Apache APISIX，请直接阅读 [第二步](getting-started.md#第二步:-创建一个-Route)
+- 熟悉`curl`命令。本文使用 [curl](https://curl.se/docs/manpage.html) 命令行进行 API 测试。您也可以使用其他工具例如 [Postman](https://www.postman.com/)等，进行测试。
```
Yiyiyimu, line 73: seems we don't need to be familiar with `curl`, run the command is enough

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,33 @@ title: 快速入门指南
...
+
+我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+
+**请求内容**
+
+请求 URL 由以下这些参数构成：
+
+- Protocol：即网络传输协议，这里使用的是最常见的HTTP协议。
+- Port：即端口，这里使用的 `80` 端口。
+- Host：即宿主机，这里的主机是 `httpbin.org`。
```
Yiyiyimu, line 42:
```suggestion
- Host：即宿主机，这里是 `httpbin.org`。
```

Yiyiyimu (commented):
I think it's not appropriate to call it "后端服务" since there is no "前端". Maybe call it "上游" is more accurate

docs/zh/latest/getting-started.md
```diff
@@ -21,17 +21,33 @@ title: 快速入门指南
...
 
-本指南旨在让大家入门 Apache APISIX，我们将配置一个对外提供公共 API 的服务，并由 API key 进行访问保护。
+## 概述
 
-另外，我们将以下面的 `echo` 端点为例，它将返回我们传递的参数。
+本文是 Apache APISIX 的快速入门指南。快速入门分为三个步骤：
 
-**Request**
+1. 通过[Docker](https://www.docker.com/) 和 [Docker Compose](https://docs.docker.com/compose/) 安装 Apache APISIX。
+1. 创建路由并绑定后端服务。
```
Yiyiyimu, line 29:
```suggestion
2. 创建路由并绑定后端服务。
```

yzeng25:
> I think it's not appropriate to call it "后端服务" since there is no "前端". Maybe call it "上游" is more accurate

Ok, I will fix them in both languages.

docs/zh/latest/getting-started.md
yzeng25: fixed

docs/zh/latest/getting-started.md
yzeng25: fixed

docs/zh/latest/getting-started.md
yzeng25: fixed

docs/zh/latest/getting-started.md
yzeng25: fixed

docs/zh/latest/getting-started.md
yzeng25: fixed

docs/zh/latest/getting-started.md
yzeng25: Doucusarus will automatically change this '111....1'to a '123...n' order. I prefer leave it as it is. 

In case we have to add some more steps to it, then I would need to only add a '1. step description' to the document, rather than change all orders one by one.

yzeng25:
@Yiyiyimu @spacewander @juzhiyuan @tao12345666333  guys, please kindly give me some review opinions/comments on English 'Getting Started' document when you are available, thanks!

tokers:
> I think it's not appropriate to call it "后端服务" since there is no "前端". Maybe call it "上游" is more accurate

If we decide to use the term "上游"，then we may use it everywhere of our documentations.

By the way, a glossary might be required.

yzeng25:
> > I think it's not appropriate to call it "后端服务" since there is no "前端". Maybe call it "上游" is more accurate
> 
> If we decide to use the term "上游"，then we may use it everywhere of our documentations.
> 
> By the way, a glossary might be required.

Yes, this PR uses "Upstream" **everywhere in 'Getting-Started' document**, except 2-3ish places where it needs to explain what is an "Upstream", or give a synonym of "Upstream". 

The current situation is that "Upstream", "后端服务" and other words are referencing to the same thing with different names (Maybe they do have some tiny differences after all). Thus, it creates confusion for me to read, identify and understand this document.

This then brings up more abstracted problems: 

1.  Similarly, here is what I would ask: is there an "upstream" since there  is no "downstream"? which of the following is more broadly accepted, "后端服务", or "上游"?
2. Where is the glossary? Yes, we definitely need a glossary to explain and define frequently used words and phrases. But it is not my top priority task yet. 
2. How many other places are referencing the same thing with different names? I don't know, it is just a matter of time to find out and fix them.
3. Why this change is only made in 'Getting-Started' document'?  Well, it is the very first document page or the 'entry' of all Apache APISIX documents. I want this page to be somewhat helpful and impressing.
4. Are we missing other sorts of documents? Along with Q1 and Q3, my answer is yes. We definitely need a product overview, product architecture, glossary, etc, to form a series of 'Product Introduction' documents. And we do need other series of documents as well. The reason why I am not doing this right now is that I am new to this team, and I want to find out how can I make good use of the existing documents. 

Hope this clarifies your concerns.

juzhiyuan:
> Similarly, here is what I would ask: is there an "upstream" since there is no "downstream"? which of the following is more broadly accepted, "后端服务", or "上游"?

Hi, AFAIK, Upstream is used in Apache APISIX (as an Entity) and Kong (as an Upstream Object), it's more technical but not easy to understand, as a Getting Started guide, Backend Server/Target Node would be more clear and easy to know what it is. Maybe we could have a table including some description for internal words?

juzhiyuan:
what do you think? @KishaniKandasamy @iamayushdas

iamayushdas:
Table is a good option but what do you think , where will you be placing that table, IMO we could place them in "general"

Yiyiyimu:
> Hi, AFAIK, Upstream is used in Apache APISIX (as an Entity) and Kong (as an Upstream Object), it's more technical but not easy to understand

Hi I think the reason we call it "upstream" is mainly because [nginx is calling it this way](http://nginx.org/en/docs/http/ngx_http_upstream_module.html)

yzeng25:
> > Hi, AFAIK, Upstream is used in Apache APISIX (as an Entity) and Kong (as an Upstream Object), it's more technical but not easy to understand
> 
> 
> 
> Hi I think the reason we call it "upstream" is mainly because [nginx is calling it this way](http://nginx.org/en/docs/http/ngx_http_upstream_module.html)

Cool, good to know that. It seems like a good choice to follow their pattern.

KishaniKandasamy:
> what do you think? @KishaniKandasamy @iamayushdas


Yes It is a much better way  to understand internal words

yzeng25:
Hi all, after this morning's discussion, we decided to use the term "Upstream service" and "上游服务" in our documents.  PR is ready for review,.
@juzhiyuan @tokers @spacewander @Yiyiyimu 
cc @membphis @iamayushdas @KafilatAdeleke

Yiyiyimu (approved):

membphis (approved):

PR #4548, opened 2021-07-06
feat: allow plugin running in the balancer phase

dmsolr:
### Pre-submission checklist:

I found, now, APISIX does not invoke plugin's balancer function.
As far as I learned, we have added `common_phase("balancer")` at the end of the balancer function.
At actually, it really works.


tokers:
@dmsolr Please add some test cases to cover it.

PR #4559, opened 2021-07-07
feat: Request-ID plugin add snowflake algorithm

dickens7:
#4209 

### Pre-submission checklist:


spacewander:
Need to make the CI pass.

dickens7:
CI  help

Yiyiyimu:
> CI help

Rerun to see how it goes

Yiyiyimu:
Hi @dickens7 do you mind also add documentations of this feature

apisix/plugins/request-id.lua
```diff
@@ -14,36 +14,163 @@
...
+                local handler = function(premature, etcd_cli, lease_id)
+                    local _, err4 = etcd_cli:keepalive(lease_id)
+                    if err4 then
+                        snowflake_init = nil
+                        worker_number = nil
+                        core.log.error("snowflake worker_number lease faild.")
+                    end
+                    core.log.info("snowflake worker_number lease success.")
+                end
+                ngx.timer.every(attr.snowflake.worker_number_interval,
```
spacewander, line 117: Can we write it via: https://github.com/apache/apisix/blob/89765ba188834c1b4c9f55088a28db25a959f39e/apisix/plugins/error-log-logger.lua#L198

apisix/plugins/request-id.lua
```diff
@@ -14,36 +14,163 @@
...
+    type = "object",
+    properties = {
+        snowflake = {
+            type = "object",
+            properties = {
+                enable = {type = "boolean"},
+                snowflake_epoc = {type = "integer", minimum = 1, default = 1609459200000},
+                node_id_bits = {type = "integer", minimum = 1, default = 5},
+                sequence_bits = {type = "integer", minimum = 1, default = 10},
+                datacenter_id_bits = {type = "integer", minimum = 1, default = 5},
```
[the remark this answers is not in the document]
spacewander: Maybe "default = 7" is better? One of the largest APISIX user have more than 2000 workers. They don't want to be in trouble when using the default value.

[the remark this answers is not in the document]
dickens7: `default=5` is the setting of standard snowflake algorithm. Setting `default=7` may be confusing for people who are not familiar with the snowflake algorithm. The Snowflake configuration is off by default, so we can introduce the settings in document clearly.

apisix/plugins/request-id.lua
```diff
@@ -65,4 +191,47 @@ function _M.header_filter(conf, ctx)
...
+        return
+    end
+    if attr.snowflake.enable then
+        if process.type() == "worker" then
+            ngx.timer.at(0, next_id)
+        end
+    end
+end
+
+function _M.api()
```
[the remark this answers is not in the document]
dickens7: The API is exposed to make it  a  global ID generator. The real applicatipon may require multiple ID while the request-id can only be passed to one proxy application.If  there is a concern about the  exposing of  the algorithm mechanism, I can implement it in other way, such as, registering a specific route and configure the request-id plugin.

apisix/plugins/request-id.lua
dickens7: > Hi @dickens7 do you mind also add documentations of this feature

ok

apisix/plugins/request-id.lua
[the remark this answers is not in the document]
spacewander: Let's make it optional and submit it in the next PR. This one is already big enough.

apisix/plugins/request-id.lua
```diff
@@ -41,9 +72,134 @@ function _M.check_schema(conf)
...
+end
+
+
+-- Split 'Worker Number' into 'Worker ID' and 'datacenter ID'
+local function split_worker_number(worker_number, node_id_bits, datacenter_id_bits)
+    local num = bit.tobit(worker_number)
+    local worker_id = bit.band(num, math_pow(2, node_id_bits) - 1) + 1
+    num = bit.rshift(num, node_id_bits)
+    local datacenter_id = bit.band(num, math_pow(2, datacenter_id_bits) - 1) + 1
+    return worker_id, datacenter_id
```
spacewander, line 153: Why start the id from 1? It seems 0 is also valid.

apisix/plugins/request-id.lua
```diff
@@ -41,9 +72,134 @@ function _M.check_schema(conf)
...
+                    if err4 then
+                        snowflake_inited = nil
+                        worker_number = nil
+                        core.log.error("snowflake worker_number: " .. id .." lease faild.")
+                    end
+                    start_at = now
+                    core.log.info("snowflake worker_number: " .. id .." lease success.")
+                end
+
+                timers.register_timer("plugin#request-id", handler)
```
spacewander, line 129: Also need to cancel it like this:
https://github.com/apache/apisix/blob/32f9d699a986864612963868d7d23408e3860550/apisix/plugins/error-log-logger.lua#L202-L205

apisix/plugins/request-id.lua
```diff
@@ -41,9 +72,134 @@ function _M.check_schema(conf)
...
 
+-- Generates the current process worker number
+local function gen_worker_number(max_number)
+    if worker_number == nil then
+        local etcd_cli, prefix = core.etcd.new()
+        local prefix = prefix .. "/plugins/request-id/snowflake/"
+        local uuid = uuid.generate_v4()
+        local id = 1
+        while (id <= max_number) do
+            ::continue::
```
spacewander, line 83: continue back may causes the id > max_number. For instance:
```
local id = 0
while id <= 3 do
    ::continue::
    if id == 3 then
        print(id)
        id = id + 1
        goto continue
    end
    print(id)
    id = id + 1
end
-- 0, ..., 4
```

apisix/plugins/request-id.lua
```diff
@@ -41,9 +72,134 @@ function _M.check_schema(conf)
...
+            local res, _ = etcd_cli:grant(attr.snowflake.worker_number_ttl)
```
tokers, line 84: I am afraid that we shouldn't ignore this error.

apisix/plugins/request-id.lua
dickens7: It's going to be a bug to start at 1 here, let me fix that

apisix/plugins/request-id.lua
dickens7: Indeed, I will fix it

apisix/plugins/request-id.lua
dickens7: Okay, I'm going to add a check

apisix/plugins/request-id.lua
dickens7: When does `_M. destroy()` triggered? I don't think there should be a need to `timers.unregister_timer` in `_M. destroy()`

spacewander: When the plugin is hot disabled.

dickens7:
TODO:  upgrade api7/lua-snowflake to 2.1-1

docs/zh/latest/plugins/request-id.md
```diff
@@ -71,6 +72,99 @@ X-Request-Id: fe32076a-d0a5-49a6-a361-6c244c1df956
 ......
 ```
 
+### 使用 snowflake 算法
```
spacewander, line 75: Let's update the English doc too. We can use an online translator for help.

apisix/plugins/request-id.lua
```diff
@@ -14,25 +14,58 @@
...
+    type = "object",
+    properties = {
+        snowflake = {
+            type = "object",
+            properties = {
+                enable = {type = "boolean"},
+                snowflake_epoc = {type = "integer", minimum = 1, default = 1609459200000},
+                data_machine_bits = {type = "integer", minimum = 1, maximum = 31, default = 12},
+                sequence_bits = {type = "integer", minimum = 1, default = 10},
+                delta_offset = {type = "integer", default = 1, enum = {1, 10, 100, 1000}},
```
spacewander, line 56: I think we can move delta_offset to the next PR. This one is big enough. Note that it is not easy to catch people to review PR.

docs/zh/latest/plugins/request-id.md
dickens7: First confirm that the document is ok, I am translating to English

apisix/plugins/request-id.lua
dickens7: ok

docs/zh/latest/plugins/request-id.md
spacewander: OK

apisix/plugins/request-id.lua
```diff
@@ -41,9 +74,142 @@ function _M.check_schema(conf)
...
+
+
+-- generate snowflake id
+local function next_id()
+    if snowflake_inited == nil then
+        snowflake_init()
+    end
+    return snowflake:next_id()
+end
+
```
nic-chen, line 200: Could we put them under `utils`? Snowflake ID generating should also be used elsewhere.

zuiyangqingzhou: agree

dickens7: Let's implement this in the next pr

docs/zh/latest/plugins/request-id.md
dickens7: The English document has been added

apisix/plugins/request-id.lua
dickens7: resolved

apisix/plugins/request-id.lua
dickens7: delta_offset has been removed from this PR

apisix/plugins/request-id.lua
```diff
@@ -245,5 +245,8 @@ function _M.init()
     end
 end
 
+function _M.destroy()
+    timers.unregister_timer("plugin#request-id")
```
spacewander, line 249: Better to check if the timer is registered?

dickens7: Is it necessary to do this?  I have not seen similar examples elsewhere.

spacewander: If we don't register a timer, why unregister it?

dickens7: You have a point, I added the logic of judgment.

apisix/plugins/request-id.lua
```diff
@@ -41,9 +74,142 @@ function _M.check_schema(conf)
...
+    if data_machine == nil then
+        local etcd_cli, prefix = core.etcd.new()
+        local prefix = prefix .. "/plugins/request-id/snowflake/"
+        local uuid = uuid.generate_v4()
+        local id = 1
+        ::continue::
+        while (id <= max_number) do
+            local res, err = etcd_cli:grant(attr.snowflake.data_machine_ttl)
+            if err then
+                core.log.error("Etcd grant failure, err: ".. err)
```
starsz, line 88: Do here need to go to continue?

apisix/plugins/request-id.lua
```diff
@@ -41,9 +74,142 @@ function _M.check_schema(conf)
...
+            else
+                data_machine = id
+
+                local _, err3 =
+                    etcd_cli:set(
+                    prefix .. tostring(id),
+                    uuid,
+                    {
+                        prev_kv = true,
+                        lease = res.body.ID
```
starsz, line 106: Need to check the `res`, it will be nil if `etcd_cli:grant` failed.

apisix/plugins/request-id.lua
dickens7: Should not continue here.
fixed.

apisix/plugins/request-id.lua
dickens7: Now `etcd_cli:grant` will exit the current loop when the return value `err` is `true`, do I need to check `res` here?

spacewander:
Please solve the conflict.

apisix/plugins/request-id.lua
```diff
@@ -41,9 +74,143 @@ function _M.check_schema(conf)
...
+        local etcd_cli, prefix = core.etcd.new()
+        local prefix = prefix .. "/plugins/request-id/snowflake/"
+        local uuid = uuid.generate_v4()
+        local id = 1
+        ::continue::
+        while (id <= max_number) do
+            local res, err = etcd_cli:grant(attr.snowflake.data_machine_ttl)
+            if err then
+                core.log.error("Etcd grant failure, err: ".. err)
+                goto continue
```
spacewander, line 89: There is a chance to cause an infinite loop?

apisix/plugins/request-id.lua
```diff
@@ -14,25 +14,58 @@
...
     }
 }
 
+local attr_schema = {
+    type = "object",
+    properties = {
+        snowflake = {
+            type = "object",
+            properties = {
+                enable = {type = "boolean"},
```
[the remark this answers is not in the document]
dickens7: fixed

docs/en/latest/plugins/request-id.md
```diff
@@ -72,6 +72,60 @@ X-Request-Id: fe32076a-d0a5-49a6-a361-6c244c1df956
...
+      sequence_bits: 10
+      data_machine_ttl: 30
+      data_machine_interval: 10
+```
+
+#### Configuration parameters
+
+| Name                | Type    | Requirement   | Default        |  Valid  | Description                    |
+| ------------------- | ------- | ------------- | -------------- | ------- | ------------------------------ |
+| enable                     | boolean  | required   | false          |  | When set it to true, enable the snowflake algorithm.  |
```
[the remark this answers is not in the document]
dickens7: Modified jsonSchema `required = {"enable", "snowflake_epoc "}`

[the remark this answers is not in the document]
spacewander: Once you provide a default value, the field is actually optional, as users don't need to specify their values.

apisix/plugins/request-id.lua
```diff
@@ -41,9 +75,144 @@ function _M.check_schema(conf)
...
+        ::continue::
+        while (id <= max_number) do
+            local res, err = etcd_cli:grant(attr.snowflake.data_machine_ttl)
+            if err then
+                id = id + 1
+                core.log.error("Etcd grant failure, err: ".. err)
+                goto continue
+            end
+
+            local _, err1 = etcd_cli:setnx(prefix .. tostring(id), uuid)
```
tokers, line 94: So we have to **write** ETCD so that we can write the uuid?

docs/en/latest/plugins/request-id.md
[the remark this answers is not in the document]
dickens7: Got it. I'll make it optional

apisix/plugins/request-id.lua
dickens7: The purpose of writing etcd here is to generate `DataMachineId` for snowflake. The value is uuid because `etcd_cli:setnx` still returns `succeeded` even if the key exists. UUID is used for verification after `etcd_cli:setnx`.

spacewander (approved):

conf/config-default.yaml
```diff
@@ -348,3 +348,12 @@ plugin_attr:
     report_ttl: 3600     # live time for server info in etcd (unit: second)
   dubbo-proxy:
     upstream_multiplex_count: 32
+  request-id:
+    snowflake:
+      enable: false
+      snowflake_epoc: 1609459200000   # the starting timestamp is expressed in milliseconds
+      data_machine_bits: 12           # data machine bit, maximum 31, because Lua cannot do bit operations greater than 31
```
tzssangglass, line 355: should check `data_machine_bits` + `sequence_bits` = 22 always? may the test case need to cover it.

docs/en/latest/plugins/request-id.md
```diff
@@ -72,6 +73,60 @@ X-Request-Id: fe32076a-d0a5-49a6-a361-6c244c1df956
...
+| data_machine_bits          | integer  | optional   | 12             |  | Maximum number of supported machines (processes) `1 << data_machine_bits` |
+| sequence_bits              | integer  | optional   | 10             |  | Maximum number of generated ID per millisecond per node `1 << sequence_bits` |
+| data_machine_ttl           | integer  | optional   | 30             |  | Valid time of registration of 'data_machine' in 'etcd' (unit: seconds) |
+| data_machine_interval      | integer  | optional   | 10             |  | Time between 'data_machine' renewal in 'etcd' (unit: seconds) |
+
+- `snowflake_epoc` default start time is  `2021-01-01T00:00:00Z`, and it can support `69 year` approximately to `2090-09-0715:47:35Z` according to the default configuration
+- `data_machine_bits` corresponds to the set of workIDs and datacEnteridd in the snowflake definition. The plug-in aslocates a unique ID to each process. Maximum number of supported processes is `pow(2, data_machine_bits)`. The default number of `12 bits` is up to `4096`.
+- `sequence_bits` defaults to `10 bits` and each process generates up to `1024` ID per second
```
tzssangglass, line 108: The `per second` here and the `per millisecond` above confuse me

t/plugin/request-id.t
```diff
@@ -470,3 +470,268 @@ GET /t
...
+                    end
+                    local id = res.headers["X-Request-Id"]
+                    if not id then
+                        return -- ignore if the data is not synced yet.
+                    end
+                    if ids[id] == true then
+                        ngx.say("ids not unique")
+                        return
+                    end
+                    ids[id] = true
```
tzssangglass, line 633: Is there any way to check that the length of the `ids` has reached a fixed value? such as 100, or at least 2. I am worried that there is no data in the ids due to sync.

tzssangglass:
LGTM

conf/config-default.yaml
dickens7: `data_machine_bits` and `sequence_bits` are not fixed and can be configured according to different requirements

docs/en/latest/plugins/request-id.md
dickens7: This is the definition of two parts of the snowflake algorithm. Millisecond is the definition of the timestamp part, which means that the timestamp part is in milliseconds. Second is the definition of the sequence number part, which means that the number of id can be generated per second

starsz (approved):

apisix/plugins/request-id.lua
```diff
@@ -53,7 +221,6 @@ function _M.rewrite(conf, ctx)
     end
 end
 
-
```
monkeyDluffy6017, line 56: why remove this blank?

PR #4560, opened 2021-07-08
ci: refactor chaos test with ginkgo

Yiyiyimu:

Split #4315 to make review easier

### Pre-submission checklist:


.github/workflows/chaos.yml
```diff
@@ -40,30 +40,35 @@ jobs:
...
-          git clone https://github.com/api7/etcd-operator.git --depth 1
-          bash etcd-operator/example/rbac/create_role.sh
-          kubectl create -f etcd-operator/example/deployment.yaml
-          bash ./t/chaos/setup_chaos_utils.sh ensure_pods_ready etcd-operator "True" 30
-          kubectl create -f etcd-operator/example/example-etcd-cluster.yaml
-          bash ./t/chaos/setup_chaos_utils.sh ensure_pods_ready etcd "True True True" 30
+          curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm install etcd bitnami/etcd --set replicaCount=3 --set auth.rbac.enabled=false
+          bash ./t/chaos/utils/setup_chaos_utils.sh ensure_pods_ready app.kubernetes.io/instance=etcd "True True True" 30
```
tokers, line 48: You can just use `kubectl wait`.

Yiyiyimu: Wow it's cool! Thanks for telling me that!

spacewander:
The test may fail sometimes? https://github.com/apache/apisix/runs/3024862769

t/chaos/killetcd/killetcd.go
```diff
@@ -0,0 +1,190 @@
...
+		gomega.Expect(err).To(gomega.BeNil())
+		listOption := client.MatchingLabels{"app": "apisix-gw"}
+		apisixPods, err := utils.GetPods(cliSet.CtrlCli, metav1.NamespaceDefault, listOption)
+		gomega.Expect(err).To(gomega.BeNil())
+		gomega.Ω(len(apisixPods)).Should(gomega.BeNumerically(">", 0))
+		apisixPod = &apisixPods[0]
+	})
+
+	stopChan := make(chan bool)
+	defer ginkgo.It("restore test environment", func() {
```
tokers, line 90: Why not use `ginkgo.AfterEach`?

t/chaos/killetcd/killetcd.go
```diff
@@ -0,0 +1,190 @@
...
+			Scheduler: &v1alpha1.SchedulerSpec{
+				Cron: "@every 10m",
+			},
+		},
+	}
+}
+
+var _ = ginkgo.Describe("Test Get Success When Etcd Got Killed", func() {
+	e := httpexpect.New(ginkgo.GinkgoT(), utils.Host)
+	eSilent := httpexpect.WithConfig(httpexpect.Config{
```
tokers, line 68: A recommended best practice is putting all your initialization logics into `ginkgo.BeforeEach`.

t/chaos/killetcd/killetcd.go
```diff
@@ -0,0 +1,190 @@
...
+		Reporter: httpexpect.NewAssertReporter(ginkgo.GinkgoT()),
+		Printers: []httpexpect.Printer{
+			utils.NewSilentPrinter(ginkgo.GinkgoT()),
+		},
+	})
+
+	var cliSet *utils.ClientSet
+	var apisixPod *v1.Pod
+	var err error
+	ginkgo.It("init client set", func() {
```
tokers, line 79: If you'd like to use `It` mechanism to separate the test suites, you may try ginkgo table.

t/chaos/killetcd/killetcd.go
Yiyiyimu: ginkgo `AfterEach` would take effect after each `It`, which I expect only work once when all `It` finish. The same with `BeforeEach` mentioned below.

Yiyiyimu:
> The test may fail sometimes? [apache/apisix/runs/3024862769](https://github.com/apache/apisix/runs/3024862769)

Test another seven times and none of it failed so I guess we could use it for now. Add a debug tool so when it failed I could ssh inside and see what's going wrong

spacewander (approved):

t/chaos/killetcd/killetcd.go
tokers: Nope, if you want to execute some codes after all the `It` blocks finishing, actually the best way is using `ginkgo.AfterSuite`, just organization these related `It` blocks into a `Describe` block.

Anyway, this is trivial. We can optimize it in the future.

tokers (approved):

t/chaos/killetcd/killetcd.go
Yiyiyimu: Thank you for the suggestion! 

But actually it's also not to execute some codes after all `It` blocks finishing, but more like to do it after each `Describe`, since I try to keep the initial status of external dependency in each test file the same. Do you know if there are some better methods to deal with this?

Issue #4552, opened 2021-07-07
bug:  request prometheus metrics.  throw etcd key not found: [/routes] warning


billnote:
### Issue description
request prometheus metrics api.  throw etcd key not found: [/routes] warning

### Environment

Bug report without environment information will be ignored or closed.

* apisix version (cmd: `apisix version`): 2.5 (i think all the version has this bug)
* OS (cmd: `uname -a`): Linux  3.10.0-1160.el7.x86_64
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):  nginx version: openresty/1.19.3.2
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API): etcd-release-3.4.13

### Minimal test code / Steps to reproduce the issue

Bug report without steps to reproduce will be ignored or closed.

1. request prometheus metrics api
2. error.log has this warn:  >>>>>> etcd key not found: [/routes] <<<<<

billnote:
This line of code causes the bug.
[https://github.com/apache/apisix/blob/3602646ff7a46b10db58741763f413b40f8d9547/apisix/plugins/prometheus/exporter.lua#L311](url)

Because etcd key has the prefix in default config. Use below code to solve this bug:
```
        local key = "/routes/"
        local local_conf = core.config.local_conf()
        if local_conf.etcd and local_conf.etcd.prefix then
            key = local_conf.etcd.prefix .. key
        end
        
        local res, _ = config:getkey(key)
```

billnote:
This line of code causes the bug.
[https://github.com/apache/apisix/blob/3602646ff7a46b10db58741763f413b40f8d9547/apisix/plugins/prometheus/exporter.lua#L311](url)

Because etcd key has the prefix in default config. Use below code to solve this bug:
```
        local key = "/routes/"
        local local_conf = core.config.local_conf()
        if local_conf.etcd and local_conf.etcd.prefix then
            key = local_conf.etcd.prefix .. key
        end
        
        local res, _ = config:getkey(key)
```

billnote:
This line of code causes the bug.
[https://github.com/apache/apisix/blob/3602646ff7a46b10db58741763f413b40f8d9547/apisix/plugins/prometheus/exporter.lua#L311](url)

Because etcd key has the prefix in default config. Use below code to solve this bug:
```
        local key = "/routes/"
        local local_conf = core.config.local_conf()
[...]

PR #4561, opened 2021-07-08
feat: add prefix when using config_etcd:getkeys

billnote:

fix: https://github.com/apache/apisix/issues/4552

### Pre-submission checklist:


apisix/plugins/prometheus/exporter.lua
```diff
@@ -308,7 +308,13 @@ function _M.collect()
                            "processing metrics endpoint: ", err)
         end
 
-        local res, _ = config:getkey("/routes")
+        local key = "/routes/"
+        local local_conf = core.config.local_conf()
```
spacewander, line 312: Better to move it into getkey

spacewander:
Where do you find "etcd key not found"  in the error log?

After consulting with the author of this feature, look like it doesn't require a real key. Actually, request any key from etcd will return the "X-Etcd-Index". A non-existed key is preferred because it doesn't return too much data, as we only care about the header.

billnote:
Sorry, I check the code, This waring will only appear in our internal version.

But  getKey function  don't need add prefix automatically?  I will close this PR, if we don't need.

apisix/core/config_etcd.lua
```diff
@@ -71,6 +71,11 @@ local function getkey(etcd_cli, key)
         return nil, "not inited"
     end
 
+    local local_conf, _ = config_local.local_conf()
```
spacewander, line 74:
```suggestion
    local local_conf = config_local.local_conf()
```

apisix/core/config_etcd.lua
```diff
@@ -71,6 +71,11 @@ local function getkey(etcd_cli, key)
...
+    if local_conf and local_conf.etcd and local_conf.etcd.prefix then
+        key = local_conf.etcd.prefix .. key
+    end
```
spacewander, line 77: Should be added to _M.getkey?

apisix/plugins/prometheus/exporter.lua
```diff
@@ -308,7 +308,7 @@ function _M.collect()
                            "processing metrics endpoint: ", err)
         end
 
-        local res, _ = config:getkey("/routes")
+        local res, _ = config:getkey("/routes/")
```
spacewander, line 311: Let's change the key name.

apisix/core/config_etcd.lua
billnote: ok

spacewander (commented):
Need a test in https://github.com/apache/apisix/blob/master/t/core/config_etcd.t

apisix/plugins/prometheus/exporter.lua
```diff
@@ -308,7 +308,8 @@ function _M.collect()
                            "processing metrics endpoint: ", err)
         end
 
-        local res, _ = config:getkey("/routes")
+        -- use phantom key to get etcd index
```
spacewander, line 311: The comment is too short compared with my explanation

spacewander (approved):

tokers (approved):

PR #4580, opened 2021-07-10
change: enable stream proxy only by default

spacewander:

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -132,6 +132,7 @@ stream {
 }
 {% end %}
 
+{% if not (stream_proxy and stream_proxy.only ~= false) then %}
 http {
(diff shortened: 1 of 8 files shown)


gxthrj:
It is confusing for using the configuration in stream-proxy to control the http-proxy.
It is better to keep configuration the same level with stream and http proxy.

spacewander:
The HTTP proxy already at the top level of apisix. There can't be "the same level" unless we downgrade the level of HTTP proxy.

tzssangglass (approved):

tzssangglass (approved):

nic-chen (approved):

starsz (approved):

oldthreefeng:
so,  You need to declare `only` this is a global variable.  otherwise confused .....

PR #4587, opened 2021-07-12
feat(plugin): Add new plugin ua-restriction for bot spider restriction

arthur-zhang:
Support bot spider restriction, eg Baidu spider、go-httpclient etc

```
Mozilla/5.0 (iPhone; CPU iPhone OS 9_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13B143 Safari/601.1 (compatible; Baiduspider-render/2.0; +http://www.baidu.com/search/spider.html)
```

### Pre-submission checklist:


apisix/plugins/bot-restriction.lua
```diff
@@ -0,0 +1,180 @@
...
+
+local _M = {
+    version = 0.1,
+    priority = 2999,
+    name = plugin_name,
+    schema = schema,
+}
+
+-- List taken from https://github.com/ua-parser/uap-core/blob/master/regexes.yaml
+local well_known_bots = {
```
spacewander, line 62: I think we should not hard code the UA list, as it could not be updated in time. It would be better to provide a mechanism but not the tool to check the UA.

t/plugin/bot-restriction.t
```diff
@@ -0,0 +1,699 @@
...
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+
+    location /disable {
```
spacewander, line 591: the /disable is unused.

t/plugin/bot-restriction.t
```diff
@@ -0,0 +1,699 @@
...
+        }
+    }
+--- response_body
+property "message" validation failed: wrong type: expected string, got number
+done
+
+
+
+=== TEST 6: set blacklist
+
```
spacewander, line 154: The blank line can be removed?

t/plugin/bot-restriction.t
```diff
@@ -0,0 +1,699 @@
...
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 7: hit route and user-agent in blacklist
+
```
spacewander, line 192: Ditto

apisix/plugins/bot-restriction.lua
```diff
@@ -0,0 +1,180 @@
...
+
+function _M.access(conf, ctx)
+    local user_agent = core.request.header(ctx, "User-Agent")
+
+    if not user_agent then
+        return
+    end
+    -- ignore multiple instances of request headers
+    if type(user_agent) == "table" then
+        return
```
spacewander, line 168: Why ignore the UA?

apisix/plugins/bot-restriction.lua
```diff
@@ -0,0 +1,180 @@
...
+    end
+    local match, err = lrucache_useragent(user_agent, conf, match_user_agent, user_agent, conf)
+    if err then
+        return
```
spacewander, line 172: Better to log the err?

apisix/plugins/bot-restriction.lua
```diff
@@ -0,0 +1,180 @@
...
+        },
+        blacklist = {
+            type = "array",
+            minItems = 1
+        },
+    },
+    additionalProperties = false,
+}
+
+local plugin_name = "bot-restriction"
```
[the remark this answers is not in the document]
arthur-zhang: I do not think so，this plugin is for spider detection，and include most common spider us.

apisix/plugins/bot-restriction.lua
arthur-zhang: Most spider bot UA contains “bot” or “spider” or “crawler” or `dev http client`, the regex expression covers most common case，it will not update very frequently.

apisix/plugins/bot-restriction.lua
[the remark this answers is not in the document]
spacewander: A real bot-detection in the industry area is not just spider detection and UA check.

conf/config-default.yaml
```diff
@@ -252,6 +252,7 @@ plugins:                          # plugin list (sorted by priority)
   - batch-requests                 # priority: 4010
   - cors                           # priority: 4000
   - ip-restriction                 # priority: 3000
+  - bot-restriction                # priority: 3000
```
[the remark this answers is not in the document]
arthur-zhang: OK

apisix/plugins/bot-restriction.lua
arthur-zhang: OK

apisix/plugins/bot-restriction.lua
spacewander: But it will be updated, isn't it? Better to require the user to choose their list instead of shipping a stale one.

apisix/plugins/bot-restriction.lua
arthur-zhang: this corner case is that the user-agent become table  when send multiple user-agent. Almost all the  bot or http-client will not send request like this，i think we ignore it is a better choice.

spacewander: So if they send other UA, the check can be bypassed? This is not a good idea, especially in an open source project...

apisix/plugins/bot-restriction.lua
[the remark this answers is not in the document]
arthur-zhang: The plugin is for the common BaiduSpider、360Spider and some dev tools  detection.
That gentleman can be prevented, the villain never fortification.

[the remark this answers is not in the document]
arthur-zhang: I found similar function in krakend

https://www.krakend.io/docs/throttling/botdetector/

apisix/plugins/bot-restriction.lua
arthur-zhang: User can update the list in whitelist or blacklist configuration to support the ua not listed in our package

t/plugin/bot-restriction.t
arthur-zhang: done

apisix/plugins/bot-restriction.lua
arthur-zhang: ok, I will check the table

apisix/plugins/bot-restriction.lua
spacewander: So why ship a stale one and ask the user to update it?

arthur-zhang: We can support other plugin for client restrction like ua, ip, or other infomartion. This plugin is just for users not want to add bunch of ua regex rules.

arthur-zhang: This plugin is only for simplify usage.
If user do not want to use the plugin, they can use other restriction plugin and modify the rules the want.

apisix/plugins/bot-restriction.lua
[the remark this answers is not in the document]
spacewander: Don't believe the nonsense bullshit. This way can only kick out script boys. But for the real hacker, checking the UA is definitely not enough.

> We have to use the product of professional security company to do the feature you mentioned.

That's it. A real bot detection system should be as professional as them instead of just doing UA checks and declaring this solves the problem. People will laugh at APISIX. We should provide a mechanism that the professional security company can use to build a gateway, but not declare we are a security gateway.

spacewander (changes requested):
Provide **a mechanism** but **not the tool** to check the UA

apisix/plugins/bot-restriction.lua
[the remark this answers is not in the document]
arthur-zhang: is it more suitable to  change plugin name to ua-restriction and remove the hard-coded ua list?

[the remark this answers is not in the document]
spacewander: Of course, yes.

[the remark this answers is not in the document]
arthur-zhang: Got it

apisix/plugins/bot-restriction.lua
```diff
@@ -0,0 +1,177 @@
...
+            type = "string",
+            minLength = 1,
+            maxLength = 1024,
+            default = "Not allowed"
+        },
+        whitelist = {
+            type = "array",
+            minItems = 1
+        },
+        blacklist = {
```
tokers, line 44: What about `allowlist` and `blocklist`, we should avoid using these sensitive words.

arthur-zhang: ok

docs/en/latest/plugins/ua-restriction.md
```diff
@@ -0,0 +1,128 @@
...
+
+The `ua-restriction` can restrict access to a Service or a Route by either `allowlist` or `denylist` `User-Agent`.
+
+## Attributes
+
+| Name      | Type          | Requirement | Default | Valid | Description                              |
+| --------- | ------------- | ----------- | ------- | ----- | ---------------------------------------- |
+| allowlist | array[string] | optional    |         |       | List of User-Agent of allowlist. |
+| denylist | array[string] | optional    |         |       | List of User-Agent of denylist. |
+| message | string | optional             | Not allowed. | [1, 1024] | Message of deny reason. |
```
spacewander, line 42: We should clarify that `[1, 1024]` is the length.

apisix/plugins/ua-restriction.lua
```diff
@@ -0,0 +1,121 @@
...
+
+    if not user_agent then
+        return
+    end
+    local match = MATCH_NONE
+    if type(user_agent) == "table" then
+        for _, v in ipairs(user_agent) do
+            if type(v) == "string" then
+                match = lrucache_useragent(v, conf, match_user_agent, v, conf)
+                if match > MATCH_ALLOW then
```
spacewander, line 107: Why not use `match > MATCH_DENY`?

apisix/plugins/ua-restriction.lua
```diff
@@ -0,0 +1,121 @@
...
+        denylist = {
+            type = "array",
+            minItems = 1
+        },
+    },
+    anyOf = {
+        {required = {"allowlist"}},
+        {required = {"denylist"}},
+    },
+    minProperties = 1,
```
spacewander, line 52: The `minProperties` is useless?

apisix/plugins/ua-restriction.lua
```diff
@@ -0,0 +1,121 @@
...
+local function match_user_agent(user_agent, conf)
+    user_agent = str_strip(user_agent)
+    if conf.allowlist then
+        for _, rule in ipairs(conf.allowlist) do
+            if re_find(user_agent, rule, "jo") then
+                return MATCH_ALLOW
+            end
+        end
+    end
+
```
tzssangglass, line 74: if a user_agent both in allowlist and denylist?

I think it's better to judge denylist first

spacewander: The allow list is more important than the deny list. Other restriction plugin also check the allow list first: https://github.com/apache/apisix/blob/master/apisix/plugins/ip-restriction.lua

tzssangglass: got

spacewander:
@arthur-zhang
Would you like to update the PR? Thanks!

arthur-zhang:
> @arthur-zhang
> Would you like to update the PR? Thanks!

I will update soon.

apisix/plugins/ua-restriction.lua
arthur-zhang: i think this check should be MATCH_ALLOW, if ua in deny list, the result should be MATCH_DENY, otherwise, the result is MATCH_ALLOW or MATCH_NONE

spacewander (approved):

t/plugin/ua-restriction.t
```diff
@@ -0,0 +1,584 @@
...
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 16: hit route and user-agent in both allowlist and denylist, part 1
```
tzssangglass, line 333:
```suggestion
=== TEST 16: hit route and user-agent in both allowlist and denylist, part 1
```

```suggestion
=== TEST 16: hit route and user-agent in both allowlist and denylist, pass(part 1)
```

t/plugin/ua-restriction.t
```diff
@@ -0,0 +1,584 @@
...
+--- request
+GET /hello
+--- more_headers
+User-Agent:foo/bar
+--- error_code: 200
+
+
+
+=== TEST 17: hit route and user-agent in both allowlist and denylist, part 2
```
tzssangglass, line 342: ditto

t/plugin/ua-restriction.t
```diff
@@ -0,0 +1,584 @@
...
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/hello",
+                    "plugins": {
+                        "ua-restriction": {
+                            "denylist": [
+                                "foo"
+                            ],
+                            "disable": true
```
tzssangglass, line 571: What does `disable` mean here? I don't see it in the plugin properties doc.

arthur-zhang: disable is built-in properties of plugin schema

t/plugin/ua-restriction.t
arthur-zhang: done

spacewander:
Let's merge master to make CI pass

arthur-zhang:
Seems CI system is not working

spacewander:
Err, since the server is changed again...

arthur-zhang:
> Err, since the server is changed again...

image

t/plugin/ua-restriction.t
```diff
@@ -0,0 +1,584 @@
...
+passed
+
+
+
+=== TEST 7: hit route and user-agent in denylist
+--- request
+GET /hello
+--- more_headers
+User-Agent:my-bot1
+--- error_code: 403
```
tokers, line 196: Also check the response body.

t/plugin/ua-restriction.t
```diff
@@ -0,0 +1,584 @@
...
+
+
+
+=== TEST 8: hit route and user-agent in denylist with multiple
+--- request
+GET /hello
+--- more_headers
+User-Agent:my-bot1
+User-Agent:my-bot2
+--- error_code: 403
```
tokers, line 206: Ditto.

apisix/plugins/ua-restriction.lua
```diff
@@ -0,0 +1,120 @@
...
+        return false, err
+    end
+
+    return true
+end
+
+function _M.access(conf, ctx)
+    local user_agent = core.request.header(ctx, "User-Agent")
+
+    if not user_agent then
```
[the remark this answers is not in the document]
spacewander: Do you mean something like this? https://github.com/apache/apisix/blob/44ead1f46ec5b89666a1f700ff16eb00b00174e8/apisix/plugins/referer-restriction.lua#L28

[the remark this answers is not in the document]
tokers: Yep.

tokers (approved):

spacewander (approved):

Issue #4583, opened 2021-07-11
request help: suggestions on improving the default_conn_delay of the limit-conn plugin


zuiyangqingzhou:
### Issue description
The default_conn_delay of the limit-conn plug-in is used to control the delay of forwarding to upstreams when the connection exceeds a certain threshold, but after actual testing, it is found that this delay time is not strictly in accordance with the configured delay time.

https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/conn.lua#L98-L101

It is found in the source code of lua-resty-limit-traffic that there is a certain calculation logic for the delay time. If req_latency is nil, the delay time is strictly in accordance with the user-defined delay time

I think a control parameter should be added in apisix to let users decide whether the delay time should be strictly in accordance with the configured time

What do you think？

spacewander:
PR is welcome!

zuiyangqingzhou:
ok, i will make a try

PR #4585, opened 2021-07-11
fix: the limit-conn plugin cannot effectively intercept requests in special scenarios

zuiyangqingzhou:
### What this PR does / why we need it:

fix [4556](https://github.com/apache/apisix/issues/4556)

ps: I will add test cases in the next two days

PR #4604, opened 2021-07-14
feat: add additional option to control the default_conn_delay option of the limit-conn plugin

zuiyangqingzhou:
feat: [issues](https://github.com/apache/apisix/issues/4583)

ps:  

I'm thinking how to do the unit test here

Because I cannot verify the logic of the third-party library [lua-resty-limit-traffic](https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/conn.lua#L89-L104), I can only verify whether nil is passed for  the req_latency parameter, but what should I do with this unit test?

Can you give some advice?


docs/en/latest/plugins/limit-conn.md
```diff
@@ -40,7 +40,7 @@ Limiting request concurrency plugin.
...
+| delay_strict_mode  | boolean | optional    | false   | [true,false]                                                                              | enable the strict mode of the latency seconds. If you set this option to `true`, it will run strictly according to the latency seconds you set without additional calculation logic.                                                                                                                                                                                                                                                                                                                      |
```
tokers, line 43: I think the option name is not so straightforward but I didn't come out with a more proper one :(

zuiyangqingzhou: 😂

spacewander: What about `only_use_default_delay`?

zuiyangqingzhou: `only_use_default_delay`,   `use_default_conn_delay`,  `apply_default_conn_delay`,  `stint_default_conn_delay`, Which of the four do you think is better, or do you have better suggestions?

spacewander: I vote for the first one. What about others' opinions?

zuiyangqingzhou: @tokers  what do you think?

tokers: Vote for the first one too, it's easy to understand.

zuiyangqingzhou: Okay, I will change it to this one. 
What about the test case I mentioned above? Any suggestions? @spacewander @tokers

spacewander: You can take a look at the test in https://github.com/apache/apisix/pull/4585

docs/zh/latest/plugins/limit-conn.md
```diff
@@ -30,7 +30,7 @@ title: limit-conn
...
-| delay_strict_mode  | boolean | optional | false  | [true,false]                                                                              | 延迟时间的严格模式. 如果设置为`true`的话，将会严格按照设置的时间来进行延迟                                                                                                                                                                                                                                                                                                                                                                           |
+| only_use_default_delay  | boolean | optional | false  | [true,false]                                                                              | 延迟时间的严格模式. 如果设置为`true`的话，将会严格按照设置的时间来进行延迟                                                                                                                                                                                                                                                                                                                                                                           |
```
tokers, line 33: `.` is the period symbol in English, shall we use `。`?

t/plugin/limit-conn2.t
```diff
@@ -178,3 +178,76 @@ qr/request latency is/
...
+=== TEST 6: hit route
+--- log_level: debug
+--- request
+GET /hello1
+--- grep_error_log eval
+qr/request latency is/
+--- grep_error_log_out
+request latency is
+request latency is
+
```
tokers, line 254: Redundant empty line.

zuiyangqingzhou: ok, i will fix it

tokers (approved):

t/plugin/limit-conn2.t
```diff
@@ -178,3 +178,76 @@ qr/request latency is/
...
+passed
+
+
+
+=== TEST 6: hit route
+--- log_level: debug
+--- request
+GET /hello1
+--- grep_error_log eval
+qr/request latency is/
```
spacewander, line 250: Need to check latency in this mode is `nil`.

zuiyangqingzhou: I don't know exactly what you mean, 

> https://github.com/apache/apisix/pull/4604/files/47dc32979a94977aa8e3d0e73270dc88a9c43149#diff-2f0a07663e8e90e9f64519ee946aaf82729c8bb43e4f9100902e87d0a325f78eR251-R253

hasn't grep_error_log_out already check that the latency is nil? 
Or would you give some guidance on how to do it ?

spacewander: The grep pattern should be `request latency is nil`.

t/plugin/limit-conn2.t
```diff
@@ -178,3 +178,76 @@ qr/request latency is/
...
+--- grep_error_log_out
+request latency is
```
tzssangglass, line 252:
```suggestion
request latency is nil
```

spacewander (approved):

PR #4606, opened 2021-07-14
fix: correct the return value checking of pcall, since apisix_ngx_cli…

RocFang:
…ent has no chance to be nil or false

correct the return value checking of pcall

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/client-control.lua
+++ b/apisix/plugins/client-control.lua
@@ -16,7 +16,7 @@
 --
 local require = require
 local core = require("apisix.core")
-local _, apisix_ngx_client = pcall(require, "resty.apisix.client")
+local ok, apisix_ngx_client = pcall(require, "resty.apisix.client")
 local tonumber = tonumber
 
 
@@ -48,7 +48,7 @@ end
 
 
 function _M.rewrite(conf, ctx)
-    if not apisix_ngx_client then
+    if not ok then
         core.log.error("need to build APISIX-OpenResty to support client restriction")
         return 503
     end


RocFang:
emm.. I thank the ci failure has nothing to do with my pr:)

apisix/plugins/client-control.lua
```diff
@@ -48,7 +48,7 @@ end
 
 
 function _M.rewrite(conf, ctx)
-    if not apisix_ngx_client then
+    if not ok then
```
tokers, line 51: Is this better to show the error message while APISIX is starting?

RocFang: @tokers thx. i guess you mean throw an error when failed calling pcall, so this plugin will not be loaded. I agree with that  actually, but only try to make the patch as small as possbile.

spacewander: The module is loaded by default, and "resty.apisix.client" is only included in APISIX-OpenResty. So showing an error message will annoy people who use regular OpenResty.

spacewander (approved):

apisix/plugins/client-control.lua
tokers: OK, got it.

tokers: But from the HTTP semantic, maybe `501` is better in such a case (than `503`).

RocFang: @tokers maybe it's another issue? :)

tokers: Got it.

tokers (approved):

PR #4660, opened 2021-07-24
docs: fix the problem of how-to-build note block display

Applenice:
`:::note Note` is not displayed and affects reading. Insert a blockquote can be solved.

from: `:::note Note Some of the tests rely` to `NOTE: Some of the tests rely`.

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/getting-started.md
+++ b/docs/en/latest/getting-started.md
@@ -76,9 +76,7 @@ curl --location --request GET "http://httpbin.org/get?foo1=bar1&foo2=bar2"
 
 - We use the [curl](https://curl.se/docs/manpage.html) command for API testing. You can also use other tools such as [Postman](https://www.postman.com/) for testing.
 
-:::note Note
-If you already have Apache APISIX installed, please skip Step 1, and go to [Step 2](getting-started.md#step-2-create-a-route) directly.
-:::
+> **_NOTE:_** If you already have Apache APISIX installed, please skip Step 1, and go to [Step 2](getting-started.md#step-2-create-a-route) directly.
 
 ## Step 1: Install Apache APISIX
 
@@ -172,9 +170,7 @@ curl "http://127.0.0.1:9080/apisix/admin/upstreams/1" -H "X-API-KEY: edd1c9f0343
 
 We use `roundrobin` as the load balancing mechanism, and set `httpbin.org:80` as our upstream target (Upstream service) with an ID of `1`. For more information on the fields, see [Admin API](./admin-api.md).
 
-:::note Note
-Creating an Upstream service is not actually necessary, as we can use [Plugin](./architecture-design/plugin.md) to intercept the request and then respond directly. However, for the purposes of this guide, we assume that at least one Upstream service needs to be set up.
-:::
+> **_NOTE:_** Creating an Upstream service is not actually necessary, as we can use [Plugin](./architecture-design/plugin.md) to intercept the request and then respond directly. However, for the purposes of this guide, we assume that at least one Upstream service needs to be set up.
 
 ### Bind the Route to the Upstream
 
@@ -275,9 +271,7 @@ Apache APISIX provides a [Dashboard](https://github.com/apache/apisix-dashboard)
 
 ![Dashboard](../../assets/images/dashboard.jpeg)
 
-:::note Note
-APISIX Dashboard is an experimental feature for now.
-:::
+> **_NOTE:_** APISIX Dashboard is an experimental feature for now.
 
 ### Troubleshooting
 
--- a/docs/en/latest/how-to-build.md
+++ b/docs/en/latest/how-to-build.md
@@ -151,9 +151,7 @@ make help
 
   - Or you can specify the NGINX binary path by running this command: `TEST_NGINX_BINARY=/usr/local/bin/openresty prove -Itest-nginx/lib -r t`.
 
-  :::note Note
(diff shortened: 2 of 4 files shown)


tokers (approved):

docs/en/latest/getting-started.md
```diff
@@ -172,9 +170,7 @@ curl "http://127.0.0.1:9080/apisix/admin/upstreams/1" -H "X-API-KEY: edd1c9f0343
 
 We use `roundrobin` as the load balancing mechanism, and set `httpbin.org:80` as our upstream target (Upstream service) with an ID of `1`. For more information on the fields, see [Admin API](./admin-api.md).
 
-:::note Note
```
spacewander, line 175: The `:::note` has special meanings. CC @juzhiyuan

juzhiyuan: image

Hi, this is a mark from Docusarus framework, which is used on our website :) e.g https://apisix.apache.org/docs/apisix/getting-started

Applenice: So now how to deal with it？Reading on Github will be affected😑

juzhiyuan: Why not explore the docs from http://apisix.apache.org/docs/apisix/getting-started ?

juzhiyuan: Docs under the Apache APISIX repo is the data source for APISIX Website.

Applenice: emmm, prefer to look on Github, the code and documentation are in one place... So `:::note` will cause a bad reading experience. Of course you can turn off this PR if you decide you don't need to change it.

spacewander: What about adding comments for them? So the new comer will understand it.

juzhiyuan: How about adding some descriptions here ?https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#contribution-guidelines-for-documentation

spacewander: Few people read the guide. It would be better to add a comment next to the mark.

juzhiyuan: let's cc @liuxiran @yzeng25  What do you think?

yzeng25: IMO we should keep it as it is right now. 

Usually, contents behind or under `Note:` are something important or needed to be emphasized.

The origin `Note:` is somewhat ambiguous (it could be note, tips, warnings or even fatal warnings, depends on how you modify or define them). Besides, plain text 'Note' is easy to be ignored or skipped. The purpose of using `:::note` and `:::` to quote a note is to make it more noticeable on docs website. 

Yes, this is not convenient for GitHub readers. But why don't you checkout APISIX document website? This alternative is better.

spacewander: We need to add a comment to clarify them so that I don't need to explain it again and again.

yzeng25: OK.

Applenice: If so, there is a problem of uniformity of formatting in the document, and it feels like the usage of Note is not the same in many places.
https://github.com/apache/apisix/blob/6af1b69607bfb26bb6b61ec3b8307e2425e1f1c3/docs/en/latest/getting-started.md#L79

https://github.com/apache/apisix/blob/e3e018315311e0ca201f17579f9106eb88a834d1/docs/en/latest/plugin-develop.md#L90

https://github.com/apache/apisix/blob/baf843403461883c1334e63d15a6bb3622c31940/docs/en/latest/plugins/proxy-cache.md#L26

https://github.com/apache/apisix/blob/64923b3190384512a85f48c425b3ace0e5b32467/docs/en/latest/install-dependencies.md#L31

liuxiran: Agree to keep it as it is right now. We can guide the user through the documentation, and actually make contributions without looking documentation is not a recommended way.

juzhiyuan: APISIX's website serves Apache APISIX and other subprojects. Those markdown files are the source data for the website. If we admit it's a **product** just like Apache APISIX, then I would still prefer to keep the mark, so documentation readers would have a better experience than origin markdown files. I encourage people to use `http://apisix.apache.org/docs` instead of markdown files.

As for people ask why there has such a mark, what I can do is only adding that notice from Docusarus to CONTRIBUTING Guide, `Few people read the guide` doesn't mean this is not appropriate.

yzeng25: True, but all 'notes', as you figured out in comments above, should transform to `:::note Note` gradually.

Applenice: I can understand, later I go to the official website to read the documentation. 😊

yzeng25: @Applenice , thanks. 
If you find any potential typo in the future, please check out our documentation website `http://apisix.apache.org/docs` first. It is very likely to be a user experience enhancement on the website.

Applenice: I think the point you made might be better if it was noted prominently in the documentation. Some people may be used to reading it on Github, or they may be reading it offline after the project is downloaded, but after stating it more people will know that the reading experience on the official website is better than now, and they may avoid PR like this.

yzeng25: Yep, that's what happened in PR https://github.com/apache/apisix/pull/4681/

Applenice: Thanks to everyone😊

PR #4661, opened 2021-07-24
feat: set the basic id attribute when creating a resource via POST

okaybase:

When fetching resource list by the admin-api, some results include the basic id attribute (which created through the manager-api), but some results don't include it (which created through the admin-api). So set the basic id attribute when creating a resource by the admin-api in order to the result consistent.

image

### Pre-submission checklist:

### DIFF
--- a/apisix/core/etcd.lua
+++ b/apisix/core/etcd.lua
@@ -310,10 +310,10 @@ function _M.push(key, value, ttl)
     -- manually add suffix
     local index = res.body.header.revision
     index = string.format("%020d", index)
-    
+
     -- set the basic id attribute
     value.id = index
-    
+
     res, err = set(key .. "/" .. index, value, ttl)
     if not res then
         return nil, err
--- a/apisix/core/etcd.lua
+++ b/apisix/core/etcd.lua
@@ -310,7 +310,10 @@ function _M.push(key, value, ttl)
     -- manually add suffix
     local index = res.body.header.revision
     index = string.format("%020d", index)
-
+    
+    -- set the basic id property
+    value.id = index
+    
     res, err = set(key .. "/" .. index, value, ttl)
     if not res then
         return nil, err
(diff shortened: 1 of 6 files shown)


t/admin/routes2.t
```diff
@@ -197,6 +197,7 @@ GET /t
             res.node.key = nil
             res.node.value.create_time = nil
             res.node.value.update_time = nil
+            res.node.value.id = nil
```
spacewander, line 200: Can we check if id is not nil?

spacewander:
FYI @juzhiyuan @nic-chen

t/admin/routes2.t
okaybase: Thanks for your review firstly. the id (like the key) is generated by the server, and different for every test.

spacewander: We can just check if it is not nil

okaybase: okay, I will add the check, thanks!

apisix/core/etcd.lua
```diff
@@ -311,6 +311,9 @@ function _M.push(key, value, ttl)
     local index = res.body.header.revision
     index = string.format("%020d", index)
 
+    -- set the basic id attribute
+    value.id = index
```
nic-chen, line 315: There is a problem, the consumer does not have an id. Adding an id directly here will cause the consumer data to become unusable.
@spacewander @okaybase

okaybase: thanks for your review firstly. 
Consumer regards the username property as the identity, so only the HTTP PUT method is supported for creating a new consumer.

This change is for creating a resource via POST. @nic-chen

nic-chen: > thanks for your review firstly. 
> 
> Consumer regards the username property as the identity, so only the HTTP PUT method is supported for creating a new consumer.
> 
> 
> 
> This change is for creating a resource via POST. @nic-chen 

👍 thanks for the explanation.

t/admin/routes2.t
okaybase: the check is done. @spacewander

nic-chen (approved):

spacewander (approved):

Issue #4703, opened 2021-07-29
request help: failed to install luarocks on CentOS 7 due to permission problems


imjoey:
### Issue description

I was going to install APISIX on CentOS 7 and followed the installation guide at https://github.com/apache/apisix/blob/master/docs/en/latest/install-dependencies.md#centos-7. While I got an error as running the script `curl https://raw.githubusercontent.com/apache/apisix/master/utils/linux-install-luarocks.sh -sL | bash -`:

```
......
+ make build
+ msg='rerun this script with '\''sudo'\'' if you failed to make install because of privilege problem.'
+ make install
+ cat build.log
mkdir -p "/usr/local/bin"
install "build/luarocks" "/usr/local/bin/luarocks"
install: cannot create regular file '/usr/local/bin/luarocks': Permission denied
make: *** [GNUmakefile:117: /usr/local/bin/luarocks] Error 1
+ echo 'rerun this script with '\''sudo'\'' if you failed to make install because of privilege problem.'
rerun this script with 'sudo' if you failed to make install because of privilege problem.
+ exit 1
```

Although the error message is very clear, while this error would still break the installation progress, not so good experience for users. This issue probably affects other OS distributions.

So how about we add the `sudo` back to the script `make install > build.log 2>&1 || (cat build.log && echo "$msg" && exit 1)` at https://github.com/apache/apisix/blob/master/utils/linux-install-luarocks.sh#L43? This might make the overall installation more smooth. 

FYI, the `sudo` was removed at PR https://github.com/apache/apisix/pull/4167. Please correct me if I missed anything. Thanks.

### Environment

Bug report without environment information will be ignored or closed.

* apisix version (cmd: `apisix version`): `master`
* OS (cmd: `uname -a`): 
```
Linux localhost 3.10.0-1127.el7.x86_64 #1 SMP Tue Mar 31 23:36:51 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): 
```
nginx version: openresty/1.19.3.2
built by gcc 9.3.1 20200408 (Red Hat 9.3.1-2) (GCC)
built with OpenSSL 1.1.1k  25 Mar 2021
TLS SNI support enabled
[...]

PR #4708, opened 2021-07-29
chore: add sudo privilege for make install command at installing luarocks

imjoey:

For now, the package `luarocks` will be installed at `/usr/local` via `make install` command. If simply running the `curl https://raw.githubusercontent.com/apache/apisix/master/utils/linux-install-luarocks.sh -sL | bash -`, we will encounter an error due to privilege errors. For details, please refer to #4703. 

As discussed with @spacewander, we decide to add the `sudo` back via this PR. This would let users install APISIX without any interruptions.

With this patch merged in my environment, I could successfully install luarocks in CentOS 7 as both **non-root** and **root** users.

#4703 

### Pre-submission checklist:

### DIFF
--- a/utils/linux-install-luarocks.sh
+++ b/utils/linux-install-luarocks.sh
@@ -39,8 +39,7 @@ fi
     > build.log 2>&1 || (cat build.log && exit 1)
 
 make build > build.log 2>&1 || (cat build.log && exit 1)
-msg="rerun this script with 'sudo' if you failed to make install because of privilege problem."
-sudo make install > build.log 2>&1 || (cat build.log && echo "$msg" && exit 1)
+sudo make install > build.log 2>&1 || (cat build.log && exit 1)
 cd .. || exit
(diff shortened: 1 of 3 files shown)


utils/linux-install-luarocks.sh
```diff
@@ -40,7 +40,7 @@ fi
 
 make build > build.log 2>&1 || (cat build.log && exit 1)
 msg="rerun this script with 'sudo' if you failed to make install because of privilege problem."
```
spacewander, line 42: We can remove this prompt and other sudo in the ci script or docs now?

imjoey: I think yes if everything is as expected. We could safely remove this prompt if it's only for `make install`. What do you think?

spacewander: Yes. We think about the same thing.

spacewander (approved):

tokers (approved):

Issue #4674, opened 2021-07-26
Proposal: Addition of an authorization plugin


rushitote:
## Background

Hi, we have been developing `casbin-authz` plugin for APISIX based on [Lua Casbin](https://github.com/casbin/lua-casbin/) which is the Lua implementation of the [Casbin](https://casbin.org/) library. Casbin is an authorization library which supports access control models like ACL, RBAC and ABAC. `casbin-authz` is a plugin for APISIX that enables authorization based on Casbin. The initial implementation is at [apisix-authz](https://github.com/casbin-lua/apisix-authz).

## Implementation and Usage 

This is what we have developed till now:

```lua
local Enforcer = require("casbin")
local core = require("apisix.core")
local get_headers = ngx.req.get_headers

local CasbinEnforcer

local plugin_name = "apisix-authz"

local schema = {
    type = "object",
    properties = {
        model_path = { type = "string" },
        policy_path = { type = "string" },
        username = { type = "string"}
    },
    required = {"model_path", "policy_path", "username"},
    additionalProperties = false
}

local _M = {
    version = 0.1,
    priority = 2560,
    type = 'auth',
    name = plugin_name,
    schema = schema
}

function _M.rewrite(conf)
    -- creates an enforcer when request sent for the first time
    if not CasbinEnforcer then
        CasbinEnforcer = Enforcer:new(conf.model_path, conf.policy_path)
    end

    local path = ngx.var.request_uri
    local method = ngx.var.request_method
    local username = get_headers()[conf.username]
    if not username then 
        username = "anonymous" 
    end

    if path and method and username then
        if not CasbinEnforcer:enforce(username, path, method) then
            return 403, {message = "Access Denied"}
        end
    else
        return 403, {message = "Access Denied"}
    end
end

local function addPolicy()
    local headers = get_headers()
    local type = headers["type"]

    if type == "p" then
        local subject = headers["subject"]
        local object = headers["object"]
        local action = headers["action"]

        if not subject or not object or not action then
            return 400, {message = "Invalid policy request."}
        end

        if CasbinEnforcer:AddPolicy(subject, object, action) then
            return 200, {message = "Successfully added policy."}
        else
            return 400, {message = "Invalid policy request."}
        end
    elseif type == "g" then
        local user = headers["user"]
        local role = headers["role"]

        if not user or not role then
            return 400, {message = "Invalid policy request."}
        end

        if CasbinEnforcer:AddGroupingPolicy(user, role) then
[...]

PR #4710, opened 2021-07-29
feat: Added authz-casbin plugin and doc and tests for it

rushitote:

fix: #4674 

This PR adds an authorization plugin: `authz-casbin` to APISIX. This is based on Lua Casbin which is the Lua implementation of Casbin library. The plugin supports enforcement of powerful authorization scenarios based on various access control models supported by Casbin.

The plugin works on the basis of a model file and a policy file. The model acts as a configuration for the policies and policy enforcement. The plugin currently also supports direct model/policy text in absence of files through plugin metadata.

An example of authz model is:

```conf
[request_definition]
r = sub, obj, act
[policy_definition]
p = sub, obj, act
[role_definition]
g = _, _
[policy_effect]
e = some(where (p.eft == allow))
[matchers]
m = (g(r.sub, p.sub) || keyMatch(r.sub, p.sub)) && keyMatch(r.obj, p.obj) && keyMatch(r.act, p.act)
```

And example of authz policy is:

```
p, *, /, GET
p, admin, *, *
g, alice, admin
```

This means that any user (subject) can access the homepage which is at `/` via `GET` request but only admins can a access other pages and use other request methods. And here, as defined in policy `alice` has a role as `admin` and hence she has admin access.

### Pre-submission checklist:


apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,347 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+local casbin          = require("casbin")
+local core            = require("apisix.core")
+local plugin          = require("apisix.plugin")
+local plugin_metadata = require("apisix.admin.plugin_metadata")
```
spacewander, line 21: The plugin_metadata should be only set on the CP side. It's forbidden to let configuration flow from the DP to the CP.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,347 @@
...
+            return 400
+        end
+    else
+        return 400, {message = "Invalid policy type."}
+    end
+end
+
+
+function _M.api()
+    return {
```
spacewander, line 317: Since we already use metadata to store the configuration, there is no need to provide APIs on the DP side. Everything should be done with the plugin metadata's API on the CP side.

rushitote: @spacewander 
Thanks for your review, so we will remove all the API functions and if the model/policy text in plugin metadata is updated (from the CP side), we will update the `CasbinEnforcer` through that model/policy too. Is this correct?

spacewander: Yes, you can use `metadata.modifiedIndex` to ensure casbin_enforcer is created with the latest model/policy, like this one:
https://github.com/apache/apisix/blob/fd8f875429bd43462430bd662788940f6a579ff9/apisix/http/router/radixtree_uri.lua#L31-L36

spacewander: And I think you can submit a minimum viable pull request so that we can check & merge it early.

rushitote: Sure I will do that.

rushitote: I added a condition that if the schema changes, the `casbin_enforcer` will be recreated.

.github/workflows/build.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           tar zxvf ${{ steps.branch_env.outputs.fullname }}
 
       - name: Linux Get dependencies
-        run: sudo apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl
+        run: sudo apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev
```
Yiyiyimu, line 92: Hi @rushitote I think you need to also install pcre lib in chaos testing ci to make it pass, see [error log](https://github.com/apache/apisix/pull/4710/checks?check_run_id=3207414954#step:4:932). 
You could add it before building docker image [here](https://github.com/apache/apisix/blob/master/.github/workflows/chaos.yml#L32)

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    return e
+end
+
+
+function _M.rewrite(conf)
+    -- creates an enforcer when request sent for the first time
+    if not casbin_enforcer then
+        casbin_enforcer = new_enforcer(conf.model_path, conf.policy_path)
```
spacewander, line 105: Multiple routes will use the same casbin enforcer? I think we should store different casbin via lrucache, like this:
https://github.com/apache/apisix/blob/9db2dd2399c3970df1e1b7fa0f8b7dcd92d26b1c/apisix/plugins/error-log-logger.lua#L144

.github/workflows/build.yml
rushitote: @Yiyiyimu Thanks for this, I have made the change.

apisix/plugins/authz-casbin.lua
rushitote: Yes, you are right. I have changed it now.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,119 @@
...
+
+    local metadata = plugin.plugin_metadata(plugin_name)
+    local casbin_enforcer = lrucache(plugin_name, metadata.modifiedIndex, new_enforcer, conf)
+
+    local path = ngx.var.request_uri
+    local method = ngx.var.request_method
+    local username = get_headers()[conf.username]
+    if not username then username = "anonymous" end
+
+    if path and method and username then
```
tokers, line 109: We don't need to check the existence for these three variables if we use them in the HTTP sub-system.

docs/en/latest/plugins/authz-casbin.md
```diff
@@ -0,0 +1,205 @@
...
+        },
+        "type": "roundrobin"
+    },
+    "uri": "/*"
+}'
+```
+
+This will create a Casbin enforcer from the model and policy files at your first request.
+
+### By using model/policy text
```
tokers, line 85: Please specify the precedence between this one and the path way.

apisix/plugins/authz-casbin.lua
rushitote: Yes, right - this is redundant here.

docs/en/latest/plugins/authz-casbin.md
rushitote: Sure, I will make it more clear there.

apisix/plugins/authz-casbin.lua
spacewander: Sorry, I rethought about it afternoon, and found it isn't a good idea.
It is more suitable to bind the casbin with the conf like:
https://github.com/apache/apisix/blob/38561dc0cd13ee412d05a96153dee80018f1e799/apisix/plugins/uri-blocker.lua#L83

We also need to ensure the global casbin is update-to-date with the metadata via https://github.com/apache/apisix/pull/4710#discussion_r679712809

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,119 @@
...
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    local ok, err = core.schema.check(schema, conf)
+    if ok then
+        return true
+    else
+        local metadata = plugin.plugin_metadata(plugin_name)
+        if metadata and metadata.value.model and metadata.value.policy and conf.username then
```
spacewander, line 67:
```suggestion
        if metadata and metadata.value and conf.username then
```

Since we have checked the metadata schema.

t/plugin/authz-casbin.t
```diff
@@ -0,0 +1,358 @@
...
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 4: Enforcer from text without files
```
spacewander, line 121: Test 4 already contains Test 3?

t/plugin/authz-casbin.t
```diff
@@ -0,0 +1,358 @@
...
+user: alice
+--- error_code: 200
+--- response_body
+hello world
+--- no_error_log
+[error]
+
+
+
+=== TEST 10: unauthorized user before policy update
```
spacewander, line 259: Need a similar test for the policy_path configuration.

t/plugin/authz-casbin.t
rushitote: Yes, removed that unnecessary part from test 4.

t/plugin/authz-casbin.t
rushitote: Added the tests for enforcer from files.

apisix/plugins/authz-casbin.lua
rushitote: @spacewander Sorry, I found this now. But is there any reason for us to not use lrucache? It worked well for two different routes with different casbin enforcers when I tested it.

.github/workflows/build.yml
Yiyiyimu: Sorry I made a mistake 😿 We need to add the dependency to the dockerfile but not on the host machine. Do you mind file a PR about the change on the dockerfile: https://github.com/apache/apisix-docker/blob/master/alpine-local/Dockerfile, which is used as building apisix image used in chaos testing

Also we need to change docs about dependencies installation to inform users to install them: https://github.com/apache/apisix/blob/master/docs/en/latest/install-dependencies.md

t/plugin/authz-casbin/policy.csv
```diff
@@ -0,0 +1,3 @@
+p, *, /, GET
+p, admin, *, *
+g, alice, admin
```
tokers, line 3: EOL symbol is required. Please check out the editor configuration.

.github/workflows/build.yml
rushitote: @Yiyiyimu Thanks again, I have created a PR for that at [apisix-docker#202](https://github.com/apache/apisix-docker/pull/202) and added PCRE in the dependency installation doc.

t/plugin/authz-casbin/policy.csv
rushitote: Added the EOL symbol in model/policy files.

apisix/plugins/authz-casbin.lua
rushitote: I have removed `lrucache` and replaced it by binding the `casbin_enforcer` to conf. Also we will use the `modifiedIndex` to keep the casbin from metadata updated.

spacewander: > 
> 
> @spacewander Sorry, I found this now. But is there any reason for us to not use lrucache? It worked well for two different routes with different casbin enforcers when I tested it.

It works fine, but makes the life cycle complex.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,118 @@
...
+    local policy_path = conf.policy_path
+
+    local e
+
+    if model_path and policy_path then
+        e = casbin:new(model_path, policy_path)
+        conf.type = "file"
+    end
+
+    local metadata = plugin.plugin_metadata(plugin_name)
```
spacewander, line 86: IMHO, it surprises me that the enforcer from metadata can override the one from the route. Normally, the route level configuration can override the global one.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,118 @@
...
+    if metadata and metadata.value.model and metadata.value.policy and not e then
+        local model = metadata.value.model
+        local policy = metadata.value.policy
+        e = casbin:newEnforcerFromText(model, policy)
+        conf.type = "metadata"
```
spacewander, line 91: We don't need to bind the enforcer from metadata with the route configuration as it is global. Recreating the global enforcer when the request hits another route is not a good idea.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,118 @@
...
+        conf.modifiedIndex = modifiedIndex
+    end
+
+    conf.casbin_enforcer = e
+end
+
+
+function _M.rewrite(conf)
```
spacewander, line 99:
```suggestion
function _M.rewrite(conf, ctx)
```

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,118 @@
...
+    -- creates an enforcer when request sent for the first time
+    local metadata = plugin.plugin_metadata(plugin_name)
+    if (not conf.casbin_enforcer) or
+    (conf.type == "metadata" and conf.modifiedIndex ~= metadata.modifiedIndex) then
+        new_enforcer(conf, metadata.modifiedIndex)
+    end
+
+    local path = ngx.var.request_uri
+    local method = ngx.var.request_method
```
spacewander, line 108:
```suggestion
    local method = ctx.var.method
```

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,118 @@
...
+
+function _M.rewrite(conf)
+    -- creates an enforcer when request sent for the first time
+    local metadata = plugin.plugin_metadata(plugin_name)
+    if (not conf.casbin_enforcer) or
+    (conf.type == "metadata" and conf.modifiedIndex ~= metadata.modifiedIndex) then
+        new_enforcer(conf, metadata.modifiedIndex)
+    end
+
+    local path = ngx.var.request_uri
```
spacewander, line 107:
```suggestion
    local path = ctx.var.uri
```

It is more suitable to use the normalized one? See http://nginx.org/en/docs/http/ngx_http_core_module.html#var_uri

spacewander:
To make the misc checker pass, you need to update https://github.com/apache/apisix/blob/9f01ef8f2ca753557289ac9dc91f5212432fae73/ci/ASF-Release.cfg#L59
To make the doc linter pass, you need to update https://github.com/apache/apisix/blob/master/docs/en/latest/config.json

apisix/plugins/authz-casbin.lua
rushitote: @spacewander so, should we create a new `casbin_enforcer` local variable specifically for global config? Then we can also have (if any) route enforcer override the global one.

rushitote:
> To make the misc checker pass, you need to update
> 
> https://github.com/apache/apisix/blob/9f01ef8f2ca753557289ac9dc91f5212432fae73/ci/ASF-Release.cfg#L59
> 
> 
> To make the doc linter pass, you need to update https://github.com/apache/apisix/blob/master/docs/en/latest/config.json

Thanks, changed it.

apisix/plugins/authz-casbin.lua
spacewander: > 
> 
> @spacewander so, should we create a new `casbin_enforcer` local variable specifically for global config? Then we can also have (if any) route enforcer override the global one.

Do you mean a new global variable?

rushitote: > > @spacewander so, should we create a new `casbin_enforcer` local variable specifically for global config? Then we can also have (if any) route enforcer override the global one.
> 
> Do you mean a new global variable?

local, similar as that of the first commit ([here](https://github.com/apache/apisix/pull/4710/commits/cb37e5fdce3df9cd42f25fa1502a5001aa25a95c#diff-b18fc191c270634f4d1d599ad10a5dbb1d646ec9ff4c9f9e6224195e45a96a8bR25))

rushitote: @spacewander I was thinking of something like this:
```lua
local casbin_enforcer

local function new_enforcer(conf, modifiedIndex)
    local model_path = conf.model_path
    local policy_path = conf.policy_path

    if model_path and policy_path then
        conf.type = "file"
        conf.casbin_enforcer = casbin:new(model_path, policy_path)
        return
    end

    local metadata = plugin.plugin_metadata(plugin_name)
    if metadata and metadata.value.model and metadata.value.policy then
        conf.type = "metadata"
        if not casbin_enforcer or casbin_enforcer.modifiedIndex ~= modifiedIndex then
            local model = metadata.value.model
            local policy = metadata.value.policy
            casbin_enforcer = casbin:newEnforcerFromText(model, policy)
            casbin_enforcer.modifiedIndex = modifiedIndex
        end
    end
end


    -- creates an enforcer when request sent for the first time
    local metadata = plugin.plugin_metadata(plugin_name)

    if (not conf.casbin_enforcer and conf.type ~= "metadata") or
    (conf.type == "metadata" and casbin_enforcer.modifiedIndex ~= metadata.modifiedIndex) then
        new_enforcer(conf, metadata.modifiedIndex)
    end

    local path = ctx.var.uri
    local method = ctx.var.method
    local username = get_headers()[conf.username]
    if not username then username = "anonymous" end

    if conf.casbin_enforcer then
        if not conf.casbin_enforcer:enforce(username, path, method) then
            return 403, {message = "Access Denied"}
        end
    else
        if not casbin_enforcer:enforce(username, path, method) then
            return 403, {message = "Access Denied"}
        end
    end
end
```

So, we divide routes into two types - either `file` or `metadata`. The `file` one has their own enforcer binded to the conf while all the `metadata` based routes use the same `casbin_enforcer`. In case of the metadata based route, enforcer is created only when either there isn't one or when the `modifiedIndex` has changed. And since this separates any route into one of the two types, one type will not override the other. Will this solve the issue?

spacewander: LGTM. The only issue is that both new_enforcer and rewrite check if we need to create a new enforcer. I think we can just leave the check in one method.

rushitote: @spacewander Can you please tell me which check we could remove? I feel all are necessary and don't see any similar checks.

spacewander: What about:
```
local casbin_enforcer

local function new_enforcer_if_need(conf)
    local model_path = conf.model_path
    local policy_path = conf.policy_path

    if model_path and policy_path then
        if not conf.casbin_enforcer then
            conf.casbin_enforcer = casbin:new(model_path, policy_path)
        end
        return true
    end

    local metadata = plugin.plugin_metadata(plugin_name)
    if not (metadata and metadata.value.model and metadata.value.policy) then
        return nil, "not enough configuration to create enforcer"
    end

    local modifiedIndex = metadata.modifiedIndex
    if not casbin_enforcer or casbin_enforcer.modifiedIndex ~= modifiedIndex then
        local model = metadata.value.model
        local policy = metadata.value.policy
        casbin_enforcer = casbin:newEnforcerFromText(model, policy)
        casbin_enforcer.modifiedIndex = modifiedIndex
    end
    return true
end


    -- creates an enforcer when request sent for the first time
    local ok, err = new_enforcer_if_need(conf)
    if not ok then
        return 503, {message = err}
    end

    local path = ctx.var.uri
    local method = ctx.var.method
    local username = get_headers()[conf.username]
    if not username then username = "anonymous" end

    if conf.casbin_enforcer then
        if not conf.casbin_enforcer:enforce(username, path, method) then
            return 403, {message = "Access Denied"}
        end
    else
        if not casbin_enforcer:enforce(username, path, method) then
            return 403, {message = "Access Denied"}
        end
    end
end
```

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,115 @@
...
+    -- creates an enforcer when request sent for the first time
+    local metadata = plugin.plugin_metadata(plugin_name)
+    if (not conf.casbin_enforcer) or
+    (conf.type == "metadata" and conf.modifiedIndex ~= metadata.modifiedIndex) then
+        new_enforcer(conf, metadata.modifiedIndex)
+    end
+
+    local path = ctx.var.uri
+    local method = ctx.var.method
+    local username = get_headers()[conf.username]
```
tzssangglass, line 106:
```suggestion
    local username = get_headers()[conf.username] or "anonymous"
```

apisix/plugins/authz-casbin.lua
rushitote: I think this is good, I have added a commit with this.

apisix/plugins/authz-casbin.lua
rushitote: Thanks, added it.

apisix/plugins/authz-casbin.lua
```diff
@@ -0,0 +1,124 @@
...
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    local ok, err = core.schema.check(schema, conf)
+    if ok then
+        return true
+    else
+        local metadata = plugin.plugin_metadata(plugin_name)
+        if metadata and metadata.value and conf.username then
```
tzssangglass, line 64: if `username` is always required, you can use `required = {"username"}` in `schema`, such as https://github.com/apache/apisix/blob/1b247c8e1ebd324efe66abbe33d5647a93f0426e/apisix/plugins/example-plugin.lua#L22-L32

t/plugin/authz-casbin.t
```diff
@@ -0,0 +1,401 @@
...
+                            "username" : "user"
+                        }
+                    },
+                    "upstream": {
+                        "nodes": {
+                            "127.0.0.1:1982": 1
+                        },
+                        "type": "roundrobin"
+                    },
+                    "uri": "/*"
```
tzssangglass, line 167:
```suggestion
                    "uri": "/hello"
```

apisix/plugins/authz-casbin.lua
rushitote: It is already there ([here](https://github.com/apache/apisix/pull/4710/files#diff-b18fc191c270634f4d1d599ad10a5dbb1d646ec9ff4c9f9e6224195e45a96a8bR33)). This was meant to check if the route will use the plugin metadata (instead of files) and hence it requires separate schema check.

tzssangglass: got

t/plugin/authz-casbin.t
```diff
@@ -0,0 +1,401 @@
...
```
tzssangglass, line 326:
```suggestion
                    "uri": "/hello"
```

t/plugin/authz-casbin.t
```diff
@@ -0,0 +1,401 @@
...
+                ngx.HTTP_PUT,
+                [[{
+                    "plugins": {},
+                    "upstream": {
+                        "nodes": {
+                            "127.0.0.1:1982": 1
+                        },
+                        "type": "roundrobin"
+                    },
+                    "uri": "/*"
```
tzssangglass, line 386: ditto

rushitote: Changed it.

docs/en/latest/plugins/authz-casbin.md
```diff
@@ -0,0 +1,207 @@
...
+
+| Name        | Type   | Requirement | Default | Valid | Description                                                  |
+| ----------- | ------ | ----------- | ------- | ----- | ------------------------------------------------------------ |
+| model_path  | string | required    |         |       | The path of the Casbin model configuration file.             |
+| policy_path | string | required    |         |       | The path of the Casbin policy file.                          |
+| username    | string | required    |         |       | The header you will be using in request to pass the username (subject). |
+
+**NOTE**: You must either specify both the `model_path` and `policy_path` in plugin config or specify both the `model` and `policy` in the plugin metadata for the schema to be valid.
+
+## Metadata
```
tokers, line 50: I thought for a while and was confused by the intention of the metadata, is it used as the default policy? And why not also support using text format in the plugin configuration?

rushitote: @tokers I think it is meant to be kind of global configuration for all routes. Should I add the support for using model/policy text from plugin conf?

tokers: @spacewander What's your opinion? If the policy in plugin metadata is used to be the default one, we'd better to keep the configuration items consistent.

spacewander: @rushitote @tokers 
Yes, let's do it.

spacewander (approved):

tzssangglass (approved):

apisix/plugins/authz-casbin.lua
```diff
@@ -28,9 +28,14 @@ local schema = {
     properties = {
         model_path = { type = "string" },
         policy_path = { type = "string" },
+        model = { type = "string" },
+        policy = { type = "string" },
         username = { type = "string"}
     },
-    required = {"model_path", "policy_path", "username"},
+    anyOf = {
```
tokers, line 35: It should be `oneOf`.

rushitote: Changed it.

docs/en/latest/plugins/authz-casbin.md
```diff
@@ -0,0 +1,250 @@
...
+        "nodes": {
+            "127.0.0.1:1980": 1
+        },
+        "type": "roundrobin"
+    },
+    "uri": "/*"
+}'
+```
+
+**NOTE**: The model/policy file paths have a higher precedence, hence if the model/policy file paths are present in the configuration the plugin will use the files accordingly.
```
tokers, line 176: It seems that now we should claim the precedence between route config and metadata config, rather than the file and text.

rushitote: @tokers So, should it be worded like this:

'The plugin route configuration has a higher precedence than the plugin metadata configuration. Hence if the model/policy configuration is present in the plugin route config, the plugin will use that instead of the metadata config.'

tokers: That's good enough 👍.

rushitote: Replaced it.

tokers (approved):

PR #4718, opened 2021-07-30
Add apsix.ssl.loose_sni configuration in config-default.yaml

TkClark:
Due to Alicloud WAF does not support SNI.
When ApiSix is the backend of Aliyun WAF and WAF is set to HTTPS, each HTTPS request returns status 502.
If set loose_sni with true, when the HTTPS request doesn't hava servername, will falling back to fake certificate. It can solve the problem that the AliCloud WAF HTTPS back source cannot be accessed correctly.


tzssangglass:
Before merge PR, there are some issues that need to be discussed.

1. if `loose_sni` is set to true, it means that all SNI-related features of APISIX are not available？How to solve this problem?

I think the core problem here is: to use the SNI features in APISIX but without the SNI(by use Alicloud WAF).

2. I am concerned that it will cause some security issues, for example, the client can bypass the TLS handshake, etc. I'm not sure this will ever happen. Just worried.

Please correct me if I am wrong.

tzssangglass:
We can use `fallback_certificate` and `fallback_certificate_key` for this case: if the `certificate` and `certificate_key` are not found because there is no SNI, we can use `fallback_certificate` and `fallback_certificate_key`.

TkClark:
> Before merge PR, there are some issues that need to be discussed.
> 
> 1. if `loose_sni` is set to true, it means that all SNI-related features of APISIX are not available？How to solve this problem?
> 
> I think the core problem here is: to use the SNI features in APISIX but without the SNI(by use Alicloud WAF).
> 
> 1. I am concerned that it will cause some security issues, for example, the client can bypass the TLS handshake, etc. I'm not sure this will ever happen. Just worried.
> 
> Please correct me if I am wrong.

1.

PR #4734, opened 2021-08-02
fix(ext-plugin): avoid using 0 as the default http status code

tzssangglass:

If the plugin-runner sends a stop request but does not set an Http Status Code, APISIX will use 0 as the default status code, causing the current request to run later phases.

see: https://github.com/apache/apisix-java-plugin-runner/issues/55
### Pre-submission checklist:

### DIFF
--- a/t/lib/ext-plugin.lua
+++ b/t/lib/ext-plugin.lua
@@ -185,7 +185,9 @@ function _M.go(case)
             local vec = builder:EndVector(len)
 
             http_req_call_stop.Start(builder)
-            http_req_call_stop.AddStatus(builder, 405)
+            if case.stop_no_code ~= true then
+                http_req_call_stop.AddStatus(builder, 405)
+            end
             http_req_call_stop.AddBody(builder, b)
             http_req_call_stop.AddHeaders(builder, vec)
             local action = http_req_call_stop.End(builder)
--- a/t/lib/ext-plugin.lua
+++ b/t/lib/ext-plugin.lua
@@ -185,7 +185,7 @@ function _M.go(case)
             local vec = builder:EndVector(len)
 
             http_req_call_stop.Start(builder)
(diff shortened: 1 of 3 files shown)


t/plugin/ext-plugin/http-req-call.t
```diff
@@ -124,7 +124,28 @@ X-Req: bar
 
 
 
-=== TEST 3: check input
+=== TEST 3: stop without setting status code
```
spacewander, line 127: Please add new test at the end of file

tzssangglass: update

t/plugin/ext-plugin/http-req-call.t
```diff
@@ -517,3 +517,24 @@ GET /hello
...
+--- extra_stream_config
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock;
+
+        content_by_lua_block {
+            local ext = require("lib.ext-plugin")
+            ext.go({stop = true, stop_no_code = true})
+        }
+    }
+--- error_code: 200
```
spacewander, line 537: We already check if error_code == 200 by default

t/lib/ext-plugin.lua
```diff
@@ -185,7 +185,9 @@ function _M.go(case)
             local vec = builder:EndVector(len)
 
             http_req_call_stop.Start(builder)
-            http_req_call_stop.AddStatus(builder, 405)
+            if case.stop_no_code ~= true then
```
spacewander, line 188:
```suggestion
            if case.check_default_status ~= true then
```

tzssangglass: update

spacewander (approved):

tokers (approved):

PR #4774, opened 2021-08-07
feat(plugin): add degradation switch for traffic control plugin

okaybase:

Add degradation switch for traffic control plugin.
When the limit function is temporarily unavailable(e.g. redis timeout for limit-count plugin), we can auto allow requests to continue exec by enable degradation switch, this is important for some scenarios(hope the request continue exec where upstream services can handle it).

### Pre-submission checklist:


apisix/plugins/limit-conn/init.lua
```diff
@@ -41,7 +41,11 @@ function _M.increase(conf, ctx)
     local lim, err = lrucache(conf, nil, create_limit_obj, conf)
     if not lim then
         core.log.error("failed to instantiate a resty.limit.conn object: ", err)
-        return 500
+        if conf.degradation then
+            return
+        else
+            return 500
+        end
```
nic-chen, line 48:
```suggestion
        if conf.degradation then
            return
        end
        return 500
```

nic-chen (approved):
LGTM

okaybase:
> LGTM

Thanks for your review~ @nic-chen

tokers (approved):

apisix/plugins/limit-conn.lua
```diff
@@ -33,6 +33,7 @@ local schema = {
         rejected_code = {
             type = "integer", minimum = 200, maximum = 599, default = 503
         },
+        degradation = {type = "boolean", default = false}
```
spacewander, line 36: Would be better to use "allow_degradation"?

t/plugin/limit-count-redis-cluster.t
```diff
@@ -377,3 +377,62 @@ code: 503
...
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 11: enable degradation switch for TEST 10
+--- request
+GET /hello
+--- error_code eval
```
spacewander, line 435: We don't need to check 200 code - it is checked by default

spacewander (commented):
Need a test in t/plugin/limit-conn.t

apisix/plugins/limit-conn.lua
okaybase: done

t/plugin/limit-count-redis-cluster.t
okaybase: done

spacewander (commented):
LGTM. Let's add new test in t/plugin/limit-conn.t

okaybase:
> LGTM. Let's add new test in t/plugin/limit-conn.t

new test added, and improve a log info. @spacewander

spacewander (approved):

PR #4782, opened 2021-08-09
feat(plugin): add limit quota header switch for limit-count

okaybase:

Add limit quota header switch for limit-count plugin.
In some scenarios, such as some crawlers, service provider don't want to show limit quota info to the response header.

### Pre-submission checklist:


tzssangglass:
why we need this? Is there a related issue?

okaybase:
> why we need this? Is there a related issue?

In some scenarios, such as some crawlers, service provider don't want to show limit quota info to the response header.

apisix/plugins/limit-count.lua
```diff
@@ -184,8 +185,10 @@ function _M.access(conf, ctx)
         return 500, {error_msg = "failed to limit count: " .. err}
     end
 
-    core.response.set_header("X-RateLimit-Limit", conf.count,
-                             "X-RateLimit-Remaining", remaining)
+    if conf.show_limit_quota_header == nil or conf.show_limit_quota_header then
```
spacewander, line 188:
```suggestion
    if conf.show_limit_quota_header then
```

The default value will be set in `check_schema` even the configuration in etcd doesn't have it.

okaybase: okay, done~

spacewander (approved):

tzssangglass (approved):

tokers (approved):

PR #4807, opened 2021-08-11
chore(limit-req): remove duplicate code

### LINKED ISSUE #2448: chore: `limit-req` plugin code optimization
### Issue description

This part of the code ( https://github.com/apache/apisix/blob/master/apisix/plugins/limit-req.lua#L70-L80)
 can be replaced with "`local key = (ctx.var[conf.key] or "") .. ctx.conf_type .. ctx.conf_version`" .

codewangg:
What this PR does / why we need it:
fix #2448


spacewander:
Look like we need to update some test to verify the new behavior.

Issue #4776, opened 2021-08-08
bug: plugins:proxy_cache, GET ,apisix-cache-status return,but POST not.


zhangdadan:
apisix:v2.8
plugins:proxy_cache
my conf: 
`
"plugins": {
    "proxy-cache": {
      "cache_method": [
        "POST",
        "GET"
      ],
      "disable": false,
      "hide_cache_headers": true
    }
  },
`
I print ngx.var find that GET `upstream_cache_status` has value,but POST not at file `/usr/local/apisix/apisix/plugins/proxy-cache.lua`

What's wrong?

tokers:
@zhangdadan Thank for your report, it seems there is a bug in the `proxy-cache` plugin, the `cache_method` setting doesn't affect the underlying `proxy_cache_methods` directive.

zhangdadan:
![image](https://user-images.githubusercontent.com/18242914/128651927-799eb107-b183-4214-a1eb-62992071bd8d.png)

I find `proxy_cache_methods`  is not dynamic at `/usr/local/apisix/conf/nginx.conf`,can it be configured?

tokers:
> > @zhangdadan Thank for your report, it seems there is a bug in the `proxy-cache` plugin, the `cache_method` setting doesn't affect the underlying `proxy_cache_methods` directive.
> 
> ![image](https://user-images.githubusercontent.com/18242914/128651927-799eb107-b183-4214-a1eb-62992071bd8d.png)
> 
> I find `proxy_cache_methods` is not dynamic at `/usr/local/apisix/conf/nginx.conf`,can it be configured?

Currently it's not dynamic, and cannot be extended by the plugin configuration.

zhangdadan:
> > > @zhangdadan Thank for your report, it seems there is a bug in the `proxy-cache` plugin, the `cache_method` setting doesn't affect the underlying `proxy_cache_methods` directive.
> > 
> > 
> > ![image](https://user-images.githubusercontent.com/18242914/128651927-799eb107-b183-4214-a1eb-62992071bd8d.png)
> > I find `proxy_cache_methods` is not dynamic at `/usr/local/apisix/conf/nginx.conf`,can it be configured?
> 
> Currently it's not dynamic, and cannot be extended by the plugin configuration.

But，I has configured by myself and add `POST` value on `proxy_cache_methods ` at  the `/usr/local/apisix/conf/nginx.conf`,then run `apisix restart`, but it's not useful.

tzssangglass:
[...]

PR #4814, opened 2021-08-12
fix(proxy-cache): keep cache_method same with nginx's proxy_cache_methods

agile6v:

Nginx's [proxy_cache_methods](https://github.com/nginx/nginx/blob/926e0aa70acd58d2a13451f40fc19aeb5393aa7e/src/http/ngx_http_upstream.c#L456
) directive only allows to specify `GET`、`POST` and `HEAD`, so we keep same with Nginx.

#4776 

### Pre-submission checklist:


spacewander (approved):

apisix/cli/ngx_tpl.lua
```diff
@@ -596,7 +596,7 @@ http {
             proxy_cache                         $upstream_cache_zone;
             proxy_cache_valid                   any {% if proxy_cache.cache_ttl then %} {* proxy_cache.cache_ttl *} {% else %} 10s {% end %};
             proxy_cache_min_uses                1;
-            proxy_cache_methods                 GET HEAD;
+            proxy_cache_methods                 GET HEAD POST;
```
tokers, line 599: Nginx by default doesn't enable the cache for POST. Why change this behavior here?

agile6v: It doesn't affect the default behavior of APISIX, which means that user can specify the `POST` method on APISIX.

tzssangglass: Is it necessary to cache a POST request?

The only I can think of for caching POST requests is that some internal specification forces the consistent use of POST requests……

tokers: > It doesn't affect the default behavior of APISIX, which means that user can specify the `POST` method on APISIX.

Here you change the Nginx template, without a configurable setting. This is a change for default havior.

agile6v: > @agile6v
> Could you sync it to APISIX.pm?

done

agile6v: > Is it necessary to cache a POST request?
> 
> The only I can think of for caching POST requests is that some internal specification forces the consistent use of POST requests……

Yes. This is a very rare scenario. The users don't enable the method on APISIX,  it doesn't take effect.  If users enable it, they need to know what they are doing.

agile6v: > > It doesn't affect the default behavior of APISIX, which means that user can specify the `POST` method on APISIX.
> 
> Here you change the Nginx template, without a configurable setting. This is a change for default havior.

Whether to enable is [here](https://github.com/agile6v/apisix/blob/master/docs/en/latest/plugins/proxy-cache.md) (cache_method)

tzssangglass (approved):

PR #4817, opened 2021-08-13
feat(gzip): support special * to match any type

RocFang:

right now, type of types attr is array, and we dont't have a method to support the special `gzip_types *` config in Nginx Core. So maybe we can change type of types to support single string, and support * too.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/gzip.lua
+++ b/apisix/plugins/gzip.lua
@@ -38,8 +38,7 @@ local schema = {
                     },
                 },
                 {
-                    type = "string",
-                    minLength = 1,
+                    enum = {"*"}
                 }
             },
             default = {"text/html"}
@@ -127,9 +126,7 @@ function _M.header_filter(conf, ctx)
             end
         end
     else
-        if content_type == types or types == "*" then
-            matched = true
-        end
+        matched = true
     end
     if not matched then
         return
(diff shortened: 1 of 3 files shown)


apisix/plugins/gzip.lua
```diff
@@ -21,17 +21,26 @@ local req_http_version = ngx.req.http_version
...
+                {
+                    type = "array",
+                    minItem = 1,
+                    items = {
+                        type = "string",
+                        minLength = 1,
+                    },
+                },
+                {
+                    type = "string",
```
spacewander, line 41: I think `enum = {"*"}` is enough?

t/plugin/gzip.t
```diff
@@ -436,6 +436,53 @@ POST /echo
 012345678
 --- more_headers
 Accept-Encoding: gzip
+Content-Type: text/html
```
spacewander, line 439: Better to use a special type to verify that any type can be matched.

apisix/plugins/gzip.lua
RocFang: @spacewander Yes, I think it's simpler to use {"*"}, this pr only try to follow nginx's config strictly.
If that not import, I will change it to {"*"}

RocFang: misunderstood you firstly, now i got it.

spacewander (approved):

shuaijinchao (approved):

PR #4844, opened 2021-08-18
ci: Remove non-exist dependency packages for CentOS 7

imjoey:

Due to the `build-essentials`, `libncurses5-dev`, `libreadline-dev`, `libssl-dev` packages do not exist in CentOS 7, which leads to CI always complain about the following error. Although this error will not make CI CentOS 7 failed, it may mislead users to do the same action. So this PR will remove the installation of these four packages.

```
# yum install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.dal.nexril.net
 * extras: repos.lax.quadranet.com
 * updates: mirror.vacares.com
No package build-essential available.
No package libncurses5-dev available.
No package libreadline-dev available.
No package libssl-dev available.
```

### Pre-submission checklist:

### DIFF
--- a/ci/centos7-ci.sh
+++ b/ci/centos7-ci.sh
@@ -39,7 +39,7 @@ install_dependencies() {
     rm -rf etcd-v3.4.0-linux-amd64
 
     # install test::nginx
-    yum install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl
+    yum install -y cpanminus perl
     cpanm --notest Test::Nginx IPC::Run > build.log 2>&1 || (cat build.log && exit 1)
 
     # install and start grpc_server_example


imjoey:
It's ready for review. Ping @spacewander @tokers @Yiyiyimu @juzhiyuan, thanks.

.github/workflows/centos8-ci.yml
```diff
@@ -0,0 +1,118 @@
+name: CI Centos8
```
spacewander, line 1: Can we share the code like ` .github/workflows/build.yml` instead of repeating centos7-ci?

spacewander (changes requested):

spacewander:
Also, to reduce CI cost, we need to run it per week or few days.

.github/workflows/centos8-ci.yml
imjoey: @spacewander Nice catch and that would be much better for multiple operating systems and architectures. I would make a change. Thanks.

spacewander (approved):

imjoey:
Ping @tokers @Yiyiyimu @juzhiyuan for review, thanks.

juzhiyuan (approved):

Issue #4853, opened 2021-08-19
request help: support configurating the node listening address


wayne-cheng:
### Issue description

I hope that apisix could  be configurated the listening address.
In some special cases, for example, I plan to use a combination of APISIX and Keepalived to achieve high availability. 
I will turn on the relevant Linux kernel parameters (to enable the program to listen the IP that does not exist on the local host):
```shell
echo "net.ipv4.ip_nonlocal_bind=1" >> /etc/sysctl.conf
```
Then, bind a virtual IP (vip) to APISIX. However, the current APISIX configuration does not support specifying the listening address, although this is a regular usage in nginx. 
I tried to modify the node_listen configuration of config.yaml like this:
```yaml
apisix:
  node_listen:
  - port: 192.168.0.244:9080
```
This trick can meet the needs of configurating listening addresses. But the listening address of https cannot be modified in this way. listen_port can only accept parameters of type number:
```yaml
apisix:
  # ...
  ssl:
    enable: true
      enable_http2: true
      listen_port: 9443
```
The parsing of config.yaml needs to be optimized. 
I am willing to submit a PR If you think it's suitable.

### Environment

- apisix version: 2.8
- OS: Linux apisix-7c455cb576-kjbjw 3.10.0-957.el7.x86_64 `#1` SMP Thu Nov 8 23:39:32 UTC 2018 x86_64 Linux
- OpenResty version: nginx version: openresty/1.19.3.1
built by gcc 10.2.1 20201203 (Alpine 10.2.1_pre1)
built with OpenSSL 1.1.1k  25 Mar 2021
TLS SNI support enabled
[...]

PR #4856, opened 2021-08-19
feat:support configurating the node listening address (#4853)

wayne-cheng:
Signed-off-by: wayne-cheng <zhengwei@tiduyun.com>

support configurating the node listening address,
The `config.yaml` example as below：
```yaml
apisix:
  node_listen:
  - 9080
  - port: 9380
    ip: 127.0.0.2            # If not given, the default is 0.0.0.0
    allow_ssl: True        # if False, the ip connot be ued to config ssl.
```
refer #4853

### Pre-submission checklist:


wayne-cheng:
@spacewander Hi, PTAL

nic-chen (commented):
@wayne-cheng  tanks.need test cases

spacewander (changes requested):
A test under t/cli is in need.

apisix/cli/ngx_tpl.lua
```diff
@@ -459,11 +459,11 @@ http {
 
     server {
         {% for _, item in ipairs(node_listen) do %}
-        listen {* item.port *} default_server {% if enable_reuseport then %} reuseport {% end %} {% if item.enable_http2 then %} http2 {% end %};
-        {% end %}
-        {% if ssl.enable then %}
+        listen {* item.ip *}:{* item.port *} default_server {% if enable_reuseport then %} reuseport {% end %} {% if item.enable_http2 then %} http2 {% end %};
+        {% if item.allow_ssl and ssl.enable then %}
```
spacewander, line 463: Better to use enable_ssl.

spacewander (changes requested):

apisix/cli/ngx_tpl.lua
wayne-cheng: ok，and should I explain these params in the doc FAQ.md ?
```yaml
apisix:
  node_listen:
  - 9080
  - port: 9380
    ip: 127.0.0.1            # If not given, the default is `0.0.0.0`.
    enable_ssl: True         # If `False`, the ip connot be ued to enable ssl. If not given, the default is `True`.
```

spacewander: IMHO, document it in the conf/config-default.yaml as a comment is enough.

spacewander: And the default value of "enable_ssl" should not be true, since `9080` doesn't have "enable_ssl", and it doesn't enable SSL by default.

spacewander: Actually, we are planning to reduce the length of the current FAQ, let each answer go to suitable places, instead of putting them together.

wayne-cheng: > And the default value of "enable_ssl" should not be true, since `9080` doesn't have "enable_ssl", and it doesn't enable SSL by default.

If  the default value of "enable_ssl" is false, the old configuration will be affected.
Considering the param `enable_ssl` should be compatible with the original param `apisix.ssl.enable`, I design the param `enable_ssl`  as a control switch that cooperates with the param  `apisix.ssl.enable`:
```lua
{% if item.enable_ssl and ssl.enable then %}
{% for _, port in ipairs(ssl.listen_port) do %}
listen {* item.ip *}:{* port *} ssl default_server {% if ssl.enable_http2 then %} http2 {% end %} {% if enable_reuseport then %} reuseport {% end %};
{% end %}
{% end %}
```
Beacause there are other params about ssl ( `enable_http2`, `listen_port` etc. ) is set in the param `apisix.ssl`, the ssl config should not be complex in the param `apisix.node_listen[]`
```yaml
apisix
  ssl:
    enable: true
    enable_http2: true
    listen_port: 9443
```

wayne-cheng: It is better to modify the config `apisix.ssl.listen_port` ,like this:
```yaml
apisix:
  ssl:
    listen_port:
    - ip: 127.0.0.1
       port: 9443
```
The ssl config in the param `apisix.node_listen[]` could be confusing，for example:
```yaml
apisix:
  node_listen:
    - ip: 127.0.0.2
      port: 9082
      enable_ssl: true
    - ip: 127.0.0.2
      port: 9083
      enable_ssl: false
```
What do you think?  @spacewander

spacewander: Agree. Personally, I prefer to use a new key "listen" since it is not just with port. Maybe we need to move "enable_http2" under "listen" too. Like this one:
```
apisix:
  ssl:
    # enable_http2: false
    # listen_port: ...
    listen:
    - ip: 127.0.0.1
      port: 9443
      enable_http2: true # per listen configuration
```

@wayne-cheng
What's your opinion?

spacewander: Yes. We need to keep backward compatibility.

wayne-cheng: @spacewander I have updated the code, please take a look.

apisix/cli/ops.lua
```diff
@@ -461,45 +461,238 @@ Please modify "admin_key" in conf/config.yaml .
...
     if type(yaml_conf.apisix.node_listen) == "number" then
 
         if ports_to_check[yaml_conf.apisix.node_listen] ~= nil then
             util.die("node_listen port ", yaml_conf.apisix.node_listen,
                     " conflicts with ", ports_to_check[yaml_conf.apisix.node_listen], "\n")
         end
 
-        local node_listen = {{port = yaml_conf.apisix.node_listen}}
+        local node_listen = {}
+        table_insert(node_listen, {ip = "0.0.0.0", port = yaml_conf.apisix.node_listen})
```
spacewander, line 475: Strictly speaking, the default value of "ip" is "*", `listen 80` is equal to `listen *:80`.

conf/config-default.yaml
```diff
@@ -113,8 +116,14 @@ apisix:
...
-    enable_http2: true
-    listen_port: 9443
+    listen:
+    - 9443
+    - ip: 127.0.0.3               # Specific IP, If not set, the default value is `0.0.0.0`.
+      port: 9444
+      enable_http2: true          # If not set, the default value is `false`.
+    #port: 9081
+    #enable_http2: true
+    #listen_port: 9443
```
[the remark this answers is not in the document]
wayne-cheng: ok，I will comment it like this:
```yaml
apisix:
  node_listen: 9080                # APISIX listening port
  # node_listen:                   # This style support multiple ports
  # - 9080
  # - ip: 127.0.0.2                # Specific IP, If not set, the default value is `0.0.0.0`.
  #   port: 9082
  ssl:
    enable: true
    # listen:                       # APISIX listening port in https.
    # - 9443
    # - ip: 127.0.0.3               # Specific IP, If not set, the default value is `0.0.0.0`.
    #   port: 9444
    #   enable_http2: true          # If not set, the default value is `false`.
    enable_http2: true              # Not recommend: This parameter should be set via the `listen`.
    listen_port: 9443               # Not recommend: This parameter should be set via the `listen`.
```

apisix/cli/ops.lua
```diff
@@ -461,45 +461,238 @@ Please modify "admin_key" in conf/config.yaml .
...
+
+                if ports_to_check[port] ~= nil then
+                    util.die("ssl listen port ", port, " conflicts with ",
+                            ports_to_check[port], "\n")
+                end
+
+                local addr = ip .. ":" .. port
+
+                if ip_port_to_check[addr] == nil then
+                    table_insert(total_ssl_listen, {ip = ip, port = port, enable_http2 = yaml_conf.apisix.ssl.enable_http2})
```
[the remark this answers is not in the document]
wayne-cheng: Thx, the code is much clearer now

conf/config-default.yaml
```diff
@@ -20,7 +20,10 @@
 #
 
 apisix:
-  node_listen: 9080                # APISIX listening port
```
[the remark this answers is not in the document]
wayne-cheng: done

conf/config-default.yaml
```diff
@@ -20,7 +20,10 @@
...
+  node_listen:                     # APISIX listening port
+  - 9080
+  - ip: 127.0.0.2                  # Specific IP, If not set, the default value is `0.0.0.0`.
```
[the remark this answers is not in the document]
wayne-cheng: done

conf/config-default.yaml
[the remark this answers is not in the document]
spacewander: I prefer to:
```yaml
apisix:
  # node_listen: 9080                # APISIX listening port
  node_listen:                       # This style support multiple ports
   - 9080
   # - ip: 127.0.0.2                # Specific IP, If not set, the default value is `0.0.0.0`.
   #   port: 9082
  ssl:
    enable: true
    listen:                       # APISIX listening port in https.
      - 9443
    # - ip: 127.0.0.3               # Specific IP, If not set, the default value is `0.0.0.0`.
    #   port: 9444
    #   enable_http2: true          # If not set, the default value is `false`.
    enable_http2: true              # Not recommend: This parameter should be set via the `listen`.
    # listen_port: 9443               # Not recommend: This parameter should be set via the `listen`.
```

Let's encourage users to use the new style.

wayne-cheng:
@spacewander A new commit has been pushed, PTAL

apisix/cli/ngx_tpl.lua
```diff
@@ -459,11 +459,11 @@ http {
 
     server {
         {% for _, item in ipairs(node_listen) do %}
-        listen {* item.port *} default_server {% if enable_reuseport then %} reuseport {% end %} {% if item.enable_http2 then %} http2 {% end %};
+        listen {* item.ip *}:{* item.port *} default_server {% if enable_reuseport then %} reuseport {% end %};
```
spacewander, line 462: http port also need to enable_http2

apisix/cli/ops.lua
```diff
@@ -461,45 +461,137 @@ Please modify "admin_key" in conf/config.yaml .
...
-            util.die("node_listen port ", yaml_conf.apisix.node_listen,
-                    " conflicts with ", ports_to_check[yaml_conf.apisix.node_listen], "\n")
+        if ports_to_check[port] ~= nil then
+            util.die(scheme, " listen port ", port, " conflicts with ",
+                ports_to_check[port], "\n")
         end
 
-        local node_listen = {{port = yaml_conf.apisix.node_listen}}
-        yaml_conf.apisix.node_listen = node_listen
+        local addr = ip .. ":" .. port
```
spacewander, line 480: We can add scheme to this key so that we can reduce a branch

apisix/cli/ngx_tpl.lua
wayne-cheng: OK, and it means that the config `apisix.node_listen[]` also has the property `enable_http2` , right?
```yaml
apisix:
  # node_listen: 9080                # APISIX listening port
  node_listen:                   # This style support multiple ports
  - 9080
  # - ip: 127.0.0.2                # Specific IP, If not set, the default value is `0.0.0.0`.
  #   port: 9082
  #   enable_http2: true  
```

apisix/cli/ops.lua
wayne-cheng: Since the config `apisix.node_listen[]` also has the property `enable_http2`, the branch can be reduced.
Port conflicts should be checked at the TCP layer, not the scheme `http/https`, so the scheme would not be add to this key.

conf/config-default.yaml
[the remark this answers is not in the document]
wayne-cheng: done

apisix/cli/ops.lua
wayne-cheng: done

apisix/cli/ngx_tpl.lua
wayne-cheng: done

wayne-cheng:
The conifg `apisix.port_admin`  also need to be compatible, now all changes have been pushed.
PTAL @spacewander

apisix/cli/ngx_tpl.lua
```diff
@@ -399,7 +399,11 @@ http {
     {% if enable_admin and port_admin then %}
     server {
         {%if https_admin then%}
-        listen {* port_admin *} ssl;
+        {% for _, item in ipairs(ssl.listen) do %}
```
spacewander, line 402: I don't like the idea that the admin requires multiple ports.

spacewander:
Please don't change the `port_admin`, thanks!

apisix/cli/ngx_tpl.lua
spacewander: Multiple admin ports are useless. You only need to set one in the API client.

wayne-cheng: I did not change the config `port_admin`. It is still the same as before, only one `port_admin` can be set. 
The idea is to match the specific ip in the config `node_listen`，not for multiple ports.

apisix/cli/ngx_tpl.lua
spacewander: > it means that the config apisix.node_listen[] also has the property enable_http2 , right?

Yes.

wayne-cheng:
@spacewander 
Please see this after that code:
```lua
{% if item.only_admin then %}
listen {* item.ip *}:{* item.port *} ssl;
{% end %}
``` 
In my environment, the apisix bind the VIP. If the `port_admin` is bound to all IP addresses, those  programs with the same port but different IP will not work.

spacewander:
Maybe we can add an "admin_listen" section like what we have done with `ssl.listen_port`? We can do it in the next PR. This PR is big enough and should pass the CI first.

wayne-cheng:
Agree，Of course this way is better. Next, Do I undo the changes about `port_admin` and push commit again？

spacewander:
Yes

wayne-cheng:
OK，PTAL @spacewander

wayne-cheng:
@spacewander 
The config `apisix.node_listen` in previous `config-default.yaml` is the type number, but the new has been the  type table. 
If a user still uses the type number to set the config `apisix.node_listen` like this:
```yaml
apisix:
  node_listen: 9081
```
The error will occur when exec `make init`，and it also appeared in the CI test:
```shell
./bin/apisix init
/usr/local/openresty/luajit/bin/luajit ./apisix/cli/apisix.lua init
failed to read local yaml config of apisix: failed to merge, path[apisix->ssl->listen] expect: table, but got: number
make: *** [Makefile:127: init] Error 1
```
This bug should have existed before, but it was missed, since the previous file `config-default.yaml` used the number type.
Should I fix the function `merge_conf` in the code file `apisix/cli/file.lua`(+151) ?
```lua
 -- ...           
            elseif type(base[key]) ~= type_val then
                if (ppath == "nginx_config" or str_sub(ppath, 1, 14) == "nginx_config->") and
                    (type_val == "number" or type_val == "string")
                then
                    base[key] = val
                else
                    local path = ppath == "" and key or ppath .. "->" .. key
                    return nil, "failed to merge, path[" .. path ..  "] expect: " ..
                                type(base[key]) .. ", but got: " .. type_val
                end
            else
                base[key] = val
            end
-- ...
```

spacewander:
Thanks for your report. Let's add it to the allow list.

I think you can submit it via a separate PR. I can merge it first, so that you are no longer a first-time contributor, then you can see the CI result without approval.

conf/config-default.yaml
```diff
@@ -115,8 +120,16 @@ apisix:
   enable_resolv_search_opt: true  # enable search option in resolv.conf
   ssl:
     enable: true
-    enable_http2: true
-    listen_port: 9443
+    # listen: 9443                # APISIX listening port in https.
```
spacewander, line 123: I think we should only support table format for the new configuration, this will make it easier to understand.

spacewander (approved):

conf/config-default.yaml
wayne-cheng: ok，I have removed it in the `config-default.yaml`.

apisix/cli/file.lua
```diff
@@ -126,6 +126,10 @@ local function path_is_multi_type(path, type_val)
         return true
     end
 
+    if path == "apisix->ssl->listen" and type_val == "number" then
```
spacewander, line 129: Could we remove it now?

apisix/cli/ops.lua
```diff
@@ -461,45 +461,132 @@ Please modify "admin_key" in conf/config.yaml .
...
+                listen_table_insert(node_listen, "http", ip, port,
+                        enable_http2, enable_ipv6)
+            end
+        end
+    end
+    yaml_conf.apisix.node_listen = node_listen
+
+    local ssl_listen = {}
+    -- listen in https, support multiple ports, support specific IP
+    if type(yaml_conf.apisix.ssl.listen) == "number" then
```
spacewander, line 540: Ditto

wayne-cheng: Got it, the related code has also been removed.

spacewander (approved):

t/cli/test_main.sh
```diff
@@ -121,6 +121,54 @@ fi
...
+echo "
+apisix:
+  node_listen:
+    - ip: 127.0.0.1
+      port: 9081
+    - ip: 127.0.0.2
+      port: 9082
+      enable_http2: true
+  ssl:
+    enable_http2: false
```
tzssangglass, line 135: Is `enable_http2` here for backwards compatibility？

tzssangglass: I get

tzssangglass (approved):

wayne-cheng:
@spacewander Thanks for your patience to review, I'll submit next PR to add the config `admin_listen`.

jagerzhang:
Hi, When will this feature be released to the docker image?

PR #4863, opened 2021-08-20
perf(kafka-logger): directly return when can't identify the broker

okaybase:

Directly return when can't identify the broker, and don't continue exec the send logic.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/kafka-logger.lua
+++ b/apisix/plugins/kafka-logger.lua
@@ -142,7 +142,7 @@ end
 
 local function send_kafka_data(conf, log_message, prod)
     if core.table.nkeys(conf.broker_list) == 0 then
-        core.log.error("failed to identify the broker specified")
+        return nil, "failed to identify the broker specified"
     end
 
     local ok, err = prod:send(conf.kafka_topic, conf.key, log_message)
--- a/t/plugin/kafka-logger.t
+++ b/t/plugin/kafka-logger.t
@@ -722,3 +722,44 @@ GET /t
 [qr/partition_id: 1/,
 qr/partition_id: 0/,
 qr/partition_id: 2/]
+
+
+
+=== TEST 20: error log, failed to identify the broker specified
+--- config
+    location /t {
+        content_by_lua_block {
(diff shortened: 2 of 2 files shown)


okaybase:
@spacewander ci timeout, please help rerun, thanks~

tokers (approved):

apisix/plugins/kafka-logger.lua
```diff
@@ -205,6 +201,11 @@ function _M.log(conf, ctx)
         end
     end
 
+    if core.table.nkeys(broker_list) == 0 then
```
spacewander, line 204: Could we validate it via JSONSchema?

okaybase: Validate it via JSONSchema will rewrite the schema definition(broker_list):
image

spacewander: Yes. But as we already check it elsewhere, check it in JSONSchema doesn't break.

okaybase: okay~

spacewander: Could you provide some examples?

okaybase: Currently the `broker_list` is `object` type, if redefine the type of `broker_list` will break(e.g. `array` type), so in this case, would you give some suggestion~

spacewander: We don't need to redefine the type to "array". We can add validation to the object type. See https://json-schema.org/understanding-json-schema/reference/object.html

okaybase: the document has been seen, and we can add `minProperties = 1` validation to the `broker_list` schema, but this logic `if core.table.nkeys(broker_list) == 0 then` is also required, because check `port` type is number.

spacewander: We can check port type too: https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties
You can find some examples in APISIX.

PR #4881, opened 2021-08-24
fix: bug about merge config node_listen

wayne-cheng:
Signed-off-by: wayne-cheng <zhengwei@tiduyun.com>
This PR is used to solve a bug found in another PR (#4856).

When the conifg `apisix.node_liste`  is set to a array in the `conifg-default.yaml`, but the user's `config.yaml` use the type number,
a error will occur.

More detail, Please see https://github.com/apache/apisix/pull/4856#issuecomment-903545790

PTAL @spacewander 

You can use the existing test cases for testing, since the vaule type of the config `apisix.node_listen` in the `config-default.yaml` is changed to the  type array.

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -112,6 +112,28 @@ local function tinyyaml_type(t)
 end
 
 
+local function path_is_multi_type(path, type_val)
+    if str_sub(path, 1, 14) == "nginx_config->" and
+            (type_val == "number" or type_val == "string") then
+        return true
+    end
+
+    if path == "apisix->node_listen" and type_val == "number" then
+        return true
+    end
+
+    if path == "apisix->ssl->listen_port" and type_val == "number" then
+        return true
+    end
+
+    if path == "apisix->ssl->listen" and type_val == "number" then
+        return true
+    end
+
+    return false
+end
+
+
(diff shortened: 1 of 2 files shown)


spacewander (commented):
Could you add a test in t/cli/test_validate_config.sh?

wayne-cheng:
OK, Please retest. @spacewander

t/cli/test_validate_config.sh
```diff
@@ -69,3 +69,23 @@ fi
...
 
 echo "passed: allow configuring address in stream_proxy"
+
+sed -i 's/^  \(node_listen:\)/  #\1/g' conf/config-default.yaml
+sed -i 's/^    \(- 9080\)/    #\1/g' conf/config-default.yaml
+sed -i 's/^  # \(node_listen: 9080\)/  \1/g' conf/config-default.yaml
+
+echo '
+apisix:
+    node_listen:
```
spacewander, line 79: Could we keep the table in conf/config-default.yaml and use number in config.yaml? Better to avoid touching config-default.yaml in the test.

spacewander: I see.

spacewander (approved):

PR #4888, opened 2021-08-24
fix(plugin_config): recover plugin when plugin_config changed

wangxiaochuang:
fix the bug: 
    route plugin cannot update when the plugin_config changed in the second time


spacewander (commented):
Could you provide a test case to indicate which situation will cause the bug?

tokers (approved):

Issue #3753, opened 2021-03-04
request help: about proxy-mirror plugin


chenyuezhou:
### Issue description

Dose proxy-mirror support set mirror rate?

### Environment

* apisix version: `2.3`
* OS: `Linux-5.4.80`
* OpenResty / Nginx version: 
```
nginx version: openresty/1.19.3.1
built by gcc 8.3.1 20191121 (Red Hat 8.3.1-5) (GCC)
built with OpenSSL 1.1.1g FIPS  21 Apr 2020
TLS SNI support enabled
configure arguments: --prefix=/home/zcy/openresty/openresty-1.19.3.1/nginx/nginx --with-cc-opt=-O2 --add-module=../ngx_devel_kit-0.3.1 --add-module=../echo-nginx-module-0.62 --add-module=../xss-nginx-module-0.06 --add-module=../ngx_coolkit-0.2 --add-module=../set-misc-nginx-module-0.32 --add-module=../form-input-nginx-module-0.12 --add-module=../encrypted-session-nginx-module-0.08 --add-module=../srcache-nginx-module-0.32 --add-module=../ngx_lua-0.10.19 --add-module=../ngx_lua_upstream-0.07 --add-module=../headers-more-nginx-module-0.33 --add-module=../array-var-nginx-module-0.05 --add-module=../memc-nginx-module-0.19 --add-module=../redis2-nginx-module-0.15 --add-module=../redis-nginx-module-0.3.7 --add-module=../rds-json-nginx-module-0.15 --add-module=../rds-csv-nginx-module-0.09 --add-module=../ngx_stream_lua-0.0.9 --with-ld-opt='-Wl,-rpath,/home/zcy/openresty/openresty-1.19.3.1/nginx/luajit/lib -Wl,-rpath,/home/zcy/openresty/openresty-1.19.3.1/nginx/luajit/lib' --with-pcre=/home/zcy/openresty/openresty-1.19.3.1/bundle/pcre-8.40 --with-stream --with-stream_ssl_module --with-stream_ssl_preread_module --with-http_ssl_module --with-http_stub_status_module --with-http_realip_module --with-http_v2_module --with-pcre-jit --with-pcre-opt=-g --with-stream
```
* etcd version:
```
{"hostname":"localhost.localdomain","version":"2.3","id":"8842cd73-9d27-4745-8005-20d8e4383c29","etcd_version":"3.5.0","up_time":7566,"last_report_time":1614841246,"boot_time":1614833686}
```
[...]

PR #4965, opened 2021-09-02
feat(proxy-mirror): support mirror requests sample_ratio

okaybase:

support mirror requests sample_ratio

use case:
In the simulation environment, user hope to use a small amount of real traffic to verify the new function, but the real traffic may be huge, which puts great pressure on the simulation service.

Fix https://github.com/apache/apisix/issues/3753
### Pre-submission checklist:


spacewander (commented):
Could you give some use cases about this feature?

okaybase:
> Could you give some use cases about this feature?

In the simulation environment, we hope to use a small amount of real traffic to verify the new function, but the real traffic may be huge, which puts great pressure on the simulation service.

spacewander:
> 
> 
> > Could you give some use cases about this feature?
> > okay~
> > In the simulation environment, we hope to use a small amount of real traffic to verify the new function, but the real traffic may be huge, which puts great pressure on the simulation service.

Personally, I prefer to sample randomly than sample the first N. As,
1. sample randomly is more accurate. It doesn't miss requests after the first N
2. we don't need a place to store the count.

What about your opinion?
@membphis @tokers @tzssangglass

okaybase:
> > > Could you give some use cases about this feature?
> > > okay~
> > > In the simulation environment, we hope to use a small amount of real traffic to verify the new function, but the real traffic may be huge, which puts great pressure on the simulation service.
> 
> Personally, I prefer to sample randomly than sample the first N. As,
> 
> 1. sample randomly is more accurate. It doesn't miss requests after the first N
> 2. we don't need a place to store the count.
> 
> What about your opinion?
> @membphis @tokers @tzssangglass

Thanks for your careful review and suggestions firstly~ :smile: 
Perhaps random sampling could be one of the policies(random/local/redis/redis-cluster).

tzssangglass:
randoml +1

tokers:
> > > Could you give some use cases about this feature?
> > > okay~
> > > In the simulation environment, we hope to use a small amount of real traffic to verify the new function, but the real traffic may be huge, which puts great pressure on the simulation service.
> 
> Personally, I prefer to sample randomly than sample the first N. As,
> 
> 1. sample randomly is more accurate. It doesn't miss requests after the first N
> 2. we don't need a place to store the count.
> 
> What about your opinion?
> @membphis @tokers @tzssangglass

I think a random sample is better than the current way. Also, if we support these rate-limiting ways, we have to change multiple plugins in the future if new rate-limiting algorithms are introduced, and it's prone to be inconsistent on the configurations items.

okaybase:
Will use random sampling percentages to achieve this function~ @spacewander @tokers @tzssangglass

apisix/plugins/proxy-mirror.lua
```diff
@@ -53,7 +60,18 @@ function _M.rewrite(conf, ctx)
     core.log.info("proxy mirror plugin rewrite phase, conf: ", core.json.delay_encode(conf))
 
     ctx.var.upstream_host = ctx.var.host
-    ctx.var.upstream_mirror_host = conf.host
+
+    if not conf.percentage or conf.percentage == 1 then
+        ctx.var.upstream_mirror_host = conf.host
+    else
+        math.randomseed(socket.gettime())
```
tokers, line 67: `randomseed` was already called in `http_init_worker`.

okaybase: thanks~ done~

apisix/plugins/proxy-mirror.lua
```diff
@@ -15,7 +15,7 @@
 -- limitations under the License.
 --
 local core          = require("apisix.core")
-
+local math = math
```
tzssangglass, line 18:
```suggestion
local math_random = math.random
```

apisix/plugins/proxy-mirror.lua
```diff
@@ -53,7 +59,17 @@ function _M.rewrite(conf, ctx)
     core.log.info("proxy mirror plugin rewrite phase, conf: ", core.json.delay_encode(conf))
 
     ctx.var.upstream_host = ctx.var.host
-    ctx.var.upstream_mirror_host = conf.host
+
+    if not conf.percentage or conf.percentage == 1 then
+        ctx.var.upstream_mirror_host = conf.host
+    else
+        local val = math.random()
```
tzssangglass, line 66:
```suggestion
        local val = math_random()
```

tzssangglass (approved):
LGTM

apisix/plugins/proxy-mirror.lua
```diff
@@ -26,6 +26,12 @@ local schema = {
             pattern = [[^http(s)?:\/\/[a-zA-Z0-9][-a-zA-Z0-9]{0,62}]]
                       .. [[(\.[a-zA-Z0-9][-a-zA-Z0-9]{0,62})+(:[0-9]{1,5})?$]],
         },
+        percentage = {
```
spacewander, line 29: IMHO, we use sample_ratio for this purpose in skywalking & zipkin. What about using this name?

t/plugin/proxy-mirror.t
```diff
@@ -445,3 +445,143 @@ GET /t
...
+GET /t
+--- error_code: 200
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 17: hit route with percentage 0.5
```
spacewander, line 582: Could we check if almost 50% requests reach the mirror?

apisix/plugins/proxy-mirror.lua
okaybase: okay~

t/plugin/proxy-mirror.t
okaybase: Logs are currently checked~

50% it might not be very very accurate in the random case~

image

image

spacewander: What about trying it 180 times? Just like https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md#send-request

spacewander: TO design the almost range, I wrote a python script:
```
#!/usr/bin/env python3
# coding: utf-8
curr = [1, 1]
turns = 180
for i in range(1, turns+1):
    prev = curr
    curr = []
    curr.append(1)
    for j in range(1, i):
        curr.append(prev[j-1] + prev[j])
    curr.append(1)
n = sum(curr)
for i in range(turns+1):
    print("%d / %d is %f" % (i, turns, curr[i] / float(n)))
exp_per = 0.999
mid = int(turns / 2)
for i in range(1, mid):
    n_per = 0
    for j in range(mid - i, mid + i + 1):
        n_per += curr[j] / float(n)
        if n_per > exp_per:
            print(mid - i, j)
            break
    if n_per > exp_per:
        break

```

okaybase: okay~ thanks a lot~ done~

okaybase:
the random sample and test cases done~ @spacewander @tokers

docs/en/latest/plugins/proxy-mirror.md
```diff
@@ -29,13 +29,14 @@ The proxy-mirror plugin, which provides the ability to mirror client requests.
 
 | Name | Type   | Requirement | Default | Valid | Description                                                                                                                 |
 | ---- | ------ | ----------- | ------- | ----- | --------------------------------------------------------------------------------------------------------------------------- |
-| host | string | optional    |         |       | Specify a mirror service address, e.g. http://127.0.0.1:9797 (address needs to contain schema: http or https, not URI part) |
+| host | string | required    |         |       | Specify a mirror service address, e.g. http://127.0.0.1:9797 (address needs to contain schema: http or https, not URI part) |
+| sample_ratio | number | optional    | 1       |  [0.00001, 1]     | the sample_ratio of mirror requests |
```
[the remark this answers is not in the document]
okaybase: thanks~ done~ @tokers

t/plugin/proxy-mirror.t
```diff
@@ -445,3 +445,139 @@ GET /t
...
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/hello"
+                   }]]
+                   )
+
+               if code == 200 then
+                   for i = 1, 200 do
+                       t('/hello?sample_ratio=0.5', ngx.HTTP_GET)
```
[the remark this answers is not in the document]
okaybase: Could you give some tips~ @spacewander 
Current the hit count is checked via: `qr/(uri: \/hello\?sample_ratio=0\.5){75,125}/`

apisix/plugins/proxy-mirror.lua
```diff
@@ -53,7 +59,18 @@ function _M.rewrite(conf, ctx)
...
 
     ctx.var.upstream_host = ctx.var.host
-    ctx.var.upstream_mirror_host = conf.host
+
+    if not conf.sample_ratio or conf.sample_ratio == 1 then
+        ctx.var.upstream_mirror_host = conf.host
+    else
+        local val = math_random()
+        core.log.info("mirror request sample_ratio conf: ", conf.sample_ratio,
+            ", random value: ", val)
```
membphis, line 68: bad indentation

membphis (approved):

apisix/plugins/proxy-mirror.lua
okaybase: okay~ done~ thanks a lot~

t/plugin/proxy-mirror.t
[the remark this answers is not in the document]
spacewander: We can add a `/inc` API, it will return an increasing number each call. Like,
```
GET /inc
1
GET /inc
2
GET /inc
3
```

So by hitting the API we can know how many requests are mirrored.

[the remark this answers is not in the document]
okaybase: Thanks firstly~
According to the document, seems the mirror address only support host~
@spacewander 

https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/proxy-mirror.md#attributes

image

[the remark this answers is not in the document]
spacewander: Yes, both the client and mirror hit the /inc API, so the number of request from mirror is total - client.

t/plugin/proxy-mirror.t
```diff
@@ -445,3 +455,163 @@ GET /t
...
+           }
+       }
+--- request
+GET /t
+--- error_code: 200
+--- response_body
+passed
+--- error_log_like eval
+qr/(uri: \/hello\?sample_ratio=0\.5){75,125}/
+--- timeout: 60
```
spacewander, line 617: We can send requests parallelly like https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md#send-request

Remember to use a high worker_connections in this case.

okaybase: okay, the test case has changed to send requests parallelly, but the ci log don't show more info about the test status 500, could you have a look at it~ @spacewander

t/plugin/proxy-mirror.t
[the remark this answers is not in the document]
okaybase: okay~ will write a new backend API~

spacewander:
Try this patch?
```
diff --git t/plugin/proxy-mirror.t t/plugin/proxy-mirror.t
index 888e9a86..a6b7e3ca 100644
--- t/plugin/proxy-mirror.t
+++ t/plugin/proxy-mirror.t
@@ -38,7 +38,7 @@ add_block_preprocessor(sub {
                 local http = require("resty.http")

                 local httpc = http.new()
-                local url = "http://127.0.0.1:1984/stat_count?action=inc"
+                local url = "http://127.0.0.1:1980/stat_count?action=inc"
                 local res, err = httpc:request_uri(url, {method = "GET"})
                 if not res then
                     core.log.error(err)
@@ -588,46 +588,14 @@ passed



-=== TEST 17: set route(stat_count)
---- config
-       location /t {
-           content_by_lua_block {
-               local t = require("lib.test_admin").test
-               local code, body = t('/apisix/admin/routes/2',
-                    ngx.HTTP_PUT,
-                    [[{
-                        "upstream": {
-                            "nodes": {
-                                "127.0.0.1:1980": 1
-                            },
-                            "type": "roundrobin"
-                        },
-                        "uri": "/stat_count"
-                   }]]
-                   )
-
-               if code >= 300 then
-                   ngx.status = code
-               end
-               ngx.say(body)
-           }
-       }
---- request
-GET /t
---- error_code: 200
---- response_body
-passed
-
-
-
-=== TEST 18: send batch requests and get mirror stat count
+=== TEST 17: send batch requests and get mirror stat count
 --- config
        location /t {
            content_by_lua_block {

                 -- reset count
-                local code, count = t('/stat_count?action=reset', ngx.HTTP_GET)
+                local code, _, count = t('/stat_count?action=reset', ngx.HTTP_GET)
                 if code >= 300 then
                    ngx.status = code
                 end
@@ -645,10 +613,12 @@ passed
                 end

                 -- get mirror stat count
-                code, count = t('/stat_count', ngx.HTTP_GET)
+                local code, _, count = t('/stat_count', ngx.HTTP_GET)
                 if code >= 300 then
                    ngx.status = code
+                   return
                 end
+                count = tonumber(count)

                 assert(count >= 75 and count <= 125, "mirror request count not in [75, 125]")
            }
@@ -657,4 +627,3 @@ passed
 GET /t
 --- error_log_like eval
 qr/(uri: \/hello\?sample_ratio=0\.5){75,125}/
---- timeout: 60
```

spacewander:
Since the patch is so big, I decide to push it to this branch instead (also trigger the CI)

spacewander:
According to the error.log, it seems that the mirror request may not
finish when the HTTP request is done, if the CPU resource is low. So just roll back to check it in the error log

spacewander (approved):

okaybase:
> According to the error.log, it seems that the mirror request may not
> finish when the HTTP request is done, if the CPU resource is low. So just roll back to check it in the error log

okay~ thanks a lot~

PR #4977, opened 2021-09-03
feat: allow route to inherit hosts from service

spacewander:

### Pre-submission checklist:

### DIFF
--- a/t/router/radixtree-host-uri2.t
+++ b/t/router/radixtree-host-uri2.t
@@ -376,3 +376,39 @@ GET /server_port
 Host: test.com
 --- no_error_log
 [error]
+
+
+
+=== TEST 14: inherit hosts from services
+--- yaml_config eval: $::yaml_config
+--- apisix_yaml
+services:
+  - id: 1
+    hosts:
+      - bar.com
+upstreams:
+  - id: 1
+    nodes:
+      "127.0.0.1:1980": 1
+    type: roundrobin
+routes:
+  -
+    service_id: 1
+    upstream_id: 1
+    uri: /hello
+    plugins:
+        proxy-rewrite:
+            uri: /hello1
+  -
+    upstream_id: 1
+    uri: /hello
+    priority: -1
+#END
+--- more_headers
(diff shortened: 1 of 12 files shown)


tokers (approved):

t/router/radixtree-host-uri2.t
```diff
@@ -376,3 +376,39 @@ GET /server_port
...
+
+
+
+=== TEST 14: inherit hosts from services
+--- yaml_config eval: $::yaml_config
+--- apisix_yaml
+services:
+  - id: 1
+    hosts:
+      - bar.com
```
tzssangglass, line 388: the service hosts here is `bar.com`, but the host in request is `www.foo.com`,how they match?

t/router/radixtree-host-uri3.t
```diff
@@ -0,0 +1,124 @@
...
+
+            if code >= 300 then
+                ngx.status = code
+                ngx.say(body)
+                return
+            end
+            ngx.sleep(0.1)
+
+            local httpc = http.new()
+            local res, err = httpc:request_uri(uri, {headers = {Host = "foo.com"}})
```
nic-chen, line 114: should be better if we have another test case without `Host`

t/router/radixtree-host-uri2.t
spacewander: Yes, we test with unmatched. If the host doesn't take effect, the route will match and the response body is "hello1 world".

t/router/radixtree-host-uri3.t
spacewander: Added.

nic-chen (approved):

tzssangglass (approved):

Donghui0:
in `plugin.merge_service_route` function,  need to add processing for route.value.timeout. Otherwise, the timeout attribute of ApisixRoute will be invalid.

I have a question in the `merge service_route` method. Why not directly core.table.deepcopy(route_conf), assign the attribute of service to new_conf ?

Issue #4441, opened 2021-06-17
request help: stream route's "remote_addr" parameter support multiple ip addresses or match rules like "192.168.0.0/16" 


Yullin:
### Environment

Request help without environment information will be ignored or closed.

* apisix version (cmd: `apisix version`): 2.6
* OS (cmd: `uname -a`): CentOS7
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): 1.19.3.2
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
* apisix-dashboard version, if have:
* luarocks version, if the issue is about installation (cmd: `luarocks --version`):

spacewander:
https://github.com/apache/apisix/blob/8578fe032e871fbcd151fa78653040c6bf2e885d/apisix/stream/router/ip_port.lua#L36

PR is welcome!

PR #4980, opened 2021-09-03
feat(stream_route): support CIDR in ip match

Zheaoli:
### Pre-submission checklist:

### DIFF
--- a/apisix/stream/router/ip_port.lua
+++ b/apisix/stream/router/ip_port.lua
@@ -187,7 +187,7 @@ function _M.routes()
     return user_routes.values, user_routes.conf_version
 end
 
-local function plugin_checker(item)
+local function stream_route_checker(item)
     if item.plugins then
         local err, message = stream_plugin_checker(item)
         if not err then
@@ -207,6 +207,7 @@ local function plugin_checker(item)
     end
     return true
 end
+_M.stream_route_checker = stream_route_checker
 
 
 function _M.stream_init_worker(filter)
@@ -214,7 +215,7 @@ function _M.stream_init_worker(filter)
     user_routes, err = core.config.new("/stream_routes", {
             automatic = true,
             item_schema = core.schema.stream_route,
-            checker = plugin_checker,
+            checker = stream_route_checker,
             filter = filter,
         })
 
--- a/apisix/stream/router/ip_port.lua
+++ b/apisix/stream/router/ip_port.lua
@@ -15,6 +15,7 @@
 -- limitations under the License.
 --
 local core      = require("apisix.core")
+local core_ip  = require("apisix.core.ip")
 local config_util = require("apisix.core.config_util")
 local plugin_checker = require("apisix.plugin").stream_plugin_checker
 local router_new = require("apisix.utils.router").new
@@ -23,7 +24,6 @@ local error     = error
 local tonumber  = tonumber
 local ipairs = ipairs
 
-
 local user_routes
 local router_ver
 local tls_router
@@ -31,16 +31,21 @@ local other_routes = {}
 local _M = {version = 0.1}
 
 
+
 local function match_addrs(route, vars)
     -- todo: use resty-ipmatcher to support multiple ip address
-    if route.value.remote_addr and
-       route.value.remote_addr ~= vars.remote_addr then
-        return false
+    if route.value.remote_addr_matcher then
+        local ok, _ = route.value.remote_addr_matcher:match(vars.remote_addr)
+        if not ok then
+            return false
+        end
     end
 
-    if route.value.server_addr and
-       route.value.server_addr ~= vars.server_addr then
-        return false
+    if route.value.server_addr_matcher then
+        local ok, _ = route.value.server_addr_matcher:match(vars.server_addr)
+        if not ok then
+            return false
+        end
     end
 
     -- todo: use resty-ipmatcher to support multiple ip address
@@ -71,6 +76,18 @@ do
             end
 
             local route = item.value
+            if item.value.remote_addr then
+                local remote_matcher = core_ip.create_ip_matcher({item.value.remote_addr})
+                if remote_matcher then
+                    item.value.remote_addr_matcher = remote_matcher
+                end
+            end
+            if item.value.server_addr then
+                local server_matcher = core_ip.create_ip_matcher({item.value.server_addr})
+                if server_matcher then
+                    item.value.server_addr_matcher = server_matcher
+                end
+            end
             if not route.sni then
                 other_routes[other_routes_idx] = item
                 other_routes_idx = other_routes_idx + 1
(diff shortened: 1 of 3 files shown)


Zheaoli:
https://github.com/apache/apisix/issues/4441

nic-chen (commented):
need test cases

apisix/stream/router/ip_port.lua
```diff
@@ -33,14 +35,38 @@ local _M = {version = 0.1}
 
 local function match_addrs(route, vars)
     -- todo: use resty-ipmatcher to support multiple ip address
-    if route.value.remote_addr and
-       route.value.remote_addr ~= vars.remote_addr then
-        return false
+    if route.value.remote_addr then
+        local ip = remote_addr_match_cache[route.value.remote_addr]
```
tokers, line 39: 1. It's dangerous to use a module-level table to cache the match result, since it may cause the memory leak;
2. The IP match overheads are not so high, the caching is unnecessary;

Zheaoli: Make sense, I'll update mycode

apisix/stream/router/ip_port.lua
```diff
@@ -33,14 +33,30 @@ local _M = {version = 0.1}
 
 local function match_addrs(route, vars)
     -- todo: use resty-ipmatcher to support multiple ip address
-    if route.value.remote_addr and
-       route.value.remote_addr ~= vars.remote_addr then
-        return false
+    if route.value.remote_addr then
```
spacewander, line 36: Better to put the init in the `create_router` instead of the match

t/stream-plugin/mqtt-proxy.t
```diff
@@ -220,7 +220,56 @@ hello world
 
 
 
-=== TEST 9: set route with invalid host
+=== TEST 9: set route with IP CIDR
```
spacewander, line 223: New test should be added from the bottom of the file. And it should be in stream-node/sanity.t

rockspec/apisix-master-0.rockspec
```diff
@@ -69,6 +69,7 @@ dependencies = {
     "ext-plugin-proto = 0.3.0",
     "casbin = 1.26.0",
     "api7-snowflake = 2.0-1",
+    "lua-resty-ipmatcher = 0.6.1"
```
spacewander, line 72: Don't add lua-resty-ipmatcher twice

apisix/stream/router/ip_port.lua
Zheaoli: Sorry I don't get your means, you mean remove the match function?

t/stream-node/sanity.t
```diff
@@ -281,3 +281,41 @@ GET /t
 hello world
 --- no_error_log
 [error]
+
+
+
+=== TEST 11: set stream route (id: 1) which uses upstream_id
```
spacewander, line 287: The name should show the difference from the others, like `CIDR`

t/stream-node/sanity.t
```diff
@@ -281,3 +281,41 @@ GET /t
...
+
+
+=== TEST 12: hit route
+--- stream_enable
+--- stream_request eval
+mmm
+--- stream_response
+hello world
+--- no_error_log
+[error]
```
tzssangglass, line 321: Add blank line at the end

tzssangglass (approved):

apisix/stream/router/ip_port.lua
```diff
@@ -71,6 +96,20 @@ do
             end
 
             local route = item.value
+
+            if not item.value.remote_addr_matcher and item.value.remote_addr then
+                local ip = create_matcher(item.value.remote_addr)
+                if ip then
```
spacewander, line 102: We can write a new checker in https://github.com/apache/apisix/blob/df0db95ce8f8a5a977481539b6af7e056d295680/apisix/stream/router/ip_port.lua#L185 to ensure the matcher can be always created?

We can use the methods in https://github.com/apache/apisix/blob/master/apisix/core/ip.lua to do this.

spacewander: We can use validate_cidr_or_ip to validate.

Zheaoli: I think it's unnecessary to use `validate_cidr_or_ip` here, we just need to check the `remote_addr` and `server_addr` is nil or not. the `ipmatcher` will help to validate the address when create a new match object by using `new` method

apisix/stream/router/ip_port.lua
```diff
@@ -71,6 +82,20 @@ do
             end
 
             local route = item.value
+
+            if not item.value.remote_addr_matcher and validate_address(item.value.remote_addr) then
```
spacewander, line 86: Please add the check to https://github.com/apache/apisix/blob/df0db95ce8f8a5a977481539b6af7e056d295680/apisix/stream/router/ip_port.lua#L185

apisix/stream/router/ip_port.lua
```diff
@@ -71,6 +76,18 @@ do
             end
 
             local route = item.value
+            if item.value.remote_addr then
+                local remote_matcher = core_ip.create_ip_matcher({item.value.remote_addr})
+                if remote_matcher then
```
spacewander, line 81: We don't need to check the returned value, it should be checked in the check function.

apisix/stream/router/ip_port.lua
```diff
@@ -176,13 +193,31 @@ function _M.routes()
     return user_routes.values, user_routes.conf_version
 end
 
+local function address_checker(item)
```
spacewander, line 196: This function doesn't only check address.
```suggestion
local function checker(item)
```

apisix/stream/router/ip_port.lua
```diff
@@ -176,13 +193,31 @@ function _M.routes()
...
+    if item.plugins then
+        return plugin_checker(item)
```
spacewander, line 198: Need to check both plugins and address

apisix/stream/router/ip_port.lua
Zheaoli: The check process in the check function just checks the address if server_addr or remote_addr is validated.

But in some circumstances, both the remote_address and server_address are not existed, but the configuration is validated. Here's the example in https://github.com/apache/apisix/blob/master/docs/zh/latest/stream-proxy.md#%E6%8E%A5%E6%94%B6-tls-over-tcp

```bash
curl http://127.0.0.1:9080/apisix/admin/stream_routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "sni": "a.test.com",
    "upstream": {
        "nodes": {
            "127.0.0.1:5991": 1
        },
        "type": "roundrobin"
    }
}'
```
So we can use the nil-check code here to avoid a useless matcher-created processes.

spacewander: I mean the returned value of create_ip_matcher

t/stream-node/sanity.t
```diff
@@ -281,3 +281,41 @@ GET /t
 hello world
 --- no_error_log
 [error]
+
+
+
+=== TEST 11: set stream route (id: 1) which uses upstream_id and remote address with IP CIDR
```
spacewander, line 287: Let's add a test that the conf is rejected by validate_cidr_or_ip

spacewander (approved):

PR #5011, opened 2021-09-07
feat(control): add new dump route config api for control api

Zheaoli:
### Pre-submission checklist:

Add new API for dumping config in control api

### DIFF
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -156,6 +156,40 @@ function _M.get_health_checker()
     return 200, info
 end
 
+local function iter_add_get_routes_info(infos, values, route_id)
+    for _, route in core.config_util.iterate_values(values) do
+        local new_route = core.table.deepcopy(route)
+        if new_route.value.upstream and new_route.value.upstream.parent then
+            new_route.value.upstream.parent = nil
+        end
+        core.table.insert(infos, new_route)
+        if route_id and route.value.id == route_id then
+            return new_route, true
+        end
+    end
+    return nil, false
+end
+
+function _M.dump_all_routes_info()
+    local routes = get_routes()
+    local infos = {}
+    iter_add_get_routes_info(infos, routes, nil)
+    return 200, infos
+end
+
+function _M.dump_route_info()
+    local routes = get_routes()
+    local uri_segs = core.utils.split_uri(ngx_var.uri)
+    local route_id = uri_segs[4]
+    local infos = {}
+    local route, flag = iter_add_get_routes_info(infos, routes, route_id)
+    if not route and not flag then
+        return 404, {error_msg = str_format("route[%s] not found", route_id)}
+    end
+    return 200, route
+end
+
+
 
 function _M.trigger_gc()
     -- TODO: find a way to trigger GC in the stream subsystem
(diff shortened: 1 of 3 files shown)


apisix/control/v1.lua
```diff
@@ -189,4 +212,10 @@ return {
         uris = {"/gc"},
         handler = _M.trigger_gc,
     },
+    -- /v1/dump
```
spacewander, line 215: Please use restful like API names for dump specific resources, like `/v1/upstreams/xxx` and `/v1/upstreams`.

apisix/control/v1.lua
```diff
@@ -156,6 +157,28 @@ function _M.get_health_checker()
     return 200, info
 end
 
+local function iter_add_get_routes_info(infos, values)
+    local temp_infos = {}
+    for _, route in core.config_util.iterate_values(values) do
+        local new_route = copy.deepcopy(route)
```
spacewander, line 163: We already have `core.table.deepcopy`

docs/en/latest/control-api.md
```diff
@@ -204,3 +204,160 @@ Introduced since `v2.8`.
...
+[
+  {
+    "value": {
+      "uris": [
+        "/hello"
+      ],
+      "id": "1",
+      "upstream": {
+        "scheme": "http",
+        "checks": {
```
spacewander, line 224: The example here is too long to focus on the point. Can we remove the healthcheck part?

t/control/routes.t
```diff
@@ -0,0 +1,171 @@
...
+    location /t {
+        content_by_lua_block {
+            local json = require("toolkit.json")
+            local t = require("lib.test_admin")
+            local code, body, res = t.test('/v1/routes',
+                ngx.HTTP_GET)
+            res = json.decode(res)
+            if res[1] then
+                local data = {}
+                data.uris = res[1].value.uris
```
spacewander, line 81: We should also check the upstream part

t/control/routes.t
```diff
@@ -0,0 +1,171 @@
...
+  -
+    id: 1
+    uris:
+        - /hello
+    upstream:
+      nodes:
+        "127.0.0.1:1980": 1
+        "127.0.0.1:1988": 1
+      type: roundrobin
+      checks:
```
spacewander, line 146: We don't need to define a long upstream for a test just check if the route exists

apisix/control/v1.lua
```diff
@@ -156,6 +156,40 @@ function _M.get_health_checker()
     return 200, info
 end
 
+local function iter_add_get_routes_info(infos, values, route_id)
```
spacewander, line 159: The infos can be created inside this function, and there is no need to pass out a flag -- use nil if route_id given but not matched.

docs/en/latest/control-api.md
Zheaoli: of course

apisix/control/v1.lua
```diff
@@ -156,6 +156,42 @@ function _M.get_health_checker()
...
+        end
+        core.table.insert(infos, new_route)
+        if route_id and route.value.id == route_id then
+            return new_route, true
+        end
+    end
+    if not route_id then
+        return infos, true
+    end
+    return nil, false
```
spacewander, line 174: We don't need a flag, just check the first result is nil or not

t/control/routes.t
spacewander: Need to check the upstream part too

spacewander (approved):

tzssangglass (approved):

PR #5136, opened 2021-09-24
docs: update how-to-build.md

candyleer:
add command make install in case of `command not found`

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/how-to-build.md
+++ b/docs/en/latest/how-to-build.md
@@ -76,6 +76,8 @@ Please refer to: [Installing Apache APISIX with Helm Chart](https://github.com/a
   cd apisix-2.9
   # Create dependencies
   make deps
+  # Install apisix command
+  make install
   ```
 
 ## Step 3: Manage Apache APISIX Server
--- a/docs/zh/latest/how-to-build.md
+++ b/docs/zh/latest/how-to-build.md
@@ -76,11 +76,14 @@ sudo yum install -y https://github.com/apache/apisix/releases/download/2.9/apisi
   cd apisix-2.9
   # 创建依赖
   make deps
+  
+  # 安装 apisix 命令
+  make install
   ```
 
 ## 步骤3：管理 Apache APISIX 服务
 
-我们可以在 Apache APISIX 的目录下使用命令初始化依赖、启动服务和停止服务，也可以通过 `make help` 命令查看所有命令和对应的功能。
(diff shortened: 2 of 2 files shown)


docs/en/latest/how-to-build.md
```diff
@@ -76,6 +76,8 @@ Please refer to: [Installing Apache APISIX with Helm Chart](https://github.com/a
   cd apisix-2.9
   # Create dependencies
   make deps
+  # Install apisix command
+  make install
```
leslie-tsang, line 80: Please update [CN version](https://github.com/apache/apisix/blob/master/docs/zh/latest/how-to-build.md) at the same time.

leslie-tsang (changes requested):

tzssangglass:
why we change `make *` to `apisix *`? for example,  if `apisix` is not in the path search variable, you will get error as `command not found`.

`make stop` = `./bin/apisix stop`, it works well.

candyleer:
> why we change `make *` to `apisix *`? for example, if `apisix` is not in the path search variable, you will get error as `command not found`.
> 
> `make stop` = `./bin/apisix stop`, it works well.

english version has been changed in https://github.com/apache/apisix/pull/5075, so chinese version follow up

tzssangglass:
> > why we change `make *` to `apisix *`? for example, if `apisix` is not in the path search variable, you will get error as `command not found`.
> > `make stop` = `./bin/apisix stop`, it works well.
> 
> english version has been changed in #5075, so chinese version follow up

got

docs/en/latest/how-to-build.md
tzssangglass: I think `make install` means installing the APISIX code into luarocks. The context here is to download the APISIX source code, so does the source code still need to be installed to luarocks? I thought it was no longer needed. what's your option? @spacewander @membphis

spacewander: Yes. See the comment here:
https://github.com/apache/apisix/blob/6162ca0ef92445fdd15b7a213655887b66cd4eea/Makefile#L167

spacewander: I changed my mind. Look like we still need this command to install APISIX to PATH even `make install` is designed for luarocks.

tzssangglass: why

spacewander: As there is no other command to do it.

spacewander (approved):

tzssangglass (approved):

PR #5257, opened 2021-10-18
fix: shm health check is not enabled after upgrading etcd client to 1.6.0

shuaijinchao:

After lua-resty-etcd is upgraded to 1.6.0, if it has been initialized in the default mode, the health check mode needs to be switched by judging whether the shared memory exists when performing the health check configuration.

https://github.com/apache/apisix/blob/93038127c54235f5ac786b7ab10aee4231cffc4c/apisix/core/config_etcd.lua#L521

https://github.com/api7/lua-resty-etcd/blob/9995f38e4da90af655528e6e8d40e3d6fa5bc946/health_check.md#L59

### Pre-submission checklist:

### DIFF
--- a/t/core/config_etcd.t
+++ b/t/core/config_etcd.t
@@ -339,6 +339,9 @@ etcd:
 GET /t
 --- response_body
 passed
+--- grep_error_log eval
+qr/healthy check use \S+ \w+/
 --- grep_error_log_out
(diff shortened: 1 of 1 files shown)


shuaijinchao:
> Could you provide a test for it?

OK

Issue #5294, opened 2021-10-20
bug: the zipkin plugin has a logic bug in the header_filter phase


zuiyangqingzhou:
### Issue description

https://github.com/apache/apisix/blob/release/2.10.0/apisix/plugins/zipkin.lua#L242-L252

There are two problems with this code：
1. `ctx.HEADER_FILTER_END_TIME` variable is not reachable at line 251
2. `ctx.HEADER_FILTER_END_TIME` is not necessary, it is only used in the header_filter phase

### Environment

- apisix version (cmd: `apisix version`):
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

### Steps to reproduce

none

### Actual result

none

### Error log

none

PR #5295, opened 2021-10-20
fix(zipkin): response_span doesn't have correct start time

zuiyangqingzhou:
fix: https://github.com/apache/apisix/issues/5294

### Pre-submission checklist:


zuiyangqingzhou:
> Could you provide a test case? You can hook the "start_child_span" to check if a valid time is given. You can refer
> 
> https://github.com/apache/apisix/blob/4dd1f6799f8207ed2f07da0e2a34c282b26cd877/t/plugin/skywalking.t#L36

OK

PR #5364, opened 2021-10-29
chore: replace submodule tmate with github action tmate

leslie-tsang:
Signed-off-by: leslie <leslie.tsang@icloud.com>

Use the github action tmate instead of the submodule tmate, as tmate only used in chaos workflow.
There is no need to consider it as a submodule of APISIX.
### Pre-submission checklist:


leslie-tsang:
> By Apache's policy, only actions from GitHub or Apache can be used in the CI. So we have to use submodule.

Issue #5149, opened 2021-09-28
request help: use name instead of id in grafana dashboard


starsz:
### Issue description

Hi, I found that now we use the service id and route id in the grafana dashboard.
Which is not friendly to read. So I recommend we can use name instead of id in the grafana dashboard.

Now:
![image](https://user-images.githubusercontent.com/25628854/135027767-62309cd3-589a-48de-baf7-2a39d6c9d55f.png)
![image](https://user-images.githubusercontent.com/25628854/135027792-b6038228-d697-49ef-b861-dec9343a8f7a.png)

### Environment

- apisix version (cmd: `apisix version`): 2.6 
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

spacewander:
The default label of prometheus is id but not name.

starsz:
Yes.I think we can use the name as the label of Prometheus if the name exists?

lifehacking:
@starsz  how solved label name view？

PR #5372, opened 2021-10-29
feat(plugins): Datadog for metrics collection

bisakhmondal:

This PR proves the backbone for integrating Datadog plugin with Apache APISIX.
At present it provides the following metrics:
- request.count - counter
- request.latency - histogram
- upstream.latency - histogram
- apisix.latency - histogram
- ingress.size - timer
- egress.size - timer

To enable datadog plugin via route, just run the following snippet
```json
{
    "name": "my_route",
    "uri": "/hello",
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:80": 1
        }
    },
    "plugins":{
        "datadog":{},
    }
}
```

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -83,7 +83,10 @@ local function generate_tag(conf, const_tags)
         tags = {}
     end
 
-    if conf.route_id and conf.route_id ~= "" then
+    -- priority on route name, if not found using the route id.
+    if conf.route_name ~= "" then
+        core.table.insert(tags, "route_name:" .. conf.route_name)
+    elseif conf.route_id and conf.route_id ~= "" then
         core.table.insert(tags, "route_id:" .. conf.route_id)
     end
 
@@ -100,12 +103,15 @@ local function generate_tag(conf, const_tags)
     if conf.response.status then
         core.table.insert(tags, "response_status:" .. conf.response.status)
     end
-
-    if tags ~= "" then
-        tags = "|#" .. concat(tags, ',')
+    if conf.scheme ~= "" then
+        core.table.insert(tags, "scheme:" .. conf.scheme)
     end
 
-    return tags
+    if #tags > 0 then
+        return "|#" .. concat(tags, ',')
+    end
+
+    return ""
 end
 
 -- remove stale objects from the memory after timer expires
@@ -136,6 +142,8 @@ function _M.log(conf, ctx)
     local entry = fetch_log(ngx, {})
     entry.upstream_latency = ctx.var.upstream_response_time * 1000
     entry.balancer_ip = ctx.balancer_ip or ""
+    entry.route_name = ctx.route_name or ""
+    entry.scheme = ctx.upstream_scheme or ""
 
     local log_buffer = buffers[conf]
     if log_buffer then
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -40,8 +40,8 @@ local schema = {
     properties = {
         buffer_duration = {type = "integer", minimum = 1, default = 60},
         inactive_timeout = {type = "integer", minimum = 1, default = 2},
-        batch_max_size = {type = "integer", minimum = 1, default = 1},
-        max_retry_count = {type = "integer", minimum = 1, default = 3},
+        batch_max_size = {type = "integer", minimum = 1, default = 5000},
+        max_retry_count = {type = "integer", minimum = 1, default = 1},
     }
 }
 
@@ -172,7 +172,7 @@ function _M.log(conf, ctx)
 
         core.log.info("datadog batch_entry: ", core.json.delay_encode(entries, true))
         for _, entry in ipairs(entries) do
-            local suffix = generate_tag(entry, metadata.value.tags)
+            local suffix = generate_tag(entry, metadata.value.constant_tags)
 
             -- request counter
             local ok, err = sock:send(format("%s:%s|%s%s", prefix ..
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -78,7 +78,7 @@ end
 local function generate_tag(conf, const_tags)
     local tags
     if const_tags and #const_tags > 0 then
-        tags = core.table.deepcopy(const_tags)
+        tags = core.table.clone(const_tags)
     else
         tags = {}
     end
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -76,32 +76,33 @@ function _M.check_schema(conf, schema_type)
 end
 
 local function generate_tag(conf, const_tags)
-    local tags = ""
-
+    local tags
     if const_tags and #const_tags > 0 then
-        tags = concat(const_tags, ",") .. ","
+        tags = core.table.deepcopy(const_tags)
+    else
+        tags = {}
     end
 
     if conf.route_id and conf.route_id ~= "" then
-        tags = tags .. "route_id:" .. conf.route_id .. ","
+        core.table.insert(tags, "route_id:" .. conf.route_id)
     end
 
     if conf.service_id and conf.service_id ~= "" then
-        tags = tags .. "service_id:" .. conf.service_id .. ","
+        core.table.insert(tags, "service_id:" .. conf.service_id)
     end
 
     if conf.consumer and conf.consumer ~= "" then
-        tags = tags .. "consumer:" .. conf.consumer .. ","
+        core.table.insert(tags, "consumer:" .. conf.consumer)
     end
     if conf.balancer_ip ~= "" then
-        tags = tags .. "balancer_ip:" .. conf.balancer_ip .. ","
+        core.table.insert(tags, "balancer_ip:" .. conf.balancer_ip)
     end
     if conf.response.status then
-        tags = tags .. "response_status:" .. conf.response.status .. ","
+        core.table.insert(tags, "response_status:" .. conf.response.status)
     end
 
     if tags ~= "" then
-        tags = "|#" .. tags:sub(1, -2)
+        tags = "|#" .. concat(tags, ',')
     end
 
     return tags
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -1,0 +1,175 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local core = require("apisix.core")
+local plugin = require("apisix.plugin")
+local send_statsd = require("apisix.plugins.udp-logger").send_udp_data
+local fetch_info = require("apisix.plugins.prometheus.exporter").parse_info_from_ctx
+local format = string.format
+local concat = table.concat
+local tostring = tostring
+local ngx = ngx
+
+
+local plugin_name = "datadog"
+
+local schema = {
+    type = "object",
+    properties = {
+        sample_rate = {type = "number", default = 1, minimum = 0, maximum = 1},
+        tags = {
+            type = "array",
+            items = {type = "string"},
+            default = {"source:apisix"}
+        }
+    }
+}
+
+local metadata_schema = {
+    type = "object",
+    properties = {
+        host = {type = "string", default = "0.0.0.0"},
+        port = {type = "integer", minimum = 0, default = 8125},
+        namespace = {type = "string", default = "apisix.dev"},
+    },
+}
+
+local _M = {
+    version = 0.1,
+    priority = 495,
+    name = plugin_name,
+    schema = schema,
+    metadata_schema = metadata_schema,
+}
+
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
+end
+
+local function generate_tag(sample_rate, tag_arr, route_id, service_id, consumer_name, balancer_ip)
+    local rate, tags = "", ""
+
+    if sample_rate and sample_rate ~= 1 then
+        rate = "|@" .. tostring(sample_rate)
+    end
+
+    if tag_arr and #tag_arr > 0 then
+        tags = "|#" .. concat(tag_arr, ",")
+    end
+
+    if route_id ~= "" then
+        tags = tags .. "route_id:" .. route_id
+    end
+
+    if service_id ~= "" then
+        tags = tags .. "service_id:" .. service_id
+    end
+
+    if consumer_name ~= "" then
+        tags = tags .. "consumer_name:" .. consumer_name
+    end
+    if balancer_ip ~= "" then
+        tags = tags .. "balancer_ip:" .. balancer_ip
+    end
+
+    if tags ~= "" and tags:sub(1, 1) ~= "|" then
+        tags = "|#" .. tags
+    end
+
+    return rate .. tags
+
+end
+
+function _M.log(conf, ctx)
+    core.log.error("conf: ", core.json.delay_encode(conf, true))
+    core.log.error("ctx: ", core.json.delay_encode(ctx, true))
+    local metadata = plugin.plugin_metadata(plugin_name)
+    if not metadata then
+        core.log.error("received nil metadata")
+    end
+
+    local udp_conf = {
+        host = metadata.value.host,
+        port = metadata.value.port
+    }
+
+    local route_id, service_id, consumer_name, balancer_ip = fetch_info(conf, ctx)
+    local prefix = ""
+
+    if metadata.value.namespace ~= "" then
+        prefix = prefix .. "."
+    end
+
+    local suffix = generate_tag(conf.sample_rate, conf.tags,
+                    route_id, service_id, consumer_name, balancer_ip)
+
+    -- request counter
+    local ok, err = send_statsd(udp_conf,
+                        format("%s:%s|%s%s", prefix .. "request.counter", 1, "c", suffix))
+    if not ok then
+        core.log.error("failed to send request_count metric to DogStatsD. err: " .. err)
+    end
+
+
+    -- request latency histogram
+    local latency = (ngx.now() - ngx.req.start_time()) * 1000
+    local ok, err = send_statsd(udp_conf,
+                        format("%s:%s|%s%s", prefix .. "request.latency", latency, "h", suffix))
+    if not ok then
+        core.log.error("failed to send request latency metric to DogStatsD. err: " .. err)
+    end
+
+    -- upstream latency
+    local apisix_latency = latency
+    if ctx.var.upstream_response_time then
+        local upstream_latency = ctx.var.upstream_response_time * 1000
+        local ok, err = send_statsd(udp_conf,
+                format("%s:%s|%s%s", prefix .. "upstream.latency", upstream_latency, "h", suffix))
+        if not ok then
+            core.log.error("failed to send upstream latency metric to DogStatsD. err: " .. err)
+        end
+        apisix_latency =  apisix_latency - upstream_latency
+        if apisix_latency < 0 then
+            apisix_latency = 0
+        end
+    end
+
+    -- apisix_latency
+    local ok, err = send_statsd(udp_conf,
+            format("%s:%s|%s%s", prefix .. "apisix.latency", apisix_latency, "h", suffix))
+    if not ok then
+        core.log.error("failed to send apisix latency metric to DogStatsD. err: " .. err)
+    end
+
+    -- request body size timer
+    local ok, err = send_statsd(udp_conf,
+            format("%s:%s|%s%s", prefix .. "ingress.size", ctx.var.request_length, "ms", suffix))
+    if not ok then
+        core.log.error("failed to send request body size metric to DogStatsD. err: " .. err)
+    end
+
+    -- response body size timer
+    local ok, err = send_statsd(udp_conf,
+            format("%s:%s|%s%s", prefix .. "egress.size", ctx.var.bytes_sent, "ms", suffix))
+    if not ok then
+        core.log.error("failed to send response body size metric to DogStatsD. err: " .. err)
+    end
+end
+
+return _M
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -16,36 +16,47 @@
 
 local core = require("apisix.core")
 local plugin = require("apisix.plugin")
-local send_statsd = require("apisix.plugins.udp-logger").send_udp_data
-local fetch_info = require("apisix.plugins.prometheus.exporter").parse_info_from_ctx
+local batch_processor = require("apisix.utils.batch-processor")
+local fetch_log = require("apisix.utils.log-util").get_full_log
+local ngx = ngx
+local udp = ngx.socket.udp
 local format = string.format
 local concat = table.concat
-local tostring = tostring
-local ngx = ngx
-
+local buffers = {}
+local ipairs = ipairs
+local stale_timer_running = false
+local timer_at = ngx.timer.at
 
 local plugin_name = "datadog"
+local defaults = {
+    host = "127.0.0.1",
+    port = 8125,
+    namespace = "apisix",
+    constant_tags = {"source:apisix"}
+}
 
 local schema = {
     type = "object",
     properties = {
+        buffer_duration = {type = "integer", minimum = 1, default = 60},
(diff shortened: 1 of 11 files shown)


bisakhmondal:
CI error not related to this PR: https://github.com/apache/apisix/runs/4064450756?check_suite_focus=true#step:14:680

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+    properties = {
+    }
+}
+
+local metadata_schema = {
+    type = "object",
+    properties = {
+        host = {type = "string"},
+        port = {type = "integer", minimum = 0},
+        namespace = {type = "string", default = "apisix.dev"},
```
spacewander, line 40:
```suggestion
        namespace = {type = "string", default = "apisix"},
```

Would be better. The idea given in the mail list is just an example...

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local core = require("apisix.core")
+local plugin = require("apisix.plugin")
+local send_statsd = require("apisix.plugins.udp-logger").send_udp_data
```
spacewander, line 19: We should not require the main module of another plugin, which will have a side effect: the plugin will be enabled once the datadog is enabled.

Considering the send_udp_data is small, we can copy it directly.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+local metadata_schema = {
+    type = "object",
+    properties = {
+        host = {type = "string"},
+        port = {type = "integer", minimum = 0},
+        namespace = {type = "string", default = "apisix.dev"},
+        sample_rate = {type = "number", default = 1, minimum = 0, maximum = 1},
```
spacewander, line 41: I think we don't need to implement sample_rate as this version, otherwise, we need to sample the request in the given rate.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+        tags = {
```
spacewander, line 42: Use name constant_tags would be better?

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+            type = "array",
+            items = {type = "string"},
+            default = {"source:apisix"}
+        }
+    },
+    required = {"host", "port"}
```
spacewander, line 48: If we provide a default value, we don't need to require the user to configure the metadata before using it. Considering statsd is usually running as a sidecar, we can use "127.0.0.1" as the default host and 8125 as the default port.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,178 @@
...
+        tags = "|#" .. tags
+    end
+
+    return rate .. tags
+
+end
+
+function _M.log(conf, ctx)
+    local metadata = plugin.plugin_metadata(plugin_name)
+    if not metadata then
```
spacewander, line 106: A more appreciated way is to prove a couple of the default values when metadata is missing. See https://github.com/apache/apisix/blob/759e6663162d43e761d556b85e3061f755ea74ae/apisix/plugins/batch-requests.lua#L213

apisix/plugins/prometheus/exporter.lua
```diff
@@ -131,6 +128,13 @@ function _M.log(conf, ctx)
             end
         end
     end
+    return route_id, service_id, consumer_name, balancer_ip
+end
+
+function _M.log(conf, ctx)
+    local vars = ctx.var
+
+    local route_id, service_id, consumer_name, balancer_ip = _M.parse_info_from_ctx(conf, ctx)
```
spacewander, line 137: For code style, we don't call a module level method inside another module level method. We use this:
```
local function fetch_req_info(conf, ctx)
end

_M.fetch_req_info = fetch_req_info
```

apisix/plugins/prometheus/exporter.lua
```diff
@@ -110,10 +110,7 @@ function _M.init()
 
 end
 
-
-function _M.log(conf, ctx)
-    local vars = ctx.var
-
+function _M.parse_info_from_ctx(conf, ctx)
```
spacewander, line 113:
```suggestion
function _M.fetch_req_info(conf, ctx)
```
would be better

t/plugin/datadog.t
```diff
@@ -0,0 +1,156 @@
...
+GET /t
+--- response_body
+property "host" is required
+done
+--- no_error_log
+[error]
+
+
+
+=== TEST 3: add plugin
```
spacewander, line 71: We need a way to test it with a backend. It would be great if you find a way to run dogstatsd locally and check the content it receives (let it report to a mock HTTP datadog backend, etc.). Another solution is writing a mock UDP backend, see https://github.com/apache/apisix/blob/759e6663162d43e761d556b85e3061f755ea74ae/t/plugin/ext-plugin/sanity.t#L34 and https://github.com/openresty/stream-lua-nginx-module

apisix/plugins/datadog.lua
bisakhmondal: :+1:

apisix/plugins/datadog.lua
bisakhmondal: Oh, I see. Definitely.

apisix/plugins/datadog.lua
bisakhmondal: okay

apisix/plugins/datadog.lua
bisakhmondal: I wrote it earlier, like the way you mentioned. But later I thought to remove it, as people may easily forget to update the meta if they are running statsd on a different port.
Anyway, updating accordingly.

apisix/plugins/prometheus/exporter.lua
bisakhmondal: Okay, Thanks.

t/plugin/datadog.t
bisakhmondal: Great idea. Let's see

t/plugin/datadog.t
```diff
@@ -0,0 +1,156 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
```
spacewander, line 19: We can avoid repeating sections by setting this:
```
add_block_preprocessor(sub {
    my ($block) = @_;

    if (!$block->request) {
    }

    if (!$block->error_log) {
    }
}
```

bisakhmondal: Okay, we can do that.

t/plugin/datadog.t
bisakhmondal: Done

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,268 @@
...
+
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
+end
+
+local function generate_tag(conf, const_tags)
+    local tags = ""
```
spacewander, line 79: Is it faster if we can use table.concat to concat the string instead?

bisakhmondal: You are right, 
> Lua strings are immutable

effectively it will be O(N^2) instead of O(N) with a table. updating accordingly

t/plugin/datadog.t
```diff
@@ -0,0 +1,187 @@
...
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:1982": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/opentracing"
+                }]],
+                [[{
```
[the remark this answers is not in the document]
bisakhmondal: is it because there are already separate tests for 'route' entities? I saw internally it perform table comparison and return "passed" or "fail" according to that - which could be used to check if the route creation was a success or not. I guess a better approach than checking error code < 300. Let me know what you think, I am okay with both of them.

t/plugin/datadog.t
```diff
@@ -0,0 +1,187 @@
...
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+
+            local code, _, body = t("/opentracing", "GET")
+             if code >= 300 then
+                ngx.status = code
+                ngx.say("fail")
+                return
+            end
+            ngx.say(body:sub(1, -2))
```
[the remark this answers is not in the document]
bisakhmondal: Nice, the subtle difference between ngx.say and ngx.print

t/plugin/datadog.t
```diff
@@ -0,0 +1,187 @@
...
+run_tests;
+
+__DATA__
+
+=== TEST 1: sanity check metadata
+--- config
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.datadog")
+            local ok, err = plugin.check_schema({host = "127.0.0.1", port = 8125}, 2)
```
spacewander, line 58:
```suggestion
            local ok, err = plugin.check_schema({host = "127.0.0.1", port = 8125}, core.schema.TYPE_METADATA)
```

t/plugin/datadog.t
```diff
@@ -0,0 +1,187 @@
...
+            ngx.say(body:sub(1, -2))
+        }
+    }
+--- wait: 3
+--- response_body
+opentracing
+--- grep_error_log eval
+qr/message received: apisix([.\w]+)/
+--- grep_error_log_out
+message received: apisix.request.counter
```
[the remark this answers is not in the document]
bisakhmondal: Done.

t/plugin/datadog.t
```diff
@@ -0,0 +1,187 @@
...
+message received: apisix.request.latency
+message received: apisix.upstream.latency
+message received: apisix.apisix.latency
+message received: apisix.ingress.size
+message received: apisix.egress.size
```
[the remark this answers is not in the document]
bisakhmondal: Done

apisix/plugins/datadog.lua
bisakhmondal: Optimized

t/plugin/datadog.t
bisakhmondal: Done.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,269 @@
...
+    if conf.response.status then
+        core.table.insert(tags, "response_status:" .. conf.response.status)
+    end
+
+    if tags ~= "" then
+        tags = "|#" .. concat(tags, ',')
+    end
+
+    return tags
+end
```
bisakhmondal, line 109:
```suggestion
    if #tags > 0 then
        return "|#" .. concat(tags, ',')
    end

    return ""
end
```

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,269 @@
...
+local function generate_tag(conf, const_tags)
+    local tags
+    if const_tags and #const_tags > 0 then
+        tags = core.table.deepcopy(const_tags)
+    else
+        tags = {}
+    end
+
+    if conf.route_id and conf.route_id ~= "" then
+        core.table.insert(tags, "route_id:" .. conf.route_id)
```
spacewander, line 87:
```
idx = #tags + 1
...
tags[idx] = "route_id:"
tags[idx + 1] = conf.route_id
idx = idx + 2
```
might be faster. You can give it a try.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,269 @@
...
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
+end
+
+local function generate_tag(conf, const_tags)
+    local tags
+    if const_tags and #const_tags > 0 then
+        tags = core.table.deepcopy(const_tags)
```
spacewander, line 81:
```suggestion
        tags = core.table.clone(const_tags)
```
is enough for the one-level tags?

apisix/plugins/datadog.lua
bisakhmondal: >tags[idx] = "route_id:"
tags[idx + 1] = conf.route_id
idx = idx + 2



The required format is comma-separated key-value pair. where a particular key and value pair is separated by a colon ( `:` ). This idx + 2 won't be able to achieve that.

TBH, the insert is a perf improvement that `tags = tags .. ""blah"` as in this case the whole tags get copied again and again. This is not the case with table.insert("key:", val), it's just O(1) and the concat is O(N) so effectively O(N).

Asymptotically, both the idx approach that you suggested and the insert approach that we have currently is O(N). But I am not sure if insert has a little overhead than tags[idx]. Please let me know, I'll be happy to update it.

the updated approach will be
tags[idx] = "route_id:"  .. conf.route_id
idx = idx + 1
..

apisix/plugins/datadog.lua
bisakhmondal: Make sense : )

t/plugin/datadog.t
```diff
@@ -0,0 +1,366 @@
...
+            -- Change the metadata
+            local code, meta_body = t('/apisix/admin/plugin_metadata/datadog',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "host":"127.0.0.1",
+                        "port": 8125,
+                        "namespace": "mycompany"
+                }]],
+                [[{
+                    "node": {
```
spacewander, line 257: We don't need to check the response of the plugin_metadata API? It has already been checked before.

t/plugin/datadog.t
```diff
@@ -0,0 +1,366 @@
...
+                 [[{
+                        "host":"127.0.0.1",
+                        "port": 8125,
+                        "constant_tags": [
+                                "source:apisix",
+                                "new_tag:must"
+                            ]
+                }]],
+                [[{
+                    "node": {
```
spacewander, line 322: Ditto

t/plugin/datadog.t
```diff
@@ -0,0 +1,366 @@
...
+message received: mycompany\.request\.latency:[\d.]+\|h\|#source:apisix,route_name:datadog,balancer_ip:[\d.]+,response_status:200,scheme:http
+message received: mycompany\.upstream\.latency:[\d.]+\|h\|#source:apisix,route_name:datadog,balancer_ip:[\d.]+,response_status:200,scheme:http
+message received: mycompany\.apisix\.latency:[\d.]+\|h\|#source:apisix,route_name:datadog,balancer_ip:[\d.]+,response_status:200,scheme:http
+message received: mycompany\.ingress\.size:[\d]+\|ms\|#source:apisix,route_name:datadog,balancer_ip:[\d.]+,response_status:200,scheme:http
+message received: mycompany\.egress\.size:[\d]+\|ms\|#source:apisix,route_name:datadog,balancer_ip:[\d.]+,response_status:200,scheme:http
+/
+
+
+
+=== TEST 6: testing behaviour with different namespace
```
spacewander, line 304:
```suggestion
=== TEST 6: testing behaviour with different tags
```

docs/en/latest/plugins/datadog.md
```diff
@@ -0,0 +1,128 @@
...
+
+To know more about how to effectively write tags, please visit [here](https://docs.datadoghq.com/getting_started/tagging/#defining-tags)
+
+## Exported Metrics
+
+Apache APISIX agent, for every request response cycle, export the following metrics to DogStatsD server if the datadog plugin is enabled:
+
+| Metric Name               | StatsD Type   | Description               |
+| -----------               | -----------   | -------                   |
+| Request Counter           | Counter       | No of requests received.   |
```
spacewander, line 72: The value of counter is not the number, but always 1?

docs/en/latest/plugins/datadog.md
```diff
@@ -0,0 +1,128 @@
...
+| Request Latency           | Histogram     | Time taken to process the request (in milliseconds). |
+| Upstream latency          | Histogram     | Time taken to proxy the request to the upstream server till a response is received (in milliseconds). |
+| APISIX Latency            | Histogram     | Time taken by APISIX agent to process the request (in milliseconds). |
+| Ingress Size              | Timer         | Request body size in bytes. |
+| Egress Size               | Timer         | Response body size in bytes. |
+
```
spacewander, line 78: Let's doc the metric tag too

apisix/plugins/datadog.lua
spacewander: The `table.insert` without idx will need to calculate the length of the table first (which is not an O(1) operation). The current behavior is fine as the optimization may be too trivial.
See:
https://github.com/openresty/luajit2/blob/cd2285f1b7792ccf61db37c10da7463b6b1a30a5/src/lib_table.c#L157
https://github.com/openresty/luajit2/blob/da4aecf2e79adaed6f6ec9726e330a77816efbe3/src/lj_tab.c#L663

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,277 @@
...
+    if not stale_timer_running then
+        -- run the timer every 30 mins if any log is present
+        timer_at(1800, remove_stale_objects)
+        stale_timer_running = true
+    end
+
+    local entry = fetch_log(ngx, {})
+    entry.upstream_latency = ctx.var.upstream_response_time * 1000
+    entry.balancer_ip = ctx.balancer_ip or ""
+    entry.route_name = ctx.route_name or ""
```
spacewander, line 145: We should fall back to route_id if the name is missing. And need a test for it.

bisakhmondal: Okay. The code block for fallback logic is at L86
Adding tests.

t/plugin/datadog.t
bisakhmondal: Ack. Removing.

docs/en/latest/plugins/datadog.md
bisakhmondal: We are pushing the "delta" to dogstatsd which will be aggregated at dogstatsd agent and will be sent to the datadog server. In our case, as it's a single request-response cycle, the value is +1 (increment) i.e. one request has been served. 
Are you asking something else, could get you totally?

apisix/plugins/datadog.lua
bisakhmondal: The code block for fallback logic is at L86
Okay, adding tests.

docs/en/latest/plugins/datadog.md
bisakhmondal: Do you mean how to write/populate "constant_tags"?
As of now, for tags, we have only L64. Please, do let me know. Thanks : )

docs/en/latest/plugins/datadog.md
spacewander: So you document it at the view of the datadog server. I see.

docs/en/latest/plugins/datadog.md
spacewander: No, I mean the tags for each metric, like the route_name.

apisix/plugins/datadog.lua
spacewander: I think we can use route_name even there is only route_id? Therefore the name of tag is consistent between routes.

docs/en/latest/plugins/datadog.md
```diff
@@ -0,0 +1,128 @@
...
+## How To Enable
+
+First of all, enable the datadog plugin in the `config.yaml`:
+
+```
+# Add this in config.yaml
+plugins:
+  - ... # plugin you need
+  - datadog
+```
```
spacewander, line 88: We don't need to do it. The datadog is enabled by default.

apisix/plugins/datadog.lua
```diff
@@ -0,0 +1,277 @@
...
+}
+
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
+end
+
+local function generate_tag(conf, const_tags)
```
spacewander, line 78: I think we can rename the `conf` to `entry` as it is not a configuration.

bisakhmondal: Updated

docs/en/latest/plugins/datadog.md
bisakhmondal: You are right, I have added the entry in default.yml.
Removed.

apisix/plugins/datadog.lua
bisakhmondal: Perfectly makes sense, Updated accordingly. Thanks for pointing it out.

docs/en/latest/plugins/datadog.md
bisakhmondal: Added

spacewander (approved):

apisix/plugins/datadog.lua
tzssangglass: @spacewander ecommend using route name, won't the same routes name be confusing?

bisakhmondal: Hi! @tzssangglass thanks for the honest concern. Yeah, in a badly designed system, it might be an issue to segregate metrics of multiple routes having the same route_name. But users can always pick multiple tags for grouping (for eg. route_name:abc, service_id:12), provided they are designing APISIX infra with service level abstraction (just one example, they can use "balancer_ip, consumer" also for filtering). 

@spacewander recommended using name over id because it's more readable to the user and there was this requirement in the Prometheus plugin (https://github.com/apache/apisix/issues/5149). Do you think that we should add a param `prefer_name` at the plugin schema or we should defer it for the upcoming versions? 
Please do let me know. Thanks.

tzssangglass: I get. I think it is up to the community to decide if `prefer_name` is needed like the prometheus plugin.I suggest adding a note about that  it could be misleading when route name could be duplicated for multiple.

docs/en/latest/plugins/datadog.md
```diff
@@ -0,0 +1,131 @@
...
+| APISIX Latency            | Histogram     | Time taken by APISIX agent to process the request (in milliseconds). |
+| Ingress Size              | Timer         | Request body size in bytes. |
+| Egress Size               | Timer         | Response body size in bytes. |
+
+The metrics will be sent to the DogStatsD agent with the following tags:
+
+> If there is no suitable value for any particular tag, the tag will simply be omitted.
+
+- **route_name**: Name specified in the route schema definition. If not present, it will fall back to the route id value.
+  - Note: If multiple routes have the same name duplicated, we suggest you to visualize graphs on the Datadog dashboard over multiple tags that could compositely pinpoint a particular route/service. If it's still insufficient for your needs, feel free to drop a feature request at [apisix/issues](https://github.com/apache/apisix/issues).
```
bisakhmondal, line 84: @tzssangglass I hope this would be sufficient. WDYT?

tzssangglass (approved):

docs/en/latest/plugins/datadog.md
```diff
@@ -0,0 +1,131 @@
...
+| buffer_duration  | integer | optional    | 60           | [1,...] | Maximum age in seconds of the oldest entry in a batch before the batch must be processed |
+| max_retry_count  | integer | optional    | 1            | [1,...] | Maximum number of retries if one entry fails to reach dogstatsd server                   |
+
+## Metadata
+
+| Name        | Type    | Requirement |     Default        | Valid         | Description                                                            |
+| ----------- | ------  | ----------- |      -------       | -----         | ---------------------------------------------------------------------- |
+| host        | string  | optional    |  "127.0.0.1"       |               | The DogStatsD server host address                                      |
+| port        | integer | optional    |    8125            |               | The DogStatsD server host port                                         |
+| namespace   | string  | optional    |    "apisix"        |               | Prefix for all the custom metrics sent by APISIX agent. Useful for finding entities for metric graph. e.g. (apisix.request.counter)                                        |
```
moonming, line 61: @bisakhmondal please add doc and examples for how to configure `host` and `port`, thanks

moonming: I means how to set the datadog endpoint

juzhiyuan: Yes, take this as an example: If I'm a newbie to use Apache APISIX and the datadog plugin, I'd like to know all the necessary steps after Apache APISIX runs successfully.

Like when I should update its metadata, and how I should setup its data with a Route.

bisakhmondal: Got your point. Thanks for the elaborate explanation.
The plugin has a default metadata schema that assumes dogstatsd is available at localhost:8125. If someone needs to reconfigure that I have covered that on the blog which is soon going to be exported at apisix blog
Here is the link: https://blog.bisakh.com/blog/Monitoring-with-Datadog-in-Apache-APISIX

Do you think, we should add it here (the `Steps to Run Datadog Agent` section & `Custom Configuration`)? Thanks guys

juzhiyuan: Ya, I have read your blog[1], I think the `Custom Configuration` part is what @moonming and I need 😄

image

[1] https://blog.bisakh.com/blog/Monitoring-with-Datadog-in-Apache-APISIX

bisakhmondal: Cool Cool, I'll open a PR with the code snippet and some extra details. Thanks, @juzhiyuan @moonming

PR #5378, opened 2021-10-30
feat(limit-count): support multiple variables as key

Xunzhuo:

feat(limit-count): support multiple variables as key

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/limit-count.lua
+++ b/apisix/plugins/limit-count.lua
@@ -36,11 +36,10 @@ local schema = {
     properties = {
         count = {type = "integer", exclusiveMinimum = 0},
         time_window = {type = "integer",  exclusiveMinimum = 0},
-        key = {
-            type = "string",
-            enum = {"remote_addr", "server_addr", "http_x_real_ip",
-                    "http_x_forwarded_for", "consumer_name", "service_id"},
-            default = "remote_addr",
+        key = {type = "string", default = "remote_addr"},
+        key_type = {type = "string",
+            enum = {"var", "var_combination"},
+            default = "var",
         },
         rejected_code = {
             type = "integer", minimum = 200, maximum = 599, default = 503
@@ -171,7 +170,29 @@ function _M.access(conf, ctx)
         return 500
     end
 
-    local key = (ctx.var[conf.key] or "") .. ctx.conf_type .. ctx.conf_version
+    local conf_key = conf.key
+    local key
+    if conf.key_type == "var_combination" then
+        local err, n_resolved
+        key, err, n_resolved = core.utils.resolve_var(conf_key, ctx.var);
+        if err then
+            core.log.error("could not resolve vars in ", conf_key, " error: ", err)
+        end
+
+        if n_resolved == 0 then
+            key = nil
+        end
+    else
+        key = (ctx.var[conf_key] or "")
+    end
+
+    if key == nil then
+        core.log.info("bypass the limit count as the key is empty")
+        -- Bypass the limit count when the key is empty.
+        -- This behavior is the same as Nginx
+        return
+    end
+    key = key .. ctx.conf_type .. ctx.conf_version
     core.log.info("limit key: ", key)
 
     local delay, remaining = lim:incoming(key, true)
--- a/apisix/plugins/limit-count.lua
+++ b/apisix/plugins/limit-count.lua
@@ -183,7 +183,7 @@ function _M.access(conf, ctx)
             key = nil
         end
     else
-        key = (ctx.var[conf_key] or "")
+        key = ctx.var[conf_key]
     end
 
     if key == nil then
--- a/docs/en/latest/plugins/limit-count.md
+++ b/docs/en/latest/plugins/limit-count.md
@@ -39,7 +39,8 @@ Limit request rate by a fixed number of requests in a given time window.
 | ------------------- | ------- | --------------------------------------- | ------------- | ------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | count               | integer | required                                |               | count > 0                                                                                               | the specified number of requests threshold.                                                                                                                                                                                                                                                                |
 | time_window         | integer | required                    |               | time_window > 0                                                                                         | the time window in seconds before the request count is reset.                                                                                                                                                                                                                                              |
-| key                 | string  | optional                                | "remote_addr" | ["remote_addr", "server_addr", "http_x_real_ip", "http_x_forwarded_for", "consumer_name", "service_id"] | The user specified key to limit the count. <br /> Now accept those as key: "remote_addr"(client's IP), "server_addr"(server's IP), "X-Forwarded-For/X-Real-IP" in request header, "consumer_name"(consumer's username) and "service_id".                                                                   |
+| key_type      | string  | optional    |   "var"   | ["var", "var_combination"] | the type of key. |
+| key           | string  | optional    |     "remote_addr"    |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable. If the `key_type` is "var_combination", the key will be a combination of variables. For example, if we use "$remote_addr $consumer_name" as keys, plugin will be restricted by two keys which are "remote_addr" and "consumer_name". |
 | rejected_code       | integer | optional                                | 503           | [200,...,599]                                                                                           | The HTTP status code returned when the request exceeds the threshold is rejected, default 503.                                                                                                                                                                                                             |
 | rejected_msg       | string | optional                                |            | non-empty                                                                                           | The response body returned when the request exceeds the threshold is rejected.                                                                                                                                                                                                             |
 | policy              | string  | optional                                | "local"       | ["local", "redis", "redis-cluster"]                                                                     | The rate-limiting policies to use for retrieving and incrementing the limits. Available values are `local`(the counters will be stored locally in-memory on the node), `redis`(counters are stored on a Redis server and will be shared across the nodes, usually use it to do the global speed limit), and `redis-cluster` which works the same as `redis` but with redis cluster. |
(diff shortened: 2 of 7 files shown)


apisix/plugins/limit-count.lua
```diff
@@ -36,11 +36,10 @@ local schema = {
...
         count = {type = "integer", exclusiveMinimum = 0},
         time_window = {type = "integer",  exclusiveMinimum = 0},
-        key = {
-            type = "string",
-            enum = {"remote_addr", "server_addr", "http_x_real_ip",
-                    "http_x_forwarded_for", "consumer_name", "service_id"},
-            default = "remote_addr",
+        key = {type = "string", default = "remote_addr"},
+        key_type = {type = "string",
+            enum = {"var", "var_combination"},
```
tokers, line 41: It looks like the constraints for `var` losing.

Xunzhuo:
cc @spacewander ,please take a look thanks. CI failed by other test case which is not related to limit-count

t/plugin/limit-count2.t
```diff
@@ -178,3 +178,137 @@ GET /hello
...
+    location /t {
+        content_by_lua_block {
+            local json = require "t.toolkit.json"
+            local http = require "resty.http"
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port
+                        .. "/hello"
+            local ress = {}
+            for i = 1, 2 do
+                local httpc = http.new()
+                local res, err = httpc:request_uri(uri, {headers = {a = 2}})
```
spacewander, line 267: What's the difference between TEST 7/8? Just because TEST 8 iterates 2 steps?

Xunzhuo: What's more, TEST 8 use the other key to count

spacewander: I think if we add two more steps, the result of TEST 8 will be the same as TEST 7?

Xunzhuo: @spacewander Yes it is, so should I remove this case or keep it? Actually I take [this](https://github.com/apache/apisix/pull/5302/files#diff-7645485edca150830424fd3151413a2eddb93d1dcea7fe694ce458113a6b3bf3) as an example.

spacewander: It is:
```
local res, err = httpc:request_uri(uri, {headers = {a = i}})
```
Not `a = 2`. And you should test it three times to trigger the limit exceed.

docs/en/latest/plugins/limit-count.md
```diff
@@ -39,7 +39,8 @@ Limit request rate by a fixed number of requests in a given time window.
...
-| key                 | string  | optional                                | "remote_addr" | ["remote_addr", "server_addr", "http_x_real_ip", "http_x_forwarded_for", "consumer_name", "service_id"] | The user specified key to limit the count. <br /> Now accept those as key: "remote_addr"(client's IP), "server_addr"(server's IP), "X-Forwarded-For/X-Real-IP" in request header, "consumer_name"(consumer's username) and "service_id".                                                                   |
+| key_type      | string  | optional    |   "var"   | ["var", "var_combination"] | the type of key. |
+| key           | string  | optional    |     "remote_addr"    |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable, like "remote_addr" or "consumer_name". If the `key_type` is "var_combination", the key will be a combination of variables, like "$remote_addr $consumer_name". |
```
moonming, line 43: we need more examples for `var_combination`, so that developers know how to use

apisix/plugins/limit-count.lua
Xunzhuo: Thanks @tokers , if we keep the enum constraints for var like before, it will be a problem to use var_combination, because var_combination provides a more flexible way to key selecting.

docs/en/latest/plugins/limit-count.md
Xunzhuo: Good point~ Thanks @moonming.

spacewander (approved):

apisix/plugins/limit-count.lua
tokers: So it's my understanding that a potential problem will be introduced here, users might configure a non-existent variable and the validation is OK, but actually, the rate-limiting won't be the effect?

I think maybe some documents should be added so that at least we can give some hints to users.

Xunzhuo: Thanks @tokers  I am working on it, update it later~

tokers: OK, I saw the changes.

tokers (approved):

apisix/plugins/limit-count.lua
```diff
@@ -171,7 +170,29 @@ function _M.access(conf, ctx)
...
+        key, err, n_resolved = core.utils.resolve_var(conf_key, ctx.var);
+        if err then
+            core.log.error("could not resolve vars in ", conf_key, " error: ", err)
+        end
+
+        if n_resolved == 0 then
+            key = nil
+        end
+    else
+        key = ctx.var[conf_key]
```
Xunzhuo, line 186: CI failure occurred, I wonder if I should use 'key = ( ctx.var[conf_key] or "")' instead of the current way. cc @spacewander

apisix/plugins/limit-count.lua
moonming: please add test cases for non-existent variable @Xunzhuo

apisix/plugins/limit-count.lua
moonming: if `ctx.var[conf_key]` is `nil`, what is the result you expect?

Xunzhuo: Yes, I got it, thanks~

docs/en/latest/plugins/limit-count.md
```diff
@@ -68,13 +69,38 @@ curl -i http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335
...
         "type": "roundrobin",
         "nodes": {
-            "39.97.63.215:80": 1
+            "127.0.0.1:9001": 1
+        }
+    }
+}'
+```
+
+Here's an example, enable the `limit count` plugin on the specified route and when setting `key_type` to `var_combination` :
```
spacewander, line 85:
```suggestion
Here's an example, enable the `limit count` plugin on the specified route when setting `key_type` to `var_combination` :
```

docs/en/latest/plugins/limit-count.md
```diff
@@ -39,7 +39,8 @@ Limit request rate by a fixed number of requests in a given time window.
...
+| key_type      | string  | optional    |   "var"   | ["var", "var_combination"] | the type of key. |
+| key           | string  | optional    |     "remote_addr"    |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable. <br />Now accept those as key: "remote_addr"(client's IP), "server_addr"(server's IP), "X-Forwarded-For/X-Real-IP" in request header, "consumer_name"(consumer's username) and "service_id". <br /> If the `key_type` is "var_combination", the key will be a combination of variables. For example,if we use "$remote_addr $consumer_name" as keys, plugin will be restricted by two keys which are "remote_addr" and "consumer_name". |
```
spacewander, line 43:
```suggestion
| key           | string  | optional    |     "remote_addr"    |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable. If the `key_type` is "var_combination", the key will be a combination of variables. For example, if we use "$remote_addr $consumer_name" as keys, plugin will be restricted by two keys which are "remote_addr" and "consumer_name". |
```

Any variable can be used as key.

docs/zh/latest/plugins/limit-count.md
```diff
@@ -42,7 +42,8 @@ title: limit-count
...
-| key                 | string  | 可选                               | "remote_addr" | ["remote_addr", "server_addr", "http_x_real_ip", "http_x_forwarded_for", "consumer_name", "service_id"] | 用来做请求计数的有效值。<br />例如，可以使用主机名（或服务器区域）作为关键字，以便限制每个主机名规定时间内的请求次数。我们也可以使用客户端地址作为关键字，这样我们就可以避免单个客户端规定时间内多次的连接我们的服务。<br />当前接受的 key 有："remote_addr"（客户端 IP 地址）, "server_addr"（服务端 IP 地址）, 请求头中的"X-Forwarded-For" 或 "X-Real-IP", "consumer_name"（consumer 的 username）, "service_id" 。 |
+| key_type      | string | 可选   |  "var"      | ["var", "var_combination"]                                          | key 的类型 |
+| key           | string  | 可选   |    "remote_addr"    |  | 用来做请求计数的依据。如果 `key_type` 为 "var"，那么 key 会被当作变量名称。<br />当前接受的 key 有："remote_addr"（客户端 IP 地址）, "server_addr"（服务端 IP 地址）, 请求头中的"X-Forwarded-For" 或 "X-Real-IP", "consumer_name"（consumer 的 username）, "service_id" 。<br />如果 `key_type` 为 "var_combination"，那么 key 会当作变量组。比如设置 "$remote_addr $consumer_name" 作为 keys，那么插件会同时受 remote_addr 和 consumer_name 两个 key 的约束。 |
```
spacewander, line 46: Ditto

Xunzhuo: @spacewander Done, Thanks~

docs/en/latest/plugins/limit-count.md
Xunzhuo: @spacewander Done thanks~

spacewander (approved):

apisix/plugins/limit-count.lua
```diff
@@ -171,7 +170,29 @@ function _M.access(conf, ctx)
...
+        key, err, n_resolved = core.utils.resolve_var(conf_key, ctx.var);
+        if err then
+            core.log.error("could not resolve vars in ", conf_key, " error: ", err)
+        end
+
+        if n_resolved == 0 then
+            key = nil
+        end
+    else
+        key = (ctx.var[conf_key] or "")
```
spacewander, line 186: When key_type is var, the key will never be nil. Could you add a test to cover it?

Xunzhuo: @spacewander Done and add more test cases when `key_type` is `var`

t/plugin/limit-count-redis-cluster.t
```diff
@@ -328,7 +328,7 @@ code: 200
                             "limit-count": {
                                 "count": ]] .. count .. [[,
                                 "time_window": 60,
-                                "key": "http_x_real_ip",
+                                "key": "remote_addr",
```
spacewander, line 331: Can we change the relative tests to use remote_addr too?

Xunzhuo: yes of course.

PR #5385, opened 2021-11-01
fix: ldap deps in `install-dependencies.sh`

leslie-tsang:
Signed-off-by: leslie <leslie.tsang@icloud.com>

* use `openldap-devel` instead of `libldap2-dev`,  `libldap2-dev` don't exist on centos
* fix file lint in `t/plugin/ldap-auth.t`, add new line at the end of file

### Pre-submission checklist:


utils/install-dependencies.sh
```diff
@@ -24,10 +24,9 @@ function install_dependencies_with_yum() {
     # add OpenResty source
     sudo yum install yum-utils
     sudo yum-config-manager --add-repo "https://openresty.org/package/${1}/openresty.repo"
-    sudo yum check-update
```
spacewander, line 27: Why remove this line?

leslie-tsang: When an update is available, `yum check-update` returns status code `100`, Which will break the execution.

spacewander, line 27: `sudo yum check-update || true` would be better?

leslie-tsang: IMO, `yum check-update` check whether there is a update or not would't bring any benefit for `install_dependencies_with_yum`.
`yum` don't need to update the repo cache after adding a new repo like `apt` does.
Shall we remove it ?

tzssangglass (approved):

spacewander (approved):

PR #5419, opened 2021-11-04
fix: add handler for invalid basic auth header values

### LINKED ISSUE #5411: bug: basic-auth plugin will cause the request abort if Authorization header value is invalid
### Issue description

Hi, there

Currently, the basic-auth plugin uses a regex pattern to fetch the part from the `Authorization` header, just like the following:

https://github.com/apache/apisix/blob/fc5f74709b44f6508ab89433e13b7613908fac5f/apisix/plugins/basic-auth.lua#L77-L83

But the code base doesn't validate the capture group number from the regex pattern, thus, if the `Authorization` header is abnormal (say, it doesn't start with the key word `Basic`), line 83 will cause the current Lua coroutine aborting and the request will be terminated with status code `500`.

### Environment

- apisix version (cmd: `apisix version`): master branch
- OS (cmd: `uname -a`):

### Steps to reproduce

1. Configure a Route and enable the `basic-auth` plugin;
2. Configure a Consumer and enable the `basic-auth` plugin;
3. Send requests with an invalid `Authorization Header` to Apache APISIX, let it hit the route created in step 1

### Actual result

Response with status `500`, and some error logs printed.

### Error log

```
2021/11/03 08:50:58 [info] 35#35: *37581 [lua] basic-auth.lua:127: phase_func(): plugin access phase, conf: {"disable":false}, client: 172.17.0.1, server: _, request: "GET /status/303 HTTP/1.1", host: "local2.httpbin.org"
2021/11/03 08:50:58 [error] 35#35: *37581 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/plugins/basic-auth.lua:83: attempt to index local 'm' (a nil value)
stack traceback:
coroutine 0:
        /usr/local/apisix/apisix/plugins/basic-auth.lua: in function 'create_obj_fun'
        /usr/local/apisix/apisix/core/lrucache.lua:92: in function 'lrucache'
        /usr/local/apisix/apisix/plugins/basic-auth.lua:99: in function 'extract_auth_header'
        /usr/local/apisix/apisix/plugins/basic-auth.lua:136: in function 'phase_func'
        /usr/local/apisix/apisix/plugin.lua:683: in function 'run_plugin'
        /usr/local/apisix/apisix/init.lua:448: in function 'http_access_phase'
        access_by_lua(nginx.conf:250):2: in main chunk, client: 172.17.0.1, server: _, request: "GET /status/303 HTTP/1.1", host: "local2.httpbin.org"
```

### Expected result

Should return `401` or `200`, and without any related error logs.

leslie-tsang:

Fix #5411 
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/basic-auth.lua
+++ b/apisix/plugins/basic-auth.lua
@@ -86,6 +86,10 @@ local function extract_auth_header(authorization)
 
         local decoded = ngx.decode_base64(m[1])
 
+        if not decoded then
+            return nil, "Failed to decode authentication header: " .. m[1]
+        end
+
         local res
         res, err = ngx_re.split(decoded, ":")
         if err then
--- a/apisix/plugins/basic-auth.lua
+++ b/apisix/plugins/basic-auth.lua
@@ -80,12 +80,19 @@ local function extract_auth_header(authorization)
             return nil, err
         end
 
+        if not m then
+            return nil, "Invalid authorization header format"
+        end
+
         local decoded = ngx.decode_base64(m[1])
 
         local res
         res, err = ngx_re.split(decoded, ":")
         if err then
-            return nil, "split authorization err:" .. err
+            return nil, "Split authorization err:" .. err
+        end
+        if #res < 2 then
+            return nil, "Split authorization err: invalid decoded data: " .. decoded
         end
 
         obj.username = ngx.re.gsub(res[1], "\\s+", "", "jo")
--- a/t/plugin/basic-auth.t
+++ b/t/plugin/basic-auth.t
@@ -163,7 +163,46 @@ GET /hello
 
 
 
-=== TEST 6: verify, invalid username
+=== TEST 6: verify, invalid basic authorization header
+--- request
+GET /hello
+--- more_headers
+Authorization: Bad_header YmFyOmJhcgo=
+--- error_code: 401
+--- response_body
+{"message":"Invalid authorization header format"}
+--- no_error_log
+[error]
+
+
+
+=== TEST 7: verify, invalid authorization value (bad base64 str)
+--- request
+GET /hello
+--- more_headers
+Authorization: Basic aaaa
+--- error_code: 401
+--- response_body eval
+qr/Split authorization err: invalid decoded data/
+--- no_error_log
+[error]
+
+
+
+=== TEST 8: verify, invalid authorization value (no password)
+--- request
+GET /hello
+--- more_headers
+Authorization: Basic YmFy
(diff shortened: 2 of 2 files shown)


apisix/plugins/basic-auth.lua
```diff
@@ -80,12 +80,19 @@ local function extract_auth_header(authorization)
             return nil, err
         end
 
+        if not m then
+            return nil, "Invalid authorization header format"
+        end
+
         local decoded = ngx.decode_base64(m[1])
```
leslie-tsang, line 87: @spacewander seems decode_base64 always return a `ffi_string`
https://github.com/openresty/lua-resty-core/blob/985eb5b323468effaa66deb5cbd4f800b99834cf/lib/resty/core/base64.lua#L60

spacewander: `ffi_string` is not a type, but a function returns a Lua string

spacewander: Oh, I know what you mean. No, you point to the wrong method. See https://github.com/openresty/lua-resty-core/blob/985eb5b323468effaa66deb5cbd4f800b99834cf/lib/resty/core/base64.lua#L107

t/plugin/basic-auth.t
```diff
@@ -163,7 +163,46 @@ GET /hello
...
+--- no_error_log
+[error]
+
+
+
+=== TEST 7: verify, invalid authorization value (bad base64 str)
+--- request
+GET /hello
+--- more_headers
+Authorization: Basic aaaa
```
spacewander, line 183: aaaa is valid base64 str

apisix/plugins/basic-auth.lua
```diff
@@ -80,12 +80,23 @@ local function extract_auth_header(authorization)
...
 
+        if not decoded then
+            return nil, "Failed to decode authentication header: " .. m[1]
+        end
+
```
leslie-tsang, line 92: @spacewander 
Is there a test case can hit this scenario ?

t/plugin/basic-auth.t
leslie-tsang: I will fix it.

apisix/plugins/basic-auth.lua
leslie-tsang: 😂 `aca` don't trigger as well, str witl `_` like `aca_a` work as expected.

spacewander (approved):

tokers (approved):

PR #5422, opened 2021-11-04
feat(limit-* plugin):  fallback to remote_addr when key is missing

Xunzhuo:

 fallback to `remote_addr` when key is missing or invalid in limit-* plugin

### Pre-submission checklist:


apisix/plugins/limit-conn/init.lua
```diff
@@ -64,10 +64,9 @@ function _M.increase(conf, ctx)
     end
 
     if key == nil then
-        core.log.info("bypass the limit conn as the key is empty")
-        -- Bypass the limit conn when the key is empty.
-        -- This behavior is the same as Nginx
-        return
+        core.log.info("The key of limit conn is empty or invalid, set $remote_addr as the key")
+        -- When the key is empty or invalid, set $remote_addr as the default key
+        key = "remote_addr"
```
spacewander, line 69: The key should be the value of remote_addr not remote_addr itself

Xunzhuo: Done. Thanks

t/plugin/limit-conn2.t
```diff
@@ -410,32 +440,102 @@ GET /t
...
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:1980": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/limit_conn"
+                }]],
+                [[{
```
spacewander, line 486: There is no need to check the response, right?

t/plugin/limit-conn2.t
```diff
@@ -321,8 +351,8 @@ request latency is nil
                  [[{
                         "plugins": {
                             "limit-conn": {
-                                "conn": 2,
-                                "burst": 1,
+                                "conn": 1,
```
spacewander, line 354: Why change this configuration?

docs/en/latest/plugins/limit-conn.md
```diff
@@ -42,14 +42,14 @@ Limiting request concurrency plugin.
...
 | key_type      | string  | optional    |   "var"   | ["var", "var_combination"] | the type of key. |
-| key           | string  | required    |         |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable, like "remote_addr" or "consumer_name". If the `key_type` is "var_combination", the key will be a combination of variables, like "$remote_addr $consumer_name". |
+| key           | string  | required    |         |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable, like "remote_addr" or "consumer_name". If the `key_type` is "var_combination", the key will be a combination of variables, like "$remote_addr $consumer_name". If the key is empty or invalid, `remote_addr` will be set as the default key.|
```
spacewander, line 45:
```suggestion
| key           | string  | required    |         |  | the user specified key to limit the rate. If the `key_type` is "var", the key will be treated as a name of variable, like "remote_addr" or "consumer_name". If the `key_type` is "var_combination", the key will be a combination of variables, like "$remote_addr $consumer_name". If the value of the key is empty, `remote_addr` will be set as the default key.|
```

apisix/plugins/limit-req.lua
```diff
@@ -102,10 +102,9 @@ function _M.access(conf, ctx)
     end
 
     if key == nil then
-        core.log.info("bypass the limit req as the key is empty")
-        -- Bypass the limit req when the key is empty.
-        -- This behavior is the same as Nginx
-        return
+        core.log.info("The key of limit req is empty or invalid, set $remote_addr as the key")
```
spacewander, line 105:
```suggestion
        core.log.info("The value of the configured key is empty, use client IP instead")
```

The value is empty, not the key itself.

Xunzhuo: Done, thanks.

docs/en/latest/plugins/limit-conn.md
Xunzhuo: Done, thanks.

t/plugin/limit-conn2.t
Xunzhuo: Sorry, I do not understand this point? Could you give me more infos?

t/plugin/limit-conn2.t
Xunzhuo: It is easier to test when value of key is empty.

t/plugin/limit-conn2.t
Xunzhuo: Do you mean that there is no need to check this?
```
--- response_body
passed
```

t/plugin/limit-conn2.t
spacewander: Why changing the configuration doesn't provide a different result?

spacewander: I see. It is already covered in the new test.

spacewander (approved):

tokers (approved):

PR #5463, opened 2021-11-09
feat: introducing prefer_name attribute in datadog plugin

bisakhmondal:

As discussed offline with @spacewander, this PR adds an additional attribute in plugin schema `prefer_name` to provide the flexibility of choosing route/service id over the name (in case of name conflict/user preference).
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -18,6 +18,7 @@ local core = require("apisix.core")
 local plugin = require("apisix.plugin")
 local batch_processor = require("apisix.utils.batch-processor")
 local fetch_log = require("apisix.utils.log-util").get_full_log
+local service_fetch = require("apisix.http.service").get
 local ngx = ngx
 local udp = ngx.socket.udp
 local format = string.format
@@ -43,6 +44,7 @@ local schema = {
         inactive_timeout = {type = "integer", minimum = 1, default = 5},
         batch_max_size = {type = "integer", minimum = 1, default = 5000},
         max_retry_count = {type = "integer", minimum = 1, default = 1},
+        prefer_name = {type = "boolean", default = true}
     }
 }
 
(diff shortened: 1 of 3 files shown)


t/plugin/datadog.t
```diff
@@ -395,3 +395,141 @@ message received: apisix\.apisix\.latency:[\d.]+\|h\|#source:apisix,new_tag:must
 message received: apisix\.ingress\.size:[\d]+\|ms\|#source:apisix,new_tag:must,route_name:1,balancer_ip:[\d.]+,response_status:200,scheme:http
 message received: apisix\.egress\.size:[\d]+\|ms\|#source:apisix,new_tag:must,route_name:1,balancer_ip:[\d.]+,response_status:200,scheme:http
 /
+
+
+
+=== TEST 8: testing behaviour with service id
```
spacewander, line 401: Is there a test for prefer_name + the route name?

apisix/plugins/datadog.lua
```diff
@@ -83,15 +85,24 @@ local function generate_tag(entry, const_tags)
...
     elseif entry.route_id and entry.route_id ~= "" then
         core.table.insert(tags, "route_name:" .. entry.route_id)
     end
 
     if entry.service_id and entry.service_id ~= "" then
-        core.table.insert(tags, "service_id:" .. entry.service_id)
+        local svc_id = entry.service_id
+
+        if entry.prefer_name then
+            local svc = service_fetch(svc_id)
```
spacewander, line 99: To prevent race and make the report code straightforward, it would be better to fetch the name in the log phase instead of the report process.

bisakhmondal: Thanks for pointing it out. Since everything is hot loaded it could be a mess. Updating.

t/plugin/datadog.t
bisakhmondal: Yes. Both enabled: test 2- 6
Prefer_name(true), route_name(false) - test 7
prefer_name(false), route_name(true) - Test-9 (route is created at test-8 with route_name)

apisix/plugins/datadog.lua
```diff
@@ -144,6 +149,12 @@ function _M.log(conf, ctx)
     entry.balancer_ip = ctx.balancer_ip or ""
     entry.route_name = ctx.route_name or ""
     entry.scheme = ctx.upstream_scheme or ""
+    entry.prefer_name = conf.prefer_name
+    -- capture service name at request response cycle if prefer_name is set.
+    if entry.prefer_name and entry.service_id and entry.service_id ~= "" then
+       local svc = service_fetch(entry.service_id)
+       entry.service_name = svc and svc.value.name or ""
```
spacewander, line 156:
```suggestion
       entry.service_id = svc and svc.value.name or ""
```
So there is no need to check the service_name in the report process.

And we can do the same thing with the route, so there is no need to pass the prefer_name configuration.

bisakhmondal: Ack. Done. Thank you.

apisix/plugins/datadog.lua
```diff
@@ -144,17 +141,22 @@ function _M.log(conf, ctx)
...
     entry.balancer_ip = ctx.balancer_ip or ""
-    entry.route_name = ctx.route_name or ""
     entry.scheme = ctx.upstream_scheme or ""
     entry.prefer_name = conf.prefer_name
 
-    -- capture service name at request response cycle if prefer_name is set.
-    if entry.prefer_name and entry.service_id and entry.service_id ~= "" then
-       local svc = service_fetch(entry.service_id)
+    -- if prefer_name is set, fetch the service/route name. If the name is nil, fall back to id.
+    if entry.prefer_name then
```
spacewander, line 148: We can use `conf.prefer_name` directly?

spacewander (approved):

tzssangglass (approved):

PR #5478, opened 2021-11-10
feat: provide skywalking logger plugin

dmsolr:
### Pre-submission checklist:

provide a new logger plugin to support log access log to SkyWalking in SkyWalking Log Format. If skywalking trace plugin is enabled or requests with SkyWalking Trace context header, this plugin will parse the trace context header and print trace context in log. So that the access log can relate to trace.

### DIFF
--- a/apisix/plugins/skywalking-logger.lua
+++ b/apisix/plugins/skywalking-logger.lua
@@ -22,11 +22,11 @@ local http            = require("resty.http")
 local url             = require("net.url")
 local plugin          = require("apisix.plugin")
 
-local base64          = require("ngx.base64") 
+local base64          = require("ngx.base64")
 local ngx_re          = require("ngx.re")
 
 local ngx      = ngx
-local tostring = tostring 
+local tostring = tostring
 local ipairs   = ipairs
 local timer_at = ngx.timer.at
 
@@ -157,16 +157,21 @@ function _M.log(conf, ctx)
     if not log_body.route_id then
         log_body.route_id = "no-matched"
     end
-    
-    local h, err = ngx.req.get_headers()
-    -- 1-TRACEID-SEGMENTID-3-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
-    local ids = ngx_re.split(h["sw8"], '-')
 
-    local trace_context = {
-        traceId = base64.decode_base64url(ids[2]),
-        traceSegment = base64.decode_base64url(ids[3]),
-        spanId = tonumber(ids[4])
-    }
+    local trace_context
+    local headers = ngx.req.get_headers()
+    if headers then
+        -- 1-TRACEID-SEGMENTID-3-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
+        local ids = ngx_re.split(headers["sw8"], '-')
+
+        trace_context = {
+            traceId = base64.decode_base64url(ids[2]),
+            traceSegment = base64.decode_base64url(ids[3]),
+            spanId = tonumber(ids[4])
+        }
+    else
+        trace_context = {}
+    end
 
     local entry = {
         traceContext = trace_context,
--- a/apisix/plugins/skywalking-logger.lua
+++ b/apisix/plugins/skywalking-logger.lua
@@ -163,7 +163,7 @@ function _M.log(conf, ctx)
         if #ids == 8 then
             trace_context = {
                 traceId = base64.decode_base64url(ids[2]),
-                traceSegment = base64.decode_base64url(ids[3]),
+                traceSegmentId = base64.decode_base64url(ids[3]),
                 spanId = tonumber(ids[4])
             }
         else
--- a/apisix/plugins/skywalking-logger.lua
+++ b/apisix/plugins/skywalking-logger.lua
@@ -1,0 +1,230 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+local batch_processor = require("apisix.utils.batch-processor")
+local log_util        = require("apisix.utils.log-util")
+local core            = require("apisix.core")
+local http            = require("resty.http")
+local url             = require("net.url")
+local plugin          = require("apisix.plugin")
+
+local base64          = require("ngx.base64") 
+local ngx_re          = require("ngx.re")
+
+local ngx      = ngx
+local tostring = tostring 
+local ipairs   = ipairs
+local timer_at = ngx.timer.at
+
+local plugin_name = "skywalking-logger"
+local stale_timer_running = false
+local buffers = {}
+
+local schema = {
+    type = "object",
+    properties = {
+        endpoint_addr = core.schema.uri_def,
+        service_name = {type = "string", default = "APISIX"},
+        service_instance_name = {type = "string", default = "APISIX Instance Name"},
+        timeout = {type = "integer", minimum = 1, default = 3},
+        name = {type = "string", default = "skywalking logger"},
+        max_retry_count = {type = "integer", minimum = 0, default = 0},
+        retry_delay = {type = "integer", minimum = 0, default = 1},
+        buffer_duration = {type = "integer", minimum = 1, default = 60},
+        inactive_timeout = {type = "integer", minimum = 1, default = 5},
+        batch_max_size = {type = "integer", minimum = 1, default = 1000},
+        include_req_body = {type = "boolean", default = false},
+    },
+    required = {"endpoint_addr"},
+}
+
+
+local metadata_schema = {
+    type = "object",
+    properties = {
+        log_format = log_util.metadata_schema_log_format,
+    },
+}
+
+
+local _M = {
+    version = 0.1,
+    priority = 410,
+    name = plugin_name,
+    schema = schema,
+    metadata_schema = metadata_schema,
+}
+
+
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
+end
+
+
+local function send_http_data(conf, log_message)
+    local err_msg
+    local res = true
+    local url_decoded = url.parse(conf.endpoint_addr)
+    local host = url_decoded.host
+    local port = url_decoded.port
+
+    core.log.info("sending a batch logs to ", conf.endpoint_addr)
+
+    local httpc = http.new()
+    httpc:set_timeout(conf.timeout * 1000)
+    local ok, err = httpc:connect(host, port)
+
+    if not ok then
+        return false, "failed to connect to host[" .. host .. "] port["
+            .. tostring(port) .. "] " .. err
+    end
+
+    local httpc_res, httpc_err = httpc:request({
+        method = "POST",
+        path = "/v3/logs",
+        body = log_message,
+        headers = {
+            ["Host"] = url_decoded.host,
+            ["Content-Type"] = "application/json",
+        }
+    })
+
+    if not httpc_res then
+        return false, "error while sending data to [" .. host .. "] port["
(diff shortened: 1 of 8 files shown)


apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+    local trace_context
+    local headers = ngx.req.get_headers()
+    if headers then
+        -- 1-TRACEID-SEGMENTID-3-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
+        local ids = ngx_re.split(headers["sw8"], '-')
+
+        trace_context = {
+            traceId = base64.decode_base64url(ids[2]),
+            traceSegment = base64.decode_base64url(ids[3]),
+            spanId = tonumber(ids[4])
```
spacewander, line 171: The span id is missing in the comment?

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+    end
+
+    local trace_context
+    local headers = ngx.req.get_headers()
+    if headers then
+        -- 1-TRACEID-SEGMENTID-3-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
+        local ids = ngx_re.split(headers["sw8"], '-')
+
+        trace_context = {
+            traceId = base64.decode_base64url(ids[2]),
```
spacewander, line 169: Need to check the length of ids and the result of decode_base64url

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+    local entry = {
+        traceContext = trace_context,
+        body = {
+            json = {
+                json = core.json.encode(log_body, true)
+            }
+        },
+        service = conf.service_name,
+        serviceInstance = conf.service_instance_name,
+        endpoint = ngx.var.uri,
```
spacewander, line 186:
```suggestion
        endpoint = ctx.var.uri,
```

docs/en/latest/plugins/skywalking-logger.md
```diff
@@ -0,0 +1,117 @@
...
+HTTP/1.1 200 OK
+...
+hello, world
+```
+
+Completion of the steps, could find the Log details on `SkyWalking UI`.
+
+## Metadata
+
+`skywalking-logger` is also support to custom log format like [http-logger](./http-logger.md).
```
spacewander, line 92:
```suggestion
`skywalking-logger` also supports to custom log format like [http-logger](./http-logger.md).
```

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+    if metadata and metadata.value.log_format
+       and core.table.nkeys(metadata.value.log_format) > 0
+    then
+        log_body = log_util.get_custom_format_log(ctx, metadata.value.log_format)
+    else
+        log_body = log_util.get_full_log(ngx, conf)
+    end
+
+    if not log_body.route_id then
+        log_body.route_id = "no-matched"
```
spacewander, line 159: Is the `no-matched` necessary for the skywalking? If not, we can leave it empty.

apisix/plugins/skywalking-logger.lua
dmsolr: I update the comment. Currently, the 4th part, `3` means span id.
```
4. Parent span ID. Must be an integer. It begins with 0. This span ID points to the parent span in parent trace segment.
```
Ref [SkyWalking Cross Process Propagation Headers Protocol](https://github.com/apache/skywalking/blob/master/docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md)

spacewander: It would be better if you also add the link to the skywalking's doc.

apisix/plugins/skywalking-logger.lua
dmsolr: update

apisix/plugins/skywalking-logger.lua
dmsolr: removed

apisix/plugins/skywalking-logger.lua
dmsolr: got it

moonming:
@wu-sheng please take a look, and welcome feedback:)

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+    end
+
+    local trace_context
+    local headers = ngx.req.get_headers()
+    if headers then
+        -- 1-TRACEID-SEGMENTID-3-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
+        local ids = ngx_re.split(headers["sw8"], '-')
```
wu-sheng: Are you sure you want to do this? SkyWalking has tracer for APISIX, which has been integrated. Why don't you try to read it from the context?

Decoding the header is fine from tech perspective, but it could be a risk when one day, Lua agent supports `namespace` concept.

wu-sheng: And all following decode(twice for trace id and span id) and splitting cost unnecessary resources.

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,236 @@
...
+                json = core.json.encode(log_body, true)
+            }
+        },
+        service = conf.service_name,
+        serviceInstance = conf.service_instance_name,
+        endpoint = ngx.var.uri,
+    }
+
+    if not stale_timer_running then
+        -- run the timer every 30 mins if any log is present
```
wu-sheng, line 196: Every 30 mins? For?

wu-sheng: I think this is a huge latency in low traffic scenario. No one is willing to wait for 30 mins for log collecting.

wu-sheng (commented):
General review from design perspective(nothing related to code details).

apisix/plugins/skywalking-logger.lua
dmsolr: I am considering about it still works when tracer plugin is not enabling. 
If we give the version matching matric, it looks not very dangerous.

wu-sheng: Your call, SkyWalking wouldn't change v3 format for a long time(already years, expect more) AFAIK. `namespace` would be an issue, but not very serious, just duplicate configurations.
I just think resource costing is very sensitive. And this is not that risk, as when the incompatible issue happens, it happens at the first time both plugin activated. So anyone will notice this immediately, they could do fix or disable one/both quickly.

apisix/plugins/skywalking-logger.lua
dmsolr: it sweeps the stale log data in every 30m. 
When the data transport fails, the data will remain in memory. Therefore, regular cleaning is required.

wu-sheng: I think this is not very general way to process logs. Usually, unless you have local cache or mmap(like SkyWalking Satellite), logs should be abandoned quickly once retry threshold reached.
The reason behind this logic is, observability presents a large data set but isn't always very usually, especially for logs. Metrics are small, and tolerable to keep latest values, but different from logs.

dmsolr: Make sense to me. @spacewander how do you think?

spacewander: Err. @dmsolr's explain is incorrect. This timer doesn't clean up the stale log. It clean up the stale buffers when they are no longer holding the log.

apisix/plugins/skywalking-logger.lua
```diff
@@ -0,0 +1,242 @@
...
+        log_body = log_util.get_full_log(ngx, conf)
+    end
+
+    local trace_context
+    local sw_header = ngx.req.get_headers()["sw8"]
+    if sw_header then
+        -- 1-TRACEID-SEGMENTID-SPANID-PARENT_SERVICE-PARENT_INSTANCE-PARENT_ENDPOINT-IPPORT
+        local ids = ngx_re.split(sw_header, '-')
+        if #ids == 8 then
+            local trace_id, err = base64.decode_base64url(ids[2])
```
spacewander, line 164: `decode_base64url` only returns str or nil

apisix/plugins/skywalking-logger.lua
wu-sheng: If it is about `clean up the stale buffers` only, than, this is APISIX's call. Nothing relates to SkyWalking or observability preference.

t/plugin/skywalking-logger.t
```diff
@@ -0,0 +1,285 @@
...
+            content_by_lua_block {
+                local core = require("apisix.core")
+
+                core.log.warn(core.json.encode(core.request.get_body(), true))
+            }
+        }
+    }
+_EOC_
+
+    $block->set_value("http_config", $http_config);
```
spacewander, line 43: New tests should contain a common section, like https://github.com/apache/apisix/blob/f06f6cc0e8e41875bf105cf2c18457339002df53/t/plugin/limit-count2.t#L38

There is no need to repeat them.

apisix/plugins/skywalking-logger.lua
dmsolr: if so, we don't need to check the result of decode_base64url, right?

docs/en/latest/plugins/skywalking-logger.md
```diff
@@ -0,0 +1,117 @@
+---
+title: http-logger
```
wu-sheng, line 2: Is this the right title?

dmsolr: fixed

t/plugin/skywalking-logger.t
dmsolr: updated

apisix/plugins/skywalking-logger.lua
spacewander: If I am correct, we should check if the trace_id is nil or not? The trace_id is required.

dmsolr: trace_context is optional. If trace_id is missing, it cannot relate to trace. But missing trace_id log can be found in `log` view.

Refer https://github.com/apache/skywalking-data-collect-protocol/blob/master/logging/Logging.proto#L99-L107

dmsolr: Currently, I prefer to collect the missing trace_context log.

docs/en/latest/plugins/skywalking-logger.md
```diff
@@ -0,0 +1,117 @@
...
+- [**Name**](#name)
+- [**Attributes**](#attributes)
+- [**How To Enable**](#how-to-enable)
+- [**Test Plugin**](#test-plugin)
+- [**Metadata**](#metadata)
+- [**Disable Plugin**](#disable-plugin)
+
+## Name
+
+`skywalking-logger` is a plugin which push Access Log data to `SkyWalking OAP` server over HTTP.
```
wu-sheng, line 35:
```suggestion
`skywalking-logger` is a plugin which push Access Log data to `SkyWalking OAP` server over HTTP. If there is tracing context existing, it sets up the trace-log correlation automatically, and relies on [SkyWalking Cross Process Propagation Headers Protocol](https://skywalking.apache.org/docs/main/latest/en/protocols/skywalking-cross-process-propagation-headers-protocol-v3/).
```

wu-sheng (approved):
Good to me. Thanks.

spacewander (approved):

t/plugin/skywalking-logger.t
```diff
@@ -0,0 +1,269 @@
...
+
+
+
+=== TEST 9: access local server and test log format
+--- request
+GET /opentracing
+--- response_body
+opentracing
+--- error_log eval
+qr/.*\{\\\"json\\\":\\\"\{(\\\\\\\"\@timestamp\\\\\\\":\\\\\\\".*\\\\\\\"|\\\\\\\"client_ip\\\\\\\":\\\\\\\"127\.0\.0\.1\\\\\\\"|\\\\\\\"host\\\\\\\":\\\\\\\"localhost\\\\\\\"|\\\\\\\"route_id\\\\\\\":\\\\\\\"1\\\\\\\"|,){7}\}/
```
membphis, line 268: Not easy to read, can there be a more readable way? @dmsolr

dmsolr: this is raw skywalking log format.
Seem like have some approach to simplify it. I will have a try later.

t/plugin/skywalking-logger.t
```diff
@@ -0,0 +1,269 @@
...
+opentracing
+--- error_log eval
+qr/.*\\\"traceContext\\\":\{(\\\"traceSegmentId\\\":\\\"ae709769-6e20-4c68-9733-10f455216cc5\\\"|\\\"traceId\\\":\\\"ae709769-6e20-4c68-9733-10f455216cc5\\\"|\\\"spanId\\\":1|,){5}\}.*/
+--- wait: 0.5
+
+
+
+=== TEST 7: test wrong trace context header
+--- request
+GET /opentracing
```
wu-sheng, line 214: Ah, I suddenly noticed this, why `/opentracing`? Rather than `/request`.

PR #5479, opened 2021-11-10
feat(plugin): azure serverless functions

bisakhmondal:
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/azure-functions.lua
+++ b/apisix/plugins/azure-functions.lua
@@ -91,7 +91,7 @@ function _M.access(conf, ctx)
         end
     end
 
-    headers["Host"],  headers["host"] = nil, nil
+    headers["host"] = nil
     local params = {
         method = ngx.req.get_method(),
         body = req_body,
--- a/apisix/plugins/azure-functions.lua
+++ b/apisix/plugins/azure-functions.lua
@@ -59,16 +59,24 @@ end
 function _M.access(conf, ctx)
     local uri_args = core.request.get_uri_args(ctx)
     local headers = core.request.headers(ctx) or {}
-    local req_body, _ = core.request.get_body()
+    local req_body, err = core.request.get_body()
 
+    if err then
+        core.log.error("error while reading request body: " .. err)
+        return 400
+    end
 
-    -- set authorization headers
-    if conf.authorization then
-        headers["x-functions-key"] = conf.authorization.apikey or ""
-        headers["x-functions-clientid"] = conf.authorization.clientid or ""
-    else
-        headers["x-functions-key"] = getenv(env_key.API)
-        headers["x-functions-clientid"] = getenv(env_key.CLIENT_ID)
+    -- set authorization headers if not already set by the client
+    -- we are following not to overwrite the authz keys
+    if not headers["x-functions-key"] and
+            not headers["x-functions-clientid"] then
+        if conf.authorization then
+            headers["x-functions-key"] = conf.authorization.apikey or ""
+            headers["x-functions-clientid"] = conf.authorization.clientid or ""
+        else
+            headers["x-functions-key"] = getenv(env_key.API)
+            headers["x-functions-clientid"] = getenv(env_key.CLIENT_ID)
+        end
     end
 
     local params = {
@@ -92,7 +100,8 @@ function _M.access(conf, ctx)
     local res, err = httpc:request_uri(conf.function_uri, params)
 
     if not res or err then
-        return core.response.exit(500, "failed to process azure function, err: " .. err)
+        core.log.error("failed to process azure function, err: " .. err)
+        return 503
     end
 
     -- setting response headers
--- a/apisix/plugins/azure-functions.lua
+++ b/apisix/plugins/azure-functions.lua
@@ -1,0 +1,105 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local core = require("apisix.core")
+local http = require("resty.http")
+local ngx  = ngx
+local getenv = os.getenv
+local plugin_name = "azure-functions"
+
+local env_key = {
+    API = "AZURE_FUNCTIONS_APIKEY",
+    CLIENT_ID = "AZURE_FUNCTIONS_CLIENTID"
+}
+
+local schema = {
+    type = "object",
+    properties = {
+        function_uri = {type = "string"},
+        authorization = {
+            type = "object",
+            properties = {
+                apikey = {type = "string"},
+                clientid = {type = "string"}
+            }
+        },
+        timeout = {type = "integer", minimum = 1000, default = 3000},
+        ssl_verify = {type = "boolean", default = false},
+        keepalive = {type = "boolean", default = true},
+        keepalive_timeout = {type = "integer", minimum = 1000, default = 60000},
+        keepalive_pool = {type = "integer", minimum = 1, default = 5}
+    },
+    required = {"function_uri"}
+}
+
+local _M = {
+    version = 0.1,
+    priority = 505,
+    name = plugin_name,
+    schema = schema,
+}
+
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+function _M.access(conf, ctx)
+    local uri_args = core.request.get_uri_args(ctx)
+    local headers = core.request.headers(ctx) or {}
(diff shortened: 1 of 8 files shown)


apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,104 @@
...
+        params.keepalive_pool = conf.keepalive_pool
+    end
+
+    local httpc = http.new()
+    httpc:set_timeout(conf.timeout)
+
+    local res, err = httpc:request_uri(conf.function_uri, params)
+
+    if not res or err then
+        return core.response.exit(500, "failed to process azure function, err: " .. err)
```
spacewander, line 95:
```suggestion
        core.log.error("failed to process azure function, err: " .. err)
        return 503
```
I think it will be safer not to return an internal error to the client.

conf/config-default.yaml
```diff
@@ -340,6 +340,7 @@ plugins:                          # plugin list (sorted by priority)
   - response-rewrite               # priority: 899
   #- dubbo-proxy                   # priority: 507
   - grpc-transcode                 # priority: 506
+  - azure-functions                # priority: 505
```
spacewander, line 343: Better to put it ahead of serverless-post-function

apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+function _M.access(conf, ctx)
+    local uri_args = core.request.get_uri_args(ctx)
+    local headers = core.request.headers(ctx) or {}
+    local req_body, _ = core.request.get_body()
```
tokers, line 62: What if the request body are large and we cannot load it into memory totally?

t/plugin/azure-functions.t
```diff
@@ -0,0 +1,226 @@
...
+
+repeat_each(1);
+no_long_string();
+no_root_location();
+no_shuffle();
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if (!defined $block->additional_http_config) {
```
spacewander, line 27: We can use `http_config` to inject HTTP config, see https://github.com/apache/apisix/blob/master/t/discovery/consul_kv.t

And I suggest you put `azure-demo` as https://github.com/apache/apisix/blob/2c12b3695302f580a1498333aaafb68da26a3096/t/lib/server.lua#L33

There is no need to introduce a new server.

apisix/plugins/azure-functions.lua
spacewander: The body will be missing when it is larger than client_body_buffer_size (8k by default)
We can read the body by ourselves via ngx.req.socket? See https://github.com/openresty/lua-resty-upload/blob/master/lib/resty/upload.lua

We can do it in the next PR.

spacewander: I just realize it uses `core.request.get_body`, which will read data from file if need. BTW, we should check the err returned by this function.

apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,104 @@
...
+
+
+    -- set authorization headers
+    if conf.authorization then
+        headers["x-functions-key"] = conf.authorization.apikey or ""
+        headers["x-functions-clientid"] = conf.authorization.clientid or ""
+    else
+        headers["x-functions-key"] = getenv(env_key.API)
```
spacewander, line 70: We need to expose the AZURE_FUNCTIONS_APIKEY via `env AZURE_FUNCTIONS_APIKEY`, if this plugin is enabled. Nginx will hide all environment variables by default.

apisix/plugins/azure-functions.lua
bisakhmondal: Ack

apisix/plugins/azure-functions.lua
bisakhmondal: thanks for pointing it out @tokers! 
@spacewander even if it reads the content from a file, it reads everything in one go
https://github.com/apache/apisix/blob/2c12b3695302f580a1498333aaafb68da26a3096/apisix/core/request.lua#L172-L181

I think we should use chunked transfer encoding here to proxy the request with a large body size. WDYT?

t/plugin/azure-functions.t
bisakhmondal: Gotcha

apisix/plugins/azure-functions.lua
bisakhmondal: You mean like this `env AZURE_FUNCTIONS_APIKEY="key" make run`?
Or you are suggesting to update something?

conf/config-default.yaml
bisakhmondal: What should be the priority value, negative?

t/plugin/azure-functions.t
bisakhmondal: Done, Is it okay now?

apisix/plugins/azure-functions.lua
spacewander: I think normally people don't need to submit big data to azure functions. We can keep this function unchanged for now.

apisix/plugins/azure-functions.lua
spacewander: Like https://github.com/apache/apisix/blob/2c12b3695302f580a1498333aaafb68da26a3096/apisix/cli/ngx_tpl.lua#L52, but only when this plugin is enabled:
https://github.com/apache/apisix/blob/2c12b3695302f580a1498333aaafb68da26a3096/apisix/cli/ngx_tpl.lua#L209

conf/config-default.yaml
spacewander: priority: -1900?

apisix/plugins/azure-functions.lua
spacewander: I change my mind. Maybe it could be easier to pass this argument via plugin metadata.

bisakhmondal: Metadata won't be a great fit, IMHO. There might be multiple instances of plugins linked to different routes and we can't force users to use the same key they put into the metadata schema(must be the master key, instead of function-specific keys). This clearly violates principle of Least Access Privileges, isn't it?

apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,113 @@
...
+            headers["x-functions-key"] = getenv(env_key.API)
+            headers["x-functions-clientid"] = getenv(env_key.CLIENT_ID)
+        end
+    end
+
+    local params = {
+        method = ngx.req.get_method(),
+        body = req_body,
+        query = uri_args,
+        headers = headers,
```
spacewander, line 86: Do we need to remove the host header? Otherwise, the host of APISIX will be used.

apisix/plugins/azure-functions.lua
tokers: Agree, but it's better to add some hints in the document. @bisakhmondal @spacewander

apisix/plugins/azure-functions.lua
bisakhmondal: Removed env vars and added metadata as a fallback option

t/plugin/azure-functions.t
```diff
@@ -0,0 +1,372 @@
...
+--- response_headers
+Content-Length: 13
+X-Extra-Header: MUST
+
+
+
+=== TEST 5: http2 check response body and headers
+--- http2
+--- exec
+curl -XGET --http2 --http2-prior-knowledge localhost:1984/azure
```
spacewander, line 198: We can use regular `--- request` to send http2 request, see
https://github.com/openresty/test-nginx/blob/9d5a20f884d67974f9eafeed5a1540cd32945fb2/lib/Test/Nginx/Socket.pm#L2310

apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,137 @@
...
+            -- plugin will fallback to the master key (if any) present inside the metadata.
+            local metadata = plugin.plugin_metadata(plugin_name)
+            if metadata then
+                headers["x-functions-key"] = metadata.value.master_apikey
+                headers["x-functions-clientid"] = metadata.value.master_clientid
+            end
+        end
+    end
+
+    headers["Host"],  headers["host"] = nil, nil
```
spacewander, line 94:
```suggestion
    headers["host"] = nil
```
is enough.

The fetched header is in lowercase.

bisakhmondal: Cool

t/plugin/azure-functions.t
bisakhmondal: Nice, it is. Updated

apisix/plugins/azure-functions.lua
```diff
@@ -0,0 +1,137 @@
...
+    return core.schema.check(schema, conf)
+end
+
+function _M.access(conf, ctx)
+    local uri_args = core.request.get_uri_args(ctx)
+    local headers = core.request.headers(ctx) or {}
+    local req_body, err = core.request.get_body()
+
+    if err then
+        core.log.error("error while reading request body: " .. err)
```
membphis, line 72: bad style, it should be `core.log.error("error while reading request body: ", err)`

bisakhmondal: Ack. Thanks for pointing it out!

bzp2010:
I have a question, should we standardize the naming of all serverless plugins? such as `xxx-function` or `xxx-serverless` 🤔

spacewander (approved):

spacewander:
@bzp2010 
I think we can use the product name directly (azure-functions in this case). It is easier to be searched.

PR #5480, opened 2021-11-11
refactor: Method for latency calculation and related refactoring

bisakhmondal:

- Separate utility method for latency calculation.
- Refactor datadog and Prometheus plugin to use the same

### Pre-submission checklist:

### DIFF
--- a/apisix/utils/log-util.lua
+++ b/apisix/utils/log-util.lua
@@ -169,11 +169,7 @@ function _M.latency_details_in_ms(ctx)
         end
     end
 
-    return {
-        latency = latency,
-        upstream_latency = upstream_latency,
-        apisix_latency = apisix_latency,
-    }
+    return latency, upstream_latency, apisix_latency
 end
 
 return _M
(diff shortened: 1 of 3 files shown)


apisix/utils/log-util.lua
```diff
@@ -153,4 +153,27 @@ function _M.get_req_original(ctx, conf)
...
+
+        -- The latency might be negative, as Nginx uses different time measurements in
+        -- different metrics.
+        -- See https://github.com/apache/apisix/issues/5146#issuecomment-928919399
+        if apisix_latency < 0 then
+            apisix_latency = 0
+        end
+    end
+
+    return {
```
spacewander, line 172: We can return multiple results directly in Lua.

bisakhmondal: Yes. But in this case, I thought a map would be efficient It's easy to pick the respective value through key without checking which param represents what latency. 
If you suggest, updating the signature.

bisakhmondal: Done

spacewander (approved):

tzssangglass (approved):

Issue #5485, opened 2021-11-11
Proposal: support Apache OpenWhisk serverless framework plugin


bzp2010:
# Proposal

## Background

OpenWhisk is an open source Faas computing platform in the Apache Foundation, announced and contributed to the open source community by IBM in 2016. In terms of running logic, OpenWhisk, like AWS Lambda, provides users with an event-driven, stateless computing model based on a variety of programming languages with direct support.
Its overall container-based runtime deployment (including traffic entry, controllers, triggers, etc.), using Docker to achieve function operation, management and scaling.
Meanwhile, in addition to container invocation and function management, OpenWhisk also includes authentication, function asynchronous triggering and other functions.

## Test Environment
Configure the standalone run as per the official documentation and execute the following command after running Docker locally.
```shell
git clone https://github.com/apache/openwhisk.git
cd openwhisk
./gradlew core:standalone:bootRun
```

## Objectives
Supports using functions in OpenWhisk as routing "upstream" via plugins to enable access to Faas functions.

## Solutions
1. plugin schema
<html>
<body>
<div class="okr-block-clipboard" data-okr="%7B%22okrDelta%22%3A%5B%7B%22lineType%22%3A%22unsupport%22%2C%22lineOptions%22%3A%7B%7D%2C%22lineContent%22%3A%5B%5D%7D%5D%2C%22businessKey%22%3A%22lark-doc%22%7D"></div><div data-zone-id="0" data-line-index="0" style="white-space: pre;">

Name | Type | Optional | Default | Description
-- | -- | -- | -- | --
api_host | string | no |   | OpenWhisk API host (eg. https://localhost:3233)
ssl_verify | bool | yes | true | Whether to verify the certificate
service_token | string | no |   | OpenWhisk ServiceToken (The format is `xxx:xxx`，Passed through Basic Auth when calling the API)
namespace | string | no |   | OpenWhisk  Namespace (eg. guest)
action | string | no |   | OpenWhisk Action (eg. hello)
result | bool | yes | true | Whether to get Action metadata (default to execute function and get response; false to get Action metadata but not execute Action, including runtime, function body, restrictions, etc.)
timeout | integer | yes | 60000ms | Functions and HTTP call timeout
keepalive | bool | yes | true | HTTP keepalive
keepalive_timeout | integer | yes | 60000ms | keepalive idle timeout
keepalive_pool | integer | yes | 5 | Connection pool limit

</div>
</body>
</html>
2. plugin implement
[...]

PR #5518, opened 2021-11-15
feat: add Apache OpenWhisk plugin

bzp2010:
Support for Apache OpenWhisk serverless platform.
Implement: #5485 

### Pre-submission checklist:


t/plugin/openwhisk.t
```diff
@@ -0,0 +1,209 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
+no_long_string();
+no_root_location();
```
spacewander, line 21: New test file should contain https://github.com/apache/apisix/blob/1018576e30477a9d1e70710386aac998a68acc68/t/wasm/route.t#L31 to reduce repeated fields.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+
+    return true
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() == "POST" then
+        if core.request.header(ctx, "Content-Type") ~= "application/json" then
+            core.log.error("only support json request body")
+            return 400, "only support json request body"
```
tokers, line 75: `415` code is used to describe this scenario: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+        namespace = {type = "string"},
+        action = {type = "string"},
+        result = {
+            type = "boolean",
+            default = true,
+        },
+        timeout = {
+            type = "integer",
+            minimum = 1000,
+            maximum = 60000,
```
spacewander, line 41: Please use the same timeout setting in azure-functions.

conf/config-default.yaml
```diff
@@ -338,6 +338,7 @@ plugins:                          # plugin list (sorted by priority)
   - traffic-split                  # priority: 966
   - redirect                       # priority: 900
   - response-rewrite               # priority: 899
+  - openwhisk                      # priority: 601
```
spacewander, line 341: Please use the similar priority like azure-functions

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+        "/actions/" .. conf.action
+
+    local httpc = http.new()
+    httpc:set_timeout(conf.timeout)
+
+    local res, err = httpc:request_uri(endpoint, params)
+
+    if not res or err then
+        core.log.error("failed to process openwhisk action, err: ", err)
+        return 503, "failed to process openwhisk action, err: " .. err
```
spacewander, line 111: See https://github.com/apache/apisix/pull/5479#discussion_r750813656

tokers, line 111: We may not expose this internal errors to the outside.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+        ssl_verify = conf.ssl_verify
+    }
+
+    if conf.keepalive then
+        params.keepalive_timeout = conf.keepalive_timeout
+        params.keepalive_pool = conf.keepalive_pool
+    end
+
+    -- OpenWhisk action endpoint
+    local endpoint = conf.api_host .. "/api/v1/namespaces/" .. conf.namespace ..
```
tokers, line 101: Maybe use the `table.concat`?

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+    if core.request.get_method() == "POST" then
+        if core.request.header(ctx, "Content-Type") ~= "application/json" then
+            core.log.error("only support json request body")
+            return 400, "only support json request body"
+        end
+    end
+
+    local params = {
+        method = "POST",
+        body = core.request.get_body(),
```
spacewander, line 81: See https://github.com/apache/apisix/pull/5479#discussion_r750883106

tokers, line 81: Although the request body for an API call won't be too large, we may still add some documents to tell the users that this is a potential problem.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+        return false, err
+    end
+
+    return true
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() == "POST" then
+        if core.request.header(ctx, "Content-Type") ~= "application/json" then
```
spacewander, line 73: Why check the Content-Type here?

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,121 @@
...
+    if not ok then
+        return false, err
+    end
+
+    return true
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() == "POST" then
```
spacewander, line 72: We should check the method to exclude GET? Or is there a special rule for PUT?

apisix/plugins/openwhisk.lua
bzp2010: Get it

apisix/plugins/openwhisk.lua
bzp2010: I think it is better to detect the request body except for GET requests. I will modify the logic and test here.

bzp2010: done

apisix/plugins/openwhisk.lua
bzp2010: I remember that previous tests showed that `..` is faster compared to `concat`.

apisix/plugins/openwhisk.lua
bzp2010: fix done

apisix/plugins/openwhisk.lua
bzp2010: fixed

apisix/plugins/openwhisk.lua
bzp2010: fixed

conf/config-default.yaml
bzp2010: Should I use the same `-1900` priority or add one `-1901`? Because they have similar functions and should not be enabled at the same time.

apisix/plugins/openwhisk.lua
bzp2010: It may not be able to complete the request within a maximum of 3000ms due to the need to download the runtime images or pull up a new service container. In addition, this timeout time acts on both HTTP requests from APISIX to OpenWhisk traffic entry and OpenWhisk internal action execution time consumption, if set to a lower value, it will cause all requests to fail.

BTW, the OpenWhisk‘s timeout support 1ms minimum, the range is 1ms - 60000ms. This is defined by OpenWhisk.

t/plugin/openwhisk.t
bzp2010: Yes, it can help us reduce duplicate code. However, can we support this default setting globally instead of adding our own block processor in each `.t` file.

apisix/plugins/openwhisk.lua
bzp2010: A reminder about the maximum request body size limit is currently written in the lua comments and documentation respectively.

bzp2010: A reminder about the maximum request body size limit is currently written in the lua comments and documentation respectively.

conf/config-default.yaml
spacewander: `-1901`.

apisix/plugins/openwhisk.lua
spacewander: Can you add your discovery as a comment?

bzp2010: Of course, I will add it to documents.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,123 @@
...
+        return false, err
+    end
+
+    return true
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() ~= "GET" then
+        if core.request.header(ctx, "Content-Type") ~= "application/json" then
```
tokers, line 73: `application/json; charset=utf-8` should also be supported.

bzp2010: Yes, I will change to `core.string.has_prefix` check `application/json` prefix.

bzp2010: @tokers done

t/plugin/openwhisk.t
spacewander: The legacy code may be a problem. But you can give it a try.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,124 @@
...
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() ~= "GET" then
+        local content_type = core.request.header(ctx, "Content-Type")
+        if not core.string.has_prefix(content_type, "application/json") then
+            core.log.error("only support json request body")
+            return 415
+        end
```
spacewander, line 77: We can only get the body after the valid check to avoid bypassing.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,124 @@
...
+
+    local res, err = httpc:request_uri(endpoint, params)
+
+    if not res or err then
+        core.log.error("failed to process openwhisk action, err: ", err)
+        return 503
+    end
+
+    -- setting response headers
+    core.response.set_header(res.headers)
```
spacewander, line 118: Should we handle HTTP2 like the azure version?

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,107 @@
+---
+title: serverless
```
spacewander, line 2: Wrong title.

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,107 @@
...
+
+The `openwhisk` plugin is used to support integration with the [Apache OpenWhisk](https://openwhisk.apache.org) serverless platform and can be set up on a route in place of Upstream, which will take over the request and send it to the OpenWhisk API endpoint.
+
+Users can call the OpenWhisk action via APISIX, pass the request parameters via JSON and get the response content.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| api_host | string | yes |   |   | OpenWhisk API host (eg. https://localhost:3233) |
```
spacewander, line 41: Can we use required/optional  like other plugins?

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,107 @@
...
+| ssl_verify | bool | no | true |   | Whether to verify the certificate |
+| service_token | string | yes |   |   | OpenWhisk ServiceToken (The format is `xxx:xxx`，Passed through Basic Auth when calling the API) |
+| namespace | string | yes |   |   | OpenWhisk  Namespace (eg. guest) |
+| action | string | yes |   |   | OpenWhisk Action (eg. hello) |
+| result | bool | no | true |   | Whether to get Action metadata (default to execute function and get response; false to get Action metadata but not execute Action, including runtime, function body, restrictions, etc.) |
+| timeout | integer | no | 60000ms | [1, 60000]ms | OpenWhisk Action and HTTP call timeout. |
```
spacewander, line 47: The default/min value of timeout doesn't match the schema

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,124 @@
...
+function _M.access(conf, ctx)
+    if core.request.get_method() ~= "GET" then
+        local content_type = core.request.header(ctx, "Content-Type")
+        if not core.string.has_prefix(content_type, "application/json") then
+            core.log.error("only support json request body")
+            return 415
+        end
+    end
+
+    -- Please avoid transferring large amounts of data or uploading files, which can
```
spacewander, line 80: The `core.request.get_body` will read data from file. This comment is incorrect. Please remove it.

t/core/request.t
```diff
@@ -437,3 +437,31 @@ c=z_z&v=x%20x
 nil
 --- error_log
 the post form is too large: request body in temp file not supported
+
+
+
+=== TEST 13: get_method
+--- config
+    location = /hello {
+        client_body_in_file_only clean;
```
spacewander, line 446: We don't need this directive, right?

bzp2010: https://github.com/apache/apisix/blob/93b33d8f5f4af73c621d52d017d114127593b936/apisix/plugins/openwhisk.lua#L72

This function is used here, and when I needed to use it, I realized that it was not wrapped in our core library, which would mix up the use of `ngx.req` and `core.request`, so I added this function and its tests.

spacewander: `client_body_in_file_only` is not relative to get_method.

apisix/plugins/openwhisk.lua
```diff
@@ -0,0 +1,124 @@
...
+    if not ok then
+        return false, err
+    end
+
+    return true
+end
+
+
+function _M.access(conf, ctx)
+    if core.request.get_method() ~= "GET" then
```
tokers, line 72: Is this enough to only read body if the request method is not `GET`, for instance, `HEAD` request should also not contain request body.

bzp2010:
## Update

Based on a tip from an Apache OpenWhisk community member in [https://github.com/apache/apisix/issues/5485#issuecomment-974602463](https://github.com/apache/apisix/issues/5485#issuecomment-974602463) and more document reading, I found that the current code implementation has some limitations and I will modify part of the implementation.

cc @spacewander @tokers

apisix/core/request.lua
```diff
@@ -285,4 +285,9 @@ function _M.get_http_version()
     return ngx.req.http_version()
 end
 
+
+function _M.get_method()
```
membphis, line 289: style:

`_M.get_method = ngx.req.get_method` is better style.

apisix/plugins/openwhisk.lua
membphis: current way is better

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,107 @@
...
+    "uri": "/hello",
+    "plugins": {
+        "openwhisk": {
+            "api_host": "http://localhost:3233",
+            "service_token": "23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP",
+            "namespace": "guest",
+            "action": "test"
+        }
+    },
+    "upstream": {
```
membphis, line 92: we should remove `upstream`

docs/zh/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,107 @@
...
+    "uri": "/hello",
+    "plugins": {
+        "openwhisk": {
+            "api_host": "http://localhost:3233",
+            "service_token": "23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP",
+            "namespace": "guest",
+            "action": "test"
+        }
+    },
+    "upstream": {
```
membphis, line 92: ditto

apisix/core/request.lua
membphis: the `_M.get_http_version` has the same issue, we can fix it in a new PR

apisix/plugins/openwhisk.lua
bzp2010: Here it is actually a requirement of OpenWhisk and does not pose a security problem.
When there is an incorrect request body, OpenWhisk will return the following.

```json
{
    "code": "M97oTvPn4Lb4TqBREemUYa5PNBImoc6F",
    "error": "The request content was malformed:\nUnexpected character 'a' at input index 0 (line 1, position 1), expected JSON Value:\naaa\n^\n"
}
```

Can we remove this check in order to use OpenWhisk's input validation directly?

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

docs/zh/latest/plugins/openwhisk.md
bzp2010: fixed

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

apisix/core/request.lua
bzp2010: fixed

t/core/request.t
bzp2010: removed

apisix/plugins/openwhisk.lua
bzp2010: removed

apisix/plugins/openwhisk.lua
bzp2010: From current tests, OpenWhisk does not generate these headers.

apisix/plugins/openwhisk.lua
spacewander: OK, we can remove it.

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,103 @@
...
+The `openwhisk` plugin is used to support integration with the [Apache OpenWhisk](https://openwhisk.apache.org) serverless platform and can be set up on a route in place of Upstream, which will take over the request and send it to the OpenWhisk API endpoint.
+
+Users can call the OpenWhisk action via APISIX, pass the request parameters via JSON and get the response content.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| api_host | string | required |   |   | OpenWhisk API host (eg. https://localhost:3233) |
+| ssl_verify | bool | no | optional |   | Whether to verify the certificate |
```
spacewander, line 42: Missing default?

tzssangglass: use `boolean` not `bool`?

apisix/plugins/openwhisk.lua
tzssangglass: I would suggest not using exclusion, just use `==` to care about what methods that needs to be cared about.

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

apisix/plugins/openwhisk.lua
bzp2010: @tzssangglass Referer here [https://github.com/apache/apisix/pull/5518#discussion_r754043953](https://github.com/apache/apisix/pull/5518#discussion_r754043953), these codes will be remove.

apisix/plugins/openwhisk.lua
bzp2010: removed

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,103 @@
...
+The `openwhisk` plugin is used to support integration with the [Apache OpenWhisk](https://openwhisk.apache.org) serverless platform and can be set up on a route in place of Upstream, which will take over the request and send it to the OpenWhisk API endpoint.
+
+Users can call the OpenWhisk action via APISIX, pass the request parameters via JSON and get the response content.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| api_host | string | required |   |   | OpenWhisk API host (eg. https://localhost:3233) |
+| ssl_verify | boolean | no | optional | true | Whether to verify the certificate |
```
spacewander, line 42: Unexpected "no"?

docs/en/latest/plugins/openwhisk.md
```diff
@@ -0,0 +1,103 @@
...
+
+Finally, you can send a request to this route and you will get the following response. And you can disable it by removing the openwhsik plugin from the route.
+
+```json
+{"ready": true}
+```
+
+## Notice
+
+When you need to submit data, you need to use a request body in JSON format. Also you should pay attention to the request body size, request body exceeding the `client_body_buffer_size` setting will be discarded completely, its size defaults to 8KiB.
```
spacewander, line 103: The notice is incorrct. We can submit data larger than 8KiB.

bzp2010: removed

docs/en/latest/plugins/openwhisk.md
bzp2010: fixed

t/plugin/openwhisk.t
bzp2010: added

t/plugin/openwhisk.t
```diff
@@ -19,7 +19,16 @@ use t::APISIX 'no_plan';
 repeat_each(1);
 no_long_string();
 no_root_location();
-run_tests;
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if ((!defined $block->error_log) && (!defined $block->no_error_log)) {
```
spacewander, line 26: Should do the same with `GET /t`

bzp2010: added

t/plugin/openwhisk.t
```diff
@@ -0,0 +1,243 @@
...
+
+=== TEST 5: hit route (with GET request)
+--- request
+GET /hello
+--- response_body chomp
+{"hello":"test"}
+
+
+
+=== TEST 6: hit route (with POST without non-json format request body)
```
tokers, line 132:
```suggestion
=== TEST 6: hit route (with POST method and non-json format request body)
```

t/plugin/openwhisk.t
```diff
@@ -0,0 +1,243 @@
...
+
+
+
+=== TEST 9: hit route (with non-existent action)
+--- request
+POST /hello
+{"name": "world"}
+--- more_headers
+Content-Type: application/json
+--- error_code: 404
```
tokers, line 196: Also check out the response body?

bzp2010: Ok, it necessary, I will add it.

bzp2010: added

t/plugin/openwhisk.t
bzp2010: fixed

spacewander (approved):

membphis (approved):

bzp2010:
## Update
Here I used force push to rebase the code to the latest code on the master branch, without code changes. I also kept the original commits as a backup, so if you don't want to see the force push used, I will restore the original commits.

bzp2010:
Hi, @spacewander @membphis @tzssangglass @tokers. The CI error has been fixed.

membphis:
we also need to update the REAME.md @bzp2010

membphis:
> we also need to update the REAME.md @bzp2010

please create a new issue about this, do this later.

membphis:
> > we also need to update the REAME.md @bzp2010
> 
> please create a new issue about this, do this later.

@bzp2010 any news?

PR #5529, opened 2021-11-16
fix(plugin): check plugin metadata  in the DP when there is a new conf

wangxiaochuang:
plugin metadata do not check it's validation

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -186,7 +186,7 @@ local function load_full_data(self, dir_res, headers)
         end
 
         if data_valid and self.checker then
-            data_valid, err = self.checker(item.value)
+            data_valid, err = self.checker(item.value, item.key)
             if not data_valid then
                 log.error("failed to check item data of [", self.key,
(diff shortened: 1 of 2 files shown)


apisix/plugin.lua
```diff
@@ -556,7 +556,14 @@ function _M.init_worker()
     end
 
     local plugin_metadatas, err = core.config.new("/plugin_metadata",
-        {automatic = true}
+        {
```
spacewander, line 559: Could you add a test for it?
You can use
```
local etcd = require("apisix.core.etcd")
assert(etcd.set("/ab", "ab"))
```
to set etcd directly.

apisix/core/config_etcd.lua
```diff
@@ -186,7 +186,7 @@ local function load_full_data(self, dir_res, headers)
         end
 
         if data_valid and self.checker then
-            data_valid, err = self.checker(item.value)
+            data_valid, err = self.checker(item.value, item.key)
```
spacewander, line 189: We should also update `config_yaml.lua` & add a test in https://github.com/apache/apisix/blob/master/t/config-center-yaml/plugin-metadata.t

PR #5537, opened 2021-11-17
chore: check the error from get_post_args

spacewander:

### Pre-submission checklist:

### DIFF
--- a/apisix/core/request.lua
+++ b/apisix/core/request.lua
@@ -161,7 +161,12 @@ function _M.get_post_args(ctx)
 
         -- use 0 to avoid truncated result and keep the behavior as the
         -- same as other platforms
-        local args = req_get_post_args(0)
+        local args, err = req_get_post_args(0)
+        if not args then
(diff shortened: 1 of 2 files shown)


tokers (approved):

bzp2010:
As far as I know `ngx.req.get_post_args` only supports form submissions using `x-www-form-urlencoded`, not for `form-data` and `json` etc. When the request is for the latter two cases, it will be recognized directly as a string or this (eg. `{"{\"a\":\"a\"}":true}`). 

Do we have to do this test?

spacewander:
> 
> 
> As far as I know `ngx.req.get_post_args` only supports form submissions using `x-www-form-urlencoded`, not for `form-data` and `json` etc. When the request is for the latter two cases, it will be recognized directly as a string or this (eg. `{"{\"a\":\"a\"}":true}`).
> 
> Do we have to do this test?

`ngx.req.get_post_args` doesn't take care of the content-type.

tzssangglass (approved):

PR #5578, opened 2021-11-23
feat(ua-restriction): sync message to rejected_msg

Xunzhuo:
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/ua-restriction.lua
+++ b/apisix/plugins/ua-restriction.lua
@@ -42,10 +42,10 @@ local schema = {
             type = "array",
             minItems = 1
         },
-        message = {
+        rejected_msg = {
             type = "string",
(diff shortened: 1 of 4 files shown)


spacewander (changes requested):
A break change which just renames the field is unacceptable.

Xunzhuo:
I also change the limit of max-length @spacewander , and I want to know why the plugins each other are not followed the same name rule?

Xunzhuo:
@spacewander And one more thing needs answers too, the max-length of rejected_msg should be 256 or 512 or 1024 or just not limited? I can see some plugins` max-length is 1024 , but some other plugins` max-length is 256, and some plugins just do not have the max length limit. Should we have a unique Rule?

PR #5589, opened 2021-11-23
fix: invalid error after passive health check is changed

### LINKED ISSUE #4077: passive healthcheck not working after configuration changing
### Issue description
When both active and passive configuration are updated from only active configuration, the http statuses of passive become an empty table which causes passive healthcheck not working. 

### Environment
* apisix version: 2.0
* OS: 3.10.0-693.el7.x86_64
* OpenResty version: openresty/1.17.8.2
* etcd version: 3.4.13
* apisix-dashboard version: 2.0

1. Set a route with two upstreams(/test.do) and healthcheck active configuration(/check.do). 
2. Send a request to trigger creating new checker.
3. Update this route only by changing healthcheck configuration to both active and passive. Let one of the upstream return 500 code to act like a unnormal node. Send some requests, but the unnormal node cannot be set to be unhealthy even reaching the passive.unhealthy.http_failures.

### What's the actual result?
I prints some debug info inside this function: https://github.com/Kong/lua-resty-healthcheck/blob/master/lib/resty/healthcheck.lua#L1372
I prints out the "defaults" table at line 1443.

After doing Step 2, the type of defaults.checks.passive.unhealthy.http_statuses changes from array to dict!

defaults.checks.passive.unhealthy.http_statuses before to_set() function:
{
     1 = 429,
     2 = 500,
     3 = 503,
}

defaults.checks.passive.unhealthy.http_statuses after to_set() function:
{
     429 = true,
     500 = true,
     503 = true,
}

It seems that to_set() changes the "self" object, result in changing "defaults" as well.

In my opinion, the "defaults" object should not be changed. Maybe there's some problem with fill_in_settings() function at line 1377. Or am I doing something wrong during these operations?

shuaijinchao:
FIX #4077

### Pre-submission checklist:


tzssangglass (approved):

membphis (approved):

bisakhmondal (approved):

t/admin/health-check.t
```diff
@@ -469,17 +434,64 @@ GET /t
...
 --- error_code: 400
 --- response_body
 {"error_msg":"invalid configuration: property \"upstream\" validation failed: property \"checks\" validation failed: object matches none of the requireds: [\"active\"] or [\"active\",\"passive\"]"}
---- no_error_log
-[error]
 
 
 
-=== TEST 13: number type timeout
+=== TEST 13: only active
```
spacewander, line 443:
```suggestion
=== TEST 13: change active & passive to only active
```

apisix/schema_def.lua
```diff
@@ -259,24 +259,38 @@ local health_checker = {
...
+            default = {
+                type = "http",
+                healthy = {
+                    http_statuses = { 200, 201, 202, 203, 204, 205, 206, 207, 208, 226,
+                                      300, 301, 302, 303, 304, 305, 306, 307, 308 },
+                    successes = 0,
+                },
+                unhealthy = {
+                    http_statuses = { 429, 500, 503 },
+                    tcp_failures = 0,
```
spacewander, line 290: Is it necessary to set a zero default? Do we need to update the doc?

shuaijinchao: Changing the value to 0 will disable the passive health check, which is necessary when only the active check is configured, documentation needs to be updated

t/admin/health-check.t
shuaijinchao: The purpose of this test is to verify that the default passive configuration will be added in the case of active only. I think this title description is not very accurate, what do you think?

apisix/schema_def.lua
shuaijinchao: document updated

spacewander (approved):

PR #5600, opened 2021-11-24
feat(ext-plugin): support to get request body

tzssangglass:

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -281,7 +281,7 @@ local function handle_extra_info(ctx, input)
 
         local err
         res, err = core.request.get_body()
-        if not res then
+        if err then
             core.log.error("failed to read request body: ", err)
         end
 
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -274,7 +274,6 @@ local function handle_extra_info(ctx, input)
         local var_name = var_req:Name()
         if var_name == "request_body" then
             res = core.request.get_body()
-            ngx.log(ngx.WARN, "res : ", require("lualib.inspect")(res))
         else
             res = ctx.var[var_name]
         end
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -279,7 +279,12 @@ local function handle_extra_info(ctx, input)
         local reqbody_req = extra_info_reqbody.New()
         reqbody_req:Init(info.bytes, info.pos)
 
-        res = core.request.get_body()
+        local err
+        res, err = core.request.get_body()
+        if not res then
+            core.log.error("failed to read request body: ", err)
+        end
+
     else
         return nil, "unsupported info type: " .. info_type
     end
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -28,6 +28,7 @@ local extra_info = require("A6.ExtraInfo.Info")
 local extra_info_req = require("A6.ExtraInfo.Req")
 local extra_info_var = require("A6.ExtraInfo.Var")
 local extra_info_resp = require("A6.ExtraInfo.Resp")
+local extra_info_reqbody = require("A6.ExtraInfo.ReqBody")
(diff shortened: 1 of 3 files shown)


tzssangglass:
I'm thinking that this implementation might not be good enough, because usually the user needs to not only read the request body, but also rewrite it.

t/lib/ext-plugin.lua
```diff
@@ -169,6 +169,8 @@ function _M.go(case)
             local entry = call_req:Args(1)
             assert(entry:Name() == "x")
             assert(entry:Value() == "z")
+        elseif case.get_request_body then
+            assert(call_req:Method() == a6_method.POST)
```
spacewander, line 173: Do we need to check the request_body's value?

t/plugin/ext-plugin/request-body.t
```diff
@@ -0,0 +1,177 @@
...
+    }
+--- error_code: 405
+--- grep_error_log eval
+qr/send extra info req successfully/
+--- grep_error_log_out
+send extra info req successfully
+
+
+
+=== TEST 4: request body(json)
```
spacewander, line 155: Better to add a test which doesn't have a value

tzssangglass: I have tried to emulate this test case：https://github.com/apache/apisix/blob/4b94b84f92d1cb6cb58e52075637a42ed76457d8/t/core/request.t#L414-L439 but failed.

should I send an empty body to test?

t/lib/ext-plugin.lua
tzssangglass: I want to validate different content-type, so the value of request body is not fixed.

tzssangglass: actions.result would check the value in request body?

t/plugin/ext-plugin/request-body.t
tzssangglass: add a test case of request body is nil

tzssangglass:
dealy to merge, is: https://github.com/api7/ext-plugin-proto/blob/main/lua/A6/ExtraInfo/ReqBody.lua used for request body?

tzssangglass:
> dealy to merge, is: [api7/ext-plugin-proto@`main`/lua/A6/ExtraInfo/ReqBody.lua](https://github.com/api7/ext-plugin-proto/blob/main/lua/A6/ExtraInfo/ReqBody.lua?rgh-link-date=2021-11-25T08%3A25%3A29Z) used for request body?

ok, after review:https://lists.apache.org/thread/m5g44gtywx05x4s262rn7qg87oxscy0v, the ReqBody in ExtraInfo is used to  holds the request body.

tzssangglass:
I will update this PR.

t/plugin/ext-plugin/request-body.t
spacewander: Yes

apisix/plugins/ext-plugin/init.lua
```diff
@@ -273,6 +274,12 @@ local function handle_extra_info(ctx, input)
 
         local var_name = var_req:Name()
         res = ctx.var[var_name]
+    elseif info_type == extra_info.ReqBody then
+        local info = req:Info()
+        local reqbody_req = extra_info_reqbody.New()
+        reqbody_req:Init(info.bytes, info.pos)
+
+        res = core.request.get_body()
```
spacewander, line 282: Better to log down err if it is not nil.

tzssangglass: done

apisix/plugins/ext-plugin/init.lua
```diff
@@ -279,7 +279,12 @@ local function handle_extra_info(ctx, input)
         local reqbody_req = extra_info_reqbody.New()
         reqbody_req:Init(info.bytes, info.pos)
 
-        res = core.request.get_body()
+        local err
+        res, err = core.request.get_body()
+        if not res then
```
spacewander, line 284: We need to check the err instead, as the body can be nil when there is no body.

tzssangglass: update

spacewander (approved):

shuaijinchao (approved):

PR #5616, opened 2021-11-25
feat: faas plugin refactoring with url path forwarding

bisakhmondal:

This PR is based on the idea of creating a single generic upstream where each faas plugin implements an authorization schema and perform the header processing.

GOAL: Reduce the same code over multiple plugins (mainly faas)

**How developers should write new (faas) plugins:**
1. They should define plugin-specific authorization schema & metadata schema(if required)
2. Write a `request_processor` function on the plugin lua file that take `plugin conf, ngx context and the URL params (that contains headers, body, urlpath, query params etc. etc.)". The function can perform a series of operations including update/adding the necessary authorization headers, performing encoding/decoding (base64 or anything) on the body.
3. And yes, that's it. (see the azure-functions Lua code for example)

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/serverless/generic-upstream.lua
+++ b/apisix/plugins/serverless/generic-upstream.lua
@@ -72,10 +72,10 @@ return function(plugin_name, version, priority, request_processor, authz_schema,
         local url_decoded = url.parse(conf.function_uri)
         local path = url_decoded.path or "/"
 
-        if ctx.curr_req_matched and ctx.curr_req_matched["*"] then
-            local end_path = ctx.curr_req_matched["*"]
+        if ctx.curr_req_matched and ctx.curr_req_matched[":ext"] then
+            local end_path = ctx.curr_req_matched[":ext"]
 
-            if path:sub(-1, -1) == "/" or end_path:sub(1, 1) == "/" then
+            if path:byte(-1) == string.byte("/") or end_path:byte(1) == string.byte("/") then
                 path = path .. end_path
             else
                 path = path .. "/" .. end_path
@@ -83,7 +83,7 @@ return function(plugin_name, version, priority, request_processor, authz_schema,
         end
 
 
-        headers["host"] = nil
+        headers["host"] = url_decoded.host
         local params = {
             method = ngx.req.get_method(),
             body = req_body,
--- a/apisix/plugins/serverless/generic-upstream.lua
+++ b/apisix/plugins/serverless/generic-upstream.lua
@@ -15,6 +15,8 @@
 -- limitations under the License.
 
 local ngx  = ngx
+local require = require
+local type = type
 
 return function(plugin_name, version, priority, header_processor, authz_schema, metadata_schema)
     local core = require("apisix.core")
@@ -59,7 +61,6 @@ return function(plugin_name, version, priority, header_processor, authz_schema, 
         local uri_args = core.request.get_uri_args(ctx)
         local headers = core.request.headers(ctx) or {}
 
-        -- core.log.error("must---> ",core.json.delay_encode(ctx.curr_req_matched, true))
         local req_body, err = core.request.get_body()
 
         if err then
@@ -70,10 +71,10 @@ return function(plugin_name, version, priority, header_processor, authz_schema, 
         -- forward the url path came through the matched uri
         local url_decoded = url.parse(conf.function_uri)
         local path = url_decoded.path or "/"
-        
+
         if ctx.curr_req_matched and ctx.curr_req_matched["*"] then
             local end_path = ctx.curr_req_matched["*"]
-            
+
             if path:sub(-1, -1) == "/" or end_path:sub(1, 1) == "/" then
                 path = path .. end_path
             else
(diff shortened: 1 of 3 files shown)


apisix/plugins/serverless/generic-upstream.lua
```diff
@@ -0,0 +1,134 @@
...
+        end
+
+        -- forward the url path came through the matched uri
+        local url_decoded = url.parse(conf.function_uri)
+        local path = url_decoded.path or "/"
+
+        if ctx.curr_req_matched and ctx.curr_req_matched["*"] then
+            local end_path = ctx.curr_req_matched["*"]
+
+            if path:sub(-1, -1) == "/" or end_path:sub(1, 1) == "/" then
```
spacewander, line 78: Using `:byte` would be better.
BTW, it is not a good idea to refactor & add feature in the same PR.

apisix/plugins/serverless/generic-upstream.lua
```diff
@@ -0,0 +1,134 @@
...
+        if err then
+            core.log.error("error while reading request body: ", err)
+            return 400
+        end
+
+        -- forward the url path came through the matched uri
+        local url_decoded = url.parse(conf.function_uri)
+        local path = url_decoded.path or "/"
+
+        if ctx.curr_req_matched and ctx.curr_req_matched["*"] then
```
spacewander, line 75: We should use `:ext` instead of `*`

t/plugin/azure-functions.t
```diff
@@ -375,3 +381,43 @@ ngx.say("Authz-Header - " .. headers["x-functions-key"] or "")
...
+            -- creating a semi path route
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "plugins": {
+                            "azure-functions": {
+                                "function_uri": "http://localhost:8765/api"
+                            }
+                        },
+                        "uri": "/azure/**"
```
spacewander, line 401:
```suggestion
                        "uri": "/azure/*"
```

apisix/plugins/serverless/generic-upstream.lua
bisakhmondal: for byte you mean like this 
`if path:byte(-1, -1) == string.byte"/" or end_path:byte(1, 1) == string.byte"/" then`

The PRs are interdependent and require one PR to be merged before another. Do you think I should split it up into 2 PRs?

spacewander: Like `if path:byte(-1) == string.byte("/") or end_path:byte(1) == string.byte("/") then`.

We can accept this one now.
Next time, please don't mix up refactor & new feature.

bisakhmondal: Cool! Thanks boss

t/plugin/azure-functions.t
```diff
@@ -375,3 +381,43 @@ ngx.say("Authz-Header - " .. headers["x-functions-key"] or "")
...
+            )
+            if code >= 300 then
+                ngx.status = code
+                ngx.say("fail")
+                return
+            end
+
+            ngx.say(body)
+
+            local code, _, body = t("/azure/httptrigger", "GET")
```
spacewander, line 412: Let's add another test which contains multiple levels of paths, like `/azure/http/trigger`.

apisix/plugins/serverless/generic-upstream.lua
bisakhmondal: Agree. I was using `**` that's why the extra arg was received through curr_req_matched. Updated

t/plugin/azure-functions.t
bisakhmondal: Done

t/plugin/azure-functions.t
bisakhmondal: Done.

bisakhmondal: As discussed offline the following cases have been covered
multiple levels of paths
multiple slashes in the paths
the wildcard matches an empty part, like /azure

apisix/plugins/serverless/generic-upstream.lua
bisakhmondal: Done

t/plugin/azure-functions.t
```diff
@@ -375,3 +393,100 @@ ngx.say("Authz-Header - " .. headers["x-functions-key"] or "")
...
+            ngx.print(body)
+        }
+    }
+--- response_body
+passed
+invocation /api/httptrigger successful
+
+
+
+=== TEST 11: check multilevel url path forwarding
```
spacewander, line 439: Next time, we can use a table drive test for those cases. See https://github.com/apache/apisix/blob/7242216076f1df4225ab5d2d742ed7fb23383306/t/plugin/gzip.t#L498

spacewander (approved):

tzssangglass (approved):

PR #5675, opened 2021-12-02
feat: support resolve default value when environment not set

kevinw66:
Thanks for review my first pull request for apisix.
And this is my first time use lua. Give me some advices if possible, thanks.

This is about read default value when ENV not set, I think this may helpful when using k8s in different environment and did not want to set so many ENV variable

result:
etcd.host = http://${{ETCD_HOST}}:2379

image

etcd.host = http://${{ETCD_HOST:123ss}}:2379
image

etcd.host = http://${{ETCD_HOST:123ss:123aa}}:2379
image

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -70,7 +70,7 @@ local function resolve_conf_var(conf)
                 local default
                 if i and j then
                     default = string.sub(var, i+1, j)
-                    var = string.sub(var, 1, i)
+                    var = string.sub(var, 1, i-1)
                 end
 
                 local v = getenv(var) or default
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -65,8 +65,15 @@ local function resolve_conf_var(conf)
             local var_used = false
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
-            local new_val = val:gsub("%$%{%{%s*([%w_]+)%s*%}%}", function(var)
-                local v = getenv(var)
+            local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%w+]*)%s*%}%}", function(var)
+                local i, j = string.find(var, "%:%w+")
+                local default
+                if i and j then
+                    default = string.sub(var, i+1, j)
+                    var = string.sub(var, 1, i)
+                end
+
+                local v = getenv(var) or default
                 if v then
                     if not exported_vars then
                         exported_vars = {}
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -65,11 +65,11 @@ local function resolve_conf_var(conf)
             local var_used = false
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
-            local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%w+]*)%s*%}%}", function(var)
-                local i, j = var:find("%:%w+")
+            local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%:[%w%-%.%:%?%/=]+]?)%s*%}%}", function(var)
+                local i, j = var:find("%:%:[%w%-%.%:%?%/=]*")
                 local default
                 if i and j then
-                    default = var:sub(i + 1, j)
+                    default = var:sub(i + 2, j)
                     var = var:sub(1, i - 1)
                 end
 
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -66,11 +66,11 @@ local function resolve_conf_var(conf)
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
             local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%w+]*)%s*%}%}", function(var)
-                local i, j = string.find(var, "%:%w+")
+                local i, j = var:find("%:%w+")
                 local default
                 if i and j then
-                    default = string.sub(var, i+1, j)
-                    var = string.sub(var, 1, i-1)
+                    default = var:sub(i + 1, j)
+                    var = var:sub(1, i - 1)
                 end
 
                 local v = getenv(var) or default
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -66,7 +66,7 @@ local function resolve_conf_var(conf)
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
             local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%:[%w%-%.%:%?%/=]+]?)%s*%}%}", function(var)
-                local i, j = var:find("%:%:[%w%-%.%:%?%/=]*")
+                local i, j = var:find("%:%:[%w%-%.%:%?%/=]+")
                 local default
                 if i and j then
                     default = var:sub(i + 2, j)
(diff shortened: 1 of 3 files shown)


apisix/cli/file.lua
```diff
@@ -65,8 +65,15 @@ local function resolve_conf_var(conf)
             local var_used = false
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
-            local new_val = val:gsub("%$%{%{%s*([%w_]+)%s*%}%}", function(var)
-                local v = getenv(var)
+            local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%w+]*)%s*%}%}", function(var)
```
spacewander, line 68: Could you add a test in https://github.com/apache/apisix/blob/b8daf65b3c0690138feccb2ce399fc58349ac7ee/t/cli/test_main.sh#L311?
Also need to update the doc in https://github.com/apache/apisix/blob/11433d740c5e19b757c61f59842d6de64e27fba8/conf/config.yaml#L24

kevinw66: OK, I'll try

kevinw66: Done

t/cli/test_main.sh
```diff
@@ -310,6 +310,35 @@ fi
...
+echo '
+etcd:
+    host:
+        - "http://${{ETCD_HOST:192.168.1.1}}:${{ETCD_PORT}}"
+' > conf/config.yaml
+
+ETCD_PORT=2379 make init
+
+if ! grep "env ETCD_HOST=192.168.1.1;" conf/nginx.conf > /dev/null; then
+    echo "failed: support default value when environment not set"
```
tzssangglass, line 323:
```suggestion
    echo "failed: should use default value when environment not set"
```

t/cli/test_main.sh
```diff
@@ -310,6 +310,35 @@ fi
...
+echo '
+etcd:
+    host:
+        - "http://${{ETCD_HOST:192.168.1.1}}:${{ETCD_PORT}}"
+' > conf/config.yaml
+
+ETCD_HOST=127.0.0.1 ETCD_PORT=2379 make init
+
+if ! grep "env ETCD_HOST=127.0.0.1;" conf/nginx.conf > /dev/null; then
+    echo "failed: support default value when environment not set"
```
tzssangglass, line 336:
```suggestion
    echo "failed: should use  environment var of ETCD_HOST"
```

t/cli/test_main.sh
```diff
@@ -310,6 +310,35 @@ fi
 
 echo "pass: support environment variables in local_conf"
 
+# support default value when environment not set
+echo '
+etcd:
+    host:
+        - "http://${{ETCD_HOST:192.168.1.1}}:${{ETCD_PORT}}"
```
tzssangglass, line 317: Can we use an other name? I'm worried that ETCD_HOST may have already been set with variables.

t/cli/test_main.sh
```diff
@@ -310,6 +310,35 @@ fi
...
+' > conf/config.yaml
+
+ETCD_PORT=2379 make init
+
+if ! grep "env ETCD_HOST=192.168.1.1;" conf/nginx.conf > /dev/null; then
```
tzssangglass, line 322: I think it would be more appropriate to check if etcd is connected properly from the APISIX error.log.

tzssangglass (approved):

t/cli/test_main.sh
kevinw66: But ETCD_HOST is used in previous tests, so I reuse this variable

tzssangglass: OK, let follow this.

t/cli/test_main.sh
kevinw66: When etcd cannot connect correctly, I found that error will print in console, not error.log

kevinw66:
The github ci test result in `Error: [error] open() "/home/runner/work/apisix/apisix/logs/nginx.pid" failed (2: No such file or directory)` is this my code problem ? I saw it failed in other testcases.

conf/config.yaml
```diff
@@ -30,6 +30,16 @@
 # And then run `export ETCD_HOST=$your_host` before `make init`.
 #
 # If the configured environment variable can't be found, an error will be thrown.
+#
+# Also, If you want to use default value when the environment variable not set,
+# Use `${{VAR::default_value}}` instead. For instance:
```
spacewander, line 35: I am curious why `::` is chosen as the separator. It is different from the common way. Is there a special reason behind it?

apisix/cli/file.lua
```diff
@@ -65,8 +65,15 @@ local function resolve_conf_var(conf)
             local var_used = false
             -- we use '${{var}}' because '$var' and '${var}' are taken
             -- by Nginx
-            local new_val = val:gsub("%$%{%{%s*([%w_]+)%s*%}%}", function(var)
-                local v = getenv(var)
+            local new_val = val:gsub("%$%{%{%s*([%w_]+[%:%:[%w%-%.%:%?%/=]+]?)%s*%}%}", function(var)
```
spacewander, line 68: Why do we need a regex to check the default variable?

conf/config.yaml
kevinw66: I want to use `:` in first time, But sometimes default value may contains it, like etcd.host=${{ETCD_HOST::demo.domain.com:1234}}

apisix/cli/file.lua
kevinw66: If we don't check, It will fail to match when config contains default value.

kevinw66:
BTW, Can you help me to check the ut? I'm confusing about the error log, and it cannot run in my local machine right now.

conf/config.yaml
spacewander: What about using `:=`? In bash, it means "Assign  Default Values".

apisix/cli/file.lua
spacewander: Sure. But we don't need to be too strict with `[%w%-%.%:%?%/=]+`? What if people want to use space in the default value?

spacewander:
Solving with https://github.com/apache/apisix/pull/5699

t/cli/test_main.sh
```diff
@@ -310,6 +310,45 @@ fi
...
+' > conf/config.yaml
+
+TEST_ENV=127.0.0.1 make init
+
+if ! grep "env TEST_ENV=127.0.0.1;" conf/nginx.conf > /dev/null; then
+    echo "failed: should use default value when environment not set"
+    exit 1
+fi
+
+echo "pass: should use default value when environment not set"
```
tzssangglass, line 350: we can use `support default value when environment not set` here. Earlier I suggested taht change `echo "failed: support default value when environment not set"` to `echo "failed: should use default value when environment not set"` above is to output exactly that failed test case when it fails. If the error output is the same for all three tests above, then when an error occurs we need to take the time to see which one is actually wrong.

I think you can tweak the error output of the test case a little.

kevinw66: Done

conf/config.yaml
kevinw66: Done

apisix/cli/file.lua
kevinw66: Done

spacewander (approved):

moonming:
@kevinw66 great work 👍

tzssangglass (approved):

Issue #5417, opened 2021-11-04
bug: auto-generated apisix id is likely to be duplicated in containerized environments


tokers:
### Issue description

Currently, Apache APISIX will generate an id while initializing if the user doesn't specify one, and it relies on the [lua-resty-jit-uuid](https://github.com/thibaultcha/lua-resty-jit-uuid) library but without an explicit seed.

https://github.com/apache/apisix/blob/4dafab5afa3293b3d72007517246e01da385f8ef/apisix/core/id.lua#L76-L78

While the jit-uuid library creates the seed by the process id and the time in ngx_lua context.

https://github.com/thibaultcha/lua-resty-jit-uuid/blob/82538049040ae85ff880b79886f21d8593140c7d/lib/resty/jit-uuid.lua#L53-L54

However, in a containerized environment, the process id (the master process) might be the same, i.e. the No. 1 process, also, if users try to deploy Apache APISIX clusters on Kubernetes through the `Deployment` resource, the time might be the same for several Pods, since `ngx.time` doesn't have enough precisions (only at milliseconds level). so the generated apisix id might be duplicated, and if the id is critical, this may cause some fatal problems in the business scenarios.

### Environment

- apisix version (cmd: `apisix version`):
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

### Steps to reproduce

N/A

### Actual result

N/A

### Error log

N/A

spacewander:
Maybe we can suffix some rand bytes from https://github.com/thibaultcha/lua-resty-jit-uuid/issues/19#issuecomment-521076462

tokers:
> Maybe we can suffix some rand bytes from [thibaultcha/lua-resty-jit-uuid#19 (comment)](https://github.com/thibaultcha/lua-resty-jit-uuid/issues/19#issuecomment-521076462)

Good idea.

spacewander:
@leslie-tsang 
Would you like to have a try? Thanks!

leslie-tsang:
> @leslie-tsang Would you like to have a try? Thanks!

Sure.

leslie-tsang:
[...]

PR #5682, opened 2021-12-03
fix(patch): add global `math.randomseed` patch support

leslie-tsang:

Kong resution Ref to #5417 
### Pre-submission checklist:

### DIFF
--- a/apisix/patch.lua
+++ b/apisix/patch.lua
@@ -20,18 +20,24 @@ local ipmatcher = require("resty.ipmatcher")
 local socket = require("socket")
 local unix_socket = require("socket.unix")
 local ssl = require("ssl")
+local ngx = ngx
 local get_phase = ngx.get_phase
 local ngx_socket = ngx.socket
 local original_tcp = ngx.socket.tcp
 local original_udp = ngx.socket.udp
 local concat_tab = table.concat
+local debug = debug
 local new_tab = require("table.new")
 local log = ngx.log
 local WARN = ngx.WARN
 local ipairs = ipairs
 local select = select
 local setmetatable = setmetatable
+local string = string
+local table = table
 local type = type
+local tonumber = tonumber
+local tostring = tostring
 
 
 local config_local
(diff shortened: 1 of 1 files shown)


leslie-tsang:
@spacewander Please take a look, Thanks !

spacewander:
Please fix the lint.

apisix/patch.lua
```diff
@@ -86,6 +92,53 @@ do
 end
 
 
+-- Inspired by kong.globalpatches
+do -- `_G.math.randomseed` patch
```
moonming, line 96: why we need this patch?

moonming (changes requested):

apisix/patch.lua
```diff
@@ -86,6 +92,53 @@ do
...
+            return
+        end
+
+        -- get randomseed
+        local bytes = resty_random.bytes(8)
+        if bytes then
+            log(ngx.INFO, "seeding from resty.random.bytes")
+
+            local t = {}
+            for i = 1, #bytes do
```
spacewander, line 119: We can check if `#str > 12` here. And we need to comment why 12 is chosen.

apisix/patch.lua
tokers: @moonming Please see https://github.com/apache/apisix/issues/5417 for details.

apisix/patch.lua
spacewander: `for i = 1, min(#bytes, 12) do` should be already OK?

.luacheckrc
```diff
@@ -2,6 +2,13 @@ std = "ngx_lua"
 unused_args = false
 redefined = false
 max_line_length = 100
+
 exclude_files = {
     "apisix/cli/ngx_tpl.lua",
 }
+
+files["apisix/patch.lua"] = {
```
spacewander, line 10: We can solve in this way, which is consistent:
https://github.com/apache/apisix/blob/6e09d56e9c93666b52fccdeb6e1d3fb1178b678c/apisix/patch.lua#L310-L311

apisix/patch.lua
```diff
@@ -86,6 +93,64 @@ do
 end
 
 
+do -- `_G.math.randomseed` patch
```
spacewander, line 96:
```suggestion
do -- `math.randomseed` patch
```

spacewander (approved):
Next time, please don't use force-push for a reviewing PR.

apisix/patch.lua
```diff
@@ -86,6 +93,64 @@ do
...
 
+do -- `math.randomseed` patch
+
+    -- Seeds the random generator, use with care.
+    -- Once - properly - seeded, this method is replaced with a stub
+    -- one. This is to enforce best-practices for seeding in ngx_lua,
+    -- and prevents third-party modules from overriding our correct seed
+    -- (many modules make a wrong usage of `math.randomseed()` by calling
+    -- it multiple times or by not using unique seeds for Nginx workers).
+    -- Inspired by kong.globalpatches
```
moonming, line 104: if these codes are copied from kong, you MUST add copyright in https://github.com/apache/apisix/blob/master/LICENSE.
if not, please remove this line.

moonming (changes requested):

apisix/patch.lua
```diff
@@ -86,6 +93,44 @@ do
...
+
+    local resty_random = require("resty.random")
+    local math_randomseed = math.randomseed
+    local seeded
+
+    -- make linter happy
+    -- luacheck: ignore
+    math.randomseed = function()
+        -- check seed mark
+        if seeded or false then
```
tzssangglass, line 111: can we just use `if not seeded then`

apisix/patch.lua
```diff
@@ -86,6 +93,44 @@ do
...
+        local t = {}
+
+        for i = 1, #random_bytes do
+            t[i] = string.byte(random_bytes, i)
+        end
+
+        local s = table.concat(t)
+
+        seeded = true
+        math_randomseed(tonumber(s))
```
tzssangglass, line 129: I would like to set `seeded = true` after `math_randomseed(tonumber(s))` 🤣

it just looks a bit strange here

leslie-tsang: Make sense. thanks for suggestion. 🥳

moonming (approved):

PR #5732, opened 2021-12-08
feat(consumer-restriction):  customize rejected_msg

qq54903099:

consumer-restriction plugin add reject_message parameter support user to custom define return message when request rejected.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -51,7 +51,8 @@ local schema = {
                 }
             }
         },
-        rejected_code = {type = "integer", minimum = 200, default = 403}
+        rejected_code = {type = "integer", minimum = 200, default = 403},
+        reject_message = {type = "string"}
     },
     anyOf = {
         {required = {"blacklist"}},
@@ -105,7 +106,11 @@ local function is_method_allowed(allowed_methods, method, user)
 end
 
 local function reject(conf)
-    return conf.rejected_code, { message = "The " .. conf.type .. " is forbidden." }
+    local reject_message = "The " .. conf.type .. " is forbidden." ;
+    if conf.reject_message then
+        reject_message = conf.reject_message;
+    end
+    return conf.rejected_code, { message = reject_message }
 end
 
 function _M.check_schema(conf)
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -52,7 +52,7 @@ local schema = {
             }
         },
         rejected_code = {type = "integer", minimum = 200, default = 403},
-        reject_message = {type = "string"}
+        rejected_message = {type = "string"}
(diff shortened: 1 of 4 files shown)


qq54903099:
how to chose reviewers?

apisix/plugins/consumer-restriction.lua
```diff
@@ -105,7 +106,11 @@ local function is_method_allowed(allowed_methods, method, user)
 end
 
 local function reject(conf)
-    return conf.rejected_code, { message = "The " .. conf.type .. " is forbidden." }
+    local reject_message = "The " .. conf.type .. " is forbidden." ;
```
spacewander, line 109: We can check `conf.reject_message` before creating new string?

spacewander (commented):
Let's check the feature in https://github.com/apache/apisix/blob/139c3972898f1e45c6d5b1f40143c504116b3486/t/plugin/consumer-restriction.t#L246

apisix/plugins/consumer-restriction.lua
qq54903099: ok

qq54903099: > 

ok

apisix/plugins/consumer-restriction.lua
```diff
@@ -51,7 +51,8 @@ local schema = {
                 }
             }
         },
-        rejected_code = {type = "integer", minimum = 200, default = 403}
+        rejected_code = {type = "integer", minimum = 200, default = 403},
+        rejected_message = {type = "string"}
```
leslie-tsang, line 55:
```suggestion
        rejected_msg = {type = "string"}
```
Would it be better to change `rejected_message` to `rejected_msg`? to keep it similar to other plugins.

t/plugin/consumer-restriction.t
```diff
@@ -263,7 +263,8 @@ Authorization: Basic amFjazIwMjA6MTIzNDU2
                             "consumer-restriction": {
                                  "blacklist": [
                                      "jack1"
-                                 ]
+                                 ],
+                                 "rejected_message": "request is forbidden"
```
leslie-tsang, line 267: New test cases are needed to ensure that the new feat works as expected.

spacewander: This change is already a new case. But we need to make sure it can pass.

spacewander: @qq54903099 
Please update the test to make CI pass: https://github.com/apache/apisix/runs/4465467353?check_suite_focus=true

apisix/plugins/consumer-restriction.lua
qq54903099: ok

spacewander (approved):

tokers (approved):

Issue #5714, opened 2021-12-06
Proposal: support OpenPolicyAgent for access control


bzp2010:
## Background
OPA is an open source lightweight general-purpose policy engine, which is a full-featured policy engine that can replace the built-in policy module in your software and decouple the service from the policy engine.
It describes policies through a policy DSL language "Rego" and stores policy data through JSON, after which the user can send a query request and OPA will combine the policy with the data and the query request entered by the user to generate policy decisions.
![image](https://user-images.githubusercontent.com/8078418/144813586-bbe724cd-5cde-4a18-9e0c-4ed205a0b66f.png)
Users can easily integrate OPA with its services, such as program libraries, HTTP API, etc. In this plugin, we will integrate OPA using HTTP API.

## Test Environment
```text
# Run the OPA throught Docker
docker run -d --name opa -p 8181:8181 openpolicyagent/opa:0.35.0 run -s

# Create policy 'example'
curl -XPUT 'localhost:8181/v1/policies/example' \
--header 'Content-Type: text/plain' \
--data-raw 'package example

default allow = false

allow {
  input.request.http.headers["test-header"] == "only-for-test"
}'

# Test: success
curl -XPOST 'localhost:8181/v1/data/example/allow' \
--header 'Content-Type: application/json' \
--data-raw '{
  "input": {
    "request": {
      "http": {
        "headers": {
          "test-header": "only-for-test"
        }
      }
    }
  }
}'

{
    "result": true
}

# Test: failed
curl -XPOST 'localhost:8181/v1/data/example/allow' \
--header 'Content-Type: application/json' \
--data-raw '{
  "input": {
    "request": {
      "http": {
        "headers": {
          "test-header": "not-for-test"
        }
      }
    }
  }
}'

{
    "result": false
}
```

### Configure schema
Name | Type | Requirement | Default | Description
-- | -- | -- | -- | --
host | string | required |   | OPA host (eg. https://localhost:8181)
ssl_verify | boolean | optional | true | Whether to verify the certificate
package | string | required |   | Policy package
decision | string | required |   | Policy rule name

### More functions (Need more discuss)
Name | Type | Requirement | Default | Description
-- | -- | -- | --  | --
with_route | boolean | optional | false | Carry current Route information in OPA API requests
with_consumer | boolean | optional | false | Carry current Consumer information in OPA API requests
with_upstream | boolean | optional | false | Carry current Upstream information in OPA API requests

### Implemention
[...]

PR #5734, opened 2021-12-08
feat: basic support OPA plugin

bzp2010:

Add OPA plugin to support API access control using OPA services.
implement #5714 

### Pre-submission checklist:


apisix/plugins/opa.lua
```diff
@@ -0,0 +1,104 @@
...
+local _M = {
+    version = 0.1,
+    priority = 2001,
+    name = "opa",
+    schema = schema,
+}
+
+
+function _M.check_schema(conf)
+    local ok, err = core.schema.check(schema, conf)
```
spacewander, line 56:
```suggestion
    return core.schema.check(schema, conf)
```

apisix/plugins/opa.lua
```diff
@@ -0,0 +1,104 @@
...
+    local res, err = httpc:request_uri(endpoint, params)
+
+    -- block by default when decision is unavailable
+    if not res or err then
+        core.log.error("failed to process OPA decision, err: ", err)
+        return 403
+    end
+
+    -- parse the results of the decision
+    local ret = core.json.decode(res.body).result
```
spacewander, line 96: Need to check if decode succeeded?

apisix/core/request.lua
```diff
@@ -269,6 +269,15 @@ function _M.get_port(ctx)
 end
 
 
+function _M.get_path(ctx, original)
```
spacewander, line 272: The `original` is unused?

apisix/plugins/opa/helper.lua
```diff
@@ -0,0 +1,62 @@
...
+local core     = require("apisix.core")
+local ngx_var  = ngx.var
+local ngx_time = ngx.time
+
+local _M = {}
+
+
+local function build_var(conf, ctx)
+    return {
+        server_addr = ngx_var.server_addr,
```
spacewander, line 27: Let's use ctx.var instead.

t/plugin/opa.t
```diff
@@ -0,0 +1,140 @@
...
+property "host" is required
+
+
+
+=== TEST 3: wrong type for `host`
+--- config
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.opa")
+            local ok, err = plugin.check_schema({host = 3233, package = "example", decision = "allow"})
```
spacewander, line 78: Can we use table drive test for check_schema?

anderseknert (commented):
Nice! I can't comment on the Lua bits, but what I read looks sensible. Some random thoughts:

* I would probably just let the user configure the full OPA URL rather than separating host, package and decision. In some cases you'll actually query a whole package, which is fine for OPA.
* The policy example (and from what I understand, the Lua code) only deals with boolean decisions. Maybe good to keep it simple, but could also be useful to allow OPA to return for example a reason for denying a request, and surface that to the caller.
* I don't see much in terms of documentation. Will that be added in a later PR?

But all things considered, great work on this! 👏

bzp2010:
Hi, @anderseknert. Glad to see your comments.

> I would probably just let the user configure the full OPA URL rather than separating host, package and decision. In some cases you'll actually query a whole package, which is fine for OPA.

Do you mean to merge host, package, decision together and use a single field (eg. `url`)?

> The policy example (and from what I understand, the Lua code) only deals with boolean decisions. Maybe good to keep it simple, but could also be useful to allow OPA to return for example a reason for denying a request, and surface that to the caller. I don't see much in terms of documentation. Will that be added in a later PR?

Features such as custom response codes, response headers and error causes will be added later in the PR. It will basically follow the roadmap section in #5714 to implement it.

anderseknert:
> Do you mean to merge host, package, decision together and use a single field (eg. `url`)?

Yeah. Either way is fine though :)
 
> Features such as custom response codes, response headers and error causes will be added later in the PR. It will basically follow the roadmap section in #5714 to implement it.

Oh, nice! I had missed the roadmap section. Thanks for bringing that to my attention! Sounds like a good plan 👍

bzp2010:
Ok, In that case, I would start by keeping the current way of combining URLs, which seems to provide better flexibility. XD

apisix/plugins/opa/helper.lua
bzp2010: This is a missing unmodified, previously I was using `reuqest_uri` var and separate `?` way to get pure paths, after checking the documentation I found that I can get pure paths using `uri` var. I will fix it.

t/plugin/opa.t
bzp2010: Yes, I think that's a good way, do we have any current examples of doing this for reference?

apisix/plugins/opa.lua
bzp2010: fixed

apisix/plugins/opa.lua
bzp2010: added

apisix/plugins/opa/helper.lua
bzp2010: fixed

apisix/core/request.lua
bzp2010: Yes, a useless code that I forgot to change.

bzp2010: fixed

anderseknert:
> Ok, In that case, I would keeping the current way of combining URLs, which seems to provide better flexibility. XD

I'm not sure that it's more flexible. It seems that currently all the fields are required to build the URL:

```lua
required = {"host", "package", "decision"}
```
But while not super common, it is perfectly legitimate to only query OPA for only a package, like:

```
http://localhost:8181/v1/data/policy
```

When OPA is queried on the package level, it will evaluate all the rules under that package and return the result. So a policy that looks like the below:

```rego
package policy

default allow = false

allow {
    input.user.roles[_] == "admin"
}

reason = "User needs to be admin" {
    not allow
}
```

When queried at `http://localhost:8181/v1/data/policy`, would respond with the result from evaluating both rules:

```json
{
    "allow": false,
    "reason": "User needs to be admin"
}
```

In this case, we don't query a rule (or decision as it's called in this code) directly, but rather all the rules in the package.

bzp2010:
Oh, I see, I will change it later. I don't know OPA very well, thank you for pointing out my problem.

t/plugin/opa.t
spacewander: You can take this as an example: https://github.com/apache/apisix/blob/c178435d7ada4eeb713d9a1688fb5f54f971abdf/t/plugin/gzip.t#L494

apisix/plugins/opa.lua
```diff
@@ -0,0 +1,104 @@
...
+        core.log.error("failed to process OPA decision, err: ", err)
+        return 403
+    end
+
+    -- parse the results of the decision
+    local data, err = core.json.decode(res.body)
+
+    if err then
+        core.log.error("invalid response body: ", res.body, " err: ", err)
+        return 403
```
tokers, line 95: Should we return `500` when the json decoding is fail.y

bzp2010: changed

apisix/plugins/opa.lua
```diff
@@ -0,0 +1,103 @@
...
+        core.log.error("failed to process OPA decision, err: ", err)
+        return 403
+    end
+
+    -- parse the results of the decision
+    local data, err = core.json.decode(res.body)
+
+    if err then
+        core.log.error("invalid response body: ", res.body, " err: ", err)
+        return 500
```
spacewander, line 94:
```suggestion
        return 503
```
for known error

t/plugin/opa.t
```diff
@@ -0,0 +1,139 @@
...
+property "policy" is required
+
+
+
+=== TEST 3: wrong type for `host`
+--- config
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.opa")
+            local ok, err = plugin.check_schema({host = 3233, policy = "example/allow"})
```
spacewander, line 78: Don't forget https://github.com/apache/apisix/pull/5734#discussion_r766265874

t/plugin/opa.t
bzp2010: changed

apisix/plugins/opa.lua
bzp2010: OK, changed

t/plugin/opa.t
bzp2010: I've switched to using TDT mode and waiting for CI to see how it turns out.

bzp2010: TDT changed

spacewander (approved):

bzp2010:
ping @tokers @tzssangglass @shuaijinchao PTAL, thanks

shuaijinchao (approved):

tzssangglass (approved):

moonming:
@bzp2010 please add OPA into the README, thanks.

PR #5745, opened 2021-12-08
feat(vault): vault lua module, integration with jwt-auth authentication plugin

bisakhmondal:

This PR integrates HashiCorp Vault with APISIX jwt-auth authentication plugin.
With this PR
- APISIX can store, delete, fetch data from vault (lua module)
- jwt-auth can retrieve private and public keys from the vault's encrypted kv storage.
- jwt-auth can store and retrieve signing secrets into and from the vault.

### Pre-submission checklist:


bisakhmondal:
PS: CI failure, not related to the introduced changes in this PR
https://github.com/apache/apisix/runs/4457908407?check_suite_focus=true#step:13:658

apisix/core/vault.lua
```diff
@@ -0,0 +1,116 @@
...
+local function set(key, data, rel_path)
+    core.log.info("stroing data into vault for key: ", key,
+                    "and value: ", core.json.delay_encode(data, true))
+
+    local res, err = make_request_to_vault("POST", key, rel_path, data)
+    if not res or err then
+        return nil, "failed to store data into vault kv engine " .. err
+    end
+
+    return {status = "success"}
```
spacewander, line 96: For write operation, better to return a single boolean, so the caller can check if the call success without knowing the internal detail.

apisix/plugins/jwt-auth.lua
```diff
@@ -198,7 +312,11 @@ end
 
 
 local function sign_jwt_with_HS(key, auth_conf, payload)
-    local auth_secret = get_secret(auth_conf)
+    local auth_secret, err = get_secret(auth_conf)
+    if not auth_secret then
+        core.log.error("failed to sign jwt, err: ", err)
+        core.response.exit(500, "failed to sign jwt")
```
spacewander, line 318:
```suggestion
        core.response.exit(503, "failed to sign jwt")
```

Use 503 for known failure.

apisix/plugins/jwt-auth.lua
```diff
@@ -76,7 +84,23 @@ local consumer_schema = {
...
                     required = {"public_key", "private_key"},
-                }
+                },
+                {
+                    properties = {
+                        vault = {
+                            type = "object",
+                            properties = {
+                                path = {type = "string"},
+                                add_prefix = {type = "boolean"}
```
spacewander, line 94: We should not allow users to specify path & add_prefix in the MVP, which will make thing very complex.
First, people may make mistakes, like using the same path according to multiple consumers.
Second, we need to enforce users to put their data under the same namespace, so we can provide a consistent token for APISIX to read the data.

apisix/admin/consumers.lua
```diff
@@ -102,6 +103,33 @@ function _M.get(consumer_name)
     end
 
     utils.fix_count(res.body, consumer_name)
+
+    if consumer_name then
+        -- if data is queried for a single consumer, and there is any plugin where the vault config
+        -- is enabled - it fetches vault data and returns combined with etcd response.
```
spacewander, line 109: IMHO, it is not a good idea to merge the vault data with etcd response. Now everyone who can access Admin API could access the vault data, which is against the reason to use vault.

apisix/plugins/jwt-auth.lua
```diff
@@ -119,29 +143,74 @@ function _M.check_schema(conf, schema_type)
...
     if not ok then
         return false, err
     end
 
-    if schema_type == core.schema.TYPE_CONSUMER then
-        if conf.algorithm ~= "RS256" and not conf.secret then
-            conf.secret = ngx_encode_base64(resty_random.bytes(32, true))
+    -- in nginx init_worker_by_lua context API calls are disabled,
+    -- also that is a costly operation during system startup.
+    if ngx.get_phase() == "init_worker" then
```
spacewander, line 155: We can check the data in Vault, but in the Admin API during creating. Sending a bundle of requests in full sync (which is not only happened during system startup) is too expensive. For instance, we don't connect to aws-lambda in the check schema. Even the data is valid, people can still change it later as the data does not come from the major data source.

apisix/core/vault.lua
bisakhmondal: Nice suggestion.

apisix/plugins/jwt-auth.lua
bisakhmondal: First:
Hi, thanks for the review. Without this feature, the MVP will be incomplete. Let me put down what it does:
suppose a user has the public key and private key stored into some arbitrary path `kv/rsa/token1`. It's very obvious users want to use it. What the plugin does is actually - if `path` is provided inside vault config it looks for fields like `public_key` and `private_key` (RS256) and `secret` (HS256 or HS512).
You can do it by
```shell
# the current directory contains public.pem and private.pem
vault kv put kv/rsa/token1 private_key=@private.pem public_key=@public.pem
```

PS: We are not storing consumer config into the vault so using the same path from multiple consumers won't be an issue as long as the user wants to do it. It just uses the secret from that path.

> people may make mistakes, like using the same path according to multiple consumers.

It's totally fine, for HS256 if the secret is not there in the auth_config and no data with key `secret` inside vault kv path - a new secret will be generated and be stored at that vault path. If someone reuses it, the stored secret inside the vault will be reused - there won't be any rewrite.

Second:
> Second, we need to enforce users to put their data under the same namespace, so we can provide a consistent token for APISIX to read the data.

Do you mean vault namespace? That's an enterprise feature [even I can't test it], we can look into it in the next version. WDYT?

Do let me know your views.

apisix/admin/consumers.lua
bisakhmondal: That's actually true. What do you suggest here, removing the part or supporting the feature in a different way?
One reason to have this feature:
- have a broader look of the data that are being used for the consumer. Instead of looking the responses of etcd and vault separately. But yes your concern is totally apt.

apisix/plugins/jwt-auth.lua
bisakhmondal: > We can check the data in Vault, but in the Admin API during creating. Sending a bundle of requests in full sync (which is not only happened during system startup) is too expensive

Yes, that's the current behaviour. For handling HTTP.PUT the ngx phase is `content_by_lua`. The full sync is done in apisix startup that is `init_worker_by_lua` phase. So in that case, the plugin simply skips the validation with external sources.
Btw, are you trying to convey something else?
> Even the data is valid, people can still change it later as the data does not come from the major data source.

Totally agree with you : )

apisix/admin/consumers.lua
spacewander: I think we can remove it.

BTW, when people store their data in protection storage, easier to access or can be written is no longer a plus for them. Restriction access is more helpful.

apisix/plugins/jwt-auth.lua
spacewander: There are two cases people what to set a common path (or ref):
1. they already have the data under another key
2. they want to share the same data among configurations

 For reason 1, since we already enforce the attributes name (secret, public key, .etc), it is not suitable to reuse the data format. And we may add new attribute requirements, which might conflict with the data already there.

For reason 2, it is quite strange to share the same configuration among consumers. Maybe we can add this feature when it's needed. (More features require more doc & test, and more bugs sometimes).

> Do you mean vault namespace? That's an enterprise feature [even I can't test it], we can look into it in the next version. WDYT?

Sorry for misleading you. Actually, I mean the keys can have a common prefix, like `/kv/apisix/`, which is easier to recognize and manage.

apisix/plugins/jwt-auth.lua
spacewander: > The full sync is done in apisix startup that is init_worker_by_lua phase.

Not exactly. The full sync can happen elsewhere like the data in etcd is compacted.
BTW, even the sync timer is triggered in the init_worker_by_lua, it runs in a separate "timer" phase. Currently, there is no way to distinguish a DP side check_schema out of the CP side.

Even we have successfully checked the data, its benefit is limited. Maybe we can just leave a note in the doc?

apisix/admin/consumers.lua
bisakhmondal: Ack

apisix/plugins/jwt-auth.lua
bisakhmondal: > There are two cases people what to set a common path (or ref):
> 
> 1. they already have the data under another key
> 2. they want to share the same data among configurations
> 
> For reason 1, since we already enforce the attributes name (secret, public key, .etc), it is not suitable to reuse the data format.

One thing you are missing here, though we are enforcing key names (public_key) on the very first version [we can add user-provided key names in later versions]...we are letting them store the keys into the vault that can be referenced from the plugin at ease. Else the feature would be too much documentation oriented (they have to know from where the plugin will perform lookup and store it there). See the test case 14,15,16.

> And we may add new attribute requirements, which might conflict with the data already there.

Okay, we are only storing value in a vault kv path if secret is not provided in HS256/HS512. Maybe for the first version, we just read (if the secret is there, it uses it. if not it generate it and stores into plugin conf. but won't write anything), then users don't have to give us write access to the kv engine. Once we build up enough trust gradually, we can introduce edit capabilities into the plugin.
What are your views here?

> For reason 2, it is quite strange to share the same configuration among consumers. Maybe we can add this feature when it's needed. (More features require more doc & test, and more bugs sometimes).



> > Do you mean vault namespace? That's an enterprise feature [even I can't test it], we can look into it in the next version. WDYT?
> 
> Sorry for misleading you. Actually, I mean the keys can have a common prefix, like `/kv/apisix/`, which is easier to recognize and manage.

By allowing custom path or namespace we are not holding them back from storing key=value into kv/apisix/*  : )
To be honest, if I were the end user, I definitely want that flexibility instead of going through the docs and figuring out for jwt key name `foo` I have to store key into kv/apisix/jwt-auth/keys/foo.

bisakhmondal: Btw, on thing I forgot to mention, both `path` and `add_prefix` are optional. Line 162

spacewander (changes requested):
@bisakhmondal 
Please don't ignore the suggestion from the maintainer, thanks!

apisix/plugins/jwt-auth.lua
spacewander: Please remove the path & add_prefix, which is too flexible, and made future evolution harder. Let's be strict from the beginning. It's easier to relax the rule than be more strict with the rule. If people really need this feature, we can add it at that time.

bisakhmondal:
> @bisakhmondal Please don't ignore the suggestion from the maintainer, thanks!

Hi @spacewander, please don't take me wrong. I never intended to ignore your suggestions. I thought we were having a discussion😅. I really am sorry if it looked that way.

Btw, updated accordingly : )

apisix/plugins/jwt-auth.lua
bisakhmondal: Done

apisix/plugins/jwt-auth.lua
bisakhmondal: Resolved. Thanks

apisix/plugins/jwt-auth.lua
```diff
@@ -28,7 +29,7 @@ local ngx_time = ngx.time
 local sub_str  = string.sub
 local plugin_name = "jwt-auth"
 local pcall = pcall
-
+local jwt_vault_prefix = "jwt-auth/keys/"
```
spacewander, line 32: What about changing it to `consumer/<username>/jwt-auth/`?
I think about it several times. Although it requires to change several places in this PR, the new format is more extendable, and easier to recognize. People will remember the username better than the key.

bisakhmondal: Thanks for the suggestion. Addressed

docs/en/latest/plugins/jwt-auth.md
```diff
@@ -110,6 +117,68 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
...
+    "plugins": {
+        "jwt-auth": {
+            "key": "key-1",
+            "vault": {}
+        }
+    }
+}'
+```
+
+Here the plugin looks up for key `secret` inside vault path (`<vault.prefix from default-conf.yaml>/jwt-auth/keys/key-1`) for key `key-1` mentioned in the consumer config and uses it for subsequent signing and jwt verification. If the key is not found in the same path, the plugin logs error and fails to perform jwt authentication.
```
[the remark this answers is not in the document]
bisakhmondal: Done

docs/en/latest/plugins/jwt-auth.md
```diff
@@ -51,6 +55,9 @@ For more information on JWT, refer to [JWT](https://jwt.io/) for more informatio
...
 | exp           | integer | optional    | 86400   | [1,...]                     | token's expire time, in seconds                                                                                                                  |
 | base64_secret | boolean | optional    | false   |                             | whether secret is base64 encoded                                                                                                                 |
+| vault | dictionary | optional    |    |                             | whether vault to be used for secret (secret for HS256/HS512  or public_key and private_key for RS256) storage and retrieval. The plugin by default uses the vault path as `kv/apisix/jwt-auth/keys/<jwt-auth.key>` for secret retrieval. |
+
+**Note**: To enable vault integration, first visit the [config-default.yaml](https://github.com/apache/apisix/blob/master/conf/config-default.yaml) and update the yaml vault attributes with your vault server configuration.
```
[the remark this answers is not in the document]
bisakhmondal: ACK

docs/en/latest/plugins/jwt-auth.md
```diff
@@ -51,6 +55,9 @@ For more information on JWT, refer to [JWT](https://jwt.io/) for more informatio
 | algorithm     | string  | optional    | "HS256" | ["HS256", "HS512", "RS256"] | encryption algorithm.                                                                                                                            |
 | exp           | integer | optional    | 86400   | [1,...]                     | token's expire time, in seconds                                                                                                                  |
 | base64_secret | boolean | optional    | false   |                             | whether secret is base64 encoded                                                                                                                 |
+| vault | dictionary | optional    |    |                             | whether vault to be used for secret (secret for HS256/HS512  or public_key and private_key for RS256) storage and retrieval. The plugin by default uses the vault path as `kv/apisix/jwt-auth/keys/<jwt-auth.key>` for secret retrieval. |
```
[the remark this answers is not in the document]
bisakhmondal: Done

t/plugin/jwt-auth-vault.t
```diff
@@ -0,0 +1,369 @@
...
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.jwt-auth")
+            local core = require("apisix.core")
+            local conf = {
+                key = "key-1",
+                algorithm = "RS256"
+            }
+
+            local ok, err = plugin.check_schema(conf, core.schema.TYPE_CONSUMER)
```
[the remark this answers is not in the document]
bisakhmondal: Addressed

bisakhmondal:
CI is flakey, failure is unrelated to the PR.

conf/config-default.yaml
```diff
@@ -281,6 +281,17 @@ etcd:
                                   # the default value is true, e.g. the certificate will be verified strictly.
     #sni:                         # the SNI for etcd TLS requests. If missed, the host part of the URL will be used.
 
+# storage backend for sensitive data storage and retrieval
+vault:
```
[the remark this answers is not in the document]
spacewander: Err. Actually, I mean like this :
```
#vault:
  # ...
```

t/plugin/jwt-auth-vault.t
```diff
@@ -0,0 +1,368 @@
...
+                    "username": "jack",
+                    "plugins": {
+                        "jwt-auth": {
+                            "key": "key-hs256",
+                            "algorithm": "HS256",
+                            "vault":{}
+                        }
+                    }
+                }]],
+                [[{
```
spacewander, line 105: Could we remove the expected response? We don't need to check this output anymore. More other tests already do it.

spacewander:
My review is almost done, only a few things need to be addressed. 😄

conf/config-default.yaml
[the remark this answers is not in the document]
bisakhmondal: Ahh, I get it now :smile:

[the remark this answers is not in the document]
bisakhmondal: Updated

t/plugin/jwt-auth-vault.t
bisakhmondal: Sure. Done

t/plugin/jwt-auth-vault.t
```diff
@@ -163,6 +148,13 @@ passed
 
 
 === TEST 4: sign a jwt and access/verify /secure-endpoint, fails as no secret entry into vault
+--- yaml_config
```
spacewander, line 151: We can share the config via extra_yaml_config
See https://github.com/apache/apisix/blob/71c256be81d95d56ea38f9874699801cbbad6a2e/t/plugin/ext-plugin/http-req-call.t#L47

bisakhmondal: Resolved

t/plugin/jwt-auth-vault.t
```diff
@@ -0,0 +1,381 @@
...
+=== TEST 5: store HS256 secret into vault
+--- exec
+VAULT_TOKEN='root' VAULT_ADDR='http://0.0.0.0:8200' vault kv put kv/apisix/consumer/jack/jwt-auth secret=$3nsitiv3-c8d3
+--- response_body
+Success! Data written to: kv/apisix/consumer/jack/jwt-auth
+
+
+
+=== TEST 6: sign a HS256 jwt and access/verify /secure-endpoint
+--- extra_yaml_config
```
spacewander, line 201: We can set up a common yaml config in this file?

t/plugin/jwt-auth-vault.t
```diff
@@ -39,6 +39,16 @@ _EOC_
...
+    my $vault_config = $block->extra_yaml_config // <<_EOC_;
+vault:
+  host: "http://0.0.0.0:8200"
+  timeout: 10
+  prefix: kv/apisix
+  token: root
+_EOC_
+
+    $block->set_value("extra_yaml_config", $vault_config);
+
```
bisakhmondal, line 51: How is it now? @spacewander

spacewander: LGTM

spacewander (approved):

t/plugin/jwt-auth-vault.t
```diff
@@ -0,0 +1,363 @@
...
+        }
+    }
+--- response_body
+failed to sign jwt
+--- error_code: 503
+--- error_log: true
+--- grep_error_log eval
+qr/failed to sign jwt, err: secret could not found in vault/
+--- grep_error_log_out
+failed to sign jwt, err: secret could not found in vault
```
tzssangglass, line 191: we can just use 
```
--- error_log eval
qr/failed to sign jwt, err: secret could not found in vault/
```
?

bisakhmondal: Done

tzssangglass (approved):

PR #5779, opened 2021-12-13
feat: supprot OPA plugin complex response

bzp2010:
Enhance OPA plug-in functionality to support overwriting of status codes, response headers, and response bodies when validation is rejected.

stage 2 of #5714 

### Pre-submission checklist:


bzp2010:
## Summary
The OPA plugin is enhanced in this PR to support overriding the status code, request header and request body in case of a failed check.
- Update OPA policy
- Update plugin

## Description
A response containing the following can be sent to APISIX in the OPA policy.
```json
{
  "result": {
    "allow": false,
    "status_code": 302,
    "headers": {
      "Location": "http://example.com/auth"
    },
    "reason": {
      "test": "test"
    }
  }
}
```
The status code, response header, and reason are all optional respectively, and the reason can be a string or object, and the plugin will automatically JSON encode according to the data type.

apisix/plugins/opa.lua
```diff
@@ -94,8 +95,27 @@ function _M.access(conf, ctx)
         return 503
     end
 
-    if not data.result then
-        return 403
+    local result = data.result
```
shuaijinchao, line 98: Will an exception occur if `data` is nil?

apisix/plugins/opa.lua
```diff
@@ -94,8 +95,27 @@ function _M.access(conf, ctx)
...
+
+    if not result.allow then
+        if result.headers then
+            core.response.set_header(result.headers)
+        end
+
+        local status_code = 403
+        local reason = ""
```
shuaijinchao, line 106: Is it okay to use `local reason` to define the default value as nil? In addition, would it be better to put the definition of `reason` above `if result.reason then`?

apisix/plugins/opa.lua
```diff
@@ -94,8 +95,27 @@ function _M.access(conf, ctx)
         return 503
     end
 
-    if not data.result then
-        return 403
+    local result = data.result
+
+    if not result.allow then
+        if result.headers then
+            core.response.set_header(result.headers)
```
spacewander, line 102: Is there a test for multiple headers in result.headers?

apisix/plugins/opa.lua
```diff
@@ -94,8 +95,27 @@ function _M.access(conf, ctx)
...
+
+        local status_code = 403
+        local reason = ""
+
+        if result.status_code then
+            status_code = result.status_code
+        end
+
+        if result.reason then
+            reason = type(result.reason) == "table"
```
spacewander, line 113: Let's add test for result.reason (string and table cases).

bzp2010: added

apisix/plugins/opa.lua
bzp2010: Got it, I agree with it. changed

apisix/plugins/opa.lua
bzp2010: Yes, it does have problems, I modified the error checking logic so that when this happens, it responds with 503 and take a log.

spacewander (approved):

tzssangglass (approved):

bzp2010:
ping @shuaijinchao PTAL thanks

shuaijinchao (approved):

apisix/plugins/opa.lua
```diff
@@ -89,13 +90,31 @@ function _M.access(conf, ctx)
     -- parse the results of the decision
     local data, err = core.json.decode(res.body)
 
-    if err then
+    if err or not data or not data.result then
         core.log.error("invalid response body: ", res.body, " err: ", err)
```
spacewander, line 94: We should distinguish "not data or not data.result" in the error log.

bzp2010: I split it into an "invalid response body", and "invalid OPA decision format: xxx err: `result` field does not exist".

bzp2010: splited

spacewander (approved):

shuaijinchao (approved):

anderseknert:
👏

PR #5782, opened 2021-12-13
fix(ext-plugin): don't use stale key

S96EA:
ext_plugin an expired key in lrucachd that could not find in go runner。
patch for [#5309](https://github.com/apache/apisix/pull/5309)

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -677,6 +677,7 @@ local function create_lrucache()
 
     lrucache = core.lrucache.new({
         type = "plugin",
+    	invalid_stale = true,
         ttl = helper.get_conf_token_cache_time(),
     })
 end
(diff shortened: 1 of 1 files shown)


apisix/plugins/ext-plugin/init.lua
```diff
@@ -677,6 +677,7 @@ local function create_lrucache()
 
     lrucache = core.lrucache.new({
         type = "plugin",
+    	invalid_stale = true,
```
spacewander, line 680: Could you do some refactor so we don't need to specify the lrucache options twice? Thanks!

spacewander (approved):
Maybe we can rename `create_lrucache` to `recreate_lrucache`?

S96EA:
> Maybe we can rename `create_lrucache` to `recreate_lrucache`?

okey, done

tzssangglass (approved):

PR #5787, opened 2021-12-13
docs: added only: false to example

danielkocot:
added "only:false" if http and stream proxy functionality is needed

For clearer documentation purposes and in reference to the general documentation of stream proxy.


spacewander (approved):

tzssangglass (approved):

juzhiyuan:
Thanks @danielkocot !

PR #5821, opened 2021-12-15
feat(plugins): make plugins more powerfull

wangxiaochuang:
we can use ngx.req.set_uri(..., true) and some other's only rewrite phase functions

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -592,7 +592,7 @@ http {
             set $dubbo_method                '';
             {% end %}
 
-            access_by_lua_block {
+            rewrite_by_lua_block {
                 apisix.http_access_phase()
             }
 


apisix/cli/ngx_tpl.lua
```diff
@@ -592,7 +592,7 @@ http {
             set $dubbo_method                '';
             {% end %}
 
-            access_by_lua_block {
+            rewrite_by_lua_block {
```
tokers, line 595: I am afraid we'd better have a discussion about this, this change breaks the original design purposes.

wangxiaochuang: how about this change

tzssangglass: APISIX does not currently mount the `rewrite_by_lua_block` phase of openresty.

PR #5823, opened 2021-12-15
docs: remove the redundant space between Chinese characters

Nidaozi:
There is an extra space after ‘该’

image

### DIFF
--- a/docs/zh/latest/architecture-design/plugin.md
+++ b/docs/zh/latest/architecture-design/plugin.md
@@ -26,8 +26,8 @@ title: Plugin
 `Plugin` 配置可直接绑定在 `Route` 上，也可以被绑定在 `Service` 或 `Consumer`上。而对于同一
 个插件的配置，只能有一份是有效的，配置选择优先级总是 `Consumer` > `Route` > `Service`。
 
-在 `conf/config.yaml` 中，可以声明本地 APISIX 节点都支持哪些插件。这是个白名单机制，不在该
-白名单的插件配置，都将会被自动忽略。这个特性可用于临时关闭或打开特定插件，应对突发情况非常有效。
+在 `conf/config.yaml` 中，可以声明本地 APISIX 节点都支持哪些插件。这是个白名单机制，不在该白名单的插件配置，
+都将会被自动忽略。这个特性可用于临时关闭或打开特定插件，应对突发情况非常有效。
 如果你想在现有插件的基础上新增插件，注意需要拷贝 `conf/config-default.yaml` 的插件节点内容到 `conf/config.yaml` 的插件节点中。
 
 插件的配置可以被直接绑定在指定 Route 中，也可以被绑定在 Service 中，不过 Route 中的插件配置


moonming:
@Nidaozi 
Please use English in the public channel, thx

tokers:
@Nidaozi Could you change the title like:

```
docs: remove the redundant space between Chinese characters
```

docs/zh/latest/architecture-design/plugin.md
```diff
@@ -26,8 +26,8 @@ title: Plugin
 `Plugin` 配置可直接绑定在 `Route` 上，也可以被绑定在 `Service` 或 `Consumer`上。而对于同一
 个插件的配置，只能有一份是有效的，配置选择优先级总是 `Consumer` > `Route` > `Service`。
 
-在 `conf/config.yaml` 中，可以声明本地 APISIX 节点都支持哪些插件。这是个白名单机制，不在该
-白名单的插件配置，都将会被自动忽略。这个特性可用于临时关闭或打开特定插件，应对突发情况非常有效。
+在 `conf/config.yaml` 中，可以声明本地 APISIX 节点都支持哪些插件。这是个白名单机制，不在该白名单的插件配置，
```
spacewander, line 29: The generated result doesn't have this problem.
https://apisix.apache.org/zh/docs/apisix/architecture-design/plugin/
微信截图_20211216091702

spacewander (changes requested):

docs/zh/latest/architecture-design/plugin.md
spacewander: Look like it is the problem of apisix-website in some browsers?
@juzhiyuan 
Could you take a look?

spacewander: 微信截图_20211216092955

The display on GitHub is OK.

juzhiyuan: image

Hi @spacewander, it's more like a Browser issue 🤔 I just checked the website & GitHub on Chrome, it has this extra space, too. Maybe that's because of the breaking line in markdowns 😄 

image

Thanks for @Nidaozi's PR and @spacewander's check, we could fix this **style** issue by this PR in my opinion, but need @Nidaozi's help to update your PR's title and description :)

juzhiyuan: I have confirmed this issue with others who're using Chrome in windows.

Nidaozi: thx; I have changed

spacewander:
```
`Plugin` 配置可直接绑定在 `Route` 上，也可以被绑定在 `Service` 或 `Consumer`上。而对于同一
个插件的配置，只能有一份是有效的，配置选择优先级总是 `Consumer` > `Route` > `Service`。
```

also have the same problem. Is it possible to add a way to detect this pattern?

Nidaozi: This problem is introduced when the browser is parsed. There are two ways to solve this problem. The first method uses regular detection to replace ‘\n’ with empty when it detects that the end is Chinese and the next line is still Chinese. The second way is to modify the CSS rendered by Markdown, set the'font-size' of the parent tag to 0 and the'font-size' of the child tag to the desired font size. https://stackoverflow.com/questions/8550112/prevent-browser-converting-n-between-lines-into-space-for-chinese-characters/8551033#8551033

juzhiyuan: Get it, I would prefer to seek help from Docusaurus 😄 Hi @slorber @Josh-Cena, would you please take look at this issue?

Josh-Cena: Hey, it's indeed a Markdown/HTML feature that they coerce line breaks to spaces. Generally we don't recommend prose wrapping for exactly this reason (and also for the sake of cleaner git diffs). You can try it in the MDX playground: https://mdx-git-renovate-babel-monorepo-mdx.vercel.app/playground

image

Maybe Firefox doesn't do that? I've tested with Chrome and Safari and they both create the spaces

juzhiyuan: image

image

Thanks to @Josh-Cena's clarification, after testing on the playground: If we use line-breaking or one space, it will be rendered in one line with a space; If we use 2 spaces and line breaking, it will be rendered in 2 lines.

If we need to fix this style rendered issue, we have to remove all line-breakings from markdown files. @Nidaozi, could you help to check if we have many cases like this? I would hope we can fix this case in one PR, but don't want it to be very lard.

Josh-Cena: > If we use 2 spaces and line breaking, it will be rendered in 2 lines.

It's actually part of the Markdown spec that two trailing spaces => `<br />` tag and two line breaks => begin a new paragraph.

juzhiyuan: 🙈🙈 How about closing this PR? 

cc @Nidaozi

Nidaozi: ok

Josh-Cena:
Hold on, why close this? Are we going to come up with a better resolution or is this just wontfix? I honestly think this is a legitimate fix and we can just fix this everywhere if needed

spacewander:
What about providing a check to the Chinese doc? We can add a script to check if
1. the last char of the line is not ascii
2. the next line is not empty
3. the first char of the next line is not ascii

If so, we can fold the Chinese phrase into one line. Such script can be integrated into the CI: https://github.com/apache/apisix/blob/fce2b44f5fe5c3358ec7dcdeb8fdb5530ee3e8a2/.github/workflows/doc-lint.yml#L22

juzhiyuan:
> I honestly think this is a legitimate fix and we can just fix this everywhere if needed

Hi @Josh-Cena, I can't clearly remember why we have breaking lines in our docs 😅 And +1 from me to improve the original docs by folding Chinese phrases into one line. @Nidaozi Would you please have a try?

image

Hi @spacewander, how about improving those markdown files first, then deciding if we need to implement the CI? Just to make sure, you mean we could use CI to check all coming Chinese documentation changes, right? 😄

Josh-Cena:
So if you use [Prettier](https://prettier.io/) in the website, you can just configure it to `proseWrap: 'none'` and it will collapse all wrapping line breaks.

juzhiyuan:
> So if you use [Prettier](https://prettier.io/) in the website, you can just configure it to `proseWrap: 'none'` and it will collapse all wrapping line breaks.

Just to make sure, do you mean to use Prettier to format all markdown files in this repo?

Josh-Cena:
Yeah, that could be helpful 🤪

juzhiyuan:
Get it, I will create an issue to track this 🥳

PR #5839, opened 2021-12-17
fix(google-cloud-logging): move ssl_verify to plugin configuration top level

### LINKED ISSUE #5838: improve: google-cloud-logging unable to set up SSL Verify
### Issue description

`google-cloud-logging` SSL verification settings currently need to be configured in the auth_config structure or auth_file configuration file to take effect. Currently, when using the file key configuration configuration, the key file downloaded from Google Cloud is usually configured directly into APISIX. The default file does not contain SSL-related settings. At this time, SSL Verify cannot be dynamically adjusted through the plugin configuration.

Solution: Adjust the configuration of SSL Verify to the primary structure of the plugin configuration.

For example:
```json
{
    "google-cloud-logging":{
        "auth_file":"/path/to/private_key.json",
        "ssl_verify":true,
        "...":"..."
    }
}
```

shuaijinchao:

FIX #5838

### Pre-submission checklist:


apisix/plugins/google-cloud-logging.lua
```diff
@@ -162,6 +163,7 @@ local function get_auth_config(config)
     end
 
     auth_config_cache = config_data
+    auth_config_cache.ssl_verify = conf.ssl_verify
```
spacewander, line 166: It seems the ssl_verify will be cached with the auth configuration. So changing the ssl_verify won't take effect immediately.

shuaijinchao: this is a good question and has been updated.

apisix/plugins/google-cloud-logging.lua
```diff
@@ -61,13 +61,13 @@ local schema = {
                     type = "string",
                     default = "https://logging.googleapis.com/v2/entries:write"
                 },
-                ssl_verify = {
```
tokers, line 64: I don't know whether this plugin is used by someone, if we just remove this setting, it's bad for these users. Or we can just reserve and deprecate it.

shuaijinchao: I think this has no substantive significance. In most cases, this field remains the default setting, and general users will not actively set this option. this plugin is a new feature released in version 2.11. In such a short period of time, I don't think it will be used in a mass production environment. In addition, it is a bug for people who use file configuration ^_^

apisix/plugins/google-cloud-logging/oauth.lua
```diff
@@ -111,10 +111,10 @@ function _M:new(config)
         access_token_expire_time = 0,
     }
 
-    if config.ssl_verify == false then
-        oauth.ssl_verify = config.ssl_verify
-    else
+    if ssl_verify == true then
         oauth.ssl_verify = true
```
spacewander, line 115: Why not `oauth.ssl_verify = ssl_verify` directly? The ssl_verify can't be nil.

spacewander:
Look like there is another issue: `auth_config_cache` is not updated even the auth_config has changed.

spacewander:
We can use `core.lrucache.plugin_ctx` to ensure each rule has its own auth object: https://github.com/apache/apisix/blob/4423509680a0c2103ce23b30a0da71735ef73e8f/apisix/plugins/limit-req.lua#L77
Also, it solves the ssl_verify cache problem.

spacewander:
Or we can bind the auth_config with the `process` pass to add_entry_to_new_processor.

apisix/plugins/google-cloud-logging.lua
tokers: Got it.

apisix/plugins/google-cloud-logging.lua
```diff
@@ -136,23 +136,23 @@ local function send_to_google(oauth, entries)
...
-        auth_config_cache = config.auth_config
+    if conf.auth_config then
+        auth_config_cache = conf.auth_config
         return auth_config_cache
     end
 
-    if not config.auth_file then
+    if not conf.auth_file then
         return nil, "configuration is not defined"
     end
```
tzssangglass: why not use if ... else

shuaijinchao: `conf.auth_config` and `conf.auth_file` is not  the same attribute

shuaijinchao:
> We can use `core.lrucache.plugin_ctx` to ensure each rule has its own auth object:
> 
> https://github.com/apache/apisix/blob/4423509680a0c2103ce23b30a0da71735ef73e8f/apisix/plugins/limit-req.lua#L77
> 
> 
> Also, it solves the ssl_verify cache problem.

good idea, updated

apisix/plugins/google-cloud-logging.lua
tzssangglass: sorry……

spacewander (approved):

tzssangglass (approved):

PR #5840, opened 2021-12-17
fix(core.log): log level need to be updated in some scenario

leslie-tsang:

log level need to be update in some scenario, such as called in init phase.

POC:
> set error_log log level as `error`
```yaml
  extra_lua_path: ""                # extend lua_package_path to load third party code
  extra_lua_cpath: ""               # extend lua_package_cpath to load third party code
  lua_module_hook: "my_project.my_hook"  # the hook module which will be used to inject third party code into APISIX
```
```lua
# code in my_project.my_hook
local core.log = require("apisix.core.log")
core_log.info("test")

# `info` func will be cached in `apisix.core.log`
# which break the expectation of the error_log log level
```

### Pre-submission checklist:

### DIFF
--- a/apisix/core/log.lua
+++ b/apisix/core/log.lua
@@ -24,6 +24,7 @@ local tostring = tostring
 local unpack = unpack
 -- avoid loading other module since core.log is the most foundational one
 local tab_clear = require("table.clear")
+local ngx_errlog = require("ngx.errlog")
 
 
 local _M = {version = 0.4}
@@ -42,17 +43,27 @@ local log_levels = {
 }
 
 
-local cur_level = ngx.config.subsystem == "http" and
-                  require "ngx.errlog" .get_sys_filter_level()
+local cur_level
+
 local do_nothing = function() end
+
+
+local function log_level_update()
+    -- Nginx use `notice` level in init phase instead of error_log directive config
+    -- Ref to src/core/ngx_log.c's ngx_log_init
+    if ngx.get_phase() ~= "init" then
+        cur_level = ngx.config.subsystem == "http" and ngx_errlog.get_sys_filter_level()
+    end
+end
 
 
 function _M.new(prefix)
     local m = {version = _M.version}
     setmetatable(m, {__index = function(self, cmd)
         local log_level = log_levels[cmd]
-
         local method
+        log_level_update()
+
         if cur_level and (log_level > cur_level)
         then
             method = do_nothing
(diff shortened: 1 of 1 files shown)


apisix/core/log.lua
```diff
@@ -42,17 +43,27 @@ local log_levels = {
...
 
 
-local cur_level = ngx.config.subsystem == "http" and
-                  require "ngx.errlog" .get_sys_filter_level()
+local cur_level
+
 local do_nothing = function() end
 
 
+local function log_level_update()
```
spacewander, line 51:
```suggestion
local function update_log_level()
```
would be better?

apisix/core/log.lua
```diff
@@ -74,8 +85,9 @@ end
 
 setmetatable(_M, {__index = function(self, cmd)
     local log_level = log_levels[cmd]
-
     local method
+    log_level_update()
+
```
spacewander, line 90: There is an issue: if we call the log.xxx in the init phase, we will always generate the method and affect the later phases.

leslie-tsang: Make sense, may be we need a new implement of this.

apisix/core/log.lua
leslie-tsang: Sure

apisix/core/log.lua
leslie-tsang: PLTA, now we don't cache the methods generated in the init phase.

apisix/core/log.lua
```diff
@@ -42,17 +43,27 @@ local log_levels = {
...
+local cur_level
+
 local do_nothing = function() end
 
 
+local function update_log_level()
+    -- Nginx use `notice` level in init phase instead of error_log directive config
+    -- Ref to src/core/ngx_log.c's ngx_log_init
+    if ngx.get_phase() ~= "init" then
+        cur_level = ngx.config.subsystem == "http" and ngx_errlog.get_sys_filter_level()
```
leslie-tsang, line 55: Shall we remove `ngx.config.subsystem == "http"` ?

apisix/core/log.lua
```diff
@@ -99,7 +102,10 @@ setmetatable(_M, {__index = function(self, cmd)
 
     -- cache the lazily generated method in our
     -- module table
-    _M[cmd] = method
+    if ngx.get_phase() ~= "init" then
```
tzssangglass, line 105: we can add `local get_phase = ngx.get_phase` at header of file, and use `get_phase()`

apisix/core/log.lua
spacewander: Yes, we can do it in the next PR.

spacewander (approved):

tzssangglass (approved):

PR #5859, opened 2021-12-20
chore: improve installation experience

leslie-tsang:

This PR make it  easy to install deps on a mini install OS.
 
### Pre-submission checklist:

### DIFF
--- a/utils/install-dependencies.sh
+++ b/utils/install-dependencies.sh
@@ -49,7 +49,7 @@ function install_dependencies_with_yum() {
     local common_dep="curl git gcc openresty-openssl111-devel unzip pcre pcre-devel openldap-devel"
     if [ "${1}" == "centos" ]; then
         # add APISIX source
-        sudo yum-config-manager --add-repo https://repos.apiseven.com/packages/centos/apache-apisix.repo
+        sudo yum install -y https://repos.apiseven.com/packages/centos/apache-apisix-repo-1.0-1.noarch.rpm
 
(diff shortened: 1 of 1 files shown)


utils/install-dependencies.sh
```diff
@@ -26,7 +26,7 @@ function detect_aur_helper() {
         AUR_HELPER=pacaur
     else
         echo No available AUR helpers found. Please specify your AUR helper by AUR_HELPER.
-        exit -1
+        exit 1
```
spacewander, line 29: Why need to change the arch Linux installation?

leslie-tsang: > Every command returns an exit status (sometimes referred to as a return status or exit code). A successful command returns a 0, while an unsuccessful one returns a **non-zero value** that usually can be interpreted as an error code. Well-behaved UNIX commands, programs, and utilities return a 0 exit code upon successful completion, though there are some exceptions.

Ref to [shell exit-status](https://tldp.org/LDP/abs/html/exit-status.html) and [SC2242](https://github.com/koalaman/shellcheck/wiki/SC2242)

IMO, We should obey `SC2242` to make it more user friendly ?

spacewander: Could we use `exit 255` to be compatible with the previous code?

spacewander (approved):

tzssangglass (approved):

tokers (approved):

PR #5867, opened 2021-12-21
change: enable HTTP when stream proxy is set and enable_admin is true

spacewander:
For development, this change let us don't need to specify `only:
false` additionally.

For DP which is used as L4 proxy, this change won't bring any
difference.

Also it makes the documentation of stream plugin simpler.

### Pre-submission checklist:

### DIFF
--- a/docs/zh/latest/stream-proxy.md
+++ b/docs/zh/latest/stream-proxy.md
@@ -40,10 +40,14 @@ apisix:
       - "127.0.0.1:9211"
 ```
 
-如果你需要同时启用 HTTP 和 stream 代理，设置 `only` 为 false：
+如果 `apisix.enable_admin` 为 true，上面的配置会同时启用 HTTP 和 stream 代理。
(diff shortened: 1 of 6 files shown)


docs/zh/latest/stream-proxy.md
```diff
@@ -40,10 +40,14 @@ apisix:
       - "127.0.0.1:9211"
 ```
 
-如果你需要同时启用 HTTP 和 stream 代理，设置 `only` 为 false：
+如果 `apisix.enable_admin` 为 true，上面的配置会同时启用 HTTP 和 stream 代理。
+If `apisix.enable_admin` is true, both HTTP and stream proxy are enabled with the configuration above.
```
leslie-tsang, line 44:
```suggestion
```

leslie-tsang (changes requested):

shuaijinchao (approved):

tzssangglass (approved):

PR #5881, opened 2021-12-22
feat(limit-count): allow sharing counter

### LINKED ISSUE #5342: request help: share limit counter between routes
### Issue description

To do this, you need to specify the key of the route's corresponding limit-count in lrucache so that the same limit object is shared across multiple routes. lrucache keys (hereafter called groups to distinguish them from limit keys) are currently generated automatically, ensuring that each route's group is independent. For this change, we need to be able to specify the group in the limit-count.

```
"limit-count": {
            "group": "group_id_blah"
            "count": 2,
            "time_window": 60,
            "rejected_code": 503,
            "key": "remote_addr"
}
```

Note that the configuration of the same group needs to be the same, which currently needs to be guaranteed by the caller, otherwise the limit object obtained by the group will be different from the configuration.

### Environment

- apisix version (cmd: `apisix version`): 2.10.1
- OS (cmd: `uname -a`):

spacewander:

Fix #5342 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/limit-count.lua
+++ b/apisix/plugins/limit-count.lua
@@ -16,6 +16,11 @@
 --
 local limit_local_new = require("resty.limit.count").new
 local core = require("apisix.core")
+local tab_insert = table.insert
+local ipairs = ipairs
+local pairs = pairs
+
+
 local plugin_name = "limit-count"
 local limit_redis_cluster_new
 local limit_redis_new
@@ -29,13 +34,60 @@ end
 local lrucache = core.lrucache.new({
     type = 'plugin', serial_creating = true,
 })
+local group_conf_lru = core.lrucache.new({
+    type = 'plugin',
+})
 
 
+local policy_to_additional_properties = {
+    redis = {
+        properties = {
+            redis_host = {
+                type = "string", minLength = 2
+            },
+            redis_port = {
+                type = "integer", minimum = 1, default = 6379,
+            },
+            redis_password = {
+                type = "string", minLength = 0,
+            },
+            redis_database = {
+                type = "integer", minimum = 0, default = 0,
+            },
+            redis_timeout = {
+                type = "integer", minimum = 1, default = 1000,
+            },
+        },
+        required = {"redis_host"},
+    },
+    ["redis-cluster"] = {
+        properties = {
+            redis_cluster_nodes = {
+                type = "array",
+                minItems = 2,
+                items = {
+                    type = "string", minLength = 2, maxLength = 100
+                },
+            },
+            redis_password = {
+                type = "string", minLength = 0,
+            },
+            redis_timeout = {
+                type = "integer", minimum = 1, default = 1000,
+            },
+            redis_cluster_name = {
+                type = "string",
+            },
+        },
+        required = {"redis_cluster_nodes", "redis_cluster_name"},
+    },
+}
 local schema = {
     type = "object",
     properties = {
         count = {type = "integer", exclusiveMinimum = 0},
         time_window = {type = "integer",  exclusiveMinimum = 0},
+        group = {type = "string"},
         key = {type = "string", default = "remote_addr"},
         key_type = {type = "string",
             enum = {"var", "var_combination"},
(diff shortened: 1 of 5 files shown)


t/plugin/limit-count2.t
```diff
@@ -480,11 +448,240 @@ passed
...
+                ngx.status = code
+            end
+            ngx.print(body)
+        }
+    }
+--- error_code: 400
+--- error_log
+[error]
+--- response_body
+{"error_msg":"failed to check the configuration of plugin limit-count err: group conf mismatched"}
```
tzssangglass, line 687: How can we update the limit-count config?

docs/zh/latest/plugins/limit-count.md
```diff
@@ -112,6 +113,54 @@ curl -i http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335
 你也可以通过 web 界面来完成上面的操作，先增加一个 route，然后在插件页面中添加 limit-count 插件：
 ![添加插件](../../../assets/images/plugin/limit-count-1.png)
 
+我们也支持在多个 Route 间共享同样的限流计数器。举个例子，
```
tzssangglass, line 116:
```suggestion
我们也支持在多个 Route 间共享同一个限流计数器。举个例子，
```

When different limit-count plugins have the same group, it means that these plugins share a unique limit counter and requests passing through them will increment on this counter.

I don't know if I'm wrong.

bzp2010 (approved):

docs/zh/latest/plugins/limit-count.md
```diff
@@ -112,6 +113,54 @@ curl -i http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335
...
+    "upstream": {
+        "type": "roundrobin",
+        "nodes": {
+            "127.0.0.1:1980": 1
+        }
+    }
+}'
+```
+
+每个配置了 `group` 为 `services_1#1640140620` 的 Route 都将共享同一个每个 IP 地址只能访问两次的计数器。
```
leslie-tsang, line 139:
```suggestion
每个配置了 `group` 为 `services_1#1640140620` 的 Route 都将共享同一个每个 IP 地址每分钟只能访问一次的计数器。
```
Would be better ?
I'm confused about the count attr of limit-count.
I don't know if I'm wrong.

spacewander: Good catch!

t/plugin/limit-count2.t
spacewander: We need to update the group too. Otherwise, the same group will have two different configuration.

docs/zh/latest/plugins/limit-count.md
spacewander: Yes, you are right.

apisix/plugins/limit-count.lua
```diff
@@ -192,7 +248,12 @@ function _M.access(conf, ctx)
         key = ctx.var["remote_addr"]
     end
 
-    key = key .. ctx.conf_type .. ctx.conf_version
+    if not conf.group then
+        key = key .. ctx.conf_type .. ctx.conf_version
+    else
+        key = key .. conf.group
```
gxthrj, line 254: So traffic will eventually be limited according to the key generated by the combination of key and group.
But `key` can not be empty (if `key` is empty, `remote_addr` will be set as the default key), 
so it is never possible to limit the rate only according to the `group`.

spacewander: Yes, the previous limit-count is two-dimension:
per-route x per-key

This PR allows to use:
group x per-key

If we want to limit only according to the group, we can use a constant variable, like hostname, so the per-key will be always a single point.

In the next PR, I will introduce a new key type that is equal to a constant variable, so we don't need a special configuration for the key part.

spacewander: Look like the hostname is not constant if the cluster limit-count is used. Anyway, we can solve it in the next PR.

tzssangglass (approved):

leslie-tsang (approved):

PR #5888, opened 2021-12-22
docs: added some explanations for the usage of the mqtt proxy

danielkocot:
added some explanations for the usage of the mqtt proxy and removed the remote_addr key from the json example configuration

The PR is needed to make the configuration of the mqtt proxy more clearer.

### DIFF
--- a/docs/en/latest/plugins/mqtt-proxy.md
+++ b/docs/en/latest/plugins/mqtt-proxy.md
@@ -23,11 +23,11 @@ title: mqtt-proxy
 
 ## Summary
 
-- [Summary](#summary)
-- [Name](#name)
-- [Attributes](#attributes)
-- [How To Enable](#how-to-enable)
-- [Delete Plugin](#delete-plugin)
+- [**Summary**](#summary)
+- [**Name**](#name)
+- [**Attributes**](#attributes)
+- [**How To Enable**](#how-to-enable)
+- [**Delete Plugin**](#delete-plugin)
 
 ## Name
 
--- a/docs/en/latest/plugins/mqtt-proxy.md
+++ b/docs/en/latest/plugins/mqtt-proxy.md
@@ -23,7 +23,6 @@ title: mqtt-proxy
 
 ## Summary
 
-- [**Summary**](#summary)
 - [**Name**](#name)
 - [**Attributes**](#attributes)
 - [**How To Enable**](#how-to-enable)
--- a/docs/en/latest/plugins/mqtt-proxy.md
+++ b/docs/en/latest/plugins/mqtt-proxy.md
@@ -23,10 +23,11 @@ title: mqtt-proxy
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


docs/en/latest/plugins/mqtt-proxy.md
```diff
@@ -23,10 +23,11 @@ title: mqtt-proxy
 
 ## Summary
 
-- [**Name**](#name)
-- [**Attributes**](#attributes)
-- [**How To Enable**](#how-to-enable)
-- [**Delete Plugin**](#delete-plugin)
+- [Summary](#summary)
```
spacewander, line 26: We don't need to add an anchor for summary?

danielkocot: You're right.

spacewander: We don't need to change the summary in this PR?

docs/en/latest/plugins/mqtt-proxy.md
```diff
@@ -23,10 +23,11 @@ title: mqtt-proxy
...
+- [Name](#name)
+- [Attributes](#attributes)
+- [How To Enable](#how-to-enable)
+- [Delete Plugin](#delete-plugin)
```
tzssangglass, line 30: why we remove `**`?

I see that all the plugin docs follows the original format

danielkocot: A lot docs also don't use `**` any longer. So in my opinion it is a general question.

tzssangglass: Maybe we should use the tool to make uniform changes. I think we can leave it as it is in this PR. 😁

juzhiyuan: agree to @tzssangglass, and @danielkocot Daniel, if not necessary, we'd better keep the Top of Contents and only change doc contents 😄 if the ToC needs to be updated, we'd better update other plugins' docs.

danielkocot: Now I found out that happened the markdown plugin in Visual Studio Code removed `**`on saving the file.

docs/en/latest/plugins/mqtt-proxy.md
```diff
@@ -23,6 +23,7 @@ title: mqtt-proxy
 
 ## Summary
 
+- [**Summary**](#summary)
```
spacewander, line 26: We don't need to add an index for the Summary. It is just two lines above.

juzhiyuan (approved):

juzhiyuan:
Once this PR gets merged, @guoqqqi could update the Chinese version 😄

tzssangglass (approved):

spacewander (approved):

PR #5890, opened 2021-12-22
fix(cors): compatible with scenarios where origin is modified

zuiyangqingzhou:
In some scenarios, if you modify the origin using proxy_rewrite plugin, the cors will not be handled correctly

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/cors.lua
+++ b/apisix/plugins/cors.lua
@@ -227,6 +227,7 @@ end
 
 
 function _M.rewrite(conf, ctx)
+    ctx["raw-request-origin"] = core.request.header(ctx, "Origin")
     if ctx.var.request_method == "OPTIONS" then
         return 200
     end
@@ -234,7 +235,7 @@ end
 
 
 function _M.header_filter(conf, ctx)
-    local req_origin = core.request.header(ctx, "Origin")
+    local req_origin =  ctx["raw-request-origin"]
     -- Try allow_origins first, if mismatched, try allow_origins_by_regex.
     local allow_origins
     allow_origins = process_with_allow_origins(conf, ctx, req_origin)
--- a/apisix/plugins/cors.lua
+++ b/apisix/plugins/cors.lua
@@ -227,7 +227,8 @@ end
 
 
 function _M.rewrite(conf, ctx)
-    ctx["raw-request-origin"] = core.request.header(ctx, "Origin")
+    -- save the original request origin as it may be changed at other phase
+    ctx.original_request_origin = core.request.header(ctx, "Origin")
     if ctx.var.request_method == "OPTIONS" then
         return 200
     end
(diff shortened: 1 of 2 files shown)


apisix/plugins/cors.lua
```diff
@@ -227,14 +227,15 @@ end
 
 
 function _M.rewrite(conf, ctx)
+    ctx["raw-request-origin"] = core.request.header(ctx, "Origin")
```
spacewander, line 230:
```suggestion
    ctx.original_request_origin = core.request.header(ctx, "Origin")
```
would be better.

Please also add a comment about this change.

tokers (commented):
It seems that all other CORS-related headers also suffer from this change.

apisix/plugins/cors.lua
zuiyangqingzhou: OK

zuiyangqingzhou:
> 



> It seems that all other CORS-related headers also suffer from this change.

Sorry, I don't quite understand what you mean, can you list some cases?

tokers:
> 

Headers like `Access-Control-Expose-Methods`, `Access-Control-Max-Age` will also be modified by `proxy-rewrite` plugin,

zuiyangqingzhou:
> > 
> 
> Headers like `Access-Control-Expose-Methods`, `Access-Control-Max-Age` will also be modified by `proxy-rewrite` plugin,

But `proxy-rewrite` plugin works in the rewrite phase, `Access-Control-Expose-Methods`, `Access-Control-Max-Age` is written to the response header in the header_filter phase,  I understand that it should not be affected

spacewander (approved):

t/plugin/cors.t
```diff
@@ -927,3 +927,70 @@ Access-Control-Max-Age:
...
+                            "allow_origins": "http://sub.domain.com",
+                            "allow_methods": "GET,POST",
+                            "allow_headers": "headr1,headr2",
+                            "expose_headers": "ex-headr1,ex-headr2",
+                            "max_age": 50,
+                            "allow_credential": true
+                        },
+                        "proxy-rewrite": {
+                            "Origin": "http://example.com"
+                        }
```
tzssangglass, line 952:
```suggestion
                        "proxy-rewrite": {
                            "headers": {
                                 "Origin": "http://example.com"
                            }
                        }
```

Modifying the header in proxy-rewrite should be configured like this?

zuiyangqingzhou: oh,  I will fix it

tokers:
> > > 
> > 
> > 
> > Headers like `Access-Control-Expose-Methods`, `Access-Control-Max-Age` will also be modified by `proxy-rewrite` plugin,
> 
> But `proxy-rewrite` plugin works in the rewrite phase, `Access-Control-Expose-Methods`, `Access-Control-Max-Age` is written to the response header in the header_filter phase, I understand that it should not be affected

Oh, forget it, it's my bad.

PR #5964, opened 2021-12-30
feat(grpc-web): support gRPC-Web Proxy

### LINKED ISSUE #5898: proposal: support grpc-web proxy
### Issue description

`gRPC` was originally developed by Google and is a high-performance remote procedure call framework based on `HTTP/2` implementation. But because the browser does not directly expose `HTTP/2`, `Web` applications cannot directly use `gRPC`.
`gRPC-Web` is a standardized protocol that solves this problem. You can use `gRPC` in the browser and provide a `JavaScript` implementation of `gRPC` for the browser client.
The principle is to create an end-to-end `gRPC` pipeline that is compatible with `HTTP/1.1` and `HTTP/2`, and then the browser sends a regular `HTTP` request to the `gRPC-Web` between the browser and the server The proxy converts the request and response.

Currently `Envoy` provides proxy and conversion capabilities of the `gRPC-Web` protocol.  I think `Apache APISIX` can provide friendly proxy support for `gRPC-Web` through plugin, which is meaningful to front-end developers. You can use a more native way to develop microservices or applications based on `gRPC`.

The following are the relevant configuration and technical details.

---

### Name:

- grpc-web

### Configuration:
```json
{
    "strip_path":false
}
```
- `strip_path`:  strip request routing prefix, If set to `true`,  stripped request path will be passed to upstream gRPC service

### Detail:

1. What Apache APISIX needs to do ?

```bash
gRPC-Web Client  <-->  Proxy  <--> gRPC Service
```

Implement Proxy, complete the decoding of data from gRPC-Web data to gRPC Server and the encoding of gRPC Server response to gRPC-Web.

2. Data Format: Use the `protoc-gen-grpc-web` plugin to  generate proto messages and the service client stub from your .proto

	2.1 protoc-gen-grpc-web mode=grpcwebtext: The default generated code sends the payload in the grpc-web-text format.
    	- Content-type: application/grpc-web-text
	    - Payload are base64-encoded.
    	- Both unary and server streaming calls are supported.

	2.2 protoc-gen-grpc-web mode=grpcweb: A binary protobuf format is also supported.
    	- Content-type: application/grpc-web+proto
	    - Payload are in the binary protobuf format.
    	- Only unary calls are supported for now.

refer to: [1]

3. CORS Support:

	3.1 Should follow the CORS spec (Mandatory)
    	- Access-Control-Allow-Credentials to allow Authorization headers
	    - Access-Control-Allow-Methods to allow POST and (preflight) OPTIONS only
    	- Access-Control-Allow-Headers to whatever the preflight request carries
	3.2 The client library is expected to support header overwrites to avoid preflight
	3.3 CSP support to be specified

refer to: [2]

4. For more information and use about grpc-web protocol, please refer to: [3]、[4]、[5]

[1] https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2
[2] https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#cors-support
[3] https://www.npmjs.com/package/grpc-web
[4] https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md
[5] https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md

### Maillist
https://lists.apache.org/thread/k6jtmdz147l0b236d77nhx197s3gxlcz

shuaijinchao:
FIX #5898 

### Pre-submission checklist:


apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,145 @@
...
+function _M.body_filter(conf, ctx)
+    local method = core.request.get_method()
+    if method ~= ALLOW_METHOD_POST then
+        return
+    end
+
+    local chunk = ngx_arg[1]
+
+    if ctx.var.grpc_web_encoding == CONTENT_ENCODING_BASE64 then
+        chunk = encode_base64(chunk)
```
spacewander, line 139: We can only reset `ngx_arg[1]` when encoding is applied.

docs/en/latest/plugins/grpc-web.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Name
+
+The `grpc-web` plugin is a proxy plugin used to process [gRPC Web](https://github.com/grpc/grpc-web) client requests to `gRPC Server`.
+
+gRPC Web Client -> APISIX -> gRPC server
+
+## Attributes
+
+- None
```
spacewander, line 40: We don't need to use list for one word?

apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,146 @@
...
+        core.log.error("reading body err, ", err)
+        return 400
+    end
+
+    if body and encoding == CONTENT_ENCODING_BASE64 then
+        body = decode_base64(body)
+    end
+
+    local ok
+    ok, err = core.request.set_raw_body(body)
```
spacewander, line 109: We don't need to make things so complex. We just read the body in core.request.get_body

apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,146 @@
...
+    core.response.set_header("Content-Type", ctx.grpc_web_mime or DEFAULT_CORS_CONTENT_TYPE)
+end
+
+function _M.body_filter(conf, ctx)
+    -- If the gRPC-Web standard MIME extension type is not obtained or
+    -- the `POST` method is not used for interaction according to the gRPC-Web specification,
+    -- the request body processing will be ignored
+    -- https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#cors-support
+    local method = core.request.get_method()
+    if not ctx.grpc_web_mime or method ~= ALLOW_METHOD_POST then
```
spacewander, line 139: If the grpc_web_mime is set, the method must be POST?

docs/en/latest/plugins/grpc-web.md
```diff
@@ -0,0 +1,85 @@
...
+## Name
+
+The `grpc-web` plugin is a proxy plugin used to process [gRPC Web](https://github.com/grpc/grpc-web) client requests to `gRPC Server`.
+
+gRPC Web Client -> APISIX -> gRPC server
+
+## How To Enable
+
+To enable the `gRPC Web` proxy plugin, routing must use the `Prefix matching` pattern (for example: `/*` or `/grpc/example/*`),
+Because the `gRPC-Web` client will pass the `package name`, `service interface name`, `method name` and other information declared in the `proto` in the URI (for example: `/path/a6.RouteService/Insert`) ,
```
spacewander, line 40:
```suggestion
Because the `gRPC Web` client will pass the `package name`, `service interface name`, `method name` and other information declared in the `proto` in the URI (for example: `/path/a6.RouteService/Insert`) ,
```

apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,145 @@
...
+
+    -- set grpc body
+    local body, err = core.request.get_body()
+    if err then
+        core.log.error("reading body err, ", err)
+        return 400
+    end
+
+    if encoding == CONTENT_ENCODING_BASE64 then
+        body = decode_base64(body)
```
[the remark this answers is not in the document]
shuaijinchao: updated

t/plugin/grpc-web.t
```diff
@@ -0,0 +1,135 @@
...
+--- response_body
+{"name":"route02","path":"path02"}
+
+
+
+=== TEST 9: Get all data through APISIX gRPC-Web Proxy
+--- exec
+node ./t/plugin/grpc-web/client.js all
+--- response_body
+[["2",{"name":"route02","path":"path02"}],["3",{"name":"route03","path":"path03"}]]
```
spacewander, line 135: Do we need two more cases:
1. check the CORS header for OPTION/non-OPTION requests
2. configure a route with absolute match

shuaijinchao: Added

apisix/plugins/grpc-web.lua
shuaijinchao: yes, updated

apisix/plugins/grpc-web.lua
shuaijinchao: done

t/plugin/grpc-web.t
```diff
@@ -0,0 +1,208 @@
...
+
+
+
+=== TEST 10: test options request
+--- request
+OPTIONS /grpc/a6.RouteService/GetAll
+--- error_code: 204
```
spacewander, line 142: We need to check the headers with `--- response_headers`

spacewander (approved):

t/plugin/grpc-web.t
shuaijinchao: ok, done

t/plugin/grpc-web.t
```diff
@@ -0,0 +1,212 @@
...
+--- response_headers
+Access-Control-Allow-Methods: POST
```
spacewander, line 144: Let's check the header for POST request too

shuaijinchao: added

apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,146 @@
...
+    local mimetype = core.request.header(ctx, "Content-Type")
+    local encoding = grpc_web_content_encoding[mimetype]
+    if not encoding then
+        core.log.error("request Content-Type: `", mimetype, "` invalid")
+        return 400
+    end
+
+    -- set grpc path
+    if not (ctx.curr_req_matched and ctx.curr_req_matched[":ext"]) then
+        core.log.error("please use matching pattern for routing URI, for example: /* or /grpc/*")
```
tzssangglass, line 83: This log looks more like it is being returned to the client.

apisix/plugins/grpc-web.lua
```diff
@@ -0,0 +1,146 @@
...
+    if path:byte(1) ~= core.string.byte("/") then
+        path = "/" .. path
+    end
+
+    req_set_uri(path)
+
+    -- set grpc body
+    local body, err = core.request.get_body()
+    if err then
+        core.log.error("reading body err, ", err)
```
tzssangglass, line 97:
```suggestion
        core.log.error("failed to read request body: ", err)
```
is better?

docs/zh/latest/plugins/grpc-web.md
```diff
@@ -0,0 +1,84 @@
...
+## 测试插件
+
+- 请求方式仅支持 `POST` 和 `OPTIONS`，参考：[CORS support](https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#cors-support) 。
+- 内容类型支持 `application/grpc-web`、`application/grpc-web-text`、`application/grpc-web+proto`、`application/grpc-web-text+proto`，参考：[Protocol differences vs gRPC over HTTP2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2) 。
+- 客户端部署，参考：[gRPC-Web Client Runtime Library](https://www.npmjs.com/package/grpc-web) 或 [Apache APISIX gRPC Web 测试框架](https://github.com/apache/apisix/tree/master/t/plugin/grpc-web) 。
+- 完成 `gRPC Web` 客户端部署后，即可通过 `浏览器` 或 `node` 向 `APISIX` 发起 `gRPC Web` 代理请求。
+
+## 禁用插件
+
+只需删除插件配置中`grpc-web`的JSON配置即可。 APISIX 插件是热加载的，所以不需要重启 APISIX。
```
tzssangglass, line 69:
```suggestion
只需删除插件配置中 `grpc-web` 的JSON配置即可。 APISIX 插件是热加载的，所以不需要重启 APISIX。
```

apisix/plugins/grpc-web.lua
shuaijinchao: thx~ updated

apisix/plugins/grpc-web.lua
shuaijinchao: yes, I updated it

docs/zh/latest/plugins/grpc-web.md
shuaijinchao: done

tzssangglass (approved):

Issue #5910, opened 2021-12-24
request help: can the log-plugins supports to logging upstream_latency and apisix_latency?


jagerzhang:
### Issue description

现在所有日志插件就只能上报 latency 字段，能否将后端耗时和APISIX本身的耗时也分开记录下，方便运营分析？

### Environment

APISIX version: 2.11.0

tokers:
You can customize the `log_format`field in plugin metadata according to your needs.

shoulda:
> ### Issue description
> 现在所有日志插件就只能上报 latency 字段，能否将后端耗时和APISIX本身的耗时也分开记录下，方便运营分析？
> 
> ### Environment
> APISIX version: 2.11.0

请教一个问题，我们这边准备用prometheus来上报日志的，但是如何数据类型是counter，是否会有溢出现象或者受限于lua_shared_dict prometheus-metrics  的大小，你们这边又遇到这个问题吗？

jagerzhang:
> You can customize the `log_format`field in plugin metadata according to your needs.

用log_format好像要把所有的字段都填写上才行吧？ 并不是追加的逻辑。

jagerzhang:
> > You can customize the `log_format`field in plugin metadata according to your needs.
> 
> 试了下没生效，配置如下：
> 
> ```
> {
>   "batch_max_size": 1,
>   "broker_list": {
>     "X.X.X.X": 9092
>   },
>   "disable": false,
>   "include_req_body": true,
>   "include_resp_body": true,
>   "kafka_topic": "xxxxx",
>   "log_format": {
>     "apisix_latency": "$apisix_latency",
>     "upstream_latency": "$upstream_latency"
>   },
>   "max_retry_count": 3
> }
> ```
> 
> 求指引~


仔细看了下好像是要 PUT 到 plugin_metadata里面，那这里有个问题，我并不知道默认的log_format是怎么写的，如果我直接改成 
```
{
"log_format": {
    "apisix_latency": "$apisix_latency",
    "upstream_latency": "$upstream_latency"
  }
}
```
那其他字段是不是都没了？只剩下这2个字段？希望可以提供一个默认的log_format，然后我们追加下这2个字段.....

jagerzhang:
> > > You can customize the `log_format`field in plugin metadata according to your needs.
> > 
> > 
> > 试了下没生效，配置如下：
> > ```
> > {
> >   "batch_max_size": 1,
> >   "broker_list": {
[...]

PR #6008, opened 2022-01-04
feat: supports to logging apisix_latency and upstream_latency

jagerzhang:
supports to logging apisix_latency and upstream_latency(copy from tapisix by [kyroslin](https://github.com/kyroslin)): https://github.com/apache/apisix/issues/5910

### Pre-submission checklist:
The original code is used directly without adding other new logic.


apisix/utils/log-util.lua
```diff
@@ -120,7 +122,9 @@ local function get_full_log(ngx, conf)
         consumer = consumer,
         client_ip = core.request.get_remote_client_ip(ngx.ctx.api_ctx),
         start_time = ngx.req.start_time() * 1000,
-        latency = (ngx_now() - ngx.req.start_time()) * 1000
+        latency = latency,
+        upstream_latency = upstream_latency,
```
spacewander, line 126: I think we should discuss it in the mail list first when adding a new field to the full log format.

Is it commonly required? Or it is just used in some companies?

spacewander (changes requested):

apisix/utils/log-util.lua
jagerzhang: ok, I'll continue to activate the previous discussion.

PR #6018, opened 2022-01-05
feat: support to add limit_size field to the hold_body_chunk method...

### LINKED ISSUE #5915: request help: logger plugins supports limit size of request/response body
### Discussed in https://github.com/apache/apisix/discussions/5892

<div type='discussions-op-text'>

<sup>Originally posted by **lrfdehao** December 23, 2021</sup>
Suppose I have two fields, `include_resp_body` as the switch, `resp_limit_size` as the limit size. After configuring these two parameters, if resp_body exceeds the size of resp_limit_size, resp_body will not be recorded in the log. The same is true for req_body</div>

kyroslin:

Fix #5915
support to add limit_size field to the hold_body_chunk method to truncate response_body

Combine limit_size with the context's body_buffer_size, body_hold_flag fields to control the truncation of response body, where limit_size is controlled by specific business plugins

### Pre-submission checklist:

### DIFF
--- a/apisix/core/response.lua
+++ b/apisix/core/response.lua
@@ -126,22 +126,18 @@ local function set_header(append, ...)
     end
 end
 
-
 function _M.set_header(...)
     set_header(false, ...)
 end
-
 
 function _M.add_header(...)
     set_header(true, ...)
 end
 
-
 function _M.get_upstream_status(ctx)
     -- $upstream_status maybe including multiple status, only need the last one
     return tonumber(str_sub(ctx.var.upstream_status or "", -3))
 end
-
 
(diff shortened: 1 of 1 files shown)


spacewander (changes requested):
I am afraid we can't accept a feature in the `core` library which is not used by any code in APISIX.

If the feature is not used by APISIX, we may consider this feature is no longer used someday in the future and remove it.

If your own plugin needs this, you can write your own `hold_body_chunk_with_limit` in a different name.

Thanks for your cooperation.

kyroslin:
Okay, thank you, let me turn off this pr first, and if there is a suitable scene later, I will look at the opportunity

PR #6023, opened 2022-01-05
fix(ipv6): allow disabling IPv6 resolve

### LINKED ISSUE #6017: request help: Provide a configuration to disable ipv6
### Issue description

When APISIX parses the upstream host, it may be parsed as `ipv6` or `ipv4`. 
Hope to provide a configuration in `config.yaml` to have the ability to disable `ipv6`.

### Environment

- apisix version (cmd: `apisix version`): master
- OS (cmd: `uname -a`): centos 7
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): 1.19.*
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API): 3.4.14
- apisix-dashboard version, if have: - 
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner): -
- luarocks version, if the issue is about installation (cmd: `luarocks --version`): -

leslie-tsang:

Sync `apisix.enable_ipv6` config to the generated `nginx.conf` resolver section

fix #6017 

### Pre-submission checklist:

### DIFF
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -552,6 +552,7 @@ Please modify "admin_key" in conf/config.yaml .
         admin_server_addr = admin_server_addr,
         control_server_addr = control_server_addr,
         prometheus_server_addr = prometheus_server_addr,
+        enable_ipv6 = yaml_conf.apisix.enable_ipv6,
     }
 
     if not yaml_conf.apisix then
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -552,7 +552,6 @@ Please modify "admin_key" in conf/config.yaml .
         admin_server_addr = admin_server_addr,
         control_server_addr = control_server_addr,
         prometheus_server_addr = prometheus_server_addr,
-        enable_ipv6 = yaml_conf.apisix.enable_ipv6,
     }
 
     if not yaml_conf.apisix then
--- a/apisix/core/dns/client.lua
+++ b/apisix/core/dns/client.lua
@@ -22,6 +22,7 @@ local insert_tab = table.insert
 local math_random = math.random
 local package_loaded = package.loaded
 local ipairs = ipairs
+local table_remove = table.remove
 local setmetatable = setmetatable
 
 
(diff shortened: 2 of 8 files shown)


apisix/cli/ops.lua
```diff
@@ -552,6 +552,7 @@ Please modify "admin_key" in conf/config.yaml .
         admin_server_addr = admin_server_addr,
         control_server_addr = control_server_addr,
         prometheus_server_addr = prometheus_server_addr,
+        enable_ipv6 = yaml_conf.apisix.enable_ipv6,
```
spacewander, line 555: The enable_ipv6 is already imported below

tzssangglass (approved):

tokers (approved):

apisix/core/dns/client.lua
```diff
@@ -130,7 +130,7 @@ end
 
 
 function _M.new(opts)
-    opts.ipv6 = true
+    opts.ipv6 = opts.ipv6 or true
```
spacewander, line 133: Why not `= local_conf.apisix.enable_ipv6`? And we need a test for it.

leslie-tsang: > Why not = local_conf.apisix.enable_ipv6?

IMO, `= local_conf.apisix.enable_ipv6` is a breaking change, Imagine a scenario where a third party or serverless plugin uses this function to parse ipv6, this change will causes an unexpected behavior, I wander if this is a fictitious demand.
Do let me know WDYT. Thanks.

> And we need a test for it.

Sure, I will add it.

spacewander: We can use
```
if opts.ipv6 == nil then
  ...
end
```
so user can specify the ipv6 option to overwrite the default one.

AFAIK, we don't promise the DNS client always handles IPv6 query and the `enable_ipv6` is enabled by default, so it's not a break change. (Even if users don't  consider it is a bug fix)

leslie-tsang: OK, I will add test case for it in next PR

apisix/core/dns/client.lua
```diff
@@ -130,7 +132,22 @@ end
 
 
 function _M.new(opts)
-    opts.ipv6 = true
+    local local_conf = config_local.local_conf()
+
+    if opts.ipv6 == nil then
```
spacewander, line 137: I spend some time digging into it. Look like it should be `enable_ipv6`, and it only works with the nameserver. See http://kong.github.io/lua-resty-dns-client/modules/resty.dns.client.html#init

This field is incorrect from the beginning...

leslie-tsang: 🤣 done

spacewander (approved):

PR #6037, opened 2022-01-06
feat: add forward-auth plugin

### LINKED ISSUE #5475: request help: Support forward-auth
### Issue description

We are using traefik and want to migrate to apisix, but traefik has a forward-auth plugin, which is a plugin that supports external authentication services, how does apisix support external authentication services?
image

### Environment

- apisix version (cmd: `apisix version`):2.10.0
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):openresty/1.19.3.1

### LINKED ISSUE #6007: Proposal: support Forward Auth authenticate plugin
## Background

Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request and replaces the result when the authentication service responds with a non-20x state. In this way, we can achieve a custom error return or user redirection to the authentication page if the authentication fails.

### Configure schema

Name | Type | Requirement | Default | Description
-- | -- | -- | -- | --
host | string | required |   | Auth service host (eg. https://localhost:8999)
ssl_verify | boolean | optional | true | Whether to verify the certificate
request_headers | array | optional | [] | Request headers allowed to be forwarded
response_headers | array | optional | [] | Response headers allowed to be returned

### Implementation

Through the plugin, part of the requested information is forwarded to the authentication service during the access phase, which will transmit the following information by way of a request header.

Method | Schema | Domain | Path | Source IP
-- | -- | -- | -- | --
X-Forwarded-Method | X-Forwarded-Proto | X-Forwarded-Host | X-Forwarded-Uri | X-Forwarded-For

At the same time, the request headers will be forwarded according to the list of request headers set in the `request_headers` in the configuration, if not set, all request headers will be forwarded. After the authentication service response, if the response code is 20x, the request will be forwarded upstream normally, otherwise the response of the authentication service will be forwarded to the client, and the response headers will be filtered using `response_headers`.

## Other
What are your ideas?

bzp2010:

Add the `forward-auth` plugin.

Fix #6007
Fix #5475
### Pre-submission checklist:


apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,145 @@
...
+    if #conf.request_headers > 0 then
+        for _, header in ipairs(conf.request_headers) do
+            auth_headers[header] = core.request.header(ctx, header)
+        end
+    else
+        auth_headers = core.table.merge(core.request.headers(), auth_headers)
+    end
+
+    local params = {
+        method = "GET",
```
shuaijinchao, line 96: Should the request method be mapped to the client request method?

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,145 @@
...
+    local httpc = http.new()
+    httpc:set_timeout(conf.timeout)
+
+    local res, err = httpc:request_uri(conf.host, params)
+
+    -- block by default when authorization service is unavailable
+    if not res or err then
+        core.log.error("failed to process forward auth, err: ", err)
+        return 403
+    end
```
shuaijinchao, line 116:
```suggestion
    if not res then
        core.log.error("failed to process forward auth, err: ", err)
        return 403
    end
```
Would it be better to judge this way?

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,143 @@
...
+        ["X-Forwarded-For"] = core.request.get_remote_client_ip(ctx),
+    }
+
+    -- append headers that need to be get from the client request header
+    if #conf.request_headers > 0 then
+        for _, header in ipairs(conf.request_headers) do
+            auth_headers[header] = core.request.header(ctx, header)
+        end
+    else
+        auth_headers = core.table.merge(core.request.headers(), auth_headers)
```
spacewander, line 92: Ditto

apisix/plugins/forward-auth.lua
bzp2010: I don't think it's needed unless we provide other special features for this, the implementation in other software is the same.

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Summary
+
+- [**Description**](#description)
+- [**Attributes**](#attributes)
+- [**Example**](#example)
+
+## Description
+
+The `forward-auth` plugin implement a classic external authentication model. We can implement a custom error return or user redirection to the authentication page if the authentication fails.
```
spacewander, line 32:
```suggestion
The `forward-auth` plugin implements a classic external authentication model. We can implement a custom error return or user redirection to the authentication page if the authentication fails.
```

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+
+Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request and replaces the result when the authentication service responds with a non-20x status.
```
spacewander, line 34:
```suggestion
Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request, and replaces the result when the authentication service responds with a non-2xx status.
```

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+        "serverless-pre-function": {
+            "phase": "rewrite",
+            "functions": [
+                "return function (conf, ctx) local core = require(\"apisix.core\"); local authorization = core.request.header(ctx, \"Authorization\"); if authorization == \"123\" then core.response.exit(200); elseif authorization == \"321\" then core.response.set_header(\"X-User-ID\", \"i-am-user\"); core.response.exit(200); else core.response.set_header(\"Location\", \"http://example.com/auth\"); core.response.exit(403); end end"
+            ]
+        }
+    },
+    "upstream": {
+        "nodes": {},
+        "scheme": "https",
```
spacewander, line 70: Let's remove useless field

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+{
+    "headers": {
+        "Authorization": "321",
+        "X-User-ID": "i-am-user",
+        "Next": "More-headers"
+    }
+}
+```
+
+3. **client_headers** Send `authorization` service response header to `client` when authorize failure
```
spacewander, line 127:
```suggestion
3. **client_headers** Send `authorization` service response header to `client` when authorizing failed
```

t/plugin/forward-auth.t
```diff
@@ -0,0 +1,157 @@
...
+    location /t {
+        content_by_lua_block {
+            local datas = {
+                {
+                    url = "/apisix/admin/routes/auth",
+                    data = [[{
+                        "plugins": {
+                            "serverless-pre-function": {
+                                "phase": "rewrite",
+                                "functions": ["return function (conf, ctx) local core = require(\"apisix.core\"); local authorization = core.request.header(ctx, \"Authorization\"); if authorization == \"123\" then core.response.exit(200); elseif authorization == \"321\" then core.response.set_header(\"X-User-ID\", \"i-am-an-user\"); core.response.exit(200); else core.response.set_header(\"Location\", \"http://example.com/auth\"); core.response.exit(403); end end"]
```
spacewander, line 76: Let's log down the X-Forwarded-XX received by the service. And better to break down the very long code.

t/plugin/forward-auth.t
```diff
@@ -0,0 +1,157 @@
...
+qr/\"x-user-id\":\"i-am-an-user\"/
+
+
+
+=== TEST 5: hit route (test client_headers)
+--- request
+GET /hello
+--- error_code: 403
+--- response_headers
+Location: http://example.com/auth
```
spacewander, line 157: Let's add a test that the client passes X-Forwarded-XX by itself

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+
+Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request and replaces the result when the authentication service responds with a non-20x status.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| host | string | required |  |  | Authorization service host (eg. https://localhost:9188) |
+| ssl_verify | boolean | optional | true |   | Whether to verify the certificate |
+| request_headers | array[string] | optional |  |  | `client` request header that will be sent to the `authorization` service |
```
spacewander, line 42: Let's doc the behavior when this field is empty

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,135 @@
...
+| upstream_headers | array[string] | optional |  |  | `authorization` service response header that will be sent to the `upstream` |
```
spacewander, line 43: Ditto

apisix/plugins/forward-auth.lua
bzp2010: I don't understand how to do it really right, I noticed that in other plugin implementations there are those that use `not res` alone, those that use `err` alone and those that use `not res or err`, which one should we take?

https://github.com/apache/apisix/blob/58fec8f5eb599c93c3491e200a4ab66907a3dd24/apisix/plugins/authz-keycloak.lua#L251-L254

https://github.com/apache/apisix/blob/58fec8f5eb599c93c3491e200a4ab66907a3dd24/apisix/plugins/serverless/generic-upstream.lua#L110-L113

https://github.com/apache/apisix/blob/58fec8f5eb599c93c3491e200a4ab66907a3dd24/apisix/plugins/wolf-rbac.lua#L145-L147

I hope to get your guidance.
cc @spacewander

apisix/plugins/forward-auth.lua
bzp2010: In our or other forward-auth plugins, authentication-related data is sent via fixed requests, not POST data, so I think that's enough too.

ping @shuaijinchao

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,143 @@
...
+            default = {},
+            items = {type = "string"},
+            description = "authorization response header that will be sent to the upstream"
+        },
+        client_headers = {
+            type = "array",
+            default = {},
+            items = {type = "string"},
+            description = "authorization response header that will be sent to"
+                           .. "the client when authorize failure"
```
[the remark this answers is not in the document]
bzp2010: changed

docs/en/latest/plugins/forward-auth.md
bzp2010: changed

docs/en/latest/plugins/forward-auth.md
bzp2010: changed

docs/en/latest/plugins/forward-auth.md
bzp2010: changed

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,143 @@
...
+        ["X-Forwarded-Method"] = core.request.get_method(),
+        ["X-Forwarded-Host"] = core.request.get_host(ctx),
+        ["X-Forwarded-Uri"] = core.request.get_uri(ctx),
+        ["X-Forwarded-For"] = core.request.get_remote_client_ip(ctx),
+    }
+
+    -- append headers that need to be get from the client request header
+    if #conf.request_headers > 0 then
+        for _, header in ipairs(conf.request_headers) do
+            auth_headers[header] = core.request.header(ctx, header)
```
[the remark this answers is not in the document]
bzp2010: fixed, because `auth_headers` has been set in our concern headers, after the judgment, the request header from the client can not override them, that's not trust the client header.

https://github.com/apache/apisix/blob/8a853d345436e842845ae5bd57de8bb60ac8aee7/apisix/plugins/forward-auth.lua#L78-L93

apisix/plugins/forward-auth.lua
bzp2010: Because, merge the list of defined headers into the client request header, any key headers passed in by the client are overwritten, so I implement the untrusted client header.

docs/en/latest/plugins/forward-auth.md
bzp2010: added

docs/en/latest/plugins/forward-auth.md
bzp2010: added

docs/en/latest/plugins/forward-auth.md
bzp2010: removed

t/plugin/forward-auth.t
bzp2010: added, I added a test case to demonstrate that these client-side incoming key headers are indeed dropped in APISIX.

t/plugin/forward-auth.t
bzp2010: I used another method to record these heads. I also added a simple `echo` function under `/auth` route.

apisix/core/request.lua
```diff
@@ -278,6 +278,15 @@ function _M.get_path(ctx)
 end
 
 
+function _M.get_uri(ctx)
```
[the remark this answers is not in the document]
bzp2010: So I should just use ctx.var.request_uri instead of wrapping it? Can I merge the functionality of `get_path` into `get_uri` and add a `with_args` parameter indicating whether or not I need to carry parameters.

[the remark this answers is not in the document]
spacewander: Don't make thing complex!
The name of request_uri and uri already show the difference. There is no need to wrap them into a function which is rarely used. Premature optimization is the source of evil.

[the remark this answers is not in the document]
spacewander: We can remove the `get_path` in the next PR. Better to get rid of it before the release, so that people won't ask why we use `get_path` in one place but use `ctx.var.uri` in the most elsewhere.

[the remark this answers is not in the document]
bzp2010: removed

apisix/plugins/forward-auth.lua
bzp2010: Then, referring to the way it is usually used, I will determine if `err` is `not nil`.

```lua
if err then
    xxx
end
```

bzp2010: changed

bzp2010: Also, I think there needs to be a specification of what rules developers should follow to handle error checking (at least on the http client), the mix of different usages is confusing.

t/plugin/forward-auth.t
```diff
@@ -0,0 +1,178 @@
...
+        content_by_lua_block {
+            local datas = {
+                {
+                    url = "/apisix/admin/routes/auth",
+                    data = [[{
+                        "plugins": {
+                            "serverless-pre-function": {
+                                "phase": "rewrite",
+                                "functions": [
+                                    "return function (conf, ctx) local core = require(\"apisix.core\"); if core.request.header(ctx, \"Authorization\") == \"111\" then core.response.exit(200); end end",
```
spacewander, line 77: Let's indent the code. Like:
```
diff --git t/plugin/forward-auth.t t/plugin/forward-auth.t
index dca35b76..285e60fb 100644
--- t/plugin/forward-auth.t
+++ t/plugin/forward-auth.t
@@ -74,7 +74,12 @@ property "request_headers" validation failed: wrong type: expected array, got st
                             "serverless-pre-function": {
                                 "phase": "rewrite",
                                 "functions": [
-                                    "return function (conf, ctx) local core = require(\"apisix.core\"); if core.request.header(ctx, \"Authorization\") == \"111\" then core.response.exit(200); end end",
+                                    "return function (conf, ctx)
+                                    local core = require(\"apisix.core\");
+                                    if core.request.header(ctx, \"Authorization\") == \"111\" then
+                                        core.response.exit(200)
+                                    end
+                                    end",
                                     "return function (conf, ctx) local core = require(\"apisix.core\"); if core.request.header(ctx, \"Authorization\") == \"222\" then core.response.set_header(\"X-User-ID\", \"i-am-an-user\"); core.response.exit(200); end end",
                                     "return function (conf, ctx) local core = require(\"apisix.core\"); if core.request.header(ctx, \"Authorization\") == \"333\" then core.response.set_header(\"Location\", \"http://example.com/auth\"); core.response.exit(403); end end",
                                     "return function (conf, ctx) local core = require(\"apisix.core\"); if core.request.header(ctx, \"Authorization\") == \"444\" then core.response.exit(403, core.request.headers(ctx)); end end"
```

spacewander: Would be better to log down the "X-Forwarded-XX" headers and check it

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,145 @@
...
+    type = "object",
+    properties = {
+        host = {type = "string"},
+        ssl_verify = {
+            type = "boolean",
+            default = true,
+        },
+        request_headers = {
+            type = "array",
+            default = {},
```
spacewander, line 32: Look like we can't support forward no headers from the client?
If this field is empty, all headers are forwarded. What about don't provide a default value? (Use empty array if no headers are need)

apisix/plugins/forward-auth.lua
```diff
@@ -0,0 +1,145 @@
...
+
+    if res.status >= 300 then
+        local client_headers = {}
+
+        if #conf.client_headers > 0 then
+            for _, header in ipairs(conf.client_headers) do
+                client_headers[header] = res.headers[header]
+            end
+        else
+            client_headers = res.headers
```
spacewander, line 128: IMHO, would be better to use allowlist? If no client_headers, no headers will be returned to the client. Some headers are not expected to be overridden, like Server / Date.

apisix/plugins/forward-auth.lua
spacewander: Currently, we have to enforce it with code review. Maybe we can invest in linter in the future.

apisix/plugins/forward-auth.lua
shuaijinchao: OK, use `GET` as a fixed request method you should remove it because the default request method is `GET`

bzp2010: > OK, use `GET` as a fixed request method you should remove it because the default request method is `GET`

Oh, I get it. Modify will be made later.

apisix/plugins/forward-auth.lua
shuaijinchao: I think it should be used

````lua
if res then
     err
end
````

instead of

```lua
if err then
     err
end
```

this way of writing may be more suitable for languages such as `Go`

the common exception function return in OpenResty is `return nil, err`, the first value is enough to know whether to handle `err`, and you can refer to more code and test cases in `lua-resty-core`, such as: https://github.com/openresty/lua-resty-core/blob/e5217414669c100b334940b250d5340911a02dd0/t/ctx.t#L133-L143

t/plugin/forward-auth.t
bzp2010: As far as I know, JSON does not support direct line feeds. And according to the code of serverless plugin, each item in functions is a function, not a combination of code execution. Perhaps it can only be written in one line. 🤔

bzp2010: Log check for `X-Forwarded-XX`, I will add it later.

bzp2010: Splitting the single line long code is done, which uses `]]..[[` truncates the data code, the code is so long that it cannot be parsed, so I modified it by referring to the explanation in this issue.

[https://github.com/openresty/lua-nginx-module/issues/1622](https://github.com/openresty/lua-nginx-module/issues/1622)

apisix/plugins/forward-auth.lua
bzp2010: I think it's rightfully so, and that's an oversight on my part. I have reversed the meaning of `request_headers` and `client_headers`, if the user does not set this parameter, APISIX will send nothing.

apisix/plugins/forward-auth.lua
bzp2010: Ditto

t/plugin/forward-auth.t
bzp2010: > ~~Log check for `X-Forwarded-XX`, I will add it later.~~

In the session of adding more cases, I found out that we won't actually use ta, and that the function to print the request header is already implemented using `echo`.

apisix/plugins/forward-auth.lua
bzp2010: removed

apisix/plugins/forward-auth.lua
bzp2010: changed

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,130 @@
...
+
+Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request, and replaces the result when the authentication service responds with a non-2xx status.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| host | string | required |  |  | Authorization service host (eg. https://localhost:9188) |
+| ssl_verify | boolean | optional | true |   | Whether to verify the certificate |
+| request_headers | array[string] | optional |  |  | `client` request header that will be sent to the `authorization` service. When it is not set, all `client` request headers are sent to the `authorization` service, except for those provided by APISIX (X-Forwarded-XXX). |
```
spacewander, line 42: Need to update the doc

t/plugin/forward-auth.t
```diff
@@ -0,0 +1,246 @@
...
+                    }]],
+                },
+                {
+                    url = "/apisix/admin/routes/auth",
+                    data = [[{
+                        "plugins": {
+                            "serverless-pre-function": {
+                                "phase": "rewrite",
+                                "functions": [
+                                    "return function(conf, ctx)
```
spacewander, line 86: Would be better to log down the APISIX generated "X-Forwarded-XX" headers and check it

bzp2010: I've merged it with the checkcase for overriding user input requests, which will check all APISIX generated headers at once.

https://github.com/apache/apisix/blob/beea582ddff8879e94484f84202307c488f30a70/t/plugin/forward-auth.t#L221-L225

docs/en/latest/plugins/forward-auth.md
bzp2010: updated

shuaijinchao (approved):

docs/en/latest/plugins/forward-auth.md
```diff
@@ -0,0 +1,130 @@
...
+
+Forward Auth cleverly moves the authentication and authorization logic to a dedicated external service, where the gateway forwards the user's request to the authentication service and blocks the original request, and replaces the result when the authentication service responds with a non-2xx status.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
+| -- | -- | -- | -- | -- | -- |
+| host | string | required |  |  | Authorization service host (eg. https://localhost:9188) |
+| ssl_verify | boolean | optional | true |   | Whether to verify the certificate |
+| request_headers | array[string] | optional |  |  | `client` request header that will be sent to the `authorization` service. When it is not set, no `client` request headers are sent to the `authorization` service, except for those provided by APISIX (X-Forwarded-XXX). |
```
spacewander, line 42: Let's provide a list of the `X-Forwarded-XXX` headers

bzp2010: the `Data Definition` added

spacewander (approved):

PR #6040, opened 2022-01-07
Dev/key auth customize error msg

nimohunter:
What this PR does / why we need it:

Add customized rejected error code and message for key-auth plugin.

Pre-submission checklist:

- [x] update doc
- [x] update t

#6025

### DIFF
--- a/apisix/plugins/key-auth.lua
+++ b/apisix/plugins/key-auth.lua
@@ -31,7 +31,7 @@ local schema = {
         query = { type = "string", default = "apikey"},
         rejected_code = {type = "integer", default = 401},
         rejected_miss_key_msg = {type = "string", default = "{\"message\":\"Missing API key found in request\"}"},
-        rejected_miss_consumer_msg = {type = "string", default = "{\"message\":\"Missing API key found in request\"}"},
+        rejected_miss_consumer_msg = {type = "string", default = "{\"message\":\"Missing related consumer\"}"},
         rejected_error_consumer_msg = {type = "string", default = "{\"message\":\"Invalid API key in request\"}"},
     },
 }
--- a/apisix/plugins/key-auth.lua
+++ b/apisix/plugins/key-auth.lua
@@ -27,14 +27,12 @@ local lrucache = core.lrucache.new({
 local schema = {
     type = "object",
     properties = {
-        header = {
-            type = "string",
-            default = "apikey",
-        },
-        query = {
-            type = "string",
-            default = "apikey",
-        },
+        header = { type = "string", default = "apikey"},
+        query = { type = "string", default = "apikey"},
+        rejected_code = {type = "integer", default = 401},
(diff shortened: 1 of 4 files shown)


leslie-tsang:
Hello there, the title and description need to be updated to make the PR more reviewer friendly.

nimohunter:
> Hello there, the title and description need to be updated to make the PR more reviewer friendly.

Sorry about it

spacewander (changes requested):
We should use a common solution to solve this problem, instead of adding similar fields to each plugin. For example, the disable in the plugin conf: https://github.com/apache/apisix/blob/18c316f8a00f6c125a92df09da95774d5f17f241/apisix/plugin.lua#L137

This has been discussed before.

nimohunter:
But I don't think the common error return code and message was conflicted with the plugin's customize return error code.

plugin doesn't return only one error message, maybe return kinds of error messages.

for example: key-auth should customize three kinds of error message : 
   1. Missing API key found in request
   2. Missing related consumer
   3. Invalid API key in request
 
so i think there is no conflict between  common error msg and plugin's customized error msg

spacewander:
If you want to customize each error msg to a different value, my suggestion is to fork and modify this plugin by yourself. It's inconvenient to maintain such an API interface. We may introduce refactor in the future and change the number of "return xxx".

How many `return xxx` in a function is an implementation detail and we can't / should not maintain a public API based on it.

nimohunter:
@leslie-tsang @spacewander  how about we walk in two steps
 first, support this function about key-auth return customized error msg 
 then, we open a new MR to fix all  plugin return error msg

nimohunter:
by the way, I think authentication into two types:

1. Personal authentication, such as QQ. in this way, the gateway should to access the other permission system to determine whether it has authority. This kind of authentication will not use the gateway auth.

2. Calls between services. For example, I have an API that I want to provide to colleagues in other departments. This logic is really suitable for key-auth and consumers.

so i think key-auth is actually a completely underestimated plugin.

membphis:
Personally, I think it’s not good to customize the rejection code. 
But I also found that some friends in the community have different opinions on this.

@nimohunter, if you are very interested in this feature, I think we can discuss it on the Apache mailing list.

nimohunter:
I agree with you, error code may not be customized. But how about the reject error msg? 

Error msg can be more friendly

membphis:
> I agree with you, error code may not be customized. But how about the reject error msg?
> 
> Error msg can be more friendly

we can discuss it in Apache mail list, follow the Apache Way. 
  
https://lists.apache.org/list.html?dev@apisix.apache.org

This seems to be a meaningful change.

nimohunter:
OK, i have send a mail about this MR...
and then?

membphis:
> OK, i have send a mail about this MR... and then?

we can discuss the idea. If we can reach a consensus, we can do things according to the consensus.

Issue #6029, opened 2022-01-06
discuss: It is not inappropriate using “apisix”  type for prometheus metrics（count first bytes from client）


zhendongcmss:
### Issue description

Because the `apisix_latency` include read contents from client, not ONLY apisix process time. It is very confused on the  dashbord and got resault that apisix lead  to high latency.
https://github.com/apache/apisix/blob/aff5d825aaea12eab312e65e82d060ab9be9b06f/apisix/utils/log-util.lua#L189
https://github.com/apache/apisix/blob/aff5d825aaea12eab312e65e82d060ab9be9b06f/apisix/plugins/prometheus/exporter.lua#L164

![image](https://user-images.githubusercontent.com/88528414/148322377-1f97f484-0e66-4f98-a022-0eb4ef17f59b.png)

### Environment

- apisix version (cmd: `apisix version`): master branch
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

### Steps to reproduce

I suggest use "apisix-and-client" type.

### Actual result

nil

### Error log

nil

spacewander:
Yes, people often found that the APISIX latency is high caused by slow clients. But changing the type will be a break change.

What about adding a note in the doc: https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/prometheus.md?

zhendongcmss:
@spacewander Add notes on doc is easy. But it doesn't dispel confused. See the Grafana dashboard persion is different with seening the document.

tzssangglass:
I prefer break change to make the metrics clearer, but of course, we need the mailing list to discuss that.

xianshun163:
@zhendongcmss @tzssangglass 

Because the `apisix_latency` include read contents from client, not ONLY apisix process time.
do you  mean read contents from client cost a  lot of time?   Why does it take so much time in  read data.


> ### Issue description
> Because the `apisix_latency` include read contents from client, not ONLY apisix process time. It is very confused on the dashbord and got resault that apisix lead to high latency.
> 
[...]

PR #6063, opened 2022-01-10
feat: supports to logging apisix_latency and upstream_latency.

jagerzhang:
supports to logging apisix_latency and upstream_latency(copy from tapisix by [@kyroslin](https://github.com/kyroslin)): https://github.com/apache/apisix/issues/5910

Discuss: https://github.com/apache/apisix/discussions/6010

### Pre-submission checklist:
The original code is used directly without adding other new logic.


apisix/utils/log-util.lua
```diff
@@ -96,6 +96,8 @@ local function get_full_log(ngx, conf)
         }
     end
 
+    local latency, upstream_latency, apisix_latency = _M.latency_details_in_ms(ctx)
```
spacewander, line 99: Better to refactor the method and avoid calling the method like `= latency_details_in_ms(ctx)`

spacewander (commented):
Let's update test to make CI pass

apisix/utils/log-util.lua
```diff
@@ -120,7 +122,9 @@ local function get_full_log(ngx, conf)
         consumer = consumer,
         client_ip = core.request.get_remote_client_ip(ngx.ctx.api_ctx),
         start_time = ngx.req.start_time() * 1000,
-        latency = (ngx_now() - ngx.req.start_time()) * 1000
+        latency = latency,
+        upstream_latency = upstream_latency,
+        apisix_latency = apisix_latency
```
zhendongcmss, line 127: I think use `apisix_latency` is confused, refer another discussment https://github.com/apache/apisix/issues/6029.
The `apisix_latenct` includes client sending data via network, not apisix exclusive. From my experience, it is large when tranfer big file.

tzssangglass: Maybe in this PR do like 
`apisix_latency = apisix_latency - upstream_latency` to get the real apisix_latency, cc @spacewander

spacewander: The latency from the client exists through the whole request process.
For file upload, it affects the apisix latency much.
But for file download, it affects the upstream latency much.
Both latencies contain the latency from the client.
So it's OK to keep the current behavior.

apisix/utils/log-util.lua
jagerzhang: ok, already update.

spacewander (approved):

tzssangglass (approved):

PR #6070, opened 2022-01-11
chore(grpc-web): update mime of response to mime of request

shuaijinchao:

update mime of response to mime of request

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/grpc-web.lua
+++ b/apisix/plugins/grpc-web.lua
@@ -59,6 +59,9 @@ function _M.check_schema(conf)
 end
 
 function _M.access(conf, ctx)
+    -- set context variable mime
+    ctx.grpc_web_mime = core.request.header(ctx, "Content-Type")
+
     local method = core.request.get_method()
     if method == ALLOW_METHOD_OPTIONS then
         return 204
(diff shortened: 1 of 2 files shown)


apisix/plugins/grpc-web.lua
```diff
@@ -123,8 +122,14 @@ function _M.header_filter(conf, ctx)
...
     end
+
     core.response.set_header("Access-Control-Allow-Origin", DEFAULT_CORS_ALLOW_ORIGIN)
-    core.response.set_header("Content-Type", ctx.grpc_web_mime or DEFAULT_CORS_CONTENT_TYPE)
+
+    local response_mime = ctx.grpc_web_mime
+    if not response_mime then
+        response_mime = core.request.header(ctx, "Content-Type")
+    end
+    core.response.set_header("Content-Type", response_mime)
```
spacewander, line 132: We can set the type only when ctx.grpc_web_mime is given?

shuaijinchao: If it is judged that only gRPC-Web type requests can get MIME, I am not sure if this is friendly to other requests.

spacewander: I am confused. The current code just get and reset the Content-Type when the grpc_web_mime is missing.

shuaijinchao: I moved mime type extraction to the plugin's entry point to ensure it is always available in context.

apisix/plugins/grpc-web.lua
```diff
@@ -60,6 +59,9 @@ function _M.check_schema(conf)
 end
 
 function _M.access(conf, ctx)
+    -- set context variable mime
+    ctx.grpc_web_mime = core.request.header(ctx, "Content-Type")
```
tzssangglass, line 63: better to add a comment explaining why you need to set the ctx.grpc_web_mime in advance

shuaijinchao: added

tzssangglass (approved):

spacewander (approved):

Issue #6112, opened 2022-01-14
Proposal: Develop Loggly logging plugin


bisakhmondal:
## Initial phase:
reference : [here](https://www.loggly.com/)
A logging plugin that works and does its job : ) 
Supported Protocol: SYSLOG-UDP

### schema:
```lua
local schema = {
    type = "object",
    properties = {
        customer_token = {type = "string"},
        severity = {
            type = "string",
            default = "INFO",
            enum = {"DEBUG", "INFO", "NOTICE", "WARNING", "ERR", "CRIT", "ALERT", "EMEGR",
                    "debug", "info", "notice", "warning", "err", "crit", "alert", "emegr"},
            description = "base severity log level",
        },
        include_req_body = {type = "boolean", default = false},
        include_resp_body = {type = "boolean", default = false},
        include_resp_body_expr = {
            type = "array",
            minItems = 1,
            items = {
                type = "array"
            }
        },
        tags = {
            type = "array",
            minItems = 1,
            items = {
                type = "string"
            }
        },
        prefer_name = {type = "boolean", default = true}
    },
    required = {"customer_token"}
}

```

### Metadata schema
```lua
local defaults = {
    host = "logs-01.loggly.com",
    port = 514,
    protocol = "syslog",
    timeout = 5000
}

local metadata_schema = {
    type = "object",
    properties = {
        host = {
            type = "string",
            default = defaults.host
        },
        port = {
            type = "integer",
            default = defaults.port
        },
        protocol = {
            type = "string",
            default = defaults.protocol,
            -- more methods coming soon
            enum = {"syslog"}
        },
        timeout = {
            type = "integer",
            minimum = 1,
            default= defaults.timeout
        },
        log_format = {
            type = "object",
        }
    }
}

```

## Phase-2
- Supported Protocol: HTTP[S]
- Support Changing Log Severity level based on HTTP response code.

This issue is for the tracking purpose of the plugin development. If you have any useful insight, feel free to let us know : )

spacewander:
I think we can remove `include_req_body` & `tags`?
Instead, we can use `log_util.get_full_log` & `log_util.get_custom_format_log`. The `tags` field can be implemented via `get_custom_format_log`.

bisakhmondal:
> I think we can remove `include_req_body` & `tags`? Instead, we can use `log_util.get_full_log` & `log_util.get_custom_format_log`. The `tags` field can be implemented via `get_custom_format_log`.

[...]

PR #6113, opened 2022-01-14
feat: development of Loggly logging plugin

bisakhmondal:
Signed-off-by: Bisakh Mondal <bisakhmondal00@gmail.com>

#6112 
### Pre-submission checklist:


bisakhmondal:
The current logger implementation
image

apisix/plugins/slslog/rfc5424.lua
```diff
@@ -81,7 +81,10 @@ local Severity = {
 local log_util = require("apisix.utils.log-util")
 
 
-local _M = { version = 0.1 }
+local _M = {
+    version = 0.1,
+    severity = Severity
```
membphis, line 86: why we need to update this code?

membphis: one PR for one thing. so I think we need to revert this PR

bisakhmondal: Hi, thanks for the code review and the concern. The current PR need to access the rfc5424 severity map to decode alert status into integer numbers.
>    local message = {
        -- facility LOG_USER - random user level message
        "<".. tostring(8 + severity[conf.severity]) .. ">1", -- <PRIVAL>1
        timestamp,                                           -- timestamp
        ctx.var.host or "-",                                 -- hostname
        "apisix",                                            -- appname
        ctx.var.pid,                                         -- proc-id
        "-",                                                 -- msgid
        "[" .. conf.customer_token .. "@41058 " .. tab_concat(taglist, " ") .. "]",
        json_str
    }

I have made the field exported here from that module. Okay, reverting the changes back with a local copy of that severity map inside the plugin. Thank you

membphis (commented):
need to update the `README.md`

image

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+}
+
+local schema = {
+    type = "object",
+    properties = {
+        customer_token = {type = "string"},
+        severity = {
+            type = "string",
+            default = "INFO",
+            enum = {"DEBUG", "INFO", "NOTICE", "WARNING", "ERR", "CRIT", "ALERT", "EMEGR",
```
spacewander, line 48: We can get the list via `severity` table?

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+    if not json_str then
+        core.log.error('error occurred while encoding the data: ', err)
+        return nil
+    end
+
+    local timestamp = log_util.get_rfc3339_zulu_timestamp()
+    local taglist = {}
+    if conf.tags then
+        for i = 1, #conf.tags do
+            if not conf.tags[i]:sub(1, 4) ~= "tag=" then
```
spacewander, line 161: Better to check it with jsonschema?

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+    properties = {
+        customer_token = {type = "string"},
+        severity = {
+            type = "string",
+            default = "INFO",
+            enum = {"DEBUG", "INFO", "NOTICE", "WARNING", "ERR", "CRIT", "ALERT", "EMEGR",
+                    "debug", "info", "notice", "warning", "err", "crit", "alert", "emegr"},
+            description = "base severity log level",
+        },
+        include_req_body = {type = "boolean", default = false},
```
spacewander, line 52: Could we refactor it to avoid adding the same group of fields to each logger?

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+
+    if metadata and metadata.value.log_format
+       and core.table.nkeys(metadata.value.log_format) > 0
+    then
+        entry = log_util.get_custom_format_log(ctx, metadata.value.log_format)
+    else
+        entry = log_util.get_full_log(ngx, conf)
+    end
+
+    if conf.prefer_name then
```
spacewander, line 136: It seems we don't need this as we support customizing format?

docs/en/latest/plugins/loggly.md
```diff
@@ -0,0 +1,157 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Summary
+
+- [Summary](#summary)
```
spacewander, line 26: Extra entry?

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+    name = plugin_name,
+    schema = batch_processor_manager:wrap_schema(schema),
+    metadata_schema = metadata_schema
+}
+
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
```
spacewander, line 121: Need to call `log_util.check_log_schema(conf)`

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,243 @@
...
+end
+
+local function generate_log_message(conf, ctx)
+    local metadata = plugin.plugin_metadata(plugin_name)
+    local entry
+
+    if metadata and metadata.value.log_format
```
spacewander, line 128: Let's refactor it in the next PR

apisix/plugins/loggly.lua
spacewander: And you miss to call `log_util.collect_body(conf, ctx)`.

apisix/plugins/loggly.lua
bisakhmondal: Yep. Sounds good

apisix/plugins/loggly.lua
bisakhmondal: > Could we refactor it to avoid adding the same group of fields to each logger?
Yes, similar to what we did to batchprocessor. Let's handle that in the next PR. 

Good catch. Adding, Thanks

apisix/plugins/loggly.lua
bisakhmondal: Ack.

docs/en/latest/plugins/loggly.md
bisakhmondal: Ahh, auto-generated from my vscode markdown extension. Removing ASAP

apisix/plugins/loggly.lua
bisakhmondal: yes

bisakhmondal:
> need to update the `README.md`
> 
> image

Yes definitely. Do you want this to be handled in this PR or as usual a small PR with a readme update?
Thanks.

apisix/plugins/loggly.lua
bisakhmondal: Done

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,256 @@
...
+    end
+
+    return res, err_msg
+end
+
+
+local function handle_log(entries)
+    local ok, err
+    for i = 1, #entries do
+        ok, err = send_data_over_udp(entries[i])
```
membphis, line 236: do we need to check the return value?

spacewander:
> > need to update the `README.md`
> 
> Yes definitely. Do you want this to be handled in this PR or as usual a small PR with a readme update? Thanks.

I suggest doing it in the next PR.

apisix/plugins/loggly.lua
spacewander: Look like we need to remove the sent entries? Otherwise, they will be retried.

apisix/plugins/loggly.lua
```diff
@@ -0,0 +1,258 @@
...
+function _M.check_schema(conf, schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+        return core.schema.check(metadata_schema, conf)
+    end
+
+    local ok, err = core.schema.check(schema, conf)
+    if not ok then
+        return nil, err
+    end
+    return log_util.check_log_schema(conf)
```
shuaijinchao, line 140:
```suggestion
    if schema_type == core.schema.TYPE_METADATA then
        return core.schema.check(metadata_schema, conf)
    end
    return core.schema.check(schema, conf)
```

can be handled in this way.

bisakhmondal: Hi, thanks for the review. As the plugin schema contains, `include_resp_body_expr`, this need to be validated. core.schema.check just performs jsonschema validation based on the constraints set to the schema.  
https://github.com/apache/apisix/blob/a813b5aa8980dd8fa5780af351b6b47b5219d0b9/apisix/core/schema.lua#L31-L65

while the check_log_schema performs validation on the custom DSL
https://github.com/apache/apisix/blob/a813b5aa8980dd8fa5780af351b6b47b5219d0b9/apisix/utils/log-util.lua#L214-L228

In this case, we need to perform validation on log_schema. Thanks.

apisix/plugins/loggly.lua
bisakhmondal: Yes, that's an issue here. The way batchprocessor is written, it is meant for consuming each batch in one go (transactional). But in the case of loggly, only bulk events can be sent through HTTP endpoints. So I can think of three potential resolution

1. if err and i==1 -> then return false else true | if some entries has been sent that's ok. we are loosing the rest of the entries. [some entries will be dropped]
2. use the current behaviour. where some entries might be retried again [duplocation].
3. Update batch processor to have this fine-grained control
https://github.com/apache/apisix/blob/a813b5aa8980dd8fa5780af351b6b47b5219d0b9/apisix/utils/batch-processor.lua#L66-L89

update 
```lua
local ok, err = self.func(batch.entries, self.batch_max_size)

-- to this
local ok, err, n = self.func(batch.entries, self.batch_max_size)
 -- shorten self.entries based on n
```

which one do you think would be good? cc @spacewander

bisakhmondal: here n denotes the number of processed entries, and the rest of the entries should be retried.

apisix/plugins/loggly.lua
shuaijinchao: OK, understood

apisix/plugins/loggly.lua
bisakhmondal: cc @spacewander

t/plugin/loggly.t
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+--- wait: 0.5
+--- response_body
+passed
+opentracing
+--- grep_error_log eval
+qr/message received: .+?(?= \{)/
+--- grep_error_log_out eval
+qr/message received: <10>1 [\d\-T:.]+Z [\d.]+ apisix [\d]+ - \[token-1\@41058 ]/
```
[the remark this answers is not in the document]
bisakhmondal: Done.

t/plugin/loggly.t
```diff
@@ -0,0 +1,344 @@
...
+            end
+            ngx.print(body)
+        }
+    }
+--- response_body
+opentracing
+--- grep_error_log eval
+qr/message received: [ -~]+/
+--- grep_error_log_out eval
+qr/message received: <10>1 [\d\-T:.]+Z [\d.]+ apisix [\d]+ - \[token-1\@41058 ] \{"apisix_latency":[\d.]*,"client_ip":"127\.0\.0\.1","latency":[\d.]*,"request":\{"headers":\{"content-type":"application\/x-www-form-urlencoded","host":"127\.0\.0\.1:1984","user-agent":"lua-resty-http\/[\d.]* \(Lua\) ngx_lua\/[\d]*"\},"method":"GET","querystring":\{\},"size":[\d]+,"uri":"\/opentracing","url":"http:\/\/127\.0\.0\.1:1984\/opentracing"\},"response":\{"headers":\{"connection":"close","content-type":"text\/plain","server":"APISIX\/[\d.]+","transfer-encoding":"chunked"\},"size":[\d]*,"status":200\},"route_id":"1","server":\{"hostname":"[ -~]*","version":"[\d.]+"\},"service_id":"","start_time":[\d]*,"upstream":"127\.0\.0\.1:1982","upstream_latency":[\d]*\}/
```
spacewander, line 344: We need to configure `include_resp_body_expr` & `include_resp_body` to get the body.

bisakhmondal: Done. Also added case for metadata.log_format

spacewander (approved):

shuaijinchao (approved):

membphis (approved):

PR #6136, opened 2022-01-17
ci: support for generating flame graph

tzssangglass:

example: https://github.com/tzssangglass/apisix/actions/runs/1709028726

### Pre-submission checklist:


ci/performance_test.sh
```diff
@@ -0,0 +1,117 @@
...
+    ulimit -n 10240
+
+    pip3 install -r t/perf/requirements.txt --user
+
+    #openresty-debug
+    export OPENRESTY_PREFIX="/usr/local/openresty-debug"
+    export PATH=$OPENRESTY_PREFIX/nginx/sbin:$OPENRESTY_PREFIX/bin:$OPENRESTY_PREFIX/luajit/bin:$PATH
+
+    mkdir output
+    python3 ./t/perf/test_http.py >$PWD/output/performance.txt 2>&1 &
```
tokers, line 85: Does this script end up automatically?

tzssangglass: yes, `duration` is 100s

tzssangglass:
Since `GITHUB_TOKEN` has only read-only access, it is not possible to send the performance test results as comments under PR.

ref: https://github.com/actions/first-interaction/issues/10

tzssangglass:
Possible solution: https://github.com/apache/incubator-doris/pull/6279

tzssangglass:
> Possible solution: [apache/incubator-doris#6279](https://github.com/apache/incubator-doris/pull/6279)

It still does not solve the problem of `GITHUB_TOKEN` permissions.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: sudo ./ci/performance_test.sh install_dependencies
+
+      - name: Install wrk2
+        run: sudo ./ci/performance_test.sh install_wrk2
+
+      - name: Install SystemTap Tools
+        run: sudo ./ci/performance_test.sh install_stap_tools
+
+      - name: Perf Test
+        run: ./ci/performance_test.sh run_performance_test
```
shuaijinchao, line 40: Is it possible to use `sudo` here and then remove `sudo` from the script?

tzssangglass: Tried it before and it didn't fit

shuaijinchao: OK

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "t/grpc_server_example"]
 	path = t/grpc_server_example
 	url = https://github.com/api7/grpc_server_example
+[submodule ".github/actions/action-create-comment"]
+	path = .github/actions/action-create-comment
+	url = https://github.com/actions-ecosystem/action-create-comment
```
shuaijinchao, line 12: If no comment is currently supported, still need to import this module?

tzssangglass: rm it

shuaijinchao (dismissed):

.github/workflows/performance.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  pull_request:
+    branches: [master, 'release/**']
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+
+jobs:
+  performance:
+    if: github.event_name == 'pull_request' || github.event.label.name == 'performance'
```
spacewander, line 12: Should we only rely on the label? The pull request can be triggered by any user.

spacewander (commented):
Better to provide some hints about where to find the output

.github/workflows/performance.yml
spacewander: We can apply a label if needed. Note that people can run anything based on their PR.

tzssangglass: > We can apply a label if needed

Now this PR is satisfied, add `performance` and it will run. To run it again you need to remove `performance` and add it again.

> Note that people can run anything based on their PR.

Yes, the idealization is so. But this requires continued enhancements to the performance testing framework.
Maybe we can change it the next time we need to run performance tests for other PRs.

spacewander (approved):

tokers (approved):

shuaijinchao (approved):

PR #6140, opened 2022-01-18
feat:(plugins) make the ip-restriction plugin support mutil-level proxy agent

hiqrf:
now the ip-restriction can not work if we use mutil-level proxy agent like slb
I chage the ngx var remote_addr to http_x_forwarded_for 
to make the ip-restriction plugin support mutil-level proxy agent 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/ip-restriction/init.lua
+++ b/apisix/plugins/ip-restriction/init.lua
@@ -89,8 +89,13 @@ end
 
 function _M.restrict(conf, ctx)
     local block = false
-    local remote_addr = ctx.var.remote_addr
-
+    local remote_addr_arr = ctx.var.http_x_forwarded_for
+    local ipList = {}
+    local reps = ","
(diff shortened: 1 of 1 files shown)


apisix/plugins/ip-restriction/init.lua
```diff
@@ -89,8 +89,13 @@ end
 
 function _M.restrict(conf, ctx)
     local block = false
-    local remote_addr = ctx.var.remote_addr
-
+    local remote_addr_arr = ctx.var.http_x_forwarded_for
```
spacewander, line 92: You can solve it with a global real_ip configuration:
https://github.com/apache/apisix/blob/3773f2626e5753f95cc040c27b59fd0cff264513/conf/config-default.yaml#L226-L230

Or a route-level solution, the `real-ip` plugin: https://apisix.apache.org/docs/apisix/plugins/real-ip

hiqrf:
think you very much ,i have already solved this problem with they real ip plugin

Issue #6137, opened 2022-01-17
Proposal: support more access control for plugin's public API


bzp2010:
## Background
Current APISIX users can register public APIs when developing plugins, such as the `sign` API in the `jwt-auth` plugin and the `login` and other API in the `wolf-rbac` plugin. In the current design, users can control their access through [API interceptor](https://apisix.apache.org/docs/apisix/plugin-interceptors), which currently only support IP restrictions via the `ip-restriction` plugin and cannot apply more complex access patterns.
Therefore, we designed the `public-api` plugin, which incorporates the APIs from user plugins into the general HTTP request processing flow and supports the configuration of all plugins currently supported by APISIX.

### Configure schema
Name | Type | Requirement | Default | Description
-- | -- | -- | -- | --
uri | string | optional |   | The URI registered for this public API (used when the user customizes the URI of the plugin's public API, i.e. a different URI than the one defined for the public API is used in general HTTP routing)

Imagine a scenario where we have a `custom-auth` plugin that provides a login interface at `/apisix/custom-auth/login`, and we want to provide this API at `/login`, so we can create a route with `uri = /login` and configure it with the `public-api` plugin where the `uri` will be configured as `/apisix/custom-auth/login`, so that when the user accesses `/login`, the request will be forwarded to the headler in the plugin.

### Principle
Currently, APISIX generates a special radixtree to handle route matching for the public API registered in the user's plugin, which is independent of the general HTTP request processing flow and takes precedence over matching. So we can't apply all the plugins to it, we need to adapt those plugins one by one.
Therefore, we need to modify the flow so that the general HTTP request route is matched before the plugin‘s public API, and when a route is matched and it has the `public-api` plugin enabled, it will call the appropriate plugin public API headler based on the plugin configuration or the URI of the current request.
This allows us to apply any plugin to a plugin public API and also allows the user to decide whether to allow the client to access a public API and what URI to use.

## Other
What are your ideas?

encanto-oss:
This is a good improvement proposal and I want to give it a try.

juzhiyuan:
Mailing List: https://lists.apache.org/thread/g25fz8ov3kjd2bf3fgrvtdcwxr2rk9kl

spacewander:
We can consider no longer exposing the public API by default. People need to expose it with the public-api plugin.

Yes, I know it's a break change. But after this change, we can be safe by default.

bzp2010:
[...]

PR #6145, opened 2022-01-19
feat: add public api plugin

bzp2010:

Implement: #6137 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/public-api.lua
+++ b/apisix/plugins/public-api.lua
@@ -49,9 +49,9 @@ function _M.access(conf, ctx)
     -- perform route matching
     if router.api.match(ctx, skip) then
         return
-    else
-        return 404
     end
+
+    return 404
 end
 
 
--- a/apisix/plugins/public-api.lua
+++ b/apisix/plugins/public-api.lua
@@ -1,0 +1,58 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+local core   = require("apisix.core")
+local router = require("apisix.router")
+
+local schema = {
+    type = "object",
+    properties = {
+        uri = {type = "string"},
+    },
+}
+
+
+local _M = {
+    version = 0.1,
+    priority = 501,
+    name = "public-api",
+    schema = schema,
+}
+
+
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+
+function _M.access(conf, ctx)
+    local local_conf = core.config.local_conf()
+
+    -- overwrite the uri in the ctx when the user has set the target uri
+    ctx.var.uri = conf.uri and conf.uri or ctx.var.uri
+    local skip = local_conf and local_conf.apisix.global_rule_skip_internal_api
+
+    -- perform route matching
+    if router.api.match(ctx, skip) then
+        return
+    else
+        return 404
+    end
+end
+
+
+return _M
--- a/apisix/plugins/public-api.lua
+++ b/apisix/plugins/public-api.lua
@@ -43,7 +43,7 @@ function _M.access(conf, ctx)
     local local_conf = core.config.local_conf()
 
     -- overwrite the uri in the ctx when the user has set the target uri
-    ctx.var.uri = conf.uri and conf.uri or ctx.var.uri
+    ctx.var.uri = conf.uri or ctx.var.uri
     local skip = local_conf and local_conf.apisix.global_rule_skip_internal_api
 
     -- perform route matching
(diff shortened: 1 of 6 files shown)


apisix/plugins/public-api.lua
```diff
@@ -0,0 +1,58 @@
...
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+
+function _M.access(conf, ctx)
+    local local_conf = core.config.local_conf()
+
+    -- overwrite the uri in the ctx when the user has set the target uri
+    ctx.var.uri = conf.uri and conf.uri or ctx.var.uri
```
spacewander, line 46:
```suggestion
    ctx.var.uri = conf.uri or ctx.var.uri
```

apisix/plugins/public-api.lua
```diff
@@ -0,0 +1,58 @@
...
+    local skip = local_conf and local_conf.apisix.global_rule_skip_internal_api
```
spacewander, line 47: We don't need to run global rules twice?

t/plugin/public-api.t
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+--- response_body eval
+"201passed\n" x 3
+
+
+
+=== TEST 3: hit route (custom-jwt-sign)
+--- request
+GET /gen_token?key=user-key
```
spacewander, line 123: Let's add assertion to make sure the jwt sign is working

t/plugin/public-api.t
```diff
@@ -0,0 +1,134 @@
...
+direct-wolf-rbac-userinfo was triggered
+
+
+
+=== TEST 4: hit route (direct-wolf-rbac-userinfo)
+--- request
+GET /apisix/plugin/wolf-rbac/user_info
+--- error_code: 401
+--- error_log
+custom-jwt-sign was triggered
```
spacewander, line 134: Let's add more cases:
1. public api not match
2. match without optional uri
3. add auth to protect public api

apisix/plugins/public-api.lua
bzp2010: fixed

apisix/plugins/public-api.lua
bzp2010: https://github.com/apache/apisix/blob/64c47df50c5607e9bf9921021391bb32c5f4d0b2/apisix/init.lua#L376-L445 

Hi, @spacewander.

Based on this part of the code, we can presume to have such a situation.

1. HTTP match, and `public-api` plugin is enabled.
This will be controlled by the configuration file whether to enforce global rules for public API calls. After the match, the subsequent request processing flow will be completely taken over by the public API handler, bypassing the rest of the code execution in the access phase.
2. HTTP not matched, public API route matched.
This will also be controlled by the configuration file whether to enforce global rules or not. 
3. Both HTTP and public APIs do not match
It will force the global rule to run and return 404 Not Route.

----------

Also, as you said in this issue https://github.com/apache/apisix/issues/6137#issuecomment-1016016565, we will adjust the public API to not be exposed by default, which will be done by completely removing the public API route match in the access phase. (Currently HTTP matching is performed followed by public API matching as a backup) In that case, the matching and processing of the public API will be done by the public-api plugin only, and if a match is made the global rules will be enforced according to the configuration.

t/plugin/public-api.t
bzp2010: added, It will check the first part of the JWT in the response body, however, because json encode cannot guarantee the order of the data in it, it matches both cases.

t/plugin/public-api.t
bzp2010: 1. added
2. I didn't understand what you meant, is it a case where there is no uri configuration in the schema? If so it is there now.
3. added, protect jwt sign by `key-auth`

apisix/plugins/public-api.lua
spacewander: > In that case, the matching and processing of the public API will be done by the public-api plugin only, and if a match is made the global rules will be enforced according to the configuration.

We can adjust it when we retire the router? The current implementation will run global rules twice in some cases.

t/plugin/public-api.t
```diff
@@ -0,0 +1,219 @@
...
+--- error_code: 401
+--- error_log
+direct-wolf-rbac-userinfo was triggered
+
+
+
+=== TEST 5: missing route
+--- request
+GET /apisix/plugin/balalbala
+--- error_code: 404
```
spacewander, line 160: Let's add more test that the uri in public API plugin matches nothing.

bzp2010: got it, adding

bzp2010: added

shuaijinchao:
I don't see the problem in the code and test cases, but I don't see the documentation for the plugin.

apisix/plugins/public-api.lua
bzp2010: I will try to fix it.

bzp2010:
> I don't see the problem in the code and test cases, but I don't see the documentation for the plugin.

@shuaijinchao 

According this comment https://github.com/apache/apisix/pull/6145#discussion_r789309852, there is a next phase of development work after this PR is completed, which will be added when everything is done.

spacewander (approved):
Approve except https://github.com/apache/apisix/pull/6145#discussion_r789717240

apisix/plugins/public-api.lua
bzp2010: ## Update
After checking again, I think there may not be a duplicate calls problem.

## Details
image

There are **4** calls to `run_global_rules` executed in the current code. One of them is in `api_router`, which is controlled by the configuration file (this is the router used by the public API), and the remaining ones are executed in the `http_access_phase` of `init`.

1. both HTTP and public API do not match, will run and return a 404 status code
2. run after HTTP matched
3. run in `common_phase` for `before_proxy`

In fact I didn't add `global_rules` to init.lua, they are still being executed according to the original logic.

If I have missed anything, please help point it out. Thanks.

shuaijinchao (approved):

Issue #6066, opened 2022-01-10
bug: request-id plugin overwrite X-Request-Id already present in the request


shelltea:
### Issue description

If `X-Request-Id` is already present in the request,the response didn't return the same value。

![image](https://user-images.githubusercontent.com/864375/148755044-6d641557-9a5d-4c6b-ad17-ae569978d487.png)

### Environment

- apisix version (cmd: `apisix version`): 2.10.2
- OS (cmd: `uname -a`): Linux apisix-85d66cc8b8-s7lb5 4.15.0-136-generic #140-Ubuntu SMP Thu Jan 28 05:20:47 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):nginx version: openresty/1.19.3.2
built by gcc 9.3.1 20200408 (Red Hat 9.3.1-2) (GCC)
built with OpenSSL 1.1.1l  24 Aug 2021
TLS SNI support enabled
[...]

PR #6160, opened 2022-01-20
fix(request-id): if the header_name is already present in the…

yujintang:
if the header_name is already present in the request uuid_val include_in_response should keep source

in request-id plugin, when header_name is already present in the request, we should keep source id 

Fix https://github.com/apache/apisix/issues/6066

### Pre-submission checklist:
no

### DIFF
--- a/apisix/plugins/request-id.lua
+++ b/apisix/plugins/request-id.lua
@@ -211,8 +211,9 @@ end
 
 function _M.rewrite(conf, ctx)
     local headers = ngx.req.get_headers()
-    local uuid_val = get_request_id(conf.algorithm)
+    local uuid_val
     if not headers[conf.header_name] then
+        uuid_val = get_request_id(conf.algorithm)
         core.request.set_header(ctx, conf.header_name, uuid_val)
     else
         uuid_val = headers[conf.header_name]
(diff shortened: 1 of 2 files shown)


spacewander (commented):
Could you add a test in https://github.com/apache/apisix/blob/master/t/plugin/request-id.t?
We can use https://github.com/apache/apisix/blob/e9fab8b2e5b5cb79a302faf8b5c9752766681c63/t/plugin/grpc-web.t#L164-L165 to send request header

tzssangglass:
LGTM

apisix/plugins/request-id.lua
```diff
@@ -214,6 +214,8 @@ function _M.rewrite(conf, ctx)
     local uuid_val = get_request_id(conf.algorithm)
     if not headers[conf.header_name] then
         core.request.set_header(ctx, conf.header_name, uuid_val)
+    else
+        uuid_val = headers[conf.header_name]
```
starsz, line 218: I think we can avoid calling get_request_id if the id exists.

spacewander (approved):

starsz (approved):

tzssangglass (approved):

tokers (approved):

apisix/plugins/request-id.lua
haoyang1994: This change should be submitted with an switch. In my situation, i can't trust the request id I received and must rewrite it with a new uuid.

PR #6199, opened 2022-01-25
fix(proto): avoid sharing state

spacewander:

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/grpc-transcode/proto.lua
+++ b/apisix/plugins/grpc-transcode/proto.lua
@@ -16,6 +16,7 @@
 --
 local core        = require("apisix.core")
 local config_util = require("apisix.core.config_util")
+local pb          = require("pb")
 local protoc      = require("protoc")
 local pcall       = pcall
 local ipairs      = ipairs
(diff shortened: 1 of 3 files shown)


apisix/plugins/grpc-transcode/proto.lua
```diff
@@ -29,6 +30,10 @@ local lrucache_proto = core.lrucache.new({
 local proto_fake_file = "filename for loaded"
 
 local function compile_proto(content)
+    -- clear pb state
+    pb.state(nil)
```
bisakhmondal, line 34: Isn't it redundant? I think Line 54 is already retrieving the current pb state and setting it to nil.

bisakhmondal (approved):
Great fix indeed :+1:

tzssangglass (approved):

leslie-tsang (approved):

apisix/plugins/grpc-transcode/proto.lua
spacewander: There is a default pb state that we need to clear.

bisakhmondal: Got it.

PR #6202, opened 2022-01-25
change(server-info): use a new approach(keepalive) to report DP info

### LINKED ISSUE #6176: Proposal: refactor server-info plugin
## Background
The current server-info plug-in adds a revision every time the status is updated, and some users set a relatively short reporting time, which may result in millions or even tens of millions of data in etcd after a long run, resulting in a slow connection and high cpu usage.

## Purpose
Modify the current server-info plugin so that it does not write a piece of data to etcd every time the system state is refreshed. When the apisix node is alive, the server-info information will always be there and will not be updated with revisions; only when the apisix node system information changes will there be a put operation. (e.g., hostname changes)

### Configura Schema
| Name    | Type | default |Description |
|:---------:|:------:|:-------:|:-------------:|
| report_ttl | integer | 36 |the live time for server info in etcd (unit: second, maximum: 86400, minimum: 3) |

## implement

Etcd has a keepalive command that can refresh the TTL of the key without generating new revisions. We can use keepalive to refresh the node information to avoid writing a lot of useless information.
1. get the server-info data, if it does not exist, initialize the server-info information
2. bind the lease to the key "server-info"
3. call keepalive to refresh the TTL to ensure the node is alive

## Server-info Flowchart
image

## Other

What are your ideas?

zaunist:

fix #6176 

### Pre-submission checklist:


zaunist:
> @zaunist any update？

I have update the plugin code, still need to update test case.

zaunist:
cc @tzssangglass @shuaijinchao. Please take a look, thanks.

zaunist:
@bisakhmondal @bzp2010

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,66 @@ local function report(premature, report_ttl)
...
+        local newres, err = core.etcd.set(key, server_info, report_ttl)
+        if not newres then
+            core.log.error("failed to set server_info: ", err)
+        end
+
+        -- set lease_id to ngx dict
+        local ok, err = internal_status:set("lease_id", newres.body.lease_id)
+        if not ok then
+            core.log.error("failed to save boot_time to shdict: ", err)
+        end
```
shuaijinchao, line 166: should not continue to set `ngx dict` if `newres` fetch fails

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,66 @@ local function report(premature, report_ttl)
         end
     end
 
-    server_info.last_report_time = ngx_time()
-
+    -- get inside etcd data, if not exist, create it
     local key = "/data_plane/server_info/" .. server_info.id
-    local ok, err = core.etcd.set(key, server_info, report_ttl)
+    local res, _ = core.etcd.get(key)
```
shuaijinchao, line 155: need to judge failure

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,66 @@ local function report(premature, report_ttl)
...
+        -- set lease_id to ngx dict
+        local ok, err = internal_status:set("lease_id", newres.body.lease_id)
+        if not ok then
+            core.log.error("failed to save boot_time to shdict: ", err)
+        end
+
+        return
+    end
+
+    local res, _ = core.etcd.get(key)
```
shuaijinchao, line 171: ditto

apisix/plugins/server-info.lua
```diff
@@ -203,8 +236,7 @@ end
 
 
 function _M.init()
-    core.log.info("server info: ", core.json.delay_encode(get()))
-
+    -- core.log.info("server info: ", core.json.delay_encode(get()))
```
shuaijinchao, line 239: remove debug and redundant code

apisix/plugins/server-info.lua
```diff
@@ -217,13 +249,14 @@ function _M.init()
         return
     end
 
+    -- local report_ttl = attr and attr.report_ttl or default_report_ttl
```
shuaijinchao, line 252: ditto

shuaijinchao (approved):

shuaijinchao:
misoperation `approved`, please correct the above problems first.

apisix/plugins/server-info.lua
```diff
@@ -87,36 +80,35 @@ local function get_boot_time()
 end
 
 
-local function uninitialized_server_info()
+local function uninitialize_server_info()
```
tokers, line 83: Rename this function is meaningless to me.

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,70 @@ local function report(premature, report_ttl)
...
+        local newres, err = core.etcd.set(key, server_info, report_ttl)
+        if not newres then
+            core.log.error("failed to set server_info: ", err)
+            return
+        end
+
+        -- set lease_id to ngx dict
+        local _, err = internal_status:set("lease_id", newres.body.lease_id)
+        if err ~= nil then
+            core.log.error("failed to save boot_time to shdict: ", err)
```
tokers, line 170: And the error log is not accurate.

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,70 @@ local function report(premature, report_ttl)
...
+        if not newres then
+            core.log.error("failed to set server_info to etcd: ", err)
+            return false, err
+        end
+
+        -- update lease_id
+        local _, err = internal_status:set("lease_id", res.body.lease_id)
+        if err ~= nil then
+            core.log.error("failed to set lease_id to shdict: ", err)
+        end
```
tokers, line 189: This paragraph is same to the one in the `if  not res.body.node then .... end` block, we can create a function to reuse it.

apisix/plugins/server-info.lua
```diff
@@ -87,36 +80,35 @@ local function get_boot_time()
...
+local function uninitialize_server_info()
     local boot_time = get_boot_time()
     return {
         etcd_version     = "unknown",
         hostname         = core.utils.gethostname(),
         id               = core.id.get(),
         version          = core.version.VERSION,
-        up_time          = ngx_time() - boot_time,
         boot_time        = boot_time,
-        last_report_time = -1,
```
tokers, line 99: Note we should also update the manager api or the server-info report cannot be shown on the dashboard.

apisix/plugins/server-info.lua
```diff
@@ -158,25 +150,70 @@ local function report(premature, report_ttl)
...
+    -- get inside etcd data, if not exist, create it
     local key = "/data_plane/server_info/" .. server_info.id
-    local ok, err = core.etcd.set(key, server_info, report_ttl)
+    local res, err = core.etcd.get(key)
+    if err ~= nil then
+        core.log.error("failed to get server_info from etcd: ", err)
+    end
+
+    if  not res.body.node then
+        local newres, err = core.etcd.set(key, server_info, report_ttl)
```
tokers, line 161: Also the `key` for individual instances should be unique but should we use the `atomic_set` method?

apisix/plugins/server-info.lua
zaunist: Got it.

apisix/plugins/server-info.lua
zaunist: Got it.

apisix/plugins/server-info.lua
zaunist: Thanks for your carefulness.

apisix/plugins/server-info.lua
zaunist: Yes, I will discard this modification.

apisix/plugins/server-info.lua
```diff
@@ -132,15 +124,24 @@ local function get_server_info()
...
 local function report(premature, report_ttl)
     if premature then
         return
     end
 
+    -- get apisix node info
     local server_info, err = get()
     if not server_info then
         core.log.error("failed to get server_info: ", err)
-        return
```
tokers, line 143: Why removing this?

apisix/plugins/server-info.lua
```diff
@@ -158,25 +159,65 @@ local function report(premature, report_ttl)
...
-
+    -- get inside etcd data, if not exist, create it
     local key = "/data_plane/server_info/" .. server_info.id
-    local ok, err = core.etcd.set(key, server_info, report_ttl)
+    local res, err = core.etcd.get(key)
+    if err ~= nil then
+        core.log.error("failed to get server_info from etcd: ", err)
+    end
+
+    if  not res.body.node then
```
tokers, line 169:
```suggestion
    if not res.body.node then
```

apisix/plugins/server-info.lua
```diff
@@ -158,25 +159,65 @@ local function report(premature, report_ttl)
...
     end
 
-    server_info.last_report_time = ngx_time()
-
+    -- get inside etcd data, if not exist, create it
     local key = "/data_plane/server_info/" .. server_info.id
-    local ok, err = core.etcd.set(key, server_info, report_ttl)
+    local res, err = core.etcd.get(key)
+    if err ~= nil then
+        core.log.error("failed to get server_info from etcd: ", err)
```
tokers, line 166: if `err` is not `nil`, why can we go ahead?

apisix/plugins/server-info.lua
```diff
@@ -158,25 +159,65 @@ local function report(premature, report_ttl)
...
+    end
+
+    if  not res.body.node then
+        local res_new, err = core.etcd.set(key, server_info, report_ttl)
```
tokers, line 170: I think the logic for initial setting can be merged to the below logic (`modified_version` is zero).

tokers: So we don't have two similar code blocks.

apisix/plugins/server-info.lua
zaunist: When a apisix node be  initialed, the ETCD don't have server-info data, I think the `get` func will report error, so I continue the code and set the info to ETCD. Maybe use `if not res` be better.

apisix/plugins/server-info.lua
zaunist: Can we set `modified_version ` to `nil`, I'm not sure if it can be set to nil.

apisix/plugins/server-info.lua
zaunist: Sorry, careless.

apisix/plugins/server-info.lua
tokers: No, if `get` method reports error when no data, you should check the error type and decide whether returning or going ahead.

apisix/plugins/server-info.lua
tokers: I have already said you can set it to `0`.

zaunist: updated it.

tzssangglass:
origin issue has ` only when the apisix node system information changes will there be a put operation. (e.g., hostname changes)`

we need to add a test case to cover this?

zaunist:
Hi, Community. 

I have encountered follow problems.

- When the server-info plugin is turned on by default, other plugins report errors, and I'm not sure if the errors reported here are the expected behaviour. https://github.com/apache/apisix/runs/5174938000?check_suite_focus=true#step:10:354
- When the server-info plugin is turned off by default, errors are also reported in other checks, here the errors are due to a test of the configuration file. https://github.com/apache/apisix/runs/5177477142?check_suite_focus=true#step:10:355

Should I choose to turn on the server-info plugin by default or turn it off by default? If it is turned on by default, I don't know how to deal with the error here, can anyone give me some help?

apisix/plugins/server-info.lua
```diff
@@ -132,51 +125,107 @@ local function get_server_info()
 end
 
 
+local function set(key,value,ttl, modified_index)
```
tzssangglass, line 128:
```suggestion
local function set(key, value, ttl, modified_index)
```

zaunist:
> origin issue has ` only when the apisix node system information changes will there be a put operation. (e.g., hostname changes)`
> 
> we need to add a test case to cover this?

How do I change the `hostname` in the .t test file? I'm not quite sure how to do it.

apisix/plugins/server-info.lua
```diff
@@ -132,51 +125,107 @@ local function get_server_info()
...
+    local res_new, err = core.etcd.atomic_set(key, value, ttl, modified_index)
+    if not res_new then
+        core.log.error("failed to set server_info: ", err)
+        return 503, {error_msg = err}
+    end
+
+    lease_id = res_new.body.lease_id
```
tzssangglass, line 135: should we check if `lease_id` is not nil?

apisix/plugins/server-info.lua
```diff
@@ -132,51 +125,107 @@ local function get_server_info()
...
+
+    -- set or update lease_id
+    local ok, err = internal_status:set("lease_id", lease_id)
+    if not ok then
+        core.log.error("failed to set lease_id to shdict: ", err)
+        return 503, {error_msg = err}
+    end
+
+    return 200, nil
+end
```
tzssangglass, line 145: It looks like the two steps should be in sync, and we can use lock in the next PR to keep them in sync.

tzssangglass:
> How do I change the hostname in the .t test file? I'm not quite sure how to do it.

It doesn't have to be hostname, it can be any other action that can be used to punish service info refresh.

zaunist:
> 

Only a change in `hostname`、`etcd_version`、`apisix_version`、`boot_time` will result in a put operation, and in the current case there doesn't seem to be a way to make any of them change.

zaunist:
> > How do I change the hostname in the .t test file? I'm not quite sure how to do it.
> 
> It doesn't have to be hostname, it can be any other action that can be used to punish service info refresh.

Only a change in `hostname`、`etcd_version`、`apisix_version`、`boot_time` will result in a put operation, and in the current case there doesn't seem to be a way to make any of them change.

tzssangglass:
> Only a change in `hostname`、`etcd_version`、`apisix_version`、`boot_time`、`id` will result in a put operation, and in the current case there doesn't seem to be a way to make any of them change.

since server-info is stored in `ngx.shared["internal-status"]`, can we update this?

.github/workflows/centos7-ci.yml
```diff
@@ -80,3 +80,8 @@ jobs:
     - name: Run test cases
       run: |
         docker exec centos7Instance bash -c "cd apisix && ./ci/centos7-ci.sh run_case"
+
+    - name: Set up tmate session
+      if: ${{ failure() }}
+      uses: ./.github/actions/action-tmate
+      timeout-minutes: 15
```
tzssangglass, line 87: missing blank line at the end

tokers:
> Hi, Community.
> 
> I have encountered follow problems.
> 
> * When the server-info plugin is turned on by default, other plugins report errors, and I'm not sure if the errors reported here are the expected behaviour. https://github.com/apache/apisix/runs/5174938000?check_suite_focus=true#step:10:354
> * When the server-info plugin is turned off by default, errors are also reported in other checks, here the errors are due to a test of the configuration file. https://github.com/apache/apisix/runs/5177477142?check_suite_focus=true#step:10:355
> 
> Should I choose to turn on the server-info plugin by default or turn it off by default? If it is turned on by default, I don't know how to deal with the error here, can anyone give me some help?

For the 1st question:

```
Error: led test 'TEST 7: hit - pattern "[error]" should not match any line in error.log but matches line "2022/02/13 17:07:31 [error] 14195#14195: *37 [lua] server-info.lua:131: set(): failed to set server_info: value changed before overwritten, context: ngx.timer" (req 0)
```

It's still related to the server-info plugin.

For the second one, this error is normal as you disable the server-info plugin, it shouldn't exist in the list.

conf/config-default.yaml
```diff
@@ -359,7 +359,7 @@ plugins:                          # plugin list (sorted by priority)
   - limit-req                      # priority: 1001
   #- node-status                   # priority: 1000
   - gzip                           # priority: 995
-  - server-info                    # priority: 990
+  # - server-info                    # priority: 990
```
tokers, line 362: Is this the result of discussion?

zaunist: No, I can't fix the test case failed, and I am looking for help.

zaunist:
In the end, I decided to use etcd.set instead of etcd.atomic_set for the following reasons.

1. Atomic_set performs transactions, and in the test case, calling Apache APISIX reload interface would result in two transactions being performed in a short period of time. Apache APISIX would start once and reload would be called again, which would make the rest of the test case unstable. The error message just like it: https://github.com/apache/apisix/runs/5179696756?check_suite_focus=true#step:10:657

    - The first init in https://github.com/apache/apisix/blob/f37fc0eb77abbf7fdf98ea40ce706d75e5582a4b/apisix/plugins/server-info.lua#L205
    - The second init in https://github.com/apache/apisix/blob/f37fc0eb77abbf7fdf98ea40ce706d75e5582a4b/apisix/plugin.lua#L152

2. Using etcd.set does the same thing as etcd.atomic_set, and there is no strong need to use etcd.atomic_set.

For server-info test-case:

I tried to change the hostname using exec, but `sudo hostname` doesn't seem to work in the test framework. https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md#executing-shell-commands

Does anyone have any suggestions?

tzssangglass:
> I tried to change the hostname using exec, but `sudo hostname` doesn't seem to work in the test framework.

may shell commands in test nginx doesn't have permission to execute `sudo`? 

can you try to modify the `conf/apisix.uid`?

zaunist:
> > I tried to change the hostname using exec, but `sudo hostname` doesn't seem to work in the test framework.
> 
> may shell commands in test nginx doesn't have permission to execute `sudo`?
> 
> can you try to modify the `conf/apisix.uid`?

Got it, I will to try.

zaunist:
Hi, @tzssangglass . I checked `core.id.get()` and `core.utils.gethostname()`. It doesn't support updating ids and hostnames, it only gets them once when apisix starts, so it's not possible to update them in real time in the server-info plugin either. I'm sorry that this doesn't do what the proposal is trying to do. Maybe we can enhance `core.id.get` and `core.utils.hostname` to support regularly update id and hostname in a subsequent PR.

nic-chen (commented):
What we need to test is keepalive instead of getting server info, so we could modify the info stored in shared DICT to test.

apisix/plugins/server-info.lua
```diff
@@ -132,51 +124,111 @@ local function get_server_info()
 end
 
 
+local function set(key, value, ttl)
+    local res_new, err = core.etcd.set(key, value, ttl)
+    if not res_new then
+        core.log.error("failed to set server_info: ", err)
+        return 503, {error_msg = err}
```
nic-chen, line 131: why should we use `503` here?

zaunist:
> What we need to test is keepalive instead of getting server info, so we could modify the info stored in shared DICT to test.

I have tried to change the server-info information in the share dict in the t-test file, but it did not work.

apisix/plugins/server-info.lua
zaunist: 503 Service Unavailable, I think it is OK to return this error when the plugin cannot provide the service, what do you suggestion?

nic-chen: I think `500` is better. `503` is commonly used for overload.

nic-chen:
> > What we need to test is keepalive instead of getting server info, so we could modify the info stored in shared DICT to test.
> 
> I have tried to change the server-info information in the share dict in the `.t` test file, but it did not work.

you could push the test case again to see why it failed.

tzssangglass:
> Hi, @tzssangglass . I checked `core.id.get()` and `core.utils.gethostname()`. It doesn't support updating ids and hostnames, it only gets them once when apisix starts, so it's not possible to update them in real time in the server-info plugin either. I'm sorry that this doesn't do what the proposal is trying to do. Maybe we can enhance `core.id.get` and `core.utils.hostname` to support regularly update id and hostname in a subsequent PR.

My mistake, I should have pointed out that the test cases are written in shell script. Let's drop this test case for now and focus on finishing this PR.

zaunist:
> > Hi, @tzssangglass . I checked `core.id.get()` and `core.utils.gethostname()`. It doesn't support updating ids and hostnames, it only gets them once when apisix starts, so it's not possible to update them in real time in the server-info plugin either. I'm sorry that this doesn't do what the proposal is trying to do. Maybe we can enhance `core.id.get` and `core.utils.hostname` to support regularly update id and hostname in a subsequent PR.
> 
> My mistake, I should have pointed out that the test cases are written in shell script. Let's drop this test case for now and focus on finishing this PR.

Yes, if we need to get the latest `id` and `hostname` etc., then we also need to update the `core.id.lua` and the code in `core.utils.lua` to support getting the latest status on a regular basis. My question is, do we have to do these changes in this PR?

tzssangglass:
> My question is, do we have to do these changes in this PR?

No need to change in this PR.

tzssangglass (dismissed):

apisix/plugins/server-info.lua
```diff
@@ -132,51 +124,111 @@ local function get_server_info()
...
-        return
+        return 500, {error_msg = err}
     end
 
     if server_info.etcd_version == "unknown" then
         local res, err = core.etcd.server_version()
         if not res then
             core.log.error("failed to fetch etcd version: ", err)
-            return
+            return 500, {error_msg = err}
```
spacewander, line 168: Please use 503 like other similar cases.

apisix/plugins/server-info.lua
```diff
@@ -132,51 +124,111 @@ local function get_server_info()
...
+        end
+
         return
     end
 
+    -- get lease_id from ngx dict
+    lease_id, err = internal_status:get("lease_id")
+    if not lease_id then
+        core.log.error("failed to get lease_id from shdict: ", err)
+    end
```
spacewander, line 213: Missing a `return`?

apisix/plugins/server-info.lua
```diff
@@ -132,51 +124,111 @@ local function get_server_info()
...
+    if not res then
+        core.log.error("send heartbeat failed: ", err)
+        return 503, {error_msg = err}
+    end
+
     local data, err = core.json.encode(server_info)
     if not data then
         core.log.error("failed to encode server_info: ", err)
-        return
+        return false, err
```
spacewander, line 225: The returned value is inconsistent?

And it seems the returned value of `report` is not checked:
https://github.com/apache/apisix/blob/d5166c4c794ffe594c4bc70811394711d59a5d9a/apisix/plugins/server-info.lua#L278

spacewander:
@zaunist 
I am curious why a shorter ttl is chosen?

zaunist:
> @zaunist I am curious why a shorter ttl is chosen?

Hi, @spacewander. Previously, server-info would write a lot of dirty data and cause performance problems with ETCD, but with keepalive, this problem is avoided and a simple http request does not take up any more performance. So a shorter ttl would allow for more timely reporting of Apache APISIX node status. One minute heartbeat is too long in my opinion, so it was changed to 36 seconds, or we could even set a shorter heartbeat, such as nacos, which has a default of 5 seconds.

https://github.com/alibaba/nacos/blob/47d3afd0e4de5019c09a921fff08ae66aed52600/api/src/main/java/com/alibaba/nacos/api/common/Constants.java#L173

zaunist:
> Yes. But keepalive request is still a write operation. Since this PR is to reduce the usage of etcd, using a lower interval will increase the number of requests to etcd.
> 
> > One minute heartbeat is too long in my opinion, so it was changed to 36 seconds, or we could even set a shorter heartbeat, such as nacos, which has a default of 5 seconds
> 
> Nacos doesn't write the heartbeat to etcd... Are there any issues relative to the heartbeat interval? We don't want to get in trouble because of an unnecessary change.

OK, I will change it to 60s.

apisix/plugins/server-info.lua
zaunist: Already add return code block

apisix/plugins/server-info.lua
zaunist: Got it.

apisix/plugins/server-info.lua
```diff
@@ -125,58 +117,119 @@ local function get_server_info()
...
     local key = "/data_plane/server_info/" .. server_info.id
-    local ok, err = core.etcd.set(key, server_info, report_ttl)
+    local res, err = core.etcd.get(key)
+    if not res or (res.status ~= 200 and res.status ~= 404) then
+        core.log.error("failed to get server_info from etcd: ", err)
+        return 503, {error_msg = err}
+    end
+
+    if not res.body.node then
+        local ok, err = set(key, server_info, report_ttl)
```
spacewander, line 188: It's strange that get returns `nil, err` while set returns `status, table`...
Can we use `nil, err` in both cases?

apisix/plugins/server-info.lua
```diff
@@ -125,58 +117,119 @@ local function get_server_info()
...
+        if not ok then
+            core.log.error("failed to set server_info to etcd: ", err)
+            return 503, {error_msg = err}
```
spacewander, line 191: Now we get a nested err table

apisix/plugins/server-info.lua
```diff
@@ -125,58 +117,119 @@ local function get_server_info()
...
+    end
+
+    local ok = core.table.deep_eq(server_info, res.body.node.value)
+    -- not equal, update it
     if not ok then
-        core.log.error("failed to report server info to etcd: ", err)
+        local ok, err = set(key, server_info, report_ttl)
+        if not ok then
+            core.log.error("failed to set server_info to etcd: ", err)
+            return 503, {error_msg = err}
```
spacewander, line 203: I think we should return directly as we don't (and can't) care about the result of `report`.

zaunist: OK, I will fix it.

apisix/plugins/server-info.lua
zaunist: Got it.

apisix/plugins/server-info.lua
zaunist: Do I just return it here without the err? That means if there is an error, it will just be printed to error.log

spacewander: see https://github.com/apache/apisix/pull/6202#discussion_r812618876

zaunist: Thanks for your reply.

shuaijinchao (approved):

PR #6203, opened 2022-01-25
feat(batchprocessor): support partial consumption of batch entries

bisakhmondal:
Signed-off-by: Bisakh Mondal <bisakhmondal00@gmail.com>

Helpful where the batch processor entries can't be processed in one go (transactional) and an error occurs during the middle of the consumption. With this PR, the batch processor takes a hint from the consumer so that the already consumed entries don't get retried in the next run.

Use case: https://github.com/apache/apisix/pull/6113#discussion_r790424879, datadog plugin etc (where entries can't be processed as a bulk)

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -23,7 +23,6 @@ local ngx = ngx
 local udp = ngx.socket.udp
 local format = string.format
 local concat = table.concat
-local ipairs = ipairs
 local tostring = tostring
 
 local plugin_name = "datadog"
@@ -64,12 +63,14 @@ local _M = {
     metadata_schema = metadata_schema,
 }
 
+
 function _M.check_schema(conf, schema_type)
     if schema_type == core.schema.TYPE_METADATA then
         return core.schema.check(metadata_schema, conf)
     end
     return core.schema.check(schema, conf)
 end
+
 
 local function generate_tag(entry, const_tags)
     local tags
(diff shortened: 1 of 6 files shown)


apisix/utils/batch-processor.lua
```diff
@@ -63,17 +63,36 @@ local function set_metrics(self, count)
...
+end
+
+
 function execute_func(premature, self, batch)
     if premature then
         return
     end
 
-    local ok, err = self.func(batch.entries, self.batch_max_size)
+    --- In case of "err" and a valid "n" batch processor considers, all n-1 entries have been
```
spacewander, line 80: Would be `first_fail` better than `n` as the var name?

apisix/utils/batch-processor.lua
```diff
@@ -63,17 +63,36 @@ local function set_metrics(self, count)
...
+    --- successfully consumed and hence reschedule the job for entries with index n to #entries
+    --- based on the current retry policy.
+    local ok, err, n = self.func(batch.entries, self.batch_max_size)
     if not ok then
-        core.log.error("Batch Processor[", self.name,
-                       "] failed to process entries: ", err)
+        if n then
+            core.log.error("Batch Processor[", self.name, "] failed to process entries [",
+                            #batch.entries + 1 - (n or -1), "/", #batch.entries ,"]: ", err)
```
spacewander, line 87: The `n` can't be `nil`, so `n or -1` is unnecessary.

spacewander:
Don't forget to mention it in the doc.

apisix/utils/batch-processor.lua
```diff
@@ -63,17 +63,36 @@ local function set_metrics(self, count)
 end
 
 
+local function slice_batch(batch, n)
+    local slice = {}
+    for i = n or 1, #batch, 1 do
+        slice[#slice+1] = batch[i]
```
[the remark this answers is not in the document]
bisakhmondal: Oh, I see. I thought len is O(1) operation but it seems, during each invocation it gets recomputed at O(logn).
Nice nit picking :+1:

apisix/utils/batch-processor.lua
bisakhmondal: Yes, defenitely

apisix/utils/batch-processor.lua
bisakhmondal: For it not to be nil, we have to update all existing instances of consumers. also, it's kind of enforcing the future implementation to return something like `false, err, 1` when the consumers consume the entries in one go and hit an error. Do we really need that sort of complication? I think it's better to leave that part to the consumers themselves like if they want to use the subtle feature or not. What do you say?

bisakhmondal: Oh damn, got your point. It's the if block, haha

apisix/plugins/datadog.lua
```diff
@@ -108,127 +109,144 @@ local function generate_tag(entry, const_tags)
...
+    if prefix ~= "" then
+        prefix = prefix .. "."
+    end
 
-        if ctx.route_name and ctx.route_name ~= "" then
-            entry.route_id = ctx.route_name
-        end
+    -- request counter
+    ok, err = sock:send(format("%s:%s|%s%s", prefix ..
+                       "request.counter", 1, "c", suffix))
```
spacewander, line 132: My bad! Actually, I mean the alignment like:
https://github.com/apache/apisix/blob/615ee41312282ca2375abddae1d074637a901d89/apisix/upstream.lua#L274-L275

spacewander (approved):

apisix/plugins/datadog.lua
bisakhmondal: Fixed

tzssangglass (approved):

PR #6214, opened 2022-01-27
fix(admin-token): fixed token is used by default

shuaijinchao:
https://lists.apache.org/thread/m6s1ozvhp8ww4mcw85q2m5mp4x7fkoxh

### Pre-submission checklist:


conf/config.yaml
```diff
@@ -43,5 +43,5 @@
 apisix:
   admin_key:
     - name: admin
-      key: edd1c9f034335f136f87ad84b625c8f1  # using fixed API token has security risk, please update it when you deploy to production environment
+      key: ${{APISIX_ADMIN_API_KEY}}
```
moonming, line 46: According to the discussion on the mailing list, the default value should be empty? I don't understand what the environment variables here mean

moonming: do we need to update https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L91-L100?

moonming: any do need to remove https://github.com/apache/apisix/blob/master/apisix/admin/init.lua#L64-L68?

moonming (changes requested):

apisix/cli/file.lua
```diff
@@ -52,6 +58,32 @@ local function tab_is_array(t)
 end
 
 
+local function generate_random_str(len)
+    local rand_str = ""
+    local rand_num = 0
+
+    -- set random seed
+    randomseed(tostring(os_time()):reverse():sub(1, 5))
```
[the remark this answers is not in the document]
shuaijinchao: the cli program runs independently of APISIX, and the logic here has not been triggered when the cli is executed.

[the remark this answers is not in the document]
moonming: https://github.com/apache/apisix/blob/master/apisix/cli/ops.lua#L149 Maybe you can put this function in cli `init`

conf/config.yaml
shuaijinchao: I think it would be more flexible to replace with an environment variable. Users can modify the admin token through system environment variables. If the environment variable is not read. At this point, APISIX will generate a random token for it and prompt it on standard output and logs.

shuaijinchao:
hi @moonming , Regarding setting the admin token to remove https://github.com/apache/apisix/pull/6214#discussion_r793302104, I think it would be more flexible to replace with an environment variable. Users can modify the admin token through system environment variables. If the environment variable is not read. At this point, APISIX will generate a random token for it and prompt it on standard output and logs.

apisix/cli/file.lua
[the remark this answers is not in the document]
shuaijinchao: APISIX and CLI are two completely separate programs. There is no sharing between them. APISIX runs on OpenResty and the CLI runs on luajit or native lua.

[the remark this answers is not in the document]
moonming: I means you can make random seed of CLI in https://github.com/apache/apisix/blob/master/apisix/cli/ops.lua#L149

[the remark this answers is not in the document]
shuaijinchao: I got it, thanks for the pointers.

apisix/init.lua
```diff
@@ -86,6 +88,12 @@ function _M.http_init(args)
             core.log.error("failed to load the configuration: ", err)
         end
     end
+
+    local is_cli_token = getenv(cli_util.ADMIN_TOKEN_TAG_KEY)
+    if is_cli_token and tonumber(is_cli_token) == 1 then
+        core.log.warn("admin token has been automatically created by the system, ",
```
tokers, line 94: system is too generic, using `apisix` will be better.

shuaijinchao: updated, thx~

apisix/init.lua
```diff
@@ -86,6 +88,12 @@ function _M.http_init(args)
             core.log.error("failed to load the configuration: ", err)
         end
     end
+
+    local is_cli_token = getenv(cli_util.ADMIN_TOKEN_TAG_KEY)
+    if is_cli_token and tonumber(is_cli_token) == 1 then
+        core.log.warn("admin token has been automatically created by the apisix, ",
+                        "value: `", getenv(cli_util.ADMIN_TOKEN_KEY), "`")
+    end
```
moonming, line 96: This is just a one-time token and should be very explicit about this

moonming: After Apache APISIX restarts, the token will change

moonming (changes requested):

ci/linux_apisix_current_luarocks_runner.sh
```diff
@@ -66,6 +67,10 @@ script() {
     ulimit -n -S
     ulimit -n -H
 
+    # set default admin token
+    sudo yamlq e -i '.apisix.admin_key[0].key = "edd1c9f034335f136f87ad84b625c8f1"' conf/config-default.yaml
```
moonming, line 71: why set token twice https://github.com/apache/apisix/pull/6214/files#diff-210dceffdb6579327475828287fd91c559b209de07ebf0323ab8005b589d9c0fR44？

shuaijinchao: There are two role users in the default configuration template of APISIX, namely `admin` and `viewer`.

shuaijinchao:
@moonming The current known security issue of `Admin API` is due to `X-REAL-IP` and has been fixed, should this PR continue?
I think the implementation of this PR can still improve the security capabilities of APISIX and does not conflict with the repair of `X-REAL-IP`.

shuaijinchao:
the problem has not been confirmed, and the PR is temporarily closed.

PR #6218, opened 2022-01-27
docs: add public-api related content

bzp2010:

Add public-api related documents, and remove obsolete content regarding plugin interceptors and plugin development.

### Pre-submission checklist:


docs/en/latest/plugins/prometheus.md
```diff
@@ -46,17 +46,16 @@ plugin_attr:
 
 Assume environment variable `INTRANET_IP` is `172.1.1.1`, now APISIX will export the metrics via `172.1.1.1:9092`.
 
-**Before version `2.6`, the metrics are exposed via the data panel port,
-you may need to use [interceptors](../plugin-interceptors.md) to protect it.**
-
-If you still want this behavior, you can configure it like this:
+If you still want expose the metrics via the data panel port, you can configure it like this:
```
juzhiyuan, line 49:
```suggestion
If you still want to expose the metrics via the data panel port, you can configure it like this:
```

docs/en/latest/plugins/public-api.md
```diff
@@ -0,0 +1,132 @@
...
+## Description
+
+The `public-api` plugin is used to enhance the plugin public API access control.
+When current users develop custom plugins, they can register some public APIs for fixed functionality, such as the `/apisix/plugin/jwt/sign` API in `jwt-auth`. These APIs can only apply limited plugins for access control (currently only `ip-restriction`) by way of plugin interceptors.
+
+With the `public-api` plugin, we put the all public API into general HTTP API router, which is consistent with the normal route registered by the user and can apply any plugin. The public API added in the user plugin is no longer open by default by APISIX, but the user manually configures the route for it, and the user can configure any uri and plugin.
+
+## Attributes
+
+| Name | Type | Requirement | Default | Valid | Description |
```
juzhiyuan, line 39:
```suggestion
| Name | Type | Required | Default | Valid | Description |
```

docs/zh/latest/plugins/batch-requests.md
```diff
@@ -52,8 +52,7 @@ title: batch-requests
 
 ## 接口
 
-插件会增加 `/apisix/batch-requests` 这个接口，你可能需要通过 [interceptors](../plugin-interceptors.md)
-来保护它。
+插件会增加 `/apisix/batch-requests` 这个接口，你可能需要通过 [public-api](public-api.md) 插件来暴露它。
```
juzhiyuan, line 55:
```suggestion
插件会增加 `/apisix/batch-requests` 这个接口，你需要通过 [public-api](public-api.md) 插件来暴露它。
```

docs/zh/latest/plugins/jwt-auth.md
```diff
@@ -51,8 +51,7 @@ title: jwt-auth
 
 ## 接口
 
-插件会增加 `/apisix/plugin/jwt/sign` 这个接口，你可能需要通过 [interceptors](../plugin-interceptors.md)
-来保护它。
+插件会增加 `/apisix/plugin/jwt/sign` 这个接口，你可能需要通过 [public-api](public-api.md) 插件来暴露它。
```
juzhiyuan, line 54:
```suggestion
插件会增加 `/apisix/plugin/jwt/sign` 这个接口，你需要通过 [public-api](public-api.md) 插件来暴露它。
```

docs/zh/latest/plugins/node-status.md
```diff
@@ -40,7 +40,7 @@ title: node-status
 
 ## 插件接口
 
-插件增加接口 `/apisix/status`，可通过 [interceptors](../plugin-interceptors.md) 保护该接口。
+插件增加接口 `/apisix/status`，你可能需要通过 [public-api](public-api.md) 插件来暴露它。
```
juzhiyuan, line 43:
```suggestion
插件增加接口 `/apisix/status`，你需要通过 [public-api](public-api.md) 插件来暴露它。
```

docs/zh/latest/plugins/wolf-rbac.md
```diff
@@ -51,7 +51,7 @@ rbac 功能由 [wolf](https://github.com/iGeeky/wolf) 提供, 有关 `wolf` 的
 * /apisix/plugin/wolf-rbac/change_pwd
 * /apisix/plugin/wolf-rbac/user_info
 
-你可能需要通过 [interceptors](../plugin-interceptors.md) 来保护它们。
+你可能需要通过 [public-api](public-api.md) 插件来暴露它。
```
juzhiyuan, line 54:
```suggestion
你需要通过 [public-api](public-api.md) 插件来暴露它。
```

docs/zh/latest/plugin-develop.md
```diff
@@ -326,8 +326,7 @@ function _M.api()
 end
 ```
 
-注意注册的接口会暴露到外网。
-你可能需要使用 [interceptors](plugin-interceptors.md) 来保护它。
+注意，注册的接口将不会默认暴露，您需要使用[public-api 插件](plugins/public-api.md)来暴露它。
```
shuaijinchao, line 329:
```suggestion
注意，注册的接口将不会默认暴露，需要使用[public-api 插件](plugins/public-api.md)来暴露它。
```
the use of honorifics for personal pronouns is not required in documentation.

docs/zh/latest/plugins/prometheus.md
```diff
@@ -46,17 +46,16 @@ plugin_attr:
...
-如果你依然想要这样的行为，你可以这么配置：
+如果你依然想要让指标暴露在数据面的端口上，你可以这么配置：
 
 ```
 plugin_attr:
   prometheus:
     enable_export_server: false
 ```
 
+同时，您还需要使用 [public-api](public-api.md) 插件来暴露它。
```
shuaijinchao, line 57:
```suggestion
同时，需要使用 [public-api](public-api.md) 插件来暴露它。
```

docs/en/latest/plugins/prometheus.md
bzp2010: @juzhiyuan Do you mean to set the `enable_export_server` default configuration to occur false, expose it to the DP port by default, and allow the user to set the route for this API?

docs/en/latest/plugins/public-api.md
bzp2010: https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/cors.md?plain=1#L36-L38

Other documents also use Requirement.

docs/zh/latest/plugin-develop.md
bzp2010: Get it

docs/en/latest/plugins/public-api.md
```diff
@@ -0,0 +1,132 @@
...
+| -- | -- | -- | -- | -- | -- |
+| uri | string | optional | "" |   | The uri of the public API. When you set up the route, you can use this to configure the original API uri if it is used in a way that is inconsistent with the original public API uri. |
```
tzssangglass, line 41:
```suggestion
| Name | Type   | Requirement | Default | Valid | Description                                                                                                                                                                            |
|------|--------|-------------|---------|-------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| uri  | string | optional    | ""      |       | The uri of the public API. When you set up the route, you can use this to configure the original API uri if it is used in a way that is inconsistent with the original public API uri. |
```

wrong table format

docs/en/latest/plugins/public-api.md
```diff
@@ -0,0 +1,132 @@
...
+```shell
+$ curl -i 'http://127.0.0.1:9080/gen_token?key=user-key'
+    -H "apikey: test-apikey"
+HTTP/1.1 200 OK
+
+# Failed request
+$ curl -i 'http://127.0.0.1:9080/gen_token?key=user-key'
+HTTP/1.1 401 UNAUTHORIZED
+```
+
```
tzssangglass, line 131: how about use 

```
$ curl 'http://127.0.0.1:9080/apisix/plugin/jwt/sign?key=user-key'
```

to get token atfer protecting this route?

docs/en/latest/plugins/public-api.md
tzssangglass: here `Requirement` is follow other docs.

docs/en/latest/plugins/public-api.md
bzp2010: @tzssangglass Is this the new specification we are implementing? I've written documents before that use this style.

docs/en/latest/plugins/public-api.md
bzp2010: This is inherited from the second case. If you don't use the `/get_token` route, you can just configure the public API route in the first way (Basic Use Case).

docs/zh/latest/plugins/prometheus.md
bzp2010: fixed

docs/zh/latest/plugin-develop.md
bzp2010: fixed, Other personal pronouns in the md file of this PR design have also been removed.

docs/zh/latest/plugins/wolf-rbac.md
bzp2010: fixed

docs/zh/latest/plugins/node-status.md
bzp2010: fixed

docs/zh/latest/plugins/jwt-auth.md
bzp2010: fixed

docs/zh/latest/plugins/batch-requests.md
bzp2010: fixed

docs/en/latest/plugins/public-api.md
tzssangglass: my bad, it's just that I noticed in the IDE that it's not formatted correctly to show up as a table.
I just checked its presentation on github and it is correct.

docs/en/latest/plugins/public-api.md
tzssangglass: Does this mean that the `/apisix/plugin/jwt/sign` is still exposed?

docs/en/latest/plugins/prometheus.md
juzhiyuan: Because we don't use `data plane` in docs frequently, how about using `via the data plane port (default: xxx)` here?

Also, it should be `plane` instead of `panel`

docs/en/latest/plugins/public-api.md
juzhiyuan: ok

docs/en/latest/plugins/prometheus.md
bzp2010: I think it is ok, I will add the description of the default port. The `panel` is a spelling mistake.

bzp2010: @juzhiyuan updated

docs/zh/latest/plugins/batch-requests.md
bzp2010: updated, please re-check and resolve it. XD

docs/en/latest/plugins/public-api.md
bzp2010: @tzssangglass 

It will still be inserted into the radixtree built in `api_router` at startup, but in the default `http_access_phase`, no match will be made. If the user configures a route for it and turns on the `public-api` plugin, we do an `api_router` match in the `public-api` plugin's access hook to process the public API request.

Therefore, the conclusion is that it is not exposed and the user needs to configure the route to access it.

docs/zh/latest/plugin-develop.md
bzp2010: ping @shuaijinchao

docs/en/latest/plugins/public-api.md
tzssangglass: got

docs/en/latest/plugins/public-api.md
```diff
@@ -0,0 +1,132 @@
...
+- [**Description**](#description)
+- [**Attributes**](#attributes)
+- [**Example**](#example)
+
+## Description
+
+The `public-api` plugin is used to enhance the plugin public API access control.
+When current users develop custom plugins, they can register some public APIs for fixed functionality, such as the `/apisix/plugin/jwt/sign` API in `jwt-auth`. These APIs can only apply limited plugins for access control (currently only `ip-restriction`) by way of plugin interceptors.
+
+With the `public-api` plugin, we put the all public API into general HTTP API router, which is consistent with the normal route registered by the user and can apply any plugin. The public API added in the user plugin is no longer open by default by APISIX, but the user manually configures the route for it, and the user can configure any uri and plugin.
```
spacewander, line 35:
```suggestion
With the `public-api` plugin, we put all public API into the general HTTP API router, which is consistent with the normal route registered by the user and can apply any plugin. The public API added in the user plugin is no longer open by default by APISIX, but the user manually configures the route for it, and the user can configure any uri and plugin.
```

juzhiyuan, line 35: I prefer to using Route instead of route, so developers could know what's indeed normal router 😉

```suggestion
With the `public-api` plugin, we put all public APIs into general HTTP API router, which is consistent with the normal Route registered by the user and can apply any plugin. The public API added in the user plugin is no longer accessible by default by APISIX, and the user has to manually configure the route for it, the user can configure any uri and plugin.
```

docs/zh/latest/plugins/node-status.md
```diff
@@ -40,7 +40,7 @@ title: node-status
 
 ## 插件接口
 
-插件增加接口 `/apisix/status`，可通过 [interceptors](../plugin-interceptors.md) 保护该接口。
+插件增加接口 `/apisix/status`，需要通过 [public-api](public-api.md) 插件来暴露它。
```
spacewander, line 43: Let's update the link to point to English version like https://github.com/apache/apisix/blob/48e8a1ee483caa7150f7ad812953730eb50324bb/docs/zh/latest/plugins/kafka-logger.md?plain=1#L183

tzssangglass (approved):

docs/zh/latest/plugins/node-status.md
bzp2010: Hi, @spacewander.

As currently used, it renders properly on our official website (when no language-specific localization documents are returned to English), and if we modify it directly, it will cause confusion on the official website, i.e. some multilingual documents will be all linked back to the English version.
It seems like we can't have it both ways, but I think it's more important to make sure the content on the official website is always available in the right way than to use the md file to read it directly.

Currently there is no Chinese documentation for the `public-api` plugin, if you think it is really necessary to change it, I will do so.

docs/en/latest/plugins/public-api.md
bzp2010: fixed

docs/zh/latest/plugins/node-status.md
juzhiyuan: How about using English only in this PR? After it gets merged, @guoqqqi and @Baoyuantop could help to translate 🤔

juzhiyuan (approved):
LGTM

LiteSun (approved):

shuaijinchao (approved):

docs/zh/latest/plugins/node-status.md
spacewander: @bzp2010 @juzhiyuan 
We already do this in other loggers. Although someone else can translate it later, but we should not invent another way to handle it, and merge a PR which will create known broken links.

bzp2010: @spacewander updated

spacewander (approved):

Issue #6175, opened 2022-01-21
request help: api-breaker plugin donot work


davidwhf:
### Issue description

```.sh
curl http://127.0.0.1:9080/apisix/admin/routes/1 \
-H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT \
-d '
{
    "plugins": {
        "api-breaker": {
            "break_response_code": 502,
            "unhealthy": {
                "http_statuses": [503],
                "failures": 2
            },
            "healthy": {
                "http_statuses": [200],
                "successes": 1
            }
        }
    },
    "uri": "/test/*"
}'
curl -i http://127.0.0.1:9080/test/index.html
curl -i http://127.0.0.1:9080/test/index.html
curl -i http://127.0.0.1:9080/test/index.html
```
I execute command curl three times, it still return 503,not 502.
```.html
HTTP/1.1 503 Service Temporarily Unavailable
Date: Fri, 21 Jan 2022 01:55:21 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 194
Connection: keep-alive
Server: APISIX/2.11.0

<html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>openresty</center>
</body>
</html>
```

### Environment

- apisix version (cmd: `apisix version`): 2.11.0
- OS (cmd: `uname -a`): Linux 5d9c3c92e8b1 3.10.0-1127.el7.x86_64 #1 SMP Tue Mar 31 23:36:51 UTC 2020 x86_64 Linux
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): 
nginx version: openresty/1.19.3.2
built by gcc 9.3.0 (Alpine 9.3.0) 
built with OpenSSL 1.1.1g  21 Apr 2020
TLS SNI support enabled
[...]

PR #6219, opened 2022-01-27
docs: optimizing api-breaker documentation

Baoyuantop:

resolve https://github.com/apache/apisix/issues/6175

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/plugins/api-breaker.md
+++ b/docs/en/latest/plugins/api-breaker.md
@@ -80,8 +80,13 @@ curl "http://127.0.0.1:9080/apisix/admin/routes/1" -H 'X-API-KEY: edd1c9f034335f
             }
         }
     },
+    "upstream": {
+        "type": "roundrobin",
+        "nodes": {
+            "127.0.0.1:1980": 1
+        }
+    },
     "uri": "/hello",
-    "host": "127.0.0.1",
 }'
 ```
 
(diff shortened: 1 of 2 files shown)


docs/en/latest/plugins/api-breaker.md
```diff
@@ -90,7 +95,7 @@ curl "http://127.0.0.1:9080/apisix/admin/routes/1" -H 'X-API-KEY: edd1c9f034335f
 Then. Like the configuration above, if your upstream service returns 500. 3 times in a row. The client will receive a 502 (break_response_code) response.
 
 ```shell
-$ curl -i -X POST "http://127.0.0.1:9080/get"
+$ curl -i -X POST "http://127.0.0.1:9080/hello"
```
zaunist, line 98: A route is not configured api-breaker and which have a unavailable upstream, APISIX will return 502 error, so your example is not enough to say it's due to the 502 error returned by the api-breaker plugin.

spacewander (approved):

docs/en/latest/plugins/api-breaker.md
spacewander: Yes. Maybe we can change the status code.

Most of the documentation is using the 1980 port as a fake upstream. So we can assume people can get used to it.

We can't use 9080 port as it may cause a loop.
`httpbin.org` is another option, but it can't be accessed in the China mainland, and it only supports `http`.

tzssangglass (approved):

PR #6228, opened 2022-01-28
feat(loggly): support HTTP response code to SYSLOG severity mapping

### LINKED ISSUE #6112: Proposal: Develop Loggly logging plugin

bisakhmondal:

closed #6112 (phase 2 step 2)
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/loggly.lua
+++ b/apisix/plugins/loggly.lua
@@ -22,7 +22,6 @@ local path = require("pl.path")
 local http = require("resty.http")
 local ngx = ngx
 local tostring = tostring
-local tonumber = tonumber
 local pairs = pairs
 local tab_concat = table.concat
 local udp = ngx.socket.udp
@@ -90,12 +89,13 @@ local schema = {
             type = "object",
             description = "upstream response code vs syslog severity mapping",
             patternProperties = {
-                [".*"] = {
+                ["^[1-5][0-9]{2}$"] = {
                     description = "keys are HTTP status code, values are severity",
                     type = "string",
                     enum = severity_enums
                 },
             },
+            additionalProperties = false
         }
     },
     required = {"customer_token"}
@@ -161,10 +161,6 @@ function _M.check_schema(conf, schema_type)
     if conf.severity_map then
         local cache = {}
         for k, v in pairs(conf.severity_map) do
-            local rcode = tonumber(k)
-            if not rcode or rcode < 100 or rcode >= 600 then
-                return nil, "expecting severity_map with http response code([100,599]) as keys"
-            end
             cache[k] = severity[v:upper()]
         end
         conf._severity_cache = cache
--- a/apisix/plugins/loggly.lua
+++ b/apisix/plugins/loggly.lua
@@ -22,6 +22,7 @@ local path = require("pl.path")
 local http = require("resty.http")
 local ngx = ngx
 local tostring = tostring
+local tonumber = tonumber
 local pairs = pairs
 local tab_concat = table.concat
 local udp = ngx.socket.udp
@@ -85,6 +86,10 @@ local schema = {
             type = "boolean",
             default = true
         },
+        severity_map = {
+            type = "object",
+            description = "upstream response code vs syslog severity mapping"
+        }
     },
     required = {"customer_token"}
 }
@@ -145,6 +150,19 @@ function _M.check_schema(conf, schema_type)
     if not ok then
         return nil, err
     end
+
+    if conf.severity_map then
+        for k, v in pairs(conf.severity_map) do
+            local rcode = tonumber(k)
+            if not rcode or rcode < 100 or rcode >= 600 then
+                return nil, "expecting severity_map with http response code([100,599]) as keys"
+            end
+            local s = severity[v:upper()]
+            if not s then
+                return nil, "expecting severity_map severity level keywords as values"
+            end
+        end
+    end
     return log_util.check_log_schema(conf)
 end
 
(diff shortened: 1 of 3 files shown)


tzssangglass (dismissed):

apisix/plugins/loggly.lua
```diff
@@ -184,9 +202,18 @@ local function generate_log_message(conf, ctx)
             core.table.insert(taglist, "tag=\"" .. conf.tags[i] .. "\"")
         end
     end
+
+    local message_severity = severity[conf.severity:upper()]
```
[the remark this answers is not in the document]
bisakhmondal: Done

apisix/plugins/loggly.lua
```diff
@@ -145,6 +150,19 @@ function _M.check_schema(conf, schema_type)
     if not ok then
         return nil, err
     end
+
+    if conf.severity_map then
+        for k, v in pairs(conf.severity_map) do
+            local rcode = tonumber(k)
```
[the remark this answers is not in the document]
bisakhmondal: Done

apisix/plugins/loggly.lua
```diff
@@ -145,6 +157,18 @@ function _M.check_schema(conf, schema_type)
...
     end
+
+    if conf.severity_map then
+        local cache = {}
+        for k, v in pairs(conf.severity_map) do
+            local rcode = tonumber(k)
+            if not rcode or rcode < 100 or rcode >= 600 then
+                return nil, "expecting severity_map with http response code([100,599]) as keys"
+            end
+            cache[k] = severity[v:upper()]
```
spacewander, line 168: We can convert k to a number?

apisix/plugins/loggly.lua
```diff
@@ -85,6 +86,17 @@ local schema = {
             type = "boolean",
             default = true
         },
+        severity_map = {
+            type = "object",
+            description = "upstream response code vs syslog severity mapping",
+            patternProperties = {
+                [".*"] = {
```
spacewander, line 93: We can use regex to verify the range of status code?

bisakhmondal: Yea, I had tested. It doesn't work that way. 
say for eg,
`["^[1-5][0-9]{2}"] = {...}` the check schema match the enums for only the set of keys where the key is a status code (regex). For other scenarios (key = "ABC", "500a") there is no matching regex, so it simply doesn't match and validate with the pattern and let it pass as valid inputs.

apisix/plugins/loggly.lua
bisakhmondal: I think I avoid that due to the default behaviour of cjson encoding `Cannot serialise table: excessively sparse array`. 
we can override the behaviour by `cjson.encode_sparse_array(true)`. In this case, I thought I don't need it that's why I am sticking with strings. WDYT?

spacewander: OK

apisix/plugins/loggly.lua
spacewander: Does `["^[1-5][0-9]{2}$"] = {...}` work for it?

bisakhmondal: @spacewander It works as well as doesn't work.
Example 1: Want :heavy_check_mark:  Got: :heavy_check_mark: 
```json
...
           "severity_map": {
                "200": "INFO"
            }
...
```

Example 2: Want :x:  Got: :x: 
```json
...
           "severity_map": {
                "200": "abc"
            }
...
```

Example 3: Want :x:  Got: :heavy_check_mark: : 
```json
...
           "severity_map": {
                "abc": "def"
            }
...
```
The problem is, if the key fails to match regex, the pattern properties don't get applied. Also, the program doesn't throw an error for the key mismatch.

spacewander: Got it. What about adding `"additionalProperties": false`? So the mismatch key will cause an error.

bisakhmondal: It worked! Awesome : )
```
// inp "severity_map": {
//               "abc": "def"
//            }
{
    "error_msg": "failed to check the configuration of plugin loggly err: property \"severity_map\" validation failed: additional properties forbidden, found abc"
}
```

spacewander (approved):

tzssangglass (approved):

PR #6246, opened 2022-02-03
ci: runner shell sudo doesn't affect redirects SC2024

kwanhur:

fix ci shell scripts warning `sudo` doesn't affect redirects [SC2024](https://github.com/koalaman/shellcheck/wiki/SC2024)

- use `tee` copies standard input to standard output, instead of `>` to write

### Pre-submission checklist:

### DIFF
--- a/ci/linux_apisix_current_luarocks_runner.sh
+++ b/ci/linux_apisix_current_luarocks_runner.sh
@@ -33,8 +33,8 @@ script() {
     sudo rm -rf /usr/local/share/lua/5.1/apisix
 
     # install APISIX with local version
-    sudo luarocks install rockspec/apisix-master-0.rockspec --only-deps | sudo tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
-    sudo luarocks make rockspec/apisix-master-0.rockspec | sudo tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
+    sudo luarocks install rockspec/apisix-master-0.rockspec --only-deps | tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
+    sudo luarocks make rockspec/apisix-master-0.rockspec | tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
     # ensure all files under apisix is installed
     diff -rq apisix /usr/local/share/lua/5.1/apisix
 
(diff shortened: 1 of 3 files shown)


ci/linux_apisix_current_luarocks_runner.sh
```diff
@@ -33,8 +33,8 @@ script() {
     sudo rm -rf /usr/local/share/lua/5.1/apisix
 
     # install APISIX with local version
-    sudo luarocks install rockspec/apisix-master-0.rockspec --only-deps > build.log 2>&1 || (cat build.log && exit 1)
-    sudo luarocks make rockspec/apisix-master-0.rockspec > build.log 2>&1 || (cat build.log && exit 1)
+    sudo luarocks install rockspec/apisix-master-0.rockspec --only-deps | sudo tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
```
spacewander, line 36: We don't need sudo to write to `build.log`? So the original code's OK.

kwanhur: The original redirects stdout into `build.log`, it isn't recommended for using `sudo`.

If it needs log into `build.log`, suggest using `tee` to make it.

spacewander: No, use no-sudo user is enough to write log to `build.log`.

ci/linux_apisix_master_luarocks_runner.sh
```diff
@@ -37,7 +37,7 @@ script() {
     cp -r ../utils ./
 
     # install APISIX by luarocks
-    sudo luarocks install $APISIX_MAIN > build.log 2>&1 || (cat build.log && exit 1)
+    sudo luarocks install $APISIX_MAIN | tee build.log > /dev/null 2>&1 || (cat build.log && exit 1)
```
spacewander, line 40: There is no need to use `tee` to redirect. The original way is good enough.

spacewander (changes requested):

spacewander:
The SC2024 requires `sudo tee` because `sudo` doesn't affect `>`.
See the example, `echo 3 | sudo tee /proc/sys/vm/drop_caches > /dev/null`.

However, as I mentioned several days ago, we don't need `sudo` to write to `build.log`. There is no need to introduce an extra cmd to do what `>` can do.

Issue #5767, opened 2021-12-10
bug: nacos.lua make some duplicate instance 


xianshun163:
### Issue description

1、deploy a  service  to register to the nacos
2、config some rule  and service and upstreams  at the apisix
3、make a test   
4、add some  log  at the balancer.lua, and  print the  upstream  nodes
5、 from the log  , that  is  3 instance  in the apisix,  but  that is  one  in the nacos.

I  read the  code  nacos.lua   ,and  find the reason:  it will get the same service  the  follow code,
local function get_nacos_services()
    local services = {}
    -- here we use lazy load to work around circle dependency
    local get_upstreams = require('apisix.upstream').upstreams
    local get_routes = require('apisix.router').http_routes
    local get_services = require('apisix.http.service').services
    local values = get_upstreams()
    iter_and_add_service(services, values)
    values = get_routes()
    iter_and_add_service(services, values)
    values = get_services()
    iter_and_add_service(services, values)
    return services
end

---------------------
and  I  try to fix it  like this, in the  local function fetch_full_registry(premature)  :
local infos = get_nacos_services()
    if #infos == 0 then
        applications = up_apps
        return
    end
    table.sort(infos,function(a,b)
        return a.namespace_id .. a.group_name .. a.service_name > b.namespace_id .. b.group_name .. b.service_name
    end)
    local last_service_info
    for _, service_info in ipairs(infos) do
        if compareService(last_service_info,service_info) then
            goto CONTINUE
        end
        last_service_info = service_info
        local data, err
        local namespace_id = service_info.namespace_id

----------------------
It  that   another  better way to fix it?  Thank you who can  fix it  in the github,     and I   can   get the code  next time。

### Environment

- apisix version (cmd: `apisix version`):2.10.1
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

### Steps to reproduce

use  the  nacos as the discovery .
and  add some code to print the upstream nodes  at the balancer.lua

### Actual result

the  instance is the same as the nacos

### Error log

no  error log

tzssangglass:
[...]

PR #6247, opened 2022-02-04
Add the de duplication function of Nacos service discovery

xianshun163:

bug: nacos.lua make some duplicate instance #5767

### Pre-submission checklist:

### DIFF
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -236,7 +236,18 @@ local function is_grpc(scheme)
 
     return false
 end
-
+local function compareService(last_service_info,service_info)
+    if not last_service_info or not service_info then
+        return false
+    end
+    if last_service_info.namespace_id == service_info.namespace_id and
+            last_service_info.group_name == service_info.group_name and
+            last_service_info.service_name == service_info.service_name then
+        return true
+    else
+        return false
+    end
+end
 local function fetch_full_registry(premature)
     if premature then
         return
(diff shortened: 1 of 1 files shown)


juzhiyuan:
Hi @xianshun163, please notice the PR's checklist :) When we change codes, we need also improve the testcases :)

apisix/discovery/nacos/init.lua
```diff
@@ -236,7 +236,18 @@ local function is_grpc(scheme)
 
     return false
 end
-
+local function compareService(last_service_info,service_info)
```
tzssangglass, line 239:
```suggestion
local function compare_service(last_service_info,service_info)
```

tzssangglass:
Need to add test cases

apisix/discovery/nacos/init.lua
```diff
@@ -259,7 +270,17 @@ local function fetch_full_registry(premature)
         return
     end
 
+    table.sort(infos,function(a,b)
```
leslie-tsang, line 273: Need to add `local table = table` to avoid a global variables access.

leslie-tsang (changes requested):

apisix/discovery/nacos/init.lua
starsz, line 239:
```suggestion
local function compare_service(last_service_info, service_info)
```

tzssangglass:
hi @xianshun163 , I have another fix idea to discuss with you.

Before adding a new service to services, we can do a de-duplication, we can add de_duplication function in https://github.com/apache/apisix/blob/06c51935aa207eef8ef75df24172e8656085c3ff/apisix/discovery/nacos/init.lua#L200-L203

L202

the code may like 

```lua
        local dup = de_duplication(services, namespace_id, group_name,
                up.service_name, up.scheme)
        if dup then
            goto CONTINUE
        end
        if up.discovery_type == 'nacos' then
```

we can compare new service and services in de_duplication function, such as:

```lua
local function de_duplication(services, namespace_id, group_name, service_name, scheme)
    for _, service in ipairs(services) do
        if service.namespace_id == namespace_id and service.group_name == group_name
                and service.service_name == service_name and service.scheme == scheme then
            return true
        end
    end
    return false
end
```

tzssangglass:
Let's turn this off, fixed by #6282

PR #6255, opened 2022-02-07
feat: release 2.10.4

bzp2010:

LTS release: 2.10.4

### Pre-submission checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ title: Changelog
 
 ## Table of Contents
 
+- [2.10.4](#2104)
 - [2.10.3](#2103)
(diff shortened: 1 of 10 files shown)


tzssangglass (approved):

spacewander (approved):

initlove:
I have a question, the way to fix this is by convert the 'real_ip_hdr' reading from configs to lower case.
Is that mean, if the config file is already in lower case, this bug will not affect old version, right?


    -- we don't need to handle '_' to '-' as Nginx won't treat 'X_REAL_IP' as 'X-Real-IP'
    real_ip_hdr = str_lower(real_ip_hdr)

PR #6270, opened 2022-02-09
feat(oidc): add set_userinfo_detail_header

yaule:
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -92,6 +92,12 @@ local schema = {
                 "header to the request for downstream.",
             type = "boolean",
             default = true
+        },
+        set_userinfo_detail_header = {
+            description = "Whether the user info token should be added in the X-Userinfo-Detail-* in scope" ..
+                "header to the request for downstream.",
+            type = "boolean",
+            default = false
         }
     },
     required = {"client_id", "client_secret", "discovery"}
@@ -316,6 +322,13 @@ function _M.rewrite(plugin_conf, ctx)
             if response.user and conf.set_userinfo_header then
                 core.request.set_header(ctx, "X-Userinfo",
                     ngx_encode_base64(core.json.encode(response.user)))
+            end
+            -- Add X-Userinfo-Detail header, maybe.
+            if response.user and conf.set_userinfo_detail_header then
+                for k in string.gmatch(conf.scope..' ', '(.-)'..' ')
+                do
+                    core.request.set_header(ctx, 'X-Userinfo-Detail-'..k,response.user[k])
+                end 
             end
         end
     end


tzssangglass:
hi @yaule , thanks for submitting the PR, can you explain the background of this PR?

apisix/plugins/openid-connect.lua
```diff
@@ -317,6 +323,13 @@ function _M.rewrite(plugin_conf, ctx)
                 core.request.set_header(ctx, "X-Userinfo",
                     ngx_encode_base64(core.json.encode(response.user)))
             end
+            -- Add X-Userinfo-Detail header, maybe.
+            if response.user and conf.set_userinfo_detail_header then
+                for k in string.gmatch(conf.scope..' ', '(.-)'..' ')
```
starsz, line 328:
```suggestion
                for k in string.gmatch(conf.scope .. ' ', '(.-)' .. ' ')
```

apisix/plugins/openid-connect.lua
```diff
@@ -317,6 +323,13 @@ function _M.rewrite(plugin_conf, ctx)
...
+                do
+                    core.request.set_header(ctx, 'X-Userinfo-Detail-'..k,response.user[k])
```
starsz, line 330:
```suggestion
                    core.request.set_header(ctx, 'X-Userinfo-Detail-' .. k, response.user[k])
```

yaule:
> hi @yaule , thanks for submitting the PR, can you explain the background of this PR?

The service behind APISIX can directly obtain user information from the header, without using base64 to encode the data, like grafana

tzssangglass:
Can you show the relevant docs or issue, I don't understand its usefulness.

yaule:
> 

https://grafana.com/docs/grafana/latest/auth/auth-proxy/

tzssangglass:
we have `set_userinfo_header` attributes in openid-connect plugin, see: https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/openid-connect.md#attributes

It looks like this feature is already available.

yaule:
Grafana requires the header is plaintext, but X-Userinfo is base64 encoded.

tzssangglass:
> Grafana requires the header is plaintext, but X-Userinfo is base64 encoded.

Another question, is there an introduction to `X-Userinfo-Detail-*`? Is this header specifically for Grafana?

yaule:
Yes, the upstream services do not need base64 decoding user information, they can be obtained directly from X-userinfo-Detail-*

tzssangglass:
> Yes, the upstream services do not need base64 decoding user information, they can be obtained directly from X-userinfo-Detail-*

My concern is that this is only grafana usage and is not a universal standard.
I would like to know if there are more common standards that support this header.

PR #6281, opened 2022-02-10
feat: compresses the redis lua script in limit-count

leslie-tsang:

This PR compresses the `limit-count` plugin's redis lua script.
This will cut the cost of network traffic by about 15%.

### Pre-submission checklist:

### DIFF
--- a/apisix/core/string.lua
+++ b/apisix/core/string.lua
@@ -21,7 +21,7 @@ local str_find = string.find
 local ffi         = require("ffi")
 local C           = ffi.C
 local ffi_cast    = ffi.cast
-
+local ngx         = ngx
 
 ffi.cdef[[
     int memcmp(const void *s1, const void *s2, size_t n);
@@ -75,6 +75,14 @@ function _M.rfind_char(s, ch, idx)
         end
     end
     return nil
+end
+
+
+-- reduce network consumption by compressing string indentation
+function _M.compress_script(s)
+    s = ngx.re.gsub(s, [[^\s+]], "", "mjo")
+    s = ngx.re.gsub(s, [[,\s+]], ",", "mjo")
+    return s
 end
 
 
--- a/apisix/core/string.lua
+++ b/apisix/core/string.lua
@@ -80,6 +80,9 @@ end
 
 
 -- reduce network consumption by compressing string indentation
+-- this method should be used with caution
+-- it will remove the spaces at the beginning of each line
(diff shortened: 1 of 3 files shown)


apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -37,6 +38,10 @@ local script = [=[
     return redis.call('incrby', KEYS[1], -1)
 ]=]
 
+--- compact the redis lua script
+script = string.gsub(script, "    ", "")
```
tokers, line 42: Can we prepare a method/function to do this?

leslie-tsang: Yes for sure, I'm wondering where it would be more appropriate to put it.

tokers: Maybe `core.utils`?

spacewander: Better to put it in `core.string`. And the `ngx.re` version will be faster?

leslie-tsang: Done

spacewander (dismissed):

apisix/plugins/limit-count/limit-count-redis-cluster.lua
leslie-tsang: @token ping

apisix/core/string.lua
```diff
@@ -78,4 +78,12 @@ function _M.rfind_char(s, ch, idx)
 end
 
 
+-- reduce network consumption by compressing string indentation
```
tokers, line 82: I think the compress way is not ideal as it might also change the literal string (spaces after `,`), so it should be used carefully, add some comments to give a hint will be better.

tokers (dismissed):

apisix/core/string.lua
```diff
@@ -78,4 +79,15 @@ function _M.rfind_char(s, ch, idx)
...
+-- this method should be used with caution
+-- it will remove the spaces at the beginning of each line
+-- remove the spaces after `,` character
```
spacewander, line 85:
```suggestion
-- and remove the spaces after `,` character
```

tzssangglass (approved):

spacewander (approved):

Issue #6288, opened 2022-02-10
bug: traffic-split plugin do not match if the first matcher failed


crazyMonkey1995:
### Issue description

According to the plugin's documentation: https://apisix.apache.org/docs/apisix/plugins/traffic-split/
I configured two rules:
```
"traffic-split": {
      "rules": [
        {
          "match": [
            {
              "vars": [
                [
                  "http_x-my-header",
                  "==",
                  "1"
                ]
              ]
            }
          ],
          "weighted_upstreams": [
            {
              "upstream_id": "demo-1",
              "weight": 100
            }
          ]
        },
        {
          "weighted_upstreams": [
            {
              "upstream_id": "demo-2",
              "weight": 50
            },
            {
              "weight": 50
            }
          ]
        }
      ]
    }
```
When the first rule does not match, the second rule also fails, I looked at the relevant code, the problem should be here:
![image](https://user-images.githubusercontent.com/21658981/153405333-2743522f-c86c-47a2-8613-e0d7fab27824.png)

### Environment

- apisix version (cmd: `apisix version`): 2.12.0

### Steps to reproduce

1、create arbitrary routes
2、add traffic-split plugin for it config like above
3、access the relevant route, intentionally not matching the first rule

### Actual result

The second rule is ignored

### Error log

2022/02/10 19:46:47 [info] 137#137: *898767 [lua] traffic-split.lua:264: phase_func(): match_passed: false, client: 127.0.0.1, server: _, request: "GET /demo/ip HTTP/1.1", host: "test.com"

### Expected result

The second rule should be matched

crazyMonkey1995:
Add "match_passed=true" before line 241 should fix this?

PR #6242, opened 2022-02-01
fix(redirect-plugin): redirection loop behind a proxy or lb

soham4abc:
### What this PR does / why we need it:
Fixes #6078 

### Pre-submission checklist:

* [ ] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #6292, opened 2022-02-11
fix(traffic-split): failed to match rule when the first rule failed

crazyMonkey1995:
fix https://github.com/apache/apisix/issues/6288

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -238,6 +238,7 @@ function _M.access(conf, ctx)
 
     for _, rule in ipairs(conf.rules) do
         if not rule.match then
+            match_passed = true
             weighted_upstreams = rule.weighted_upstreams
             break
         end
--- a/t/plugin/traffic-split5.t
+++ b/t/plugin/traffic-split5.t
@@ -403,7 +403,6 @@ passed
                 local port = tonumber(res.body)
                 ports[i] = port
             end
-            table.sort(ports)
 
             local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello?id=2"
             for i = 3, 4 do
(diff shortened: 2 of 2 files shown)


apisix/plugins/traffic-split.lua
```diff
@@ -238,6 +238,7 @@ function _M.access(conf, ctx)
 
     for _, rule in ipairs(conf.rules) do
         if not rule.match then
+            match_passed = true
```
spacewander, line 241: The match_passed  is true by default:
https://github.com/apache/apisix/blob/50d6465fd7b6288f517fd5c408a3171ace7882dd/apisix/plugins/traffic-split.lua#L237

crazyMonkey1995: But if match_passed is set to false on the first loop, it will remain false instead of true on the second loop
image

spacewander: I see. Can you add a test to cover it?

spacewander: You can refer to https://github.com/apache/apisix/pull/6242#issuecomment-1028625979 and the https://github.com/apache/apisix/blob/master/t/plugin/traffic-split5.t

crazyMonkey1995: ok, I'll look into it

crazyMonkey1995: Take a look please

t/plugin/traffic-split5.t
```diff
@@ -311,3 +311,110 @@ passed
...
+
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello?id=1"
+            local ports = {}
+            local res, err
+            for i = 1, 2 do
+                res, err = httpc:request_uri(uri)
+                local port = tonumber(res.body)
+                ports[i] = port
+            end
+            table.sort(ports)
```
starsz, line 406: I think we can omit this call.

starsz (approved):

spacewander (approved):

PR #6306, opened 2022-02-13
build: rm luarocks tar file behind successful installation

kwanhur:

remove `luarocks` downloaded tar file

1. successfully install it, no need to reserve it.
2. install script remove the untar files (line#45)[https://github.com/apache/apisix/blob/master/utils/linux-install-luarocks.sh#L45], so remove the tar file simultaneously.

### Pre-submission checklist:

### DIFF
--- a/utils/linux-install-luarocks.sh
+++ b/utils/linux-install-luarocks.sh
@@ -25,6 +25,7 @@ fi
 LUAROCKS_VER=3.8.0
 wget https://github.com/luarocks/luarocks/archive/v"$LUAROCKS_VER".tar.gz
 tar -xf v"$LUAROCKS_VER".tar.gz
+rm -f luarocks-"$LUAROCKS_VER".tar.gz
 cd luarocks-"$LUAROCKS_VER" || exit
 
 OR_BIN="$OPENRESTY_PREFIX/bin/openresty"
@@ -42,7 +43,7 @@ fi
 make build > build.log 2>&1 || (cat build.log && exit 1)
 sudo make install > build.log 2>&1 || (cat build.log && exit 1)
 cd .. || exit
-rm -rf luarocks-"$LUAROCKS_VER" luarocks-"$LUAROCKS_VER".tar.gz
+rm -rf luarocks-"$LUAROCKS_VER"
 
 mkdir ~/.luarocks || true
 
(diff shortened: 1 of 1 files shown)


utils/linux-install-luarocks.sh
```diff
@@ -42,7 +42,7 @@ fi
 make build > build.log 2>&1 || (cat build.log && exit 1)
 sudo make install > build.log 2>&1 || (cat build.log && exit 1)
 cd .. || exit
-rm -rf luarocks-"$LUAROCKS_VER"
+rm -rf luarocks-"$LUAROCKS_VER" luarocks-"$LUAROCKS_VER".tar.gz
```
spacewander, line 45: Could we remove it just after untar?

kwanhur: > Could we remove it just after untar?

on this way, if `configure` failed, it need to download archive file `v3.8.0.tar.gz` repeatedly.

spacewander: OK. So that we can also handle https://github.com/apache/apisix/issues/6307 in this PR.

kwanhur: no, #6307 occurs at `configure` failure, so download twice.

spacewander: Yes. But they change the same thing twice. So why not put together?

kwanhur: ok, remove it behind untar was better, I'll update this logic.

utils/linux-install-luarocks.sh
```diff
@@ -25,6 +25,7 @@ fi
 LUAROCKS_VER=3.8.0
 wget https://github.com/luarocks/luarocks/archive/v"$LUAROCKS_VER".tar.gz
 tar -xf v"$LUAROCKS_VER".tar.gz
+rm -f luarocks-"$LUAROCKS_VER".tar.gz
```
leslie-tsang, line 28:
```suggestion
rm -f v"$LUAROCKS_VER".tar.gz
```

leslie-tsang (changes requested):

leslie-tsang (approved):

spacewander (approved):

PR #6314, opened 2022-02-14
fix(hmac-auth): hmac-auth plugin sort array param

### LINKED ISSUE #6313: bug: hmac-auth plugin will not sort array value in param
### Issue description

When I passing url like /api/xxx?age=123&name=123&a=2&a=1
this should be sorted as a=1&a=2&age=123&name=123
but I got a=2&a=1&age=2&name=1

### Environment

- OS (cmd: `uname -a`):

### Steps to reproduce

same as description

### Actual result

same as description

### Error log

same as description

### Expected result

_No response_

kevinw66:

fixes #6313 

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/hmac-auth.lua
+++ b/apisix/plugins/hmac-auth.lua
@@ -244,7 +244,13 @@ local function generate_signature(ctx, secret_key, params)
 
             -- whether to encode the uri parameters
             if type(param) == "table" then
+                local vals = {}
                 for _, val in pairs(param) do
+                    core.table.insert(vals, val)
+                end
+                core.table.sort(vals)
+
+                for _, val in pairs(vals) do
                     core.table.insert(query_tab, encode_or_not(key) .. "=" .. encode_or_not(val))
                 end
             else


apisix/plugins/hmac-auth.lua
```diff
@@ -244,7 +244,13 @@ local function generate_signature(ctx, secret_key, params)
 
             -- whether to encode the uri parameters
             if type(param) == "table" then
+                local vals = {}
```
tylitianrui, line 247: i  wonder that  why not use table.new

kevinw66: Just keep the same code style like table `keys`

starsz (commented):
Hi, @kevinw66.Can you add a test case to cover it.

kevinw66:
> Hi, @kevinw66.Can you add a test case to cover it.

Can you teach me how to run unit tests in my local machine ? I left a comment on issue.
Thanks.

kevinw66:
> 



> Hi! Thanks for detecting the issue and the PR. After looking into the codebase, I find the way of forming the canonical query string unnecessarily complex yet it's not accurate. I would go with a simpler approach rather than pathing this up.
> 
> We can replace:
> 
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/hmac-auth.lua#L219-L255
> 
> 
> with
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/aws-lambda.lua#L123-L130
> 
> 
> It's very simple, yet effective.
> cc @spacewander to share your views.

This may still need to check table param
```lua
for k, v in pairs(args) do
    if type(v) == "table" then
        for _, sv in pairs(v) do
            canonical_qs[#canonical_qs+1] = ngx.unescape_uri(k) .. "=" .. ngx.unescape_uri(sv)
        end
    else
        canonical_qs[#canonical_qs+1] = ngx.unescape_uri(k) .. "=" .. ngx.unescape_uri(v)
    end
end
```

kevinw66:
> Hi! Thanks for detecting the issue and the PR. After looking into the codebase, I find the way of forming the canonical query string unnecessarily complex yet it's not accurate. I would go with a simpler approach rather than pathing this up.
> 
> We can replace:
> 
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/hmac-auth.lua#L219-L255
> 
> 
> with
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/aws-lambda.lua#L123-L130
> 
> 
> It's very simple, yet effective.
> cc @spacewander to share your views.

This may not work correctly in some cases, in ASCII table, the '=' symbol is behind number sequence, so if I pass param like `a1a=2&a1a=1&a=2`, it will sort as `a1a=1&a1a=2&a=2`, but I think the result `a=2&a1a=1&a1a=2` may correct.

spacewander:
> Hi! Thanks for detecting the issue and the PR. After looking into the codebase, I find the way of forming the canonical query string unnecessarily complex yet it's not accurate. I would go with a simpler approach rather than pathing this up.
> 
> We can replace:
> 
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/hmac-auth.lua#L219-L255
> 
> with
> 
> https://github.com/apache/apisix/blob/605141cba33beccf71100fdea03c8e9608d4852f/apisix/plugins/aws-lambda.lua#L123-L130
> 
> It's very simple, yet effective.
> cc @spacewander to share your views.

Yes, the way in aws-lambda still needs to handle multiple values in the same query key.

apisix/plugins/hmac-auth.lua
```diff
@@ -244,7 +244,13 @@ local function generate_signature(ctx, secret_key, params)
...
                 for _, val in pairs(param) do
+                    core.table.insert(vals, val)
```
spacewander, line 249: Remember to convert boolean to `''` before sorting. `table.sort` only supports sorting in the same type.

kevinw66: done

apisix/plugins/hmac-auth.lua
tzssangglass: Not necessary to keep old code style here. We can specify the initialization size when we create a new table by using `core.table.new(narray, nhash)`

spacewander (approved):

bisakhmondal:
> This may still need to check table param
> 
> ```lua
> for k, v in pairs(args) do
>     if type(v) == "table" then
>         for _, sv in pairs(v) do
>             canonical_qs[#canonical_qs+1] = ngx.unescape_uri(k) .. "=" .. ngx.unescape_uri(sv)
>         end
>     else
>         canonical_qs[#canonical_qs+1] = ngx.unescape_uri(k) .. "=" .. ngx.unescape_uri(v)
>     end
> end
> ```

Ohh, Thanks! I missed the scenario : )
Btw, in that case,  something still looks odd! We are considering one extra level nesting in query params. Query params with nesting level > 2 would fail. 

For example: (random example, taken from web)
```json
main_key[other_key][main_array][0][name]=Bob&main_key[other_key][main_array][0][nickname]=bobby&main_key[other_key][main_array][1][name]=Tom&main_key[other_key][main_array][1][nickname]=Tommy

{
    "main_key": {
        "other_key": {
            "main_array": [{
                    "name": "Bob",
                    "nickname": "bobby"
                },
                {
                    "name": "Tom",
                    "nickname": "Tommy"
                }
            ]
        }
    }
}

```

I think we need to perform a recursive DFS! Else it will introduce some sort of indeterminism
```shell
🔥➜ ~ resty -e 'print(ngx.escape_uri({["a"]="b"}))' 

table%3A%200x7f5aa8abc608
🔥➜ ~ resty -e 'print(ngx.escape_uri({["a"]="b"}))'

table%3A%200x7f7eca005608
```

spacewander:
AFAIK, the `ngx.req.get_uri_args` only returns two level table at most.


> Multiple occurrences of an argument key will result in a table value holding all the values for that key in order.

https://github.com/openresty/lua-nginx-module#ngxreqget_uri_args

Here is the source: https://github.com/openresty/lua-resty-core/blob/97d2e5e08d9ddf328e33d2b5f1710a9ce3d8bbd0/lib/resty/core/request.lua#L237

PR #6325, opened 2022-02-15
feat(wasm): support getting request body

spacewander:

### Pre-submission checklist:


tzssangglass (dismissed):

bzp2010 (approved):

t/APISIX.pm
```diff
@@ -217,6 +217,13 @@ if ($version =~ m/\/apisix-nginx-module/) {
 _EOC_
 }
 
+my $a6_ngx_vars = "";
+if ($version =~ m/\/apisix-nginx-module/) {
+    $a6_ngx_vars = <<_EOC_;
```
leslie-tsang, line 222:
```suggestion
    $a6_ngx_vars .= <<_EOC_;
```
Would this be better ? If there is a new setting that needs to be applied, we don't need to change it.

t/wasm/request-body.t
```diff
@@ -0,0 +1,152 @@
...
+
+
+
+=== TEST 5: hit
+--- request
+POST /hello
+hello
+--- grep_error_log eval
+qr/request get body: \w+/
+--- grep_error_log_out
```
leslie-tsang, line 152: Is it possible to remove this ?

t/APISIX.pm
```diff
@@ -217,6 +217,13 @@ if ($version =~ m/\/apisix-nginx-module/) {
 _EOC_
 }
 
+my $a6_ngx_vars = "";
```
starsz, line 220: `a6_ngx_vars` looks strange.
Maybe `apisix_ngx_vars`  would be better?

t/APISIX.pm
spacewander: There is no need to append the str to `$a6_ngx_vars` which is empty if apisix-nginx-module is missing.

t/APISIX.pm
spacewander: There is already a `a6_ngx_directives`, so `a6_ngx_vars` is fine

t/wasm/request-body.t
spacewander: No. A empty `grep_error_log_out` means "should not grep any result with the pattern given in grep_error_log".

leslie-tsang (approved):

tzssangglass (approved):

bzp2010 (approved):

starsz (approved):

shuaijinchao (approved):

PR #6327, opened 2022-02-15
change(prometheus): replace wrong apisix_nginx_http_current_connections{state="total"} label

### LINKED ISSUE #6301: bug: some mistakes in prometheus plugin
### Issue description

The APISIX prometheus plugin uses the `/nginx_status` endpoint provided by nginx to get some internal metrics, which return the data like
```
Active connections: 291
server accepts handled requests
 16630948 16630948 31070465
Reading: 6 Writing: 179 Waiting: 106
```
According to the [nginx documentation](https://nginx.org/libxslt/en/docs/http/ngx_http_stub_status_module.html):
> The following status information is provided:
> 
> Active connections
> The current number of active client connections including Waiting connections.
> accepts
> The total number of accepted client connections.
> handled
> The total number of handled connections. Generally, the parameter value is the same as accepts unless some resource limits have been reached (for example, the worker_connections limit).
> requests
> The total number of client requests.
> Reading
> The current number of connections where nginx is reading the request header.
> Writing
> The current number of connections where nginx is writing the response back to the client.
> Waiting
> The current number of idle client connections waiting for a request.

The third number in the third line is `The total number of client requests`, but APISIX now puts this metric in `nginx_http_current_connections{label="total"}`, which is obviously wrong. 
https://github.com/apache/apisix/blob/0920d1d21fba36947b4ce13438040874c5d37610/apisix/plugins/prometheus/exporter.lua#L86-L88
https://github.com/apache/apisix/blob/0920d1d21fba36947b4ce13438040874c5d37610/apisix/plugins/prometheus/exporter.lua#L174-L206

Also the three metrics in the third line are all cumulative, should we use the counter type instead of the guage for those metrics? Which is more in line with the prometheus definition of metric types.

So I want to introduce two new metric: 
1. `nginx_http_requests_total`(counter type)
2. `nginx_http_connections_total`(counter type) with a `type` label, which has two values: `accepts`, `handled`

If there are no problems, I can complete the fix.
### Environment

- apisix version (cmd: `apisix version`): 2.10.1

### Steps to reproduce

Just enable prometheus plugin.

### Actual result

/

### Error log

/

### Expected result

_No response_

nic-6443:

Fix #6301 
### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -87,6 +87,9 @@ function _M.init()
             "Number of HTTP connections",
             {"state"})
 
+    metrics.requests = prometheus:gauge("http_requests_total",
+            "The total number of client requests")
+
     metrics.etcd_reachable = prometheus:gauge("etcd_reachable",
             "Config server etcd reachable from APISIX, 0 is unreachable")
 
@@ -199,9 +202,12 @@ local function nginx_status()
             break
         end
 
-        label_values[1] = name
-        metrics.connections:set(val[0], label_values)
-
+        if name == "total" then
+            metrics.requests:set(val[0])
+        else
+            label_values[1] = name
+            metrics.connections:set(val[0], label_values)
+        end
     end
 end
 
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -88,7 +88,7 @@ function _M.init()
             {"state"})
 
     metrics.requests = prometheus:gauge("http_requests_total",
-            "The total number of client requests")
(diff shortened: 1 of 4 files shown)


apisix/plugins/prometheus/exporter.lua
```diff
@@ -87,6 +87,9 @@ function _M.init()
             "Number of HTTP connections",
             {"state"})
 
+    metrics.requests = prometheus:gauge("http_requests_total",
+            "The total number of client requests")
```
tokers, line 91:
```suggestion
            "The total number of client requests since APISIX started")
```

tzssangglass (approved):

spacewander (approved):

Issue #6225, opened 2022-01-28
request help: install failed via source on Mac(2.12.0)


kwanhur:
### Issue description

after `make deps` successfully, but `make install` or `sudo make install` failed.

it seems no authority to mkdir dir `/usr/share/lua` . 
Installed lua with brew, so this related dir should locate at `/usr/local/Cellar/lua/5.1/share` .

```bash
➜  apisix git:(master) sudo make install
Password:
install -d /usr/local/apisix/
install -d /usr/local/apisix/logs/
install -d /usr/local/apisix/conf/cert
install conf/mime.types /usr/local/apisix/conf/mime.types
install conf/config.yaml /usr/local/apisix/conf/config.yaml
install conf/config-default.yaml /usr/local/apisix/conf/config-default.yaml
install conf/debug.yaml /usr/local/apisix/conf/debug.yaml
install conf/cert/* /usr/local/apisix/conf/cert/
install -d /usr/share/lua/5.1/apisix
install: mkdir /usr/share/lua: Operation not permitted
make: *** [install] Error 71
```

### Environment

- apisix version (cmd: `apisix version`): 2.12.0
- OS (cmd: `uname -a`): Darwin Kwanhur-MacBook-Pro.local 20.6.0 Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64 x86_64
- luarocks version, if the issue is about installation (cmd: `luarocks --version`): /usr/local/bin/luarocks 3.7.0

```bash
➜ brew info lua
lua: stable 5.4.3 (bottled)
Powerful, lightweight programming language
https://www.lua.org/
/usr/local/Cellar/lua/5.4.3 (29 files, 746.3KB) *
  Poured from bottle on 2021-09-19 at 21:07:48
From: git://mirrors.ustc.edu.cn/homebrew-core.git/Formula/lua.rb
License: MIT
```

```bash
➜ brew info luarocks
luarocks: stable 3.7.0 (bottled), HEAD
Package manager for the Lua programming language
https://luarocks.org/
/usr/local/Cellar/luarocks/3.7.0 (105 files, 715KB) *
  Poured from bottle on 2021-09-19 at 21:22:43
From: git://mirrors.ustc.edu.cn/homebrew-core.git/Formula/luarocks.rb
License: MIT
```

Chever-John:
Could it be that you don't have operation permissions for the file?
Try "ls -" :)

[...]

PR #6363, opened 2022-02-17
docs(build): add guide for install failed on MacOS

kwanhur:

Add user guide, explain failure reason when `make install` under MacOS

#6225 
### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/how-to-build.md
+++ b/docs/en/latest/how-to-build.md
@@ -117,6 +117,10 @@ Please refer to: [Installing Apache APISIX with Helm Chart](https://github.com/a
 
         Solution: set `LDAP_DIR` with `luarocks config` manually, for example `luarocks config variables.LDAP_DIR /usr/local/opt/openldap/`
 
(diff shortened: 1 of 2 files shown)


tzssangglass:
I think we don't need to add tips on building on Macos to this document. This document is not geared towards MacOS.

kwanhur:
> I think we don't need to add tips on building on Macos to this document. This document is not geared towards MacOS.

enn, maybe users face the same problem under different os, not only MacOS.

To be continue, I want to known that:

1. any need to add cognate guides?
2. where place it will be better, like [FAQ](https://github.com/apache/apisix/blob/master/docs/en/latest/FAQ.md) or other?

spacewander:
I think people can find https://github.com/apache/apisix/issues/6225 by the search engine, so there is no need to add it in the doc. And it seems more like a general Mac permission problem instead of APISIX's problem.

kwanhur:
Hope #6225 helps those trapped.

Issue #3931, opened 2021-03-29
redirect plugins add curl test for `http_to_https`


tokers:
# Improve Docs

## Please describe which part of docs should be improved or typo fixed

The [document](http://apisix.apache.org/docs/apisix/plugins/redirect) about redirect plugin contains the configuration about using `http_to_https` in redirect plugin, but it doesn't add a curl test case to show the result.

![image](https://user-images.githubusercontent.com/10428333/112805052-afe6aa80-90a7-11eb-9b4e-c19d33c9a55c.png)

## Describe the solution you'd like

We should add a curl test case to show that a request is really redirected since the scheme is `HTTP`.

kartikeysaran:
Hey ! Is this issue still unresolved ?

tzssangglass:
> 嘿 ！这个问题还没有解决吗？

PR is welcome!

leslie-tsang:
AFAIK, @kartikeysaran , not yet, would you have a try ?

kartikeysaran:
Yeah sure

tzssangglass:
hi @kartikeysaran, assigned you

leslie-tsang:
@kartikeysaran Hi there, Are there any issues that are preventing you from moving forward?

Do let me know if any help in needed.

mangoGoForward:
I tested it like this:

1. Create a route

```bash
$ curl http://127.0.0.1:9080/apisix/admin/routes/1  -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "uri": "/hello",
    "plugins": {
        "redirect": {
            "http_to_https": true
        }
    }
}'
```

2. Test.

```bash
$ curl http://127.0.0.1:9080/hello -i
HTTP/1.1 301 Moved Permanently
Date: Fri, 18 Feb 2022 07:48:32 GMT
Content-Type: text/html
Content-Length: 166
Connection: keep-alive
Location: https://127.0.0.1/hello
Server: APISIX/2.12.0

<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>openresty</center>
</body>
</html>
```

How about add the second step to the [redirect document](http://apisix.apache.org/docs/apisix/plugins/redirect)? @tzssangglass @leslie-tsang

leslie-tsang:
> I tested it like this:
> 
> 1. Create a route
> 
> ```shell
[...]

PR #6378, opened 2022-02-18
docs: add curl test for redirect plugin's `http_to_https`.

mangoGoForward:
Add curl test for redirect plugin's `http_to_https`, please see #3931.

### Pre-submission checklist:


docs/en/latest/plugins/redirect.md
```diff
@@ -124,6 +124,17 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1  -H 'X-API-KEY: edd1c9f034335f1
 }'
 ```
 
+Testing based on the above examples :
+
+```shell
+$ curl http://127.0.0.1:9080/hello -i
+HTTP/1.1 301 Moved Permanently
+...
+Location: https://127.0.0.1/hello
```
tokers, line 133: Since we use `9080` to accept HTTP traffic, should we use `9443` here?

mangoGoForward: Why should be `9443`? and the default port of `https` is `443`.
When `http_to_https` is set to true and the request is HTTP, it will be automatically redirected to HTTPS with 301 response code, and the URI will keep the same as client request.

https://github.com/apache/apisix/blob/9b98f1d13187c58195543bd48cb8b66e33aeab27/apisix/plugins/redirect.lua#L155-L166

tzssangglass: > Why should be `9443`

The default port for the APISIX proxy https protocol is 9443.

tokers: @mangoGoForward Just for the consistence of the documentation, look through the whole doc you'll find we don't use the default `80` port.

mangoGoForward: > The default port for the APISIX proxy https protocol is 9443.

@tokers See the `Location` header of response, it redirected to `https`. And I found some test cases in `t/plugin/redirect.t`:
https://github.com/apache/apisix/blob/42e84e437fa532f24523f5b38015436e1fcc013f/t/plugin/redirect.t#L431-L438

Seems we don't need to update?

tokers: Personally I think there is a bug, The generated Location header should respect the setting of SSL port, we cannot assume the SSL port is `443`.

Also, my original meaning is quite simple, just because we use `9080` as the sample port to accept HTTP traffic, for the sake for consistence, I think using `9443` is better.

mangoGoForward: Sorry, I don't know how to use `9443` in this test case.

Do we need create a upstream?

tzssangglass: This PR does not need to be modified in the test case to 9443, just use 9443 in the documentation? cc @tokers

spacewander: Ping @tokers

tokers: @tzssangglass is correct, just use `9443` in documentation is enough.

tzssangglass, line 133:
```suggestion
Location: https://127.0.0.1:9443/hello
```

tzssangglass (changes requested):

docs/zh/latest/plugins/redirect.md
```diff
@@ -115,6 +115,17 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1  -H 'X-API-KEY: edd1c9f034335f1
 }'
 ```
 
+基于上述例子的测试示例：
+
+```shell
+$ curl http://127.0.0.1:9080/hello -i
+HTTP/1.1 301 Moved Permanently
+...
+Location: https://127.0.0.1/hello
```
tzssangglass, line 124:
```suggestion
Location: https://127.0.0.1:9443/hello
```

tzssangglass (changes requested):

starsz (approved):

tzssangglass (approved):

PR #6385, opened 2022-02-20
docs: optimization curl command

xuanyanwow:

In some special env. the test curl command may get the error result.
optimization the command in docs.
``` sh
# curl "http://127.0.0.1:9080/apisix/admin/services/" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1'
{"error_msg":"failed to check token"}
curl: (6) Could not resolve host: edd1c9f034335f136f87ad84b625c8f1'
```
references： https://blog.csdn.net/u011035397/article/details/109452045

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/getting-started.md
+++ b/docs/en/latest/getting-started.md
@@ -111,7 +111,7 @@ Once the download is complete, execute the `curl` command on the host running Do
 
 ```bash
 # Note: Please execute the curl command on the host where you are running Docker.
-curl "http://127.0.0.1:9080/apisix/admin/services/" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1'
+curl "http://127.0.0.1:9080/apisix/admin/services/" -H "X-API-KEY: edd1c9f034335f136f87ad84b625c8f1"
 ```
 
 The following data is returned to indicate that Apache APISIX was successfully started:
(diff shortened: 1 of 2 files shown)


kwanhur:
`curl` examples appear everywhere in `docs`, should be optimized along with all the others.

spacewander (changes requested):
This doesn't bring any difference whether we use the single quote for the header or not.

spacewander:
Please don't trust a piece of very doubtful information copied from someone's blog. It is a waste of both your time and mine.

spacewander:
I guess you forgot to paste a quote during copy & paste, so `edd1c9f034335f136f87ad84b625c8f1` is not quoted and is treated as the target.

Please learn the difference between single quotes and double quotes via `man bash`, instead of believing an idea from an unreliable platform.

xuanyanwow:
I make sure what i have not forgot the quote, and i try to use -d or other param with single quote, only the -H working error.
And this question only show in my one computer, another is normal.
I will try to find more info and test about this question, and then take a discuss here.
Thank you all.

xuanyanwow:
Because i test in windows with docker.

Now i can make sure the question by use ` fiddler ` to get the  request info.
It is because windows'cmd cannot parse single quote ,but it's have a little diffrent .

``` 
curl -H 'X: 123'
```
it can not parse ,throw the error `curl: (6) Could not resolve host:...`

```
curl -H 'X:123'
```
it will not throw the error ,but it's not a normal http's header.

PR #6390, opened 2022-02-20
ci: split the full CI into two parts (running t/plugin and the other)

### LINKED ISSUE #5861: request help: split the full CI into two parts (running t/plugin and the other)
### Issue description

Sometimes the full CI "linux_openresty" https://github.com/apache/apisix/blob/4423509680a0c2103ce23b30a0da71735ef73e8f/.github/workflows/build.yml#L27
takes 50 minutes to complete.

Let's split it into multiple parts so the max single job time can be reduced.

We can pass an environment variable as the test file range in https://github.com/apache/apisix/blob/4423509680a0c2103ce23b30a0da71735ef73e8f/ci/linux_openresty_common_runner.sh#L86

### Environment

- OS (cmd: `uname -a`):

tzssangglass:
Closes #5861

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,13 +111,13 @@ jobs:
           ./ci/${{ matrix.os_name }}_runner.sh do_install
 
       - name: Linux Script
-      if: steps.filter.outputs.plugins == 'false'
-      env:
-        TEST_FILE_SUB_DIR: ""
-        run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
+        if: steps.filter.outputs.plugins == 'false'
+        env:
+          TEST_FILE_SUB_DIR: ""
+          run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
 
       - name: Linux Script
-      if: steps.filter.outputs.plugins == 'true'
-      env:
-        TEST_FILE_SUB_DIR: "plugin"
-        run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
+        if: steps.filter.outputs.plugins == 'true'
+        env:
+          TEST_FILE_SUB_DIR: "plugin"
+          run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
           ./ci/${{ matrix.os_name }}_runner.sh do_install
 
       - name: Linux Script
-        run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
+        run: sudo -E ./ci/${{ matrix.os_name }}_runner.sh script
 
 
   build-2:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -39,6 +39,14 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Paths filter
+        uses: ./.github/actions/paths-filter
```
spacewander, line 43: Is it possible to remove this 3rd party action?

spacewander:
> This PR doesn't solve the #5861. When non-plugin code is changed, we still need to run all the test cases.
> 
> I think we need to use matrix to run CI parallelly, and reduce the max run time.

Quoted from that issue:

> Let's split it into multiple parts so the max single job time can be reduced.

tzssangglass:
> When non-plugin code is changed, we still need to run all the test cases.

yes, only the test files in t/plugin will be run if only the files in the plugins directory are modified.

> > Let's split it into multiple parts so the max single job time can be reduced.

Are you saying try to make the files in the apisix/ directory correspond to the files in the t/ directory? For example, if we modify the files in the `apisix/core` directory, then just run the test files under `t/plugin`.

I had considered doing this, and I was concerned that there wasn't enough coverage.

.github/workflows/build.yml
tzssangglass: Then we need to implement a similar function ourselves, and I tried to find github action support for a similar function: trigger step by file modification path, but not found. I can't find the environment variable that indicates the file modification.

kwanhur:
Multi-jobs run in parallel by default, see it [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobs). Splitting into two jobs(like build and build-plugin), then they'll cover all the scope together and reduce total cost time.

- build-plugin job proves `t/plugin` cases with preserving env variable value `true` or `on` etc

Hope this helps you.

tzssangglass:
With the current split, in ci centos7 for example, `t/plugin` will take about 35 minutes and the other two will take about 20 minutes, I'm going to split `t/plugin` into two directories `t/plugin_1` and `t/plugin_2 `and run them in two jobs, what do you think? cc @spacewander @bzp2010 @shuaijinchao

.github/workflows/build.yml
```diff
@@ -104,4 +105,186 @@ jobs:
...
-        run: sudo ./ci/${{ matrix.os_name }}_runner.sh script
+        run: sudo -E ./ci/${{ matrix.os_name }}_runner.sh script
+
+
+  build-2:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - ubuntu-18.04
```
spacewander, line 116: Why not use matrix to solve the problem? We can use "test-dir" like how we use the `platform`

tzssangglass: fixed

spacewander (approved):

bzp2010 (approved):

PR #6320, opened 2022-02-14
ci(apisix test): grep search [error] from file error.log

kwanhur:
### What this PR does / why we need it:

grep search `[error]` from `logs/error.log` file instead of `cat logs/error.log` stdin

- reduce one command operation

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #6428, opened 2022-02-23
ci: add shellcheck linter

kwanhur:

use [ShellCheck](https://github.com/koalaman/shellcheck) to analysis shell scripts and gives warnings and suggestions.

intention: import shell script's quality

test case [SC2148](https://github.com/kwanhur/apisix/runs/5301576428?check_suite_focus=true)

`.shellcheckrc` to enable/disable shellcheck code(equals rule), content now comes from existed shell scripts.

- cause all of them had run successfully, so a lot of rules disable

https://github.com/apache/apisix/pull/6320#issuecomment-1039907522

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -28,3 +28,20 @@ jobs:
         . ./ci/common.sh
         export_or_prefix
         make lint
+
+  sc-lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2.4.0
+        with:
+          submodules: true
+
+      - name: Shellcheck code
+        run: |
+          scversion="latest"
+          wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+          cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
+          shellcheck --version
+          shellcheck **/*.sh
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -42,4 +42,4 @@ jobs:
           wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
           cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
           shellcheck --version
-          shellcheck **/*.sh
+          shellcheck benchmark/**/*.sh bin/apisix ci/**/*.sh t/**/*.sh utils/*.sh
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -42,4 +42,4 @@ jobs:
           wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
           cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
           shellcheck --version
-          shellcheck benchmark/*.sh bin/apisix ci/*.sh ci/pod/nacos/healthcheck/*.sh t/cli/*.sh t/plugin/grpc-web/*.sh utils/*.sh
+          git ls-files | grep ".sh$" | xargs -t shellcheck
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -42,4 +42,4 @@ jobs:
           wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
           cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
           shellcheck --version
-          shellcheck benchmark/**/*.sh bin/apisix ci/**/*.sh t/**/*.sh utils/*.sh
+          shellcheck benchmark/*.sh bin/apisix ci/*.sh ci/pod/nacos/healthcheck/*.sh t/cli/*.sh t/plugin/grpc-web/*.sh utils/*.sh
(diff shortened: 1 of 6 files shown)


spacewander (commented):
Good, let's add https://github.com/apache/apisix/pull/6320 to this PR.

.github/workflows/code-lint.yml
```diff
@@ -28,3 +28,20 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2.4.0
+        with:
+          submodules: true
+
+      - name: Shellcheck code
+        run: |
+          scversion="latest"
+          wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
```
leslie-tsang, line 44:
```suggestion
          wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
```
Why use `-J` here ?

Can it replace with `-z` ?

.github/workflows/code-lint.yml
```diff
@@ -28,3 +28,20 @@ jobs:
...
+          cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
```
leslie-tsang, line 45: can we use `apt` to install `spellcheck` ?
like
```bash
apt install -y shellcheck
```

.github/workflows/code-lint.yml
kwanhur: Because it's format is `tar.xz` not `tar.gz`

- `-J` filter the archive through xz
- `-z` filter the archive through gzip

.github/workflows/code-lint.yml
kwanhur: This way is ok to install `shellcheck`, but I'm not sure its repository updates to latest version.

1. latest version may add new check rules and it covers the before.
2. different developers use different version, the latest covers them.

The package must uniformly install from repo unless no supported, I'll improve installation.

leslie-tsang: LGTM, let keep it this way. 

Thanks for replying.

kwanhur:
mentioned PRs differ goal:

- #6320 modify search error logicality, reduce one command.
- #6428 add code(shell script) limitation.

I think they're should stay with two, one PR reaches one goal. Of cause, still need they squash into one, I'll take it.

kwanhur:
Anyway, they had been done.

ci/common.sh
```diff
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+# shellcheck disable=SC2148
```
spacewander, line 18: I think it is quite strange to add a warning and disable it later. Let's remove it.

kwanhur: SC2148(Add a shebang) rule has enabled, remove it `ci code-lint` will run failed. 
The key point is that, all the scripts have the shebang but `ci/common.sh` and `t/cli/common.sh`.

`disable` in script file, just scope the file, no influences on the other.

To fix the check failure, three solutions to choose:

1. enable SC2148 and add shellcheck disable declaration(just like now), test case [SC2148](https://github.com/kwanhur/apisix/runs/5301576428?check_suite_focus=true).
2. enable SC2148 and add a shebang in `common.sh`, but it seems not so suitable, like https://github.com/apache/apisix/pull/6340#discussion_r809653666.
3. disable SC2148, this will not find it when add a script without shebang.

Summary, solution 1 could be the better.
Hope your voice.

spacewander: Sure, but why use

> enable=SC2148

in https://github.com/apache/apisix/pull/6428/files#diff-c71502ae2fb6ee770263dbf93031c6bb1549ef6b2970b862daf24d01a1ded277R22?

Could we keep it disabled? Look like the check is enabled by you.

kwanhur: Had disable it.

Actually, all the rules enable by default. Here, I mark it obviously.

kwanhur:
[build (ubuntu-18.04, linux_openresty_1_17, t/plugin)](https://github.com/apache/apisix/runs/5355589130?check_suite_focus=true#logs) runs failed, it's not related to `shellcheck`, please trigger re-run the job.

from the logs, seems that grpc server launch abnormally.

.github/workflows/code-lint.yml
```diff
@@ -28,3 +28,20 @@ jobs:
...
+          shellcheck --version
+          shellcheck **/*.sh
```
spacewander, line 47: Let's limit the scope. We don't want to lint the code from git submodule.

kwanhur: Ok, I got the meaning that checkout repo without submodule, so it won't lint the code from git submodule.

kwanhur: > Let's limit the scope. We don't want to lint the code from git submodule.

Done.

spacewander: > shellcheck **/*.sh

? It is still linting all the shell files in this repo, including the one from submodule...

.github/workflows/code-lint.yml
```diff
@@ -28,3 +28,18 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v2.4.0
+
+      - name: Shellcheck code
+        run: |
+          scversion="latest"
+          wget -O- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+          cp -av "shellcheck-${scversion}/shellcheck" /usr/local/bin/
+          shellcheck --version
+          shellcheck benchmark/*.sh bin/apisix ci/*.sh ci/pod/nacos/healthcheck/*.sh t/cli/*.sh t/plugin/grpc-web/*.sh utils/*.sh
```
spacewander, line 45: Would be better to use `git ls-files | grep "\.sh$" | xargs`

spacewander (approved):

.github/workflows/code-lint.yml
kwanhur: Yep, it's better. `xargs -t shellcheck` echo the command to be executed, it'll show which files to check.

spacewander (approved):

tzssangglass (approved):

PR #6487, opened 2022-03-02
test(http-logger): fetch request and response body test case

shuaijinchao:
FIX https://the-asf.slack.com/archives/CUC5MN17A/p1646136677114399

### Pre-submission checklist:

### DIFF
--- a/t/lib/server.lua
+++ b/t/lib/server.lua
@@ -46,6 +46,14 @@ local function inject_headers()
     end
 end
 
+local function str_split(str, reps)
+    local str_list = {}
+    string.gsub(str, '[^' .. reps .. ']+', function(w)
+        table.insert(str_list, w)
+    end)
+    return str_list
+end
+
 
 function _M.hello()
     ngx.req.read_body()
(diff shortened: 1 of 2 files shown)


t/lib/server.lua
```diff
@@ -46,6 +46,14 @@ local function inject_headers()
     end
 end
 
+local function str_split(str, reps)
```
spacewander, line 49: Let's add the mock API in the test file like:
https://github.com/apache/apisix/blob/4feab8e6bf1cf8ffd16440848b693ce14a792fbc/t/plugin/clickhouse-logger.t#L39

It's not commonly used.

shuaijinchao: done

t/plugin/http-logger2.t
```diff
@@ -32,6 +32,64 @@ add_block_preprocessor(sub {
...
+                local query = args.query or nil
+                ngx.req.read_body()
+                local body = ngx.req.get_body_data()
+
+                if query then
+                    if type(query) == "string" then
+                        query = {query}
+                    end
+
+                    local data, err = require("cjson").decode(body)
```
starsz, line 65: Hi, Why do we need to decode the body?

tzssangglass (approved):

t/plugin/http-logger2.t
shuaijinchao: The log format data is JSON, and we need to get the information in the structure.

t/plugin/http-logger2.t
```diff
@@ -32,6 +32,64 @@ add_block_preprocessor(sub {
...
+                        local val
+                        for j = 1, #fields do
+                            local key = fields[j]
+                            if j == 1 then
+                                val = data[key]
+                            else
+                                val = val[key]
+                            end
+                        end
+                        ngx.log(ngx.ERR ,query[i], ":", val)
```
spacewander, line 81: Better to use WARN log level

t/plugin/http-logger2.t
```diff
@@ -141,3 +199,86 @@ GET /opentracing
...
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:12001": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/http-logger/test"
+                }]],
+                [[{
```
spacewander, line 234: We should remove the response body check

t/plugin/http-logger2.t
starsz: Got it.

starsz (approved):

tzssangglass (approved):

spacewander (approved):

PR #6500, opened 2022-03-03
fix(tracing): should trace request stopped by extern plugin

spacewander:

### Pre-submission checklist:


spacewander:
I don't change the skywalking/opentracing because increasing the priority will make the body_filter call earlier.

Can we remove the `finish` in the body_filter and only finish the span in the log phase?
@dmsolr @yangxikun
Please guide me the right way.

spacewander:
Or maybe we can switch to an onion model for the filter?

dmsolr:
I think finish() should be after `ext-plugin`, but start() should be before `ext-plugin`.
Moving the finish() to the log phase represents increasing the response time in trace view.

the trace includes 2 spans in APISIX, 
the first span, entry span, is over the whole request. It means from the request accepted to the response returned.
the 2nd span, exit span, is only from the sending request to accepting the response.
（Now, those spans start and stop at the same time.）

tzssangglass (approved):

shuaijinchao (approved):

spacewander:
I will open another issue to trace https://github.com/apache/apisix/pull/6500#issuecomment-1058748024

PR #2898, opened 2020-11-30
feat: allow create consumers with multiple auth plugins

spacewander:
Fix #2850

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #6502, opened 2022-03-03
feat: rerun rewrite phase for newly added plugins in consumer

### LINKED ISSUE #6405: request help: The plugins that act on the specified consumer do not take effect
### Issue description

origin mail: https://lists.apache.org/thread/o0zh8pddh50ytcf99g8jr5wnxbf9p9t3

User Description: 
```
And when I use the same config :

\"proxy-rewrite\": {
      \"headers\": {
        \"X-USER-ID\": \"USER1-ID\"
      }
    }

In the route config it works :

curl --location --request PUT
"http://$APISIX_URL:$APISIX_PORT/apisix/admin/routes/api-v1" \
--header "X-API-KEY: $APISIX_API_ADMIN_KEY" \
--header "Content-Type: application/json" \
--data-raw "{
  \"uri\": \"/v1/*\",
  \"host\": \"$APISIX_HOST\",
  \"plugins\": {
    \"key-auth\": {},
    \"consumer-restriction\": {
        \"whitelist\": [
            \"user1\",
            \"user2\"
        ]
    },
    \"proxy-rewrite\": {
      \"headers\": {
        \"X-USER-ID\": \"USER1-ID\"
      }
    }
  },
  \"service_id\": \"api-v1\"
}"
```

and

```
I mean that when I configure the plugin proxy-rewrite on the routes I have
the header X-USER-ID set with the value USER1-ID (everything is fine) and
when I configure it on the consumer user1 and I call the endpoint with the
apikey of user1, the header X-USER-ID does not exist.
```

### Environment

- OS (cmd: `uname -a`):

tzssangglass:
Closes #6405

### Pre-submission checklist:


tzssangglass:
only call core.consumer.attach_consumer is considered to be an auth class plugin.

ref: https://github.com/apache/apisix/pull/2898

apisix/init.lua
```diff
@@ -454,8 +454,15 @@ function _M.http_access_phase()
 
             if changed then
                 api_ctx.matched_route = route
+                local runned_plugins = core.table.deepcopy(api_ctx.plugins)
                 core.table.clear(api_ctx.plugins)
-                api_ctx.plugins = plugin.filter(api_ctx, route, api_ctx.plugins)
+                local unrunn_plugins
```
spacewander, line 459: Maybe we can add a flag in https://github.com/apache/apisix/blob/c4229d15efe8cac6677f8124cf92d3512d1b0180/apisix/plugin.lua#L505 so that we can save allocation?

tzssangglass: update

apisix/init.lua
```diff
@@ -456,6 +456,8 @@ function _M.http_access_phase()
                 api_ctx.matched_route = route
                 core.table.clear(api_ctx.plugins)
                 api_ctx.plugins = plugin.filter(api_ctx, route, api_ctx.plugins)
+                -- rerun rewrite phase for newly added plugins in consumer
+                plugin.rerun_plugins_of_consumer(api_ctx.plugins, api_ctx)
```
tokers, line 460: Currently, plugins in APISIX can only be run once, but after introducing this method, some plugins (in rewrite phase) can be run twice.

This inconsistence will confuse the users. I think either we only allow running each plugin for just one time or we need to redesign the plugin system so that plugins can be run multiple times.

tzssangglass: In fact, no plug-in will run twice in this case. The reason for this misunderstanding is that

1. APISIX runs the rewrite phase of plugins on route first;
2. APISIX handles the logic for consumer authentication based on the auth plugins on the route;
3. then APISIX merges the plugins on consumer and route; 
4. this causes the rewrite phase to be missed for some non-auth plugins configured on the consumer.

apisix/plugin.lua
```diff
@@ -365,17 +365,7 @@ function _M.filter(ctx, conf, plugins, route_conf, runned_plugins)
...
-    if runned_plugins and type(runned_plugins) == "table" then
-        runned_plugins_names = core.table.new(#runned_plugins / 2, 0)
-        for i = 1, #runned_plugins, 2 do
-            core.table.insert(runned_plugins_names, runned_plugins[i].name)
-        end
-        unrunn_plugins = core.table.new(4, 0)
-    end
-
-    local route_plugin_conf = route_conf and route_conf.value and route_conf.value.plugins
+    local route_plugin_conf = route_conf and route_conf.value.plugins
```
shuaijinchao, line 368:
```suggestion
    local route_plugin_conf = route_conf and route_conf.value and route_conf.value.plugins
```
Keeping the old way should be more rigorous.

tzssangglass: hi @shuaijinchao what you reviewed is already the previous version.

apisix/plugin.lua
```diff
@@ -503,6 +503,11 @@ local function merge_consumer_route(route_conf, consumer_conf)
         end
```
spacewander, line 504: we can use `if new_route_conf.value.plugins[name] == nil then conf._from_consumer = true`?

apisix/plugin.lua
```diff
@@ -805,4 +810,32 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
 end
 
 
+function _M.rerun_plugins_of_consumer(plugins, api_ctx)
+    for i = 1, #plugins, 2 do
+        -- no need to rerun the auth plugins
+        if plugins[i + 1]["_from_consumer"] and plugins[i].type ~= "auth" then
```
spacewander, line 816: Would be better to merge it in https://github.com/apache/apisix/blob/492f782d9fae719b3a20e0d65c275962bcdcaab9/apisix/plugin.lua#L736?

apisix/init.lua
tokers: Thanks for the explanation.

apisix/plugin.lua
tzssangglass: tried, Test 2 will fail. We need to merge the `new_route_conf.value.plugins["proxy-rewrite"],  it will not be nil.

apisix/plugin.lua
tzssangglass: I had thought about it, but `if plugins[i + 1]["_from_consumer"] and plugins[i].type ~= "auth" then` is difficult to port. Unless we add a flag or something like that. This makes run_plugin look even more complicated.

Another way is as follows:

```
    for i = 1, #plugins, 2 do
        -- no need to rerun the non-auth plugins
        if plugins[i + 1]["_from_consumer"] and plugins[i].type ~= "auth" then
            local new_plugins = core.table.new(2, 0)
            core.table.insert(new_plugins, plugins[i])
            core.table.insert(new_plugins, plugins[i + 1])
            _M.run_plugin("rewrite", new_plugins, api_ctx)
        end
    end
    return api_ctx
```

this adds some table overhead.

tzssangglass: The main reason is that the `plugins` here are used in the `api_ctx` in order to reduce the table overhead, so use the judgment condition.

apisix/plugin.lua
spacewander: What about this change?
```
diff --git apisix/plugin.lua apisix/plugin.lua
index d66ec045..190b0833 100644
--- apisix/plugin.lua
+++ apisix/plugin.lua
@@ -502,12 +502,11 @@ local function merge_consumer_route(route_conf, consumer_conf)
             new_route_conf.value.plugins = {}
         end

-        new_route_conf.value.plugins[name] = conf
-
-        if (route_conf and route_conf.value and route_conf.value.plugins)
-                and not route_conf.value.plugins[name] then
-            new_route_conf.value.plugins[name]["_from_consumer"] = true
+        if new_route_conf.value.plugins[name] == nil then
+            conf._from_consumer = true
         end
+
+        new_route_conf.value.plugins[name] = conf
     end

     core.log.info("merged conf : ", core.json.delay_encode(new_route_conf))
```

apisix/plugin.lua
spacewander: Maybe we can add another fake phase (rewrite_in_consumer) in `run_plugin` so we don't need to duplicate the conditions?

tzssangglass: update

apisix/plugin.lua
tzssangglass: yes, update

spacewander (approved):

leslie-tsang (approved):

shuaijinchao (approved):

PR #6509, opened 2022-03-04
change: backport server-info to 2.10

zaunist:

Currently, server-info is refactored, so we should backport to 2.10 release. Because 2.10 is a LTS version.

### Pre-submission checklist:

### DIFF
--- a/apisix/core/etcd.lua
+++ b/apisix/core/etcd.lua
@@ -207,6 +207,7 @@ local function set(key, value, ttl)
             return nil, grant_err
         end
         res, err = etcd_cli:set(prefix .. key, value, {prev_kv = true, lease = data.body.ID})
+        res.body.lease_id = data.body.ID
     else
         res, err = etcd_cli:set(prefix .. key, value, {prev_kv = true})
     end
@@ -357,6 +358,21 @@ function _M.server_version()
     end
 
     return etcd_cli:version()
+end
+
+
+function _M.keepalive(id)
+    local etcd_cli, _, err = new()
+    if not etcd_cli then
+        return nil, err
+    end
+
+    local res, err = etcd_cli:keepalive(id)
+    if not res then
+        return nil, err
+    end
+
+    return res, nil
 end
 
 
--- a/apisix/core/etcd.lua
+++ b/apisix/core/etcd.lua
@@ -372,7 +372,7 @@ function _M.keepalive(id)
         return nil, err
     end
 
-    return res, nil
+    return res
 end
 
 
--- a/t/plugin/server-info.t
+++ b/t/plugin/server-info.t
@@ -34,6 +34,20 @@ no_long_string();
 no_root_location();
 no_shuffle();
 
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if (!defined $block->request) {
+        $block->set_value("request", "GET /t");
+    }
+
+    if (!defined $block->error_log && !defined $block->no_error_log) {
+        $block->set_value("no_error_log", "[error]");
+    }
+
+    $block;
+});
+
 run_tests;
 
 __DATA__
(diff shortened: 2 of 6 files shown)


zaunist:
cc @tzssangglass @spacewander

nic-chen (approved):

apisix/core/etcd.lua
```diff
@@ -360,4 +361,19 @@ function _M.server_version()
...
+    if not etcd_cli then
+        return nil, err
+    end
+
+    local res, err = etcd_cli:keepalive(id)
+    if not res then
+        return nil, err
+    end
+
+    return res, nil
```
membphis, line 375: remove `, nil`

tzssangglass (approved):

apisix/core/etcd.lua
zaunist: If I don't return nil, how can I be sure there is no err?

membphis: the default value is `nil` in this case

t/plugin/server-info.t
```diff
@@ -65,58 +79,12 @@ location /t {
...
-}
---- request
-GET /t
---- response_body
-integral
---- no_error_log
-[error]
---- error_log
-timer created to report server info, interval: 60
+qr/^{"boot_time":\d+,"etcd_version":"[\d\.]+","hostname":"[a-zA-Z\-0-9\.]+","id":[a-zA-Z\-0-9]+,"version":"[\d\.]+"}$/
```
moonming, line 83: I am not sure why you remove test cases when you add many codes?

zaunist: The up_time field has been removed and this test is meaningless here.

apisix/core/etcd.lua
zaunist: I checked the other methods in etcd.lua, push, delete, set, etc. all return nil, and I think I'm correct in returning nil here. Shouldn't the code style of a repository be uniform? Even if the default value is nil, it should be explicitly shown in the code, so that the logic of the code is clearer.

t/plugin/server-info.t
moonming: So please add more test cases for the added codes

zaunist: In this change, we are using `keepalive` to replace the `put` operation, and testing keepalive requires verification of revision changes in etcd. However, since test-nginx reloads nginx every time you run a test, this may cause the revision to change, and I have not found a way to verify the keepalive.

apisix/core/etcd.lua
membphis: `, nil` is useless, so we should remove it and keep the code simpler

zaunist: I checke APISIX repo, other file has the same situation. Example:
- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/balancer/ewma.lua#L93

- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/balancer/ewma.lua#L100
- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/core/schema.lua#L56
- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/core/utils.lua#L137
- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/plugins/basic-auth.lua#L112
- https://github.com/apache/apisix/blob/cb2ba6819f72d75053f8f95aa781975dca2223fd/apisix/plugins/openid-connect.lua#L159

There are more places where there is a similar situation, where nil is returned when err equals nil, do we need to be uniform for those as well?

membphis: we can create a new issue, fix them later

zaunist: Got it.

membphis (approved):
LGTM

zaunist:
I have create an issue for uniform code style https://github.com/apache/apisix/issues/6541. So we can do it in other PRs.

PR #6512, opened 2022-03-04
feat: support google reCAPTCHAP

Dog-Lee:

Google reCAPTCHA is a popular human-identify service in the world. It protects API from spam and abuse.

It will be great if it's integrated into official plugins. And This helps people who love APISIX use reCAPTCHA out of the box without any coding.

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/recaptcha.lua
+++ b/apisix/plugins/recaptcha.lua
@@ -1,0 +1,122 @@
+local radix = require("resty.radixtree")
+local core = require("apisix.core")
+local http = require("resty.http")
+
+local schema = {
+    type = "object",
+    properties = {
+        recaptcha_secret_key = { type = "string" },
+        apis = {
+            type = "array",
+            items = {
+                type = "object",
+                properties = {
+                    path = { type = "string" },
+                    methods = { type = "array", items = { type = "string" }, minItems = 1 },
+                    param_from = { type = "string", default = "header", enum = { "header", "query" } },
+                    param_name = { type = "string", default = "captcha" },
+                }
+            },
+            minItems = 1
+        },
+        response = {
+            type = "object",
+            properties = {
+                content_type = { type = "string", default = "application/json; charset=utf-8" },
+                status_code = { type = "number", default = 400 },
+                body = { type = "string", default = '{"message": "invalid captcha"}' }
+            }
+        },
+
+    },
+    additionalProperties = false,
+    required = { "recaptcha_secret_key" },
+}
+
+local recaptcha_url = "https://www.recaptcha.net"
+
+local _M = {
+    version = 0.1,
+    priority = 700,
+    name = "recaptcha",
+    schema = schema,
+}
+
+function _M.check_schema(conf, schema_type)
+    return core.schema.check(schema, conf)
+end
+
+local function build_radixtree(apis)
+    local items = {}
+    for _, api in ipairs(apis) do
+        local item = {
+            paths = { api.path },
+            methods = api.methods,
+            metadata = api,
+        }
+        table.insert(items, item)
+    end
+    return radix.new(items)
+end
+
+local function find_api(request, apis)
+    local rx = build_radixtree(apis)
+    return rx:match(request.path, { method = request.method })
+end
+
+local function retrieve_captcha(ctx, api)
+    local captcha
+    if api.param_from == "header" then
+        captcha = core.request.header(ctx, api.param_name)
+    elseif api.param_from == "query" then
+        local uri_args = core.request.get_uri_args(ctx) or {}
+        captcha = uri_args[api.param_name]
+    end
+    return captcha
+end
+
+function _M.access(conf, ctx)
+    local path = ctx.var.uri
+    local method = core.request.get_method()
+
+    core.log.debug("path: ", path, ", method: ", method, ", conf: ", core.json.encode(conf))
+
+    local api = find_api({ path = path, method = method }, conf.apis)
+    if not api then
+        return
+    end
+
+    core.log.debug("api found: ", core.json.encode(api))
+
+    local invalid_captcha = true
+    local captcha = retrieve_captcha(ctx, api)
+    if captcha ~= nil and captcha ~= "" then
+        local httpc = http.new()
+        local secret = conf.recaptcha_secret_key
+        local remote_ip = core.request.get_remote_client_ip(ctx)
+        local res, err = httpc:request_uri(recaptcha_url .. "/recaptcha/api/siteverify", {
+            method = "POST",
+            body = "secret=" .. secret .. "&response=" .. captcha .. "&remoteip=" .. remote_ip,
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+            },
+            ssl_verify = false
+        })
+        if err then
+            core.log.warn("request faield: ", err)
+            return core.response.exit(500), err
+        end
+        core.log.debug("recaptcha veirfy result: ", res.body)
+        local recaptcha_result = core.json.decode(res.body)
+        if recaptcha_result.success == true then
+            invalid_captcha = false
+        end
+    end
+
+    if invalid_captcha then
+        core.response.set_header("Content-Type", conf.response.content_type)
+        return core.response.exit(conf.response.status_code, conf.response.body)
+    end
+end
+
+return _M


membphis (commented):
missing test cases

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,122 @@
...
+        }
+        table.insert(items, item)
+    end
+    return radix.new(items)
+end
+
+local function find_api(request, apis)
+    local rx = build_radixtree(apis)
+    return rx:match(request.path, { method = request.method })
+end
```
bzp2010, line 65: Why would you want to manage the api manually through your plugin? With all due respect, this does not match the design of APISIX.

`Route` should be set up and plugins enabled for them first, rather than having your global plugins manage the routes.

Dog-Lee: Good question. 
This is designed on purpose. In enterprise architecture, you may have hundreds of microservice in the backend. the api gateway dispatches the requests to those services based on the prefix of URL.

For example
```
/user/* -> user service
/book/* -> book service
...
```
Let's say if every service has one or two API need to be verified by recaptcha. If we try to add recaptcha plugin in every service/route, I think it will end up a mess, and it's not good for maintaining the plugin(or secret key in the plugin) in one place. 

perhaps we can add an option property to verify every request once the plugin is executed. just like the other plugins.

moonming:
Can this plugin be used with the rate limit plugin?
I feel like this would be a good example of plugin orchestration

Dog-Lee:
> Can this plugin be used with the rate limit plugin? I feel like this would be a good example of plugin orchestration

@moonming The priority of limit-* is 1001-1003 and the recaptcha is 700(I'm not sure if it's appropriate). I think recaptcha plugin will be executed after rate limit plugins. So every request in recaptcha should be qualified for the quota of rete limiting.

Dog-Lee:
> missing test cases

@membphis Have some trouble for write test cases. Google reCAPTCHA is aimed at protecting API from bots, so it's almost impossible to obtain a valid recaptcha token from google by programing. It's kind of a paradox. What can I do is only writing the failure test cases

nic-6443:
> > missing test cases
> 
> @membphis Have some trouble for write test cases. Google reCAPTCHA is aimed at protecting API from bots, so it's almost impossible to obtain a valid recaptcha token from google by programing. It's kind of a paradox. What can I do is only writing the failure test cases

FYI: https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
We can save some credential information in the secret of Github action, so that it is private to CI.
And how to test for failure scenarios can be found here: https://stackoverflow.com/questions/43397678/is-it-possible-to-force-fail-a-recaptcha-v2-for-testing-purposes-i-e-pretend

spacewander:
> We can save some credential information in the secret of GitHub action, so that it is private for CI.

Is it safe if we need to run the CI for pull requests? Attackers might update the test script and send the credential to their servers.

apisix/plugins/recaptcha.lua
bzp2010: I think this is creating a model that is rarely used and doesn't fit the existing concept and needs @spacewander to be looked at.

nic-6443:
> > We can save some credential information in the secret of GitHub action, so that it is private for CI.
> 
> Is it safe if we need to run the CI for pull requests? Attackers might update the test script and send the credential to their servers.

Makes sense, this may help us deal with this security risk: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-codeowners-to-monitor-changes

Dog-Lee:
Hey guys. I've submitted `recaptcha.t` file. Can somebody help to review it?

moonming:
> Hey guys. I've submitted `recaptcha.t` file. Can somebody help to review it?

great work 👍

membphis:
> Hey guys. I've submitted `recaptcha.t` file. Can somebody help to review it?

You can look at the output of CI, many fail. You need to fix them at first, if you need help, pls let us know ^_^

apisix/plugins/recaptcha.lua
spacewander: I prefer to add an expression from lua-resty-expr to filter out the requests that need to be reCAPTCHA, instead of using a mini router matcher (which is an anti-pattern).

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,147 @@
...
+local core = require("apisix.core")
+local http = require("resty.http")
+
+local ipairs = ipairs
+local table = table
+
+local schema = {
+    type = "object",
+    properties = {
+        recaptcha_secret_key = { type = "string" },
```
spacewander, line 27: let's remove `recaptcha_` in the conf of recaptcha plugin.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,147 @@
...
+            type = "object",
+            properties = {
+                content_type = { type = "string", default = "application/json; charset=utf-8" },
+                status_code = { type = "number", default = 400 },
+                body = { type = "string", default = '{"message": "invalid captcha"}' }
+            }
+        },
+
+    },
+    additionalProperties = false,
```
spacewander, line 55: Need to remove this option.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,147 @@
...
+            method = "POST",
+            body = "secret=" .. secret .. "&response=" .. captcha .. "&remoteip=" .. remote_ip,
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+            },
+            ssl_verify = false
+        })
+        if err then
+            core.log.error("request failed: ", err)
+            return 500
```
spacewander, line 130:
```suggestion
            return 503
```

apisix/plugins/recaptcha.lua
spacewander: Actually, it isn't a problem as we can configure the basic plugins in service and configure reCAPTCHA in route.

apisix/plugins/recaptcha.lua
Dog-Lee: I think here is more like an Internal server error(500) rather than Service Unavailable(503). The recaptcha layer is part of the request, and 500 seems will be more reasonable for the user.

What do you think? 

BTW, Do you think it's a good practice to add a failure tolerance option to let the request continue even recaptcha API is unreachable?

apisix/plugins/recaptcha.lua
Dog-Lee: OK

apisix/plugins/recaptcha.lua
Dog-Lee: Sure.

apisix/plugins/recaptcha.lua
Dog-Lee: @spacewander I just dig into lua-resty-expr document, and found that it seems only support arg_xxx now(please correct me if I'm wrong). Usually, the recaptcha API depends on the path rather than arg.

And yes, this is an anti-pattern. I have discussed this with @shuaijinchao before. To match APISIX pattern, I can add a match_all(boolean, default true) property to verify every request once the plugin is executed. Just like the other plugins.

What do you think?

apisix/plugins/recaptcha.lua
Dog-Lee: Done.

apisix/plugins/recaptcha.lua
Dog-Lee: Done.

apisix/plugins/recaptcha.lua
Dog-Lee: > Actually, it isn't a problem as we can configure the basic plugins in service and configure reCAPTCHA in route.

Sorry, any example of this?

apisix/plugins/recaptcha.lua
leslie-tsang: > BTW, Do you think it's a good practice to add a failure tolerance option to let the request continue even recaptcha API is unreachable?

IMO, I afraid it's not good practice for this scenario, user should remove the reCAPTCHA plugin instead of tolerate the failure.

Dog-Lee: To make sure we are on the same page, something needs to be clarified. Actually, the behavior of manually removing reCAPTCHA plugin or automatically failure tolerated is just the same. And we are using HTTP to the verify captcha token. Since the network instability, sometimes it fails. the users might won't be able to remove this plugin in time.

Dog-Lee: > IMO, I afraid it's not good practice for this scenario, user should remove the reCAPTCHA plugin instead of tolerate the failure.

To make sure we are on the same page, something needs to be clarified. Actually, the behavior of manually removing reCAPTCHA plugin or automatically failure tolerated is just the same. And we are using HTTP to the verify captcha token. Since the network instability, sometimes it fails. the users might not be able to remove this plugin in time. Just like fault tolerant in the rate-limiting scene.

Dog-Lee:
Hey guys, this plugin has been refactored. 
Have been discussed with @membphis. In the current stages, we all agree with keeping this plugin as simple and easy as possible to be used. And we'll receive feedback from the community and then decide what features will be added in the next stage.

apisix/plugins/recaptcha.lua
shuaijinchao: > I think here is more like an Internal server error(500) rather than Service Unavailable(503). The recaptcha layer is part of the request, and 500 seems will be more reasonable for the user.
> 
hi @Dog-Lee When responding with a 500 error, it is intercepted by Nginx's default error_page, so a 503 response would be more appropriate.
Check it out here https://github.com/apache/apisix/blob/c84158c40f1bc3eeb0683310a5ae5e04cce275a9/apisix/cli/ngx_tpl.lua#L315

Dog-Lee: @shuaijinchao Thanks. Then I have no concern about this now.

membphis:
> Have been discussed with @membphis. In the current stages, we all agree with keeping this plugin as simple and easy as possible to be used. And we'll receive feedback from the community and then decide what features will be added in the next stage.

nice ^_^

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+done
+--- no_error_log
+[error]
+
+== TEST 2: invalid secret_key
```
membphis, line 58: need 3 blank lines between different test cases

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,241 @@
...
+                           },
+                           "uri": "/login"
+                   }]=]
+                   )
+               if code >= 300 then
+                   ngx.status = code
+                   ngx.say(body)
+               end
+
+                code, body = t('/apisix/admin/routes/2',
```
membphis, line 162: bad indentation

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,241 @@
...
+--- request
+POST /login
+--- error_code: 400
+--- response_headers
+Content-Type: application/json; charset=utf-8
+--- response_body
+{"message":"invalid captcha"}
+--- no_error_log
+[error]
+
```
membphis, line 212: remove this useless line

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,101 @@
...
+        local httpc = http.new()
+        local secret = conf.secret_key
+        local remote_ip = core.request.get_remote_client_ip(ctx)
+        local res, err = httpc:request_uri(recaptcha_url .. "/recaptcha/api/siteverify", {
+            method = "POST",
+            body = "secret=" .. secret .. "&response=" .. captcha .. "&remoteip=" .. remote_ip,
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+            },
+            ssl_verify = false
```
shuaijinchao, line 80: It can be set into the plugin configuration as optional configuration.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,101 @@
...
+
+    core.log.debug("path: ", path, ", method: ", method, ", conf: ", core.json.encode(conf))
+
+    local invalid_captcha = true
+    local captcha = retrieve_captcha(ctx, conf)
+    if captcha ~= nil and captcha ~= "" then
+        local httpc = http.new()
+        local secret = conf.secret_key
+        local remote_ip = core.request.get_remote_client_ip(ctx)
+        local res, err = httpc:request_uri(recaptcha_url .. "/recaptcha/api/siteverify", {
```
shuaijinchao, line 74: It should be better if the URI is set into the plugin configuration, if the authentication service API is updated, we don't need to modify the code to configure it.

Dog-Lee: In my experience, recaptcha is a very stable service. If we provide customize URL, I think it might add a potential security breach that allows the bad guy to change the URL and mock response, and the customer might lose their profit eventually.

Let's move it into the next phase. : )

apisix/plugins/recaptcha.lua
Dog-Lee: https://github.com/apache/apisix/pull/6512#discussion_r823289710

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,242 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+log_level('debug');
```
spacewander, line 19: Is it necessary to use debug log level?

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,242 @@
...
+                }
+            })
+            if not ok then
+                ngx.say(err)
+            end
+
+            ngx.say("done")
+        }
+    }
+--- request
```
spacewander, line 51: Let's use this script to simplify the duplicate fields:
```
#!/usr/bin/env python
# coding: utf-8
# Usage: ./simpify.py $test_filename
import sys
with open(sys.argv[1]) as f:
    lines = [l.rstrip() for l in f.readlines()]
newlines = []
for i, l in enumerate(lines):
    if ((l == "GET /t" and lines[i-1] == "--- request") or
        (l == "[error]" and lines[i-1] == "--- no_error_log")):
        newlines = newlines[:-1]
        continue
    newlines.append(l)
res = "\n".join(newlines)
# print(res)
with open(sys.argv[1], "w") as f:
   f.write(res + '\n') 
```

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,242 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+log_level('debug');
+repeat_each(1);
+no_long_string();
+no_root_location();
```
spacewander, line 22: Let's add 
```
add_block_preprocessor(sub {
    my ($block) = @_;

    if (!$block->request) {
    }

    if (!$block->no_error_log && !$block->error_log) {
    }
});
```
to remove the duplicate fields.

spacewander:
@nic-6443 
Would you like to take a look? Thanks!

t/plugin/recaptcha.t
Dog-Lee: Sorry. Can you explain more? Anything I need to update?

apisix/plugins/recaptcha.lua
shuaijinchao: I think whether `ssl_verify` needs to be verified, the choice should be left to the user.

apisix/plugins/recaptcha.lua
shuaijinchao: @spacewander What do you think about this question?

nic-6443:
> @nic-6443 Would you like to take a look? Thanks!

OK.

t/plugin/recaptcha.t
nic-6443: What he means is that there are duplicate fields in these test cases, e.g.
```
--- request
GET /t
--- no_error_log
[error]
```
You can use the script provided above to remove these duplicate fields, and then use `add_block_preprocessor` to take care of them.

See:
https://github.com/apache/apisix/blob/0c2ba4bca6192fc200a34ff7b7bebf44a079f0df/docs/en/latest/internal/testing-framework.md?plain=1#L47-L63

apisix/plugins/recaptcha.lua
nic-6443: I don't think there's any security risk in supporting the configuration of `recaptcha_url`, because if this "bad guy" can already maliciously modify your gateway configuration, he could do something even worse than mock recaptcha. 

APISIX has a lot of plugins that need to interact to external services, and it is normal to configure the address of external services. I looked at Google's documentation, the official has provided at least two connection addresses for recaptcha: "www.google.com" and "www.recaptcha.net", it is not clear whether the enterprise version will have other connection addresses. So I think it is necessary to have configurable connection addresses.

apisix/plugins/recaptcha.lua
nic-6443: > [#6512 (comment)](https://github.com/apache/apisix/pull/6512#discussion_r823289710)

Instead, `ssl_verify = false` is a real security risk, which could lead to man-in-the-middle attacks.

nic-6443 (commented):
I read through the reCAPTCHA documentation and there should be two versions of the API currently: v2(Verify requests with challenge) and v3(Verify requests with a score). 
What we are implementing here is supposed to be the v2 version, so would it be better to clarify this in the documentation?

Dog-Lee:
> I read through the reCAPTCHA documentation and there should be two versions of the API currently: v2(Verify requests with challenge) and v3(Verify requests with a score). What we are implementing here is supposed to be the v2 version, so would it be better to clarify this in the documentation?

Hi, @nic-6443. You are right. This plugin is for v2, and I'd modified the document.  Please check it out and let me know if needs to be improved.

Dog-Lee:
Hi guys. Any progress on this PR? When this PR can be merged?

moonming:
> 

CI has not passed, do you need help with this?

Dog-Lee:
```
Failed test 't/plugin/api-breaker.t TEST 19: hit route 20 times, confirm the breaker time - pattern "phase_func(): breaker_time: 2" should match a line in error.log (req 0)'
```

The CI output is confusing. I don't think it's related to the recaptcha. Please correct me if I'm wrong.

apisix/plugins/recaptcha.lua
nic-6443: @Dog-LeeWant to reconsider this review? I think it makes sense to make the server-side address configurable.

Dog-Lee: https://developers.google.com/recaptcha/docs/faq?hl=en#can-i-use-recaptcha-globally
> please use "www.recaptcha.net" in your code in circumstances when "www.google.com" is not accessible.

Due to the GWF, the Chinese server might not be able to connect with www.google.com(and it reproduces in my local env). I think this might be the reason why it has www.recaptcha.net domain.

@nic-6443 I would say it's not a must-have feature in the current stage, and I can add it in the next version. Can we? : )

nic-6443: ok

nic-6443 (dismissed):

apisix/plugins/recaptcha.lua
spacewander: @nic-6443 
Why do you approve this PR even the author doesn't resolve your comment?

Dog-Lee: @nic-6443 @spacewander I have just pushed a commit to support customize `ssl_verify` and default to `true`.  Thanks for pointing out your concern.

nic-6443: > @nic-6443 Why do you approve this PR even the author doesn't resolve your comment?

Sorry, my bad.

t/plugin/recaptcha.t
nic-6443: We can optimize this later, do you think it's OK? @spacewander

nic-6443 (dismissed):

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+            method = "POST",
+            body = "secret=" .. secret .. "&response=" .. captcha .. "&remoteip=" .. remote_ip,
```
spacewander, line 77: Let's use `ngx.encode_args` to build the body

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+            },
+            ssl_verify = conf.ssl_verify
+        })
+        if err then
+            core.log.error("request failed: ", err)
+            return 503
+        end
+        core.log.debug("recaptcha veirfy result: ", res.body)
+        local recaptcha_result = core.json.decode(res.body)
```
spacewander, line 88:
```suggestion
        local recaptcha_result, err = core.json.decode(res.body)
```
Need to check the result

docs/en/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Attributes
+
+| Name      | Type          | Requirement | Default    | Valid                                                                    | Description                                                                                                                                         |
+| --------- | ------------- | ----------- | ---------- | ------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| secret_key | string        | required    |            |  | The secret key of the Google reCAPTCHA v2 service. |
+| parameter_source | string | optional | header | | The enum of captcha parameter source. Only `header`, `query` are supported. |
+| parameter_name | string | optional | captcha | | The name of captcha parameter. |
+| response | object | optional    | content_type  = `application/json; charset=utf-8`<br />status_code = `400`<br />body = `{"message":"invalid captcha"}` |  | The response of invalid recaptcha token. |
```
spacewander, line 36:
```suggestion
| response | object | optional    | content_type  = `application/json; charset=utf-8`<br />status_code = `400`<br />body = `{"message":"invalid captcha"}` |  | The response for invalid recaptcha token. |
```

docs/en/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,114 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Name
```
spacewander, line 25:
```suggestion
## Description
```

t/plugin/recaptcha.t
spacewander: It's easy to do it. Here is the script:
```
#!/usr/bin/env python
# coding: utf-8
# Usage: ./simpify.py $test_filename
import sys
with open(sys.argv[1]) as f:
    lines = [l.rstrip() for l in f.readlines()]
newlines = []
for i, l in enumerate(lines):
    if ((l == "GET /t" and lines[i-1] == "--- request") or
        (l == "[error]" and lines[i-1] == "--- no_error_log")):
        newlines = newlines[:-1]
        continue
    newlines.append(l)
res = "\n".join(newlines)
# print(res)
with open(sys.argv[1], "w") as f:
   f.write(res + '\n') 
```

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,110 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## 简介
```
spacewander, line 25:
```suggestion
## 描述
```

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,110 @@
...
+
+通过向 Google reCAPTCHA 服务校验客户端传递的验证码来限制对上游服务的访问。插件支持自定义无效校验码的响应体。注意，此插件只支持 Google reCAPTCHA v2 版本。
```
leslie-tsang, line 27:
```suggestion
> 注意：此插件只支持 Google reCAPTCHA v2 版本。

通过向 Google reCAPTCHA 服务校验客户端传递的验证码来限制对上游服务的访问。插件支持自定义无效校验码的响应体。
```
Would be better ?

apisix/plugins/recaptcha.lua
Dog-Lee: will you prefer to return 500 or 503 if JSON decode failed?

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,110 @@
...
+## 测试插件
+
+使用 `curl` 访问：
+
+```shell
+curl -X POST 'http://127.0.0.1:9080/login'
+{"message":"invalid captcha"}
+
+curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_invalid_captcha'
+{"message":"invalid captcha"}
```
leslie-tsang, line 92:
```suggestion
$ curl -X POST 'http://127.0.0.1:9080/login'
{"message":"invalid captcha"}

$ curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_invalid_captcha'
{"message":"invalid captcha"}
```
Need to use `$` to distinguish between command and the output, :)

docs/zh/latest/plugins/recaptcha.md
Dog-Lee: How about this?  

通过向 Google reCAPTCHA 服务校验客户端传递的验证码来限制对上游服务的访问。插件支持自定义无效校验码的响应体。

>注意，此插件只支持 Google reCAPTCHA v2 版本。

leslie-tsang: LGTM

apisix/plugins/recaptcha.lua
membphis: I think we should use `400` here because the request body is invalid.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+
+local _M = {
+    version = 0.1,
+    priority = 700,
+    name = "recaptcha",
+    schema = schema,
+}
+
+function _M.check_schema(conf, schema_type)
+    return core.schema.check(schema, conf)
```
tzssangglass, line 49: `schema_type` is unuesd?

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+        captcha = uri_args[conf.parameter_name]
+    end
+    return captcha
+end
+
+function _M.access(conf, ctx)
+    local path = ctx.var.uri
+    local method = core.request.get_method()
+
+    core.log.debug("path: ", path, ", method: ", method, ", conf: ", core.json.encode(conf))
```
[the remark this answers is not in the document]
membphis: nice catch

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+local function retrieve_captcha(ctx, conf)
+    local captcha
+    if conf.parameter_source == "header" then
+        captcha = core.request.header(ctx, conf.parameter_name)
+    elseif conf.parameter_source == "query" then
+        local uri_args = core.request.get_uri_args(ctx) or {}
+        captcha = uri_args[conf.parameter_name]
+    end
+    return captcha
+end
```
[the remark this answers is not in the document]
membphis: another style and cleaer:

```lua
local function retrieve_captcha(ctx, conf)
    if conf.parameter_source == "header" then
        return core.request.header(ctx, conf.parameter_name)
    end

    if conf.parameter_source == "query" then
        local uri_args = core.request.get_uri_args(ctx)
        return uri_args and uri_args[conf.parameter_name]
    end
end
```

apisix/plugins/recaptcha.lua
Dog-Lee: Please check this out. https://apisix.apache.org/docs/apisix/plugin-develop

At the same time, we need to implement the check_schema(conf) method to complete the specification verification.
```
function _M.check_schema(conf, schema_type)
end
```
Note: the project has provided the public method "core.schema.check", which can be used directly to complete JSON verification.

apisix/plugins/recaptcha.lua
Dog-Lee:
```
$ curl --location --request POST 'https://www.recaptcha.net/recaptcha/api/siteverify'
{
    "success": false,
    "error-codes": [
        "missing-input-secret"
    ]
}
```
I think recaptcha API always returns JSON response even the request is invalid. So I suggest logging the `err` and then taking this case as invalid_captcha since the request is not qualify the recaptcha API. Do you guys agree?

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+        local remote_ip = core.request.get_remote_client_ip(ctx)
+        local res, err = httpc:request_uri(recaptcha_url .. "/recaptcha/api/siteverify", {
+            method = "POST",
+            body = "secret=" .. secret .. "&response=" .. captcha .. "&remoteip=" .. remote_ip,
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+            },
+            ssl_verify = conf.ssl_verify
+        })
+        if err then
```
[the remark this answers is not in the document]
Dog-Lee: Can you please explain why?

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,102 @@
...
+            core.log.error("request failed: ", err)
+            return 503
+        end
+        core.log.debug("recaptcha veirfy result: ", res.body)
+        local recaptcha_result = core.json.decode(res.body)
+        if recaptcha_result.success == true then
+            invalid_captcha = false
+        end
```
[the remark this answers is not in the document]
Dog-Lee: I prefer to handle all of the invalid captcha cases in the last.
```
 if invalid_captcha then
      core.response.set_header("Content-Type", conf.response.content_type)
      return conf.response.status_code, core.utils.resolve_var(conf.response.body, ctx.var)
  end
```

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
Dog-Lee: These are legacy code in the first version. already removed.

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
tzssangglass: `res` does not always have a value, and may be nil

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
tzssangglass: But you've only set `invalid_captcha == false` in one place, why not just handle it here?

apisix/plugins/recaptcha.lua
membphis: APISIX mainly follow REST API design. I think use `400` is clearer.

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
vm-001:
```lua
function _M.access(conf, ctx)
    local invalid_captcha = true
    local captcha = retrieve_captcha(ctx, conf)
    if captcha == nil or captcha == "" then
        -- duplicated !
        core.response.set_header("Content-Type", conf.response.content_type)
        return conf.response.status_code, core.utils.resolve_var(conf.response.body, ctx.var)
    end

    local httpc = http.new()
    local secret = conf.secret_key
    local remote_ip = core.request.get_remote_client_ip(ctx)
    local res, err = httpc:request_uri(recaptcha_url .. "/recaptcha/api/siteverify", {
        method = "POST",
        body = ngx.encode_args({ secret = secret, response = captcha, remoteip = remote_ip }),
        headers = {
            ["Content-Type"] = "application/x-www-form-urlencoded",
        },
        ssl_verify = conf.ssl_verify
    })
    if not res then
        core.log.error("request failed: ", err)
        return 503
    end
    core.log.debug("recaptcha verify result: ", res.body)
    local recaptcha_result, err = core.json.decode(res.body)
    if err then
        core.log.error("faield to decode the recaptcha response json: ", err)
    end
    if recaptcha_result.success ~= true then
        -- duplicated !
        core.response.set_header("Content-Type", conf.response.content_type)
        return conf.response.status_code, core.utils.resolve_var(conf.response.body, ctx.var)
    end

    return
end
```

Hi @tzssangglass, two duplicated return codes.

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
vm-001:
```
if not res then
    core.log.error("request failed: ", err)
    return 503
end
```

I see.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,101 @@
...
+    priority = 700,
+    name = "recaptcha",
+    schema = schema,
+}
+
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+local function retrieve_captcha(ctx, conf)
```
leslie-tsang, line 53:
```suggestion

local function retrieve_captcha(ctx, conf)
```
Please follow the [CODE_STYLE.md](https://github.com/apache/apisix/blob/master/CODE_STYLE.md#blank-line).

> The functions needs to be separated by two blank lines:

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,101 @@
...
+        return core.request.header(ctx, conf.parameter_name)
+    end
+
+    if conf.parameter_source == "query" then
+        local uri_args = core.request.get_uri_args(ctx) or {}
+        return uri_args[conf.parameter_name]
+    end
+end
+
+function _M.access(conf, ctx)
```
leslie-tsang, line 64:
```suggestion

function _M.access(conf, ctx)
```
Please follow the [CODE_STYLE.md](https://github.com/apache/apisix/blob/master/CODE_STYLE.md#blank-line).

> The functions needs to be separated by two blank lines:

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,112 @@
...
+> 注意，此插件只支持 Google reCAPTCHA v2 版本。
+
+## 属性
+
+| Name      | Type          | Requirement | Default    | Valid                                                                    | Description                         |
+| --------- | ------------- | ----------- | ---------- | ------------------------------------------------------------------------ |-------------------------------------|
+| secret_key | string        | 必须    |            |  | Google reCAPTCHA v2 的 secret key    |
+| parameter_source | string | 可选 | header | | 验证码参数的来源枚举值。当前仅支持 `header`, `query` |
+| parameter_name | string | 可选 | captcha | | 验证码参数的名称                            |
+| response | object | 可选    | content_type  = `application/json; charset=utf-8`<br />status_code = `400`<br />body = `{"message":"invalid captcha"}` |  | 无效验证码的 HTTP 响应体                     |
```
leslie-tsang, line 38: Why `<br>` here ?

leslie-tsang (changes requested):

docs/zh/latest/plugins/recaptcha.md
vm-001: The response is an object type and it has three properties. And I just want to make sure it looks good. Has APISIX had any guidance on this? Or any suggestions from you?

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
tzssangglass: `local invalid_captcha = true` is where you declare invalid_captcha, I mean `invalid_captcha == false` only one place.

docs/zh/latest/plugins/recaptcha.md
leslie-tsang: > Has APISIX had any guidance on this ?

Not yet.

> Or any suggestions from you ?

I'm worried that this might cause exceptions in other markdown renderers, shall we remove it ?

vm-001: Sure.

apisix/plugins/recaptcha.lua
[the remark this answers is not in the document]
vm-001: No. if request didn't pass the captcha parameter in the header(by default) also cause invalid captcha. and it will be handled in the last in function.

t/plugin/recaptcha.t
```diff
@@ -0,0 +1,241 @@
...
+captcha: test
+--- error_code: 404
+
+
+
+=== TEST 7: recaptcha valid
+--- request
+POST /active?captcha=test
+--- error_code: 404
+--- no_error_log
```
spacewander, line 240: We don't need this?

docs/en/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,116 @@
...
+## Test Plugin
+
+Use curl to request:
+
+```shell
+$ curl -X POST 'http://127.0.0.1:9080/login'
+{"message":"invalid captcha"}
+
+$ curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_invalid_captcha'
+{"message":"invalid captcha"}
```
spacewander, line 94: Need a successful case

t/plugin/recaptcha.t
vm-001: If you are talking about 
```
--- no_error_log
[error]
```
then yes.

If you are talking `TEST 6` and `TEST 7`. We need these cases for test passed cases. (because the test data had been added on `TEST4`, so I just omit it.)

spacewander: Yes, I just mean the `--- no_error_log`.

docs/en/latest/plugins/recaptcha.md
vm-001: Sorry. It's been a while. I have some personal tasks need to work on.  I just add a successful case, Can you please help to review it?

moonming: @vm-001 great work 👍

spacewander:
CI failed

apisix/plugins/recaptcha.lua
vm-001: @leslie-tsang I had fixed the code style in another commit. Can you please redo the "requested changes"? because it still exists here.

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,103 @@
...
+            ssl_verify = conf.ssl_verify
+        })
+        if not res then
+            core.log.error("request failed: ", err)
+            return 503
+        end
+        core.log.debug("recaptcha verify result: ", res.body)
+        local recaptcha_result, err = core.json.decode(res.body)
+        if err then
+            core.log.error("failed to decode the recaptcha response json: ", err)
```
spacewander, line 88: We need to return from here as the recaptcha_result is nil.
And please add a test to cover this.

docs/en/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,126 @@
...
+| secret_key | string        | required    |                                                                                                                |  | The secret key of the Google reCAPTCHA v2 service. |
+| parameter_source | string | optional | header                                                                                                         | | The enum of captcha parameter source. Only `header`, `query` are supported. |
+| parameter_name | string | optional | captcha                                                                                                        | | The name of captcha parameter. |
+| response | object | optional    | content_type  = `application/json; charset=utf-8`; status_code = `400`; body = `{"message":"invalid captcha"}` |  | The response for invalid recaptcha token. |
+| ssl_verify | boolean | optional | true                                                                                                           | | verify if SSL cert matches hostname. |
```
spacewander, line 39: The ssl_verify does not just verify the hostname...

tzssangglass:
Can we mock a successful response and add a test case that verifies success?

you can ref: https://github.com/apache/apisix/blob/943436a16c3820f6f5921c2d98528468a89e323b/t/plugin/azure-functions.t#L31-L68

apisix/plugins/recaptcha.lua
```diff
@@ -0,0 +1,103 @@
...
+        end
+        if recaptcha_result and recaptcha_result.success == true then
+            invalid_captcha = false
+        end
+    end
+
+    if invalid_captcha then
+        core.response.set_header("Content-Type", conf.response.content_type)
+        return conf.response.status_code, core.utils.resolve_var(conf.response.body, ctx.var)
```
tzssangglass, line 97: why use `core.utils.resolve_var(conf.response.body, ctx.var)`?

moonming:
@leslie-tsang please review again, thx

docs/en/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,126 @@
...
+
+$ curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_valid_captcha'
+<html>
+<head><title>503 Service Temporarily Unavailable</title></head>
+<body>
+<center><h1>503 Service Temporarily Unavailable</h1></center>
+<hr><center>openresty</center>
+</body>
+</html>
+(Request has been proxied to the upstream service. Here 503 code is because we didn't set up an HTTP service on 127.0.0.1:1980)
```
leslie-tsang, line 104:
```suggestion
$ curl -I -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_valid_captcha'
HTTP/1.1 200 OK
...
```
Would be better to add a normal upstream response case ?

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,122 @@
...
+使用 `curl` 访问：
+
+```shell
+$ curl -X POST 'http://127.0.0.1:9080/login'
+{"message":"invalid captcha"}
+
+$ curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_invalid_captcha'
+{"message":"invalid captcha"}
+
+$ curl -X POST 'http://127.0.0.1:9080/login' -H 'captcha: the_valid_captcha'
```
leslie-tsang, line 96: Ditto

apisix/plugins/recaptcha.lua
leslie-tsang: It seems that the author wants to put some dynamic var in the response body, would it be better to use static ones?

docs/zh/latest/plugins/recaptcha.md
```diff
@@ -0,0 +1,122 @@
...
+| --------- | ------------- | ----------- |----------------------------------------------------------------------------------------------------------------| ------------------------------------------------------------------------ |-------------------------------------|
+| secret_key | string        | 必须    |                                                                                                                |  | Google reCAPTCHA v2 的 secret key    |
+| parameter_source | string | 可选 | header                                                                                                         | | 验证码参数的来源枚举值。当前仅支持 `header`, `query` |
+| parameter_name | string | 可选 | captcha                                                                                                        | | 验证码参数的名称                            |
+| response | object | 可选    | content_type  = `application/json; charset=utf-8`; status_code = `400`; body = `{"message":"invalid captcha"}` |  | 无效验证码的 HTTP 响应体                     |
+| ssl_verify | boolean | 可选 | true                                                                                                           | | 验证 SSL 证书与主机名是否匹配 |
```
leslie-tsang, line 39: > The ssl_verify does not just verify the hostname...

Ditto

leslie-tsang (commented):
Hello there, Would you like to take some to continue this PR ?
Please do let me know if any help in needed. :)

Issue #6538, opened 2022-03-08
docs: remove invalid response body data of test files


Chever-John:
### Improved or typo fixed.

In the test file, there is a lot of redundant request response body data, and some just duplicate the input content. For this reason, a PR is needed to improve the test example situation.

For example, this following section of the test file fits the above description, and all you need to do is remove the following section.

```shell
=== TEST 5: set route(id: 1) to etcd cluster with TLS
--- yaml_config
apisix:
  node_listen: 1984
  admin_key: null
etcd:
  host:
    - "https://127.0.0.1:12379"
  tls:
    verify: false
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin").test
            local code, body = t('/apisix/admin/routes/1',
                ngx.HTTP_PUT,
                [[{
                    "methods": ["GET"],
                    "upstream": {
                        "nodes": {
                            "127.0.0.1:8080": 1
                        },
                        "type": "roundrobin"
                    },
                    "desc": "new route",
                    "uri": "/index.html"
                }]],
                # The response body below is what needs to be deleted
                [[{
                    "node": {
                        "value": {
                            "methods": [
                                "GET"
                            ],
                            "uri": "/index.html",
                            "desc": "new route",
                            "upstream": {
                                "nodes": {
                                    "127.0.0.1:8080": 1
                                },
                                "type": "roundrobin"
                            }
                        },
                        "key": "/apisix/routes/1"
                    },
                    "action": "set"
                }]]
                )

            ngx.status = code
            ngx.say(body)
        }
    }
--- request
GET /t
--- response_body
passed
--- no_error_log
[error]
```

### Solution

You can solve this problem by writing scripts in languages that include but are not limited to **Python**.

The logic of a scripting language can be as follows:
1. Determine the scope of the script;
2. Identifying marker(need to find out by yourself);
3. Delete the associated row.

Your suggestions and more efficient solutions are welcome!

Chever-John:
Please assigne this issue to me.

PR #6545, opened 2022-03-08
test: remove unnecessary response body data check

Chever-John:

In the test file, there is a lot of redundant request response body data, and some just duplicate the input content. For this reason, a PR is needed to improve the test example situation.

Fix https://github.com/apache/apisix/issues/6538

### Pre-submission checklist:


spacewander (commented):
Look like we need to update:
```
ngx.status = code
```
to
```
if code >= 300 then
    ngx.status = code
end
```

Because the response data check will convert the status code to 200:
https://github.com/apache/apisix/blob/9b0460fb68e26d3fc1e8eac6659794598ce859ac/t/lib/test_admin.lua#L204-L214

spacewander (approved):

shuaijinchao (approved):

Issue #6574, opened 2022-03-10
feat: keycloak 1. Support of password grant type for token generation 2. Support for User Info Injection in the request


azilentech:
### Issue description

In Keycloak plugin,

The following features can be helpful:
1. There can be feature to support "password" grant type which can generate token via Keycloak token endpoint and return it in response.
-  There can be configuration setting "token_generation_endpoint", which will contain URL endpoint. If incoming URL request match with this URL, identify the user name and password from incoming request parameters and call "token_endpoint". 

2. There can be an optional configuration that can allow injecting Keycloak User Info in the incoming request.
- There can be configuration settings: "include_user_info" and "token_userinfo_endpoint". Based on these values, the plugin will call Token Userinfo Endpoint and inject a new header "X-User-Info" in the request before forwarding it for further process.

azilentech:
Hi,
We have implemented this feature for our internal scenario already and If it can be helpful to a larger audience, we have created PR for review.
We have kept only a single feature, as you suggested in the PR.
Please let me know, your comments for the same.

We will also create separate PR for Dashboard changes so that it can support these keyclaok plugin changes.

azilentech:
In the code, instead of taking "token_generation_endpoint" as configuration setting name, I have used "password_grant_token_generation_incoming_uri" as setting name.
As the previous name was making confusion with "token_endpoint" which was referring to keycloak url.

PR #6586, opened 2022-03-11
feat: add support for password grant in keycloak plugin

azilentech:
Support of password grant type for token generation (#6574)

Committing changes after adding new feature for generating token based on `user id/password` and also taking support of grant type `password`

feat: #6574

### Pre-submission checklist:

### DIFF
--- a/apisix/plugins/authz-keycloak.lua
+++ b/apisix/plugins/authz-keycloak.lua
@@ -19,10 +19,12 @@ local http      = require "resty.http"
 local sub_str   = string.sub
 local type      = type
 local ngx       = ngx
+local util = require("apisix.cli.util")
 local plugin_name = "authz-keycloak"
 
 local log = core.log
 local pairs = pairs
+local ipairs = ipairs
 
 local schema = {
     type = "object",
@@ -700,14 +702,6 @@ local function fetch_jwt_token(ctx)
     return token
 end
 
-local function split_string(s, delimiter)
-    local result = {}
-    for match in (s..delimiter):gmatch("(.-)"..delimiter) do
-        table.insert(result, match)
-    end
-    return result
-end
-
 -- To get new access token by calling get token api
 local function generate_token_using_password_grant(conf,ctx)
     log.warn("generate_token_using_password_grant Function Called")
@@ -718,18 +712,23 @@ local function generate_token_using_password_grant(conf,ctx)
     local username = nil
     local password = nil
     --split by &
-    local parameters_array = split_string(request_body, "&")
-    if parameters_array then
+    local parameters_array = util.split(request_body, "&")
+
+    if #parameters_array == 2 then
         for k, parameter in ipairs(parameters_array) do
             if string.find(parameter, "username") then
                 --split by =
-                local username_value_array = split_string(parameter, "=")
-                username = username_value_array[2]
+                local username_value_array = util.split(parameter, "=")
+                if #username_value_array == 2 then
+                    username = username_value_array[2]
+                end
             end
             if string.find(parameter, "password") then
                 --split by =
-                local password_value_array = split_string(parameter, "=")
-                password = password_value_array[2]
+                local password_value_array = util.split(parameter, "=")
+                if #password_value_array == 2 then
+                    password = password_value_array[2]
+                end
             end
         end
     end
--- a/apisix/plugins/authz-keycloak.lua
+++ b/apisix/plugins/authz-keycloak.lua
@@ -380,7 +380,7 @@ local function authz_keycloak_ensure_sa_access_token(conf)
 
                 local params = {
                     method = "POST",
-                    body =  ngx.encode_args({
+                    body = ngx.encode_args({
                         grant_type = "refresh_token",
                         client_id = client_id,
                         client_secret = conf.client_secret,
@@ -456,7 +456,7 @@ local function authz_keycloak_ensure_sa_access_token(conf)
 
         local params = {
             method = "POST",
-            body =  ngx.encode_args({
+            body = ngx.encode_args({
                 grant_type = "client_credentials",
                 client_id = client_id,
                 client_secret = conf.client_secret,
@@ -644,7 +644,7 @@ local function evaluate_permissions(conf, ctx, token)
 
     local params = {
         method = "POST",
-        body =  ngx.encode_args({
+        body = ngx.encode_args({
             grant_type = conf.grant_type,
             audience = authz_keycloak_get_client_id(conf),
             response_mode = "decision",
@@ -732,13 +732,13 @@ local function generate_token_using_password_grant(conf,ctx)
     if not token_endpoint then
         local err = "Unable to determine token endpoint."
         log.error(err)
-        return 500, err
+        return 503, err
     end
     local httpc = authz_keycloak_get_http_client(conf)
 
     local params = {
         method = "POST",
-        body =  ngx.encode_args({
+        body = ngx.encode_args({
             grant_type = "password",
             client_id = client_id,
             client_secret = conf.client_secret,
@@ -775,9 +775,11 @@ local function generate_token_using_password_grant(conf,ctx)
 end
 
 function _M.access(conf, ctx)
+    local headers = core.request.headers(ctx)
     if conf.password_grant_token_generation_incoming_uri and
-        ngx.var.request_uri:upper() ==
-        conf.password_grant_token_generation_incoming_uri:upper() and
+        ngx.var.request_uri ==
+        conf.password_grant_token_generation_incoming_uri and
+        headers["content-type"] == "application/x-www-form-urlencoded" and
         core.request.get_method() == "POST" then
             return generate_token_using_password_grant(conf,ctx)
     end
--- a/apisix/plugins/authz-keycloak.lua
+++ b/apisix/plugins/authz-keycloak.lua
@@ -67,7 +67,11 @@ local schema = {
         access_token_expires_leeway = {type = "integer", minimum = 0, default = 0},
         refresh_token_expires_in = {type = "integer", minimum = 1, default = 3600},
         refresh_token_expires_leeway = {type = "integer", minimum = 0, default = 0},
-        token_generation_endpoint = {type = "string", minLength = 1, maxLength = 4096},
+        password_grant_token_generation_incoming_uri = {
+            type = "string",
+            minLength = 1,
+            maxLength = 4096
+        },
     },
     allOf = {
         -- Require discovery or token endpoint.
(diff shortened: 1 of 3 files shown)


soulbird:
Missing unit tests

azilentech:
Hi,

I will need little guidance for the CLA signing process from your side.

In PR, I have made the following changes.

- Added test cases. As these test cases require keycloak installation, I tested with my local environment by adjusting parameters and configuration settings. Please let me know If I need to perform further steps.
- I updated code by using better configuration setting "password_grant_token_generation_incoming_uri" instead of using "token_generation_endpoint" which can be confusing with "token_endpoint".
- I updated the documentation by following contribution guidelines.
- I also ran "make lint" and followed all the suggestions.

Please let me know the next actions for me.

spacewander:
> I will need little guidance for the CLA signing process from your side.

There is no CLA file for APISIX :)

> Added test cases. As these test cases require keycloak installation, I tested with my local environment by adjusting parameters and configuration settings. Please let me know If I need to perform further steps.

As the keycloak environment is set up by @sshniro,
@sshniro 
Could you provide some help for @azilentech to test this feature? Thanks!

niketrk:
@sshniro 

I checked the logs of above test runs.
error is: "status=401, body={\"error\":\"invalid_grant\",\"error_description\":\"Invalid user credentials\"}"

I have used following settings in my test scenario:
"client_id": "course_management"
"client_secret": "d1ec69e9-55d2-4109-a3ea-befa071579d5"
username = "teacher@gmail.com"
password = "123456"
grant_type = "password"

I took the above settings from other Keycloak test cases.
I am doubting that these settings are not proper for the test Keycloak instance. In my local environment, I put my Keycloak settings and then run test scenario via
`TEST_NGINX_BINARY=/usr/local/bin/openresty prove -Itest-nginx/lib -r t/plugin/authz-keycloak.t`

And it worked properly.

-------------
Another issue is reported as: "ERROR: apisix/plugins/authz-keycloak.lua: line 719: getting the Lua global "string""

For which I made fix "local str_find = string.find", so it should be fixed now.

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +703,107 @@ local function fetch_jwt_token(ctx)
     return token
 end
 
+-- To get new access token by calling get token api
+local function generate_token_using_password_grant(conf,ctx)
+    log.warn("generate_token_using_password_grant Function Called")
```
starsz, line 708: Why should we use `warn` level ?

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +703,107 @@ local function fetch_jwt_token(ctx)
...
+    --Read Body
+    ngx.req.read_body()
+    --Get Body Data
+    local request_body=ngx.req.get_body_data()
```
starsz, line 712: You can use `core.request.get_body()` instead.

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +703,107 @@ local function fetch_jwt_token(ctx)
...
+    local json, err = authz_keycloak_parse_json_response(res)
+
+    if not json then
+        err = "Could not decode JSON from response"
+              .. (err and (": " .. err) or '.')
+        log.error(err)
+        return 401, {message = err}
+    end
+
+    return  res.status, res.body
```
starsz, line 794:
```suggestion
    return res.status, res.body
```

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +703,107 @@ local function fetch_jwt_token(ctx)
...
+end
 
 function _M.access(conf, ctx)
+
+    if conf.password_grant_token_generation_incoming_uri then
+        if ngx.var.request_uri:upper()
+                == conf.password_grant_token_generation_incoming_uri:upper() then
```
starsz, line 801: Use `and` would be better.

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +703,107 @@ local function fetch_jwt_token(ctx)
...
+local function generate_token_using_password_grant(conf,ctx)
+    log.warn("generate_token_using_password_grant Function Called")
+    --Read Body
+    ngx.req.read_body()
+    --Get Body Data
+    local request_body=ngx.req.get_body_data()
+    local username = nil
+    local password = nil
+    --split by &
+    local parameters_array = util.split(request_body, "&")
```
starsz, line 716: Why the username, password in the request body and use & to concat.

apisix/plugins/authz-keycloak.lua
azilentech: Agree... it should be debug level log. I changed it.

apisix/plugins/authz-keycloak.lua
azilentech: Done. I changed it.

apisix/plugins/authz-keycloak.lua
azilentech: Done.

apisix/plugins/authz-keycloak.lua
azilentech: Good Suggestion and changed it to use and for all three conditions.

apisix/plugins/authz-keycloak.lua
azilentech: As url encoded parameters were coming, they were automatically appended using '&'. Having said that, your comment helped me to utilize "ngx.decode_args" which makes code much cleaner. Thanks!

sshniro:
> scenario

Hi, are you getting the same error when you run the docker image instance (sshniro/keycloak-apisix:latest) locally?

azilentech:
@sshniro 
I didn't test that way earlier, but just now, I did that and it is working. Also when I see the above checks, that test case is starting to pass now. 

I have got the idea what things might cause that. Earlier I was using '&' to split the incoming parameters. but now I am using "ngx.decode_args" ( https://github.com/apache/apisix/pull/6586#discussion_r825597553 ).

Because of that earlier, username (which is email) wouldn't parse properly.
That should be the root cause of the same.

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,88 @@ local function fetch_jwt_token(ctx)
...
+
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+
+    if conf.password_grant_token_generation_incoming_uri and
+        ngx.var.request_uri:upper() ==
+        conf.password_grant_token_generation_incoming_uri:upper() and
+        ctx.curr_req_matched["_method"]:upper() == "POST" then
```
starsz, line 782: `ctx.curr_req_matched["_method"]:upper()` looks strange.
I think you can use `core.req.get_method` instead.

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,88 @@ local function fetch_jwt_token(ctx)
...
+              .. (err and (": " .. err) or '.')
+        log.error(err)
+        return 401, {message = err}
+    end
+
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+
```
starsz, line 778: redundant line ?

docs/en/latest/plugins/authz-keycloak.md
```diff
@@ -122,6 +123,27 @@ of the same name. The scope is then added to every permission to check.
 If `lazy_load_paths` is `false`, the plugin adds the mapped scope to any of the static permissions configured
 in the `permissions` attribute, even if they contain one or more scopes already.
 
+### Password Grant Token Generation Incoming URI
+
+If you want to generate a token using `password` grant, you can set value of `password_grant_token_generation_incoming_uri`.
```
starsz, line 128:
```suggestion
If you want to generate a token using `password` grant, you can set the value of `password_grant_token_generation_incoming_uri`.
```

docs/en/latest/plugins/authz-keycloak.md
```diff
@@ -122,6 +123,27 @@ of the same name. The scope is then added to every permission to check.
...
+
+Incoming request URI will be matched with this value and if matched, it will generate token using `Token Endpoint`.
```
starsz, line 130:
```suggestion
Incoming request URI will be matched with this value and if matched, it will generate a token using `Token Endpoint`.
```

starsz (approved):

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,87 @@ local function fetch_jwt_token(ctx)
...
+    if not token_endpoint then
+        local err = "Unable to determine token endpoint."
+        log.error(err)
+        return 500, err
+    end
+    local httpc = authz_keycloak_get_http_client(conf)
+
+    local params = {
+        method = "POST",
+        body =  ngx.encode_args({
```
spacewander, line 741:
```suggestion
        body = ngx.encode_args({
```

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,87 @@ local function fetch_jwt_token(ctx)
...
+    end
+
+    local client_id = authz_keycloak_get_client_id(conf)
+
+    local token_endpoint = authz_keycloak_get_token_endpoint(conf)
+
+    if not token_endpoint then
+        local err = "Unable to determine token endpoint."
+        log.error(err)
+        return 500, err
```
spacewander, line 735:
```suggestion
        return 503, err
```

New feature is encouraged to avoid 500

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,87 @@ local function fetch_jwt_token(ctx)
...
+    end
+
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+    if conf.password_grant_token_generation_incoming_uri and
+        ngx.var.request_uri:upper() ==
+        conf.password_grant_token_generation_incoming_uri:upper() and
+        core.request.get_method() == "POST" then
```
spacewander, line 781: We should also check the "Content-Type"?

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,87 @@ local function fetch_jwt_token(ctx)
...
+        return 401, {message = err}
+    end
+
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+    if conf.password_grant_token_generation_incoming_uri and
+        ngx.var.request_uri:upper() ==
+        conf.password_grant_token_generation_incoming_uri:upper() and
```
spacewander, line 780: Is it required to compare case-insensitive?
Also, the `var.request_uri` will contain `?argument=part`, is it desirable?

t/plugin/authz-keycloak.t
```diff
@@ -621,3 +622,104 @@ GET /t
...
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:1982": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/api/token"
+                }]],
+                [[{
```
spacewander, line 656: We need to remove the echo response data as https://github.com/apache/apisix/pull/6545

apisix/plugins/authz-keycloak.lua
azilentech: done and also incorporated this in same file at other places.

apisix/plugins/authz-keycloak.lua
azilentech: Valid point. I updated it. I didn't update at other places because there can be backward compatibility issues and there should be separate PR for the same.

apisix/plugins/authz-keycloak.lua
azilentech: Yes. it was missed. I introduced that now.

apisix/plugins/authz-keycloak.lua
azilentech: URL should be case sensitive so removed that "upper" from code.
For the part of request_uri, I believe, as we are checking uri, method, and header, all three parts, this should be fine.
Basically for the operation of token generation, exact uri matching will be helpful.
Your thoughts?

t/plugin/authz-keycloak.t
azilentech: Alright. I removed it from this new test case.

apisix/plugins/authz-keycloak.lua
spacewander: The `request_uri` will contain the query string part of the uri. For example,
With request `/path?query=string`
The `ngx.var.request_uri` will return `/path?query=string`, and `ctx.var.uri` will return `/path`. By looking at the doc of `password_grant_token_generation_incoming_uri`, I guess we can use `ctx.var.uri` so both `/path?query=string` and `/path` can match.

spacewander (approved):
Approved except https://github.com/apache/apisix/pull/6586#discussion_r829659989

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,89 @@ local function fetch_jwt_token(ctx)
...
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+    local headers = core.request.headers(ctx)
+    if conf.password_grant_token_generation_incoming_uri and
+        ngx.var.request_uri ==
+        conf.password_grant_token_generation_incoming_uri and
+        headers["content-type"] == "application/x-www-form-urlencoded" and
+        core.request.get_method() == "POST" then
```
tzssangglass, line 783:
```suggestion
    local headers = core.request.headers(ctx)
    local need_grant_token = conf.password_grant_token_generation_incoming_uri and
            ngx.var.request_uri ==
                    conf.password_grant_token_generation_incoming_uri and
            headers["content-type"] == "application/x-www-form-urlencoded" and
            core.request.get_method() == "POST"
    if need_grant_token then
```

apisix/plugins/authz-keycloak.lua
azilentech: @spacewander I understood your idea. I am curious, as to why we should treat both strings (/path?query=string or /path) the same way and proceed with token generation.

Token generation is a very specific use case, and shouldn't we say if and only if a specific URL will be submitted then we will generate token and submit result back.
 
if the request is made with /path?query=string, it shouldn't be assumed for token generation and it should be treated the same as any other request. (e.g. proceed to forward to upstream for further processing).

Please share your view. Accordingly, I can make changes, if needed.

apisix/plugins/authz-keycloak.lua
azilentech: Done. This one is giving better readability. Thanks!

apisix/plugins/authz-keycloak.lua
```diff
@@ -695,8 +700,89 @@ local function fetch_jwt_token(ctx)
...
+        return 401, {message = err}
+    end
+
+    return res.status, res.body
+end
 
 function _M.access(conf, ctx)
+    local headers = core.request.headers(ctx)
+    local need_grant_token = conf.password_grant_token_generation_incoming_uri and
+        ngx.var.request_uri == conf.password_grant_token_generation_incoming_uri and
```
spacewander, line 780:
```suggestion
        ctx.var.request_uri == conf.password_grant_token_generation_incoming_uri and
```
would be better, which provides a cache of the var.

spacewander (dismissed):

apisix/plugins/authz-keycloak.lua
azilentech: Done. Made the change.

spacewander (approved):

tzssangglass (approved):

lijing-21:
Hi @azilentech , thank you for your contribution!

Here is the Contributor T-shirt form[1], if you're interested, kindly take a look :)

[1] https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#contributor-t-shirt

PR #6592, opened 2022-03-12
docs: update issue and pr templates

### LINKED ISSUE #6583: docs: Make issue forms consistent
### Improved or typo fixed.

The current issue forms are inconsistent with each different scenarios.

### Solution

Make the issue form consistent across all the different issue types and scenarios.

navendu-pottekkat:
Signed-off-by: Navendu Pottekkat <navendupottekkat@gmail.com>

This PR updates the issue and PR templates and makes it consistent.

Fixes #6583

### DIFF
--- a/.github/ISSUE_TEMPLATE/request_help.yml
+++ b/.github/ISSUE_TEMPLATE/request_help.yml
@@ -1,35 +1,36 @@
-name: "Request help"
-description: Please use this template for requesting help.
-title: "request help: "
+name: "Request Help"
+description: Stuck? Ask for help!
+title: "help request: "
 body:
   - type: markdown
     attributes:
       value: |
-        The more information you fill in, the better we can help you.
+        _The more information you share, the faster we can help you._
 
-        - Please use English on public channel.
-        - Encourage using words rather than pictures to describe errors.
-        - Encourage pre-searching with keywords in the issue.
+        Prior to opening the issue, please make sure that you:
+
+        - Use English to communicate.
+        - Search the [open issues](https://github.com/apache/apisix/issues) and [discussion forum](https://github.com/apache/apisix/discussions) to avoid duplicating the issue.
 
   - type: textarea
     id: description
     attributes:
-      label: Issue description
-      description: Provide a clear and concise description of the help you need.
+      label: Description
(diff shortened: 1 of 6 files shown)


navendu-pottekkat:
Hmm, seems like there are a lot of CI checks being run which are unnecessary. We may need to blacklist changes to this path so that it doesn't trigger these workflows.

.github/ISSUE_TEMPLATE/request_help.yml
```diff
@@ -1,35 +1,36 @@
...
     id: environment
     attributes:
       label: Environment
-      description: Request help without environment information will be ignored or closed.
+      description: Share your environment details. Reports without proper environment details will likely be closed.
       value: |
-        - apisix version (cmd: `apisix version`):
-        - OS (cmd: `uname -a`):
-        - OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
-        - etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
```
spacewander, line 30: Why remove the `get the info from server-info API`?

navendu-pottekkat: `etcd version` implies that you have to get it. And the API is self documented in my opinion and it is obvious what it does. So, an additional sentence might be redundant? What do you think? Should I add back that sentence?

spacewander: I see.

spacewander (approved):

navendu-pottekkat:
> Hmm, seems like there are a lot of CI checks being run which are unnecessary. We may need to blacklist changes to this path so that it doesn't trigger these workflows.

@spacewander What do you think about this?

spacewander:
> > Hmm, seems like there are a lot of CI checks being run which are unnecessary. We may need to blacklist changes to this path so that it doesn't trigger these workflows.
> 
> @spacewander What do you think about this?

There are ignores:
https://github.com/apache/apisix/blob/a490e0a2570d916921b4a4e50f3c308291ba7dd9/.github/workflows/build.yml#L7
but not in a fine grain size

tokers (approved):

Issue #6595, opened 2022-03-12
feat: Changing the limit-count plugin allow_degradation parameter default value to true.


fatihbm:
### Issue description

The main goal here should be to minimize the risk of downtime in the production environment due to a third-party (redis) service. 
This is a configurable setting but if the default value is set to true it could be a more safe way.

spacewander:
Normally we can't accept a break change, even turning false to true by default.

PR #6596, opened 2022-03-12
feat: Changing the limit-count plugin allow_degradation parameter default value to true.

fatihbm:
feat: #6595 

### Pre-submission checklist:


membphis:
I think we should add some test cases to confirm your changes.

spacewander (changes requested):
I am afraid we can't accept a break change.

Issue #6570, opened 2022-03-10
feat: How to make apisix be the gateway for tars


zhixiongdu027:
### Issue description

# Feat: How to make apisix be the gateway for tars

## Background：

TARS is a Linux Foundation project. It is a high-performance RPC framework based on name service and Tars protocol, also integrated administration platform, and implemented
hosting-service via flexible schedule.

Now I want to implement Tars proxy function with apisix, and make apisix be the gateway for tars

### a service discovery plugin (can both work in http module and stream module)

rpc calls to tars are addressed through servants. this discovery plugin will provide translation services from servants to nodes, like:

```lua
  local function discovery.nodes("Test.TestServer.HelloObj")
    return {
      { host = "172.16.8.33", port = 12334, weight =100,},
      { host = "172.16.8.33", port = 12334, weight =100,},
    }
  end
```

### a tars http proxy

There are two problems in providing tcp services externally:

1. Limited exposed ports 
2. Poor synergy with different frameworks

So there needs to be a gateway to provide http->tcp conversion and proxy
Tars already has an official TarsGateway to do this work, I will refer to it to implement APISix Tars Proxy

### a tars tcp proxy

It is also necessary to support tcp proxy:

+ RPC request proxy in the environment of internal and external network isolation
+ gateway can support server push

### discovery configuration as follows:

```yaml
discovery:
  tars:
    db_conf:
      host: 172.16.1.99
      port: 3306
      database: db_tars
      user: tars
      password: tars2022    
```

### http proxy configuration as follows:
configuration format is based on TarsGateway, with slight changes
```yaml
  tars-proxy:
    hosts: [ "*.tars.com","a.rpc.com" ]
    tup_uri: /tup
    json_uri: /tup-json
    auto_proxy: false
    default:
      balancer: chash
      hash_on: head
      key: XGUID
      timeout:
        connect: 1000
        send: 1000
        read: 1000
      discovery: tars
      forward_headers: [ "XGUID","REMOTE_IP" ]
    proxy:
      hello:test:
        servant: Test.TestServer.HelloObj
        nodes:
          - host: 172.16.33.8
            port: 9927
      hello:
        servant: Test.TestServer.HelloObj
      sum:
        servant: Test.GetSumServer.GetSumObj
        balancer: roudrobin
      user:
        servant: Test.UserServer.LoginObj
        forward_headers: [ "GUID","token" ]
```

### tcp proxy configuration as follows:
todo

spacewander:
[...]

PR #6599, opened 2022-03-13
feat: add tars discovery

zhixiongdu027:
### Pre-submission checklist:

#6570 

Hello everyone:

  This is a PR for the tars discovery module, which is part of "tars-proxy".
  
  Work on "tars-proxy" continues,  but maybe we can finish tars discovery first


t/tars/discovery/tars.t
```diff
@@ -0,0 +1,409 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
+log_level('debug');
```
spacewander, line 20: Is it necessary to use debug level?

t/tars/discovery/tars.t
```diff
@@ -0,0 +1,409 @@
...
+    $block->set_value("config", $config);
+});
+
+run_tests();
+
+__DATA__
+
+=== TEST 1: create initial server and servant
+--- timeout: 12
+--- yaml_config eval: $::yaml_config
```
spacewander, line 148: We already set the yaml_config?

t/tars/discovery/tars.t
```diff
@@ -0,0 +1,409 @@
...
+]
+--- response_body eval
+[
+    "DONE\n",
+    "DONE\n",
+    "DONE\n",
+    "DONE\n",
+    "{ 1 1 1 }\n",
+]
+--- no_error_log
```
spacewander, line 185: Let's save the no_error_log like https://github.com/apache/apisix/blob/1ac7166e80fdd502cc6a8a60ef58f415d3c0e56c/t/plugin/gzip.t#L40

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+
+
+function _M.nodes(servant)
+    return get_endpoint(servant)
+end
+
+
+function _M.init_worker()
+    -- TODO: maybe we can read dict name from discovery config
+    endpoint_dict = ngx.shared.discovery
```
spacewander, line 262: Let's use a new shdict instead of hijacking authz-keycloak's

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+    if not endpoint_dict then
+        error("failed to get nginx shared dict: discovery, please check your APISIX version")
+    end
+
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    local conf = local_conf.discovery.tars
+    default_weight = local_conf.discovery.tars.default_weight or 100
```
spacewander, line 272:
```suggestion
    default_weight = conf.default_weight
```
The default weight is set already.

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+    if last_fetch_error or last_fetch_full_time + conf.full_fetch_interval <= now then
+        last_fetch_full_time = now
+        last_fetch_error = fetch_full(db_cli)
+    else
+        last_fetch_error = fetch_incremental(db_cli)
+    end
+
+    if not last_fetch_error then
+        db_cli:set_keepalive(120 * 1000, 1)
+    end
```
spacewander, line 251: Need `close` if err?

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+
+
+local function create_endpoint_lrucache(servant)
+    local endpoint_content = endpoint_dict:get_stale(servant)
+    if not endpoint_content then
+        core.log.error("get empty endpoint content from discovery DICT, servant: ", servant)
+        return nil
+    end
+
+    local endpoint = core.json.decode(endpoint_content)
```
spacewander, line 102: Let's get the err and put it in the msg

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+
+local endpoint_lrucache = core.lrucache.new({
+    ttl = 300,
+    count = 1024
+})
+
+local activated_buffer = core.table.new(10, 0)
+local nodes_buffer = core.table.new(0, 5)
+
+local endpoints_pattern = core.table.concat(
```
spacewander, line 64: Could you show us the data which is matched by this regex?

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+
+
+local function delete_endpoint(servant)
+    core.log.info("delete servant ", servant)
+    endpoint_dict:delete(servant .. "#version")
+    endpoint_dict:delete(servant)
+end
+
+
+local function create_endpoint_lrucache(servant)
```
spacewander, line 95:
```suggestion
local function add_endpoint_to_lrucache(servant)
```

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,283 @@
...
+        core.log.error("decode endpoint content failed, content: ", endpoint_content)
+        return nil
+    end
+
+    return endpoint
+end
+
+
+local function get_endpoint(servant)
+    local endpoint_version = endpoint_dict:get_stale(servant .. "#version")
```
spacewander, line 113: Could you tell me why get_statle is used instead of get?

zhixiongdu027:
@spacewander 
Hi, all comment issues have been taken care of. please help review. 
Many TKS.

zhixiongdu027:
image

@spacewander @membphis 

Sorry, I don't understand how this failure happened. 
Please give a hand. TKS

apisix/discovery/tars/init.lua
spacewander: Need to handle this.

conf/config-default.yaml
```diff
@@ -266,6 +266,7 @@ nginx_config:                     # config for render the template to generate n
       introspection: 10m
       access-tokens: 1m
       ext-plugin: 1m
+      tars: 1m
```
spacewander, line 269: Let's make it configurable in the next PR.

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+        core.log.error("decode json failed, content: ", endpoint_content, ", err: ", err)
+        return nil
+    end
+
+    return endpoint
+end
+
+
+local function get_endpoint(servant)
+    local endpoint_version, err = endpoint_dict:get_stale(servant .. "#version")
```
spacewander, line 119: Why use get_stale when we don't have an expiration time?

t/tars/discovery/tars.t
```diff
@@ -0,0 +1,381 @@
...
+  enable_admin: false
+discovery:
+  tars:
+    db_conf:
+      host: 127.0.0.1
+      port: 3306
+      database: db_tars
+      user: root
+      password: tars2022
+    full_fetch_interval: 90
```
spacewander, line 41: Could we use a shorter interval instead of waiting > 90 in the test?

t/tars/discovery/tars.t
```diff
@@ -0,0 +1,381 @@
...
+
+    $block->set_value("apisix_yaml", $apisix_yaml);
+
+    my $config = $block->config // <<_EOC_;
+        location /count {
+            content_by_lua_block {
+              local core = require("apisix.core")
+              local d = require("apisix.discovery.tars")
+
+              ngx.sleep(11)
```
spacewander, line 60: Ditto

t/tars/discovery/tars.t
zhixiongdu027: this value Is check by schema ,
can we use an fake schema.lua  int test ?

apisix/discovery/tars/init.lua
zhixiongdu027: from SQL queries and have no expiration time in the cache sense.

Next:
When we read the results of fectch_full the logic is as follows:

1. endpoint_dict:flush_all() 2. for (result : results ){ endpoint_dict[servant]=nodes } 3. endpoint_dict: flush_expired()

get_endpoint may happen at 2,
we have to use endpoint_dict:get_stale

conf/config-default.yaml
zhixiongdu027: > Need to handle this.

Sorry，I did not get your point
Your means we can keep using this db hanlder event get last_db_error ?

spacewander: We need to close the handler when getting last_db_error?

apisix/discovery/tars/init.lua
spacewander: We don't set any expiration for the shdict's data, right? So why do we need to use get_stale & flush_expired?

t/tars/discovery/tars.t
spacewander: Maybe we can reduce the min in the schema? A sane user may not use a low value.

apisix/discovery/tars/init.lua
zhixiongdu027: Consider the following situation:

If the current endpoint is as follows:
k1: v1, k2: v2, k3: v3 ,k4: v4

Then we full fetch to get the following results
k1: v1, k4: v4, k5: v5

At this time, we need
1: add k4:v4, k5:v5
2: delete k2:v2, k3:v3

flush_all and flush_expired can achieve this

t/tars/discovery/tars.t
zhixiongdu027: > A sane user may not use a low value.
Yes, I agree with this

But I think the meaning of scheame's existence is to ensure that this value is within a reasonable range,

Is it not appropriate to modify the schema if it is only for testing

conf/config-default.yaml
zhixiongdu027: You mean we need to decide whether to close or not based on the value of error ?

Currently the case of error=='again' has been handled
But So Sorry for I don't know what other error values do not need to be closed.

zhixiongdu027:
@spacewander

Hi, I still have some questions to ask, please take a look at the comments I wrote.
thank you very much.

apisix/discovery/tars/init.lua
spacewander: Thanks for your explanation. Let's enrich the comment in the code.

t/tars/discovery/tars.t
spacewander: What about rewriting schema in the test?
Like https://github.com/apache/apisix/blob/9d450d7fe3169a77727df28696d083809b93977a/t/plugin/skywalking.t#L34-L35

zhixiongdu027: ok, I will try

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+            core.log.error("get empty endpoint version, servant: ", servant, ", err: ", err)
+        end
+        return nil
+    end
+    return endpoint_lrucache(servant, endpoint_version, add_endpoint_to_lrucache, servant)
+end
+
+
+local function extract_endpoint(query_result)
+    for _, p in ipairs(query_result) do
```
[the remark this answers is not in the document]
zhixiongdu027: I do not think so. 
Because the result of query_result is the return by db_cli, there is no type other than table.
But you reminded me that query_result maybe is {}

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+    local res, err, errcode, sqlstate = db_cli:query(incremental_query_sql)
+    if not res then
+        core.log.error("query failed, error: ", err, ", ", errcode, " ", sqlstate)
+        return err
+    end
+
+    core.table.clear(activated_buffer)
+    extract_endpoint(res)
+
+    while err == "again" do
```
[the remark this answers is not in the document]
zhixiongdu027: Actually I don't know much about this db:read_result
This code is copied from lua_resty_mysql

And according to the code, as long as res == nil , loop will break.

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+    end
+
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    local conf = local_conf.discovery.tars
+    default_weight = conf.default_weight
+
+    core.log.info("conf ", core.json.encode(conf, true))
```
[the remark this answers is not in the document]
zhixiongdu027: I will update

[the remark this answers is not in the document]
zhixiongdu027: > 
I will update

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+    if not endpoint_dict then
+        error("failed to get nginx shared dict: tars, please check your APISIX version")
+    end
+
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    local conf = local_conf.discovery.tars
+    default_weight = conf.default_weight
```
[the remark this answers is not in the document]
zhixiongdu027: I get your point.
But other discovery plugin is also used like this
https://github.com/apache/apisix/blob/c7755543ae2e097942da4bbc5a9c6136de90fd78/apisix/discovery/eureka/init.lua#L209
https://github.com/apache/apisix/blob/c7755543ae2e097942da4bbc5a9c6136de90fd78/apisix/discovery/nacos/init.lua#L379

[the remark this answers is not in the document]
tzssangglass: Thanks for the reminder, I will think about whether it is appropriate.

zhixiongdu027:
@spacewander @tzssangglass 
Hi , I have submitted a new submission,
Please help to review, thank you very much

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,332 @@
...
+select application, server_name from t_adapter_conf
+where registry_timestamp > now() - interval %d second
+)
+group by servant, activated order by activated desc ]]
+
+local _M = {
+    version = 0.1,
+}
+
+local endpoint_dict
```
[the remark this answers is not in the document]
tzssangglass: Sorry for my mistake. The original way here was fine. Because when we don't use tars as a discovery type, we can do without the declared ngx.shared.tars.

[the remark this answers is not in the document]
zhixiongdu027: OK，I will update

apisix/discovery/tars/init.lua
spacewander: @zhixiongdu027 
Please add close to the db_cli, thanks.

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,340 @@
...
+    while err == "again" do
+        res, err, errcode, sqlstate = db_cli:read_result()
+        if not res then
+            if err then
+                core.log.error("read result failed, error: ", err, ", ", errcode, " ", sqlstate)
+            end
+            return err
+        end
+        extract_endpoint(res)
+    end
```
spacewander, line 281: For style, let's add `return nil` at the end

apisix/discovery/tars/init.lua
spacewander: @zhixiongdu027 
Please add some comment about the expiration operation, thanks.

apisix/discovery/tars/init.lua
```diff
@@ -0,0 +1,340 @@
...
+    endpoint_dict:flush_expired()
```
spacewander, line 233: For style, let's add return nil at the end

conf/config-default.yaml
spacewander: According to the example, we need to close if we don't keepalive?
https://github.com/openresty/lua-resty-mysql#synopsis
```
				-- put it into the connection pool of size 100,
                -- with 10 seconds max idle timeout
                local ok, err = db:set_keepalive(10000, 100)
                if not ok then
                    ngx.say("failed to set keepalive: ", err)
                    return
                end

                -- or just close the connection right away:
                -- local ok, err = db:close()
                -- if not ok then
                --     ngx.say("failed to close: ", err)
                --     return
                -- end
```

zhixiongdu027: Sorry,  I  think this place doesn't actively call db:close() should be acceptable, 
[this link](https://forum.openresty.us/d/2151-f9a6aa2bb662b5ab77ee279e6dbd3ed6/4)  provides agentzh's view

apisix/discovery/tars/init.lua
zhixiongdu027: Sorry, I add comment in function fetch_full and forget  function get_endpoint()

I will update

apisix/discovery/tars/init.lua
zhixiongdu027: > Please add some comment about the expiration operation, thanks.

Sorry, I add comment in function fetch_full and forget function get_endpoint()

I will update

apisix/discovery/tars/init.lua
zhixiongdu027: > Need `close` if err?

Sorry, I think this place doesn't actively call db:close() should be acceptable,
[this link](https://forum.openresty.us/d/2151-f9a6aa2bb662b5ab77ee279e6dbd3ed6/4) provides agentzh's view

tzssangglass:
hi @zhixiongdu027 , file conflict, you need to merge master branch to your PR.

zhixiongdu027:
> hi @zhixiongdu027 , file conflict, you need to merge master branch to your PR.

Done

spacewander (approved):

tzssangglass (approved):

PR #6668, opened 2022-03-21
docs: translate some zh/*.md table properties to chinese

liangliang4ward:
### Description
translate some zh/*.md table properties to chinese

Fixes # (issue)

### Checklist


docs/zh/latest/plugins/csrf.md
```diff
@@ -29,7 +29,7 @@ title: csrf
 
 ## 属性
 
-| Name             | Type    | Requirement | Default | Valid | Description                                                  |
+| 名称             | 类型    | 必选 | 默认值 | 有效值 | 描述                                                 |
```
juzhiyuan, line 32:
```suggestion
| 名称             | 类型    | 必选项 | 默认值 | 有效值 | 描述                                                 |
```

juzhiyuan (commented):
Thanks! Also cc @hf400159 @kwanhur to take a look, for unifying Chinese Documentation Style 😄

kwanhur:
Translate [`optional` --> `可选`](https://github.com/apache/apisix/blob/master/docs/zh/latest/plugins/redirect.md?plain=1#L37 ) at the same.

docs/zh/latest/plugins/csrf.md
liangliang4ward: done

liangliang4ward:
> Translate [`optional` --> `可选`](https://github.com/apache/apisix/blob/master/docs/zh/latest/plugins/redirect.md?plain=1#L37) at the same.

done

docs/zh/latest/plugins/csrf.md
```diff
@@ -29,11 +29,11 @@ title: csrf
...
 
-| Name             | Type    | Requirement | Default | Valid | Description                                                  |
+| 名称             | 类型    | 必选项 | 默认值 | 有效值 | 描述                                                 |
 | ---------------- | ------- | ----------- | ------- | ----- | ------------------------------------------------------------ |
-|   name   |  string |    optional    | `apisix-csrf-token`  |    | 生成的 Cookie 中的 token 的名字，需要使用这个名字在请求头携带 Cookie 中的内容 |
-| expires |  number | optional | `7200` | | CSRF Cookie 的过期时间(秒) |
-| key | string | required |  |  | 加密 token 的秘钥 |
+|   name   |  string |    可选    | `apisix-csrf-token`  |    | 生成的 Cookie 中的 token 的名字，需要使用这个名字在请求头携带 Cookie 中的内容 |
+| expires |  number | 可选 | `7200` | | CSRF Cookie 的过期时间(秒) |
+| key | string | 必选 |  |  | 加密 token 的秘钥 |
```
hf400159, line 36:
```suggestion
|   name   |  string |    否    | `apisix-csrf-token`  |    | 生成的 Cookie 中的 token 的名字，需要使用这个名字在请求头携带 Cookie 中的内容 |
| expires |  number | 否 | `7200` | | CSRF Cookie 的过期时间(秒) |
| key | string | 是 |  |  | 加密 token 的秘钥 |
```

hf400159: I think it should be changed to “是” or “否”. This is more suitable for "必选项".

liangliang4ward: @juzhiyuan what you think?  if we changed to "是" or "否"，we should changed all docs includ en&zh

juzhiyuan: I just checked Stripe and GitHub's API References, they're different from Apache APISIX API References.

I have two styles:

1. Required (必填) -> Yes, No
2. Optional (可选) -> Required, Optional

After consideration, how about using `Required` as column name? also cc @navendu-pottekkat

liangliang4ward: If we do, I'll change it all on this PR

juzhiyuan: > translate some zh/*.md table properties to chinese

Please wait for @navendu-pottekkat, @kwanhur, and @hf400159's confirmation, because if we change Chinese, we should also update English version :)

navendu-pottekkat: I would suggest Required -> Yes/No

kwanhur: No, I don't we should change them on this PR, before the document specification comes out, I prefer that keep consistent format.

kwanhur: No, I don't we should change them on this PR, before the document specification comes out, I prefer that keep consistent format.

liangliang4ward: > No, I don't we should change them on this PR, before the document specification comes out, I prefer that keep consistent format.

yes;  I aggree with you

navendu-pottekkat: I agree. And for the specification we can follow this I would suggest Required -> Yes/No.

liangliang4ward:
So let me conclude.
the parameter table head  column "is required" 
In en/*.md    is  `Required ` and value is `Yes/No`;
In zh/*.md    is  `必填` and value is `是/否`.

@kwanhur @juzhiyuan @navendu-pottekkat @hf400159 

I will close this PR. we can renew another one.

docs/zh/latest/plugins/csrf.md
juzhiyuan: OK!

PR #6670, opened 2022-03-21
feat: Add the function to hide header for key-auth plugin

bin-ya:
### Description

Add the function of hiding header for key-auth plugin

### Checklist


spacewander:
Let's merge master to make CI pass.

apisix/plugins/key-auth.lua
```diff
@@ -110,6 +114,10 @@ function _M.rewrite(conf, ctx)
     end
     core.log.info("consumer: ", core.json.delay_encode(consumer))
 
+    if conf.hide_credentials then
+        core.request.set_header(ctx, "apikey", nil)
```
tzssangglass, line 118:
```suggestion
        core.request.set_header(ctx, conf.header, nil)
```

tzssangglass:
hi @bin-ya PTAL https://github.com/apache/apisix/runs/5624966335?check_suite_focus=true

tzssangglass (approved):

docs/en/latest/plugins/key-auth.md
```diff
@@ -41,6 +41,7 @@ For route side:
 | ---- | ------ | ----------- | ------- | ----- | ---------------------------------------------------------------------------- |
 | header  | string | optional    | apikey        |       | the header we get the key from |
 | query   | string | optional    | apikey        |       | the query string we get the key from, which priority is lower than `header` |
+| hide_credentials   | bool | optional    | false        |       | Whether to pass the apikey request headers to the upstream. |
```
tokers, line 44: It's confusing, the header is not necessary to be "apikey", it's specified by the `header`. Also, the query should also be removed.

bin-ya: ok

bin-ya: Query is the content modified by others. Can I delete it?

juzhiyuan (commented):
Great! Would you like to share your Apache APISIX Development experience?

tzssangglass:
oo, another file conflict @bin-ya

docs/zh/latest/plugins/key-auth.md
```diff
@@ -41,6 +41,7 @@ router 端配置：
 | ---- | ------ | ------ | ------ | ------ | ------------------------------------------------------------------------------------------------------------- |
 | header  | string | 可选| apikey |        | 设置我们从哪个 header 获取 key。 |
 | query  | string | 可选 | apikey |        | 设置我们从哪个 query string 获取 key，优先级低于 `header` |
+| hide_credentials  | bool | 可选 | false |        | 是否将请求头传递给 upstream。 |
```
tzssangglass, line 44:
```suggestion
| hide_credentials  | bool | 可选 | false |        | 是否将含有认证信息的请求头传递给 upstream。 |
```

tzssangglass:
File conflicts have not been resolved @bin-ya

docs/zh/latest/plugins/key-auth.md
bin-ya: ok

docs/en/latest/plugins/key-auth.md
tokers: You should only handle the query that used to carry the credential.

docs/en/latest/plugins/key-auth.md
```diff
@@ -41,6 +41,7 @@ For route side:
 | ---- | ------ | ----------- | ------- | ----- | ---------------------------------------------------------------------------- |
 | header  | string | optional    | apikey        |       | the header we get the key from |
 | query   | string | optional    | apikey        |       | the query string we get the key from, which priority is lower than `header` |
+| hide_credentials   | bool | optional    | false        |       | Whether to pass the request header containing authentication information to upstream. |
```
starsz, line 44: Need to align the `|`.

starsz (approved):

apisix/plugins/key-auth.lua
```diff
@@ -110,6 +114,10 @@ function _M.rewrite(conf, ctx)
     end
     core.log.info("consumer: ", core.json.delay_encode(consumer))
 
+    if conf.hide_credentials then
```
spacewander, line 117: We should remove the key according to where the key comes from.

bin-ya: ok, I will.

apisix/plugins/key-auth.lua
```diff
@@ -110,6 +114,12 @@ function _M.rewrite(conf, ctx)
...
+        core.request.set_header(ctx, conf.header, nil)
+        core.request.get_uri_args(ctx)
+        core.request.set_uri_args(ctx, {})
```
tzssangglass, line 120: What is the purpose of doing this?

bin-ya: Hide information from query

tzssangglass: I think we don't necessary to remove the all query params
If the key is in the header, we only need to delete the key in the headers.
If the key is in the query, we only need to delete the key in the query params.

cc @spacewander right?

spacewander: Yes, we should only remove the key.

bin-ya: OK, I'll revise it right away.

tzssangglass (approved):

starsz (approved):

spacewander (approved):

PR #6686, opened 2022-03-22
feat(redirect): set redirect server port when enable http_to_https

### LINKED ISSUE #4400: request help: Can the port be set for http_to_https in the redirect plugin?
### Issue description
Can the port be set for http_to_https in the redirect plugin?

The default https port of apisix is 9443, but after I set http_to_https to true, 
visiting "http://test.com:9080/service/"
will redirect to "https://test.com/service/",
my expected to jump to "https://test.com:9443/service/"

### Environment

* apisix version (cmd: `apisix version`):2.6
* OS (cmd: `uname -a`):`Linux xxxxx 4.19.90-2012.5.0.0054.oe1.x86_64 #1 SMP Tue Dec 22 15:58:47 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
```
nginx version: openresty/1.19.3.1
built by gcc 8.3.1 20190311 (Red Hat 8.3.1-3) (GCC)
built with OpenSSL 1.1.1h  22 Sep 2020 (running with OpenSSL 1.1.1k  25 Mar 2021)
```
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
```
{"hostname":"xxxxxx","boot_time":1623142227,"etcd_version":"3.4.0","up_time":93656,"last_report_time":1623235829,"id":"xxxxx","version":"2.6"}
```
* apisix-dashboard version, if have:
* luarocks version, if the issue is about installation (cmd: `luarocks --version`):`/usr/local/bin/luarocks 3.4.0`

kwanhur:
### Description

Support set redirect server port and retrieve it from header `x-forwarded-port`, instead of only the well-known port `443`.

1. When enable APISIX default https server, its port `9443`, `x-forwarded-port` default value `$server_port` equals to `9443`, redirect to `9443` could meet the expectation.
So support to fetch the redirect server port on specified `x-forwarded-port`, it only works on enable `http_to_https` to `true`.
    - `x-forwarded-port` should be number and range 1 to 65535.

2. Fixed example `Test Plugin` code demo.

Fixes #4400 

### Checklist


tokers:
@kwanhur IMHO specify a port is really hacky, what if users say we need to customize the host in the Location, instead of the one passed from the client? Do we need to add another `ret_host` field?

Why not extend the `uri` field, let it also be effective in the `http_to_https` mode, so that we can custom the Location.

kwanhur:
From plugin redirect docs, [rule 1](https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/redirect.md)
> Only one of `http_to_https`, `uri` or `regex_uri` can be specified.

1. Under `uri` mode, it had supported to specify redirect port or host, like [test case 15-16](https://github.com/apache/apisix/blob/master/t/plugin/redirect.t#L346-L394).

    - specify redirect port just update plugin's uri like(the other field is also the same solution)
```shell
 "plugins": {
        "redirect": {
        "uri": "https://$host:9443$request_uri",
        "ret_code": 301
     }
}
```

2. Under `http_to_https` mode, extend `uri` field, it seems to break the rule 1 and bring a big change.

To extend the other field, what about new attribute `location`?
- If empty, its default value `https://$host$request_uri`.
- If not empty, it goes to the same with `uri`.

tokers:
> Under http_to_https mode, extend uri field, it seems to break the rule 1 and bring a big change.

It's a big change but not broken? Let's hear more voices from the community :)

leslie-tsang:
> I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

Looks like a breaking change here, discussion in mail list first in needed.

tokers:
@kwanhur Any updates?

tokers:
> 

Indeed it's not a broken change, [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) is an identifier for a resource, not a component of a URL (e.g., the path). Could you explain the details about the broken change?

Oh, by the way, maybe we should talk about this on the mailing list.

kwanhur:
[test case 21](https://github.com/apache/apisix/blob/master/t/plugin/redirect.t#L487) shows the incompatibility
https://github.com/apache/apisix/blob/dbe7eeebba06229d4a8df75263f2a78301cc1ca0/t/plugin/redirect.t#L487

https://github.com/apache/apisix/blob/dbe7eeebba06229d4a8df75263f2a78301cc1ca0/t/plugin/redirect.t#L515-L516

tokers:
> 

Thanks, that really change the rule, but on the other hand, even if we remove this limitation, all configurations set before can still be applied successfully in the new version. @spacewander What do you think?

kwanhur:
We also need to take some compatibility points on `admin-api` `dashboard` `ingress` etc when removing the limitation.

tokers:
> We also need to take some compatibility points on `admin-api` `dashboard` `ingress` etc when removing the limitation.

All of them are on the caller's side, could you give a specific example of the incompatibility?

spacewander:
> > 
> 
> Thanks, that really change the rule, but on the other hand, even if we remove this limitation, all configurations set before can still be applied successfully in the new version. @spacewander What do you think?

I can't catch up with the point. The new field is optional and it is not a break change?

tokers:
@kwanhur Another way is that we can check the X-Forwarded-Port in this plugin.

kwanhur:
Have a glance at `x-forwarded-port` definition, it's default value `$server_port` which default is `9443`, using it means that the default redirect server port will be `$server_port`, not now the well-known port `443`.

Using it means need to update the existed test cases, please note that.

https://github.com/apache/apisix/blob/4afc8a7fbf82d4311504abd00f08156a806e5b40/apisix/cli/ngx_tpl.lua#L668

tokers:
> Just like the test log [here](https://gist.github.com/kwanhur/edb695c852cce66e361386a6c4cbb810), many cases failed with inconsistent port, `X-Forwarded-Port` from `$server_port` in `t/APISIX.pm` is `1984` but the original expected well-known `443`.
> 
> How about updating these test cases' expected redirect server port? @tokers

I think so, if the downstream really carries the `X-Forwarded-Port`.

kwanhur:
Three scenarios will force `x-forwarded-port` value into well-known port `443`:

1. non-number
2. less than and zero
3. greater than 65535

docs/en/latest/plugins/redirect.md
```diff
@@ -45,6 +45,8 @@ The `redirect` Plugin can be used to configure redirects.
 
 Only one of `http_to_https`, `uri` and `regex_uri` can be configured.
 
+* When enable `http_to_https`, you can specify redirect server port with the header `X-Forwarded-Port`.
```
spacewander, line 48:
```suggestion
* When enabling `http_to_https`, the port in the redirect URL will be the value of header `X-Forwarded-Port` or the port of the server.
```

docs/zh/latest/plugins/redirect.md
```diff
@@ -32,19 +32,21 @@ description: 本文介绍了关于 Apache APISIX `redirect` 插件的基本信
...
+| regex_uri           | array[string] | 是   |       |              | 将来自客户端的 URL 与正则表达式匹配并重定向。当匹配成功后使用模板替换发送重定向到客户端，如果未匹配成功会将客户端请求的 URI 转发至上游。 和 `regex_uri` 不可以同时存在。例如：["^/iresty/(.)/(.)/(.*)","/$1-$2-$3"] 第一个元素代表匹配来自客户端请求的 URI 正则表达式，第二个元素代表匹配成功后发送重定向到客户端的 URI 模板。 |
+| ret_code            | integer       | 是   | 302   | [200, ...]   | HTTP 响应码                                                                                                                                                                                            |
+| encode_uri          | boolean       | 是   | false | [true,false] | 当设置为 `true` 时，对返回的 `Location` Header 按照 [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)的编码格式进行编码。                                                                                          |
+| append_query_string | boolean       | 是   | false | [true,false] | 当设置为 `true` 时，将原始请求中的查询字符串添加到 `Location` Header。如果已配置 `uri` 或 `regex_uri` 已经包含查询字符串，则请求中的查询字符串将附加一个`&`。如果你已经处理过查询字符串（例如，使用 NGINX 变量 `$request_uri`），请不要再使用该参数以避免重复。                                 |
 
 :::note
 
 `http_to_https`、`uri` 和 `regex_uri` 只能配置其中一个属性。
 
+* 当开启 `http_to_https`，你可以设置请求头 `X-Forwarded-Port` 指定重定向的服务器端口。
```
spacewander, line 48:
```suggestion
* 当开启 `http_to_https` 时，重定向 URL 中的端口将是 `X-Forwarded-Port` 请求头的值或服务器的端口。
```

spacewander (approved):

kwanhur:
Workflow failed on `t/plugin/api-breaker.t`, not the plugin redirect.
The administrator, please trigger it to rerun the cases, thanks.

apisix/plugins/redirect.lua
```diff
@@ -157,7 +157,7 @@ function _M.rewrite(conf, ctx)
     if conf.http_to_https and _scheme == "http" then
         -- TODO： add test case
         -- PR: https://github.com/apache/apisix/pull/1958
-        if ret_port == 443 then
+        if ret_port == 443 or ret_port == 0 then
```
tokers: Why check if it's equal to `0`? may be `if not ret_port or ret_port == 443`?

kwanhur: `0` can pass through `x-forwarded-port`

checking condition `ret_port == nil or ret_port == 443 or ret_port <= 0 or ret_port > 65535`

spacewander (approved):

tokers (approved):

apisix/plugins/redirect.lua
tokers: Gotcha.

tzssangglass (approved):

apisix/plugins/redirect.lua
```diff
@@ -147,6 +148,7 @@ function _M.rewrite(conf, ctx)
     core.log.info("plugin rewrite phase, conf: ", core.json.delay_encode(conf))
 
     local ret_code = conf.ret_code
+    local ret_port = tonumber(ctx.var["var_x_forwarded_port"])
```
crazyMonkey1995, line 151: Recently, I upgraded apisix with the latest code, and found that there was an error in converting http to https. When redirecting, I always bring a port 80. It should be caused by this change, because I am listening on port 80 and port 443, but ngx_tpl There is a piece of code in it that is "set $var_x_forwarded_port $server_port;", which means that the port from http to https is always the same. I don't think this change is very reasonable?

kwanhur: Oh, seems I got what you mean. The server port `80` and then enable `http_to_https`, it'll always redirect to `https://host:80` then get `400 bad request`. cc @tokers @spacewander @tzssangglass 

We need to take more consideration about https://github.com/apache/apisix/pull/6686#issuecomment-1113004299

tzssangglass: Can we set a default value for this case?
If there is no `X-Forwarded-Port` in clinet reuqest headers, then the `redirect` plugin use `443` as the default https port?

kwanhur: It has default as the server_port

https://github.com/apache/apisix/blob/1b5c1900da517bb7d34d4b221aace7242345e78a/apisix/cli/ngx_tpl.lua#L668

As mentioned before https://github.com/apache/apisix/pull/6686#issuecomment-1112812875

tzssangglass: > It has default as the server_port

I mean in the `redirect` plugin, if someone enable `http_to_https` and the client request doesn't pass X`-Forwarded-Port`, use 443 as the default port in `Location` and don't care about the `server_port` variable.

kwanhur: As the conversation before https://github.com/apache/apisix/pull/6686#issuecomment-1075798013

The original implement with an extra attribute `ret_port`, but no acceptable.

spacewander: Maybe we can read the correct port from `config.yaml`?

atlassianss: 把这个set $var_x_forwarded_port       $server_port;   改成set $var_x_forwarded_port       443;

PR #6698, opened 2022-03-23
docs: fix table garbled

hf400159:
In the Apache APISIX official website, this table format error, because there is no blank line between the table and the text.

image


liangliang4ward:
hi. In this https://github.com/apache/apisix/pull/6668/files#diff-379fb7b0afff59b9ef8020983270b775f669391bb7d0f585ad9fd32c462f9dd9

the table head I already fixed.  

and "可选" -> “否”，I think we should after unify document style to modify.

@juzhiyuan What do you think?

hf400159:
> hi @hf400159 . In this https://github.com/apache/apisix/pull/6668/files#diff-379fb7b0afff59b9ef8020983270b775f669391bb7d0f585ad9fd32c462f9dd9
> 
> the table head I already fixed.
> 
> and "可选" -> “否”，I think we should after unify document style to modify all docs
> 
> @juzhiyuan What do you think?

ok，I will close

Issue #3039, opened 2020-12-14
request help: question about time zone


ziyou434:
### Issue description
I find there are some code about time zone，but I don't know what they can do or where they can be use.
```
containers:
        - env:
            - name: TZ
              value: "Asia/Shanghai"

...

hostPath:
            path: /etc/localtime
            type: File
          name: localtime
```
My company is sensitive to time zone，I want to insure these configs  have no effect .

### Environment

* apisix version (cmd: `apisix version`): 2.0
* OS: linux

spacewander:
```
hostPath:
            path: /etc/localtime
            type: File
          name: localtime
```

This part of code mounts the host's /etc/localtime to the pod. Quoted from the doc:
> If the TZ environment variable does not have a value, the operation chooses a time zone by default. In the GNU C Library, the default time zone is like the specification ‘TZ=:/etc/localtime’ (or ‘TZ=:/usr/local/etc/localtime’, depending on how the GNU C Library was configured; see Installation). Other C libraries use their own rule for choosing the default time zone, so there is little we can say about them.

https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

```
containers:
        - env:
            - name: TZ
              value: "Asia/Shanghai"
```
I think we should remove this, since it is bad idea to hard code timezone. 
@membphis @gxthrj What about your guys' opinion?

moonming:
agree, we can not hard code TZ

PR #6705, opened 2022-03-24
docs: FAQ add items for modifying the timezone

Gallardot:
### Description

ref https://github.com/apache/apisix-helm-chart/pull/232

### Checklist

### DIFF
--- a/docs/zh/latest/FAQ.md
+++ b/docs/zh/latest/FAQ.md
@@ -566,3 +566,23 @@ apisix:
 
  - 插件实例作用范围：`plugin-metadata` 作用于该插件的所有配置实例。`plugin-configs` 作用于其下配置的插件配置实例。
  - 绑定主体作用范围：`plugin-metadata` 作用于该插件的所有配置实例绑定的主体。`plugin-configs` 作用于绑定了该 `plugin-configs` 的路由。
+
+## 如何修改APISIX的时区
+
+你可能需要将APISIX的时区设置为与你本地时区一致，否则在检索APISIX日志的时候会得到不一致的时区的日志数据，特别在你使用apisix日志插件的时候。
+
+如若你是通过 RPM 仓库或者源码方式安装部署，你需要在服务器上执行如下的命令再启动 APISIX 。
+```shell
+export TZ='Asia/Shanghai'
+apisix start
+```
+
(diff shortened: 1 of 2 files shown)


docs/zh/latest/FAQ.md
```diff
@@ -566,3 +566,24 @@ apisix:
 
  - 插件实例作用范围：`plugin-metadata` 作用于该插件的所有配置实例。`plugin-configs` 作用于其下配置的插件配置实例。
  - 绑定主体作用范围：`plugin-metadata` 作用于该插件的所有配置实例绑定的主体。`plugin-configs` 作用于绑定了该 `plugin-configs` 的路由。
+
+## 如何修改APISIX的时区
```
juzhiyuan, line 570:
```suggestion
## 如何修改 Apache APISIX 的时区
```

docs/zh/latest/FAQ.md
```diff
@@ -566,3 +566,24 @@ apisix:
...
+
+你可能需要将 APISIX 的时区设置为与你本地时区一致，否则在检索 APISIX 日志的时候会得到不一致的时区的日志数据，特别在你使用 APISIX 日志插件的时候。
```
juzhiyuan, line 572: Hi, just to make sure: is it's necessary to use the same timezone as local? If yes, then please use `必须` or `应当` instead of `可能`.

juzhiyuan (commented):
BTW, please update this PR's checklist, thanks!

docs/zh/latest/FAQ.md
Gallardot: > Hi, just to make sure: is it's necessary to use the same timezone as local? If yes, then please use `必须` or `应当` instead of `可能`.

@juzhiyuan 
I think it depends on the user's actual situation. After all, in the current APISIX Docker image, the default time zone is UTC, but our own scene needs to use the same timezone as local. To make it easier to view the logs.

So I think `可能` is OK.

juzhiyuan: Just a little confusing for me, cc @bzp2010 to have a check. 😅

moonming:
the same as https://github.com/apache/apisix/issues/3039

moonming:
I am not sure we need to add this to FAQ for Apache APISIX

Gallardot:
> I am not sure we need to add this to FAQ for Apache APISIX

@tokers PTAL.

PR #6807, opened 2022-04-08
fix: inappropriate default timeout of syslog schema

nic-6443:
### Description

We use 3 as the default timeout configuration in the syslog plugin. This configuration will eventually be used as the timeout configuration of the lua-resty-logger-socket library, and its unit is milliseconds. 

It is obviously inappropriate to use 3 milliseconds as the default configuration, because it is easy to trigger a timeout. I think the author of the syslog plugin should have mistakenly thought that the timeout unit in lua-resty-logger-socket is seconds, so 3 is used as the default value.

So this PR modifies the default value to 3000, and clarifies the timeout unit in the documentation.

https://github.com/apache/apisix/blob/51eec3ea64bba464ab4a6741acc318c502cab26c/apisix/plugins/syslog.lua#L36

https://github.com/apache/apisix/blob/51eec3ea64bba464ab4a6741acc318c502cab26c/apisix/plugins/syslog.lua#L90-L101

https://github.com/cloudflare/lua-resty-logger-socket/blob/master/README.md?plain=1#L102-L104
> timeout
>
> Sets the timeout (in ms) protection for subsequent operations, including the connect method. Default value is 1000 (1 sec).

### Checklist


spacewander (dismissed):

spacewander (approved):

starsz (approved):
Good capture.

leslie-tsang (approved):

shuaijinchao (approved):

PR #6811, opened 2022-04-08
fix: add support for multi-header with same key

rampagecong:
### Description

add support for multi-header with same key like `set-cookie`

Fixes # (issue)
https://github.com/apache/apisix-go-plugin-runner/issues/74
### Checklist

### DIFF
--- a/t/lib/ext-plugin.lua
+++ b/t/lib/ext-plugin.lua
@@ -367,12 +367,14 @@ function _M.go(case)
             local action = http_req_call_rewrite.End(builder)
             build_action(action, http_req_call_action.Rewrite)
 
-        elseif case.rewrite_resp_header == true or case.rewrite_vital_resp_header == true then
+        elseif case.rewrite_resp_header == true or case.rewrite_vital_resp_header == true or case.rewrite_same_resp_header then
             local hdrs = {
                 {"X-Resp", "foo"},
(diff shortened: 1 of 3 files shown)


rampagecong:
@spacewander Could you please take a look?

t/lib/ext-plugin.lua
```diff
@@ -373,6 +373,40 @@ function _M.go(case)
                 {"X-Req", "bar"},
                 {"Content-Type", "application/json"},
                 {"Content-Encoding", "deflate"},
+                {"X-Same", "one"},
```
spacewander, line 376: We don't need X-Same for other header tests?

rampagecong: Sorry, I optimized the code.I divided it into three test cases based on this feature.
TEST 19: rewrite response header and call the upstream service
TEST 20: rewrite non-important response headers and call the upstream service
TEST 25: rewrite same response headers and call the upstream service

spacewander (approved):

t/lib/ext-plugin.lua
spacewander: Don't worry, it's just a small problem.

spacewander:
@rampagecong 
Could you apply the same fix to other headers? Thanks!

PR #6822, opened 2022-04-11
feat: check plugin init type

tylitianrui:
### Description
- check  the  type  of  `plugin.init` whether it  is  `function` or  not.   `plugin.init`  may  be defined as  other  data types 
 in custom plugins
-  remove  redundant  code

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -150,10 +150,13 @@ local function load_plugin(name, plugins_list, plugin_type)
     core.table.insert(plugins_list, plugin)
 
     if plugin.init then
-        plugin.init()
+        if  type(plugin.init) == "function" then
+            plugin.init()
+        else
+            core.log.warn("plugin [", name,
+                           "]: field 'init' is not  a function")
+        end
     end
-
-    return
 end
 
 


tokers:
@tylitianrui The issue link is broken, could you fix it?

apisix/plugin.lua
```diff
@@ -150,10 +150,13 @@ local function load_plugin(name, plugins_list, plugin_type)
     core.table.insert(plugins_list, plugin)
 
     if plugin.init then
-        plugin.init()
+        if  type(plugin.init) == "function" then
+            plugin.init()
```
soulbird, line 154: Why do we need this?

tylitianrui: check type  of `plugin.init`.   customer plugin may  define it  as other  data  types

spacewander: We can add it in https://github.com/apache/apisix/blob/761c34a292d187fbc49ddb06f1be0d7790df3815/apisix/plugin.lua#L114 and reject invalid plugin

tzssangglass:
> plugin.init may be defined as other data types in custom plugins

Frankly, I don't think it makes sense.

If init needs to be checked, then `rewrite`, `access`, etc. all need to be checked.

PR #6862, opened 2022-04-15
fix(grpc-transcode): fetch the response body cached in the ctx

### LINKED ISSUE #6848: bug: http-logger 和grpc-transcode  同时使用BUG
### Current Behavior

http-logger 开启 include_resp_body = true 记录返回body 
同时
使用 grpc-transcode 
接口返回数据错错误 表现为返回 proto 定义的默认值

### Expected Behavior

正常解析grpc

### Error Logs

_No response_

### Steps to Reproduce

1.开启插件 http-logger 开启 include_resp_body = true 记录返回body 
2.编写一个grpc 服务通过 grpc-transcode解析
3.  请求后返回默认值
4.  插件 http-logger 设置 include_resp_body = false  请求返回正常

详细步骤,代码:https://github.com/aixichen/apisix_bug/blob/master/README.md

### Environment

- APISIX version (run `apisix version`):2.13.0
- Operating system (run `uname -a`):root

tzssangglass:
Closes #6848

### Description
When the logger-* plugins works on global rule, then they would read response body first, and then the `grpc-transcode` plugin would read the body failed second.

Fixes # (issue)

### Checklist

We need to keep an eye on all plugins that use `core.response.hold_body_chunk(ctx)`, they may have similar issuess so that I can handle with different PRs?


tzssangglass:
## bug cause

when the logger plugin is bound to global_rule and the grpc-transcode is bound to route

1. in the first `body_filter` phase, run the logger plugin which is bound to global_rule first to get the response body by: https://github.com/apache/apisix/blob/605c2c40c46dcd4cd3a4d551423345faa01b1a72/apisix/utils/log-util.lua#L255
2. then run `grpc-transcode` on route and get the response body by: https://github.com/apache/apisix/blob/605c2c40c46dcd4cd3a4d551423345faa01b1a72/apisix/plugins/grpc-transcode/response.lua#L25
3. in the second `body_filter` phase,  run the logger plugin which is bound to global_rule, it would got `eof` is true and make `ctx._body_buffer = nil` see:https://github.com/apache/apisix/blob/605c2c40c46dcd4cd3a4d551423345faa01b1a72/apisix/core/response.lua#L180-L189
4. in the second `body_filter` phase,  run `grpc-transcode` on route and get the response body, it would got `eof` is true, and got `ctx._body_buffer` is nil that changed by the logger plugin, the got `chunk` is an empty string.

tzssangglass:
The current fix only makes use of the response body's cache, which may not be the most appropriate solution, as the logger class plugin is configured in global_rule
and route will result in different logging content, because if logger plugin configured on route, `grpc-transcode` is executed before logger plugin, so `grpc-transcode` decode the content from grpc to text by: https://github.com/apache/apisix/blob/605c2c40c46dcd4cd3a4d551423345faa01b1a72/apisix/plugins/grpc-transcode/response.lua#L59, then logger plugin got the decoded `ngx.arg[1]`

t/plugin/grpc-transcode-with-logger-plgin.t
```diff
@@ -0,0 +1,232 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+BEGIN {
+    if ($ENV{TEST_NGINX_CHECK_LEAK}) {
+        $SkipReason = "unavailable for the hup tests";
+
+    } else {
+        $ENV{TEST_NGINX_USE_HUP} = 1;
```
spacewander, line 22: Look like TEST_NGINX_USE_HUP is unnecessary? We can merge it into t/plugin/grpc-transcode2.t

spacewander (approved):

shuaijinchao (approved):

leslie-tsang (approved):

PR #6908, opened 2022-04-22
fix: upstream nodes cant't include ipv6 addr

### LINKED ISSUE #6905: bug: host schema validation doesn't support ipv6
### Current Behavior

image

### Expected Behavior

_No response_

### Error Logs

_No response_

### Steps to Reproduce

```
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "uri": "/*",
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "[::1]:8882": 1
        }
    },
     "plugins":{"example-plugin":{"i": 123, "ip": "[::1]", "port": 8881}}
}'
```

### Environment

- APISIX version (run `apisix version`): 2.10.1

soulbird:
### Description
Make a simple modification to make the upstream nodes host support ipv6

Fixes #6905 

### Checklist

### DIFF
--- a/apisix/schema_def.lua
+++ b/apisix/schema_def.lua
@@ -37,7 +37,7 @@ local id_schema = {
     }
 }
 
-local host_def_pat = "^\\*?[0-9a-zA-Z-._]+$"
+local host_def_pat = "^\\*?[0-9a-zA-Z-._\\[\\]:]+$"
 local host_def = {
     type = "string",
     pattern = host_def_pat,
--- a/t/APISIX.pm
+++ b/t/APISIX.pm
@@ -200,7 +200,7 @@ if ($version =~ m/\/apisix-nginx-module/) {
     $a6_ngx_directives = <<_EOC_;
     apisix_delay_client_max_body_check on;
     apisix_mirror_on_demand on;
-    wasm_vm wasmtime;
+    #wasm_vm wasmtime;
 _EOC_
 }
 
--- a/t/APISIX.pm
+++ b/t/APISIX.pm
@@ -200,7 +200,7 @@ if ($version =~ m/\/apisix-nginx-module/) {
     $a6_ngx_directives = <<_EOC_;
     apisix_delay_client_max_body_check on;
     apisix_mirror_on_demand on;
-    #wasm_vm wasmtime;
+    wasm_vm wasmtime;
 _EOC_
 }
 
--- a/t/admin/upstream-array-nodes.t
+++ b/t/admin/upstream-array-nodes.t
@@ -476,11 +476,12 @@ GET /t
                 )
 
             ngx.status = code
-            ngx.print(body)
+            ngx.say(body)
         }
     }
 --- request
 GET /t
---- error_code: 201
+--- response_body
+passed
 --- no_error_log
 [error]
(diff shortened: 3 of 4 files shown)


apisix/schema_def.lua
```diff
@@ -37,7 +37,7 @@ local id_schema = {
     }
 }
 
-local host_def_pat = "^\\*?[0-9a-zA-Z-._]+$"
+local host_def_pat = "^\\*?[0-9a-zA-Z-._\\[\\]:]+$"
```
zhendongcmss, line 40: Can validate ipv4/ipv6 more detail ?

t/APISIX.pm
```diff
@@ -200,7 +200,7 @@ if ($version =~ m/\/apisix-nginx-module/) {
     $a6_ngx_directives = <<_EOC_;
     apisix_delay_client_max_body_check on;
     apisix_mirror_on_demand on;
-    wasm_vm wasmtime;
+    #wasm_vm wasmtime;
```
spacewander, line 203: Why remove that?

apisix/schema_def.lua
spacewander: I would prefer to a simple regex check as a complex check is error-prone (may reject valid input)

membphis: IPv6 schema: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L56

soulbird: > IPv6 schema: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L56
> 
> can we reuse it?

Let my try

t/plugin/referer-restriction.t
```diff
@@ -173,7 +173,6 @@ hello world
             local cases = {
                 "x.*",
                 "~y.xn",
-                "::1",
```
spacewander, line 176: The `::1` is an invalid referer.

apisix/schema_def.lua
zhendongcmss: My point that it is hard to use the same regular exprssion to valitate ipv4 and ipv6.

t/plugin/referer-restriction.t
soulbird: In fact, referer is a URL, and the current shcema check is not suitable. Maybe we need another PR to fix this

apisix/schema_def.lua
soulbird: > I would prefer to a simple regex check as a complex check is error-prone (may reject valid input)

So I'm leaning towards a more conservative way of fixing it, like this currently. @zhendongcmss

spacewander (approved):

t/admin/upstream-array-nodes.t
```diff
@@ -450,3 +450,38 @@ GET /t
...
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                    "upstream": {
+                        "nodes": [
+                            {
+                                "host":"[::1]",
```
bzp2010, line 467: Is it valid only when using object's node? What happens if I use an array pattern like `[::1]:8080`?

soulbird: It will not affect object's node, see here: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L308

bzp2010 (approved):

spacewander (approved):

PR #6925, opened 2022-04-25
chore: replace ngx.decode_args to core.string.decode_args

soulbird:
### Description

Replace replace ngx.decode_args to core.string.decode_args, so that we can support more than 100 raw args.

### Checklist

### DIFF
--- a/apisix/core/string.lua
+++ b/apisix/core/string.lua
@@ -105,6 +105,15 @@ function _M.compress_script(s)
 end
 
 
+---
+-- Decodes a URI encoded query-string into a Lua table.
+-- All request arguments received will be decoded by default.
+--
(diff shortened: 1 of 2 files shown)


apisix/core/string.lua
```diff
@@ -103,4 +105,15 @@ function _M.compress_script(s)
 end
 
 
+function _M.decode_args(args)
```
spacewander, line 108: Let's add doc like https://github.com/apache/apisix/blob/fd9488da7c72f06bf102dad20b3c7dee19b0b40c/apisix/core/request.lua#L89

tzssangglass (approved):

spacewander (approved):

PR #6934, opened 2022-04-26
fix(balancer): Add test cases  when setting upstream doesn't specify priority

zhendongcmss:
### Description

The `transform_node` function doesn't check the priority. If the priority is `nil`, apisix will print a error.

image

### Checklist

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -50,12 +50,17 @@ local function transform_node(new_nodes, node)
         new_nodes._priority_index = {}
     end
 
-    if not new_nodes[node.priority] then
-        new_nodes[node.priority] = {}
-        core.table.insert(new_nodes._priority_index, node.priority)
+    local node_priority = 0   -- if nodes don't specify priority using default
+    if node.priority then
+        node_priority = node.priority
     end
 
-    new_nodes[node.priority][node.host .. ":" .. node.port] = node.weight
+    if not new_nodes[node_priority] then
+        new_nodes[node_priority] = {}
+        core.table.insert(new_nodes._priority_index, node_priority)
+    end
+
+    new_nodes[node_priority][node.host .. ":" .. node.port] = node.weight
     return new_nodes
 end
 
--- a/t/admin/balancer.t
+++ b/t/admin/balancer.t
@@ -251,3 +251,118 @@ host: 39.97.63.217 count: 4
 host: 39.97.63.218 count: 12
 --- no_error_log
 [error]
+
+
+
+=== TEST 6: don't specify the priority
+--- config
+    location /t {
+        content_by_lua_block {
+            local up_conf = {
+                type = "roundrobin",
+                nodes = {
+                    {host = "39.97.63.215", port = 80, weight = 1},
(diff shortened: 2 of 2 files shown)


zhendongcmss:
> hi @zhendongcmss we need test cases to cover this.

will update

tzssangglass:
I'm curious about the trigger condition because priority is set as a default in the JSON schema.
see: https://github.com/apache/apisix/blob/e56d23950ff18f7656a605ad80b5c9a2576d3c01/apisix/schema_def.lua#L331-L335

zhendongcmss:
> I'm curious about the trigger condition because priority is set as a default in the JSON schema. see:
> 
> https://github.com/apache/apisix/blob/e56d23950ff18f7656a605ad80b5c9a2576d3c01/apisix/schema_def.lua#L331-L335

I know that.
But, Somtimes, we develop a new plugin on access phase  and use `upstream.set(ctx, "xxxx", "xxx", up_conf)` to rewrite the upstreams, if the `up_conf` doesn't include the priority(want to use the default priority) will get a error.

tzssangglass:
> I know that.
> But, Somtimes, we develop a new plugin on access phase and use `upstream.set(ctx, "xxxx", "xxx", up_conf)` to rewrite the upstreams, if the `up_conf` doesn't include the priority(want to use the default priority) will get an error.

I don't really agree with the way this was handled. I think it's too late to check the parameters here.

In fact, for all configurations with default attributes that are not verified by the json schema, there are similar problems?

zhendongcmss:
> schema

I think it is necessary to check the passed  parameter to aviod thread crash error.

tzssangglass:
Let's hear what others think.

apisix/balancer.lua
```diff
@@ -50,12 +50,17 @@ local function transform_node(new_nodes, node)
         new_nodes._priority_index = {}
     end
 
-    if not new_nodes[node.priority] then
-        new_nodes[node.priority] = {}
-        core.table.insert(new_nodes._priority_index, node.priority)
+    local node_priority = 0   -- if nodes don't specify priority using default
+    if node.priority then
```
spacewander, line 54: Is it possible to call https://github.com/apache/apisix/blob/353e10b60e3604f80bae173a07c3c570f38ae367/apisix/upstream.lua#L179?

We already have a fixing logic in https://github.com/apache/apisix/blob/353e10b60e3604f80bae173a07c3c570f38ae367/apisix/upstream.lua#L219.

It will be terrible to fill a default value in three places.

zhendongcmss: If you don't agree with that add the priority. I think it is necessary to add test cases to test it.

zhendongcmss:
@tzssangglass @spacewander please review

spacewander:

zhendongcmss:

There no default priority checking the test cases don't pass.   How about you that the test cases are necessary and reasonable ？

The  test cases call test->balancer.pick_sever -> create_server_picker -> fetch_health_nodes, there are no default priority check.

I think it is reasonable because some times we call the API function not only using jsom conf.

PR #6949, opened 2022-04-27
api-breaker plugin returns default response body

### LINKED ISSUE #6923: feat: As a user, I want to let api-breaker plugin return default response body
### Description

refer to the document [api-breaker](https://apisix.apache.org/docs/apisix/plugins/api-breaker)
when upstream is in the unhealthy state, the api-breaker returns unhealthy.http_statuses only, without a response body.
In order to be more compatible to the client, return a default response body is useful.

qihaiyan:
### Description

When upstream is in the unhealthy state, the api-breaker returns unhealthy.http_statuses only, without a response body.
In order to be more compatible to the client, return a default response body is useful.

Fixes #6923

### Checklist

### DIFF
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -46,10 +46,12 @@ local schema = {
                 type = "object",
                 properties = {
                     key = {
-                        type = "string"
+                        type = "string",
+                        minLength = 1
                     },
                     value = {
-                        type = "string"
+                        type = "string",
+                        minLength = 1
                     }
                 }
             }
@@ -178,8 +180,8 @@ function _M.access(conf, ctx)
     -- breaker
     if lasttime + breaker_time >= ngx.time() then
         if conf.break_response_body then
-            if conf.headers then
-                for _, value in ipairs(conf.headers) do
+            if conf.break_response_headers then
+                for _, value in ipairs(conf.break_response_headers) do
                     local val = core.utils.resolve_var(value.value, ctx.var)
                     core.response.add_header(value.key, val)
                 end
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -40,7 +40,7 @@ local schema = {
         break_response_body = {
             type = "string"
         },
-        headers = {
+        break_response_headers = {
             type = "array",
             items = {
                 type = "object",
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -132,12 +132,7 @@ local _M = {
 
 
 function _M.check_schema(conf)
-    local ok, err = core.schema.check(schema, conf)
-    if not ok then
-        return false, err
-    end
-
-    return true
+    return core.schema.check(schema, conf)
 end
 
 
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -20,6 +20,8 @@ local plugin_name = "api-breaker"
 local ngx = ngx
 local math = math
 local error = error
+local ipairs = ipairs
+
 
 local shared_buffer = ngx.shared["plugin-".. plugin_name]
 if not shared_buffer then
@@ -38,9 +40,19 @@ local schema = {
         break_response_body = {
             type = "string"
         },
-        break_response_content_type = {
-            type = "string",
-            default = "application/json"
+        headers = {
+            type = "array",
+            items = {
+                type = "object",
+                properties = {
+                    key = {
+                        type = "string"
+                    },
+                    value = {
+                        type = "string"
+                    }
+                }
+            }
         },
         max_breaker_sec = {
             type = "integer",
@@ -120,7 +132,12 @@ local _M = {
 
 
 function _M.check_schema(conf)
-    return core.schema.check(schema, conf)
+    local ok, err = core.schema.check(schema, conf)
+    if not ok then
+        return false, err
+    end
+
+    return true
 end
 
 
@@ -166,9 +183,11 @@ function _M.access(conf, ctx)
     -- breaker
     if lasttime + breaker_time >= ngx.time() then
         if conf.break_response_body then
-            core.response.clear_header_as_body_modified()
-            if conf.break_response_content_type then
-                core.response.set_header("Content-Type", conf.break_response_content_type)
+            if conf.headers then
+                for _, value in ipairs(conf.headers) do
+                    local val = core.utils.resolve_var(value.value, ctx.var)
+                    core.response.add_header(value.key, val)
+                end
             end
             return conf.break_response_code, conf.break_response_body
         end
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -35,6 +35,13 @@ local schema = {
             minimum = 200,
             maximum = 599,
         },
+        break_response_body = {
+            type = "string"
+        },
+        break_response_content_type = {
+            type = "string",
+            default = "application/json"
+        },
         max_breaker_sec = {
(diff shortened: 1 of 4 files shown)


apisix/plugins/api-breaker.lua
```diff
@@ -35,6 +35,13 @@ local schema = {
             minimum = 200,
             maximum = 599,
         },
+        break_response_body = {
+            type = "string"
+        },
+        break_response_content_type = {
```
tokers, line 41: Apache APISIX has a few plugins that define the response body but without setting the content type.

I'm thinking about that, if we need to keep the behaviors consistence?

qihaiyan: If the body is json, and the content-type is text/html, some APP clients may report error, and the terminal user would be effected.

tokers: This problem cannot be solved even if we provide this field. People have to know how to configure it.

For these kinds of returning body, I think we may need a unified way to handle the `Content-Type`.

qihaiyan: > This problem cannot be solved even if we provide this field. People have to know how to configure it.
> 
> For these kinds of returning body, I think we may need a unified way to handle the `Content-Type`.

A unified way is good, but it's not that easy. Modern API returns application/json, but some legacy API returns application/soap+xml like SOAP.

apisix/plugins/api-breaker.lua
```diff
@@ -158,6 +165,11 @@ function _M.access(conf, ctx)
 
     -- breaker
     if lasttime + breaker_time >= ngx.time() then
+        if conf.break_response_body then
+            core.response.clear_header_as_body_modified()
+            ngx.header["Content-Type"] = conf.break_response_content_type
```
tzssangglass, line 170:
```suggestion
            if conf.break_response_content_type then
                core.response.set_header("Content-Type", conf.break_response_content_type)
            end
```

tzssangglass:
Hi @qihaiyan , can we add a test cases to cover `break_response_content_type` ?

qihaiyan:
> Hi @qihaiyan , can we add a test cases to cover `break_response_content_type` ?

Sure, i'll add the test cases tomorrow.

apisix/plugins/api-breaker.lua
qihaiyan: I think the API owner should be responsible to configure route on the apigageway correctly , it's part of the developing work, and only the API owner knows what the correct Content-Type should be. If there's no way to config the Content-Type correctly, when break state is open, the client APP used this API may crash because of the unexpected Content-Type.

qihaiyan:
Can you help to check why the chaos-test failed? @tzssangglass 
https://github.com/apache/apisix/runs/6222619168?check_suite_focus=true
image

spacewander:
> Can you help to check why the chaos-test failed? @tzssangglass https://github.com/apache/apisix/runs/6222619168?check_suite_focus=true image

The chaos test is not so stable. I have rerun it.

qihaiyan:
> The chaos test is not so stable. I have rerun it.
ok, thanks.

apisix/plugins/api-breaker.lua
tokers: Fair enough. But would you like to submit an issue to track the content-type field for other plugins?

qihaiyan: > Fair enough. But would you like to submit an issue to track the content-type field for other plugins?

Ok, i'll browse all the plugins which should be with a content-type field and submit an issue with a checklist.

apisix/plugins/api-breaker.lua
```diff
@@ -158,6 +165,13 @@ function _M.access(conf, ctx)
 
     -- breaker
     if lasttime + breaker_time >= ngx.time() then
+        if conf.break_response_body then
+            core.response.clear_header_as_body_modified()
```
spacewander, line 169: We don't need this if the body is not from the upstream

apisix/plugins/api-breaker.lua
spacewander: > Ok, i'll browse all the plugins which should be with a content-type field and submit an issue with a checklist.

It is not a good idea, as this check won't fix the non-open-source plugins.

One solution is to guess:
1. according to the response body: https://github.com/apache/apisix/blob/cc88a8db0ba64e1cb351e389ca5ec64c516073d6/apisix/control/router.lua#L81 or using https://github.com/spacewander/lua-resty-mime-sniff
2. according to the Accept header or the ext part of the uri

Another solution is to configure it in the route

spacewander: As for this plugin, I would suggest adding a field to add any response headers instead of just content-type.

tokers (dismissed):

apisix/plugins/api-breaker.lua
qihaiyan: > As for this plugin, I would suggest adding a field to add any response headers instead of just content-type.

Is there any other plugin witch such a field? I'm not sure of which field is suitable for representing any response headers.

qihaiyan: > As for this plugin, I would suggest adding a field to add any response headers instead of just content-type.

Can we use the ```headers``` field to represent any response headers? I found the ```echo``` plugin and also the ```response-rewrite``` plugin using this attribute. @spacewander 

Name | Type | Requirement | Default | Valid | Description
-- | -- | -- | -- | -- | --
headers | object | optional |   |   | New headers for response

spacewander: Yes. And we should use an array so multiple response headers with the same name can be used.

tzssangglass (approved):

qihaiyan:
This PR‘s status is ```Changes approved``` current now. Can i still add more commits to this PR to aim at @spacewander 's suggestion? @tzssangglass . 

> As for this plugin, I would suggest adding a field to add any response headers instead of just content-type.

spacewander (commented):
Yes, let's address https://github.com/apache/apisix/pull/6949#discussion_r867478450

apisix/plugins/api-breaker.lua
qihaiyan: removed this line(```core.response.clear_header_as_body_modified()```) in the new commit

apisix/plugins/api-breaker.lua
```diff
@@ -35,6 +37,23 @@ local schema = {
             minimum = 200,
             maximum = 599,
         },
+        break_response_body = {
+            type = "string"
+        },
+        headers = {
```
spacewander, line 43:
```suggestion
        break_response_headers = {
```
would be better

apisix/plugins/api-breaker.lua
```diff
@@ -35,6 +37,23 @@ local schema = {
...
+            type = "array",
+            items = {
+                type = "object",
+                properties = {
+                    key = {
+                        type = "string"
```
spacewander, line 49: Let's add "minLength = 1" to avoid empty string

apisix/plugins/api-breaker.lua
```diff
@@ -35,6 +37,23 @@ local schema = {
...
+                    },
+                    value = {
+                        type = "string"
```
spacewander, line 52: Ditto

apisix/plugins/api-breaker.lua
qihaiyan: fixed with a new commit

apisix/plugins/api-breaker.lua
qihaiyan: fixed with a new commit

spacewander (approved):

tzssangglass (approved):

PR #6953, opened 2022-04-27
feat(stream): port syslog plugin

tzssangglass:
### Description

Fixes # (issue)

### Checklist


apisix/stream/plugins/syslog.lua
```diff
@@ -0,0 +1,64 @@
...
+    flush_syslog = syslog.flush_syslog,
+}
+
+
+function _M.check_schema(conf)
+    return core.schema.check(schema, conf)
+end
+
+
+-- log phase in APISIX
```
spacewander, line 57: We can remove this?

apisix/stream/plugins/syslog.lua
```diff
@@ -0,0 +1,64 @@
...
+function _M.log(conf, ctx)
+    local entry = log_util.get_full_log_in_stream(ngx)
```
spacewander, line 59: I think we should only allow custom log format in the stream, to avoid keeping adding new fields to the default log. And people won't ask how to combine the custom log format with the default one

apisix/utils/batch-processor.lua
```diff
@@ -181,7 +186,7 @@ function batch_processor:push(entry)
         return
     end
 
-    if not batch_metrics and prometheus.get_prometheus() and self.name
+    if is_http and not batch_metrics and prometheus.get_prometheus() and self.name
```
spacewander, line 189: Check if prometheus exists will be more straightforward?

tzssangglass: update

apisix/stream/plugins/syslog.lua
tzssangglass: update

apisix/stream/plugins/syslog.lua
tzssangglass: rm

t/stream-plugin/syslog.t
```diff
@@ -0,0 +1,328 @@
...
+
+
+
+=== TEST 4: hit
+--- stream_request eval
+mmm
+--- stream_response
+hello world
+--- error_log
+sending a batch logs to 127.0.0.1:5044
```
spacewander, line 132: Is it possible to check the log content?

shuaijinchao (approved):

PR #6960, opened 2022-04-28
feat(xRPC): support log filter

tzssangglass:
### Description

Fixes # (issue)

### Checklist


apisix/schema_def.lua
```diff
@@ -807,6 +807,26 @@ local xrpc_protocol_schema = {
             description = "protocol-specific configuration",
             type = "object",
         },
+        logger = {
```
spacewander, line 810: Better to use array so that we can add multiple logger in the future

apisix/stream/xrpc/runner.lua
```diff
@@ -70,10 +74,54 @@ local function put_req_ctx(session, ctx)
...
+    local logger = session._route.protocol.logger
+
+    local pkg_name = "apisix.stream.plugins." .. logger.name
+    local ok, plugin = pcall(require, pkg_name)
+
+    core.ctx.set_vars_meta(ctx)
+    ctx.conf_id = tostring(logger.conf)
+    ctx.conf_type = "xrpc-logger"
+
+    if not ok then
```
spacewander, line 102: Better to check ok just after pcall

apisix/stream/xrpc/runner.lua
```diff
@@ -70,10 +74,54 @@ local function put_req_ctx(session, ctx)
...
+        core.log.error("failed to load plugin [", plugin, "] err: ", plugin)
```
spacewander, line 103:
```suggestion
        core.log.error("failed to load plugin [", logger.name, "] err: ", plugin)
```

apisix/stream/xrpc/runner.lua
```diff
@@ -70,10 +74,54 @@ local function put_req_ctx(session, ctx)
 end
 
 
+local function log_filter(session, ctx)
```
spacewander, line 77:
```suggestion
local function filter_logger(session, ctx)
```
would be better?

apisix/stream/xrpc/runner.lua
```diff
@@ -70,10 +74,54 @@ local function put_req_ctx(session, ctx)
...
+
 local function finish_req(protocol, session, ctx)
     ctx._rpc_end_time = ngx_now()
 
-    protocol.log(session, ctx)
+    local matched = log_filter(session, ctx)
+    core.log.info("log filter result: ", matched)
+
+    if matched then
+        protocol.log(session, ctx)
```
spacewander, line 121: We need to run `protocol.log` whether the logger is matched or not, otherwise the reuse logic in 
https://github.com/apache/apisix/blob/2800a06f00ff9010d893eb8c4b9c94359cc3a96b/apisix/stream/xrpc/protocols/redis/init.lua#L287 won't work.

apisix/schema_def.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/schema_def.lua
```diff
@@ -807,6 +807,28 @@ local xrpc_protocol_schema = {
...
+            type = "array",
+            items = {
+                properties = {
+                    name = {
+                        type = "string",
+                    },
+                    filter = {
+                        description = "log filter rules",
```
spacewander, line 818:
```suggestion
                        description = "logger filter rules",
```

apisix/stream/xrpc/runner.lua
spacewander: Missing this one?

apisix/stream/xrpc/runner.lua
```diff
@@ -70,9 +75,53 @@ local function put_req_ctx(session, ctx)
...
+    local pkg_name = "apisix.stream.plugins." .. logger.name
+    local ok, plugin = pcall(require, pkg_name)
+    if not ok then
+        core.log.error("failed to load plugin [", plugin, "] err: ", plugin)
+        return
+    end
+
+    core.ctx.set_vars_meta(ctx)
+    ctx.conf_id = tostring(logger.conf)
+    ctx.conf_type = "xrpc-logger"
```
spacewander, line 102: Can we run these once instead of per-logger?

t/xrpc/pingpong2.t
```diff
@@ -139,3 +151,548 @@ lua tcp socket send timeout: 60000
...
+
+
+=== TEST 6: log filter matched successful
+--- request eval
+"POST /t
+" .
+"pp\x02\x00\x00\x00\x00\x00\x00\x03ABC"
+--- stream_conf_enable
+--- error_log
+call pingpong's log
```
spacewander, line 257: This error log only means `protocol.log` is called, but not the logger.

tzssangglass: update

apisix/schema_def.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: There is a problem here, since it is possible to run the plugin in `logger`, `metrics` stage, so `conf_id`, `conf_type` can only be set here. We need to find a way to solve this problem.

apisix/stream/xrpc/runner.lua
```diff
@@ -93,11 +93,13 @@ local function run_log_plugin(ctx, logger)
...
+        core.log.error("failed to load plugin [", logger.name, "] err: ", plugin)
         return
     end
 
-    core.ctx.set_vars_meta(ctx)
+    -- we choose to initialize conf_id and conf_type here
+    -- because conf here refers specifically to the conf in the log phase,
+    -- to avoid overwriting the conf in other parts of the protocol
     ctx.conf_id = tostring(logger.conf)
     ctx.conf_type = "xrpc-logger"
```
spacewander, line 104: We should add it to https://github.com/apache/apisix/blob/1b5c1900da517bb7d34d4b221aace7242345e78a/apisix/stream/xrpc/sdk.lua#L110

t/xrpc/apisix/stream/xrpc/protocols/pingpong/init.lua
```diff
@@ -76,6 +76,12 @@ local function to_int32(p, idx)
 end
 
 
+local function init_ctx_fileds(ctx)
+    -- init some ctx fields
+    core.ctx.set_vars_meta(ctx)
```
spacewander, line 81: Ditto

apisix/stream/xrpc/runner.lua
tzssangglass: update

t/xrpc/apisix/stream/xrpc/protocols/pingpong/init.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
```diff
@@ -70,9 +74,49 @@ local function put_req_ctx(session, ctx)
 end
 
 
+local function filter_logger(ctx, logger)
+    if not logger or not logger.filter or #logger.filter == 0 then
```
spacewander, line 78: The logger can't be nil, and we should return true when filter is missing?

apisix/stream/xrpc/sdk.lua
```diff
@@ -108,6 +108,10 @@ function _M.get_req_ctx(session, id)
     local ctx = core.tablepool.fetch("xrpc_ctxs", 4, 4)
     -- fields start with '_' should not be accessed by the protocol implementation
     ctx._id = id
+    core.ctx.set_vars_meta(ctx)
+    ctx.conf_id = tostring(session._route.protocol)
```
spacewander, line 112: Better to set a per-logger conf_id because multiple instances of the same  logger plugin can be used together.

apisix/stream/xrpc/runner.lua
```diff
@@ -70,9 +74,49 @@ local function put_req_ctx(session, ctx)
...
+        return false
+    end
+
+    local expr, err = expr.new(logger.filter)
```
spacewander, line 82: Better avoid using the module name as the var name.
And let's introduce a cache to avoid creating new obj per operations in the next PR.

tzssangglass: update

apisix/stream/xrpc/sdk.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
```diff
@@ -71,9 +79,58 @@ local function put_req_ctx(session, ctx)
...
+       return false
+    end
+
+    if not logger.filter or #logger.filter == 0 then
+        -- no valid filter, default execution plugin
+        return true
+    end
+
+    -- key and version are divided by per-logger level
+    local key = "xrpc-logger" .. ctx.conf_id
```
spacewander, line 93: We can use ctx.conf_id directly?

apisix/stream/xrpc/runner.lua
```diff
@@ -71,9 +79,58 @@ local function put_req_ctx(session, ctx)
...
+    local version = tostring(logger.filter)
+    local filter_expr, err = logger_expr_cache(key, version, expr.new, logger.filter)
+    if not filter_expr or err then
```
spacewander, line 96:
```suggestion
    if not filter_expr then
```
is enough?

apisix/stream/xrpc/runner.lua
```diff
@@ -71,9 +79,58 @@ local function put_req_ctx(session, ctx)
...
+    end
+
+    if not logger.filter or #logger.filter == 0 then
+        -- no valid filter, default execution plugin
+        return true
+    end
+
+    -- key and version are divided by per-logger level
+    local key = "xrpc-logger" .. ctx.conf_id
+    local version = tostring(logger.filter)
```
spacewander, line 94: Actually, we can use the key directly as the version so we can save a `tostring`.
As each logger's conf's memory address is independent.

apisix/stream/xrpc/runner.lua
```diff
@@ -14,14 +14,22 @@
...
 local ngx_now = ngx.now
 local OK = ngx.OK
 local DECLINED = ngx.DECLINED
 local DONE = ngx.DONE
+local pcall = pcall
+local ipairs = ipairs
+local tostring = tostring
 
+local logger_expr_cache = core.lrucache.new({
+    ttl = 300, count = 32
```
spacewander, line 31: Personally speaking, what about using a larger count like 1024?

apisix/stream/xrpc/runner.lua
tzssangglass: But the meaning of `version` here should be related to `filter`. If `conf` does not change but `filter` changed, then `version` should change and a new expr should be generated?

However, as currently designed, if the `conf` is changed via the admin api, the `filter` will also change and use the new memory address.

I was trying to make the place more intuitive.

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
```diff
@@ -90,7 +90,7 @@ local function filter_logger(ctx, logger)
     end
 
     -- key and version are divided by per-logger level
-    local key = "xrpc-logger" .. ctx.conf_id
+    local key = tostring(ctx.conf_id)
```
spacewander, line 93: We can save `tostring` here as `ctx.conf_id` is already a string?

spacewander (dismissed):

apisix/stream/xrpc/runner.lua
tzssangglass: update

spacewander (approved):

PR #6968, opened 2022-04-29
feat: support hook response body for ext-plugin

soulbird:
### Description

support hook response body for ext-plugin

### Checklist


apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,103 @@
...
+        host = ctx.picked_server.host,
+        port = ctx.picked_server.port,
+    })
+
+    if not ok then
+        return nil, err
+    end
+
+    local params = {
+        path = ctx.var.uri,
```
spacewander, line 45: Should use upstream_uri

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,103 @@
...
+    return core.schema.check(_M.schema, conf)
+end
+
+
+function _M.response(conf, ctx)
+     -- TODO: request
+     local res, err = get_response(ctx)
+    if not res or err then
+        core.log.error("failed to request: ", err or "")
+        return core.response.exit(503)
```
spacewander, line 81: Use 502 would be better?

conf/config-default.yaml
```diff
@@ -397,6 +397,7 @@ plugins:                          # plugin list (sorted by priority)
   - openwhisk                      # priority: -1901
   - serverless-post-function       # priority: -2000
   - ext-plugin-post-req            # priority: -3000
+  #- ext-plugin-post-resp          # priority: -4000
```
spacewander, line 400: We can enable it by default?

apisix/plugins/ext-plugin/init.lua
```diff
@@ -682,6 +683,96 @@ local rpc_handlers = {
 
         return true
     end,
+    nil,
```
spacewander, line 686: Better to add a comment about the skipped key.

t/plugin/ext-plugin/response_body.t
```diff
@@ -0,0 +1,141 @@
...
+            ext.go({check_input = true})
+        }
+    }
+--- error_code: 200
+--- response_body
+hello world
+
+
+
+=== TEST 3: motify body
```
spacewander, line 125: Need other test cases:
1. modify header
2. modify status
3. upstream failed to response
4. runner failed to response

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,103 @@
...
+     local headers = res.headers
+     local code, body = ext.communicate(conf, ctx, name, constants.RPC_HTTP_RESP_CALL)
+     if code or body then
+         -- TODO: chunk
+         return code, body
+     end
+     core.log.info("ext-plugin will send response")
+     -- send origin response
+     -- TODO: chunk
+     core.response.set_header(headers)
```
spacewander, line 95: We should hide some headers like Nginx's proxy_hide_header directive?

apisix/plugins/ext-plugin/init.lua
```diff
@@ -682,6 +683,96 @@ local rpc_handlers = {
...
+        end
+
+        local buf = flatbuffers.binaryArray.New(resp)
+        local call_resp = http_resp_call_resp.GetRootAsResp(buf, 0)
+        local len = call_resp:HeadersLength()
+        if len > 0 then
+            local resp_headers = {}
+            for i = 1, len do
+                local entry = call_resp:Headers(i)
+                local name = str_lower(entry:Name())
```
spacewander, line 753: We should exclude some headers like https://github.com/apache/apisix/blob/cc88a8db0ba64e1cb351e389ca5ec64c516073d6/apisix/plugins/ext-plugin/init.lua#L68?

conf/config-default.yaml
soulbird: All back-to-origin logic will be hooked by `resty.http`. So I think it is not appropriate to enable this plugin by default until the relevant logic is perfect.

spacewander: Only route that contains this plugin should use `resty.http` instead.

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,180 @@
...
+
+    ngx.status = res.status
+
+    local reader = res.body_reader
+    repeat
+        local chunk, ok, read_err, print_err
+
+        chunk, read_err = reader()
+        if read_err then
+            core.error.log("read response failed: ", read_err)
```
spacewander, line 161: Ouch. We need tests to cover this branch.

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,180 @@
...
+    -- Filter out exclude_resp_header headeres
+    for k, v in pairs(res.headers) do
+        if not exclude_resp_header[str_lower(k)] then
+            core.response.set_header(k, v)
+        end
+    end
+
+    local code, body = ext.communicate(conf, ctx, name, constants.RPC_HTTP_RESP_CALL)
+    if code or body then
+        -- TODO: chunk
```
spacewander, line 147: Why do we need to handle chunk in this branch?

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,180 @@
...
+                core.error.log("output response failed: ", print_err)
+            end
+        end
+
+        if read_err or print_err then
+            return 502
+        end
+    until not chunk
+
+    ngx_flush(true)
```
spacewander, line 176: The flush should be called after print, not here

t/plugin/ext-plugin/response_body.t
```diff
@@ -0,0 +1,139 @@
+#
```
spacewander, line 1: Should call it `response.t` as it is not only for the body

t/plugin/ext-plugin/response_body.t
```diff
@@ -0,0 +1,139 @@
...
+            ext.go({check_input = true})
+        }
+    }
+--- error_code: 200
+--- response_body
+hello world
+
+
+
+=== TEST 3: motify body
```
spacewander, line 123:
```suggestion
=== TEST 3: modify body
```

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,180 @@
...
+    for k, v in pairs(res.headers) do
+        if not exclude_resp_header[str_lower(k)] then
+            core.response.set_header(k, v)
+        end
+    end
+
+    local code, body = ext.communicate(conf, ctx, name, constants.RPC_HTTP_RESP_CALL)
+    if code or body then
+        -- TODO: chunk
+        return code, body
```
spacewander, line 148: Should we close the reader here?

apisix/plugins/ext-plugin-post-resp.lua
spacewander: Also, need to log the err of ngx_flush if happened.

apisix/init.lua
```diff
@@ -471,6 +481,9 @@ function _M.http_access_phase()
     if api_ctx.dubbo_proxy_enabled then
         return ngx.exec("@dubbo_pass")
     end
+
+    -- plugin runner response call
+    ext_response_phase("response")
```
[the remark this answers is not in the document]
spacewander: We need to delay the `require` so the ext-plugin module won't be loaded if it is disabled. Or maybe there is a better way.

apisix/plugins/ext-plugin-post-resp.lua
soulbird: Now close first, keepalive will be processed later

t/plugin/ext-plugin/response.t
```diff
@@ -0,0 +1,420 @@
...
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 2: check input
+--- request
+GET /hello
```
spacewander, line 107: Is it possible to check if the args in the uri are sent to the upstream correctly?

apisix/plugins/ext-plugin/init.lua
```diff
@@ -682,6 +682,107 @@ local rpc_handlers = {
...
+        if len > 0 then
+            -- TODO: support empty body
+            body = call_resp:BodyAsString()
+        end
+        local code = call_resp:Status()
+        core.log.info("recv resp, code: ", code, " body: ", body, " len: ", len)
+
+        if code == 0 then
+            -- runner change body only, we should set code.
+            code = body and res.status or nil
```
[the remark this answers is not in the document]
spacewander: Ping @soulbird

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -0,0 +1,172 @@
...
+    end
+    core.log.info("ext-plugin will send response")
+
+    -- send origin response, status maybe changed.
+    err = send_response(res, code)
+    close(http_obj)
+
+    if err then
+        core.log.error(err)
+        return ngx.headers_sent and nil or 502
```
spacewander, line 165: 502 will always return here

apisix/plugins/ext-plugin-post-resp.lua
```diff
@@ -162,7 +162,7 @@ function _M.before_proxy(conf, ctx)
 
     if err then
         core.log.error(err)
-        return ngx.headers_sent and nil or 502
+        return not ngx.headers_sent and 502
```
spacewander, line 165: It would return `false` instead of `nil`?

tzssangglass (approved):

PR #6980, opened 2022-05-04
refactor(ops): penlight path instead of shell command mkdir

kwanhur:
### Description

Using Lua library(penlight, luafilesystem) to do file system related operations.

* If shell command mkdir not exist, command `mkdir -p path` will failed.
* Avoid `mkdir` dependency, use penlight path library, it's a dependent library of APISIX.

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -713,15 +713,14 @@ local function start(env, ...)
         util.die("Error: It is forbidden to run APISIX in the /root directory.\n")
     end
 
-    local logs = env.apisix_home .. "/logs"
-    local logs_path = pl_path.exists(logs)
-    if not logs_path then
-        local _, err = pl_path.mkdir(logs)
+    local logs_path = env.apisix_home .. "/logs"
+    if not pl_path.exists(logs_path) then
+        local _, err = pl_path.mkdir(logs_path)
         if err ~= nil then
-            util.die("failed to mkdir" .. logs .. ", error: ", err)
+            util.die("failed to mkdir ", logs_path, ", error: ", err)
         end
     elseif not pl_path.isdir(logs_path) and not pl_path.islink(logs_path) then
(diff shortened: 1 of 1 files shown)


apisix/cli/ops.lua
```diff
@@ -713,8 +713,16 @@ local function start(env, ...)
         util.die("Error: It is forbidden to run APISIX in the /root directory.\n")
     end
 
-    local cmd_logs = "mkdir -p " .. env.apisix_home .. "/logs"
-    util.execute_cmd(cmd_logs)
+    local logs = env.apisix_home .. "/logs"
```
spacewander, line 716: Using name `logs_path` would be better?

apisix/cli/ops.lua
```diff
@@ -713,8 +713,16 @@ local function start(env, ...)
...
+    local logs_path = pl_path.exists(logs)
+    if not logs_path then
```
spacewander, line 718: We can use `if not pl_path.exists`? The `exists` will return the input argument back if it exists, so there is no need to reassign it back.

spacewander (approved):

apisix/cli/ops.lua
```diff
@@ -713,8 +713,16 @@ local function start(env, ...)
...
+        local _, err = pl_path.mkdir(logs)
+        if err ~= nil then
+            util.die("failed to mkdir" .. logs .. ", error: ", err)
```
spacewander, line 721:
```suggestion
            util.die("failed to mkdir " .. logs .. ", error: ", err)
```

spacewander: BTW, it's no need to concat the params in `die`.

spacewander:
Approved the solution but some code style issues need to be addressed.

spacewander (approved):

tokers (approved):

tzssangglass (approved):

Issue #6602, opened 2022-03-14
request help: real_ip_recursive for real-ip plugin


crazyMonkey1995:
### Issue description

I found that there was once a PR about real_ip_recursive: https://github.com/apache/apisix/pull/4332
But it seems that the real-ip plugin does not have recursive related configuration items. In fact, I looked at the real-ip code and it seems that there is no related implementation. Is there any other way to achieve it? The current scenario is to configure it in the global plugin. real-ip for dynamic modification, looking forward to reply, thanks!

### Environment

- apisix version (cmd: `apisix version`): latest
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
- apisix-dashboard version, if have:
- the plugin runner version, if the issue is about a plugin runner (cmd: depended on the kind of runner):
- luarocks version, if the issue is about installation (cmd: `luarocks --version`):

tokers:
@crazyMonkey1995 I think it can be implemented in the real-ip plugin.

crazyMonkey1995:
> 

Am I missing something? But the logic here is just to return the last IP of XFF
![image](https://user-images.githubusercontent.com/21658981/158164111-a2d0b0d2-6bf6-484b-a4e6-c251ae71be61.png)

tokers:
> > 
> 
> Am I missing something? But the logic here is just to return the last IP of XFF ![image](https://user-images.githubusercontent.com/21658981/158164111-a2d0b0d2-6bf6-484b-a4e6-c251ae71be61.png) @tokers

No, the current implementation of `real-ip` plugin is not totally same to the Nginx real-ip module. We may support the `real_ip_header` and `real_ip_recursive` settings.

crazyMonkey1995:
> > > 
> > 
> > 
> > Am I missing something? But the logic here is just to return the last IP of XFF ![image](https://user-images.githubusercontent.com/21658981/158164111-a2d0b0d2-6bf6-484b-a4e6-c251ae71be61.png) @tokers
> 
> No, the current implementation of `real-ip` plugin is not totally same to the Nginx real-ip module. We may support the `real_ip_header` and `real_ip_recursive` settings.

Ok, I misunderstood.

tokers:
@crazyMonkey1995 Would you like to try it?

crazyMonkey1995:
[...]

PR #6988, opened 2022-05-06
feat(real-ip): support search recursive

crazyMonkey1995:
### Description

Support real ip recusive, see http://nginx.org/en/docs/http/ngx_http_realip_module.html

Fixes https://github.com/apache/apisix/issues/6602
### Checklist


apisix/plugins/real-ip.lua
```diff
@@ -67,6 +73,19 @@ function _M.check_schema(conf)
...
 
 
+local function addr_match(conf, addr)
+    if not conf.trusted_addresses then
+        return false
+    end
+
+    if not conf.matcher then
+        conf.matcher = core.ip.create_ip_matcher(conf.trusted_addresses)
+    end
```
tzssangglass, line 83: It is bad practice to add properties to conf, and we should keep conf pure. How about making `matcher` a module-level variable?

crazyMonkey1995: Changed it to local variable.

apisix/plugins/real-ip.lua
```diff
@@ -84,6 +104,22 @@ local function get_addr(conf, ctx)
...
         end
 
+        if conf.recursive and conf.trusted_addresses then
+            local split_addrs = {}
+            local match_itr = str_gmatch(addrs, "[^,%s*]+")
+            for itr in match_itr do
+                tb_insert(split_addrs, itr)
+            end
+
+            for i = #split_addrs, 1, -1 do
```
spacewander, line 114:
```suggestion
            for i = #split_addrs, 2, -1 do
```
We don't care if the first one can be trusted.

apisix/plugins/real-ip.lua
```diff
@@ -84,6 +104,22 @@ local function get_addr(conf, ctx)
             return addrs
         end
 
+        if conf.recursive and conf.trusted_addresses then
+            local split_addrs = {}
+            local match_itr = str_gmatch(addrs, "[^,%s*]+")
```
tzssangglass, line 109: can we use `ngx.re.gmatch` to replace `string.gmatch`

tzssangglass:
LGTM

spacewander (dismissed):

apisix/plugins/real-ip.lua
crazyMonkey1995: Solved.

apisix/plugins/real-ip.lua
```diff
@@ -106,7 +106,7 @@ local function get_addr(conf, ctx)
 
         if conf.recursive and conf.trusted_addresses then
             local split_addrs = {}
-            local match_itr = str_gmatch(addrs, "[^,%s*]+")
+            local match_itr = ngx.re.gmatch(addrs, "[^,%s*]+", "jo")
```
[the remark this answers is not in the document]
spacewander: And the return value is different.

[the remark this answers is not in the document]
crazyMonkey1995: Sorry for that, I should have tested it first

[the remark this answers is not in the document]
crazyMonkey1995: I replaced it with ngx.re.split

apisix/plugins/real-ip.lua
```diff
@@ -84,6 +103,17 @@ local function get_addr(conf, ctx)
             return addrs
         end
 
+        if conf.recursive and conf.trusted_addresses then
+            local split_addrs, _ = ngx_re_split(addrs, ",\\s*", "jo")
```
soulbird, line 107: `local split_addrs = ngx_re_split(addrs, ",\\s*", "jo")` better? Or deal with the error?

apisix/plugins/real-ip.lua
```diff
@@ -105,12 +135,8 @@ function _M.rewrite(conf, ctx)
...
-        if not conf.matcher then
-            conf.matcher = core.ip.create_ip_matcher(conf.trusted_addresses)
-        end
-
         local remote_addr = ctx.var.remote_addr
-        local trusted = conf.matcher:match(remote_addr)
+        local trusted = addr_match(conf, remote_addr)
         if not trusted then
             return
         end
```
soulbird, line 142:
```lua
if not addr_match(conf, remote_addr) then
    return
end
```

apisix/plugins/real-ip.lua
crazyMonkey1995: Thanks for the review, the above if condition has already dealt with the absence of commas, so there should be no need to deal with err here? I'm not very familiar with lua, but this `_` may indeed be removed.

tzssangglass (approved):

apisix/plugins/real-ip.lua
```diff
@@ -67,6 +73,19 @@ function _M.check_schema(conf)
 end
 
 
+local function addr_match(conf, addr)
+    if not conf.trusted_addresses then
+        return false
+    end
+
+    if not matcher then
```
spacewander, line 81: We should not use global matcher. Otherwise different configuration will use the same matcher.

crazyMonkey1995: But I defined it as a local variable above.

spacewander: Different requests with different configurations will share this as it is defined as a module variable.

crazyMonkey1995: I see, can I use lru cache to solve it? Or do you have any suggestions?

crazyMonkey1995: Use lru cache instead, refer to the implementation of ip-restriction

apisix/plugins/real-ip.lua
```diff
@@ -67,6 +75,22 @@ function _M.check_schema(conf)
 end
 
 
+local function addr_match(conf, addr)
+    if not conf.trusted_addresses then
+        return false
+    end
+
+    local matcher, err = lrucache(conf.trusted_addresses, nil,
+                                  core.ip.create_ip_matcher, conf.trusted_addresses)
```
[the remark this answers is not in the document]
spacewander: Better to use https://github.com/apache/apisix/blob/4690feb421779f5b79e8dd990dc00f4d3f1052d0/apisix/core/lrucache.lua#L171 instead.

apisix/plugins/real-ip.lua
```diff
@@ -67,6 +75,22 @@ function _M.check_schema(conf)
 end
 
 
+local function addr_match(conf, ctx, addr)
+    if not conf.trusted_addresses then
```
spacewander, line 79: We already check the trusted_addresses before: https://github.com/apache/apisix/blob/025710564ac717f99fb96dd19dc1fdd1344bc7ef/apisix/plugins/real-ip.lua#L107?

tzssangglass (approved):

spacewander (approved):

PR #7007, opened 2022-05-08
change(opentelemetry): the default network port for OTLP/HTTP is changed to 4318

soulbird:
### Description

opentelemetry's default network port for OTLP/HTTP is 4318 not 4317.

The default network port for OTLP/gRPC is 4317.

### Checklist

### DIFF
--- a/apisix/plugins/opentelemetry.lua
+++ b/apisix/plugins/opentelemetry.lua
@@ -72,7 +72,7 @@ local attr_schema = {
             type = "object",
             description = "opentelemetry collector",
             properties = {
-                address = {type = "string", description = "host:port", default = "127.0.0.1:4317"},
+                address = {type = "string", description = "host:port", default = "127.0.0.1:4318"},
                 request_timeout = {type = "integer", description = "second uint", default = 3},
                 request_headers = {
                     type = "object",
@@ -82,7 +82,7 @@ local attr_schema = {
                    },
                 }
             },
-            default = {address = "127.0.0.1:4317", request_timeout = 3}
+            default = {address = "127.0.0.1:4318", request_timeout = 3}
         },
         batch_span_processor = {
             type = "object",
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -434,7 +434,7 @@ plugin_attr:
     resource:
       service.name: APISIX
     collector:
-      address: 127.0.0.1:4317
+      address: 127.0.0.1:4318
       request_timeout: 3
       request_headers:
         Authorization: token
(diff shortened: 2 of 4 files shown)


tokers:
@soulbird Shall we also update the corresponding test cases?

soulbird:
> @soulbird Shall we also update the corresponding test cases?
Looks like no corresponding test cases need to be updated.

tzssangglass:
ref: https://github.com/open-telemetry/opentelemetry-specification/pull/1970

tzssangglass (approved):

moonming:
> 

Why？ No test cases for opentelemetry？

moonming:
> > @soulbird Shall we also update the corresponding test cases?
> > Looks like no corresponding test cases need to be updated.

Why？ No test cases for opentelemetry？

moonming:
the title of this PR should add `opentelemetry`

soulbird:
> > > @soulbird Shall we also update the corresponding test cases?
> > > Looks like no corresponding test cases need to be updated.
> 
> Why？ No test cases for opentelemetry？

In the test cases of opentelemetry, the data is reported to the black hole, and there is no port-related test. Look: https://github.com/apache/apisix/blob/master/t/plugin/opentelemetry.t#L41

soulbird:
> > > > > @soulbird Shall we also update the corresponding test cases?
> > > > > Looks like no corresponding test cases need to be updated.
> > > 
> > > 
> > > Why？ No test cases for opentelemetry？
> > 
> > 
> > In the test cases of opentelemetry, the data is reported to the black hole, and there is no port-related test. Look: https://github.com/apache/apisix/blob/master/t/plugin/opentelemetry.t#L41
> 
> Is opentelemetry really started? If yes, there should be has port

No services related to opentelemetry were found to be started.

moonming:
This port change problem should be covered with test cases

soulbird:
> This port change problem should be covered with test cases

I think it is not necessary to add test coverage related to the default port, and the success or failure of the report will not affect the operation of APISIX. You can see that the `do_request` function does not return any relevant error information, it just output a log. Look: https://github.com/yangxikun/opentelemetry-lua/blob/main/lib/opentelemetry/trace/exporter/http_client.lua#L30
cc @spacewander

soulbird:
> This port change problem should be covered with test cases

I think it is not necessary to add test coverage related to the default port, and the success or failure of the report will not affect the operation of APISIX. You can see that the `do_request` function does not return any relevant error information, it just output a log. Look: https://github.com/yangxikun/opentelemetry-lua/blob/main/lib/opentelemetry/trace/exporter/http_client.lua#L30
cc @spacewander

moonming:
> 

So should this function return an error message?

spacewander:
> > This port change problem should be covered with test cases
> 
> I think it is not necessary to add test coverage related to the default port, and the success or failure of the report will not affect the operation of APISIX. You can see that the `do_request` function does not return any relevant error information, it just output a log. Look: https://github.com/yangxikun/opentelemetry-lua/blob/main/lib/opentelemetry/trace/exporter/http_client.lua#L30 cc @spacewander

In the test, the `do_request` is replaced to a stub: https://github.com/apache/apisix/blob/0d1280cde7fcd463f2e86941935e053561bcf331/t/plugin/opentelemetry2.t#L43.

So changing the port won't affect the behavior of the test, and the correctness of trace reporting is covered by lua-opentelemetry itself.

On the other hand, as the default network port for OTLP/HTTP is changed from 4317 to 4318 in https://github.com/open-telemetry/opentelemetry-specification/pull/1970, it is necessary to change APISIX to adapt to the new break change.

CC the author of lua-opentelemetry and this plugin: @yangxikun

spacewander (approved):

yangxikun:
Maybe its my careless to use 4317 port as default value. Change 4317 to 4318 is fine. Users usually configure the plugin option `collector.address` explicitly instead of using the default values.
I think test case is not need for this case.

PR #7036, opened 2022-05-12
feat(xrpc): register variable rpc_time & redis_cmd_line

tzssangglass:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/stream/xrpc/runner.lua
+++ b/apisix/stream/xrpc/runner.lua
@@ -29,15 +29,15 @@ local tostring = tostring
 
 
 core.ctx.register_var("rpc_time", function(ctx)
-    local session = ctx.xrpc_session
-    local curr_ctx_id = session._currr_ctx_id
-    local curr_ctx = session._ctxs[curr_ctx_id]
+    local curr_ctx = ctx.xrpc_session and ctx.xrpc_session._curr_ctx
 
     if not curr_ctx then
         core.log.warn("can't find current context by id: ", curr_ctx_id)
         return nil
     end
-    return curr_ctx._rpc_end_time * 1000 - curr_ctx._rpc_start_time * 1000
+    --use second as the unit, like the request_time
+    local time = (curr_ctx._rpc_end_time * 1000 - curr_ctx._rpc_start_time * 1000) / 1000
+    return time
 end)
 
 local logger_expr_cache = core.lrucache.new({
@@ -129,8 +129,7 @@ end
 
 local function finish_req(protocol, session, ctx)
     ctx._rpc_end_time = ngx_now()
-    session._currr_ctx_id = ctx._id
-
+    session._curr_ctx = ctx
     local loggers = session.route.protocol.logger
     if loggers and #loggers > 0 then
         for _, logger in ipairs(loggers) do
--- a/apisix/stream/xrpc/runner.lua
+++ b/apisix/stream/xrpc/runner.lua
@@ -27,6 +27,19 @@ local pcall = pcall
 local ipairs = ipairs
 local tostring = tostring
 
+
+core.ctx.register_var("rpc_time", function(ctx)
+    local session = ctx.xrpc_session
+    local curr_ctx_id = session._currr_ctx_id
+    local curr_ctx = session._ctxs[curr_ctx_id]
+
+    if not curr_ctx then
+        core.log.warn("can't find current context by id: ", curr_ctx_id)
+        return nil
+    end
+    return curr_ctx._rpc_end_time * 1000 - curr_ctx._rpc_start_time * 1000
+end)
+
 local logger_expr_cache = core.lrucache.new({
     ttl = 300, count = 1024
 })
@@ -95,7 +108,7 @@ local function filter_logger(ctx, logger)
         core.log.error("failed to validate the 'filter' expression: ", err)
         return false
     end
-    return filter_expr:eval(ctx)
+    return filter_expr:eval(ctx.var)
 end
 
 
@@ -116,6 +129,7 @@ end
 
 local function finish_req(protocol, session, ctx)
     ctx._rpc_end_time = ngx_now()
+    session._currr_ctx_id = ctx._id
 
     local loggers = session.route.protocol.logger
     if loggers and #loggers > 0 then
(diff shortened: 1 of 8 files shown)


docs/en/latest/apisix-variable.md
```diff
@@ -39,5 +39,7 @@ List in alphabetical order:
 | route_name       | core    | name of `route`        |   |
 | service_id       | core    | id of `service`        |   |
 | service_name     | core    | name of `service`      |   |
+| rpc_time         | xRPC    | time spent at the rpc request level |   |
+| redis_cmd_line   | redis   | the type of Redis command |   |
```
spacewander, line 43:
```suggestion
| redis_cmd_line   | Redis   | the content of Redis command |   |
```

docs/en/latest/xrpc.md
```diff
@@ -131,6 +131,52 @@ One specifies the `superior_id`, whose corresponding value is the ID of another
...
+                }
+            }
+        }
+    }
+}
+```
+
+This configuration means that when the `rpc_time` is greater than 10 ms, xPRC reports the request log to the log server via the `syslog` plugin. `conf` is the configuration of the logging server required by the `syslog` plugin.
+
+Unlike standard TCP proxies, which only execute a logger when the connection is broken, xRPC's executed logger at the end of each 'request'.
```
spacewander, line 169:
```suggestion
Unlike standard TCP proxies, which only execute a logger when the connection is closed, xRPC's executed logger at the end of each 'request'.
```

docs/en/latest/xrpc.md
```diff
@@ -131,6 +131,52 @@ One specifies the `superior_id`, whose corresponding value is the ID of another
...
+
+The protocol itself defines the granularity of the specific request, and the xRPC extension code implements the request's granularity.
+
+For example, in the Redis protocol, the execution of a command is considered a request.
+
+### Exposes variables
```
spacewander, line 175: We can omit this section as it is written in docs/en/latest/apisix-variable.md

apisix/stream/xrpc/runner.lua
```diff
@@ -27,6 +27,19 @@ local pcall = pcall
...
+core.ctx.register_var("rpc_time", function(ctx)
+    local session = ctx.xrpc_session
+    local curr_ctx_id = session._currr_ctx_id
+    local curr_ctx = session._ctxs[curr_ctx_id]
+
+    if not curr_ctx then
+        core.log.warn("can't find current context by id: ", curr_ctx_id)
+        return nil
+    end
+    return curr_ctx._rpc_end_time * 1000 - curr_ctx._rpc_start_time * 1000
```
spacewander, line 40: Let's use second as the unit, like the `request_time`

apisix/stream/xrpc/runner.lua
```diff
@@ -27,6 +27,19 @@ local pcall = pcall
 local ipairs = ipairs
 local tostring = tostring
 
+
+core.ctx.register_var("rpc_time", function(ctx)
+    local session = ctx.xrpc_session
+    local curr_ctx_id = session._currr_ctx_id
+    local curr_ctx = session._ctxs[curr_ctx_id]
```
spacewander, line 34: I think we can implement a special var meta for the xRPC ctx?
A new version of https://github.com/apache/apisix/blob/4690feb421779f5b79e8dd990dc00f4d3f1052d0/apisix/core/ctx.lua#L216.

Therefore we can avoid looking up current xRPC ctx in every vars.

BTW, the `release_vars` is not called for xRPC ctx.

spacewander: Or we can pass ctx in https://github.com/apache/apisix/blob/4690feb421779f5b79e8dd990dc00f4d3f1052d0/apisix/core/ctx.lua#L343 as var._ctx

docs/en/latest/apisix-variable.md
tzssangglass: update

docs/en/latest/xrpc.md
tzssangglass: update

docs/en/latest/xrpc.md
tzssangglass: remove this

apisix/stream/xrpc/runner.lua
tzssangglass: update

apisix/stream/xrpc/runner.lua
tzssangglass: due to:https://github.com/apache/apisix/blob/4690feb421779f5b79e8dd990dc00f4d3f1052d0/apisix/core/ctx.lua#L345

in `core.ctx.register_var`, the `ctx.var._ctx` would point to `ngx.ctx.api_ctx`, so I choose  to pass th xRPC ctx in `session._curr_ctx`

apisix/stream/xrpc/runner.lua
```diff
@@ -27,6 +27,19 @@ local pcall = pcall
...
+
+core.ctx.register_var("rpc_time", function(ctx)
+    local curr_ctx = ctx.xrpc_session and ctx.xrpc_session._curr_ctx
+
+    if not curr_ctx then
+        core.log.warn("can't find current context by id: ", curr_ctx_id)
+        return nil
+    end
+    --use second as the unit, like the request_time
+    local time = (curr_ctx._rpc_end_time * 1000 - curr_ctx._rpc_start_time * 1000) / 1000
```
spacewander, line 39: Why do we multiply 1000 first and divide by 1000 later?

t/xrpc/redis.t
```diff
@@ -565,3 +577,126 @@ passed
 --- response_body
 ok
 --- stream_conf_enable
+
+
+
+=== TEST 12: set custom log format
```
spacewander, line 583: Maybe we can split the redis variable into another PR and the test can be put into `redis2.t`. Otherwise, there will be merge conflict.

tzssangglass: split this PR, so close.

apisix/stream/xrpc/runner.lua
spacewander: @tzssangglass 
The `register_var` doesn't rewrite the `ctx.var`. `ctx.var._ctx` will point to the ctx passed via `set_vars_meta(ctx)`.

PR #7049, opened 2022-05-14
feat(ops): allow access admin loopback IP address

kwanhur:
### Description
Allow [IPv4 loopback address](https://datatracker.ietf.org/doc/html/rfc5735#section-3) 127.0.0.0/8 and [IPv6 loopback address](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.3) to access admin.

Now, if not the address `127.0.0.0/24`, it'll get a warning tip like 

```shell
WARNING: using fixed Admin API token has security risk.
Please modify "admin_key" in conf/config.yaml .
```

* support single IP address and CIDR format
* support both IPv4 and IPv6
* remove hardcode 127.0.0.0/24 when checking

### Checklist


spacewander:
Would you please add test to cover it?

kwanhur:
Test cases cover IPv4 IPv6 CIDR and IPv6 loopback single address.

I'm not sure if need to cover non-loopback address cases.

t/cli/test_admin.sh
```diff
@@ -112,6 +112,48 @@ if [ $count -eq 0 ]; then
     exit 1
 fi
 
+echo "
```
spacewander, line 115: The newer added tests don't check the behavior of `checked_admin_key = true`

spacewander:
Let's merge master to make CI pass.

kwanhur:
Done.

t/cli/test_admin.sh
kwanhur: Thx. Fixed.

spacewander: Err... Even reverting the feature, the newer added tests can still pass, as they don't check the behavior of `checked_admin_key = true`.

kwanhur: Ohh, I got what you mean, now all the cases are `checked_admin_key = true`, should make it `checked_admin_key = false`, add more test cases later.

t/cli/test_admin.sh
```diff
@@ -112,6 +113,66 @@ if [ $count -eq 0 ]; then
...
+apisix:
+    enable_admin: true
+    allow_admin:
+        - 127.0.0.0/8
+" > conf/config.yaml
+
+make init
+
+count=`grep -c "allow 127.0.0.0/8" conf/nginx.conf`
+if [ $count -eq 0 ]; then
```
tzssangglass, line 126: Not sure I understand what the purpose of using count here is? Worried about other configurations also 
`listen 127.0.0.0/8` ?

kwanhur: `count` used to find the target address render or not. It'll place with `allow ip-address or cidr;`.

https://github.com/apache/apisix/blob/6ebe02797fb564c84b30df1865adef7f473880bf/apisix/cli/ngx_tpl.lua#L515-L528

t/cli/test_admin.sh
kwanhur: Added case to cover it. Plz take a view.

kwanhur: Case as below I think it's ok.

The detail here, `yaml_conf.apisix.enable_admin` to `true`, `yaml_conf.apisix.allow_admin` not empty, so it through into foreach element to check `loopback` or not, but value `all` still stays `checked_admin_key` to `false`. 

So this case satisfies `if yaml_conf.apisix.enable_admin and not checked_admin_key then` and walks inside.

Am I missing other? Please point it out, thanks very much.

```shell
echo "
apisix:
    enable_admin: true
    allow_admin:
        - all
" > conf/config.yaml
```

spacewander: The tests can be passed even without the new feature.

kwanhur: Added case value `10.0.0.1`.

The branch `if yaml_conf.apisix.enable_admin and not checked_admin_key then` seems be out of range changed feature from `if allow_ip == "127.0.0.0/24" then` to `if _ip and _ip:is_loopback() then`.

I think the new test case should cover `if _ip and _ip:is_loopback() then`, like the value `all` `10.0.0.1` `127.0.0.9` `127.0.0.1/8` `::1` `::1/128`, these cases are `non-IP address` `non-loopback` `IPv4 loopback address` `IPv4 loopback CIDR` `IPv6 loopback address` `IPv6 loopback CIDR`.

Maybe I'm confused now.

t/cli/test_admin.sh
```diff
@@ -112,6 +113,67 @@ if [ $count -eq 0 ]; then
...
+    echo "failed: not found 'allow 127.0.0.0/8;' in conf/nginx.conf"
+    exit 1
+fi
+
+echo "
+apisix:
+    enable_admin: true
+    allow_admin:
+        - ::1
```
spacewander, line 135: Could you rewrite or remove test cases which are not relative to the new feature?

kwanhur: To make some confirmation, the not relative is that `allow_admin` should be limited on which option?

1. Option A `IPv4 address` and `IPv4 CIDR`, the others are not relative.
2. Option B `IPv6 address` and `IPv6 CIDR` are not relative, the others are ok.
3. Option C `all` is not relative.

[allow](https://nginx.org/en/docs/http/ngx_http_access_module.html#allow) (ngx_http_access_module)

```
Syntax: | allow address | CIDR | unix: | all;
```

PR #7080, opened 2022-05-19
chore(ci): apt update before install

tzssangglass:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/.github/workflows/tars-ci.yml
+++ b/.github/workflows/tars-ci.yml
@@ -48,8 +48,7 @@ jobs:
 
       - name: Linux Install
         run: |
-          sudo ./ci/${{ matrix.os_name }}_runner.sh linux_get_dependencies
-          sudo cpanm --notest Test::Nginx >build.log 2>&1 || (cat build.log && exit 1)
+          sudo ./ci/${{ matrix.os_name }}_runner.sh before_install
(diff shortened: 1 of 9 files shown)


.github/workflows/tars-ci.yml
```diff
@@ -48,6 +48,7 @@ jobs:
 
       - name: Linux Install
         run: |
+          sudo apt-get update
           sudo apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev
```
spacewander, line 52: Would be better to put repeated commands in https://github.com/apache/apisix/blob/master/ci/common.sh?

tzssangglass: update

ci/linux_openresty_common_runner.sh
```diff
@@ -101,6 +101,10 @@ after_success() {
     echo "done"
 }
 
+linux_get_dependencies() {
+    install_dependencies
```
spacewander, line 105: We can put this in each runner.sh's `do_install`? So there is no need to put `sudo ./ci/${{ matrix.os_name }}_runner.sh linux_get_dependencies` in each workflow.

spacewander (approved):

bzp2010 (approved):

soulbird (approved):

PR #7150, opened 2022-05-27
chore: disable etcd dir init operation in DP mode

### LINKED ISSUE #6735: feat: As a user, I want to start Apache APISIX with a etcd user (who only have read permission of etcd), so that we can improve the security and avoid changing the etcd in DP side
### Description

Hello,  I config the Apache APISIX with a user (who only has read permission of `/apisix/`), and when I start Apache APISIX.
I meet this problem: 
```
got malformed key-put message: "{"error":"etcdserver: permission denied","code":7,"message":"etcdserver: permission denied"}" from etcd "http://127.0.0.1:2379/v3/kv/put
```

The code is as follow:
https://github.com/apache/apisix/blob/55eb58ab85ac7e108de36271fd9cf29fe2fa3324/apisix/cli/etcd.lua#L310-L365

I think we can avoid adding the `init_dir` if the directory exists.

leslie-tsang:
### Description

Fixes #6735

### Checklist

### DIFF
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -241,6 +241,11 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    if not yaml_conf.apisix.enable_admin then
+        print("skip etcd dir init operation as DP enabled")
+        return
+    end
+
     local etcd_ok = false
     for index, host in ipairs(etcd_healthy_hosts) do
         local is_success = true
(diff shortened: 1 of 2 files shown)


apisix/cli/etcd.lua
```diff
@@ -241,6 +241,11 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    if not yaml_conf.apisix.enable_admin then
```
nic-chen, line 244: it's a break change, should use a new field to do this

nic-chen (changes requested):

apisix/cli/etcd.lua
leslie-tsang: Makes sense, let's keep it as a demo for those who want to use `read-only` DP. :)

PR #7221, opened 2022-06-09
feat(ssl): support get upstream cert from ssl object

soulbird:
### Description

Support get upstream cert from ssl object:
1、Add `ssl.type` field to indicate the certificate type in the ssl object.
2、Only when `ssl.type=="client"`, the SSL object can be referenced.

### Checklist


apisix/admin/ssl.lua
```diff
@@ -46,7 +46,7 @@ local function check_conf(id, conf, need_id)
     conf.id = id
 
     core.log.info("schema: ", core.json.delay_encode(core.schema.ssl))
-    core.log.info("conf  : ", core.json.delay_encode(conf))
+    core.log.info("conf : ", core.json.delay_encode(conf))
```
spacewander, line 49:
```suggestion
    core.log.info("conf: ", core.json.delay_encode(conf))
```

apisix/init.lua
```diff
@@ -512,6 +512,22 @@ function _M.http_access_phase()
         return pubsub_kafka.access(api_ctx)
     end
 
+    if api_ctx.matched_upstream and api_ctx.matched_upstream.tls_id then
```
spacewander, line 515: Could we move this just after the upstream is found:
https://github.com/apache/apisix/blob/975038ea70dc47e0613c42b3eabb8dcc730216a4/apisix/init.lua#L497

apisix/schema_def.lua
```diff
@@ -722,6 +731,13 @@ _M.ssl = {
     type = "object",
     properties = {
         id = id_schema,
+        type = {
+            description = "ssl certificate type, " ..
+                            "0 to server certificate, " ..
+                            "1 to client certificate for upstream",
+            type = "integer",
+            enum = {0, 1}
```
spacewander, line 739: Please use enum string like `server` & `client`, thanks!

spacewander: Better to provide a default value, so we don't need to handle an unset situation.

docs/en/latest/admin-api.md
```diff
@@ -541,8 +541,9 @@ In addition to the equalization algorithm selections, Upstream also supports pas
...
+| tls.client_cert_id          | optional, can't be used with `tls.client_cert` and `tls.client_key`       | Set the referenced [SSL](#ssl) id.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |                                                                                                                                            |
```
membphis, line 546: @tokers pls confirm this first

apisix/init.lua
```diff
@@ -495,6 +495,24 @@ function _M.http_access_phase()
                                    or route_val.upstream
     end
 
+    if api_ctx.matched_upstream and api_ctx.matched_upstream.tls and
+        api_ctx.matched_upstream.tls.client_cert_id then
+
+        local cert_id = api_ctx.matched_upstream.tls.client_cert_id
+        local upstream_ssl = router.router_ssl.get_by_id(cert_id)
+        if not upstream_ssl or upstream_ssl.type ~= "client" then
```
tokers, line 503: here dd a log will be better.

docs/en/latest/admin-api.md
tokers: No problem but currently it looks strange as we use an` SSL` object while the name is `client_cert_id`.

docs/en/latest/admin-api.md
```diff
@@ -570,6 +571,8 @@ You can set the `scheme` to `tls`, which means "TLS over TCP".
 
 To use mTLS to communicate with Upstream, you can use the `tls.client_cert/key` in the same format as SSL's `cert` and `key` fields.
 
+Or you can reference SSL object by `tls.client_cert_id` to set SSL cert and key. The SSL object can be referenced only if the `type` field is `client`. Only `cert` and `key` will be used in the SSL object.
```
tokers, line 574: **Otherwise the request will be rejected by APISIX**.

I think we should tell users the result.

apisix/upstream.lua
```diff
@@ -415,6 +425,29 @@ local function check_upstream_conf(in_dp, conf)
             return false, "invalid configuration: " .. err
         end
 
+        local ssl_id = conf.tls and conf.tls.client_cert_id or nil
```
spacewander, line 428:
```suggestion
        local ssl_id = conf.tls and conf.tls.client_cert_id
```
is enough

t/node/upstream-mtls.t
```diff
@@ -545,3 +545,68 @@ GET /t
 GET /hello_chunked
 --- response_body
 hello world
+
+
+
+=== TEST 13: get cert by tls.client_cert_id
```
spacewander, line 551: Let's add a case that there is no SSL object according to the client_cert_id

spacewander (dismissed):

docs/en/latest/admin-api.md
tokers: @soulbird What's the result?

docs/en/latest/admin-api.md
```diff
@@ -792,7 +792,7 @@ Currently, the response is returned from etcd.
...
 | update_time  | False    | Auxiliary                | Epoch timestamp (in seconds) of the updated time. If missing, this field will be populated automatically.         | 1602883670                                       |
-| type         | False    | Certificate type         | Identifies the type of certificate, default  `server`.                                                                             | `client` Indicates that the certificate is a client certificate, which is used when APISIX accesses the upstream; `server` Indicates that the certificate is a server-side certificate, which is used by APISIX when verifying client requests.     |
+| type         | False    | Certificate position         | Identifies the type of certificate, default  `server`.                                                                             | `client` Indicates that the certificate is a client certificate, which is used when APISIX accesses the upstream; `server` Indicates that the certificate is a server-side certificate, which is used by APISIX when verifying client requests.     |
```
spacewander, line 795: What does `Certificate position` mean? Is there any source for this concept?

soulbird: In fact, I didn't find any official concept to describe client short certificate and server certificate, do you have any suggestion?

spacewander: If this field is controversial, what about using `Auxiliary`?

soulbird: That's great!

spacewander (approved):

PR #7240, opened 2022-06-12
feat: allow APISIX_PROFILE to be an empty string

### LINKED ISSUE #7216: When APISIX_PROFILE is used, no empty judgment is performed, resulting in an unsightly display format of the configuration file name
### Description

1. When I am using APISIX_PROFILE, if a value is assigned, it will be displayed as follows
```
    a. config-dp.yaml
    b. config-cp.yaml
    c. ...[others]
```

2. However, when I don't do the assignment, I expect to be `config.yaml`, but I actually get `config-.yaml`.
3. It works fine, but the file names look ugly. As someone with code cleanliness, it's hard to accept it.

sunhao-java:
### Description
When APISIX_PROFILE is set to empty string, the configuration file name is set to config-.yaml

Fixes #7216


tzssangglass:
we should update docs:https://github.com/apache/apisix/blob/master/docs/en/latest/profile.md, and add test case, like: https://github.com/apache/apisix/blob/master/t/core/profile.t

sunhao-java:
> we should update docs:https://github.com/apache/apisix/blob/master/docs/en/latest/profile.md, and add test case, like: https://github.com/apache/apisix/blob/master/t/core/profile.t

Is there some documentation that can help me to write unit test cases

tzssangglass:
> Is there some documentation that can help me to write unit test cases

https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md

t/core/profile.t
```diff
@@ -32,3 +32,19 @@ __DATA__
...
 GET /t
 --- error_code: 404
+
+=== TEST 2: set env "APISIX_PROFILE" to Empty String
+--- config
+    location /t {
+        content_by_lua_block {
+            local profile = require("apisix.core.profile")
+            profile.apisix_home = "./test/"
+            profile.profile = ""
```
spacewander, line 42: We can check if `profile.profile` is equal to `""` here?

sunhao-java: `profile.profile = ""` means we set env "APISIX_PROFILE" to empty string.
This is just to simulate a scenario where the environment variable is empty string.

sunhao-java: > We can check if `profile.profile` is equal to `""` here?

profile.profile = "" means we set env "APISIX_PROFILE" to empty string.
This is just to simulate a scenario where the environment variable is empty string.

spacewander (approved):

tzssangglass (approved):

PR #7252, opened 2022-06-14
fix(skywalking-logger, error-log-logger): skywalking service_instance_name $hostname is supported

### LINKED ISSUE #7251: request help: skywalking-logger and error-log-logger plugin service_instance_name set $hostname invalid
### Current Behavior

skywalking-logger and error-log-logger plugins set service_instance_name to `$hostname` and report to skywalking

instance is represented as `"$hostname"`string

image

### Expected Behavior

Expect service instances to be a string of $hostname instead of "$hostname" after reporting to Skywalking

image

### Error Logs

no error logs

### Steps to Reproduce

We can refer to the skywalking plugin processing method

https://github.com/apache/apisix/blob/45816272a80d6d2045259d32acdabbec4039ed1e/apisix/plugins/skywalking.lua#L130-L132

### Environment

- APISIX version (run `apisix version`): 2.13.1
- Operating system (run `uname -a`): Linux apisix-844457679-8242t 5.10.102.1-microsoft-standard-WSL2+ #1 SMP Mon Jun 13 04:56:19 UTC 2022 x86_64 Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): openresty/1.19.9.1

dickens7:
### Description

Fixes #7251

### Checklist

### DIFF
--- a/apisix/plugins/error-log-logger.lua
+++ b/apisix/plugins/error-log-logger.lua
@@ -180,15 +180,15 @@ local function send_to_skywalking(log_message)
     httpc:set_timeout(config.timeout * 1000)
 
     local entries = {}
-    local serviceInstanceName = config.skywalking.service_instance_name
-    if serviceInstanceName == "$hostname" then
-        serviceInstanceName = core.utils.gethostname()
+    local service_instance_name = config.skywalking.service_instance_name
+    if service_instance_name == "$hostname" then
+        service_instance_name = core.utils.gethostname()
     end
 
     for i = 1, #log_message, 2 do
         local content = {
             service = config.skywalking.service_name,
-            serviceInstance = serviceInstanceName,
+            serviceInstance = service_instance_name,
             endpoint = "",
             body = {
                 text = {
--- a/apisix/plugins/error-log-logger.lua
+++ b/apisix/plugins/error-log-logger.lua
@@ -180,10 +180,15 @@ local function send_to_skywalking(log_message)
     httpc:set_timeout(config.timeout * 1000)
 
     local entries = {}
+    local serviceInstanceName = config.skywalking.service_instance_name
+    if serviceInstanceName == "$hostname" then
+        serviceInstanceName = core.utils.gethostname()
+    end
+
     for i = 1, #log_message, 2 do
         local content = {
             service = config.skywalking.service_name,
-            serviceInstance = config.skywalking.service_instance_name,
+            serviceInstance = serviceInstanceName,
             endpoint = "",
             body = {
                 text = {
--- a/apisix/plugins/skywalking-logger.lua
+++ b/apisix/plugins/skywalking-logger.lua
@@ -143,6 +143,11 @@ function _M.log(conf, ctx)
         end
     end
 
+    local serviceInstanceName = conf.service_instance_name
+    if serviceInstanceName == "$hostname" then
+        serviceInstanceName = core.utils.gethostname()
+    end
+
     local entry = {
         traceContext = trace_context,
         body = {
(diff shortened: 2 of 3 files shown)


apisix/plugins/error-log-logger.lua
```diff
@@ -180,10 +180,15 @@ local function send_to_skywalking(log_message)
     httpc:set_timeout(config.timeout * 1000)
 
     local entries = {}
+    local serviceInstanceName = config.skywalking.service_instance_name
```
spacewander, line 183: We should call it in the `init` method, instead of in the per-report way

spacewander (commented):
Let's submit the PR to master, and backport it after it's merged

apisix/plugins/error-log-logger.lua
spacewander: I change my mind. It is a route-level configuration and may be changed on the fly.

tokers:
Can we add some test cases to cover the changes?

dickens7:
> Can we add some test cases to cover the changes?

I don't know how this test case can be written to cover this scenario.

Shall we query the data received by skywaling?

soulbird:
> > Can we add some test cases to cover the changes?
> 
> I don't know how this test case can be written to cover this scenario.
> 
> Shall we query the data received by skywaling?

Just for reference:  https://github.com/apache/apisix/blob/master/t/plugin/skywalking-logger.t#L124-L196.
However, in the instance of executing the github action, the hostname may not be determined

tzssangglass:
we can verify `hostname` like: https://github.com/apache/apisix/blob/master/t/plugin/server-info.t#L108

dickens7:
> we can verify `hostname` like:
> 
> https://github.com/apache/apisix/blob/afce02cea506eb4c70f2dd9d5230d18c7ab8b51e/t/plugin/server-info.t#L108

This won't work, because `$hostname` can also be matched but we expect the value to be something other than `$hostname`

tzssangglass:
how about `^[a-zA-Z\-0-9\.]+`

dickens7:
> how about `^[a-zA-Z\-0-9\.]+`

`service_instance_name` is user-defined and can be set to any value such as `#xxx`.
This is just for testing, it should work.

But I think my current use case should be ok, we just need to judge that service_instance_name is not equal to the string `$hostname` and not equal to the empty string.

```
--- no_error_log eval
qr/\\\"serviceInstance\\\":\\\"\$hostname\\\"/
qr/\\\"serviceInstance\\\":\\\"\\\"/
```

spacewander (approved):

tzssangglass:
I reruned the failed CI and if CI passed, we can merge it.

tzssangglass (approved):

apisix/plugins/error-log-logger.lua
starsz, line 183: IMO, we should use this code style, like "service_instance_name`

dickens7: Yes, we should use snake case naming

dickens7:
please rerun ci

spacewander (approved):

starsz (approved):

apisix/plugins/error-log-logger.lua
```diff
@@ -180,10 +180,15 @@ local function send_to_skywalking(log_message)
     httpc:set_timeout(config.timeout * 1000)
 
     local entries = {}
+    local service_instance_name = config.skywalking.service_instance_name
+    if service_instance_name == "$hostname" then
+        service_instance_name = core.utils.gethostname()
+    end
```
leslie-tsang, line 186: Hello there, we need a test case to ensure it work as expected. :)

leslie-tsang (commented):
LGTM. 👍

PR #7266, opened 2022-06-16
feat: export some importent params for kafka-client

mikawudi:
export 4 importent params about kafka-producter for kafka-logger plugins
if no problem on this commit, i will edit docs later

### DIFF
--- a/apisix/plugins/kafka-logger.lua
+++ b/apisix/plugins/kafka-logger.lua
@@ -84,10 +84,10 @@ local schema = {
         -- see https://github.com/doujiang24/lua-resty-kafka#new-1
         cluster_name = {type = "integer", minimum = 1, default = 1},
         -- config for lua-resty-kafka, default value is same as lua-resty-kafka
-        kafka_batch_num = {type = "integer", minimum = 1, default = 200},
-        kafka_batch_size = {type = "integer", minimum = 0, default = 1048576},
-        kafka_max_buffering = {type = "integer", minimum = 1, default = 50000},
-        kafka_time_linger = {type = "integer", minimum = 1, default = 1}
+        producer_batch_num = {type = "integer", minimum = 1, default = 200},
+        producer_batch_size = {type = "integer", minimum = 0, default = 1048576},
+        producer_max_buffering = {type = "integer", minimum = 1, default = 50000},
+        producer_time_linger = {type = "integer", minimum = 1, default = 1}
     },
     required = {"broker_list", "kafka_topic"}
 }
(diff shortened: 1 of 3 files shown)


apisix/plugins/kafka-logger.lua
```diff
@@ -83,6 +83,11 @@ local schema = {
         -- in lua-resty-kafka, cluster_name is defined as number
         -- see https://github.com/doujiang24/lua-resty-kafka#new-1
         cluster_name = {type = "integer", minimum = 1, default = 1},
+        -- config for lua-resty-kafka, default value is same as lua-resty-kafka
+        kafka_batch_num = {type = "integer", minimum = 1, default = 200},
```
spacewander, line 87: Would be better to use `producer` prefix instead of `kafka`? Since this plugin is already a kafka logger.

apisix/plugins/kafka-logger.lua
```diff
@@ -83,6 +83,11 @@ local schema = {
...
+        kafka_batch_size = {type = "integer", minimum = 0, default = 1048576},
+        kafka_max_buffering = {type = "integer", minimum = 1, default = 50000},
+        kafka_time_linger = {type = "integer", minimum = 1, default = 1}
```
spacewander, line 90: Why don't we call it `flush_time`?

spacewander (commented):
Let's add some doc for the new fields

apisix/plugins/kafka-logger.lua
mikawudi: good idea,i will change it

apisix/plugins/kafka-logger.lua
mikawudi: because this params named “linger.ms” in official kafka client producer and other popular kafka client like sarama,
and i don‘t know why named “flush_time” in resty-kafka......i think call it like “linger” is friendly for user google it....

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -47,6 +47,10 @@ For more info on Batch-Processor in Apache APISIX please refer.
 | include_resp_body| boolean | optional    | false         | [false, true] | Whether to include the response body. The response body is included if and only if it is `true`. |
 | include_resp_body_expr  | array  | optional    |          |         | When `include_resp_body` is true, control the behavior based on the result of the [lua-resty-expr](https://github.com/api7/lua-resty-expr) expression. If present, only log the response body when the result is true. |
 | cluster_name     | integer | optional    | 1              | [0,...] | the name of the cluster. When there are two or more kafka clusters, you can specify different names. And this only works with async producer_type.|
+| producer_batch_num | integer | optional    | 200 | [1,...] | `batch_num` param in [lua-resty-kafka],merge message and batch send to server, unit is message count |
```
spacewander, line 50: Should be `[lua-resty-kafka](link to lua-resty-kafka)`?

spacewander (commented):

spacewander (approved):

tzssangglass (approved):

PR #7273, opened 2022-06-17
feat: allows users to specify plugin execution priority

### LINKED ISSUE #6580: feat: Redesign the architecture of the plugin system
## Current situation

At the moment, you add a plugin in the request flow by referencing with its name and setting the configuration. The plugin is effectively a **singleton**. 

This approach has some downsides:

* You can use a plugin only once in the request flow. Users cannot add the same plugin multiple times with different configurations.
* For maintainers, it means that they need to duplicate code and reference it with different names and priorities. For example, the [ext-plugin-pre-req](https://github.com/apache/apisix/blob/master/apisix/plugins/ext-plugin-pre-req.lua) and the [ext-plugin-post-req](https://github.com/apache/apisix/blob/master/apisix/plugins/ext-plugin-post-req.lua) do the same - run arbitrary Lua code.
* Users cannot change the priority of a plugin, it's hard-coded (see #6504)
* Plugin developers need to handle every possible configuration case (see #6197)

## Proposal

In Object-Oriented speak, make `Plugin` a class and introduce `PluginInstance` "objects" (or something similarly named):

* `Plugin` defines the code
* `PluginInstance` references the `Plugin` and associates it with a priority and a configuration

Decoupling the code from the configuration/priority would bring much more freedom for users.

### LINKED ISSUE #7248: help request: proxy-rewrite plugin impact on authz-keycloak plugin
### Description

in my scene，the route enable proxy-rewrite and authz-keycloak plugins both  
```
{
  "uri": "/elastic-compute1/*",
  "name": "test1",
  "methods": [
    "GET",
    "POST",
    "PUT",
    "DELETE",
    "PATCH",
    "HEAD",
    "OPTIONS",
    "CONNECT",
    "TRACE"
  ],
  "plugins": {
    "authz-keycloak": {
      "client_id": "xx",
      "client_secret": "yyyyy",
      "disable": false,
      "http_method_as_scope": true,
      "lazy_load_paths": true,
      "resource_registration_endpoint": "https://keycloak.paas-middleware:8443/auth/realms/xx/authz/protection/resource_set",
      "ssl_verify": false,
      "token_endpoint": "https://keycloak.paas-middleware:8443/auth/realms/xx/protocol/openid-connect/token"
    },
    "key-auth": {
      "disable": false
    },
    "proxy-rewrite": {
      "headers": {
        "X-App-Id": "$consumer_name"
      },
      "regex_uri": [
        "^/elastic-compute1/(.*)",
        "/elastic-compute/$1"
      ]
    }
  },
  "upstream_id": "1",
  "status": 1
}
```
i notice that the proxy-rewrite plugin priority is smaller than authz-keycloak, which means do the proxy-rewrite plugin process first.
the request is ```/elastic-compute1/v1/clusters```
so the ctx.var.upstream_uri match proxy-rewrite regex_uri ,became to ```/elastic-compute/v1/clusters```
BUT the authz-keycloak plugin use ctx.var.request_uri as resource_id send to keycloak get permissions
```
        -- Resolve URI to resource(s).
        permission, err = authz_keycloak_resolve_resource(conf, ctx.var.request_uri,
                                                          sa_access_token)
```
Is it designed this way?
or maybe we can pass request_uri replaced by upstream_uri when they are not same(usually means processed by proxy-rewrite before)

### Environment

- APISIX version (run `apisix version`):2.14.1
- Operating system (run `uname -a`):centos 3.10.0-1160.el7.x86_64
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): 1.19.9.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 3.5.4

tzssangglass:
### Description

Fixes #6580
Fixes #7248

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -68,6 +68,10 @@ local function sort_plugin(l, r)
     return l.priority > r.priority
 end
 
+local function custom_sort_plugin(l, r)
+    return l._meta.priority > r._meta.priority
+end
+
 
 local PLUGIN_TYPE_HTTP = 1
 local PLUGIN_TYPE_STREAM = 2
@@ -365,7 +369,7 @@ local function trace_plugins_info_for_debug(ctx, plugins)
 end
 
 
-function _M.filter(ctx, conf, plugins, route_conf)
+function _M.filter(ctx, conf, plugins, route_conf, phase)
     local user_plugin_conf = conf.value.plugins
     if user_plugin_conf == nil or
        core.table.nkeys(user_plugin_conf) == 0 then
@@ -375,6 +379,7 @@ function _M.filter(ctx, conf, plugins, route_conf)
         return plugins or core.tablepool.fetch("plugins", 0, 0)
     end
 
+    local custom_sort = false
     local route_plugin_conf = route_conf and route_conf.value.plugins
     plugins = plugins or core.tablepool.fetch("plugins", 32, 0)
     for _, plugin_obj in ipairs(local_plugins) do
@@ -389,6 +394,9 @@ function _M.filter(ctx, conf, plugins, route_conf)
                 end
             end
 
+            if plugin_conf._meta and plugin_conf._meta.priority then
+                custom_sort = true
+            end
             core.table.insert(plugins, plugin_obj)
             core.table.insert(plugins, plugin_conf)
 
@@ -397,6 +405,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
     end
 
     trace_plugins_info_for_debug(ctx, plugins)
+
+
+    if custom_sort then
+        local tmp_plugin_objs = core.tablepool.fetch("tmp_plugin_objs", 0, #plugins / 2)
+        local tmp_plugin_confs = core.tablepool.fetch("tmp_plugin_confs", #plugins / 2, 0)
+
+        for i = 1, #plugins, 2 do
+            local plugin_obj = plugins[i]
+            local plugin_conf = plugins[i + 1]
+
+            -- in the rewrite phase, the plugin executes in the following order:
+            -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
+            -- 2. merge plugins from consumer and route
+            -- 3.execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
+            -- in this case, we need to skip the plugins that was already executed(step 1)
+            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
+                plugin_conf._skip_rewrite_in_consumer = true
+            end
+
+            tmp_plugin_objs[tostring(plugin_conf)] = plugin_obj
+            core.table.insert(tmp_plugin_confs, plugin_conf)
+
+            if not plugin_conf._meta or not plugin_conf._meta.priority then
+                plugin_conf._meta = core.table.new(0, 1)
+                plugin_conf._meta.priority = plugin_obj.priority
+            end
+
+        end
+
+        sort_tab(tmp_plugin_confs, custom_sort_plugin)
+
+        core.table.clear(plugins)
+
+        for i = 1, #tmp_plugin_confs do
+            local plugin_conf = tmp_plugin_confs[i]
+            local plugin_obj = tmp_plugin_objs[tostring(plugin_conf)]
+            core.table.insert(plugins, plugin_obj)
+            core.table.insert(plugins, plugin_conf)
+        end
+
+        core.tablepool.release("tmp_plugin_objs", tmp_plugin_objs)
+        core.tablepool.release("tmp_plugin_confs", tmp_plugin_confs)
+    end
 
     return plugins
 end
@@ -754,6 +805,11 @@ function _M.run_plugin(phase, plugins, api_ctx)
                 phase = "rewrite"
             end
             local phase_func = plugins[i][phase]
+
+            if phase == "rewrite" and plugins[i + 1]._skip_rewrite_in_consumer then
+                goto CONTINUE
+            end
+
             if phase_func then
                 plugin_run = true
                 local conf = plugins[i + 1]
@@ -781,6 +837,8 @@ function _M.run_plugin(phase, plugins, api_ctx)
                     end
                 end
             end
+
+            ::CONTINUE::
         end
         return api_ctx, plugin_run
     end
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -419,14 +419,14 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
             -- 2. merge plugins from consumer and route
             -- 3. execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
             -- in this case, we need to skip the plugins that was already executed(step 1)
-            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
+            if phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
                 plugin_conf._skip_rewrite_in_consumer = true
             end
 
             tmp_plugin_objs[plugin_conf] = plugin_obj
             core.table.insert(tmp_plugin_confs, plugin_conf)
 
-            if not plugin_conf._meta or type(plugin_conf._meta) ~= "table" then
+            if not plugin_conf._meta then
                 plugin_conf._meta = core.table.new(0, 1)
                 plugin_conf._meta.priority = plugin_obj.priority
             else
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -438,7 +438,7 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
 
         sort_tab(tmp_plugin_confs, custom_sort_plugin)
 
-        local index = 0
+        local index
         for i = 1, #tmp_plugin_confs do
             index = i * 2 - 1
             local plugin_conf = tmp_plugin_confs[i]
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -418,7 +418,7 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
             -- in the rewrite phase, the plugin executes in the following order:
             -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
             -- 2. merge plugins from consumer and route
-            -- 3.execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
+            -- 3. execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
             -- in this case, we need to skip the plugins that was already executed(step 1)
             if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
                 plugin_conf._skip_rewrite_in_consumer = true
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -406,7 +406,6 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
 
     trace_plugins_info_for_debug(ctx, plugins)
 
-
     if custom_sort then
         local tmp_plugin_objs = core.tablepool.fetch("tmp_plugin_objs", 0, #plugins / 2)
         local tmp_plugin_confs = core.tablepool.fetch("tmp_plugin_confs", #plugins / 2, 0)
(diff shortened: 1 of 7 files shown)


apisix/plugin.lua
```diff
@@ -398,6 +406,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+            -- in the rewrite phase, the plugin executes in the following order:
+            -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
+            -- 2. merge plugins from consumer and route
+            -- 3. execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
+            -- in this case, we need to skip the plugins that was already executed(step 1)
+            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
+                plugin_conf._skip_rewrite_in_consumer = true
+            end
+
+            tmp_plugin_objs[tostring(plugin_conf)] = plugin_obj
```
spacewander, line 427: We can directly use table as key

apisix/plugin.lua
```diff
@@ -398,6 +406,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+
+        end
+
+        sort_tab(tmp_plugin_confs, custom_sort_plugin)
+
+        core.table.clear(plugins)
+
+        for i = 1, #tmp_plugin_confs do
+            local plugin_conf = tmp_plugin_confs[i]
+            local plugin_obj = tmp_plugin_objs[tostring(plugin_conf)]
```
spacewander, line 443: Ditto

apisix/plugin.lua
```diff
@@ -398,6 +406,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+            if not plugin_conf._meta or not plugin_conf._meta.priority then
+                plugin_conf._meta = core.table.new(0, 1)
+                plugin_conf._meta.priority = plugin_obj.priority
+            end
+
+        end
+
+        sort_tab(tmp_plugin_confs, custom_sort_plugin)
+
+        core.table.clear(plugins)
```
spacewander, line 439: We can skip the `clear` and assign to the original plugins by index

t/plugin/custom_sort_plugins.t
```diff
@@ -0,0 +1,611 @@
...
+=== TEST 5: verify order
+--- request
+GET /hello
+--- response_body
+serverless-post-function
+serverless-pre-function
+
+
+
+=== TEST 6: merge plugins form custom and route, execute the rewrite phase
```
spacewander, line 195:
```suggestion
=== TEST 6: merge plugins from consumer and route, execute the rewrite phase
```

apisix/plugin.lua
```diff
@@ -398,6 +406,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+            -- in this case, we need to skip the plugins that was already executed(step 1)
+            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
+                plugin_conf._skip_rewrite_in_consumer = true
+            end
+
+            tmp_plugin_objs[tostring(plugin_conf)] = plugin_obj
+            core.table.insert(tmp_plugin_confs, plugin_conf)
+
+            if not plugin_conf._meta or not plugin_conf._meta.priority then
+                plugin_conf._meta = core.table.new(0, 1)
```
spacewander, line 431: This will override the existed fields in _meta?

docs/en/latest/terminology/plugin.md
```diff
@@ -95,6 +95,43 @@ the configuration above means customizing the error response from the jwt-auth p
...
+| priority       | integer        | Custom plugin priority |
+
+### Custom Plugin Priority
+
+All plugins have a default priority, but it is possible to customize the plugin priority to change the plugin's execution order.
+
+```json
+ {
+    "serverless-post-function": {
+        "_meta": {
```
nfrankel, line 107: As I already mentioned, this is poor Developer Experience. It makes the use of the configuration more complex for no reason except that it's easier to implement.

docs/en/latest/terminology/plugin.md
```diff
@@ -95,6 +95,43 @@ the configuration above means customizing the error response from the jwt-auth p
...
+        },
+        "phase": "rewrite",
+        "functions": ["return function(conf, ctx)
+                    ngx.say(\"serverless-pre-function\");
+                    end"]
+    }
+}
+```
+
+The default priority of serverless-pre-function is 10000, and the default priority of serverless-post-function is -2000. By default, the serverless-pre-function plugin will be executed first, and serverless-post-function plugin will be executed next.
```
nfrankel, line 127: At this point, why do we keep two different functions? We can keep a single one as call it whenever we required it. Can we change the sample?

nfrankel: A good use-case would be the combination of `proxy-mirror` plugin and the `proxy-rewrite` one.

nfrankel (changes requested):

docs/en/latest/terminology/plugin.md
```diff
@@ -95,6 +95,43 @@ the configuration above means customizing the error response from the jwt-auth p
...
+
+The above configuration means setting the priority of the serverless-pre-function plugin to -2000 and the priority of the serverless-post-function plugin to 10000. The serverless-post-function plugin will be executed first, and serverless-pre-function plugin will be executed next.
+
+Note:
+
+- Custom plugin priority only affects the current object(route, service ...) of the plugin instance binding, not all instances of that plugin. For example, if the above plugin configuration belongs to Route A, the order of execution of the plugins serverless-post-function and serverless-post-function on Route B will not be affected and the default priority will be used.
+- Custom plugin priority does not apply to the rewrite phase of some plugins configured on the consumer. The rewrite phase of plugins configured on the route will executed first, and then the rewrite phase of plugins(exclude auth plugins) on the consumer will be executed.
```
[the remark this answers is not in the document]
tzssangglass: fixed

apisix/plugin.lua
tzssangglass: fixed

apisix/plugin.lua
tzssangglass: fixed

apisix/plugin.lua
tzssangglass: updated

t/plugin/custom_sort_plugins.t
tzssangglass: fixed

apisix/plugin.lua
tzssangglass: fixed

apisix/plugin.lua
```diff
@@ -398,6 +406,49 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+            local plugin_obj = plugins[i]
+            local plugin_conf = plugins[i + 1]
+
+            -- in the rewrite phase, the plugin executes in the following order:
+            -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
+            -- 2. merge plugins from consumer and route
+            -- 3. execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
+            -- in this case, we need to skip the plugins that was already executed(step 1)
+            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
+                plugin_conf._skip_rewrite_in_consumer = true
```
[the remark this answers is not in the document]
tzssangglass: No, the `_skip_rewrite_in_consumer` flag is present in conf and it will only find plugins from the route during the `rewrite_in_consumer` phase so that they can be skipped during the `rewrite_in_consumer` phase.

The test case(Test 6&7) has been modified to cover it.

docs/en/latest/terminology/plugin.md
tzssangglass: I also think `priority` should not be in `_meta`, it doesn't fit the definition of `_meta`, it only works for plugin bound objects. How about a custom priority directly as a base property of the plugin? ref: https://github.com/apache/apisix/issues/6580#issuecomment-1089871187

cc @spacewander @membphis

docs/en/latest/terminology/plugin.md
tzssangglass: > A good use-case would be the combination of `proxy-mirror` plugin and the `proxy-rewrite` one.

I tried it and the two plugins don't show well to work together by customizing the priority.

Maybe you're talking about changing the request path via the `proxy-rewrite` plugin, and then `proxy-mirror` using that new path.

But `proxy-rewrite` plugin update uri on the var `ctx.var.upstream_uri`, but `proxy-mirror` get the uri by the var `ctx.var.uri`, this is the difference in implementation.

docs/en/latest/terminology/plugin.md
spacewander: The plugin property will affect every plugin configuration, not just what we configure for.

apisix/plugin.lua
spacewander: Why add `type(plugin_conf._meta) ~= "table"`?

apisix/schema_def.lua
```diff
@@ -952,6 +952,11 @@ _M.plugin_injected_schema = {
                     { type = "object" },
                 }
             },
+            priority = {
+                description = "priority of plugins by customized order",
+                type = "integer",
+                default = 0,
```
spacewander, line 958: We should not use a default value here (it should depend on the plugin)

apisix/plugin.lua
```diff
@@ -398,6 +406,51 @@ function _M.filter(ctx, conf, plugins, route_conf)
...
+        for i = 1, #plugins, 2 do
+            local plugin_obj = plugins[i]
+            local plugin_conf = plugins[i + 1]
+
+            -- in the rewrite phase, the plugin executes in the following order:
+            -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
+            -- 2. merge plugins from consumer and route
+            -- 3. execute the rewrite phase of the plugins on consumer(phase: rewrite_in_consumer)
+            -- in this case, we need to skip the plugins that was already executed(step 1)
+            if phase and phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
```
spacewander, line 422:
```suggestion
            if phase == "rewrite_in_consumer" and not plugin_conf._from_consumer then
```

tzssangglass: fixed

apisix/schema_def.lua
tzssangglass: remove it

docs/en/latest/terminology/plugin.md
tzssangglass: After discussing with @spacewander , I changed my mind.

`_meta` is based on the concept of abstraction of all plugin configurations, which are available to all plugins. So to distinguish it, it should have a separate configuration field, i.e. `_meta`.

nfrankel: Sorry, but again, you speak from an _engineering_ perspective. I'm talking from a _user_ perspective. As a user, it's not fundamental whether the configuration parameter can be used across multiple plugins.

Here, I need to know **and** remember:

1. The parameter name, _i.e._, `priority`
2. The parent's name, _i.e._, `meta`
3. The prefix, _i.e._, `_` 

Only #1 is mandatory. You make #2 and #3 necessary and impair the ease of use for engineering reasons. While I understand there's a tension between the ease of implementation and ease of use, I don't see any good reason to force those additional requirements onto users at this point.

docs/en/latest/terminology/plugin.md
nfrankel: >I tried it and the two plugins don't show well to work together by customizing the priority.

That's actually the origin of my initial requirement.

>Maybe you're talking about changing the request path via the proxy-rewrite plugin, and then proxy-mirror using that new path.

Indeed!

Imagine that I receive a request for `v1/hello`. Now, my requirements are:

1. The prefix `v1` to be removed
2. The request to be forwarded to the upstream
3. The request to be duplicated to the mirror

>But proxy-rewrite plugin update uri on the var ctx.var.upstream_uri, but proxy-mirror get the uri by the var ctx.var.uri, this is the difference in implementation.

Again, how are users supposed to know this? It's an implementation detail. 

The requirements make sense from a functional point of view.

tzssangglass: Since `proxy-mirror` support rewrite `path`, so we can implement the above requirements, but we just need to configure them separately on `proxy-mirror` and `proxy-rewrite`.

Of course, we can also modify the `proxy-mirror` plugin to read the `ctx.var.upstream_uri` variable. Of course, it won't be in this PR.

tzssangglass: > we can also modify the proxy-mirror plugin to read the ctx.var.upstream_uri variable

in this way, we can match the requirements

docs/en/latest/terminology/plugin.md
tzssangglass: Yes, it is most directly to make `priority` one of the properties of the plugin. just like `disable`, what your option? @spacewander

spacewander: No. Setting the new field directly on the configuration may override people's current configuration. We should avoid break change

spacewander: I have submitted a PR to explain it in the comment: https://github.com/apache/apisix/pull/7290

bzp2010 (approved):

PR #7316, opened 2022-06-23
rockspec: Add rsa rockspec

zhengwe:
Add Asymmetric encryption and decryption RSA library
hope to Add this library, Can be used when customizing plugins


spacewander (changes requested):
We don't add dependency which is not used in APISIX.
For your custom plugin, please add it to your own rockspec.

PR #7317, opened 2022-06-23
bugfix: when both route and global rules setted proxy-rewrite plugin,…

panhow:
when proxy-rewrite plugin is both set in route and global rule, uri rewrite won't work, cause `ctx.var.upstream_uri` will only sync once to `ngx.var.upstream_uri`

### Description

Fixes # (issue)
How to reproduce it?
```bash
# create a global_rules
curl 127.0.0.1:8081/apisix/admin/global_rules/proxy_set_header -H 'x-api-key: xxx' -XPUT \
	-d '{"plugins":{"proxy-rewrite":{"headers":{"bb": "cc"}}}}'

# create a route for debug
curl 127.0.0.1:8081/apisix/admin/routes/proxy-rewrite-test -H 'x-api-key: xxx'  -XPUT -d '{
	"uri": "/server/*",
	"host": "apisix.debug",
	"plugins": {
		"proxy-rewrite":{
			"regex_uri": [
			    "/server/(.*)",
                             "/$1"
			]
		}
	},
	"upstream": {
		"type": "roundrobin",
		"nodes": {
			"127.0.0.1:7777": 1
		}
	}
}'

# run an http server
python -m SimpleHTTPServer 7777 &

# send a requests
curl 127.0.0.1/server/123123 -H 'host: apisix.debug'
```

```bash
# expect access log
/123123

# actual log
/server/123123
```

```bash
# remove global rules, everything works fine
curl 127.0.0.1:8081/apisix/admin/global_rules/proxy_set_header -H 'x-api-key: xxx' -DELETE
curl 127.0.0.1/server/123123 -H 'host: apisix.debug'
```

### Checklist

### DIFF
--- a/apisix/plugins/proxy-rewrite.lua
+++ b/apisix/plugins/proxy-rewrite.lua
@@ -185,6 +185,7 @@ function _M.rewrite(conf, ctx)
         upstream_uri = core.utils.uri_safe_encode(upstream_uri)
     end
 
+    ctx.var.upstream_uri = nil
     if ctx.var.is_args == "?" then
         if index then
             ctx.var.upstream_uri = upstream_uri .. "&" .. (ctx.var.args or "")


withlin:
@tokers  @spacewander ping

tokers:
@panhow We need some test cast cases to verify the changes are effective.

apisix/plugins/proxy-rewrite.lua
```diff
@@ -185,6 +185,7 @@ function _M.rewrite(conf, ctx)
         upstream_uri = core.utils.uri_safe_encode(upstream_uri)
     end
 
+    ctx.var.upstream_uri = nil
```
tzssangglass, line 188: If this line is added, it seems that `upstream_uri` will be set to nil whenever the `proxy-rewrite` plugin is used.

It doesn't look like a good fix. We should consider multiple scenarios.

Jary-mf: maybe ctx.var.upstream_uri = nil is fine ? because there will be unconditionally assigned to ctx.var.upstream_uri again。see lines 189 to 197

panhow: it DOES NOT a good fix, but works

actually there are two problems to fix:
1. `proxy-rewrite` should not set `ctx.var.upstream_uri` when the `uri` or `regex_uri` isn't set
2. `ngx.var.xxx` can be set once at the time when `ctx.var.xxx` is first assign, after the metamthod `__newIndex` run , other assignment won't pass to the `ngx.var.xxx`

Jary-mf:
BTW，although there is no uri or regex_uri configuration in global_rules, the code logic still affects upstream_uri, which seems a bit unreasonable :)

apisix/plugins/proxy-rewrite.lua
spacewander: The fix LGTM ( I can't give a better fix yet). Would be better to add a comment that we purge the cache by assigning a nil value.

spacewander:
> BTW，although there is no uri or regex_uri configuration in global_rules, the code logic still affects upstream_uri, which seems a bit unreasonable :)

We submit another PR to solve it.

apisix/plugins/proxy-rewrite.lua
spacewander: > @panhow We need some test cast cases to verify the changes are effective.

@panhow 
You can add the test in https://github.com/apache/apisix/blob/master/t/plugin/proxy-rewrite3.t.
You can refer to https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md to write and run the test locally.

spacewander:
> @panhow We need some test cast cases to verify the changes are effective.

@panhow 
You can add the test in https://github.com/apache/apisix/blob/master/t/plugin/proxy-rewrite3.t.
You can refer to https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md to write and run the test locally.

PR #7355, opened 2022-06-29
feat: add hostname config

mikawudi:
hostname use to tag the node
config it in local config file is easy then change server hostname.....
and sometime we can‘t change the server’s hostname

### DIFF
--- a/apisix/core/utils.lua
+++ b/apisix/core/utils.lua
@@ -253,6 +253,13 @@ function _M.gethostname()
         return hostname
     end
 
+    local local_conf = config_local.local_conf()
+    local config_hostname = table.try_read_attr(local_conf, "apisix", "hostname")
+    if config_hostname then
+        hostname = config_hostname
+        return hostname
+    end
+
     local hd = io_popen("/bin/hostname")
     local data, err = hd:read("*a")
     if err == nil then
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -20,6 +20,7 @@
 #
 
 apisix:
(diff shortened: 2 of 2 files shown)


tokers:
@mikawudi Could you add some test cases to cover it?

mikawudi:
@tokers any docs for write test?.......i can't understanding the  program language of .t file

apisix/core/utils.lua
```diff
@@ -253,6 +253,13 @@ function _M.gethostname()
         return hostname
     end
 
+    local local_conf = config_local.local_conf()
```
spacewander, line 256: Why do we need to configure the hostname in an API gateway instead of the environment it is running on? This way is an anti-pattern.

spacewander (changes requested):
I think adding hostname in the application is not a good idea, and it can't solve the problem if other code read the hostname from the environment directly.

apisix/core/utils.lua
tzssangglass: In some runtime scenarios, the hostname is not known before startup, but only at runtime, and the hostname cannot be configured before startup.

mikawudi: emm,英文不大好,描述不清晰撒
我们存在这种场景,以前有的机器没修改过hostname,阿里云默认给的是一堆乱码,然后这个乱码的host被别的组拿去解析了,现在没法修改hostname,然后apisix部署在这个机器上...日志里面的和dashboard里看到的hostname就都是乱码了,确实场景不是很合适的话我这边直接把pr关了撒~周五周六忙apisix落地没啥时间回复信息,见谅~

PR #7384, opened 2022-07-05
fix(limit-count): conf with group and disable can't be configured

### LINKED ISSUE #7375: bug: limit-count plugin,"group" peoperty result in 404 code
### Current Behavior

hello , when I use limit count plugin, I add two routes in etcd，the config like this below
route A:
`
{"id":"414182934216442808","create_time":1656401866,"update_time":1656908750,"uri":"/saveJsonUser*","name":"test1111","methods":["GET","POST","PUT","DELETE","PATCH","HEAD","OPTIONS","CONNECT","TRACE"],"plugins":{"limit-count":{"count":2,"disable":false,"group":"123456","key_type":"var","policy":"local","rejected_code":503,"time_window":30}},"upstream":{"nodes":{"172.xxx.xxx.xxx:8081":1},"timeout":{"connect":6,"send":6,"read":6},"type":"roundrobin","scheme":"http","pass_host":"pass","keepalive_pool":{"idle_timeout":60,"requests":1000,"size":320}},"status":1}
`
Route B:
`
 {"id":"414220935969637304","create_time":1656424517,"update_time":1656901963,"uri":"/getIp","name":"getIp","priority":2,"methods":["GET","POST","PUT","DELETE","PATCH","HEAD","OPTIONS","CONNECT","TRACE"],"plugins":{"limit-count":{"count":2,"disable":false,"group":"123456","key_type":"var","policy":"local","rejected_code":503,"time_window":30}},"upstream":{"nodes":{"172.xxx.xxx.xxx:8081":1},"timeout":{"connect":6,"send":6,"read":6},"type":"roundrobin","scheme":"http","pass_host":"pass","keepalive_pool":{"idle_timeout":60,"requests":1000,"size":320}},"status":1}
`
  you can see, I opend limit-count with the same propery "group" and other properties, but when I test invoke /getIp and /saveJsonUser, only one path cloud access, the other return 404, but when I just config one route (such as only routeA or routeB)，the access is OK. 

### Expected Behavior

I expect when I invoke routeA , it returns OK,  when I invoke routeB , it returns OK also, when I used the count up ，it returns 503，because the two routes is in same "group"

### Error Logs

 I add some logs in source code, I find the problem occurs in the following:
   apisix/plugins/limit-count.lua  -> method name: check_schema
   `

     for _, field in ipairs(fields) do
            if not core.table.deep_eq(prev_conf[field], conf[field]) then
                core.log.error("previous limit-conn group ", prev_conf.group,
                        " conf: ", core.json.encode(prev_conf))
                core.log.error("current limit-conn group ", conf.group,
                        " conf: ", core.json.encode(conf))
                return false, "group conf mismatched"
            end
        end
   `
 "prev_conf" has property "disable", but "conf" does not has propety "disable", so deep_eq return false.

 then I use some way to avoid this, I find in apisix/plugin.lua -> method name:check_schema  has following codes:
`
 
    if plugin_obj.check_schema then
            local disable = plugin_conf.disable
            plugin_conf.disable = nil

            local ok, err = plugin_obj.check_schema(plugin_conf, schema_type)
            if not ok then
                return false, "failed to check the configuration of plugin "
                              .. name .. " err: " .. err
            end

            plugin_conf.disable = disable
        end
`
  I do not understand why should "plugin_conf.disable = nil" before invoke "plugin_obj.check_schema", when I remove the "plugin_conf.disable = nil", the bug is disappear. I want to konw why should "plugin_conf.disable = nil"

 I do not know if this is a bug, If not ,please tell me if my config has error or I ingore some other things.

Thank you!!!

### Steps to Reproduce

1. create two routes
2. add limit count in the two routes,as below:

`

    {"limit-count":{"count":2,"disable":false,"group":"123456","key_type":"var","policy":"local","rejected_code":503,"time_window":30}}

`
3.invode routeA and routeB , only one route could access, the other return 404

### Environment

- APISIX version (run `apisix version`):v2.12.0
- Operating system (run `uname -a`):3.10.0-957.21.3.el7.x86_64
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):openresty/1.19.9.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):3.5.0

soulbird:
### Description

Fixes #7375 

### Checklist

### DIFF
--- a/apisix/plugins/limit-count.lua
+++ b/apisix/plugins/limit-count.lua
@@ -139,7 +139,7 @@ local _M = {
 
 
 local function group_conf(conf)
-    return conf
+    return core.table.deepcopy(conf)
 end
 
 
--- a/apisix/plugins/limit-count.lua
+++ b/apisix/plugins/limit-count.lua
@@ -129,6 +129,7 @@ local schema = {
     }
 }
 
+local schema_copy = core.table.deepcopy(schema)
 
 local _M = {
     version = 0.4,
@@ -139,7 +140,7 @@ local _M = {
 
 
 local function group_conf(conf)
-    return core.table.deepcopy(conf)
+    return conf
 end
 
 
@@ -151,7 +152,10 @@ function _M.check_schema(conf)
 
     if conf.group then
         local fields = {}
-        for k in pairs(schema.properties) do
+        -- When the goup field is configured,
+        -- we will use schema_copy to get the whitelist of properties,
+        -- so that we can avoid getting injected properties.
+        for k in pairs(schema_copy.properties) do
             tab_insert(fields, k)
         end
         local extra = policy_to_additional_properties[conf.policy]
--- a/t/plugin/limit-count2.t
+++ b/t/plugin/limit-count2.t
@@ -767,7 +767,7 @@ limit key: afafafhao2:remote_addr
 
 
 
-=== TEST 21: group with disable
+=== TEST 22: group with disable
 --- config
     location /t {
         content_by_lua_block {
--- a/t/plugin/limit-count2.t
+++ b/t/plugin/limit-count2.t
@@ -764,3 +764,40 @@ limit key: afafafhao2:remote_addr
 limit key: afafafhao2:remote_addr
 --- response_body
 [200,200,503,503]
+
+
+
+=== TEST 21: group with disable
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/services/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                    "plugins": {
+                        "limit-count": {
+                            "count": 2,
+                            "time_window": 60,
+                            "rejected_code": 503,
+                            "group": "abcd",
+                            "disable": false
+                        }
+                    },
+                    "upstream": {
+                        "nodes": {
+                            "127.0.0.1:1980": 1
+                        },
+                        "type": "roundrobin"
+                    }
+                }]]
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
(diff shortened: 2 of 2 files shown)


spacewander (commented):
https://github.com/apache/apisix/blob/2f7833e07479610ed5c8148399c347df85c20b96/apisix/plugins/limit-count.lua#L153-L168

We use a whitelist for the fields which need to be compared, and `disable` is expected not in the list?

soulbird:
> https://github.com/apache/apisix/blob/2f7833e07479610ed5c8148399c347df85c20b96/apisix/plugins/limit-count.lua#L153-L168
> 
> We use a whitelist for the fields which need to be compared, and `disable` is expected not in the list?

Yes, it's injected in: https://github.com/apache/apisix/blob/master/apisix/plugin.lua#L139-L162

soulbird:
The whitelist for the fields  does not work as it should, because the schema may be injected into other fields, currently there are `disable` `_meta` `$comment`. A better approach is to move all injected fields into _meta and manually filter out this field in the whitelist. But this would introduce a break change, not sure if it's worth it. cc @spacewander

apisix/plugins/limit-count.lua
```diff
@@ -139,7 +139,7 @@ local _M = {
 
 
 local function group_conf(conf)
-    return conf
+    return core.table.deepcopy(conf)
```
spacewander, line 142: What about maintaining a deep copy of schema, so we only need to deep copy once?
https://github.com/apache/apisix/blob/a8a692d9a58a9a0185e22cfebbe2c0f000ba41f3/apisix/plugins/limit-count.lua#L86

spacewander (approved):

tzssangglass (approved):

PR #7385, opened 2022-07-05
docs: fix command format

oil-oil:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/docs/en/latest/plugins/basic-auth.md
+++ b/docs/en/latest/plugins/basic-auth.md
@@ -97,7 +97,7 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
 After you have configured the Plugin as mentioned above, you can make a request to the Route as shown below:
 
 ```shell
-curl -i -ufoo:bar http://127.0.0.1:9080/hello
+curl -i -u foo:bar http://127.0.0.1:9080/hello
 ```
 
 ```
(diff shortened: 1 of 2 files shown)


docs/en/latest/plugins/basic-auth.md
```diff
@@ -97,7 +97,7 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f13
 After you have configured the Plugin as mentioned above, you can make a request to the Route as shown below:
 
 ```shell
-curl -i -ufoo:bar http://127.0.0.1:9080/hello
+curl -i -u foo:bar http://127.0.0.1:9080/hello
```
spacewander, line 100: Both `-ufoo:bar` and `-u foo:bar` are equal?

oil-oil: > Both `-ufoo:bar` and `-u foo:bar` are equal?

I'm not sure if it's equal, but I think adding a space would make it clearer.

tzssangglass: `-u foo:bar` means means `foo:bar` is `<user:password>`, and here `-u` equal `--user`? see: https://curl.se/docs/manpage.html and search `-u, --user <user:password>`

tokers: I think it's equal but `-u foo:bar` will be clear.

PR #7412, opened 2022-07-08
feat(promethues): add ngx.shared.DICT status

### LINKED ISSUE #5837: request: Expect monitoring metrics to add monitoring of nginx shared memory usage
### Issue description

期待监控指标添加对nginx共享内存使用情况的监控

### Environment

- OS (cmd: `uname -a`):

ccxhwmy:
### Description

Fixes #5837 
Use the request header to carry the name of ngx.shared.DICT that needs to be fetched, use `ngx.shared.DICT.capacity` and `ngx.shared.DICT.free_space` to get the current status of `ngx.shared.DICT`, and use `gauge` to store the status
### Checklist

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -33,7 +33,6 @@ local get_services = require("apisix.http.service").services
 local get_consumers = require("apisix.consumer").consumers
 local get_upstreams = require("apisix.upstream").upstreams
 local clear_tab = core.table.clear
-local concat_tab = core.table.concat
 local get_stream_routes = router.stream_routes
 local get_protos = require("apisix.plugins.grpc-transcode.proto").protos
 local service_fetch = require("apisix.http.service").get
@@ -123,9 +122,13 @@ function _M.http_init(prometheus_enabled_in_stream)
             "Etcd modify index for APISIX keys",
             {"key"})
 
-    metrics.shared_dict = prometheus:gauge("shared_dict",
-            "nginx shared DICT of APISIX",
-            {"key"})
+    metrics.shared_dict_capacity_bytes = prometheus:gauge("shared_dict_capacity_bytes",
+            "capacity every moment of nginx shared DICT since APISIX start",
+            {"name"})
+
+    metrics.shared_dict_free_space_bytes = prometheus:gauge("shared_dict_free_space_bytes",
+            "the current free space for nginx shared DICT in APISIX",
+            {"name"})
 
     -- per service
 
@@ -356,42 +359,40 @@ local function etcd_modify_index()
 
 end
 
-local key = {}
-local combine_param = {}
 
 local function shared_dict_status()
-    local header_of_shared_dict = core.request.header(ngx.ctx, "shared_dict")
-    if not header_of_shared_dict then
+    local attr = plugin.plugin_attr(plugin_name)
+    if not attr or not attr.shared_DICT then
         return
     end
 
+    local name = {}
     local set_shared_dict_param = function (shared_dict_name)
         if type(shared_dict_name) ~= "string" then
             return
         end
+
+        name[1] = shared_dict_name
         local share_dict = ngx.shared[shared_dict_name]
         if share_dict then
-            combine_param[1] = shared_dict_name
-            combine_param[2] = "capacity"
-            key[1] = concat_tab(combine_param, "_")
-            metrics.shared_dict:set(share_dict:capacity(), key)
-
-            combine_param[2] = "free_space"
-            key[1] = concat_tab(combine_param, "_")
-            metrics.shared_dict:set(share_dict:free_space(), key)
+            metrics.shared_dict_capacity_bytes:set(share_dict:capacity(), name)
+            metrics.shared_dict_free_space_bytes:set(share_dict:free_space(), name)
         end
     end
 
-    if type(header_of_shared_dict) == "string" then
-        set_shared_dict_param(header_of_shared_dict)
+    local shared_dicts = attr.shared_DICT
+
+    if type(shared_dicts) == "string" then
+        set_shared_dict_param(shared_dicts)
     end
 
-    if type(header_of_shared_dict) == "table" then
-        for _, shared_dict_name in ipairs(header_of_shared_dict) do
+    if type(shared_dicts) == "table" then
+        for _, shared_dict_name in ipairs(shared_dicts) do
             set_shared_dict_param(shared_dict_name)
         end
     end
 end
+
 
 local function collect(ctx, stream_only)
     if not prometheus or not metrics then
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -360,7 +360,7 @@ local key = {}
 local combine_param = {}
 
 local function shared_dict_status()
-    local header_of_shared_dict = ngx.req.get_headers()["shared_dict"]
+    local header_of_shared_dict = core.request.header(ngx.ctx, "shared_dict")
     if not header_of_shared_dict then
         return
     end
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -34,7 +34,6 @@ local get_services = require("apisix.http.service").services
 local get_consumers = require("apisix.consumer").consumers
 local get_upstreams = require("apisix.upstream").upstreams
 local clear_tab = core.table.clear
-local insert_tab = core.table.insert
 local get_stream_routes = router.stream_routes
 local get_protos = require("apisix.plugins.grpc-transcode.proto").protos
 local service_fetch = require("apisix.http.service").get
@@ -362,32 +361,13 @@ local function etcd_modify_index()
 end
 
 
-local shared_dicts
 local function shared_dict_status()
-    local config_indexs = {"http", "stream", "meta"}
-    if not shared_dicts then
-        shared_dicts = {}
-        local existed = {}
-        for _, config_index in ipairs(config_indexs) do
-            local config = plugin.nginx_config(config_index)
-            if config and config.lua_shared_dict then
-                for shared_dict_name, _ in pairs(config.lua_shared_dict) do
-                    if existed[shared_dict_name] == nil then
-                        insert_tab(shared_dicts, shared_dict_name)
-                        existed[shared_dict_name] = true
-                    end
-                end
-            end
-        end
-    end
-
     local name = {}
-    for _, shared_dict_name in ipairs(shared_dicts) do
-        name[1] = shared_dict_name
-        local share_dict = ngx.shared[shared_dict_name]
-        if share_dict then
-            metrics.shared_dict_capacity_bytes:set(share_dict:capacity(), name)
-            metrics.shared_dict_free_space_bytes:set(share_dict:free_space(), name)
+    for shared_dict_name, shared_dict in pairs(ngx.shared) do
+        if shared_dict then
+            name[1] = shared_dict_name
+            metrics.shared_dict_capacity_bytes:set(shared_dict:capacity(), name)
+            metrics.shared_dict_free_space_bytes:set(shared_dict:free_space(), name)
         end
     end
 end
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -399,7 +399,7 @@ local function collect(ctx, stream_only)
         return 500, {message = "An unexpected error occurred"}
     end
 
-    --
+    -- collect ngx.shared.DICT status
     shared_dict_status()
 
     -- across all services
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -18,6 +18,7 @@ local base_prometheus = require("prometheus")
 local core      = require("apisix.core")
 local plugin    = require("apisix.plugin")
 local ipairs    = ipairs
+local pairs     = pairs
 local ngx       = ngx
 local re_gmatch = ngx.re.gmatch
 local ffi       = require("ffi")
@@ -33,6 +34,7 @@ local get_services = require("apisix.http.service").services
 local get_consumers = require("apisix.consumer").consumers
 local get_upstreams = require("apisix.upstream").upstreams
 local clear_tab = core.table.clear
+local insert_tab = core.table.insert
 local get_stream_routes = router.stream_routes
 local get_protos = require("apisix.plugins.grpc-transcode.proto").protos
 local service_fetch = require("apisix.http.service").get
@@ -123,11 +125,11 @@ function _M.http_init(prometheus_enabled_in_stream)
             {"key"})
 
     metrics.shared_dict_capacity_bytes = prometheus:gauge("shared_dict_capacity_bytes",
-            "capacity every moment of nginx shared DICT since APISIX start",
+            "The capacity of each nginx shared DICT since APISIX start",
             {"name"})
 
     metrics.shared_dict_free_space_bytes = prometheus:gauge("shared_dict_free_space_bytes",
-            "the current free space for nginx shared DICT in APISIX",
+            "The free space of each nginx shared DICT since APISIX start",
             {"name"})
 
     -- per service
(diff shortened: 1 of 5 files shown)


apisix/plugins/prometheus/exporter.lua
```diff
@@ -351,6 +356,42 @@ local function etcd_modify_index()
 
 end
 
+local key = {}
+local combine_param = {}
+
+local function shared_dict_status()
+    local header_of_shared_dict = ngx.req.get_headers()["shared_dict"]
+    if not header_of_shared_dict then
```
soulbird, line 364: Let's use `core.request.header`, so that we can use cache.

t/plugin/prometheus.t
```diff
@@ -632,3 +632,61 @@ GET /apisix/prometheus/metrics
...
+--- more_headers
+Shared_DICT: worker-events
+Shared_DICT: upstream-healthcheck
+Shared_DICT: internal-status
+--- response_body_like
+.*TYPE apisix_shared_dict gauge.*upstream-healthcheck_capacity.*upstream-healthcheck_free_space.*
+
+
+
+=== TEST 45: fetch the prometheus multiple shared dict data
```
soulbird, line 672: Why use the same title? Are their test intents the same? It's confusing

docs/zh/latest/plugins/prometheus.md
```diff
@@ -186,8 +186,11 @@ scrape_configs:
     | node         | 消费者节点 IP 地址。 |
 
 - etcd reachability: APISIX 连接 etcd 的可用性，用 0 和 1 来表示，`1` 表示可用，`0` 表示不可用。
+
```
tzssangglass, line 189: we don't need a blank line here, elsewhere as well

apisix/plugins/prometheus/exporter.lua
```diff
@@ -351,6 +356,42 @@ local function etcd_modify_index()
...
+            combine_param[1] = shared_dict_name
+            combine_param[2] = "capacity"
+            key[1] = concat_tab(combine_param, "_")
+            metrics.shared_dict:set(share_dict:capacity(), key)
+
+            combine_param[2] = "free_space"
+            key[1] = concat_tab(combine_param, "_")
+            metrics.shared_dict:set(share_dict:free_space(), key)
+        end
+    end
```
tzssangglass, line 383: can we move this function between `local combine_param = {}` and `local function shared_dict_status()`?

apisix/plugins/prometheus/exporter.lua
```diff
@@ -351,6 +356,42 @@ local function etcd_modify_index()
 
 end
 
+local key = {}
+local combine_param = {}
```
tzssangglass, line 360: can we init `key` and `combine_param` as table if `header_of_shared_dict` is true?

ccxhwmy: > can we init `key` and `combine_param` as table if `header_of_shared_dict` is true?

I consider the problem of repeated malloc for table `key` and `combine_param`.

tzssangglass: Do you mean that you avoid creating tables every time? Sounds reasonable.

apisix/plugins/prometheus/exporter.lua
ccxhwmy: > can we move this function between `local combine_param = {}` and `local function shared_dict_status()`?

can you tell me why do this?

apisix/plugins/prometheus/exporter.lua
```diff
@@ -122,6 +123,10 @@ function _M.http_init(prometheus_enabled_in_stream)
             "Etcd modify index for APISIX keys",
             {"key"})
 
+    metrics.shared_dict = prometheus:gauge("shared_dict",
+            "nginx shared DICT of APISIX",
+            {"key"})
```
spacewander, line 128: This is not a standard way to name a prometheus metric.
We should distinguish them with name, instead of abusing the label.

They should be:
```
apisix_shared_dict_free_space{name="internal-status"}
...
```
and the unit should be byte.

See https://prometheus.io/docs/practices/naming/#metric-names

apisix/plugins/prometheus/exporter.lua
```diff
@@ -351,6 +356,42 @@ local function etcd_modify_index()
 
 end
 
+local key = {}
+local combine_param = {}
+
+local function shared_dict_status()
+    local header_of_shared_dict = core.request.header(ngx.ctx, "shared_dict")
```
spacewander, line 363: People don't use Prometheus in that way. Usually the application developers decide what metrics will be exposed via their local configuration, but not Prometheus's configuration.

spacewander: Not even said, the Prometheus doesn't support custom header AFAIK.

apisix/plugins/prometheus/exporter.lua
ccxhwmy: > 

Modify the initialization with
```lua
    metrics.shared_dict_capacity_bytes = prometheus:gauge("shared_dict_capacity_bytes",
            "nginx shared DICT capacity of APISIX",
            {"name"})

    metrics.shared_dict_free_space_bytes = prometheus:gauge("shared_dict_free_space_bytes",
            "nginx shared DICT free_space of APISIX",
            {"name"})
```
Is my understanding correct?

spacewander: Yes. The description can be more clear, like `The total number of client requests since APISIX started`

apisix/plugins/prometheus/exporter.lua
tzssangglass: I thought it over and it didn't seem necessary. ignore this.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -122,6 +122,14 @@ function _M.http_init(prometheus_enabled_in_stream)
             "Etcd modify index for APISIX keys",
             {"key"})
 
+    metrics.shared_dict_capacity_bytes = prometheus:gauge("shared_dict_capacity_bytes",
+            "capacity every moment of nginx shared DICT since APISIX start",
```
spacewander, line 126: I am afraid I can't understand what this means...
Could you change the description?

apisix/plugins/prometheus/exporter.lua
```diff
@@ -352,13 +360,50 @@ local function etcd_modify_index()
...
 
 
+local function shared_dict_status()
+    local attr = plugin.plugin_attr(plugin_name)
+    if not attr or not attr.shared_DICT then
+        return
+    end
+
+    local name = {}
+    local set_shared_dict_param = function (shared_dict_name)
```
spacewander, line 370: We can put this in the for loop instead of defining a new function to solve it?
The function definition can't be JIT and should be avoided.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -352,13 +360,50 @@ local function etcd_modify_index()
 end
 
 
+local function shared_dict_status()
+    local attr = plugin.plugin_attr(plugin_name)
```
spacewander, line 364: We can list the shdict in https://github.com/apache/apisix/blob/master/apisix/cli/ngx_tpl.lua, instead of requiring the users to configure them? Since the usage of shdict is an implementation detail.

apisix/plugins/prometheus/exporter.lua
ccxhwmy: > We can put this in the for loop instead of defining a new function to solve it? The function definition can't be JIT and should be avoided.

Would you provide doc about it? I would like to learn more about it.

apisix/plugins/prometheus/exporter.lua
spacewander: Something like:
```
{
    "prometheus-metrics",
    "internal-status",
     ...
}
```

To make it simple, we can only provide the must-have shdicts in this first version.

spacewander: I change my mind. We can iterate `ngx.shared` to get all the shdict.

apisix/plugins/prometheus/exporter.lua
spacewander: http://wiki.luajit.org/NYI#bytecode, see FNEW.

apisix/plugins/prometheus/exporter.lua
ccxhwmy: > We can list the shdict in https://github.com/apache/apisix/blob/master/apisix/cli/ngx_tpl.lua, instead of requiring the users to configure them? Since the usage of shdict is an implementation detail.

This is what I was trying to do at first, but I don't know how to get all the shdicts.

ccxhwmy: > I change my mind. We can iterate `ngx.shared` to get all the shdict.

I modified to get all ngx.shared.DICT, I found it not a hard job.

apisix/plugins/prometheus/exporter.lua
ccxhwmy: > http://wiki.luajit.org/NYI#bytecode, see FNEW.

Thank you~

apisix/plugins/prometheus/exporter.lua
```diff
@@ -352,13 +362,47 @@ local function etcd_modify_index()
 end
 
 
+local shared_dicts
+local function shared_dict_status()
+    local config_indexs = {"http", "stream", "meta"}
+    if not shared_dicts then
+        shared_dicts = {}
+        local existed = {}
```
tzssangglass, line 370: what `existed` is used for? It looks like `shared_dicts`  will only be `nil` or has `shared_dict_name`.

ccxhwmy: > what `existed` is used for? It looks like `shared_dicts` will only be `nil` or has `shared_dict_name`.

I found that we will configure duplicate `lua_shared_dict` in https://github.com/apache/apisix/blob/master/conf/config-default.yaml, like `prometheus-metrics`, `existed` is used to deduplicate.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -352,13 +362,47 @@ local function etcd_modify_index()
 end
 
 
+local shared_dicts
+local function shared_dict_status()
+    local config_indexs = {"http", "stream", "meta"}
+    if not shared_dicts then
```
spacewander, line 368: Can't we just iterate the `ngx.shared`?

ccxhwmy: > Can't we just iterate the `ngx.shared`?

I was stupid...

tzssangglass (approved):

t/plugin/prometheus.t
```diff
@@ -632,3 +632,53 @@ GET /apisix/prometheus/metrics
...
+apisix_shared_dict_free_space_bytes{name="worker-events"} \d+.*/,
+qr/.*apisix_shared_dict_capacity_bytes{name="upstream-healthcheck"} \d+(?:.|\n)*
+apisix_shared_dict_free_space_bytes{name="upstream-healthcheck"} \d+.*/,
+qr/.*apisix_shared_dict_capacity_bytes{name="internal-status"} \d+(?:.|\n)*
+apisix_shared_dict_free_space_bytes{name="internal-status"} \d+.*/]
+
+
+
+=== TEST 44: fetch the prometheus multiple shared dict data contain not exist shared dict
+--- yaml_config
```
soulbird, line 673: Looks like we don't need this test case anymore?

ccxhwmy: I consider it as an example to show that `ngx.shared` is determined not only by the `yaml` file but also by the `ngx_tpl.lua` file.

soulbird: 1. The final result of `ngx_tpl.lua` depends on the configuration of `yaml`, I don't understand what you mean, and your test cases are written like this:
````
--- yaml_config
nginx_config:
   http:
     lua_shared_dict:
````
2. You better indicate your test intent in the title of the test case.

Comment (tzssangglass) [reply t/plugin/prometheus.t:673]: > 1. The final result of `ngx_tpl.lua` depends on the configuration of `yaml`, I don't understand what you mean, and all of your test cases are written like this:

Not exactly true, we have some `lua_shared_dict` defined directly in `ngx_tpl.lua` which does not appear in `config-default.yaml`, like `xds-config` `xds-config-version`, but maybe we can optimize it later.

Comment (soulbird) [reply t/plugin/prometheus.t:673]: > nginx_config:
>   http:
>     lua_shared_dict:
>       worker-events: 10m
>       upstream-healthcheck: 10m
>       not-exist-shared-dict: 10m
> --- request
> GET /apisix/prometheus/metrics
> --- response_body_unlike
> qr/.*apisix_shared_dict_capacity_bytes{name="not-exist-shared-dict"} \d+(?:.|\n)*
> apisix_shared_dict_free_space_bytes{name="not-exist-shared-dict"} \d+.*/

Got it, but I don't think it's necessary to test whether shared dict works or not.

Comment (soulbird) [reply t/plugin/prometheus.t:673]: > > 1. The final result of `ngx_tpl.lua` depends on the configuration of `yaml`, I don't understand what you mean, and all of your test cases are written like this:
> 
> Not exactly true, we have some `lua_shared_dict` defined directly in `ngx_tpl.lua` which does not appear in `config-default.yaml`, like `xds-config` `xds-config-version`, but maybe we can optimize it later.

Got it, but I don't think it's necessary to test whether shared dict works or not.

Comment (tzssangglass) [reply t/plugin/prometheus.t:673]: I think this is just another way of describing the correctness of the returned data

Comment (ccxhwmy) [reply t/plugin/prometheus.t:673]: `yaml_config` can determine whether `ngx.shared` exists. For custom `ngx.shared`, the configuration of `yaml_config` is a necessary condition, but it is not a sufficient condition. Here it can be verified that `yaml_config` is not sufficient condition.

Comment (spacewander) [reply t/plugin/prometheus.t:673]: Also the lua_shared_dict here is just to control the size of shdict. Adding the configuration or not doesn't affect the lua_shared_dict  in `nginx.conf`.

Comment (ccxhwmy) [reply t/plugin/prometheus.t:673]: > > Here it can be verified that yaml_config is not sufficient condition.
> 
> We already have some lua_shared_dict defined in the yaml by default. Is it necessary to override the default yaml to verify it?

I tested and found that APISIX will use `config-default.yaml` as the configuration when running the test case, my understanding was wrong.

Comment (ccxhwmy) [reply t/plugin/prometheus.t:673]: I tested and found that APISIX will use `config-default.yaml` as the configuration when running the test case, my understanding was wrong.

**Review by spacewander — COMMENTED** [review]

**Comment by ccxhwmy** [comment]

I don't know why this [ci error](https://github.com/apache/apisix/runs/7362090026?check_suite_focus=true#step:18:829) is reported here, Would you give me some suggestion.

**Comment by spacewander** [comment]
> > Please make the CI pass, thanks!
> 
> I don't know why this [ci error](https://github.com/apache/apisix/runs/7362090026?check_suite_focus=true#step:18:829) is reported here, Would you give me some suggestion.

The error message is:
```
Bailout called.  Further testing stopped:  t/plugin/prometheus.t TEST 42: fetch the prometheus multiple shared dict data - You asked for many requests, the expected results should be arrays as well.
FAILED--Further testing stopped: t/plugin/prometheus.t TEST 42: fetch the prometheus multiple shared dict data - You asked for many requests, the expected results should be arrays as well.
```

You can look up it.

**Comment by tzssangglass** [comment]
> The error message report that `You asked for many requests, the expected results should be arrays as well.`, but the test case is:

It does seem strange, so if you don't know how to modify it, consider a different type of validation.

**Comment by ccxhwmy** [comment]
> > The error message report that `You asked for many requests, the expected results should be arrays as well.`, but the test case is:
> 
> It does seem strange, so if you don't know how to modify it, consider a different type of validation.

You are right! I found a new type of validation, it maybe more suitable.

**Review by spacewander — DISMISSED** [review]

File: t/plugin/prometheus.t [hunk 14]
```diff
@@ -632,3 +632,14 @@ GET /apisix/prometheus/metrics
...
+
+
+
+=== TEST 42: fetch the prometheus shared dict internal-status data
+--- http_config
+lua_shared_dict test-shared-dict 10m;
+--- request
+GET /apisix/prometheus/metrics
+--- response_body_like
+.*apisix_shared_dict_capacity_bytes{name="test-shared-dict"} \d+(?:.|\n)*
```
spacewander, line 644: Would be better to check the num instead of `\d+`

ccxhwmy:
Is the [error](https://github.com/apache/apisix/runs/7391688526?check_suite_focus=true) a false report?

tzssangglass:
> Is the [error](https://github.com/apache/apisix/runs/7391688526?check_suite_focus=true) a false report?

rerun

spacewander (approved):

soulbird:
@ccxhwmy  could you update the [apisix-grafana-dashboard](https://github.com/apache/apisix/blob/master/docs/assets/other/json/apisix-grafana-dashboard.json) meta for shared dict?

ccxhwmy:
> @ccxhwmy  could you update the [apisix-grafana-dashboard](https://github.com/apache/apisix/blob/master/docs/assets/other/json/apisix-grafana-dashboard.json) meta for shared dict?

Yes, I can.

soulbird:
> > @ccxhwmy  could you update the [apisix-grafana-dashboard](https://github.com/apache/apisix/blob/master/docs/assets/other/json/apisix-grafana-dashboard.json) meta for shared dict?
> 
> Yes, I can.

Thanks a lot, let's do it in another PR.

ccxhwmy:
I have update the [apisix-grafana-dashboard](https://github.com/apache/apisix/blob/master/docs/assets/other/json/apisix-grafana-dashboard.json) meta for shared dict in the (PR)[https://github.com/apache/apisix/pull/7529].

Issue #7178, opened 2022-06-01
help request: about eureka when apisix reload


kongjun01:
### Description

when apisix the reload, the eureka data is not ready (fetch_full_registry is not over), and there are many 502. Is there any solution.

### Environment

- APISIX version (run `apisix version`):2.10.5
- Operating system (run `uname -a`):
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

tzssangglass:
I think discovery mode has this problem if it's a large amount of data.

We should allow users to specify that they have to wait for the fetch data to finish before providing the service, then APISIX will be interrupted during reload.

soulbird:
> One solution is to let APISIX fetch all data from eureka in the `init_worker` phase. Just like the way we load data from ETCD.

Looks like that's what it's doing now.

tzssangglass:
> Looks like that's what it's doing now.

For eureka, it's not.

kongjun01:
> I think discovery mode has this problem if it's a large amount of data.
> 
> We should allow users to specify that they have to wait for the fetch data to finish before providing the service, then APISIX will be interrupted during reload.

I considered that before.Our most front-end is Ali Cloud SLB, with a heartbeat mechanism.When the release is released, [-s quit], wait for SLB to think that APISIX has been offline, but there is a waiting time for heartbeat detection, during which SLB will still have traffic forward

kongjun01:
> One solution is to let APISIX fetch all data from eureka in the `init_worker` phase. Just like the way we load data from ETCD.

The etcd is also asynchronous with the ngx_timer_at function, so how does it guarantee that the apisix provides the service after the data is fully loaded with the etcd?

tokers:
> > One solution is to let APISIX fetch all data from eureka in the `init_worker` phase. Just like the way we load data from ETCD.
> 
> The etcd is also asynchronous with the ngx_timer_at function, so how does it guarantee that the apisix provides the service after the data is fully loaded with the etcd?

See https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L159. That's the default behavior.

soulbird:
[...]

PR #7432, opened 2022-07-11
bugfix: resolution when apisix reload eureka does not load the data

kongjun01:
### Description

https://github.com/apache/apisix/issues/7178

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/discovery/eureka/init.lua
+++ b/apisix/discovery/eureka/init.lua
@@ -24,7 +24,6 @@ local tostring           = tostring
 local type               = type
 local math_random        = math.random
 local ngx                = ngx
-local ngx_timer_at       = ngx.timer.at
 local ngx_timer_every    = ngx.timer.every
 local string_sub         = string.sub
 local str_find           = core.string.find
--- a/apisix/discovery/eureka/init.lua
+++ b/apisix/discovery/eureka/init.lua
@@ -16,7 +16,7 @@
 --
 
 local local_conf         = require("apisix.core.config_local").local_conf()
-local http               = require("resty.http")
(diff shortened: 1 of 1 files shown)


apisix/discovery/eureka/init.lua
```diff
@@ -16,15 +16,14 @@
 --
 
 local local_conf         = require("apisix.core.config_local").local_conf()
-local http               = require("resty.http")
+local socket_http        = require("socket.http")
 local core               = require("apisix.core")
```
soulbird, line 20: We don't need to use blocking sockets at every phase, you can take a look at https://github.com/apache/apisix/blob/master/apisix/patch.lua and make some adjustments

tzssangglass:
In the init and init_worker phases, we can't use `lua-resty-http`, so we use `luasocket`.
But we don't need to replace the runtime procedure that uses `lua-resty-http`.

kongjun01:
> In the init and init_worker phases, we can't use `lua-resty-http`, so we use `luasocket`. But we don't need to replace the runtime procedure that uses `lua-resty-http`.

This method is public at init and runtime.

apisix/discovery/eureka/init.lua
kongjun01: > We don't need to use blocking sockets at every phase, you can take a look at https://github.com/apache/apisix/blob/master/apisix/patch.lua and make some adjustments

Ok，let me try.

tzssangglass:
> This method is public at init and runtime.

I don't know what this means.

But replacing all `lua-resty.http` with `luasocket` is not acceptable. `luasocket` is a blocking call.

Issue #7450, opened 2022-07-13
docs: highlight introductions that need to be configured in different locations


juzhiyuan:
### Background

In the OpenTelemetry plugin's document[1], there have three locations that need to be configured:

1. When creating a Route/Service/Global Plugin, we need to configure attributes like `sampler.xxx`.
2. This plugin is disabled by default, we need to enable this plugin in the `config.yaml` file.
3. When using custom OpenTelemetry Collector's configurations, we need to configure `plugin_attr.opentelemetry.xxx` in the `config.yaml` file.

This case[2] shows that some users will ignore configuration locations when using plugins, even though there has one sentence showing that `You can set the collecting by specifying the configuration in conf/config.yaml.`.

**Wrong Usage:**

![image](https://user-images.githubusercontent.com/2106987/178638833-c0909985-484f-4f61-be8b-7b5e921b84df.png)

[1] https://apisix.apache.org/docs/apisix/plugins/opentelemetry
[2] https://the-asf.slack.com/archives/CUC5MN17A/p1657549138916759

### Desired State

We have nearly 90 plugins, and some other plugins need to be configured separately. I suggest:

1. For those plugins, we need to highlight that the plugin has multiple locations to configure.
2. In the document's different sections, e.g., `Attributes`, `How to enable`, and `How to set collecting`, we should use Docusaurus's `admonitions` feature[1] to highlight which place to configure.

![image](https://user-images.githubusercontent.com/2106987/178639322-fd95a3cc-be6b-4ec4-b125-5b5e7d16bfde.png)

### Related Links

1. https://github.com/apache/apisix/issues/7448
2. https://github.com/apache/apisix-helm-chart/issues/312

[1] https://docusaurus.io/docs/markdown-features/admonitions

PR #7451, opened 2022-07-13
docs: improve OpenTelemetry documentation

juzhiyuan:
### Description

This PR aims to improve the OpenTelemetry plugin's documentation.

Related Issue: https://github.com/apache/apisix/issues/7450

### Checklist

### DIFF
--- a/docs/en/latest/plugins/opentelemetry.md
+++ b/docs/en/latest/plugins/opentelemetry.md
@@ -1,12 +1,13 @@
 ---
-title: opentelemetry
+title: OpenTelemetry
 keywords:
   - APISIX
-  - Plugin
+  - Observability
   - OpenTelemetry
-  - opentelemetry
-description: This document contains information about the Apache opentelemetry Plugin.
+  - API Gateway
+description: OpenTelemetry is an open source Observability project. Follow this documentation to integrate OpenTelemetry with API gateway Apache APISIX to measure application performance.
 ---
+
 <!--
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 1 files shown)


docs/en/latest/plugins/opentelemetry.md
```diff
@@ -28,31 +29,59 @@ description: This document contains information about the Apache opentelemetry P
...
 
-You can set up the collector by configuring it in you configuration file (`conf/config.yaml`):
+```yaml title="conf/config.yaml"
+plugins:
+  - ...
+  - opentelemetry
+  - ...
+```
+
+## Plugin YAML Attributes
```
juzhiyuan, line 57: I was using `Plugin Meta Attributes` to refer to configurations under the `conf/config.yaml` file, but I come up with the `_meta` attributes in plugin, so I use `Plugin YAML Attributes` instead, but I think it's still wired 🤔 

Do you have a better choice? cc @navendu-pottekkat @nfrankel @tzssangglass @bzp2010

docs/en/latest/plugins/opentelemetry.md
```diff
@@ -124,7 +151,7 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1  -H 'X-API-KEY: edd1c9f034335f1
 
 ## Disable Plugin
```
juzhiyuan, line 152: @navendu-pottekkat I'm not sure if we should use `Disabling Plugin` or using `Disable Plugin` 😅 I would prefer `Disable Plugin` because it's shorter and easy to read, but I notice some other products' documentations are using `xxx-doing`.

navendu-pottekkat: Disable Plugin is good IMO as well.

hf400159 (commented):
We should discuss the structure of this type of document. Maybe we can put Load Plugin and Plugin YAML Attributes under "Related Configuration".

docs/en/latest/plugins/opentelemetry.md
```diff
@@ -124,7 +151,7 @@ curl http://127.0.0.1:9080/apisix/admin/routes/1  -H 'X-API-KEY: edd1c9f034335f1
...
 
-To disable the `opentelemetry` Plugin, you can delete the corresponding JSON configuration from the Plugin configuration. APISIX will automatically reload and you do not have to restart for this to take effect.
+To disable this plugin, you can delete the corresponding JSON configuration from the Plugin configuration. APISIX will automatically reload and you do not have to restart for this to take effect.
```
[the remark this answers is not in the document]
hf400159: image
Other APISIX documentation uses this description.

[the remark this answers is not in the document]
juzhiyuan: Why should we keep repeating `the "opentelemetry" plugin` in the documentation? 🤔 

Documentation should be correct, and straightforward IMO.

hf400159:
> Please consider this a Template Update PR, and I take the `opentelemetry` documentation as an example. For this kind of plugin (needs multiple configurations), this PR's structure:

Most plugins require these operations to load plugins. Instead of repeating the description in every document, we can create a document like `Plugin Basic Operation` to illustrate this operation. cc @juzhiyuan

docs/en/latest/plugins/opentelemetry.md
tzssangglass: `Plugin Metadata` is ok for me

juzhiyuan:
> Most plugins require these operations to load plugins. Instead of repeating the description in every document, we can create a document like Plugin Basic Operation to illustrate this operation. cc @juzhiyuan

From the count: in the 80+ plugins, there have only a few plugins that aren't loaded by default in the `config-default.yaml` file.

From users' view:

1. The documentation should be straightforward, correct, transparent, and rich.
2. As a user, I want to follow the documentation to use this feature smoothly without errors. I want to find out all the information I need. I don't like visiting different links here and there. It's a good link to other documentation because they're detailed and could reduce duplicate content. Still, I suggest only picking the essential part of this plugin's documentation.
3. For this specific kind of documentation (I mean `Plugin`), keep the most essential part in the documentation and leave other `Read more` content elsewhere.

juzhiyuan:
Another question: In this post[1], it says, `The opentelemetry plugin is located on the Agent side in the above figure, but currently only supports the trace part, and does not support the logs and metrics protocols of OpenTelemetry.`, may I know current if it supports Logs and Metrics? cc @tzssangglass 

[1] https://apisix.apache.org/blog/2022/02/28/apisix-integration-opentelemetry-plugin/

spacewander:
> > We should discuss the structure of this type of document. Maybe we can put Load Plugin and Plugin YAML Attributes under "Related Configuration".
> 
> Please consider this a Template Update PR, and I take the `opentelemetry` documentation as an example. For this kind of plugin (needs multiple configurations), this PR's structure:
> 
>     1. Description
> 
>     2. Load Plugin
> 
>     3. Plugin YAML Attributes (optional)
> 
>     4. Plugin Attributes
> 
>     5. Enable Plugin
> 
>     6. Disable Plugin
> 
> 
> This structure's flow is the same as real use cases, but if you move Step 2 and Step 3 to another place, it breaks the use flow IMO. cc @navendu-pottekkat @spacewander @membphis

The template looks good to me.

tzssangglass:
> may I know currently if it supports Logs and Metrics

not support yet

PR #7453, opened 2022-07-13
feat: supports dynamic control of whether plugins are running

### LINKED ISSUE #7144: help request: How to add filter conditions for plugins' usage
### Description

Hello I am new user for apisix.
I want to use `jwt-auth` and `proxy-rewrite` plugins for my service.

Suppose it is an user service and it supports several apis like `/register`, `/login`, `/rename` and so on.
Routes like `/register` and `login` will not use `jwt-auth` plugin but others need. So I add a script like this:
```shell
#!/bin/bash
curl http://127.0.0.1:9080/apisix/admin/routes/route1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "methods": ["GET", "POST", "PUT", "DELETE"],
    "uris": [
        "/api/rename*",
        "/api/list*",
        "/api/etc*"
    ],
    "plugins": {
        "jwt-auth": {},
        "proxy-rewrite": {
              "/regex_uri": ["/api/", "/svc/"]
        }
    },
    "upstream" : {
        "type": "roundrobin",
        "nodes": {
            "node1:12345": 1,
            "node2:12345": 1
        }
    }
}'

curl http://127.0.0.1:9080/apisix/admin/routes/route2 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "methods": ["GET", "POST", "PUT", "DELETE"],
    "uris": [
        "/api/register*",
        "/api/login*"
    ],
    "plugins": {
        "proxy-rewrite": {
              "/regex_uri": ["/api/", "/svc/"]
        }
    },
   "upstream" : {
        "type": "roundrobin",
        "nodes": {
            "node1:12345": 1,
            "node2:12345": 1
        }
    }
}'
```

The difference between `route1` and `route2` is only whether using `jwt-auth` plugin. 
I want to make my start-up script more concise and decrease apisix routes setting, so is there a better way to filter plugins' usage by specific condition? (such as url regular expression)

Thanks in advance.

### Environment

- APISIX version (run `apisix version`): 2.13.0
- Operating system (run `uname -a`): Linux 0c06ebd3850b 5.10.76-linuxkit # 1 SMP Mon Nov 8 10:21:19 UTC 2021 x86_64 Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): openresty/1.19.3.2

soulbird:
### Description

Fixes #7144

### Checklist


t/plugin/plugin.t
```diff
@@ -284,7 +284,7 @@ GET /hello
                     error_response = "OK"
                 }},
             }) do
-                local code, body = t('/apisix/admin/global_rules/1',
+                local code, body = t('/apisix/admin/plugin_configs/1',
```
tzssangglass, line 287: why do we change `global_rules` to `plugin_configs` here?

soulbird: > why do we change `global_rules` to `plugin_configs` here?

The global jwt plugin affects later test cases. Here is just a test schema, it is also possible to use ·`plugin_configs`

docs/en/latest/terminology/plugin.md
```diff
@@ -90,12 +90,29 @@ Some common configurations can be applied to the plugin configuration. For examp
 
 the configuration above means customizing the error response from the jwt-auth plugin to '{"message": "Missing credential in request"}'.
 
+```
+{
+    "jwt-auth": {
+        "_meta": {
+            "filter": {
```
spacewander, line 97: Why don't we put the expression under filter directly?

apisix/plugin.lua
```diff
@@ -371,6 +375,28 @@ local function trace_plugins_info_for_debug(ctx, plugins)
...
 
+local function meta_filter(plugin_name, plugin_conf)
+    local filter_vars = plugin_conf._meta and
+                plugin_conf._meta.filter and plugin_conf._meta.filter.vars
+    if not filter_vars then
+        return true
+    end
+
+    local ex, ok, err
+    ex, err = expr_lrucache(plugin_name, plugin_conf, expr.new, filter_vars)
```
spacewander, line 386: We can't use plugin_name as a key unless there is only one filter for a plugin.

t/plugin/plugin.t
```diff
@@ -306,3 +306,139 @@ GET /hello
...
+passed
+
+
+
+=== TEST 13: hit route: run proxy-rewrite plugin
+--- request
+GET /hello?version=v2
+--- response_headers
+x-api-version: v2
+--- no_error_log
```
spacewander, line 433: It is already set: https://github.com/apache/apisix/blob/ea6b55db7d50133c6cf6adb32539223c45cb0d38/t/plugin/plugin.t#L22

apisix/plugin.lua
soulbird: What I plan is only one filter for a plugin.Is this reasonable？

spacewander: However, we need one filter for each plugin configuration (route/x -> jwt-auth), not a particular plugin like `jwt-auth`

apisix/schema_def.lua
```diff
@@ -956,6 +956,11 @@ _M.plugin_injected_schema = {
                 description = "priority of plugins by customized order",
                 type = "integer",
             },
+            filter = {
+                description = "filter determines whether the plugin "..
+                                "needs to be executed at runtime",
+                type  = "array",
+            }
```
tzssangglass, line 963:
```suggestion
            filter = {
                description = "filter determines whether the plugin "..
                                "needs to be executed at runtime",
                type  = "array",
                maxItems = 20,
                items = {
                    type = "array",
                },
            }
```

apisix/plugin.lua
```diff
@@ -375,14 +375,15 @@ local function trace_plugins_info_for_debug(ctx, plugins)
     end
 end
 
-local function meta_filter(plugin_name, plugin_conf)
+local function meta_filter(route, plugin_name, plugin_conf)
```
spacewander, line 378: The `route` actually can be `route` or `global_rule`?

apisix/schema_def.lua
soulbird: What is the basis for this (`maxItems = 20`) restriction?

apisix/plugin.lua
soulbird: How about using `api_ctx.conf_id` and `plugin_name` as union key.

apisix/schema_def.lua
tzssangglass: It is just a convention within APISIX to create expr, which in general does not require many child elements.

apisix/plugin.lua
```diff
@@ -371,6 +375,33 @@ local function trace_plugins_info_for_debug(ctx, plugins)
...
+    end
+    if not ex then
+        core.log.warn("failed to get the 'vars' expression: ",
+                 err or "", " plugin_name: ", plugin_name)
+        return true
+    end
+    ok, err = ex:eval()
+    if err then
+        core.log.warn("failed to run the 'vars' expression: ",
+                 err or "", " plugin_name: ", plugin_name)
```
spacewander, line 400:
```suggestion
                 err, " plugin_name: ", plugin_name)
```

apisix/plugin.lua
```diff
@@ -371,6 +375,33 @@ local function trace_plugins_info_for_debug(ctx, plugins)
...
 
+local function meta_filter(ctx, plugin_name, plugin_conf)
+    local filter = plugin_conf._meta and plugin_conf._meta.filter
+    if not filter then
+        return true
+    end
+
+    local ex, ok, err
+    if ctx then
+        local key = core.table.concat({plugin_name, ctx.conf_type,
```
spacewander, line 386: Use `..` could be faster, without creating new table

apisix/schema_def.lua
soulbird: I think this can be omitted first, it has no effect on apisix, and users can choose by themselves. What do you think?

spacewander (approved):

apisix/schema_def.lua
tzssangglass: From a prudential point of view, I would like to add, in order to gradually unify.

spacewander (approved):

tzssangglass (approved):

PR #7514, opened 2022-07-21
feat: add plugin_metadata into control API

kingluo:
### Description

Add plugin_metadata into control API, so that you could retrieve all plugin metadatas or specific plugin metadata from control API.

### Checklist

@spacewander


apisix/control/v1.lua
```diff
@@ -276,6 +277,28 @@ function _M.dump_service_info()
     return 200, info
 end
 
+function _M.dump_all_plugin_metadata()
+    local names = plugins.get_plugins_list()
```
spacewander, line 281: Better to use another way to do it. Maybe a new method in the `plugin` module or some other way.
We can't let the module in control-api require the module in admin-api. They are at the same level.

docs/en/latest/control-api.md
```diff
@@ -451,3 +451,42 @@ Dumps the Upstream with the specified `upstream_id`:
    "modifiedIndex":1225
 }
 ```
+
+
+### GET /v1/plugin_metadatas
+
+Introduced in [v2.15](https://github.com/apache/apisix/releases/tag/2.15).
```
spacewander, line 458: Should be `2.15.0`. We have a patch version for the latest releases.

t/control/plugin-metadata.t
```diff
@@ -0,0 +1,124 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+repeat_each(1);
+no_long_string();
+no_root_location();
+no_shuffle();
+log_level("info");
```
spacewander, line 23: We don't need to repeat the settings in https://github.com/apache/apisix/blob/44d20c391767d3d42c5111ea7ef6318e84633996/t/APISIX.pm#L23-L28

t/control/plugin-metadata.t
```diff
@@ -0,0 +1,124 @@
...
+            local _, body = t('/apisix/admin/plugin_metadata/file-logger',
+                ngx.HTTP_PUT,
+                [[
+                {"log_format": {"upstream_response_time": "$upstream_response_time"}}
+                ]]
+                )
+            ngx.say(body)
+        }
+    }
+--- request
```
spacewander, line 62: We can set the repeated sections in `add_block_preprocessor` like the other test files:
https://github.com/apache/apisix/blob/44d20c391767d3d42c5111ea7ef6318e84633996/t/deployment/conf_server.t#L28

t/control/plugin-metadata.t
```diff
@@ -0,0 +1,118 @@
...
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local _, body = t('/apisix/admin/plugin_metadata/example-plugin',
+                ngx.HTTP_PUT,
+                [[{
+                    "skey": "val",
+                    "ikey": 1
+                }]],
+                [[{
```
spacewander, line 34: We don't need to check the response of admin-api in control-api's test? This response is already checked in admin-api's test.

t/control/plugin-metadata.t
spacewander: @kingluo Please update the test

t/control/plugin-metadata.t
kingluo: > We don't need to check the response of admin-api in control-api's test? This response is already checked in admin-api's test.

So just check the status code?

spacewander: Yes

t/control/plugin-metadata.t
```diff
@@ -0,0 +1,121 @@
...
+                    "node": {
+                        "value": {
+                            "skey": "val",
+                            "ikey": 1
+                        }
+                    },
+                    "action": "set"
+                }]]
+                )
+            if code ~= 200 and code ~= 201 then
```
spacewander, line 44: We can use this way like the other tests?
https://github.com/apache/apisix/blob/44d20c391767d3d42c5111ea7ef6318e84633996/t/plugin/csrf.t#L83

spacewander (approved):

tzssangglass (approved):

PR #7529, opened 2022-07-24
feat(grafana json): update grafana json for nginx shared dict in prometheus

ccxhwmy:
update grafana json for nginx shared dict in prometheus

### Description
update grafana json to add nginx shared dict, which default config is `internal-status|upstream-healthcheck|worker-events|lrucache-lock|balancer-ewma|balancer-ewma-locks|balancer-ewma-last-touched-at|etcd-cluster-health-check`.

Fixes # (issue)

### Checklist

### DIFF
--- a/docs/assets/other/json/apisix-grafana-dashboard.json
+++ b/docs/assets/other/json/apisix-grafana-dashboard.json
@@ -1703,9 +1703,9 @@
       },
       "yaxes": [
         {
-          "$$hashKey": "object:1148",
+          "$$hashKey": "object:420",
           "decimals": null,
-          "format": "bytes",
+          "format": "decbytes",
           "label": "",
           "logBase": 1,
           "max": null,
@@ -1713,7 +1713,7 @@
           "show": true
         },
         {
-          "$$hashKey": "object:1149",
+          "$$hashKey": "object:421",
           "decimals": null,
           "format": "Misc",
           "label": "",
@@ -1808,9 +1808,9 @@
       },
       "yaxes": [
         {
-          "$$hashKey": "object:1209",
+          "$$hashKey": "object:481",
           "decimals": null,
-          "format": "bytes",
+          "format": "decbytes",
           "label": "",
           "logBase": 1,
           "max": null,
@@ -1818,7 +1818,7 @@
           "show": true
         },
         {
-          "$$hashKey": "object:1210",
+          "$$hashKey": "object:482",
           "decimals": null,
           "format": "Misc",
           "label": "",
--- a/docs/assets/other/json/apisix-grafana-dashboard.json
+++ b/docs/assets/other/json/apisix-grafana-dashboard.json
@@ -1622,6 +1622,212 @@
       "timeShift": null,
       "title": "Nginx metric errors",
       "type": "stat"
+    },
+    {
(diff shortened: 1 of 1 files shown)


soulbird:
image
We can't see the change of `apisix_shared_dict_free_space_bytes` through the ordinate, can you adjust the unit? That would be more intuitive. Please note that we cannot determine the size of the shared memory in the actual environment, it may be 1MB or 100MB, so pay attention to the generality of the unit.

docs/assets/other/json/apisix-grafana-dashboard.json
```diff
@@ -1622,6 +1622,216 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "sum(apisix_shared_dict_free_space_bytes{name=~\"internal-status|upstream-healthcheck|worker-events|lrucache-lock|balancer-ewma|balancer-ewma-locks|balancer-ewma-last-touched-at|etcd-cluster-health-check\", instance=~\"$instance\"}) by (name)",
```
spacewander, line 1783: We don't need to hardcode the name of each shdict, right?

BTW, we should provide an available percent (free / capacity) for each shdict than the sum. The sum is not helpful.

ccxhwmy: > We don't need to hardcode the name of each shdict, right?
> 
> BTW, we should provide an available percent (free / capacity) for each shdict than the sum. The sum is not helpful.

I tryed to do it but failed, would you give me some suggestion(like use which function in grafana) in caculate the percent (free/capacity) for each shdict, thank you very much ! : )

spacewander: You can learn about the syntax of PromQL and try for it. AFAIK, a division between free and capacity is enough.

docs/assets/other/json/apisix-grafana-dashboard.json
```diff
@@ -1622,6 +1622,111 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "(apisix_shared_dict_free_space_bytes * 100) / on (name) apisix_shared_dict_capacity_bytes",
```
spacewander, line 1678: If you are worried about showing instances will make the graph too complex, maybe we can add a comment that people can modify this query to show per-instance graph?

spacewander (approved):
Anyway, LGTM

soulbird (approved):

docs/assets/other/json/apisix-grafana-dashboard.json
ccxhwmy: > If you are worried about showing instances will make the graph too complex, maybe we can add a comment that people can modify this query to show per-instance graph?

Does you mean add a comment like `# we can modify the expr to show per-instance graph` in json file?

PR #7549, opened 2022-07-27
feat(prometheus): extra labels from nginx var for http request metrics

### LINKED ISSUE #4273: discuss: enrich the Available metrics of plugin prometheus
### Issue description

About the plugin prometheus, in the original metric data of APISIX, the number of the available metrics is limited.

In some practical application scenarios, request_method and remote_addr are very useful monitoring metrics.

Is there a plan to enrich the [Available metrics](https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/prometheus.md#available-metrics)，such as adding remote_addr, request_method in Status codes?
pe

### Environment

apisix version (cmd: apisix version): 2.5

kingluo:
close #4273

### Description

Add extra labels from nginx variables for http request metrics.

Fixes #4273

### Checklist

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -78,13 +78,13 @@ local function extra_labels(name, ctx)
     if metrics and metrics[name] and metrics[name].extra_labels then
         local labels = metrics[name].extra_labels
         for _, kv in ipairs(labels) do
-			local val, v = next(kv)
-			if ctx then
-				val = ctx.var[v:sub(2)]
-				if val == nil then
-					val = ""
-				end
-			end
+            local val, v = next(kv)
+            if ctx then
+                val = ctx.var[v:sub(2)]
+                if val == nil then
+                    val = ""
+                end
+            end
             core.table.insert(extra_labels_tbl, val)
         end
     end
@@ -166,7 +166,7 @@ function _M.http_init(prometheus_enabled_in_stream)
     metrics.status = prometheus:counter("http_status",
             "HTTP status codes per service in APISIX",
             {"code", "route", "matched_uri", "matched_host", "service", "consumer", "node",
-			unpack(extra_labels("http_status"))})
+            unpack(extra_labels("http_status"))})
 
     metrics.latency = prometheus:histogram("http_latency",
         "HTTP request latency in milliseconds per service in APISIX",
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -40,6 +40,7 @@ local service_fetch = require("apisix.http.service").get
 local latency_details = require("apisix.utils.log-util").latency_details_in_ms
 local xrpc = require("apisix.stream.xrpc")
 local unpack = unpack
+local next = next
 
 
 local ngx_capture
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -73,10 +73,10 @@ local function ngx_var_label_values(ctx, name)
     clear_tab(ngx_var_label_values_tbl)
 
     local attr = plugin.plugin_attr("prometheus")
-    local ngx_var_labels = attr.ngx_var_labels
+    local custom_labels = attr.custom_labels
 
-    if ngx_var_labels and ngx_var_labels[name] then
-        local labels = ngx_var_labels[name]
+    if custom_labels and custom_labels[name] then
+        local labels = custom_labels[name]
         for _, name in ipairs(labels) do
             local val = ctx.var[name]
             if val == nil then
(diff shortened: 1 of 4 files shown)


conf/config-default.yaml
```diff
@@ -522,6 +522,14 @@ plugin_attr:
     export_addr:
       ip: 127.0.0.1
       port: 9091
+    # ngx_var_labels:
```
tokers, line 525: I think we don't need to highlight the `var`. Instead, we can just use the name `custom_labels`.

conf/config-default.yaml
```diff
@@ -522,6 +522,14 @@ plugin_attr:
     export_addr:
       ip: 127.0.0.1
       port: 9091
+    # custom_labels:
```
spacewander, line 525: Would be better to use:
```
metrics:
    xxx:
       extra_labels:
           label_name_xxx: $label_var
     yyy:
```

So we can add more options to a metric in the future. And we don't need to require the label name should be the variable name.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -65,6 +67,28 @@ local function gen_arr(...)
...
+
+    local attr = plugin.plugin_attr("prometheus")
+    local custom_labels = attr.custom_labels
+
+    if custom_labels and custom_labels[name] then
+        local labels = custom_labels[name]
+        for _, name in ipairs(labels) do
+            local val = ctx.var[name]
+            if val == nil then
+                val = "nil"
```
spacewander, line 83: I would suggest using `""`, which is language-independent.

conf/config-default.yaml
spacewander: And please add a comment to show that the commented configurations are the example but not the default value.

t/plugin/prometheus4.t
```diff
@@ -0,0 +1,144 @@
...
+                    "uri": "/apisix/prometheus/metrics"
+                }]]
+                )
+            if code >= 300 then
+                ngx.status = code
+                return
+            end
+        }
+    }
+--- error_code: 200
```
spacewander, line 66: Actually we don't need to specify `--- error_code: 200` as it is checked by default

t/plugin/prometheus4.t
```diff
@@ -0,0 +1,144 @@
...
+
+
+=== TEST 5: client request, label with nonexist ngx variable
+--- yaml_config
+plugin_attr:
+    prometheus:
+        custom_labels:
+            http_status:
+                - dummy
+            bandwidth:
```
spacewander, line 131: We don't need the bandwidth?

conf/config-default.yaml
kingluo: > 

The labels of extra_labels should be a list, because fields of dictionary in lua table are out of order, which would cause the metric full name may change each time it reloads the conf.

spacewander: We can provide order on the implementation side, for example sorting them before using them.

docs/en/latest/plugins/prometheus.md
```diff
@@ -53,6 +53,29 @@ plugin_attr:
     export_uri: /apisix/metrics
 ```
 
+### Specifying `metrics`
+
+For http request related metrics, you could specify extra labels, which match the nginx variables.
+
+If you specify label for nonexist nginx variable, the label value would be "nil".
```
tzssangglass, line 60:
```suggestion
If you specify label for nonexist nginx variable, the label value would be "".
```

tzssangglass (dismissed):

t/plugin/prometheus4.t
```diff
@@ -0,0 +1,142 @@
...
+plugin_attr:
+    prometheus:
+        metrics:
+            bandwidth:
+                extra_labels:
+                    - upstream_addr: $upstream_addr
+                    - upstream_status: $upstream_status
+--- request
+GET /hello
+--- error_code: 200
```
spacewander, line 112: Actually we don't need to specify `--- error_code: 200` as it is checked by default

t/plugin/prometheus4.t
```diff
@@ -0,0 +1,142 @@
...
+            http_status:
+                extra_labels:
+                    - dummy: $dummy
+--- request
+GET /hello
+--- error_code: 200
+
+
+
+=== TEST 6: fetch the prometheus metric data, with nil label
```
spacewander, line 138: It is no longer a nil label?

docs/en/latest/plugins/prometheus.md
```diff
@@ -53,6 +53,29 @@ plugin_attr:
     export_uri: /apisix/metrics
 ```
 
+### Specifying `metrics`
+
+For http request related metrics, you could specify extra labels, which match the nginx variables.
```
spacewander, line 58: Let's use `APISIX variable` like elsewhere in the doc.

spacewander (approved):

soulbird (approved):

PR #7554, opened 2022-07-27
feat(key-auth): support get key from variable combination

### LINKED ISSUE #7530: feat: As a user, I want to authenticate clients by client_id and client_secret headers
### Description

We would like to authenticate clients by client_id and client_secret headers. But one key-auth plugin offers just one header to use. I tried to connect 2 key-auth plugins in plugin orchestration section but that didn't work

soulbird:
### Description

Fixes #7530 

### Checklist


t/plugin/key-auth.t
```diff
@@ -585,3 +585,84 @@ GET /hello?auth=auth-one
...
+--- request
+GET /t
+--- response_body
+passed
+--- no_error_log
+[error]
+
+
+
+=== TEST 28: verify auth get key  from var_combination
```
tzssangglass, line 661:
```suggestion
=== TEST 28: verify auth get key from var_combination
```

docs/en/latest/plugins/key-auth.md
```diff
@@ -45,8 +45,9 @@ For Route:
...
-| header | string | optional    | apikey  |       | The header to get the key from.                                                                                                                                                                                                                                               |
-| query  | string | optional    | apikey  |       | The query string to get the key from. Lower priority than header.                                                                                                                                                                                                             |
+| var_combination | string | optional |      |    | Set which variable combination we get the key from. For example, "$remote_addr $consumer_name" will be combined into a key after parsing. highest priority.                  |
```
moonming, line 48: no test cases for `priority`

t/plugin/key-auth.t
```diff
@@ -585,3 +585,84 @@ GET /hello?auth=auth-one
...
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "plugins": {
+                        "key-auth": {
+                            "var_combination": "${http_client_id}:${http_client_secret}"
```
moonming, line 633: The example in the document is "$remote_addr $consumer_name", and there is an extra colon in the test case. So which one is right?

docs/en/latest/plugins/key-auth.md
```diff
@@ -45,8 +45,9 @@ For Route:
...
+| header | string | optional    | apikey  |       | The header to get the key from. Lower priority than `var_combination`.                                                                                                                                                                                                                                              |
```
moonming, line 49: need to add example in doc.

moonming (changes requested):

apisix/plugins/key-auth.lua
```diff
@@ -90,12 +93,28 @@ end
...
+        local err, n_resolved
+        key, err, n_resolved = core.utils.resolve_var(conf.var_combination, ctx.var)
+        if err then
+            core.log.error("could not resolve vars in ",
+                                conf.var_combination, " error: ", err)
+        end
+
+        if n_resolved == 0 then
+            key = nil
+        end
```
moonming, line 108:
```suggestion
        if err then
            core.log.error("could not resolve vars in ",
                                conf.var_combination, " error: ", err)

        elseif n_resolved == 0 then
            key = nil
        end
```

t/plugin/key-auth.t
soulbird: The documentation is just an example of what can be configured, it is not the same as the test case, and the documentation and the test case are not the same for the users. There is no right or wrong here. But I can unify them.

moonming: This confuses me, how should I configure it?

soulbird: OK. I'll add some examples and test cases later.

soulbird:
This function is not needed temporarily, reopen it when necessary.

PR #7584, opened 2022-08-01
fix(redirect):  compatible with old version data

### LINKED ISSUE #7583: bug: Incompatible redirect plugin parameters after upgrade to 2.15
### Current Behavior

redirect plugin parameters incompatible

### Expected Behavior

_No response_

### Error Logs

failed to check the configuration of plugin redirect err: value should match only one schema, but matches none

### Steps to Reproduce

1、create route in version 2.14
2、upgrade version to 2.15
3、route init error

Here is the bug:
The old version will have an append_ query_ string parameter, which is false by default. After upgrading, http_ to_ https and append_ query_ string parameters cannot coexist, even if The value of append_ query_string is false, which means that the old version created the plugin and append_ query_ string is false, the parameter verification also fails. This compatibility is not very reasonable.

### Environment

- APISIX version (run `apisix version`): 2.15

tzssangglass:
### Description

Fixes #7583

### Checklist

### DIFF
--- a/t/plugin/redirect2.t
+++ b/t/plugin/redirect2.t
@@ -79,3 +79,32 @@ GET /test/hello?o=apache
 --- response_headers
 Location: http://test.com/hello?q=apisix&o=apache
 --- error_code: 302
+
+
+
+=== TEST 3: compatible with old version configuration
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/hello",
+                    "host": "foo.com",
+                    "plugins": {
+                        "redirect": {
+                            "http_to_https": true,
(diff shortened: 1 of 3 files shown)


t/plugin/redirect2.t
```diff
@@ -79,3 +79,32 @@ GET /test/hello?o=apache
 --- response_headers
 Location: http://test.com/hello?q=apisix&o=apache
 --- error_code: 302
+
+
+
+=== TEST 3: compatible with old version configuration
```
tokers, line 85: Also test the case that `http_to_https` is `true` and `append_query_string` is also `true`.

tzssangglass: Already have, ref:https://github.com/apache/apisix/pull/7584/files#diff-97c7ac2f91e3c7a47c9832ebc576be52e9d69af74dde165200128e0f056531b6R1120-R1149

spacewander (approved):

soulbird (approved):

PR #7590, opened 2022-08-02
feat(ldap-auth): use lua-resty-ldap instead of lualdap

kingluo:
### Description

Use lua-resty-ldap to do nonblocking ldap auth.

### Checklist


apisix/plugins/ldap-auth.lua
```diff
@@ -30,11 +30,13 @@ local schema = {
     title = "work with route or service object",
     properties = {
         base_dn = { type = "string" },
-        ldap_uri = { type = "string" },
+        ldap_host = { type = "string" },
```
spacewander, line 33: We should keep the compatibility of the configuration

apisix/plugins/ldap-auth.lua
```diff
@@ -139,8 +141,20 @@ function _M.rewrite(conf, ctx)
...
+    local ldapconf = {
+        timeout = 10000,
+        start_tls = false,
+        ldap_host = conf.ldap_host,
+        ldap_port = conf.ldap_port,
+        ldaps = conf.use_tls,
+        verify_ldap_host = conf.verify_ldap_host,
+        base_dn = conf.base_dn,
+        attribute = uid,
+        keepalive = 60000,
```
spacewander, line 153: Let's doc the newly added timeout and keepalive

apisix/plugins/ldap-auth.lua
```diff
@@ -139,8 +141,20 @@ function _M.rewrite(conf, ctx)
...
+    }
+    local res, err = ldap.ldap_authenticate(user.username, user.password, ldapconf)
+    if not res then
+        core.log.error(err)
```
spacewander, line 157: We use warning level log for this situation:
https://github.com/apache/apisix/blob/5cbcf246021b9f5a97ce98e999a7ed65a1cbc117/apisix/plugins/jwt-auth.lua#L369
and it would be better to add a prefix in the error message

apisix/plugins/ldap-auth.lua
kingluo: These items are not exported to admin api, we only use default values. So we need to doc the default values? But it's kind of internal parameters.

apisix/plugins/ldap-auth.lua
```diff
@@ -30,13 +30,12 @@ local schema = {
     title = "work with route or service object",
     properties = {
         base_dn = { type = "string" },
-        ldap_host = { type = "string" },
-        ldap_port = { type = "number" },
+        ldap_uri = { type = "string" },
```
tzssangglass, line 33: Deleting these two options seems incompatible, can't we keep them?

kingluo: What do you mean? Those are just new config items which make it incompatible, so I revert to use the old one.

tzssangglass: I thought `ldap_host` and `ldap_port` became `ldap_uri` now, and I was wondering if they were compatible. For example, there is now an old copy of the data in etcd and they contain `ldap_host` and `ldap_port`. when updating the APISIX version, the user needs to reset the `ldap_uri`?

kingluo: `ldap_uri` is the original configuration item. Nothing change here.

tzssangglass: got it.

tzssangglass (approved):

apisix/plugins/ldap-auth.lua
```diff
@@ -138,9 +139,23 @@ function _M.rewrite(conf, ctx)
...
+        ldap_port = ldap_port or 389,
+        ldaps = conf.use_tls,
+        verify_ldap_host = conf.verify_ldap_host,
+        base_dn = conf.base_dn,
+        attribute = uid,
+        keepalive = 60000,
+    }
+    local res, err = ldap.ldap_authenticate(user.username, user.password, ldapconf)
+    if not res then
+        core.log.warn("ldap-auth: ", err)
```
spacewander, line 158: The prefix should be more meaningful

docs/en/latest/plugins/ldap-auth.md
```diff
@@ -49,7 +49,8 @@ For Route:
 |----------|---------|----------|---------|------------------------------------------------------------------------|
 | base_dn  | string  | True     |         | Base dn of the LDAP server. For example, `ou=users,dc=example,dc=org`. |
 | ldap_uri | string  | True     |         | URI of the LDAP server.                                                |
-| use_tls  | boolean | False    | `true`  | If set to `true` uses TLS.                                             |
+| use_tls  | boolean | False    | `false` | If set to `true` uses TLS.                                             |
+| verify_ldap_host| boolean  | False     | `false`        | Whether to verify the server certificate when `use_tls` is enabled; If set to `true`, you must set `ssl_trusted_certificate` in `config.yaml`, and make sure the host of `ldap_uri` matches the host in server certificate. |
```
spacewander, line 53: This field doesn't have default value in the code. Why add a default value in the doc?

spacewander, line 53: According to the description, we should name this field `tls_verify`? It doesn't verify the host but the TLS relative stuff.

spacewander (approved):

tzssangglass (approved):

PR #7594, opened 2022-08-02
fix(balancer): quote ipv6 address

### LINKED ISSUE #7563: bug: APISIX 2.10.5 (LTS) can't support IPv6 Upstream host IP
### Current Behavior

etcd data:
```
{
    "scheme":"http",
    "name":"safebox-esp-web_web-console-cns2waf_5009",
    "desc":"Created by apisix-ingress-controller, DO NOT modify it manually",
    "id":"a48bc649",
    "update_time":1659003639,
    "create_time":1659003457,
    "nodes":[
        {
            "weight":100,
            "priority":0,
            "host":"2409:XXXX:XXXX:2000:0:1:afd:7d20",
            "port":4009
        },
        {
            "weight":100,
            "priority":0,
            "host":"2409:XXXX:XXXX:2000:0:1:afd:7d68",
            "port":4009
        }
    ],
    "type":"roundrobin",
    "labels":{
        "managed-by":"apisix-ingress-controller"
    },
    "hash_on":"vars",
    "pass_host":"pass"
}
```

### Expected Behavior

_No response_

### Error Logs

_No response_

### Steps to Reproduce

I had modify **/usr/local/apisix/apisix/schema_def.lua  LINE 40**
```
-- local host_def_pat = "^\\*?[0-9a-zA-Z-._]+$"
local host_def_pat = "^\\*?[0-9a-zA-Z-._\\[\\]:]+$"
```

AND   **/usr/local/apisix/apisix/balancer.lua LINE 294**
```
            local has_ipv6 = string.find(server.host, ":", 1, true)
            if (has_ipv6 ~= nil) then
                local is_ipv6_plus = string.find(server.host, "[", 1, true)
                if (is_ipv6_plus == nil) then
                    server.host = "["..server.host.."]"
                end
            end

            local ok, err = balancer.set_current_peer(server.host, server.port, pool_opt)
```

When upstream hosts count =1，its working;
But upstream hosts  number >=2 ，error log:
```
2022/07/28 10:07:09 [error] 51#51: *13796 [lua] balancer.lua:370: run(): failed to set server peer [[2409:XXXX:XXXX:2000:0:1:afd:7d65:4009]:nil] err: invalid IPv6 address while connecting to upstream, client: 10.253.91.92, server: _, request: "GET /web-console-cnswaf/ HTTP/1.1", host: "xxx.xxx.com"
```
server.host = 2409:XXXX:XXXX:2000:0:1:afd:7d65:4009  (4009 is port)
server.port = nil 

### Environment

- APISIX version: 2.10.5uanme
- Operating system: CentOS 7 , Linux 4.19
- OpenResty version: 1.19.9.1
- etcd version, if relevant: 3.4.0

kingluo:
Fixes #7563

### Description

quote ipv6 address with square brackets before picking server.

Fixes #7563

### Checklist


docs/zh/latest/admin-api.md
```diff
@@ -532,7 +532,7 @@ APISIX 的 Upstream 除了基本的负载均衡算法选择外，还支持对上
...
-| nodes          | 必需，不能和 `service_name` 一起用 | Node           | 哈希表或数组。当它是哈希表时，内部元素的 key 是上游机器地址列表，格式为`地址 +（可选的）端口`，其中地址部分可以是 IP 也可以是域名，比如 `192.168.1.100:80`、`foo.com:80`等。value 则是节点的权重。当它是数组时，数组中每个元素都是一个哈希表，其中包含 `host`、`weight` 以及可选的 `port`、`priority`。`nodes` 可以为空，这通常用作占位符。客户端命中这样的上游会返回 502。                                        | `192.168.1.100:80`                               |
+| nodes          | 必需，不能和 `service_name` 一起用 | Node           | 哈希表或数组。当它是哈希表时，内部元素的 key 是上游机器地址列表，格式为`地址 +（可选的）端口`，其中地址部分可以是 IP 也可以是域名，比如 `192.168.1.100:80`、`foo.com:80`等。如果地址部分是 IPv6 地址，则必须用中括号将其括起来。value 则是节点的权重。当它是数组时，数组中每个元素都是一个哈希表，其中包含 `host`、`weight` 以及可选的 `port`、`priority`。`nodes` 可以为空，这通常用作占位符。客户端命中这样的上游会返回 502。                                        | `192.168.1.100:80`, `[::1]:80`                               |
```
tokers, line 535: We quote the IPv6 address in the code if it doesn't wrap by the double quotes, but why here do we ask users to add the double quotes?

kingluo: There is two types of `nodes`:
If it's an array, then user have to specify address and port seperately, we could quote the ipv6 address with square brackets if not quoted yet.
However, if it's a hash table, the key format is "address:port", then if user do not quote the ipv6 address, we could not tell the address and port from it. For example, `::1:1980` and `::1` are both valid ipv6 address, how to know whether 1980 is a port or not?

docs/en/latest/admin-api.md
```diff
@@ -524,7 +524,7 @@ In addition to the equalization algorithm selections, Upstream also supports pas
...
+| nodes                       | required, can't be used with `service_name` | IP addresses (with optional ports) of the Upstream nodes represented as a hash table or an array. If the IP address is IPv6 address, it (without port) must be quoted with square brackets. In the hash table, the key is the IP address and the value is the weight of the node for the load balancing algorithm. In the array, each item is a hash table with keys `host`, `weight`, and the optional `port` and `priority`. Empty nodes are treated as placeholders and clients trying to access this Upstream will receive a 502 response.                                                                                                                                                                                                                                                                             | `192.168.1.100:80`, `[::1]:80`                                                                                                                         |
```
spacewander, line 527: > IP addresses (with optional ports) of the Upstream nodes represented as a hash table or an array. If the IP address is IPv6 address, it (without port) must be quoted with square brackets.

The quote is not needed in the array format? Only when using in hash format, and the host contains a port, the quote is required to distinguish it between `::1:1980` and `[::1]:1980`.

kingluo: @spacewander yes.

docs/zh/latest/admin-api.md
spacewander: `::1` is a valid format as we claim the port is optional. So it's the user's choice to quote it or not. Would be better to use `suggest` instead of `must` in the doc.

spacewander: Or we can use `must` when the user needs to use a port, like `如果地址部分是 IPv6 地址加端口，则必须用中括号将 IPv6 地址括起来。`

tzssangglass (dismissed):

docs/en/latest/admin-api.md
```diff
@@ -524,7 +524,7 @@ In addition to the equalization algorithm selections, Upstream also supports pas
...
+| nodes                       | required, can't be used with `service_name` | IP addresses (with optional ports) of the Upstream nodes represented as a hash table or an array. For hash table case, if the key is IPv6 address with port, then the IPv6 address must be quoted with square brackets. In the hash table, the key is the IP address and the value is the weight of the node for the load balancing algorithm. In the array, each item is a hash table with keys `host`, `weight`, and the optional `port` and `priority`. Empty nodes are treated as placeholders and clients trying to access this Upstream will receive a 502 response.                                                                                                                                                                                                                                                                             | `192.168.1.100:80`, `[::1]:80`                                                                                                                         |
```
spacewander, line 527: > For hash table case, if the key is IPv6 address with port, then the IPv6 address must be quoted with square brackets.

Could we move this sentence back so the English version is similar with the Chinese version?

spacewander (approved):

tzssangglass (approved):

PR #7613, opened 2022-08-05
feat: add version header for Admin API

bzp2010:
### Description

We designed a new Admin API data format for APISIX v3 that is not compatible with what was once v2. We now lack the means to distinguish which version of the Admin API is currently in use, so I will add a default response header for Admin APIia that contains its version number information.

No issue was created.

### Checklist

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -18,6 +18,7 @@ local require = require
 local core = require("apisix.core")
 local route = require("apisix.utils.router")
 local plugin = require("apisix.plugin")
+local v3_adapter = require("apisix.admin.v3_adapter")
 local ngx = ngx
 local get_method = ngx.req.get_method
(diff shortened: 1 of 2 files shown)


apisix/admin/init.lua
```diff
@@ -188,6 +189,11 @@ local function run()
     local code, data = resource[method](seg_id, req_body, seg_sub_path,
                                         uri_args)
     if code then
+        if v3_adapter.enable_v3() then
+            core.response.set_header("X-API-VERSION", "3")
```
soulbird, line 193: `v2` or `v3` ?

bzp2010: maybe better, let me change

bzp2010: done

soulbird (approved):

spacewander (approved):

PR #7648, opened 2022-08-11
fix: use a new way to manage clean_handler

### LINKED ISSUE #7633: bug: config_util.lua:71: attempt to call local 'f' (a nil value)
### Current Behavior

The service is configured with a health check. When the upstream node changes and the concurrent request is bright, there will be a large number of 500 errors（lua entry thread aborted because function 'f' is nil）

### Expected Behavior

The function 'f' should not be nil, the request should be completed normally

### Error Logs

The error log is found as follows:
2022/08/08 10:29:46 [error] 46#46: *144287093 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/core/config_util.lua:71: attempt to call local 'f' (a nil value)
stack traceback:
coroutine 0:
        /usr/local/apisix/apisix/core/config_util.lua: in function 'cancel_clean_handler'
        /usr/local/apisix/apisix/upstream.lua:119: in function 'fetch_healthchecker'
        /usr/local/apisix/apisix/upstream.lua:327: in function 'set_upstream'
        /usr/local/apisix/apisix/init.lua:515: in function 'http_access_phase'
        access_by_lua(nginx.conf:253):2: in main chunk, client: 172.31.xx.xx, server: _, request: "GET xxxxxxx

According to the error message, find the location of the error code:
file path: apisix/core/config_util.lua

function _M.cancel_clean_handler(item, idx, fire)
    local f = item.clean_handlers[idx]
    core_tab.remove(item.clean_handlers, idx)
    if fire then
        f(item)
    end
end

### Steps to Reproduce

1. Install apisix to EKS using the helm package
2. Service discovery using kubernetes
3. Routed services enable health checks
4. Initiate concurrent requests, no such error message appears
5. Adjust (increase or decrease) upstream nodes, and then initiate concurrent requests, a large number of such error logs appear, and the requester receives a large number of 500 errors

### Environment

- APISIX version (run `apisix version`):2.14.1
- Operating system (run `uname -a`): Linux apisix-58d87884d-l4qkw 5.4.129-63.229.amzn2.x86_64 #1 SMP Tue Jul 20 21:22:08 UTC 2021 x86_64 Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.19.9.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):3.4.0
- APISIX Dashboard version, if relevant:2.13.0
- LuaRocks version, for installation issues (run `luarocks --version`): 3.8.0

spacewander:

### Description

Fixes #7633

### Checklist

### DIFF
--- a/apisix/core/config_util.lua
+++ b/apisix/core/config_util.lua
@@ -20,8 +20,10 @@
 -- @module core.config_util
 
 local core_tab = require("apisix.core.table")
+local log = require("apisix.core.log")
 local str_byte = string.byte
 local str_char = string.char
+local ipairs = ipairs
 local setmetatable = setmetatable
 local tostring = tostring
 local type = type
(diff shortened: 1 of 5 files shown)


apisix/core/config_util.lua
```diff
@@ -56,23 +58,56 @@ end
...
 function _M.add_clean_handler(item, func)
-    local idx = #item.clean_handlers + 1
-    item.clean_handlers[idx] = func
-    return idx
+    if not item.clean_handlers._id then
+        item.clean_handlers._id = 1
+    end
+
+    local id = item.clean_handlers._id
+    item.clean_handlers._id = item.clean_handlers._id + 1
```
starsz, line 66: If `item.clean_handlers._id` is nil. The result will be two because we set `item.clean_handlers._id = 1` at line 62.

tzssangglass: What is the `clean_handlers._id` used for?

spacewander: > What is the `clean_handlers._id` used for?

It is used to provide id for the handlers.

spacewander: > If `item.clean_handlers._id` is nil. The result will be two because we set `item.clean_handlers._id = 1` at line 62.

Yes. But we store the id (which is assigned before `item.clean_handlers._id = item.clean_handlers._id + 1`), so the first id is still 1.

tzssangglass (approved):

apisix/core/config_util.lua
starsz: > > If `item.clean_handlers._id` is nil. The result will be two because we set `item.clean_handlers._id = 1` at line 62.
> 
> Yes. But we store the id (which is assigned before `item.clean_handlers._id = item.clean_handlers._id + 1`), so the first id is still 1.

Got it.

starsz (approved):

zhangliweixyz:
image

Hi, as for this bug, can I change the code like this? if f not nil and call f, and I'm not sure if there is other problem

PR #7690, opened 2022-08-16
fix(openid-connect): the default 'redirect_uri' (#2426)

### LINKED ISSUE #2426: bug：enable openid-connect plugin without redirect_uri got 500 error
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

### Bug
- Which version of Apache APISIX Dashboard, OS, and Browser?

1. create a route, enable openid-connect plugin, configs:
```json
{
	"name": "api2",
	"route_group_id": "de45ec14-317f-4177-8a22-10ab95bd45f7",
	"route_group_name": "group1",
	"status": true,
	"desc": "",
	"protocols": ["http"],
	"paths": ["/get1"],
	"methods": ["GET"],
	"upstream_protocol": "https",
	"type": "roundrobin",
	"plugins": {
		"openid-connect": {
			"client_id": "kbyuFDidLLm280LIwVFiazOqjO3ty8KH",
			"client_secret": "60Op4HFM0I8ajz0WdiStAbziZ-VFQttXuxixHHs2R7r7-CW8GR79l-mmLqMhc-Sa",
			"discovery": "https://samples.auth0.com/.well-known/openid-configuration"
		}
	},
	"script": {},
	"uris": ["/get1"],
	"vars": [],
	"upstream": {
		"type": "roundrobin",
		"nodes": {
			"httpbin.org:443": 1
		},
		"timeout": {
			"connect": 6000,
			"send": 6000,
			"read": 6000
		}
	},
	"upstream_header": {},
	"upstream_path": {
		"to": "/get"
	}
}
```

2. visit this route via chrome, got 500 error

2020-10-10 14-52-43屏幕截图

3. got the error log in apisix error.log:
```shell
2020/10/10 06:47:45 [error] 26#26: *6045490 [lua] openidc.lua:1378: authenticate(): request to the redirect_uri path but there's no session state found, client: 192.168.1.145, server: , request: "GET /get HTTP/1.1", host: "192.168.1.145:9
2020/10/10 06:47:45 [error] 26#26: *6045490 [lua] openid-connect.lua:163: phase_func(): failed to authenticate in openidc: request to the redirect_uri path but there's no session state found, client: 192.168.1.145, server: , request: "GET
2020/10/10 06:47:45 [error] 25#25: *6045493 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/plugins/echo.lua:96: attempt to index local 'conf' (a userdata value)
stack traceback:                                                                                                                                            
coroutine 0:                                                                                                                                             
        /usr/local/apisix/apisix/plugins/echo.lua: in function 'phase_fun'                                                                                     
        [string "route#41ee9eef-564e-4208-86e6-fc6834bd7178"]:17: in function 'phase_func'                                                                  
        /usr/local/apisix/apisix/script.lua:52: in function 'run'                                                                                                 
        /usr/local/apisix/apisix/init.lua:508: in function 'http_access_phase'                                                                                 
        access_by_lua(nginx.conf:173):2: in main chunk, client: 192.168.1.145, server: , request: "GET /favicon.ico HTTP/1.1", host: "192.168.1.145:9080", referrer: "http://192.168.1.145:9080/get"
```
**`redirect_uri` should be a reqired param in openid-connect plugin**

liweitianux:
### Description

Previously the `redirect_uri` was set to `ngx.var.request_uri` if not
configured.  However, it caused the underlying `lua-resty-openidc`
module to raise this error:

```
request to the redirect_uri path but there's no session state found
```

because `lua-resty-openidc` would think it was the redirection response
from OP when the `redirect_uri` equals `ngx.var.request_uri`.

Although the OAuth 2.0 Security Best Current Practice [1] recommends
that the `redirect_uri` should be explicitly specified to prevent
malicious redirection attacks, it would also be handy for APISIX to
properly determine a default one if `redirect_uri` not given.

Therefore, append the `.apisix/redirect` suffix to the current request
URI to determine the default `redirect_uri`.  It makes
`lua-resty-openidc` happy and it's almost unlikely to conflict with
user's URIs.

Also note that the OP should be properly configured to accept such
auto-determined redirect URIs.

Update the documentation accordingly.

Fix #2426.

[1] https://datatracker.ietf.org/doc/draft-ietf-oauth-security-topics/

### Checklist


docs/en/latest/plugins/openid-connect.md
```diff
@@ -198,3 +198,4 @@ In this example, the Plugin can enforce that the access token, the ID token, and
 
 - `redirect_uri` needs to be captured by the route where the current APISIX is located. For example, the `uri` of the current route is `/api/v1/*`, `redirect_uri` can be filled in as `/api/v1/callback`;
 - `scheme` and `host` of `redirect_uri` (`scheme:host`) are the values required to access APISIX from the perspective of the identity provider.
+- See also [this GitHub issue](https://github.com/apache/apisix/issues/2426), especially these comments: [@starsz](https://github.com/apache/apisix/issues/2426#issuecomment-1091021687), [@david-woelfle](https://github.com/apache/apisix/issues/2426#issuecomment-1090675455), [@liweitianux (1)](https://github.com/apache/apisix/issues/2426#issuecomment-1206107085), [@liweitianux (2)](https://github.com/apache/apisix/issues/2426#issuecomment-1207423283)
```
spacewander, line 201: Could you write the conclusion in a short sentence, instead of multiple reference links?

apisix/plugins/openid-connect.lua
```diff
@@ -65,7 +65,7 @@ local schema = {
         },
         redirect_uri = {
             type = "string",
-            description = "use ngx.var.request_uri if not configured"
+            description = "auto append '.apisix/redirect' if not configured"
```
tokers, line 68: But it cannot make sure the new URI will be matched by the same route, it actually adds an implicit condition that the route match type is prefix instead of exact.

liweitianux: Well. If the route is of type **exact**, then it cannot actually use this `openid-connect` plugin, because the `redirect_uri` must be different from the request URI.

starsz (commented):
Hi @liweitianux .Can you add a test case to cover it ?

apisix/plugins/openid-connect.lua
```diff
@@ -263,7 +263,22 @@ function _M.rewrite(plugin_conf, ctx)
...
+            -- This is the redirection response from the OIDC provider.
+            conf.redirect_uri = uri
+        else
+            if string.sub(uri, -1, -1) == "/" then
+                conf.redirect_uri = string.sub(uri, 1, -2) .. suffix
+            else
+                conf.redirect_uri = uri .. suffix
+            end
+        end
+        core.log.debug("auto set redirect_uri: ", conf.redirect_uri)
```
tzssangglass, line 281: When there is no `conf.redirect_uri`, it looks like we generate `redirect_uri` for each request(except for the real redirect from IdP).

Can we inject a default value for `redirect_uri` when we add the `openid-connect` plugin.

For example:

1. we require that the match condition for a road has to be a prefix match, like `/openid/*`, then redirect_uri defaults to `/openid/.apisix/redirect`.

Listen to what others say.

apisix/plugins/openid-connect.lua
tokers: @liweitianux Could you write some hints about this in the OpenID Connect doc?

liweitianux: Sure. I'll update the doc later.

apisix/plugins/openid-connect.lua
shreemaan-abhishek: @liweitianux any thoughts about this?

liweitianux: I don't think we can set a proper default value for `redirect_uri`, because a plugin can be configured globally, per-route, and per-service.  An exception is that the `openid-connect` plugin is configured for a specific route, so we can determine a `redirect_uri` that's covered by the route.

Please correct me if I'm mistaken.

shreemaan-abhishek: I agree with you.

shreemaan-abhishek: We need a test case to show that not providing `redirect_uri` works fine.

liweitianux: I changed the job this year, and I haven't been working with APISIX for more than one year now. I'm unfamiliar with the test framework, and that was the main reason that this PR was lacking a test case.  It would be great if you can rejuvenate this PR.  Thanks.

Sn0rt (approved):

monkeyDluffy6017:
Please resolve the conflicts and make the ci pass

docs/zh/latest/plugins/openid-connect.md
```diff
@@ -208,3 +208,4 @@ curl http://127.0.0.1:9180/apisix/admin/routes/1 \
 
 - `redirect_uri` 需要能被当前 APISIX 所在路由捕获，比如当前路由的 `uri` 是 `/api/v1/*`, `redirect_uri` 可以填写为 `/api/v1/callback`；
 - `redirect_uri`（`scheme:host`）的 `scheme` 和 `host` 是身份认证服务视角下访问 APISIX 所需的值。
+- 另请参考[此 GitHub 问题](https://github.com/apache/apisix/issues/2426), 尤其是下述评论: [@starsz](https://github.com/apache/apisix/issues/2426#issuecomment-1091021687), [@david-woelfle](https://github.com/apache/apisix/issues/2426#issuecomment-1090675455), [@liweitianux (1)](https://github.com/apache/apisix/issues/2426#issuecomment-1206107085), [@liweitianux (2)](https://github.com/apache/apisix/issues/2426#issuecomment-1207423283)
```
kayx23, line 211:
```suggestion
- 另请参考[此 GitHub 问题](https://github.com/apache/apisix/issues/2426), 尤其是下述评论：[@starsz](https://github.com/apache/apisix/issues/2426#issuecomment-1091021687), [@david-woelfle](https://github.com/apache/apisix/issues/2426#issuecomment-1090675455), [@liweitianux (1)](https://github.com/apache/apisix/issues/2426#issuecomment-1206107085), [@liweitianux (2)](https://github.com/apache/apisix/issues/2426#issuecomment-1207423283)
```
fix for CI

docs/en/latest/plugins/openid-connect.md
kayx23:
```suggestion
- `redirect_uri` should not be the same as the URI of the route. This is because when a user initiates a request to visit the protected resource, the request directly hits the redirection URI with no session cookie in the request, which leads to the `no session state found` error.
```

shreemaan-abhishek:
@luoluoyuyu sure.

luoluoyuyu:
@shreemaan-abhishek
It seems I don't have permission to add commits. i may need to submit a new PR.

shreemaan-abhishek:
@liweitianux could you invite @luoluoyuyu with access to your repo?

t/plugin/openid-connect2.t
```diff
@@ -334,3 +334,58 @@ passed
...
+
+=== TEST 10: OpenID server does not have a corresponding redirect_uri, should return 400.
+--- config
+    location /t {
+        content_by_lua_block {
+            local http = require "resty.http"
+            local httpc = http.new()
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello"
+            local res, err = httpc:request_uri(uri, {method = "GET"})
+            ngx.say(res.status)
```
Vacant2333, line 383: Redundant code

t/plugin/openid-connect2.t
```diff
@@ -334,3 +334,58 @@ passed
...
+            if res.status == 400 then
+                ngx.say(res.status)
```
Vacant2333, line 385: the response_body was `true`

Vacant2333 (changes requested):

apisix/plugins/openid-connect.lua
```diff
@@ -86,7 +86,7 @@ local schema = {
         },
         redirect_uri = {
             type = "string",
-            description = "use ngx.var.request_uri if not configured"
+            description = "auto append '.apisix/redirect' if not configured"
```
monkeyDluffy6017, line 89: Could you clarify what you will append `'.apisix/redirect` to ?

luoluoyuyu: > Well. If the route is of type **exact**, then it cannot actually use this `openid-connect` plugin, because the `redirect_uri` must be different from the request URI, as required by `lua-resty-openidc`.

I think this explanation is correct

monkeyDluffy6017 (approved):

moonming (approved):

PR #7691, opened 2022-08-16
feat(openid-connect): Add `symmetric_key` option to support verifying tokens with symmetric algorithms

liweitianux:
### Description

The underlying `lua-resty-openidc` module already supports the
`symmetric_key` option to specify the HMAC key for verifying HS???
tokens.  However, note that `lua-resty-openidc` just passes the
`symmetric_key` value as-is to HMAC.  So we choose to accept a
base64url-encoded secret, which is easier to obtain from OP and
configure, and then decode it before passing it to `lua-resty-openidc`.

### Checklist


liweitianux:
**NOTE**: the `lua-resty-openidc` is having a bug in verifying HS??? tokens, and I've submitted a PR to fix it: https://github.com/zmartzone/lua-resty-openidc/pull/447

apisix/plugins/openid-connect.lua
```diff
@@ -71,6 +72,10 @@ local schema = {
             type = "string",
             description = "the URI will be redirect when request logout_path",
         },
+        symmetric_key = {
+            type = "string",
+            description = "base64url-encoded secret for verifying HS??? tokens",
```
spacewander, line 77: What's `HS???` token?

spacewander (commented):
Could you add some tests for this feature? How can we test it?

apisix/plugins/openid-connect.lua
liweitianux: like HS256, HS512

liweitianux:
> Could you add some tests for this feature? How can we test it?

Hmm, I'll have to first check out the existing tests (e.g., for RS256) and figure out one for the HS256/HS512 tokens.

apisix/plugins/openid-connect.lua
tzssangglass: we can just use `HS256, HS512`, not `HS???`,  it looks unclear

liweitianux: ok. sure :)

liweitianux:
> **NOTE**: the `lua-resty-openidc` is having a bug in verifying HS??? tokens, and I've submitted a PR to fix it: [zmartzone/lua-resty-openidc#447](https://github.com/zmartzone/lua-resty-openidc/pull/447)

Well. `lua-resty-openidc` is actually good. It's the Keycloak (I was testing with) having the issue with HS256 (and other HS algorithms) tokens.

By the way, I'll have to consider how to best pass the `symmetric_key` option (e.g., whether to encode or not).

spacewander:
> It's the Keycloak (I was testing with) having the issue with HS256 (and other HS algorithms) tokens.

Is there any link for this issue in Keycloak? Would you explain it in depth?

spacewander:
> > It's the Keycloak (I was testing with) having the issue with HS256 (and other HS algorithms) tokens.
> 
> Is there any link for this issue in Keycloak? Would you explain it in depth?

Solved it: https://github.com/keycloak/keycloak/issues/13823.

apisix/plugins/openid-connect.lua
kayx23: I updated

kayx23:
Re-opened as this is a meaningful contribution.

Helped resolve conflicts and update the explanation for the `symmetric_key` to be a bit more readable, because the PR author mentioned in a different PR they've changed jobs and might not be able to continue the effort here, especially with adding test cases.

Revolyssup:
@liweitianux Can you no longer work on this PR?

liweitianux:
> @liweitianux Looks like you're busy. We will assign this issue to someone else.

Hi @Revolyssup, sorry that I haven't been using APISIX for some time, and I currently don't have the time for this PR.  Please help reassign it. Thanks.

PR #7639, opened 2022-08-10
feat: convert grpc-status-details-bin in header to HTTP response body…

monkeyDluffy6017:
### Description

Convert `grpc-status-details-bin` in response header to response body, so that users could avoid parsing header to read error response.

Fixes #6996 

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #7728, opened 2022-08-18
fix: recovery old pb state after pubsub.new

monkeyDluffy6017:
### Description

recovery old pb state after pubsub.new, so that it won't affect other module who use pb state.

### Checklist

### DIFF
--- a/apisix/core/pubsub.lua
+++ b/apisix/core/pubsub.lua
@@ -34,7 +34,7 @@ local mt = { __index = _M }
 local pb_state
 local function init_pb_state()
     -- clear current pb state
-    pb.state(nil)
+    local old_pb_state = pb.state(nil)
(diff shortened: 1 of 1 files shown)


spacewander (dismissed):

apisix/core/pubsub.lua
```diff
@@ -49,7 +49,7 @@ local function init_pb_state()
         return "failed to load pubsub protocol: " .. err
```
spacewander, line 49: Maybe we need to apply the same fix in https://github.com/apache/apisix/pull/7639#discussion_r948679096 too?

monkeyDluffy6017: Done

spacewander (approved):

PR #7747, opened 2022-08-19
fix:When the token in the configuration file is a numeric type, the verification fails due to the inconsistency of the types. The reasons are as follows: the requested type (req_token) = string, the configured type (token) = number

funnyboy0128:
### Description

fix:When the token in the configuration file is a numeric type, the verification fails due to the inconsistency of the types. The reasons are as follows: the requested type (req_token) = string, the configured type (token) = number

### Checklist

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -76,7 +76,7 @@ local function check_token(ctx)
 
     local admin
     for i, row in ipairs(local_conf.apisix.admin_key) do
-        if req_token == row.key then
+        if req_token == tostring(row.key) then
             admin = row
             break
         end


apisix/admin/init.lua
```diff
@@ -76,7 +76,7 @@ local function check_token(ctx)
 
     local admin
     for i, row in ipairs(local_conf.apisix.admin_key) do
-        if req_token == row.key then
+        if req_token == tostring(row.key) then
```
spacewander, line 79: We can wrap the key with `""` to ensure key is a string?
Like `"edd1c9f034335f136f87ad84b625c8f1"`

spacewander (commented):
https://github.com/apache/apisix/blob/29f21a0eceb8eedc5d40ba68af154698f0eb70f4/apisix/cli/schema.lua#L193
Would you like to add schema check to admin_key, so that we can reject number type from the beginning?
Add type conversion when using it may make thing complex.

PR #7760, opened 2022-08-22
feat: add workflow plugin(mvp)

tzssangglass:
### Description

I want to combine the existing DSL to support more complex limiting scenarios, this PR is just the first step in providing workflow plugins to implement basic orchestration capabilities.

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/workflow.lua
+++ b/apisix/plugins/workflow.lua
@@ -1,0 +1,118 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core       = require("apisix.core")
+local expr       = require("resty.expr.v1")
+local ipairs     = ipairs
+local tonumber   = tonumber
+local type       = type
+
+local schema = {
+    type = "object",
+    properties = {
+        rules = {
+            type = "array",
+            items = {
+                type = "object",
+                properties = {
+                    case = {
+                        type = "array",
+                        items = {
+                            type = "array",
+                        },
+                        minItems = 1,
+                    },
+                    actions = {
+                        type = "array",
+                        items = {
+                            type = "array",
+                            minItems = 2
+                        }
+                    }
+                },
+                required = {"case", "actions"}
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


apisix/plugins/workflow.lua
```diff
@@ -0,0 +1,123 @@
...
+
+local _M = {
+    version = 0.1,
+    priority = 1006,
+    name = plugin_name,
+    schema = schema
+}
+
+local support_action = {
+    ["return"] = true,
```
soulbird, line 62: It is better to register a handler for each action.

apisix/plugins/workflow.lua
```diff
@@ -0,0 +1,123 @@
...
+
+    return true
+end
+
+
+local function do_action(actions)
+    for _, action in ipairs(actions) do
+        if action[1] == "return" then
+            local code = tonumber(action[2].code)
+            return core.response.exit(code)
```
soulbird, line 105: It's better to also include error messages instead of just code.

apisix/plugins/workflow.lua
```diff
@@ -0,0 +1,123 @@
...
+
+            if action[1] == "return" then
+                if not action[2].code then
+                    return false, "bad actions, code is needed if action is return"
+                end
+
+                if type(action[2].code) ~= "number" then
+                    return false, "bad code, the required type of code is number"
+                end
+            end
```
soulbird, line 93: It's better to add a separate schema validation function for each action, it will look more elegant.

apisix/plugins/workflow.lua
spacewander, line 105: Better to return a status code and do the exiting outside the plugin. Therefore, we can enable features like custom error response. And other actions in the future will need this change too.

apisix/plugins/workflow.lua
```diff
@@ -0,0 +1,123 @@
...
+                        items = {
+                            type = "array",
+                        },
+                        minItems = 1,
+                    },
+                    actions = {
+                        type = "array",
+                        items = {
+                            type = "array",
+                            minItems = 2
```
spacewander, line 42:
```suggestion
                            minItems = 1
```
is enough? Maybe we will add action without args in the future.

apisix/plugins/workflow.lua
tzssangglass: updated

apisix/plugins/workflow.lua
tzssangglass: updated

apisix/plugins/workflow.lua
tzssangglass: updated

apisix/plugins/workflow.lua
tzssangglass: updated

apisix/plugins/workflow.lua
```diff
@@ -0,0 +1,132 @@
...
+            minimum = 100,
+            maximum = 599
+        }
+    },
+    required = {"code"}
+}
+
+
+local function exit(conf)
+    local code = tonumber(conf.code)
```
spacewander, line 75: Is tonumber necessary after the schema check?

spacewander (approved):

apisix/plugins/workflow.lua
tzssangglass: no need, fix it on next PR.

soulbird (approved):

PR #7775, opened 2022-08-24
feat(admin-api): support check conf only, don't write etcd

soulbird:
### Description

Just check resources conf only, don't write it to etcd.

Fixes # (issue)

### Checklist


apisix/admin/init.lua
```diff
@@ -130,13 +130,18 @@ local function run()
...
+    local check_only = false
     -- /apisix/admin/schema/route
     local seg_res, seg_id = uri_segs[4], uri_segs[5]
     local seg_sub_path = core.table.concat(uri_segs, "/", 6)
     if seg_res == "schema" and seg_id == "plugins" then
         -- /apisix/admin/schema/plugins/limit-count
         seg_res, seg_id = uri_segs[5], uri_segs[6]
         seg_sub_path = core.table.concat(uri_segs, "/", 7)
+
+    elseif seg_res == "check_conf" then
```
bzp2010, line 142: I think some comments are needed here to indicate what the check conf interface uri of the Admin API is.

like `-- /apisix/admin/schema/route`

bzp2010:
Is it necessary to add the documentation for these Admin API?

soulbird:
> Is it necessary to add the documentation for these Admin API?

Yes, the documentation is done in another PR

moonming (commented):
I don't understand what problem this PR solves, where is the documentation?

soulbird:
> Description



> I don't understand what problem this PR solves, where is the documentation?

The title already describes the function of this PR. In some cases, such as the API design stage, the user just wants to verify whether the configuration is correct, and does not want to actually submit the configuration to APISIX. This PR provides an API to help users do just that.

soulbird:
> I don't understand what problem this PR solves, where is the documentation?

The title already describes the function of this PR. In some cases, such as the API design stage, the user just wants to verify whether the configuration is correct, and does not want to actually submit the configuration to APISIX. This PR provides an API to help users do just that.

moonming (commented):
You should use the CLI instead of the admin API to solve this problem. Using the admin API to do the test work, is a very weird thing

soulbird:
> You should use the CLI instead of the admin API to solve this problem. Using the admin API to do the test work, is a very weird thing

It should be noted that this is not only to verify the schema, but also includes some reference verification between objects, SSL certificate verification, etc. Even if the CLI is used, APISIX also needs to provide the ability first.

t/admin/check_conf.t
```diff
@@ -0,0 +1,457 @@
...
+{"error_msg":"404 Route Not Found"}
+
+
+
+=== TEST 3: check routes conf: wrong method
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/check_conf/routes/1',
```
starsz, line 83: Add `check_conf` into path.
Maybe a param like `dry-run` would be better?

soulbird:
Introducing a new API to support is not a good method. We will initiate a discussion in the #7790  and reopen it after the solution is confirmed.

PR #7778, opened 2022-08-24
feat: remove useless field in Admin API v3

bzp2010:
### Description

There are still parts of our current Admin API v3 implementation that don't make sense, they exist for compatibility with the etcd v2 protocol, and we need to remove those parts to simplify the Admin API and make it fit our plans.

- remove nested `node` struct on get single record, create, update
- remove empty `node` field on delete
- remove `count` field and add `total` field that depend by paging

This PR completes them and modifies all the test cases involved, so this PR is large, which is mainly all about the data format modification of the test cases.

Fix problem1 in #7789

### Checklist

### DIFF
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -185,7 +185,7 @@ function _M.filter(body)
     filter(body, args)
 
     -- recalculate the amount of filtered data
-    body.count = #body.list
+    body.count = body.list and #body.list or 0
 end
 
 
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -171,6 +171,9 @@ local function filter(body, args)
             table.remove(body.list, i)
         end
     end
+
+    -- recalculate the amount of filtered data
+    body.count = #body.list
 end
 
 
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -171,9 +171,6 @@ local function filter(body, args)
             table.remove(body.list, i)
         end
     end
-
-    -- recalculate the amount of filtered data
-    body.count = #body.list
 end
 
 
@@ -186,6 +183,9 @@ function _M.filter(body)
 
     pagination(body, args)
     filter(body, args)
+
+    -- recalculate the amount of filtered data
+    body.count = #body.list
 end
 
 
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -181,11 +181,13 @@ function _M.filter(body)
 
     local args = request.get_uri_args()
 
-    pagination(body, args)
     filter(body, args)
 
-    -- recalculate the amount of filtered data
-    body.count = body.list and #body.list or 0
+    -- recalculate the total amount of filtered data and remove count field
+    body.total = body.list and #body.list or 0
+    body.count = nil
+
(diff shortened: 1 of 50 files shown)


t/admin/filter.t
```diff
@@ -752,3 +752,129 @@ passed
...
+            ngx.status = code
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 16: pagination data count
```
spacewander, line 821: TEST 16 is equal to 15? And it seems there is no need to recreated the routes.

bzp2010: It looks like I made an incorrect copy and overwrote the code.

bzp2010: changed

apisix/admin/v3_adapter.lua
```diff
@@ -183,6 +183,9 @@ function _M.filter(body)
 
     pagination(body, args)
     filter(body, args)
+
+    -- recalculate the amount of filtered data
+    body.count = body.list and #body.list or 0
```
starsz, line 188: But I think the count should be the total object of the resource. Not the count of the list.

bzp2010: ping @starsz @soulbird

Yes, as I mentioned in the issue, I was thinking the same thing, do you think it would be better to use `count` or add another `total` field?

apisix/admin/v3_adapter.lua
```diff
@@ -181,8 +181,31 @@ function _M.filter(body)
...
+
+        pagination(body, args)
+
+        -- remove the count field returned by etcd
+        -- we don't need a field that reflects the length of the currently returned data,
+        -- it doesn't make sense
+        body.count = nil
+    end
+
+    return body
```
soulbird, line 208: Can we create a new body and then populate its fields? I don't think it's intuitive to change fields on the old body structure. I can't see what the final output data structure looks like with the current code, it's really hard to maintain. cc @tzssangglass

bzp2010: Yes, I also suffer from it, sometimes I do field overwriting and it does not take effect, the problem is strange.

apisix/admin/v3_adapter.lua
```diff
@@ -180,9 +181,33 @@ function _M.filter(body)
     end
 
     local args = request.get_uri_args()
+    local processed_body = deepcopy(body)
```
spacewander, line 184: Why do we need to copy the body?

spacewander: OK, I got it now.

bzp2010: @spacewander In the comment above, @soulbird  points out that we need to copy the body to avoid logical and coding problems.
https://github.com/apache/apisix/pull/7778#discussion_r957974650

Do you think we need it?

spacewander:
Is it possible to move the filter to https://github.com/apache/apisix/blob/7a24ea622b22aa4bf37c3fe19c228505cac003eb/apisix/admin/init.lua#L195?

bzp2010:
@spacewander Currently we don't have a filter for all the data, so I had to filter here to execute the logic separately.

spacewander (approved):

apisix/admin/v3_adapter.lua
tzssangglass: +1

tzssangglass (approved):

PR #7782, opened 2022-08-24
feat(kafka-logger): support sasl conf

biubiue:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/t/plugin/kafka-logger.t
+++ b/t/plugin/kafka-logger.t
@@ -642,12 +642,10 @@ passed
 
 === TEST 21: inject create producer
 --- extra_init_by_lua
-    local setmetatable_p = setmetatable
-    local setmetatable_c = setmetatable
     local producer = require("resty.kafka.producer")
     producer.new = function(self, broker_list, producer_config, cluster_name)
         local opts = producer_config or {}
-        local cli = setmetatable_c({
+        local cli = {
                 broker_list = broker_list,
                 topic_partitions = {},
                 brokers = {},
@@ -661,8 +659,8 @@ passed
                     ssl_verify = opts.ssl_verify or false,
                     resolver = opts.resolver or nil
                 }
-            }, { __index =  { _VERSION = "0.20" } })
-        return setmetatable_p({
+            }
+        return {
             client = cli,
             correlation_id = 1,
             request_timeout = opts.request_timeout or 2000,
@@ -677,7 +675,7 @@ passed
             _timer_flushing_buffer = false,
             ringbuffer = ringbuffer:new(opts.batch_num or 200, opts.max_buffering or 50000),
             sendbuffer = sendbuffer:new(opts.batch_num or 200, opts.batch_size or 1048576)
-        }, { __index =  { _VERSION = "0.20" } })
+        }
     end
 --- request
 GET /hello


apisix/plugins/kafka-logger.lua
```diff
@@ -205,7 +215,8 @@ function _M.log(conf, ctx)
     for host, port in pairs(conf.broker_list) do
         local broker = {
             host = host,
-            port = port
+            port = port,
+            sasl_config = conf.sasl_config or nil
```
spacewander, line 219: Would you please add test to cover it?
To make things simple, you can validate the behavior by injecting the producer.new and checking if sasl_config is passed through.

You can learn about the test framework with:
https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md

t/plugin/kafka-logger.t
```diff
@@ -592,3 +592,40 @@ qr/partition_id: 2/]
...
+                    sasl_config = {
+                        mechanism = "PLAIN",
+                        strategy = "sasl",
+                        user = "admin",
+                        password = "admin-secret"
+                    }
+                }
+            }
+            local message = "halo world"
+            local p = producer:new(broker_list)
```
spacewander, line 617: The test doesn't pass through the code you add.

We can inject a hook or mock into the code. Like:
https://github.com/apache/apisix/blob/b3d4b464045b9bdfed3170cbd7a343ed5c40b24a/t/plugin/zipkin2.t#L35
https://github.com/apache/apisix/blob/b3d4b464045b9bdfed3170cbd7a343ed5c40b24a/t/plugin/opentelemetry.t#L553

t/plugin/kafka-logger.t
```diff
@@ -592,3 +592,44 @@ qr/partition_id: 2/]
...
+=== TEST 20: user sasl create producer
+--- extra_init_by_lua
+local producer = require("resty.kafka.producer")
+producer.send = function()
+    return true
+end
+--- config
+    location /t {
+        content_by_lua_block {
+            local producer = require "resty.kafka.producer"
```
spacewander, line 607: We should test the kafka-logger plugin directly like other tests, instead of using the producer outside of the plugin.

t/plugin/kafka-logger.t
```diff
@@ -592,3 +592,111 @@ qr/partition_id: 2/]
...
+passed
+
+
+
+=== TEST 21: access
+--- request
+GET /hello
+--- response_body
+hello world
+--- wait: 2
```
tzssangglass, line 649: we can inject the producer.new like: https://github.com/apache/apisix/blob/2aa5c48d3acc9bea9b61958a09b51ae99c8a528b/t/plugin/tencent-cloud-cls.t#L220-L236

apisix/plugins/kafka-logger.lua
```diff
@@ -51,6 +51,16 @@ local schema = {
                 },
             },
         },
+        sasl_config = {
+            type = "object",
+            description = "sasl config",
+            properties = {
+                mechanism = { type = "string", description = "mechanism" },
+                password =  { type = "string", description = "password" },
+                user = { type = "string", description = "user" },
```
tokers, line 60: These fields should be required if the sasl config exists?

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -38,6 +38,11 @@ It might take some time to receive the log data. It will be automatically sent a
...
+| sasl_config            | object  | False    |                |                       | Kafka sasl conf.                                                                                                                                                                                                                                                                                                                   |
+| sasl_config.mechanism  | string  | False    |                |                       | Kafka mechanism.                                                                                                                                                                                                                                                                                                                   |
```
tokers, line 42: Add the default value here.

apisix/plugins/kafka-logger.lua
```diff
@@ -51,6 +51,16 @@ local schema = {
                 },
             },
         },
+        sasl_config = {
+            type = "object",
+            description = "sasl config",
+            properties = {
+                mechanism = { type = "string", description = "mechanism", default = "PLAIN" },
```
tokers, line 58: The `mechanism` should be restricted to only use a few enumerations like `PLAIN`, `SCRAM` and so on.

tokers: OK, `SCRAM-SHA-256` is copied from Confluent docs. No worries at all.

spacewander:

t/plugin/kafka-logger.t
```diff
@@ -592,3 +592,118 @@ qr/partition_id: 2/]
...
+
+=== TEST 21: inject create producer
+--- extra_init_by_lua
+    local producer = require("resty.kafka.producer")
+    producer.new = function(self, broker_list, producer_config, cluster_name)
+        if (#broker_list) then
+            ngx.log(ngx.ERR, "unexpected broker_list length: ", #broker_list)
+            return nil
+        end
+        return producer
```
tzssangglass, line 651: The `producer` should not be returned here, you can see that the original returned is a table.

PR #3624, opened 2021-02-21
feat: add pre-commit framework

jbampton:
### What this PR does / why we need it:

From the pre-commit site -> "Git hook scripts are useful for identifying simple issues before submission to code review. We run our hooks on every commit to automatically point out issues in code such as missing semicolons, trailing whitespace, and debug statements. By pointing these issues out before code review, this allows a code reviewer to focus on the architecture of a change while not wasting time with trivial style nitpicks."

refs #3289 

https://pre-commit.com/

The next image shows the final pre-commit run for this pull request:

Screen Shot 2021-02-22 at 3 14 50 am

### Pre-submission checklist:

* [X] Did you explain what problem does this PR solve? Or what new features have been added?
* [ ] Have you added corresponding test cases?
* [ ] Have you modified the corresponding document?
* [ ] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix/tree/master#community) first**

PR #7821, opened 2022-08-30
chore: Add devcontainer to ease out for contributors

jaysonsantos:
### Description
Hi there, first thank you very much for this project.
I had a bit of trouble trying to set-up the project for local development and tried to put things in a standardize way to help onboarding new developers.

### Checklist


nic-6443:
> @nic-6443 could you approve again so the markdown linter runs again?

ok.

jaysonsantos:
@nic-6443 i've added the missing licenses

.devcontainer/change-uid.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function change_gid() {
+    local gid="$1"
+    if [ -z "$gid" ]; then
+        echo "Not changing group id"
+        return 0
+    fi
+    groupmod --gid "$CHANGE_USER_GID" "$USERNAME"
+}
```
tzssangglass, line 40: add a blank line at the end of file

tzssangglass:
CI error: https://github.com/apache/apisix/runs/8125739733?check_suite_focus=true

jaysonsantos:
@tzssangglass how do you feel about adding something like [pre-commit](https://pre-commit.com/) to add all the linters in it?
`make lint` does not cover eclint or the one for markdown

tzssangglass:
> @tzssangglass how do you feel about adding something like [pre-commit](https://pre-commit.com/) to add all the linters in it?
> `make lint` does not cover eclint or the one for markdown

I'm not familiar with this, I remember we had a discussion about pre-commit. ref: https://github.com/apache/apisix/pull/3624

IMO, I think it will increase the maintenance burden. After adding it, we need to at least distinguish which files don't need to be checked.

Anyway, you can open a new issue and let's hear what others think.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY utils ./utils
+RUN ./utils/install-dependencies.sh
+RUN luarocks install luacheck
+
+RUN mkdir /workspaces && chown vscode: /workspaces
+WORKDIR /workspaces
+USER vscode
+
+RUN mkdir -p /home/vscode/.ssh && ssh-keyscan github.com > /home/vscode/.ssh/known_hosts
+RUN git clone https://github.com/openresty/test-nginx && sudo cpanm ./test-nginx
```
soulbird, line 46: Refer to here https://github.com/apache/apisix/blob/master/docs/en/latest/building-apisix.md#building-runtime-for-apisix, you also need to install the dependencies of test-nginx.

soulbird:
I tried it and it doesn't work under Mac(Intel). I got errors like the following:
````
ERROR: for etcd_tls Cannot start service etcd_tls: Mounts denied:
The path /workspaces/apisix/t/certs is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.
````

**Comment by membphis** [comment]
The devcontainer is not good way for APISIX developer, it depends on Node.js, Python and C++ compiler, but APISIX do not need them.

Here is the link of devcontainer:

https://code.visualstudio.com/docs/remote/devcontainer-cli#_system-requirements

File: .devcontainer/Dockerfile (hunk 2 above)
Comment (jaysonsantos) [reply .devcontainer/Dockerfile:46]: hi @soulbird i had a feeling that cpanm already does install the dependencies of the package.
i've managed to run the tests with this same container, did something happen there when you tried?

**Comment by jaysonsantos** [comment]
@soulbird 
> I tried it and it doesn't work under Mac(Intel). 
When does that happen? what `devcontainer build .` returns? Or the problem is after it starts? IIRC, the only time it runs etcd is after create (which probably should be after attach to be honest).

@membphis those dependencies are only for the cli but does not invalidate the use of the dockerfile to install all dependencies which is a bit overwhelming for people that are just starting (in my experience at least).
the good thing is that you can open github codespaces (or vscode) with this repo, and you would have the whole setup done. but if the dependencias are too much and one does not care much about the niceties it add, a simple `docker build -t apisix -f .devcontainer/Dockerfile . && docker run -w $PWD -v $PWD:$PWD -u $(id -u) -g $(id -g) apisix` should drop a shell with some development capabilities out of the box (it still would miss docker in docker for the dependencies).

File: .devcontainer/Dockerfile (hunk 2 above)
Comment (soulbird) [reply .devcontainer/Dockerfile:46]: It seems that some dependent packages are not installed. After I use this method, this step can be passed through
```shell
RUN sudo cpanm --notest Test::Nginx IPC::Run
RUN git clone https://github.com/openresty/test-nginx && sudo cpanm ./test-nginx
``

Comment (jaysonsantos) [reply .devcontainer/Dockerfile:46]: Funky because IPC::Run is one of the first dependencies it tries to download.

```
STEP 12/15: RUN git clone https://github.com/openresty/test-nginx && cpanm ./test-nginx
Cloning into 'test-nginx'...
--> Working on ./test-nginx
Configuring /workspaces/test-nginx ... OK
==> Found dependencies: LWP::UserAgent, IPC::Run, Test::Base, Test::LongString, Text::Diff
--> Working on LWP::UserAgent
...
```
not sure why it failed to you, do you still have a log about that?

Comment (soulbird) [reply .devcontainer/Dockerfile:46]:
```shell
#0 72.73 Building and testing HTTP-Daemon-6.14 ... OK
#0 77.57 Successfully installed HTTP-Daemon-6.14
#0 77.62 ! Installing the dependencies failed: Module 'Net::HTTP' is not installed
#0 77.62 ! Bailing out the installation for libwww-perl-6.67.
#0 77.71 ! Installing the dependencies failed: Module 'LWP::UserAgent' is not installed
#0 77.71 ! Bailing out the installation for Test-Nginx-0.30.
#0 77.71 18 distributions installed
------
Dockerfile-with-features:32
--------------------
  30 |     RUN mkdir -p /home/vscode/.ssh && ssh-keyscan github.com > /home/vscode/.ssh/known_hosts
  31 |     #RUN sudo cpanm --notest Test::Nginx IPC::Run
  32 | >>> RUN git clone https://github.com/openresty/test-nginx && sudo cpanm ./test-nginx
  33 |     ENV PATH=$PATH:/usr/local/openresty/nginx/sbin
  34 |     
--------------------
```

PR #7822, opened 2022-08-30
feat: Add ability to inject headers via prefix to otel traces

jaysonsantos:
### Description
This adds a new feature that allows to inject headers into traces, like, if you have internal headers that would make sense to live as an attribute like x-client-uuid, x-client-version.
It also adds the ability to using prefixes to get the variables, like: `x-client-*`

### Checklist


apisix/plugins/opentelemetry.lua
```diff
@@ -273,6 +280,26 @@ local function create_tracer_obj(conf)
...
+        local is_key_a_match = #key >= 2 and string.sub(key, -1, -1) == "*"
+        local prefix = string.sub(key, 0, -2)
+        local prefix_size = #prefix
+        local val = source[key]
+        if val then
+            core.table.insert(attributes, attr.string(key, val))
+        end
+        if is_key_a_match then
+            for possible_key, value in pairs(source) do
+                if string.sub(possible_key, 0, prefix_size) == prefix then
```
tokers, line 294: We already have a `has_prefix` method in `apisix.core.string`.

apisix/plugins/opentelemetry.lua
```diff
@@ -273,6 +280,25 @@ local function create_tracer_obj(conf)
 end
 
 
+local function inject_attributes(attributes, wanted_attributes, source)
+    for _, key in ipairs(wanted_attributes) do
+        local is_key_a_match = #key >= 2 and string.sub(key, -1, -1) == "*"
+        local prefix = string.sub(key, 0, -2)
```
tzssangglass, line 286:
```suggestion
        local prefix = key:sub(0, -2)
```

spacewander (commented):
I am afraid we can't iterate the `ctx.var` and filter with prefixes to get the variables as it uses a lazy-load mechanism. The wanted attribute may not be loaded when we iterate this object.

Moreover, we already support injecting header via variable `http_x_client_version`.

However, the headers can be filtered by prefixes. So there is still a new feature "inject headers via prefix".  Maybe you can change the title of this PR and also the configuration?

spacewander:
> @spacewander do you prefer me to change the function `inject_attributes` to have a parameter like `with_prefix` and when calling with ctx.var, i could send it as false but true for headers?

LGTM

docs/en/latest/plugins/opentelemetry.md
```diff
@@ -46,6 +46,8 @@ The Plugin only supports binary-encoded [OLTP over HTTP](https://opentelemetry.i
...
+| additional_header_attributes          | array[string] | False    |                                                 |                                                              | Variables and its values which will be appended to the trace span.                                                                                                                                                                           |
+| additional_header_attributes[0]       | string        | True     |                                                 |                                                              | Request headers. For example, `x-my-header"` or `x-my-headers-*` to include all headers with the prefix `x-my-headers-`.                                                                                                                                                                         |
```
starsz, line 50:
```suggestion
| additional_header_attributes[0]       | string        | True     |                                                 |                                                              | Request headers. For example, `x-my-header` or `x-my-headers-*` to include all headers with the prefix `x-my-headers-`.                                                                                                                                                                         |
```

Haha.

apisix/plugins/opentelemetry.lua
```diff
@@ -273,6 +280,25 @@ local function create_tracer_obj(conf)
 end
 
 
+local function inject_attributes(attributes, wanted_attributes, source)
+    for _, key in ipairs(wanted_attributes) do
+        local is_key_a_match = #key >= 2 and key:sub(-1, -1) == "*"
+        local prefix = key:sub(0, -2)
```
tokers, line 286: We can only calculate the `prefix` only if `is_key_a_match` is `true` since it will generate a new string object. What about re-organizing the code to:

```lua
local is_prefix_key = #key >= 2 and key:byte(-1, -1) == str_byte("*")

if is_prefix_key then
    local prefix = key:sub(0, -2)
    for possible_key, value in pairs(source) do
        if core.string.has_prefix(possible_key, prefix) then
            core.table.insert(attributes, attr.string(possible_key, value))
        end
    end
else

    local val = source[key]
    if val then
        core.table.insert(attributes, attr.string(key, val))
    end
end
```

Note it's just a pseudo code.

Makefile
```diff
@@ -460,3 +468,17 @@ ci-env-down:
 	@$(call func_echo_status, "$@ -> [ Start ]")
 	$(ENV_DOCKER_COMPOSE) down
 	@$(call func_echo_success_status, "$@ -> [ Done ]")
+
+
+### eclint: Validate files against editorconfig
```
spacewander, line 473: Could you move the dev kit stuff to a separate PR? Thanks!

apisix/plugins/opentelemetry.lua
```diff
@@ -169,6 +169,13 @@ local schema = {
                 type = "string",
                 minLength = 1,
             }
+        },
+        additional_header_attributes = {
```
spacewander, line 173: Better to rename this conf to show it is used for prefix. Otherwise, people will ask why `additional_attributes` can use `http_header` but we have another conf called additional_header_attributes.

apisix/plugins/opentelemetry.lua
```diff
@@ -273,6 +280,27 @@ local function create_tracer_obj(conf)
 end
 
 
+local function inject_attributes(attributes, wanted_attributes, source, with_prefix)
+    for _, key in ipairs(wanted_attributes) do
+        local is_key_a_match = #key >= 2 and key:sub(-1, -1) == "*" and with_prefix
```
spacewander, line 285: `key:sub(-1, -1)`
Would be better to use `:byte`?

Makefile
jaysonsantos: i've dropped the commit

spacewander (approved):

starsz (approved):

PR #7826, opened 2022-08-31
fix: wait old nginx quit when compress in log roate

monkeyDluffy6017:
### Description

1. If we don't wait for the old nginx to exit first before compressing, we may lose the logs
2. refactor code to ensure that the signal is sent once at a time

### Checklist

### DIFF
--- a/apisix/plugins/log-rotate.lua
+++ b/apisix/plugins/log-rotate.lua
@@ -37,7 +37,9 @@ local str_format = string.format
 local str_reverse = string.reverse
 local tab_insert = table.insert
 local tab_sort = table.sort
-
+local new_tab = require "table.new"
+local ngx_sleep = require("apisix.core.utils").sleep
+local isempty = require "table.isempty"
 local local_conf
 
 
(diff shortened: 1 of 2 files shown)


apisix/plugins/log-rotate.lua
```diff
@@ -219,24 +247,41 @@ end
...
+    core.log.warn("send USR1 signal to master process [", pid, "] for reopening log file")
+    local ok, err = signal.kill(pid, signal.signum("USR1"))
+    if not ok then
+        core.log.error("failed to send USR1 signal for reopening log file: ", err)
+    end
+
+    if enable_compression then
+        -- to avoid losing logs during compression
+        ngx_sleep(0.5)
+        wait_old_nginx_exit()
```
tzssangglass, line 277: sleeping 0.5s then 10s seems strange, sleeping 10s directly?

apisix/plugins/log-rotate.lua
```diff
@@ -169,6 +171,32 @@ local function rename_file(log, date_str)
...
+    local file_path = prefix .. "/logs/nginx.pid.oldbin"
+
+    for i = 1, max_try do
+        -- nginx will delete nginx.pid.oldbin after old master process quit
+        found = file_exists(file_path)
+        if found then
+            if i <= max_try then
+                ngx_sleep(1)
+            end
+
```
tzssangglass, line 187: delete this blank line

apisix/plugins/log-rotate.lua
```diff
@@ -169,6 +171,32 @@ local function rename_file(log, date_str)
 end
 
 
+local function wait_old_nginx_exit()
+    local found = true
+    local max_try = 10  -- 10 seconds
+    local prefix = ngx.config.prefix()
+    local file_path = prefix .. "/logs/nginx.pid.oldbin"
```
tzssangglass, line 178: Just curious, where would renamed the old `nginx.pid` be to `nginx.pid.oldbin`

monkeyDluffy6017: where or why?
1. where 
Before executing the new Nginx binary.
https://github.com/nginx/nginx/blob/f7ea8c76b55f730daa3b63f5511feb564b44d901/src/core/nginx.c#L717
2. why
It's for rollback, if the new Nginx binary doesn't work, we could kill the new one, and use the old Nginx instead.

tzssangglass: I get

PR #7827, opened 2022-08-31
feat(saml-auth): add saml-auth plugin

kingluo:
### Description

Add saml-auth plugin.

### Checklist


apisix/plugins/saml-auth.lua
```diff
@@ -0,0 +1,92 @@
...
+local lrucache = core.lrucache.new({
+    ttl = 300, count = 512
+})
+
+local schema = {
+    type = "object",
+    title = "work with route or service object",
+    properties = {
+        sp_issuer = { type = "string" },
+        idp_uri = { type = "string" },
```
soulbird, line 31: Should be added with strict verification, like`coree.schema_def.uri_def`

kingluo: @soulbird `uri_def` is not suitable in this case, because all uri related properties could be in either full qualified address with host or path only.

soulbird: It's just an example, I think it's a bit too rough to only limit the string.

apisix/plugins/saml-auth.lua
```diff
@@ -0,0 +1,92 @@
...
+            return 500
+        end
+        is_resty_saml_init = true
+    end
+
+    core.log.info("plugin rewrite phase, conf: ", core.json.delay_encode(conf))
+
+    local saml = core.lrucache.plugin_ctx(lrucache, ctx, nil, create_saml_obj, conf)
+
+    local data = saml:authenticate()
```
spacewander, line 87: It's weird to let a plugin hijack the request flow when the authentication is failed. It would bypass APISIX's execution code. lua-resty-openidc does this at some corners, but since lua-resty-saml is owned by us, we can do better.

Maybe we can do it in the future PR?

apisix/cli/ngx_tpl.lua
```diff
@@ -660,6 +664,10 @@ http {
             set $dubbo_method                '';
             {% end %}
 
+            {% if enabled_plugins["saml-auth"] then %}
+            set $saml_data_dir               {*apisix_lua_home*}/deps/share/lua/5.1/resty/saml;
```
spacewander, line 668: Look like we can put the generated saml_data_dir in the Lua code? So there is no need to use extra variable.

apisix/plugins/saml-auth.lua
spacewander: Can any value be suitable as the input? For example, is `""` suitable for sp_issuer?

apisix/cli/ngx_tpl.lua
kingluo: It depends on `apisix_lua_home`, how to determine it in lua code?

.github/workflows/fuzzing-ci.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
         sudo apt-get -y install software-properties-common
         sudo add-apt-repository -y "deb http://openresty.org/package/ubuntu $(lsb_release -sc) main"
         sudo apt-get update
-        sudo apt-get install -y git openresty curl openresty-openssl111-dev unzip make gcc libldap2-dev
+        sudo apt-get install -y git openresty curl openresty-openssl111-dev unzip make gcc libldap2-dev libxml2-dev libxslt-dev libssl-dev
```
membphis, line 53: I don't like this change, it greatly increases the difficulty of building a user's local development environment.

apisix/cli/ngx_tpl.lua
```diff
@@ -432,6 +436,11 @@ http {
             dns_resolver = dns_resolver,
         }
         apisix.http_init(args)
+
+        {% if enabled_plugins["saml-auth"] then %}
+        local constants = require("apisix.constants")
+        constants.saml_data_dir = "{*apisix_lua_home*}/deps/share/lua/5.1/resty/saml"
```
membphis, line 442: confused
Why can't Lua find the default path?

apisix/plugins/saml-auth.lua
```diff
@@ -0,0 +1,92 @@
...
+    priority = 2598,
+    name = plugin_name,
+    schema = schema,
+}
+
+function _M.check_schema(conf, schema_type)
+    return core.schema.check(schema, conf)
+end
+
+local function create_saml_obj(conf)
```
membphis, line 66: `local create_saml_obj = resty_saml.new`

which is better

apisix/plugins/saml-auth.lua
```diff
@@ -0,0 +1,92 @@
...
+        if err then
+            core.log.error("saml init: ", err)
+            return 503
+        end
+        is_resty_saml_init = true
+    end
+
+    core.log.info("plugin rewrite phase, conf: ", core.json.delay_encode(conf))
+
+    local saml = core.lrucache.plugin_ctx(lrucache, ctx, nil, create_saml_obj, conf)
```
membphis, line 85: do we need to check the `saml` object first?

ci/centos7-ci.sh
```diff
@@ -23,7 +23,7 @@ install_dependencies() {
 
     # install build & runtime deps
     yum install -y wget tar gcc automake autoconf libtool make unzip \
-        git sudo openldap-devel which
+        git sudo openldap-devel which libxml2-devel openssl-devel libxslt-devel
```
membphis, line 26: ditto

ci/centos7-ci.sh
```diff
@@ -84,6 +84,9 @@ run_case() {
     export_or_prefix
     make init
     set_coredns
+
+    wget https://raw.githubusercontent.com/api7/lua-resty-saml/main/t/lib/keycloak.lua -O t/lib/keycloak2.lua
```
membphis, line 88: it seems hack

docs/en/latest/plugins/saml-auth.md
```diff
@@ -0,0 +1,123 @@
...
+{
+    "methods": ["GET"],
+    "uri": "/anything/*",
+    "plugins": {
+          "saml-auth": {
+            "idp_cert": "-----BEGIN CERTIFICATE-----\nMIIClzCCAX8CBgGC7zvjWjANBgkqhkiG9w0BAQsFADAPMQ0wCwYDVQQDDAR0ZXN0\nMB4XDTIyMDgzMDE0NDkxNVoXDTMyMDgzMDE0NTA1NVowDzENMAsGA1UEAwwEdGVz\ndDCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAJelXTy4j6pP++7Bq9Yp\nmkuDYoBQRhUUO/RNVNOBPitVisNEQgdpKS2CFYLENZePo0xsDrN+5IB/8MTEqo3V\ntVu2BYQtq5tS0SUEgH1TLIt6gcnOuqgmmk3+RfUM3WGaUlbEN/j1C50RepFJm03i\nAM8dWCsCTCnZQLq/T4T7wwEIIOuAfz7Zip9+8HuVTOUoC3SxwVqTNb0pVRvdg7gU\nYh0f4CnckhlZG+4AgFvyW5LXxyUWbBtIpYEJN8DZEbEg+QmjuUHT0w+Apdtr6MSM\nKT87tQMxRlIYCqjmwPB9xjW1eZ9FdTGu81WeIhvs3iwTBoH6qcBR9ZJltY/VXEhg\neGECAwEAATANBgkqhkiG9w0BAQsFAAOCAQEAhbVRy4EoJzr/B84g9V6aAobUHoEF\ntv02G2GS/3UxrcaBXMLUl/tJlgti+PwQGDzIEXP0b0LLdAh/5LZFeDFVbphXUk26\nXxj3T2ej9ZcsRbaF+yHi0iisyMNdx9gbosDYg8yx9+XWImQMe6+T6+7fjYnilWE8\n0nYUuxyf1GF20EwQ0sjcJbxuq/K/2DZfOT8eo8a5c1oJFB20rE+KEpZnhBnDTQ+s\nUXPjN0AYTCKtPiAd7D190Pk1CQUbfAUmeII9WstCJhmGqIeiSpHsa9mAa50GVT3B\nnHm5obfM+sBEmy78MnC+s9Kt/EquakxeEi+YZEVfgyORq+LwNaef5MuedA==\n-----END CERTIFICATE-----",
```
membphis, line 60: we provide an example, it is enough, which is clear:

```json
 "idp_cert": "-----BEGIN CERTIFICATE----... ...-----END CERTIFICATE-----",
```

t/APISIX.pm
```diff
@@ -724,6 +729,7 @@ _EOC_
             set \$upstream_no_cache              '';
             $a6_ngx_vars
 
+
```
membphis, line 732: it is useless, pls remove this line

docs/en/latest/plugins/saml-auth.md
spacewander: Yes, let's do the same for other cert/key in this file

apisix/cli/ngx_tpl.lua
```diff
@@ -432,6 +436,11 @@ http {
             dns_resolver = dns_resolver,
         }
         apisix.http_init(args)
+
+        {% if enabled_plugins["saml-auth"] then %}
+        local saml = require("apisix.plugins.saml-auth")
```
spacewander, line 441: We should not require a plugin here. This will make it unable to disable the plugin.

Makefile
```diff
@@ -151,6 +151,7 @@ deps: runtime
 	$(eval ENV_LUAROCKS_VER := $(shell $(ENV_LUAROCKS) --version | grep -E -o "luarocks [0-9]+."))
 	@if [ '$(ENV_LUAROCKS_VER)' = 'luarocks 3.' ]; then \
 		mkdir -p ~/.luarocks; \
+		$(ENV_LUAROCKS) config $(ENV_LUAROCKS_FLAG_LOCAL) variables.OPENSSL_DIR $(ENV_OPENSSL_PREFIX); \
```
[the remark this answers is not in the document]
kingluo: The previous commit already add `OPENSSL_DIR` in `linux-install-luarocks.sh`:
https://github.com/apache/apisix/pull/7827/files#diff-4fc86c1a9aa8e4ff0b1e5767d2d2bfa32579829c477d034bea4e8f96fe7ab69d

apisix/cli/ngx_tpl.lua
kingluo: So where do you prefer to set the data_dir? Define `set_data_dir()` in lua-resty-saml and call it here? The `apisix.constants` is not suitable according to @membphis.

utils/install-dependencies.sh
```diff
@@ -84,13 +84,14 @@ function install_dependencies_with_apt() {
...
-    sudo apt-get install -y git openresty curl openresty-openssl111-dev make gcc libpcre3 libpcre3-dev libldap2-dev unzip
+    sudo apt-get install -y git openresty curl openresty-openssl111-dev make gcc libpcre3 libpcre3-dev libldap2-dev unzip libxml2-dev libxslt-dev
 }
 
 # Install dependencies on mac osx
 function install_dependencies_on_mac_osx() {
     # install OpenResty, etcd and some compilation tools
-    brew install openresty/brew/openresty luarocks lua@5.1 wget curl git pcre openldap
+    brew install openresty/brew/openresty luarocks lua@5.1 wget curl git pcre openldap libxml2 libxslt
+:qa
```
tzssangglass, line 94: typo?

t/plugin/saml-auth.t
```diff
@@ -0,0 +1,215 @@
...
+            assert(res.status == 200)
+
+            -- SLO (single logout)
+            res, err = kc.single_logout(uri .. "/logout", saml_cookie, saml_cookie2, keycloak_cookie)
+            if err or res.headers['Location'] ~= "/logout_ok" then
+                ngx.log(ngx.ERR, err)
+                ngx.exit(500)
+            end
+        }
+    }
```
tzssangglass, line 215: add some login failed cases?

utils/install-dependencies.sh
kingluo: @tzssangglass yes, it's typo. thanks.

apisix/cli/ngx_tpl.lua
membphis: we can do this in `init` phase

kingluo: @membphis No matter which place to init the plugin, we need to determine `apisix_home` before apisix start, because saml init needs the absolute path of `xsd/` dir.

spacewander: What about putting the apisix_lua_home in the constants module, so we can use it in the plugin?

kingluo: That's ok.

Issue #7722, opened 2022-08-17
bug: when use response rewrite  plugin, scenarios that return an empty body cannot be supported


Hazel6869:
### Current Behavior

➜  example git:(master) ✗ curl http://127.0.0.1:9080/apisix/admin/routes/1 \
-H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "methods": ["GET"],
    "uri": "/get",
    "plugins": {
        "response-rewrite": {
         "body_base64":false,
            "body": "",
            "headers": {
                "X-Server-id": 3,
                "X-Server-status": "on",
                "X-Server-balancer_addr": "$balancer_ip:$balancer_port"
            },
            "vars":[
                [ "status","==",200 ]
            ]
        }
    },
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "httpbin.org": 1
        }
    }
}'

➜  example git:(master) ✗ curl -X GET -i http://127.0.0.1:9080/get
HTTP/1.1 502 Bad Gateway
Connection: keep-alive
Keep-Alive: timeout=4
Proxy-Connection: keep-alive
Content-Length: 0

### Expected Behavior

➜  example git:(master) ✗ curl -X GET -i http://127.0.0.1:9080/get
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Type: application/json
Date: Wed, 17 Aug 2022 08:44:10 GMT
Keep-Alive: timeout=4
Proxy-Connection: keep-alive
Server: APISIX/2.15.0
X-Server-Balancer-Addr: 34.227.213.82:80
X-Server-Id: 3
X-Server-Status: on

### Steps to Reproduce

 I will fix this

### Environment

- APISIX version (run `apisix version`): version 2.15.0
- Operating system (run `uname -a`): Linux lan 5.15.0-41-generic #44-Ubuntu SMP Wed Jun 22 14:20:53 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): ginx version: openresty/1.21.4.1
built with OpenSSL 1.1.1n  15 Mar 2022
- etcd version, if relevant (run ` `):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
[...]

PR #7836, opened 2022-09-01
fix: response-rewrite plugin might cause Apache AIPSIX hanging

Hazel6869:
### Description

Fixes  https://github.com/apache/apisix/issues/7722

### Checklist

### DIFF
--- a/t/plugin/response-rewrite.t
+++ b/t/plugin/response-rewrite.t
@@ -699,3 +699,50 @@ X-A: 127.0.0.1
 X-B: from 127.0.0.1 to 127.0.0.1:1980
 --- no_error_log
 [error]
+
+
+
+=== TEST 25: set empty body
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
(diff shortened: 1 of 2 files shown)


spacewander:
Could you use `Fixes #7722` in the comment message?

t/plugin/response-rewrite.t
```diff
@@ -699,3 +699,50 @@ X-A: 127.0.0.1
 X-B: from 127.0.0.1 to 127.0.0.1:1980
 --- no_error_log
 [error]
+
+
+
+=== TEST 25: set empty body
```
spacewander, line 705: I run your test locally, and it can be passed even without newly added modification.

Hazel6869: Are there any relevant examples? We tested and found that the original code would go wrong when running this case

Hazel6869: @spacewander  hi you can see https://github.com/Hazel6869/apisix/runs/8244062115?check_suite_focus=true

spacewander (approved):

tokers (approved):

PR #7850, opened 2022-09-02
fix: don't enable passive healthcheck by default

spacewander:
This issue is introduced in the bugfix:
https://github.com/apache/apisix/pull/5589

This PR fixes the bug added by that bugfix, and also
introduce a new test (TEST 6) to cover issue
https://github.com/apache/apisix/issues/4077.

The fix of that issue is submitted in
https://github.com/api7/lua-resty-healthcheck/pull/5
which fixes the bug in an appropriate way.

### Description

Fixes # (issue)

### Checklist


apisix/schema_def.lua
```diff
@@ -277,20 +277,6 @@ local health_checker = {
                     }
                 }
             },
```
leslie-tsang, line 279:
```suggestion
            }
```
Seems we can remove it as well.

tokers (approved):

apisix/schema_def.lua
spacewander: IMHO, I like to keep a comma at the end, so adding a new field doesn't need to touch the original code and break the `git blame`.

leslie-tsang (approved):

soulbird (approved):

PR #7859, opened 2022-09-04
feat: proxy-mirror plugin resolve host

### LINKED ISSUE #7796: feat: As a user, I want to let proxy-mirror plugin resolve the host, so that the DNS resolver in APISIX can be used.
### Description

The proxy-mirror doesn't support brighter DNS resolution, as its feature is currently based on Nginx's directive, which uses Nginx's own DNS resolution.

As Nginx's own DNS resolution doesn't support modern DNS features, this brings some inconvenience, especially under k8s.

To solve the problem, we need to resolve the host ourselves, and pass the result to the Nginx variable `upstream_mirror_uri`.
https://github.com/apache/apisix/blob/3c0757477a8c4e24910bb998d06e1d13ec283738/apisix/plugins/proxy-mirror.lua#L63-L65

We can use the method below to resolve the host:
https://github.com/apache/apisix/blob/3c0757477a8c4e24910bb998d06e1d13ec283738/apisix/core/resolver.lua#L44

Strangevy:
### Description

Fixes #7796 

### Checklist

### DIFF
--- a/t/plugin/proxy-mirror.t
+++ b/t/plugin/proxy-mirror.t
@@ -696,3 +696,49 @@ passed
 passed
 passed
 passed
+
+
+
+=== TEST 23: set mirror requests host to domain
+--- config
+       location /t {
+           content_by_lua_block {
+               local t = require("lib.test_admin").test
+               local code, body = t('/apisix/admin/routes/1',
+                    ngx.HTTP_PUT,
+                    [[{
+                        "plugins": {
+                            "proxy-mirror": {
+                               "host": "http://httpbin.org",
+                               "parh": "/get"
+                            }
+                        },
+                        "upstream": {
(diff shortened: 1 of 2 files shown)


t/plugin/proxy-mirror.t
```diff
@@ -696,3 +696,49 @@ passed
...
+       location /t {
+           content_by_lua_block {
+               local t = require("lib.test_admin").test
+               local code, body = t('/apisix/admin/routes/1',
+                    ngx.HTTP_PUT,
+                    [[{
+                        "plugins": {
+                            "proxy-mirror": {
+                               "host": "http://httpbin.org",
+                               "parh": "/get"
```
tokers, line 713: Typo

PR #7869, opened 2022-09-06
fix: reload once when log rotate

monkeyDluffy6017:
### Description

1.  Waiting for nginx reopen files to avoid losing logs during compression
2.  Refactor code to ensure that the signal is sent once at a time

### Checklist

### DIFF
--- a/apisix/plugins/log-rotate.lua
+++ b/apisix/plugins/log-rotate.lua
@@ -37,7 +37,9 @@ local str_format = string.format
 local str_reverse = string.reverse
 local tab_insert = table.insert
 local tab_sort = table.sort
-
+local new_tab = require "table.new"
+local ngx_sleep = require("apisix.core.utils").sleep
+local isempty = require "table.isempty"
 local local_conf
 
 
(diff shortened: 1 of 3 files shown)


monkeyDluffy6017:
> Is it feasible to add some test cases?
We have log-rotate.t, log-rotate2.t, log-rotate3.t  3 test files to test its functionality.
But I have no idea how to add test case about losing log file or sending several reopen signal.

apisix/plugins/log-rotate.lua
```diff
@@ -37,7 +37,9 @@ local str_format = string.format
 local str_reverse = string.reverse
 local tab_insert = table.insert
 local tab_sort = table.sort
-
+local new_tab = require "table.new"
+local ngx_sleep = require("apisix.core.utils").sleep
+local isempty = require "table.isempty"
```
spacewander, line 42: Better to introduce it via `core.table `

apisix/plugins/log-rotate.lua
```diff
@@ -218,25 +221,42 @@ local function file_size(file)
...
+    core.log.warn("send USR1 signal to master process [", pid, "] for reopening log file")
+    local ok, err = signal.kill(pid, signal.signum("USR1"))
+    if not ok then
+        core.log.error("failed to send USR1 signal for reopening log file: ", err)
+    end
+
+    if enable_compression then
+        -- Waiting for nginx reopen files
+        -- to avoid losing logs during compression
+        ngx_sleep(wait_time)
```
spacewander, line 252: Will the sleep block the timer and cause the next rotation time incorrect?

monkeyDluffy6017: `wait_time` is just 0.5s by default. rotation time is much lager than 0.5s, right?

apisix/plugins/log-rotate.lua
monkeyDluffy6017: Done

tzssangglass (dismissed):

apisix/plugins/log-rotate.lua
spacewander: @monkeyDluffy6017 
Currently, there is no limitation for the user-specific wait_time. BTW, we need to doc it if this argument is exposed to the user.

spacewander: If the wait time is only for hardcode, maybe we can add extra sleep in the test so we don't need to set a special value.

monkeyDluffy6017: ok

monkeyDluffy6017: Done

soulbird (approved):

spacewander (approved):

PR #7884, opened 2022-09-08
fix(file-loger): use no buffering model when open file

### LINKED ISSUE #7839: bug:  [file-logger] two log messages mix-output in one line（两条不同的日志混合在一行输出）
### Current Behavior

I added file-logger plugins in global_rules, and I foud that two different log are mixed in one line.
我这里把file-logger插件添加到全局规则，我发现存在部分日志内容会把两个不同的日志消息混合在一行中输出。
I guess is it related to buffer size which is used to cache log message or something else ?
我猜测这个会不会用于缓存日志消息的buffer size或其他东西有关？
My log message is more than 4K bytes for this case.
在这个bug中，我的日志信息都超过了4K字节大小。

### Expected Behavior

I expected each log message should be output in sperated line.
我期望每条日志记录应该独立的输出到一行。

### Error Logs

no error log, just the behavior.
没有错误日志， 只是看到了错误现象。

### Steps to Reproduce

It happens accidently, and appears  1/1000 in my environment.
这是个偶发现象， 在我的环境中出现概率约为千分之一。

### Environment

- APISIX version 2.15.0:
   apisix is run in docker with image [apache/apisix:2.15.0-centos]

   Linux gateway-web-2-15-0-86687b9f68-xzhdc 4.14.105-19-0024 #1 SMP Tue Jun 14 16:39:55 CST 2022 x86_64 x86_64 x86_64 GNU/Linux
   apisix is run in docker with image [apache/apisix:2.15.0-centos]

   nginx version: openresty/1.21.4.1

   {"version":"2.15.0","etcd_version":"3.4.0","id":"2a887545-0b32-4060-bd5d-5db1057b9db2","boot_time":1661732855,"hostname":"gateway-web-2-15-0-86687b9f68-xzhdc"}

  v2.13.0

  matched to apisix v2.15.0

monkeyDluffy6017:
### Description
  use no buffering model when open file
  Avoid error:
  If the data written to the log file is larger than the default buffer, it will be flushed into the file several times, and other log may be flushed in during this period

Fixes #7839 

### Checklist


apisix/plugins/file-logger.lua
```diff
@@ -116,11 +119,16 @@ local function write_file_data(conf, log_message)
...
+        -- which will cause problem with the log output
+        -- it should be atomic
+        msg = msg .. "\n"
+        local ok, err = file:write(msg)
         if not ok then
             core.log.error("failed to write file: ", conf.path, ", error info: ", err)
         else
-            file:flush()
+            -- No buffering model don't need flush, write to file directly
+            -- file:flush()
```
bzp2010, line 131: If they are no longer needed, do we need to delete it?

tokers:
@monkeyDluffy6017 I don't think just changing the flush mechanism is a good idea, but it can be an alternative for the users. Also, the default Glibc buffer size should be shown (or the way we can calculate it should be given if the buffer size varies on different OS).

monkeyDluffy6017:
> @monkeyDluffy6017 I don't think just changing the flush mechanism is a good idea, but it can be an alternative for the users. Also, the default Glibc buffer size should be shown (or the way we can calculate it should be given if the buffer size varies on different OS).

I don't think we need buffer here, because we write and flush immediately. cc @spacewander

tokers:
> > @monkeyDluffy6017 I don't think just changing the flush mechanism is a good idea, but it can be an alternative for the users. Also, the default Glibc buffer size should be shown (or the way we can calculate it should be given if the buffer size varies on different OS).
> 
> I don't think we need buffer here, because we write and flush immediately. cc @spacewander

The number of access logs can be tremendous if the QPS are high, in such a case, the performance will suffer from a serious degradation. Access logs are different from error logs.

spacewander:
What about using a bigger buffer, like 64KB in Nginx's access_log?

IMHO, there still will have an issue that a log > 64KB might be broken. Maybe using `C.write` and managing the buffer by ourselves via Lua table will be a better idea...

spacewander:
After rethinking this problem, I have a new opinion:
As currently we create the file handler per route instead of per file, it is hard to flush the log in order if we use a buffer per route.

Since the original code doesn't use buffer (it flushes the buffer immediately), I think we can focus on solving the bug first.

monkeyDluffy6017:
@tokers   do you agree with @spacewander ? 
```
Since the original code doesn't use buffer (it flushes the buffer immediately), I think we can focus on solving the bug first.
```

tokers:
> @tokers do you agree with @spacewander ?
> 
> ```
> Since the original code doesn't use buffer (it flushes the buffer immediately), I think we can focus on solving the bug first.
> ```

 Got it.

tokers:
> > @tokers do you agree with @spacewander ?
> > ```
> > Since the original code doesn't use buffer (it flushes the buffer immediately), I think we can focus on solving the bug first.
> > ```
> 
> Got it.

But we may still open an eye about the performance.

tokers (approved):

tzssangglass (approved):

spacewander (approved):

PR #7891, opened 2022-09-09
fix(http-logger): explicitly read the req body in rewrite phase

### LINKED ISSUE #7890: bug: http logger req body empty
### Current Behavior

**English description:**
When the pre plugins returns, the following collection request will be empty
Such as the pre jwt plugin in rewrite phase:
`return 401, {message = "Missing JWT token in request"}`
@see log-util line 179 or line 183
`local body = req_get_body_data() or local body_file = ngx.req.get_body_file()`
body or body_file is empty

**Chinese description:**
如果使用了前置jwt-auth或者其他自定义的鉴权插件，使用如下的退出
`return 401, {message = "Missing JWT token in request"}`
会导致在log阶段调用 log-util 179行 或者 183行读取请求体为空
`local body = req_get_body_data() or local body_file = ngx.req.get_body_file()`

### Expected Behavior

Explicitly read the req body

### Error Logs

_No response_

### Steps to Reproduce

apiVersion: apisix.apache.org/v2beta3
kind: ApisixRoute
metadata:
  name: {{route_name}}
spec:
  http:
  - name:  {{route_name-rule}}
    match:
      paths:
      - "/*"
    backends:
       - serviceName:  {{svcName}}
         servicePort: {{svcPort}}
    plugins:
       - name: jwt-auth
         enable: true
         config:
           key:  {{key}}
           secret: {{secret}}
       - name: http-logger
         enable: true
         config:
           uri: {{url}}
           include_req_body: true
           include_resp_body: true

### Environment

- APISIX version 2.15.0-alpine
- K8s version v1.24.3

cj2a7t:
### Description

Explicitly read the req body: 
When the pre plugins returns, the following collection request will be empty
Such as the pre jwt plugin in rewrite phase:
`return 401, {message = "Missing JWT token in request"}`
@see log-util line 179 or line 183
`local body = req_get_body_data() or local body_file = ngx.req.get_body_file()`
body or body_file is empty

如果使用了前置jwt-plugin或者其他自定义的鉴权插件，使用如下的退出
`return 401, {message = "Missing JWT token in request"}`
会导致在log阶段调用 log-util 179行 或者 183行读取请求体为空
`local body = req_get_body_data() or local body_file = ngx.req.get_body_file()`

Fixes #7890

### Checklist

### DIFF
--- a/apisix/plugins/http-logger.lua
+++ b/apisix/plugins/http-logger.lua
@@ -150,6 +150,22 @@ local function send_http_data(conf, log_message)
 end
 
 
+function _M.rewrite(conf, ctx)
+
+    if conf.include_req_body then
+        -- explicitly read the req body
+        -- When the pre plugins returns, the following collection request will be empty
+        -- Such as the pre jwt plugin in rewrite phase:
+        -- return 401, {message = "Missing JWT token in request"}
+        -- @see log-util line 179 or line 183
+        -- local body = req_get_body_data() or local body_file = ngx.req.get_body_file()
+        -- body or body_file is empty
+        ngx.req.read_body()
+    end
+
+end
+
+
 function _M.body_filter(conf, ctx)
     log_util.collect_body(conf, ctx)
 end


spacewander (commented):
Personally, I don't like this change. It will force Nginx to read unused body, even from an untrusted client.

Maybe we can add it as a case "request body is missing" in the doc: https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/http-logger.md#attributes

cj2a7t:
@spacewander
 I see what you mean.
However, in most of our scenarios, the authentication plugin will go ahead. In the rewrite phase, the logger's rewrite is post-existing and does not exist untrusted client.
Whether to turn it on as a schema configuration，give the user the read_body option to enable configuration.
image

spacewander:
If the authentication plugin goes ahead and Nginx reads the body, it should be available in the logger, unless the body is too big or proxy_request_buffering is disabled.

It is not a good idea to force reading an unused body just because the logger needs it.

cj2a7t:
hi , @spacewander  oh yes, I see what you mean. 
This is not the case that body is too big or proxy_request_buffering is disabled.
You can try the combination of http-logger plugin and jwt-auth plugin.
image
In this case, I want to get the request body from the http-logger(jwt-auth plugin goes ahead return). Is there any other way?

tzssangglass:
> However, in most of our scenarios, the authentication plugin will go ahead.

If you always need to read the body in your scenario, you can try the `lua_need_request_body` directive, but it is not recommended.

cj2a7t:
@tzssangglass  Is it more unfriendly to control through lua_need_request_body command

cj2a7t:
From the perspective of the plug-in only, include_req_body = true.
However, the request body cannot be obtained, which seems difficult for the user to understand.

tzssangglass:
> From the perspective of the logger plugin only, include_req_body = true. However, the request body cannot be obtained, which is ambiguous for users.
This is from your point of view. Would this problem still occur without `jwt-auth` plugin?

I was thinking maybe we could try `ngx.req.read_body()` when get body is nil, now this fix is not reasonable.

cj2a7t:
Copy that. I'll try. Is there any better plan @tzssangglass

tzssangglass:
> Copy that. I'll try. Is there any better plan @tzssangglass

No more idea for now, let's hear from others.

spacewander (changes requested):
I don't think it needs to be fixed.

The log code should avoid changing the behavior of the normal code.

When event A happens, we can log event A down. But we can't say because we want to log event A, so event A should happen.

spacewander:
> > From the perspective of the logger plugin only, include_req_body = true. However, the request body cannot be obtained, which is ambiguous for users.
> > This is from your point of view. Would this problem still occur without `jwt-auth` plugin?
> 
> I was thinking maybe we could try `ngx.req.read_body()` when get body is nil, now this fix is not reasonable.

There are many situations that `get body is nil`, for example:
1. Nginx doesn't read the body (this case)
2. the body is too large so Nginx stores it in a temp file
3. the request doesn't have a body
4. Nginx is configured to pass the request body streamingly (which will be broken if we always force Nginx to read the body just because someone wants to log the body)
5. Nginx is configured to always buffer the request body in a temp file

Not every situation needs to be "fixed"

Issue #7569, opened 2022-07-29
Apisix with multiple K8s clusters as a source for service discovery


afbm08:
### Description

As a user, I need to use multiple K8s clusters as a source for service discovery.

In forum, Zeping Bai told me that at this time it's not possible to create a multiple similar service discovery.

It's possible develop that feature?

zhixiongdu027:
I plan to complete this feature, but
The configuration form may be worth discussing.

1: Use a new discovery name such as multi-kubernetes, but only supports multi-cluster

2: The current kubernetes discovery supports both single and multiple clusters, but the compatibility cost may be high

3: Bread update kubernetes discovery, supports both single and multi-cluster

Maybe someone can suggest a new direction

@afbm08 @tao12345666333 @tokers @tzssangglass

tao12345666333:
Maybe it's not just the needs of a specific module

tokers:
Well, I don't see we need to add another service discovery module as its core capability is same to the existing one.

tokers:
As for the compatibility, just adding another field: `clusters`, which is an array is enough. One can configure it via configuring a bunch of kubernetes clusters via the `clusters` array or using the existing one. Sub fields in the `clusters` are same to the flatten fields.

spacewander:
> As for the compatibility, just adding another field: `clusters`, which is an array is enough. One can configure it via configuring a bunch of kubernetes clusters via the `clusters` array or using the existing one. Sub fields in the `clusters` are same to the flatten fields.

+1

zhixiongdu027:
Okay, Let's move on

zhixiongdu027:
Hello, every one:

   This work is continuing, and I hope you can continue to discuss the configuration issue



At present, the demo I have completed supports the configuration in the following two modes:

# Singleton Mode

compatible with current configuration, except for add shared_size field

## config.yaml :
```yaml
kubernets:
    schema:  
      ....   
    services:
      ....   
    client:
      ....  
    shared_size: 1m 
```

and rendered out nginx.conf as follows
## nginx.conf:

```
http {

  ...
  
  lua_shared_dict kubernetes 1m;

  ....
}
```

# Multiple Mode

configuration same as the singleton pattern except that they under an arrays

## config.yaml

```yaml
kubernets:
  - services: ...
    client: ...
    id: "release"
    shared_size: 2m
  - services: ...
    client: ...
    id: "debug"
    shared_size: 1m
```


[...]

Issue #7199, opened 2022-06-06
docs: about apisix discovery for kubernetes


liangyawang:
### Current State

> Hello team members, I had the following problems when I made automatic discovery based on apifix and kubernetes according to the official documents
> 1. use the token file as the interactive voucher between apifix and kuberentes. The missing field in the error header cannot be connected normally, and then replace it with a token string to complete the link
In my token file at that time, only the token string was pasted without any other fields
> 2. when I call the API to automatically discover the route of type kubernetes, I use curl to request and give an error prompt for the non optional upstream
When I help with troubleshooting through team members, the troubleshooting process is as follows:
> 1. service_ The name field is incorrectly filled in. For example: /test/test app
> 2. not in service_ Add the corresponding service port in the name field
> 3. we mistakenly think that we need to add a service port directly. After troubleshooting, we mapped it through the port name corresponding to the port in the endpoints
> The team is expected to automatically find kubernetes in the official document
> 1. for token_ File problem
> 2. add a corresponding example in this discovery type document to create an automatic discovery route with kubernetes type through API
> 3. the example should give an example and explain the detailed configuration format of each field. For example, today's port will be checked for a while
> 4. the detailed information of the corresponding namespace service port shall be displayed during the example for reference
> 5. the log should be printed for each environment and adjusted according to the level of the configuration file, because today I want to see if he really gets the data, but I don't know where to add the log to print for users who are unfamiliar with Lua

### Desired State

- Display service information
```
[root@ ~]# kubectl get pods -n test |grep test-app
test-app-6f57f6fffb-gfblz                   1/1     Running            0          6d23h

[root@ ~]# kubectl get endpoints -n test test-app
NAME          ENDPOINTS             AGE
test-app   10.223.216.201:8080   1d

[root@ ~]# kubectl describe endpoints -n test test-app
Name:         test-app
Namespace:    test
Labels:       app=test-app
Annotations:  endpoints.kubernetes.io/last-change-trigger-time: 2022-06-02T19:07:21+08:00
Subsets:
  Addresses:          10.223.216.201
  NotReadyAddresses:  <none>
  Ports:
    Name     Port  Protocol
    ----     ----  --------
    8080tcp  8080  TCP                        #Note the port name here

Events:  <none>
```

- Show the route creation method of the corresponding service
Since the dashboard does not support routes of kubernetes type, it is created by calling API
```
[...]

Issue #7185, opened 2022-06-02
bug: Independent external APIs cannot be automatically discovered by linking APIs servers


liangyawang:

spacewander:
Could you add a log 
```
    local cjson = require("cjson.safe"); ngx.log(ngx.EMERG, "AA", cjson.encode(apiserver))
```
in https://github.com/apache/apisix/blob/4562875348fdaab3cb05e6df8aecf0604fa6e157/apisix/discovery/kubernetes/informer_factory.lua#L48?

liangyawang:
您好 请问您会讲中文吗 如果会的话 我向您提问一下 这行代码加在lua文件的什么位置？
还有个问题，当我把配置中的token_file改位token时它不再出现报错 但是我也不能确定它是否获取到了对应的endpoints

> 2022年6月2日 下午5:45，罗泽轩 ***@***.***> 写道：
> 
> local cjson = require("cjson.safe"); ngx.log(ngx.EMERG, "AA", cjson.encode(apiserver))

tzssangglass:
> 您好 请问您会讲中文吗 如果会的话 我向您提问一下 这行代码加在lua文件的什么位置？

In the line above this line of code: https://github.com/apache/apisix/blob/4562875348fdaab3cb05e6df8aecf0604fa6e157/apisix/discovery/kubernetes/informer_factory.lua#L48

The new code was added to print what `apiserver ` is.

liangyawang:
请回复一下这个问题

> 2022年6月3日 下午7:35，梁亚旺 ***@***.***> 写道：
> 
> 我尝试通过api创建一个路由 因为dashboard还没办法创建 命令如下
> <粘贴的图形-1.tiff>
> 
> 但是 当我尝试通过apisix访问域名时 发现apisix报错如下
> 2022/06/03 19:25:53 [error] 3135#3135: *10445111 [lua] init.lua:517: http_access_phase(): failed to set upstream: no valid upstream node: nil, client: 100.100.100.100, server: _, request: "GET /actuator/health HTTP/1.1", host: "teacher-test.xxxx.cn <http://teacher-test.xxxx.cn/>”
> 
> 官网中没有对kubernetes的自动发现做详细的描述 所以此刻 我不清楚问题出在了哪里 
> 上图中 service_name字段 yace是我k8s中的namespace teacher-app是我的endpoints名称 
>> 2022年6月2日 下午5:45，罗泽轩 ***@***.*** ***@***.***>> 写道：
[...]

PR #7895, opened 2022-09-11
feat: support multiple kubernets clusters discovery

zhixiongdu027:
### Description

Fixes # (issue)
 #7569 ->  -- Support mulitple kubernets cluster
 #7199 , #7185 ->  -- Remove possible extra spaces in client.token
### Checklist

- [y] I have explained the need for this PR and the problem it solves
- [y] I have explained the changes or the new features added to this PR
- [y] I have added tests corresponding to this change
- [] I have updated the documentation to reflect this change
- [y] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

 
### Todo
  Doc ...


zhixiongdu027:
@spacewander @tokers @tzssangglass 
Hello:
  
  I think the code part is already ready for review,
  
  The documentation part will be submitted later
  
  Please help for review ,Many TKS.

apisix/discovery/kubernetes/init.lua
```diff
@@ -298,88 +301,227 @@ local function create_endpoint_lrucache(endpoint_key, endpoint_port)
...
+    for _, conf in ipairs(confs) do
+        local id = conf.id
+
+        if ctx[id] then
+            error("duplicate id value")
+        end
+
+        local endpoint_dict = ngx.shared["kubernetes-" .. id]
+        if not endpoint_dict then
+            error(core.table.concat {
```
spacewander, line 440: I would recommend using `string.format` which is more common

t/kubernetes/discovery/kubernetes2.t
```diff
@@ -0,0 +1,671 @@
...
+
+BEGIN {
+    our $token_file = "/tmp/var/run/secrets/kubernetes.io/serviceaccount/token";
+    our $token_value = eval {`cat $token_file 2>/dev/null`};
+
+    our $yaml_config = <<_EOC_;
+apisix:
+  node_listen: 1984
+  config_center: yaml
+  enable_admin: false
```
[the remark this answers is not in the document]
spacewander: I found an one-line script to remove them:
`grep -lP 'config_center: yaml\n\s+enable_admin: false' t/*/*.t |xargs -r sed -i -r -z 's/\n\s+enable_admin: false\n/\n/g'`

apisix/discovery/kubernetes/schema.lua
```diff
@@ -28,113 +28,228 @@ local port_patterns = {
...
                         type = "string",
-                        pattern = namespace_regex_pattern
+                        pattern = [[^[a-z0-9]{1,6}$]]
                     },
-                    minItems = 1
-                },
-                not_match = {
-                    type = "array",
-                    items = {
+                    service = {
```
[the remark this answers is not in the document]
zhixiongdu027: Maybe not a good idea , because service had default value in single mode but no default value in multiple mode .

t/kubernetes/discovery/kubernetes2.t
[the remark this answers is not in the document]
zhixiongdu027: seems like remove "enable_admin: false" will cause all test case fail

image

apisix/discovery/kubernetes/schema.lua
```diff
@@ -28,113 +28,228 @@ local port_patterns = {
...
-                host = "${KUBERNETES_SERVICE_HOST}",
-                port = "${KUBERNETES_SERVICE_PORT}",
-            }
-        },
-        client = {
-            type = "object",
-            properties = {
-                token = {
-                    type = "string",
+                namespace_selector = {
```
spacewander, line 104: What about reusing the namespace_selector schema?

zhixiongdu027: ok, I'll try to optimize the schema fully

apisix/discovery/kubernetes/init.lua
```diff
@@ -298,88 +302,217 @@ local function create_endpoint_lrucache(endpoint_key, endpoint_port)
...
+        return nil
+    end
+
+    return endpoint_lrucache(service_name, endpoint_version,
+            create_endpoint_lrucache, endpoint_dict, endpoint_key, endpoint_port)
+end
+
+
+function _M.init_worker()
+    local discovery_conf = local_conf.discovery.kubernetes
```
tzssangglass, line 507: we should update `config-default.yaml`, added `id` attribute

zhixiongdu027: sorry i didn't understand what you mean

tzssangglass: here: https://github.com/apache/apisix/blob/2a44ba61b25224f585dc67caed7131d248cf9792/conf/config-default.yaml#L307-L336

only single_mode_nodes configuration here

tzssangglass:
A discussion:

In the current implementation, single cluster and multiple clusters are implemented separately in the code.

Can we support only multiple clusters, the original single cluster just has a cluster of multiple clusters.

apisix/discovery/kubernetes/init.lua
zhixiongdu027: Ok ,get your point .

zhixiongdu027:
> "Can we support only multiple clusters"

Makes me think it has two meanings.

1: Abandon compatibility with the current single cluster configuration, users need to manually modify the previous configuration after the upgrade, otherwise it cannot be used

2: Compatible with the current single cluster configuration, but modify the code to treat the single cluster as a multi-cluster configuration with only one array element

I think you may mean 2,

However, there is no full benefit, but it brings performance loss.
Because the performance when dealing with multi-cluster configurations is lower than that of single-cluster configurations

apisix/discovery/kubernetes/init.lua
```diff
@@ -298,88 +302,217 @@ local function create_endpoint_lrucache(endpoint_key, endpoint_port)
...
+        end
 
-    return endpoint_lrucache(service_name, endpoint_version,
-            create_endpoint_lrucache, endpoint_key, endpoint_port)
+        local ok, status = pcall(handle.list_watch, handle, handle.apiserver)
+
+        local retry_interval = 0
+        if not ok then
+            core.log.error("list_watch failed, kind: ", handle.kind,
+                    ", reason: ", "RuntimeException", ", message : ", status)
```
tzssangglass, line 323: `"RuntimeException"` does not look like APISIX style logs

zhixiongdu027: But i think is kubernetes style

apisix/cli/ops.lua
```diff
@@ -653,36 +653,52 @@ Please modify "admin_key" in conf/config.yaml .
         end
     end
 
-    -- inject kubernetes discovery environment variable
+    -- inject kubernetes discovery shared dict and environment variable
     if enabled_discoveries["kubernetes"] then
 
-        local kubernetes_conf = yaml_conf.discovery["kubernetes"]
+        if not sys_conf["discovery_shared_dict"] then
```
tokers, line 659: Use plural: `discovery_shared_dicts`.

apisix/cli/ngx_tpl.lua
```diff
@@ -238,8 +238,11 @@ http {
     lua_shared_dict balancer-ewma-last-touched-at {* http.lua_shared_dict["balancer-ewma-last-touched-at"] *};
     lua_shared_dict etcd-cluster-health-check {* http.lua_shared_dict["etcd-cluster-health-check"] *}; # etcd health check
 
-    {% if enabled_discoveries["kubernetes"] then %}
-    lua_shared_dict kubernetes {* http.lua_shared_dict["kubernetes"] *};
+    # for discovery shared dict
+    {% if discovery_shared_dict then %}
+    {% for key, size in pairs(discovery_shared_dict) do %}
+    lua_shared_dict {*key*} {*size*};
+    {% end %}
```
tzssangglass, line 245: We can add test cases to cover here?

apisix/discovery/kubernetes/init.lua
tzssangglass: People view APISIX logs from here, and we don't use their style of logs for every new ecology we dock.

tzssangglass:
> > "Can we support only multiple clusters"
> 
> Makes me think it has two meanings.
> 
> 1: Abandon compatibility with the current single cluster configuration, users need to manually modify the previous configuration after the upgrade, otherwise it cannot be used
> 
> 2: Compatible with the current single cluster configuration, but modify the code to treat the single cluster as a multi-cluster configuration with only one array element
> 
> I think you may mean 2,
> 
> However, there is no full benefit, but it brings performance loss. Because the performance when dealing with multi-cluster configurations is lower than that of single-cluster configurations

I mean 1. 

> but it brings performance loss.
> Because the performance when dealing with multi-cluster configurations is lower than that of single-cluster configurations

It's persuasive. Can you talk more about `performance loss`.

zhixiongdu027:
> I mean 1.
> 
Maybe this needs more people to join the discussion


> It's persuasive. Can you talk more about `performance loss`.

Additional id detection and table field access are required in the multiple_mode_nodes function, obviously there is a certain performance loss

apisix/discovery/kubernetes/init.lua
zhixiongdu027: I think it's best practice to keep things as they are

In the List-Watch process, there are many error messages that may appear in many Kubernetes responses, all of them are in this format (Reason and Message)

It's not necessary to change the raw error message content for "apisix style"

spacewander (approved):

tzssangglass (approved):

spacewander:
@zhixiongdu027 
We are going to release a new version next week. Would you like to add a doc before that?

zhixiongdu027:
I will asap, it may take three to four days

wolgod:
i see  lua_shared_dict kubernetes default is 1m , If there are many services in multiple cluster, it may not be enough, but it cannot be set very large. How to evaluate the memory usage of a service?

zhixiongdu027:
> i see lua_shared_dict kubernetes default is 1m , If there are many services in multiple cluster, it may not be enough, but it cannot be set very large. How to evaluate the memory usage of a service?

@wolgod 

Please don't worry. 
Usually, 1m of memory can store about 1000 endpoints information

And in multi-cluster mode, each cluster uses its own ngx.shared.DICT.
For example, if you define a release cluster and a debug cluster,
Then there will be two ngx.shared.DICT: 
   [ kubernetes-release ,  kubernetes-debug ] ,
and you can specify the memory size separately

In addition, you can use namespace_selector and label_selector to filter endpoint information to reduce memory

wolgod:
> 
thanks, I wonder if the community will support apisix-seed service discovery for kubernetes discovery in the future?

zhixiongdu027:
@wolgod 
This is not a place for discussion, you should create a ISSUE

wolgod:
thanks,  Will there be a release for 2.15.x support multiple kubernets clusters discovery ？ because apisix 2.x can't switch to version 3 seamlessly

tzssangglass:
> thanks, Will there be a release for 2.15.x support multiple kubernets clusters discovery ？ because apisix 2.x can't switch to version 3 seamlessly

No, functional PRs are not back-ported.

PR #7923, opened 2022-09-15
fix(cors): allow origins and regex conflict

bzp2010:
### Description

The default value of the current `allow_origins` configuration conflicts with the value of `allow_origins_by_regex`, causing the value entered by the user to be ignored.

This PR will proactively remove the default value of `allow_origins` when the `allow_origins_by_regex` configuration exists. A new test case has been added and no changes have been made to the old one, which is now fully compatible.

### Checklist


apisix/plugins/cors.lua
```diff
@@ -164,13 +164,11 @@ function _M.check_schema(conf, schema_type)
...
-        if conf.allow_origins == "*" or conf.allow_methods == "*" or
-            conf.allow_headers == "*" or conf.expose_headers == "*" then
-            return false, "you can not set '*' for other option when 'allow_credential' is true"
-        end
-    end
+
+    -- When allow_origins_by_regex is present we need to clear the default value of allow_origins
+    -- first to avoid errors in the check in allow_credential below.
     if conf.allow_origins_by_regex then
+        conf.allow_origins = nil
```
spacewander, line 171: Maybe we should only clear when the value is `*`?
Otherwise, conf like `allow_origins = "apple', allow_origins_by_regex = "bana*"` will not work.

bzp2010: @spacewander 

In my opinion, `allow_origins_by_regex` is a more advanced form of allow_origins. Once the user decides to use `allow_origins_by_regex`, then we should not use `allow_origins` at same time. With regular expressions, the user can implement whatever he wants to implement with allow_origins.

So my implementation is as you can see, I fully believe that we should not mix these two patterns, it will bring extra mental cost to the user, the user needs to understand in particular, how the execution order of the text pattern and the regular pattern is.

Even if possible, I would like to add mandatory checks directly via jsonschema to make sure users don't set them at the same time. What do you think?

tokers: @bzp2010 Then we should also show the relationship between these two fields in the doc.

spacewander: > Once the user decides to use allow_origins_by_regex, then we should not use allow_origins at same time.

This will be a break change. Note that we are going to backport this fix. Maybe we can discuss this issue in the mail list.

bzp2010: OK, I will start a discussion in the mail list.

Qiuqiu0505:
Hi，I also met the same preblem and are there any new conclusions now? Thanks

PR #7925, opened 2022-09-15
feat: support ssl key-encrypt-salt rotation

monkeyDluffy6017:
### Description

Currently, the key_encryption_salt cannot be changed once users use it to encrypt the private key, otherwise, Apache APISIX cannot decrypt the private key correctly. This may become a pain point when the user leaks the salt.
As a user, I can configure multiple key_encryption_salt for Apache APISIX, and Apache APISIX will use them to decrypt private keys in turn

### Checklist


conf/config-default.yaml
```diff
@@ -114,9 +114,10 @@ apisix:
     ssl_session_tickets: false              #  disable ssl_session_tickets by default for 'ssl_session_tickets' would make Perfect Forward Secrecy useless.
                                             #  ref: https://github.com/mozilla/server-side-tls/issues/135
 
-    key_encrypt_salt: edd1c9f0985e76a2      #  If not set, will save origin ssl key into etcd.
-                                            #  If set this, must be a string of length 16. And it will encrypt ssl key with AES-128-CBC
-                                            #  !!! So do not change it after saving your ssl, it can't decrypt the ssl keys have be saved if you change !!
+    key_encrypt_salt:             #  If not set, will save origin ssl key into etcd.
```
membphis, line 117: Is this an incompatible change?

monkeyDluffy6017: it's backward compatible, we could use 2 ways at the same time

monkeyDluffy6017: I have test case about this case

apisix/ssl.lua
```diff
@@ -55,23 +56,37 @@ function _M.server_name()
...
+        local ivs = core.table.try_read_attr(local_conf, "apisix", "ssl", "key_encrypt_salt")
+        local type_ivs = type(ivs)
+
+        if type_ivs == "table" then
+            for index, iv in ipairs(ivs) do
+                if type(iv) == "string" and #iv == 16 then
+                    local aes_with_iv = assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv}))
+                    core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
+                else
+                    core.log.error("the key_encrypt_salt does not meet the "
```
spacewander, line 73: We can use `,` instead of `..` as separator

apisix/ssl.lua
```diff
@@ -86,32 +101,32 @@ function _M.aes_encrypt_pkey(origin)
...
+local function aes_decrypt_pkey(origin)
+    if core.string.has_prefix(origin, "---") then
+        return origin
     end
 
-    local decrypted = iv:decrypt(decoded_key)
-    if not decrypted then
-        core.log.error("decrypt ssl key failed. key[", key, "] ")
+    local aes_128_cbc_with_iv_tbl = get_aes_128_cbc_with_iv()
+    if core.table.isempty(aes_128_cbc_with_iv_tbl) then
```
spacewander, line 110: Use `#` would be better as the table must be array

t/admin/ssl4.t
```diff
@@ -0,0 +1,382 @@
...
+                },
+                "key": "/apisix/ssls/1"
+            }]]
+            )
+
+        ngx.status = code
+        ngx.say(body)
+    }
+}
+--- request
```
spacewander, line 128: Let's move the common part to the top like other test files

apisix/ssl.lua
```diff
@@ -55,23 +56,37 @@ function _M.server_name()
...
-        if type(iv) =="string" and #iv == 16 then
-            _aes_128_cbc_with_iv = assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv}))
-        else
-            _aes_128_cbc_with_iv = nil
+        local ivs = core.table.try_read_attr(local_conf, "apisix", "ssl", "key_encrypt_salt")
+        local type_ivs = type(ivs)
+
+        if type_ivs == "table" then
+            for index, iv in ipairs(ivs) do
+                if type(iv) == "string" and #iv == 16 then
```
spacewander, line 69: Better to check this via schema in https://github.com/apache/apisix/blob/master/apisix/cli/schema.lua

tokers: +1

conf/config-default.yaml
```diff
@@ -114,9 +114,10 @@ apisix:
...
+      - edd1c9f0985e76a2          #  If set this, must be a string of length 16. And it will encrypt ssl key with AES-128-CBC
+                                  #  !!! So do not change it after saving your ssl, it can't decrypt the ssl keys have be saved if you change !!
+                                  #  Only use the first key to encrypt, and decrypt in the order of the array.
```
tokers, line 120: The comment is not accurate to describe the field, since the comment says it "must be a string of length 16".

apisix/ssl.lua
monkeyDluffy6017: Done

t/admin/ssl4.t
monkeyDluffy6017: Done

apisix/ssl.lua
monkeyDluffy6017: Done

apisix/ssl.lua
monkeyDluffy6017: we need backward compatibility, how to support both table and string if use schema?

monkeyDluffy6017: Done

conf/config-default.yaml
monkeyDluffy6017: Done

apisix/cli/schema.lua
```diff
@@ -222,7 +222,25 @@ local config_schema = {
                                     }
                                 }
                             }
-                        }
+                        },
+                        key_encrypt_salt = {
+                            anyOf = {
+                                {
+                                type = "array",
```
spacewander, line 229: Let's add length check for the array

apisix/cli/schema.lua
```diff
@@ -222,7 +222,25 @@ local config_schema = {
...
+                                    maxLength = 16
+                                    }
+                                },
+                                {
+                                    type = "string",
+                                    minLength = 16,
+                                    maxLength = 16
+                                }
+                            }
+
```
spacewander, line 242: Why add an extra blank line?

apisix/ssl.lua
```diff
@@ -55,23 +56,32 @@ function _M.server_name()
...
-            _aes_128_cbc_with_iv = nil
+        local ivs = core.table.try_read_attr(local_conf, "apisix", "ssl", "key_encrypt_salt")
+        local type_ivs = type(ivs)
+
+        if type_ivs == "table" then
+            for _, iv in ipairs(ivs) do
+                local aes_with_iv = assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv}))
+                core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
+            end
+        elseif type_ivs == "string" and #ivs == 16 then
```
spacewander, line 72: We can remove the `#ivs == 16` check as we already checked it in the schema.

apisix/cli/schema.lua
monkeyDluffy6017: minItems = 1? why need this check?

apisix/cli/schema.lua
monkeyDluffy6017: Done

apisix/ssl.lua
monkeyDluffy6017: Done

t/admin/ssl4.t
```diff
@@ -0,0 +1,358 @@
...
+
+
+
+=== TEST 3: client request with the old style key_encrypt_salt
+--- yaml_config
+apisix:
+    node_listen: 1984
+    ssl:
+        key_encrypt_salt: "edd1c9f0985e76a1"
+--- response_body eval
```
soulbird, line 181: The judgment of the response body does not seem necessary.

t/admin/ssl4.t
```diff
@@ -0,0 +1,358 @@
...
+                ngx.say("failed to send http request: ", err)
+                return
+            end
+
+            ngx.say("sent http request: ", bytes, " bytes.")
+
+            while true do
+                local line, err = sock:receive()
+                if not line then
+                    -- ngx.say("failed to receive response status line: ", err)
```
soulbird, line 76: Please remove redundant comments.

apisix/cli/schema.lua
```diff
@@ -222,7 +222,25 @@ local config_schema = {
                                     }
                                 }
                             }
-                        }
+                        },
+                        key_encrypt_salt = {
+                            anyOf = {
```
tokers, line 227: I know `anyOf` can satisfy the schema check but it should be `oneOf` in this scene?

conf/config-default.yaml
tokers: The comment is still wrong.

The `key_encrypt_salt` should either be a string whose size is `16` or an array whose elements are string, and the size is also `16`.

apisix/ssl.lua
```diff
@@ -55,23 +56,32 @@ function _M.server_name()
...
+        local type_ivs = type(ivs)
+
+        if type_ivs == "table" then
+            for _, iv in ipairs(ivs) do
+                local aes_with_iv = assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv}))
+                core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
+            end
+        elseif type_ivs == "string" then
+            local aes_with_iv = assert(aes:new(ivs, nil, aes.cipher(128, "cbc"), {iv = ivs}))
+            core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
```
tzssangglass, line 74: how about `type_ivs` is not `table` or `string`?

t/admin/ssl4.t
monkeyDluffy6017: Just a double check, i prefer to leave it there

t/admin/ssl4.t
monkeyDluffy6017: Done

apisix/cli/schema.lua
monkeyDluffy6017: But this is the conventional way in APISIX, it's used everywhere, so i think i should follow this.

apisix/ssl.lua
monkeyDluffy6017: It will raise an error because the limitation in schema.lua.

apisix/ssl.lua
```diff
@@ -55,23 +56,32 @@ function _M.server_name()
 end
 
 
-local _aes_128_cbc_with_iv = false
+local _aes_128_cbc_with_iv_tbl = false
```
spacewander, line 59: Better to avoid using boolean value as the default value of a table?

monkeyDluffy6017: Done

spacewander (approved):

tzssangglass (approved):

conf/config-default.yaml
kayx23: I'm updating this to `a hexadecimal string of length 16`. It's technically a 8 byte value in hex representation. Size isn't accurate.

PR #7926, opened 2022-09-15
fix: check config.yaml plugins list is nil

zhendongcmss:
### Description

when config.yaml plugin list is nil, apisix will print an error.
```
22/09/15 08:46:34 [warn] 49#49: *7 [lua] plugin.lua:271: load(): failed to read stream plugin list from local file, context: init_worker_by_lua*
2022/09/15 08:46:34 [error] 46#46: *101 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/admin/init.lua:318: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
coroutine 0:
        [C]: in function 'ipairs'
        /usr/local/apisix/apisix/admin/init.lua:318: in function 'sync_local_conf_to_etcd'
        /usr/local/apisix/apisix/admin/init.lua:421: in function </usr/local/apisix/apisix/admin/init.lua:415>, context: ngx.timer
```

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -258,21 +258,17 @@ local function sync_local_conf_to_etcd(reset)
     local local_conf = core.config.local_conf()
 
     local plugins = {}
-    if local_conf.plugins then
-        for _, name in ipairs(local_conf.plugins) do
-            core.table.insert(plugins, {
-                name = name,
-            })
-        end
+    for _, name in ipairs(local_conf.plugins) do
+        core.table.insert(plugins, {
+            name = name,
+        })
     end
 
-    if local_conf.stream_plugins then
-        for _, name in ipairs(local_conf.stream_plugins) do
-            core.table.insert(plugins, {
-                name = name,
-                stream = true,
-            })
-        end
+    for _, name in ipairs(local_conf.stream_plugins) do
+        core.table.insert(plugins, {
+            name = name,
+            stream = true,
+        })
     end
 
     if reset then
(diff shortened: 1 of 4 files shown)


tokers (commented):
We may add a case to cover it.

apisix/admin/init.lua
```diff
@@ -258,17 +258,21 @@ local function sync_local_conf_to_etcd(reset)
     local local_conf = core.config.local_conf()
 
     local plugins = {}
-    for _, name in ipairs(local_conf.plugins) do
-        core.table.insert(plugins, {
-            name = name,
-        })
+    if local_conf.plugins then
```
spacewander, line 261: In what situation people will use an empty plugin list?

zhendongcmss: In most cases the list of http plugins is not empty, but the list of stream plugins is empty.

zhendongcmss: In the production environment we turned off all unused plugins.

apisix/admin/init.lua
```diff
@@ -258,17 +258,21 @@ local function sync_local_conf_to_etcd(reset)
...
+            })
+        end
     end
 
-    for _, name in ipairs(local_conf.stream_plugins) do
-        core.table.insert(plugins, {
-            name = name,
-            stream = true,
-        })
+    if local_conf.stream_plugins then
```
spacewander, line 269: Maybe we can add a default empty array in https://github.com/apache/apisix/blob/9129572f6302300ddeee3f2234393445d8ead781/apisix/cli/schema.lua#L247?

So that we don't need to introduce extra checks.

zhendongcmss: good idea

tokers (approved):

spacewander (approved):

PR #7935, opened 2022-09-18
feat: support configuring the order of the upstream dns service discovery resolve

monkeyDluffy6017:
…very resolve

### Description

Some users are having problems with the upstream DNS service discovery feature, because the current DNS resolution order is fixed:  1. "last" 2. "SRV" 3. "A" 4. "AAAA"  5. "CNAME".
"last" means the last successful resolving method; that is, if the last resolving is successful, the last resolving method will be used first.
This causes the problem that if for some reason the A record is resolved successfully, the higher priority SRV record will not be resolved in the future, and the user expects to get the result of resolving the SRV record.

As a user, I want to configure how and in what order the upstream DNS service discovery function resolves, and APISIX is able to resolve in the way and in the order I configure.

### Checklist


tokers (commented):
@monkeyDluffy6017 We should also update the https://apisix.apache.org/docs/apisix/discovery/dns/#service-discovery-via-dns.

apisix/discovery/dns/schema.lua
```diff
@@ -24,6 +24,18 @@ return {
...
             },
         },
+        order = {
+            type = "array",
+            minItems = 1,
+            maxItems = 5,
+            uniqueItems = true,
+            items = {
+                enum = {"last", "SRV", "A", "AAAA", "CNAME"},
+     --           ["not"] = {
```
spacewander, line 34: Just for debug?

t/discovery/dns/sanity.t
```diff
@@ -318,3 +318,24 @@ qr/upstream nodes: \{[^}]+\}/
...
+    - service_name: "srv-a.test.local"
+      discovery_type: dns
+      type: roundrobin
+      id: 1
+discovery:
+    dns:
+        servers:
+            - "127.0.0.1:1053"
+        order:
+            - A
```
spacewander, line 336: Is it possible to add some test about invalid input, like `a` or repeated record type?

apisix/discovery/dns/schema.lua
monkeyDluffy6017: Done

t/discovery/dns/sanity.t
monkeyDluffy6017: Done

t/discovery/dns/sanity.t
```diff
@@ -318,3 +326,115 @@ qr/upstream nodes: \{[^}]+\}/
...
+            - SRV
+--- apisix_yaml
+upstreams:
+    - service_name: "srv-a.test.local"
+      discovery_type: dns
+      type: roundrobin
+      id: 1
+--- must_die
+--- error_log
+matches none of the enum values
```
tzssangglass, line 440: Can we add a test case: DNS Server returns SRV records first for the same domain name, and the next query returns A records?

monkeyDluffy6017:
> 

Done

spacewander (approved):

membphis (approved):

tzssangglass (approved):

PR #7946, opened 2022-09-20
test: fix config error in test cases

monkeyDluffy6017:
### Description

Some yaml config in test cases are incorrect because we don't handle the result of schema validatation.

### Checklist


t/tars/discovery/tars.t
```diff
@@ -41,8 +41,8 @@ discovery:
       database: db_tars
       user: root
       password: tars2022
-    full_fetch_interval: 3
-    incremental_fetch_interval: 1
+    full_fetch_interval: 90
```
zhixiongdu027, line 44: Please see [link](https://github.com/apache/apisix/pull/6599#discussion_r827587517)
   >     "full_fetch_interval: 3"
   >     "incremental_fetch_interval: 1"
   I think we should keep it as it is

monkeyDluffy6017: The APISIX will fail before modify the schema, maybe we should figure out other way to reduce the interval.

monkeyDluffy6017:
```   
 -- fill the default value by the schema
    schema.validate(default_conf)
```
We didn't handle the result of schema validation before, that's why your solution work, we need to throw error now

zhixiongdu027: maybe we should find another way
@spacewander @monkeyDluffy6017

spacewander: > The APISIX will fail before modify the schema, maybe we should figure out another way to reduce the interval.

@monkeyDluffy6017 
Interesting. The modification is done in the init_by_lua phase. It is strange that the schema validation happens before it. Could you dig deeper into it?

monkeyDluffy6017: Sure, i will check this later

apisix/core/config_local.lua
```diff
@@ -65,7 +65,10 @@ function _M.local_conf(force)
     end
 
     -- fill the default value by the schema
-    schema.validate(default_conf)
+    local ok, err = schema.validate(default_conf)
+    if not ok then
+        ngx.log(ngx.ERR, "======================== levy, err: ", err)
```
tokers, line 70: levy?

monkeyDluffy6017: It will be changed to error(),  I just want to find out how many test cases will failed and fix them.

monkeyDluffy6017: I have changed this pr to draft.

t/APISIX.pm
```diff
@@ -502,6 +503,8 @@ _EOC_
 
     require "resty.core"
 
+    $extra_init_by_lua_start
```
spacewander, line 506: Could we just move $extra_init_by_lua to this place?

spacewander: Ping @monkeyDluffy6017

monkeyDluffy6017: Done

t/tars/discovery/tars.t
monkeyDluffy6017: The modification should be put before apisix = require("apisix")，the validation is called in that.

tzssangglass (approved):

monkeyDluffy6017:
test

apisix/core/config_local.lua
monkeyDluffy6017: Done

t/APISIX.pm
```diff
@@ -493,6 +493,11 @@ _EOC_
 
     $block->set_value("main_config", $main_config);
 
+    # The new directive is introduced here to modify the upper and
```
spacewander, line 496: We can modify the schema, not only the upper and lower limits

t/APISIX.pm
```diff
@@ -493,6 +493,11 @@ _EOC_
...
+    # lower limits of the schema before apisix validate in require("apisix")
+    # Todo: merge extra_init_by_lua_start and extra_init_by_lua
```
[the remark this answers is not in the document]
monkeyDluffy6017: https://github.com/apache/apisix/discussions/7988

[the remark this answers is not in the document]
monkeyDluffy6017: https://github.com/apache/apisix/discussions/7988

t/APISIX.pm
monkeyDluffy6017: Done

spacewander (approved):

tzssangglass (approved):

PR #7999, opened 2022-09-27
feat: add brokers field to support set broker with the same host in kafka-logger plugin

### LINKED ISSUE #7998: feat: As a user, I want to set more than one broker in kafka-logger plugin, so that we can push message to them at a time
### Description

Hello, folks,
When I use "kafka-logger" plugin in Apache APISIX. I found that we can't set up more than one broker with the same host.

The reason is that we use a table to store the broker list, but the key is the host (which will be the same if we deploy more Kafka on one node), so I think we should support it.

By the way, the docs are incorrect： https://apisix.apache.org/docs/apisix/next/plugins/kafka-logger/#enabling-the-plugin

```
"broker_list" :
  {
    "127.0.0.1":9092,
    "127.0.0.1":9093
  },
```

What I want is: 

```
"broker_list" :
[
    {
        "host":"127.0.0.1",
        "port":9092
    },
    {
        "host":"127.0.0.1",
        "port":9093
    }
]

```

starsz:
### Description

Fix #7998

Because the key in the JSON object should be different.
And the original design we use map to store the broker list, so we can't create a broker with the same host.

Now, I add a new filed `brokers` (It's an array type), then we can create brokers with the same host.

### Checklist


apisix/plugins/kafka-logger.lua
```diff
@@ -251,3 +284,4 @@ end
 
 
 return _M
+
```
tokers, line 287: Remove the empty line.

apisix/plugins/kafka-logger.lua
```diff
@@ -51,6 +52,26 @@ local schema = {
                 },
             },
         },
+        brokers = {
```
tokers, line 55: Please mark the `broker_list` in the schema as deprecated.

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -37,7 +37,10 @@ It might take some time to receive the log data. It will be automatically sent a
...
+| brokers                | array   | True     |                |                       | List of Kafka brokers (nodes).                                                                                                                                                                                                                                                                                                                   |
+| brokers.host           | string  | True     |                |                       | The host of Kafka broker                                                                                                                                                                                                                                                                                                                   |
```
tokers, line 42:
```suggestion
| brokers.host           | string  | True     |                |                       | The host of Kafka broker, e.g., `192.168.1.1`.                                                                                                                                                                                                                                                                                                                 |
```

t/plugin/kafka-logger2.t
```diff
@@ -340,6 +340,50 @@ qr/not found topic, retryable: true, topic: undefined_topic, partition_id: -1/
                         key= "key1",
                     },
                 },
+                {
+                    input = {
+                        brokers = {
+                        },
+                        kafka_topic = "test",
```
tokers, line 347: Add another test case that the `host` or the `port` field is missing.

apisix/plugins/kafka-logger.lua
starsz: Fixed.

apisix/plugins/kafka-logger.lua
starsz: Fixed.

docs/en/latest/plugins/kafka-logger.md
starsz: Fixed.

t/plugin/kafka-logger2.t
starsz: Fixed.

tokers (dismissed):

tzssangglass (dismissed):

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -37,7 +37,10 @@ It might take some time to receive the log data. It will be automatically sent a
...
-| broker_list            | object  | True     |                |                       | List of Kafka brokers (nodes).                                                                                                                                                                                                                                                                                                                   |
+| broker_list            | deprecated  | True |                |                       | Use `brokers` instead. List of Kafka brokers.  (nodes).                                                                                                                                                                                                                                                                                                                   |
```
spacewander, line 40: It would be better to put the "deprecated" in the Description instead of the Type?

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -37,7 +37,10 @@ It might take some time to receive the log data. It will be automatically sent a
...
+| brokers                | array   | True     |                |                       | List of Kafka brokers (nodes).                                                                                                                                                                                                                                                                                                                   |
+| brokers.host           | string  | True     |                |                       | The host of Kafka broker, e.g, `192.168.1.1`.                                                                                                                                                                                                                                                                                                                   |
```
spacewander, line 42: Do we need to update the Chinese version?
CC @SylviaBABY

apisix/plugins/kafka-logger.lua
```diff
@@ -199,15 +225,24 @@ function _M.log(conf, ctx)
     end
 
     -- reuse producer via lrucache to avoid unbalanced partitions of messages in kafka
-    local broker_list = core.table.new(core.table.nkeys(conf.broker_list), 0)
+    local length = conf.broker_list and core.table.nkeys(conf.broker_list) or #conf.brokers
```
spacewander, line 228: Just FYI: if we don't need to rewrite the content of table, we can use core.table.clone instead of `alloc + insert in loop`.

docs/en/latest/plugins/kafka-logger.md
starsz: Let me update the Chinese version.

apisix/plugins/kafka-logger.lua
starsz: Let me update it.

starsz: Thanks for the remind. Improved.

docs/en/latest/plugins/kafka-logger.md
starsz: Fixed.

docs/en/latest/plugins/kafka-logger.md
starsz: Fixed.

spacewander (approved):

SylviaBABY (approved):

Issue #7795, opened 2022-08-25
bug: the b3 header data should be consistently sent downstream in zipkin plugin


Gallardot:
### Current Behavior

Enabel zipkin plugin.

Set `sample_ratio` to not equal to 1, for example, 0.1. APISIX will not send B3 header data downstream when a request does not require sampling.

### Expected Behavior

Enabel zipkin plugin.

Set `sample_ratio` to not equal to 1, for example, 0.1. Even when a request does not require sampling, Apisix sends B3 header data downstream.

### Steps to Reproduce

ref: https://apisix.apache.org/docs/apisix/plugins/zipkin/

### Environment

- APISIX version (run `apisix version`):
- Operating system (run `uname -a`):
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

Gallardot:
ref: https://github.com/openzipkin/b3-propagation#sampling-state

```
Sampling is a mechanism to reduce the volume of data that ends up in the tracing system. 
In B3, sampling applies consistently per-trace: once the sampling decision is made, the same value should be consistently sent downstream. 
```

Even if the data is not sampled, the logic of the relevant phase should still be executed. But the final span data is not sent to the zipkin server.

Gallardot:
[opentelemetry plugin](https://apisix.apache.org/docs/apisix/plugins/opentelemetry/) has the same problem.

spacewander:
If I am correct, do you mean:
When there is a b3 header sent by the client, we should sample it regardless of the APISIX's configuration?
If so, why should not we report the span?

Gallardot:
> If I am correct, do you mean: When there is a b3 header sent by the client, we should sample it regardless of the APISIX's configuration? If so, why should not we report the span? As the behavior of "sample" is to report the span to the zipkin server.

Thanks @spacewander 
I mean if the http request is to be sampled, then we have to report span data to the zipkin Server. There's no problem with that. 
However, we should send `B3 header` data to `upstream` regardless of whether the http request needs to be sampled or not.

spacewander:
I got your point.
[...]

PR #8099, opened 2022-10-17
fix(zipkin): send trace IDs with a reject sampling decision

Gallardot:
### Description
https://github.com/apache/apisix/issues/7795

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/zipkin.lua
+++ b/apisix/plugins/zipkin.lua
@@ -100,6 +100,21 @@ local function parse_b3(b3)
     return nil, pieces[1], pieces[2], pieces[3], pieces[4]
 end
 
+local function inject_header(ctx)
+    local opentracing = ctx.opentracing
+    local tracer = opentracing.tracer
+    local outgoing_headers = {}
+
+    local span = opentracing.request_span
+    if ctx.opentracing_sample then
+        span = opentracing.proxy_span
+    end
+    tracer:inject(span, "http_headers", outgoing_headers)
+
+    for k, v in pairs(outgoing_headers) do
+        core.request.set_header(ctx, k, v)
+    end
+end
 
 function _M.rewrite(plugin_conf, ctx)
     local conf = core.table.clone(plugin_conf)
@@ -151,24 +166,6 @@ function _M.rewrite(plugin_conf, ctx)
         request_span_id = headers["x-b3-spanid"]
     end
 
-    if sampled == "1" or sampled == "true" then
-        per_req_sample_ratio = 1
-    elseif sampled == "0" or sampled == "false" then
-        per_req_sample_ratio = 0
-    end
-
-    ctx.opentracing_sample = tracer.sampler:sample(per_req_sample_ratio or conf.sample_ratio)
-    if not ctx.opentracing_sample then
-        core.request.set_header(ctx, "x-b3-sampled", "0")
-        -- pass the trace ids even the sample is rejected
-        -- see https://github.com/openzipkin/b3-propagation#why-send-
-        -- trace-ids-with-a-reject-sampling-decision
-        core.request.set_header(ctx, "x-b3-traceid", trace_id)
-        core.request.set_header(ctx, "x-b3-parentspanid", parent_span_id)
-        core.request.set_header(ctx, "x-b3-spanid", request_span_id)
-        return
-    end
-
     local zipkin_ctx = core.tablepool.fetch("zipkin_ctx", 0, 3)
     zipkin_ctx.trace_id = trace_id
     zipkin_ctx.parent_span_id = parent_span_id
@@ -198,11 +195,24 @@ function _M.rewrite(plugin_conf, ctx)
         request_span = request_span,
     }
 
+    -- Process sampled
+    if sampled == "1" or sampled == "true" then
+        per_req_sample_ratio = 1
+    elseif sampled == "0" or sampled == "false" then
+        per_req_sample_ratio = 0
+    end
+
+    ctx.opentracing_sample = tracer.sampler:sample(per_req_sample_ratio or conf.sample_ratio)
+    if not ctx.opentracing_sample then
+        request_span:set_baggage_item("x-b3-sampled","0")
+        return
+    end
+    request_span:set_baggage_item("x-b3-sampled","1")
+
     local request_span = ctx.opentracing.request_span
     if conf.span_version == ZIPKIN_SPAN_VER_1 then
         ctx.opentracing.rewrite_span = request_span:start_child_span("apisix.rewrite",
                                                                      start_timestamp)
-
         ctx.REWRITE_END_TIME = tracer:time()
         ctx.opentracing.rewrite_span:finish(ctx.REWRITE_END_TIME)
     else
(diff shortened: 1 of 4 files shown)


apisix/plugins/zipkin.lua
```diff
@@ -163,9 +165,9 @@ function _M.rewrite(plugin_conf, ctx)
         -- pass the trace ids even the sample is rejected
         -- see https://github.com/openzipkin/b3-propagation#why-send-
         -- trace-ids-with-a-reject-sampling-decision
-        core.request.set_header(ctx, "x-b3-traceid", trace_id)
-        core.request.set_header(ctx, "x-b3-parentspanid", parent_span_id)
-        core.request.set_header(ctx, "x-b3-spanid", request_span_id)
+        core.request.set_header(ctx, "x-b3-traceid", trace_id or to_hex(rand_bytes(16)))
```
spacewander, line 168: I am confused. Why do we generate a trace ID when the sampling is rejected?

spacewander: After reading the OpenTracing material, I can understand the reason.

The current implementation of APISIX is incomplete. For example, we don't have the ability to generate a root span. We require the client to pass a trace ID & span ID. Maybe we should add the ability to generate trace ID/span ID if not given instead of just generating a trace ID when the sampling is rejected

spacewander: Yes, I am wrong about the root span.

After some investigation, I figure out that:
Currently, APISIX handles root span like this:

1. check if need to sample or not
2. if not, set some upstream header and go to the next plugin
3. generate root span

The correct behavior should be:
1. check if need to sample or not
2. generate root span
3. set some upstream header (as the sampling should only affect the tracing data report)

In step 2 of APISIX, the root span is missing so we have to create a fake span ID in this PR.

Maybe we need to do some bigger changes to solve this problem?

For example, update
https://github.com/apache/apisix/blob/490e0ba995fb0bfe4c1b22eec424f58ebef39446/apisix/plugins/zipkin/codec.lua#L101
https://github.com/apache/apisix/blob/490e0ba995fb0bfe4c1b22eec424f58ebef39446/apisix/plugins/zipkin/reporter.lua#L58
to make it work correctly.

Otherwise, the generated fake span in this PR doesn't have a real span associated with it.

Gallardot: I'm sorry, I didn't get your point. @spacewander 

Why do we need to generate a real root span when no sampling is required and the client does not pass a trace ID & span ID? Where does this real root span need to be used when no sampling is required?  We do not need to report these spans to the zipkin server.

A fake trace ID & span ID are sufficient. Only `upstream headers` and `apisix's logs` need these ids

spacewander: There is no standard way to do sampling in zipkin. But by reading the code of Kong / Envoy / Traefik, they both choose the "correct steps" we described in https://github.com/apache/apisix/pull/8099#discussion_r999020747.

The benefit of this way is that we don't need to duplicate header set operations in both sampling and no sampling path, especially when we need to introduce more flags in the future. The fake span id solution is not so popular.

apisix/plugins/zipkin/codec.lua
```diff
@@ -97,11 +97,13 @@ local function new_injector()
...
-        -- when we call this function, we already start to sample
-        headers["x-b3-sampled"] = "1"
+        headers["x-b3-sampled"] = span_context:get_baggage_item("x-b3-sampled")
         for key, value in span_context:each_baggage_item() do
-            -- XXX: https://github.com/opentracing/specification/issues/117
-            headers["uberctx-"..key] = ngx.escape_uri(value)
+            -- skip x-b3-sampled baggage
+            if key ~= "x-b3-sampled" then
+                -- XXX: https://github.com/opentracing/specification/issues/117
+                headers["uberctx-"..key] = ngx.escape_uri(value)
```
spacewander, line 105: Why does the uberctx depend on "x-b3-sampled"?

spacewander (commented):
Let's update the test file to make CI pass.

apisix/plugins/zipkin/codec.lua
Gallardot: The uberctx does not depend on "x-b3-sampled".

ref:  https://github.com/apache/apisix/pull/8099/files#diff-d7b683eff2eae613bb495064140483bd2d223010b83563842f57433f1501cc64R205-R210

baggage is used to pass  "x-b3-sampled", so we need to skip this item. Or any suggestions?  Looking forward to your reply. @spacewander 

ps: I will fix the CI later

spacewander: @Gallardot 
Sorry, it is my mistake, I didn't realize it. The current solution is good enough to me.

Gallardot: @spacewander I have no idea why the CI about Kafka is fail. Can you help to run it again? PTAL.

Gallardot:
> Let's update the test file to make CI pass.

@spacewander All CI pass. PTAL.

t/plugin/zipkin2.t
```diff
@@ -126,8 +126,9 @@ b3: 80f198ee56343ba864fe8b2a57d3eff7-e457b5a2e4d86bd1-0-05e3ac9a4f6e3b90
 --- response_headers
 x-b3-sampled: 0
 x-b3-traceid: 80f198ee56343ba864fe8b2a57d3eff7
-x-b3-parentspanid: 05e3ac9a4f6e3b90
-x-b3-spanid: e457b5a2e4d86bd1
+x-b3-parentspanid: e457b5a2e4d86bd1
+--- raw_response_headers_unlike
+x-b3-spanid:
```
spacewander, line 131: It can pass even without x-b3-spanid.
I will recommend:
```
--- raw_response_headers_like
.*x-b3-spanid:.*
```

t/plugin/zipkin2.t
```diff
@@ -136,6 +137,7 @@ x-b3-spanid: e457b5a2e4d86bd1
 b3: 0
 --- response_headers
 x-b3-sampled: 0
+--- raw_response_headers_unlike
```
spacewander, line 140: Let's use `raw_response_headers_like` too

t/plugin/zipkin2.t
```diff
@@ -126,8 +126,9 @@ b3: 80f198ee56343ba864fe8b2a57d3eff7-e457b5a2e4d86bd1-0-05e3ac9a4f6e3b90
 --- response_headers
 x-b3-sampled: 0
 x-b3-traceid: 80f198ee56343ba864fe8b2a57d3eff7
-x-b3-parentspanid: 05e3ac9a4f6e3b90
-x-b3-spanid: e457b5a2e4d86bd1
+x-b3-parentspanid: e457b5a2e4d86bd1
+--- raw_response_headers_like:
```
spacewander, line 130: It should be:
```
--- raw_response_headers_like
.*x-b3-spanid:.*
```

Gallardot: @spacewander  Sorry, I'm not familiar with this testing framework. I need some help. 
The expected test result is that the `x-b3-spanid`  header exists and is not empty.  Because `x-b3-spanid` header's value is a random value.

May be look like this  ?
```
--- raw_response_headers_like: x-b3-spanid: .*
```

t/plugin/zipkin2.t
Gallardot: done

spacewander (approved):

t/plugin/zipkin2.t
spacewander: LGTM

membphis (approved):

PR #8113, opened 2022-10-18
feat(ai): add dynamic route cache key

kingluo:
### Description

add dynamic route cache key

### Checklist

### DIFF
--- a/apisix/core/ai.lua
+++ b/apisix/core/ai.lua
@@ -20,6 +20,7 @@ local template        = require("resty.template")
 local ipairs          = ipairs
 local pcall           = pcall
 local assert          = assert
+local loadstring      = loadstring
 
 local get_cache_key_func
 
@@ -98,9 +99,10 @@ function  _M.routes_analyze(routes)
         core.log.info("use ai plane to match route")
         router.match = ai_match
         local str = get_cache_key_func_def_render({route_flags = route_flags})
-        local ok, func = pcall(str)
+        local func = loadstring(str)
+        assert(func)
+        local ok, get_cache_key_func = pcall(func)
         assert(ok)
-        get_cache_key_func = func()
     end
 end
 
--- a/apisix/core/ai.lua
+++ b/apisix/core/ai.lua
@@ -75,15 +75,17 @@ local function gen_get_cache_key_func(route_flags)
     local str = get_cache_key_func_def_render({route_flags = route_flags})
     local func, err = loadstring(str)
     if func == nil then
-        return err
+        return false, err
     else
         local ok
         ok, get_cache_key_func = pcall(func)
         if not ok then
             local err = get_cache_key_func
-            return err
+            return false, err
         end
     end
+
+    return true
 end
 
 
@@ -119,8 +121,8 @@ function  _M.routes_analyze(routes)
         core.log.info("use ai plane to match route")
         router.match = ai_match
 
-        local err = gen_get_cache_key_func(route_flags)
-        if err then
+        local ok, err = gen_get_cache_key_func(route_flags)
+        if not ok then
             core.log.error("generate get_cache_key_func failed:", err)
             router.match = orig_router_match
         end
--- a/apisix/core/ai.lua
+++ b/apisix/core/ai.lua
@@ -26,12 +26,13 @@ local get_cache_key_func_def_render
 
 local get_cache_key_func_def = [[
 return function(ctx)
-    return ctx.var.uri
+    local var = ctx.var
+    return var.uri
         {% if route_flags["methods"] then %}
-        .. "\0" .. ctx.var.method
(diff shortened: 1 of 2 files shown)


apisix/core/ai.lua
```diff
@@ -16,8 +16,27 @@
...
 local ipairs          = ipairs
 
+local get_cache_key_func
+
+local get_cache_key_func_def_render = template.compile([[
+return function(ctx)
+    local key = ctx.var.uri
+
+    {% if route_flags["methods"] then %}
+    key = key .. "\0" .. ctx.var.method
```
membphis, line 31: pls change `"\0"` to a visible characters

apisix/core/ai.lua
```diff
@@ -16,8 +16,27 @@
...
+    {% end %}
+
+    {% if route_flags["host"] then %}
+    key = key .. "\1" .. ctx.var.host
```
membphis, line 35: the second code block is better performance

```lua
local a = "ccc"
a = a .. "bbb"
a = a .. "ccc"
```

```lua
local a = "aaa" .. "bbb" .. "ccc"
```

membphis (changes requested):

apisix/core/ai.lua
```diff
@@ -71,6 +97,26 @@ function  _M.routes_analyze(routes)
     else
         core.log.info("use ai plane to match route")
         router.match = ai_match
+
+        if get_cache_key_func_def_render == nil then
+            local template = require("resty.template")
```
membphis, line 102: can we do this `template` at top level?

membphis (commented):
pls add some test cases to check it works fine

apisix/core/ai.lua
```diff
@@ -71,6 +118,12 @@ function  _M.routes_analyze(routes)
     else
         core.log.info("use ai plane to match route")
         router.match = ai_match
+
+        local err = gen_get_cache_key_func(route_flags)
```
membphis, line 122: ditto

membphis (changes requested):

apisix/core/ai.lua
```diff
@@ -48,10 +66,41 @@ local function ai_match(ctx)
...
+        get_cache_key_func_def_render = template.compile(get_cache_key_func_def)
+    end
+
+    local str = get_cache_key_func_def_render({route_flags = route_flags})
+    local func, err = loadstring(str)
+    if func == nil then
+        return false, err
+    else
+        local ok
+        ok, get_cache_key_func = pcall(func)
```
spacewander, line 81: Better to use a temp var instead of hijacking a module var to hold the potential err message

apisix/core/ai.lua
```diff
@@ -17,6 +17,24 @@
...
+local pcall           = pcall
+local loadstring      = loadstring
+local encode_base64   = ngx.encode_base64
+
+local get_cache_key_func
+local get_cache_key_func_def_render
+
+local get_cache_key_func_def = [[
+return function(ctx)
+    return ctx.var.uri
```
spacewander, line 29: We can make it a little faster, by using `local var = ctx.var` to reduce table lookup.

t/core/ai.t
```diff
@@ -474,3 +474,177 @@ qr/use ai plane to match route/
...
+
+            if code >= 300 then
+                ngx.status = code
+                ngx.say(body)
+                return
+            end
+            ngx.sleep(1)
+
+            -- enable
+            do
```
spacewander, line 626: Let's exact this common part into a function, instead of repeating it three times.

membphis (approved):

spacewander (approved):

apisix/core/ai.lua
```diff
@@ -36,8 +55,8 @@ end
 
 
 local function ai_match(ctx)
-    -- TODO: we need to generate cache key dynamically
-    local key = ctx.var.uri .. "-" .. ctx.var.method .. "-" .. ctx.var.host
+    local key = get_cache_key_func(ctx)
+    core.log.info("route cache key: ", core.log.delay_exec(encode_base64, key))
```
membphis, line 59: @kingluo after my benchmark, we have to remove this line.

so we have to find a new way for testing.

apisix/core/ai.lua
```diff
@@ -17,6 +17,25 @@
...
+
+local get_cache_key_func
+local get_cache_key_func_def_render
+
+local get_cache_key_func_def = [[
+return function(ctx)
+    local var = ctx.var
+    return var.uri
+        {% if route_flags["methods"] then %}
+        .. "\0" .. var.method
```
membphis, line 32: use `#` , should be enough

membphis: the `uri`, `host` and `method`, they should generally not contain `#`

PR #8119, opened 2022-10-19
fix: renew route lrucache when the routes change

kingluo:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -45,9 +45,7 @@ return function(ctx)
 end
 ]]
 
-local route_lrucache = core.lrucache.new({
-    count = 512
-})
+local route_lrucache
 
 local schema = {}
 
@@ -76,8 +74,7 @@ end
 local function ai_match(ctx)
     local key = get_cache_key_func(ctx)
     core.log.info("route cache key: ", key)
-    local ver = router.router_http.user_routes.conf_version
-    local route_cache = route_lrucache(key, ver,
+    local route_cache = route_lrucache(key, nil,
                                        match_route, ctx)
     -- if the version has not changed, use the cached route
     if route_cache then
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -46,8 +46,6 @@ end
 ]]
 
 local route_lrucache = core.lrucache.new({
-    -- TODO: we need to set the cache size by count of routes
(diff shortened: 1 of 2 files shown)


membphis:
@kingluo pls rebase your branch

membphis (dismissed):

apisix/plugins/ai.lua
```diff
@@ -132,6 +132,12 @@ local function routes_analyze(routes)
         core.log.info("use ai plane to match route")
         router.router_http.match = ai_match
 
+        local count = #routes + 3000
```
spacewander, line 135: I am not familiar with the route cache design, so the questions below need to be answered by someone who is familiar with it. CC @membphis @tzssangglass 

1. do we need to clean the cache when we switch 'cache' to 'no cache'?
2. is it desirable to recreate cache once the router is rebuilt (a new route is added)?

spacewander: If it is desirable, we need to switch the cache lookup way from `key + ver` to `key` as each ver has its own cache now. Also, we may not need to init the cache as it may be created in the first router building.

tzssangglass: > do we need to clean the cache when we switch 'cache' to 'no cache'?

I don't think it's needed because 'cache' to 'no cache' means `router.router_http.user_routes.conf_version` has changed, refer to the explanation above.

Do we need to recreate the lrucache when switching 'cache' to 'no cache'? @membphis

membphis: for the better performance and logic simple, I think we should renew the lrucache and ignore `router.router_http.user_routes.conf_version`, which is more simpler to read and understanding

spacewander: > I don't think it's needed because 'cache' to 'no cache' means router.router_http.user_routes.conf_version has changed, refer to the explanation above.

So it is acceptable to keep the unused cache when we move to 'no cache'?

BTW, look like we can remove the init in https://github.com/apache/apisix/blob/3887162bc387b89c121f1ccbd823848a388cd6f5/apisix/plugins/ai.lua#L48 as there will always create a router cache here?

tzssangglass: > So it is acceptable to keep the unused cache when we move to 'no cache'?

I think it's needed.

membphis (approved):

PR #8130, opened 2022-10-20
perf: simple setup upstream

tzssangglass:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -21,11 +21,12 @@ local router          = require("apisix.router")
 local event           = require("apisix.core.event")
 local load_balancer   = require("apisix.balancer")
 local balancer        = require("ngx.balancer")
+local is_http         = ngx.config.subsystem == "http"
+local enable_keepalive = balancer.enable_keepalive and is_http
 local ipairs          = ipairs
 local pcall           = pcall
 local loadstring      = loadstring
 local type            = type
-local encode_base64   = ngx.encode_base64
 
 local get_cache_key_func
 local get_cache_key_func_def_render
@@ -115,13 +116,17 @@ end
 local pool_opt = { pool_size = 320 }
 local function ai_balancer_run(route)
     local server = route.value.upstream.nodes[1]
-    local ok, err = balancer.set_current_peer(server.host, server.port or 80, pool_opt)
-    if not ok then
-        core.log.error("failed to set server peer [", server.host, ":",
-                       server.port, "] err: ", err)
-        return ok, err
+    if enable_keepalive then
+        local ok, err = balancer.set_current_peer(server.host, server.port or 80, pool_opt)
+        if not ok then
+            core.log.error("failed to set server peer [", server.host, ":",
+                           server.port, "] err: ", err)
+            return ok, err
+        end
+        balancer.enable_keepalive(60, 1000)
+    else
+        balancer.set_current_peer(server.host, server.port)
     end
-    balancer.enable_keepalive(60, 1000)
 end
 
 local function routes_analyze(routes)
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -221,10 +221,8 @@ local function routes_analyze(routes)
             and not route_up_flags["timeout"]
             and not route_up_flags["keepalive"] then
             -- replace the upstream module
-        ngx.log(ngx.WARN, "replace the upstream module")
         apisix.handle_upstream = ai_upstream
     else
-        ngx.log(ngx.WARN, "origin the upstream module")
         apisix.handle_upstream = orig_handle_upstream
     end
 end
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -15,6 +15,7 @@
 -- limitations under the License.
 --
 local require         = require
+local apisix          = require("apisix")
 local core            = require("apisix.core")
 local router          = require("apisix.router")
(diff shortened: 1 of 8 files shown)


t/debug/dynamic-hook.t
```diff
@@ -405,7 +405,8 @@ hook_test:                      # module and function list, name: hook_test
                     "uri": "/hello",
                     "upstream": {
                         "nodes": {
-                            "127.0.0.1:1980": 1
+                            "127.0.0.1:1980": 1,
```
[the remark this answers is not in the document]
tzssangglass: done

membphis (dismissed):

apisix/plugins/ai.lua
```diff
@@ -100,33 +108,120 @@ local function gen_get_cache_key_func(route_flags)
...
+            return ok, err
+        end
+        balancer.enable_keepalive(60, 1000)
+    else
+        balancer.set_current_peer(server.host, server.port or 80)
+    end
+end
+
 local function routes_analyze(routes)
     -- TODO: need to add a option in config.yaml to enable this feature(default is true)
```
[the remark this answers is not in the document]
tzssangglass: next PR fix it.

apisix/plugins/ai.lua
```diff
@@ -100,33 +108,120 @@ local function gen_get_cache_key_func(route_flags)
...
+local function ai_balancer_run(route)
+    local server = route.value.upstream.nodes[1]
+    if enable_keepalive then
+        local ok, err = balancer.set_current_peer(server.host, server.port or 80, pool_opt)
+        if not ok then
+            core.log.error("failed to set server peer [", server.host, ":",
+                           server.port, "] err: ", err)
+            return ok, err
+        end
+        balancer.enable_keepalive(60, 1000)
```
spacewander, line 126: Why don't we use the keepalive configured in the yaml?

apisix/plugins/ai.lua
```diff
@@ -100,33 +108,120 @@ local function gen_get_cache_key_func(route_flags)
...
+    else
+        balancer.set_current_peer(server.host, server.port or 80)
+    end
```
spacewander, line 129: We can move:
```
if enable_keepalive then
```
to here?
Like:
```
if enable_keepalive then
    balancer.enable_keepalive(60, 1000)
```

apisix/plugins/ai.lua
```diff
@@ -138,6 +233,27 @@ local function routes_analyze(routes)
...
+            and not route_flags["upstream_id"]
+            and not route_flags["enable_websocket"]
+            and not route_flags["plugins"]
+            and not route_up_flags["has_domain"]
+            and route_up_flags["pass_host"]
+            and route_up_flags["scheme"]
+            and not route_up_flags["checks"]
+            and not route_up_flags["retries"]
+            and not route_up_flags["timeout"]
+            and not route_up_flags["timeout"]
```
spacewander, line 248: Repeated condition?

apisix/plugins/ai.lua
```diff
@@ -138,6 +233,27 @@ local function routes_analyze(routes)
...
+            and not route_up_flags["keepalive"] then
```
spacewander, line 249: Is it possible to use allow list instead of a pile of deny list?

apisix/plugins/ai.lua
```diff
@@ -100,33 +108,120 @@ local function gen_get_cache_key_func(route_flags)
...
 local function routes_analyze(routes)
     -- TODO: need to add a option in config.yaml to enable this feature(default is true)
-    local route_flags = core.table.new(0, 2)
+    local route_flags = core.table.new(0, 5)
+    local route_up_flags = core.table.new(0, 8)
     for _, route in ipairs(routes) do
-        if route.methods then
-            route_flags["methods"] = true
-        end
+        if type(route) == "table" then
```
membphis, line 137: we can change it to a better style, which is simpler:

```lua
local route_flags = {}

for _, route in ipairs(routes) do
    for key, value in pairs(route.value) do
        -- collect route flags
        if key == "host" or key = "hosts" then
            route_flags["host"] = true
        elseif key == "remote_addr" or key = "remote_addrs" then
            route_flags["remote_addr"] = true
        else
            route_flags[key] = true
        end

        -- ... ...
    end

end
```

apisix/plugins/ai.lua
membphis: We should enable `keepalive` by default

tzssangglass: done

apisix/plugins/ai.lua
tzssangglass: fixed

apisix/plugins/ai.lua
tzssangglass: changed

apisix/plugins/ai.lua
tzssangglass: maybe not, follow: https://github.com/apache/apisix/blob/ecdc209523f011b5fb4a0fee3a8e867da9c53fb5/apisix/balancer.lua#L291-L338

if `enable_keepalive`, would add `pool_opt` as the third argument to `balancer.set_current_peer`.

Do we fix the call to `set_current_peer` here and in `balancer.lua`, pass `pool_opt` even if there is disable `enable_keepalive`, otherwise it will all become non-persistent connection.

membphis (approved):

spacewander (approved):

PR #8149, opened 2022-10-23
perf: allow skip body filter

tzssangglass:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/ai.lua
+++ b/apisix/plugins/ai.lua
@@ -19,10 +19,11 @@ local apisix          = require("apisix")
 local core            = require("apisix.core")
 local router          = require("apisix.router")
 local event           = require("apisix.core.event")
-local load_balancer   = require("apisix.balancer")
 local balancer        = require("ngx.balancer")
+local ngx             = ngx
 local is_http         = ngx.config.subsystem == "http"
 local enable_keepalive = balancer.enable_keepalive and is_http
+local is_apisix_or, response = pcall(require, "resty.apisix.response")
 local ipairs          = ipairs
 local pcall           = pcall
 local loadstring      = loadstring
@@ -65,7 +66,7 @@ local _M = {
 
 local orig_router_match
 local orig_handle_upstream = apisix.handle_upstream
-local orig_balancer_run = load_balancer.run
+local orig_http_balancer_phase = apisix.http_balancer_phase
 
 local default_keepalive_pool = {}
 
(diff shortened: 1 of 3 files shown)


apisix/plugins/ai.lua
```diff
@@ -19,10 +19,11 @@ local apisix          = require("apisix")
 local core            = require("apisix.core")
 local router          = require("apisix.router")
 local event           = require("apisix.core.event")
-local load_balancer   = require("apisix.balancer")
 local balancer        = require("ngx.balancer")
+local ngx             = ngx
 local is_http         = ngx.config.subsystem == "http"
 local enable_keepalive = balancer.enable_keepalive and is_http
+local is_apisix_or, response = pcall(require, "resty.apisix.response")
```
membphis, line 26: this should be a core API

apisix/plugins/ai.lua
```diff
@@ -116,7 +117,26 @@ end
...
-local function ai_balancer_run(route)
+local function ai_http_balancer_phase()
+    local api_ctx = ngx.ctx.api_ctx
+    if not api_ctx then
+        core.log.error("invalid api_ctx")
+        return core.response.exit(500)
+    end
+
+    if is_apisix_or then
+        local ok, err = response.skip_header_filter_by_lua()
```
membphis, line 128: Only the current request skips the ***_filter, all right?

apisix/plugins/ai.lua
membphis: what do you thing? @spacewander

tzssangglass: You mean move `pcall(require, "resty.apisix.response")` on apisix core level?

apisix/plugins/ai.lua
tzssangglass: no, it effects global

apisix/plugins/ai.lua
```diff
@@ -116,7 +117,21 @@ end
...
-local function ai_balancer_run(route)
+local function ai_http_balancer_phase()
+    local api_ctx = ngx.ctx.api_ctx
+    if not api_ctx then
+        core.log.error("invalid api_ctx")
+        return core.response.exit(500)
+    end
+
+    if is_apisix_or then
+        local ok, err = response.skip_body_filter_by_lua()
```
spacewander, line 128: No skip_header_filter_by_lua?

apisix/plugins/ai.lua
spacewander: It is possible but not valuable to do this, as we don't want to support multiple versions of APISIX-Base, only support the latest APISIX-Base. Maybe a simple API to detect if apisix-base / openresty is used is enough.

apisix/plugins/ai.lua
tzssangglass: cancel skip_header_filter_by_lua, because we add `APISIX/2.99`in the header_filterphase, and `enable_server_tokens = false` cannot be used as a flag to skip header_filter_by_lua because it only removes the version in server token, not delete the server token unless we add another switch for removing the server token completely.

membphis (approved):

t/debug/dynamic-hook.t
```diff
@@ -30,6 +30,11 @@ run_tests();
 __DATA__
 
 === TEST 1: dynamic enable
+# ai module would conflict with the debug module
+--- extra_yaml_config
+plugins:
+    #ai
```
spacewander, line 36:
```suggestion
    #- ai
```

t/debug/dynamic-hook.t
```diff
@@ -86,6 +91,11 @@ call require("apisix").http_access_phase() args:{}
 
 
 === TEST 2: dynamic enable by per request and disable after handle request
+# ai module would conflict with the debug module
+--- extra_yaml_config
+plugins:
+    #ai
```
spacewander, line 97: Ditto

spacewander (dismissed):

t/debug/dynamic-hook.t
tzssangglass: fixed

membphis (approved):

spacewander (approved):

PR #8159, opened 2022-10-24
feat(prometheus): extra_labels of prometheus plugin_attr support const

### LINKED ISSUE #8146: help request: what extra_labels of prometheus plugin i can use?
### Description

Hi, i want to add some labels to metrics, and some labels are literals instead of vars. I find it will be an empty string when i use literals.  
Besides, i cannot find how to define vars like $upstream_addr. 
```yaml
plugin_attr:
  prometheus:
    metrics:
        http_status:
            extra_labels:
                - upstream_addr: $upstream_addr
                - upstream_status: $upstream_status

```
Any reply would be greatly appreciated :)

### Environment

xinxinh2020:
### Description

Fixes #8146 (issue)

### Checklist

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -80,7 +80,7 @@ local function extra_labels(name, ctx)
         for _, kv in ipairs(labels) do
             local val, v = next(kv)
             if ctx then
-                if v:find("$",1, true) == 1 then
+                if v:byte(1) == string.byte("$") then
                     val = ctx.var[v:sub(2)]
                     if val == nil then
                         val = ""
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -80,7 +80,7 @@ local function extra_labels(name, ctx)
         for _, kv in ipairs(labels) do
             local val, v = next(kv)
             if ctx then
-                if v:find("$",1, true) then
+                if v:find("$",1, true) == 1 then
                     val = ctx.var[v:sub(2)]
                     if val == nil then
                         val = ""
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -80,9 +80,14 @@ local function extra_labels(name, ctx)
         for _, kv in ipairs(labels) do
             local val, v = next(kv)
             if ctx then
-                val = ctx.var[v:sub(2)]
-                if val == nil then
-                    val = ""
+                if v:find("$",1, true) then
+                    val = ctx.var[v:sub(2)]
+                    if val == nil then
+                        val = ""
+                    end
+                else
+                    -- if label value is a literal const, use the literal const directly
+                    val = v
                 end
             end
             core.table.insert(extra_labels_tbl, val)
--- a/docs/en/latest/plugins/prometheus.md
+++ b/docs/en/latest/plugins/prometheus.md
@@ -59,6 +59,8 @@ For http request related metrics, you could specify extra labels, which match th
 
 If you specify label for nonexist APISIX variable, the label value would be "".
 
+If you specify label not started with "$", the label value would be a constant value.
+
 Currently, only below metrics are supported:
 
 * http_status
--- a/t/plugin/prometheus5.t
+++ b/t/plugin/prometheus5.t
@@ -1,0 +1,120 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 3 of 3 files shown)


tzssangglass:
Why support const? What scenarios require monitoring a constant value?

xinxinh2020:
> Why support const? What scenarios require monitoring a constant value?

In my case, i need to identify which apisix cluster the metrics come from。and every apisix clusters，which are managed by  k8s operator,  have it own labels(such as which k8s CR it belongs to). I think it would be useful to tag these labels in metrics to make us more easy to query metrics of specific custom labels.

tzssangglass:
I don't think it's a good idea to support constants on `extra_labels`, we have metrics like `apisix_node_info`, maybe we should extend it to support more constants.

xinxinh2020:
> I don't think it's a good idea to support constants on `extra_labels`, we have metrics like `apisix_node_info`, maybe we should extend it to support more constants.

As far as i am concerned, `apisix_node_info` is used to show all of the apisix info,  and the `apisix_node_info` is not associated with other metrics like `http_status`. 
Maybe we can add a variable like $node_info_id, so we can refer the `apisix_node_info` in other metrics. 
But i still argue that adding apisix-cluster-level const in some metrics is useful. For example, we have four apisix clusters with lables:
- cluster1: deploy=vm,  provider=aws, type=public
- cluster2: deploy=k8s, provider=aws, type=private
- cluster3: deploy=k8s, provider=azure, type=private
- cluster4: deploy=vm, provider=azure, type=public

Even with extended `apisix_node_info`, how can i query metrics group by  any combination of these labels, such as provider=aws, or deploy=vm&type=public ?

If we tag these labels in metrics, it would be quite convenient and a promql like `apisix_http_status{provider="aws"}` would be enough.

tzssangglass:
> Even with extended `apisix_node_info`, how can i query metrics group by any combination of these labels, such as provider=aws, or deploy=vm&type=public ?

Understood, then you need to add test cases and update the documentation.

xinxinh2020:
> 
That's great. 
And how to add a test case in apisix? Any example ?

xinxinh2020:
> > Even with extended `apisix_node_info`, how can i query metrics group by any combination of these labels, such as provider=aws, or deploy=vm&type=public ?
> 
> Understood, then you need to add test cases and update the documentation.

That's great.
And how to add a test case in apisix? Any example ?

tzssangglass:
> That's great.
> And how to add a test case in apisix? Any example ?

ref: https://github.com/apache/apisix/blob/master/docs/en/latest/plugin-develop.md#write-test-case

xinxinh2020:
>

xinxinh2020:
> > That's great.
> > And how to add a test case in apisix? Any example ?
> 
> ref: https://github.com/apache/apisix/blob/master/docs/en/latest/plugin-develop.md#write-test-case

Test case & doc are provided. Pls review it :)

apisix/plugins/prometheus/exporter.lua
```diff
@@ -80,9 +80,14 @@ local function extra_labels(name, ctx)
         for _, kv in ipairs(labels) do
             local val, v = next(kv)
             if ctx then
-                val = ctx.var[v:sub(2)]
-                if val == nil then
-                    val = ""
+                if v:find("$",1, true) then
```
spacewander, line 83: Will "a$b" match?

xinxinh2020: not, "a$b" will be considered as a const now

spacewander: @xinxinh2020 
This branch will match `a$b`.

spacewander:
``` 
￥ cat /tmp/x.lua
#!/usr/bin/env luajit

local v = "a$b"
print(v:find("$",1, true))

 ￥ /tmp/x.lua
2       2
```

xinxinh2020: > @xinxinh2020 This branch will match `a$b`.

Thinks for reminding.  I fix it already.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -80,9 +80,14 @@ local function extra_labels(name, ctx)
         for _, kv in ipairs(labels) do
             local val, v = next(kv)
             if ctx then
-                val = ctx.var[v:sub(2)]
-                if val == nil then
-                    val = ""
+                if v:find("$",1, true) == 1 then
```
spacewander, line 83: Would be better to use string.byte which is O(1).

xinxinh2020: good advice & work done

Issue #8087, opened 2022-10-14
bug: Invalid filter with response-rewrite in apache/apisix:2.99.0-debian


chyoong:
### Current Behavior

Hi, I would also like to use nginx sub_filter in apisix, but it still doesn't work when I use it in apache/apisix:2.99.0-debian，and I see an identical issue and have been merged，there https://github.com/apache/apisix/issues/5451

### Expected Behavior

Before content:
```html
<head>
    <title>test</title>
    <meta charset="utf-8">
</head>
```

ideally, use the content after `filter`:
```html
<head>
    <title>test</title>
    <meta charset="utf-8">
    <script>console.log(1)</script>
</head>
```

However, it does not take effect

### Steps to Reproduce

1. Use docker-compose deployment
2. In the `plugin/market` path, enable the `response-rewrite` plugin
3. configuration parameters are as follows
```json
{
  "disable": false,
  "filters": [
    {
      "regex": "</head>",
      "replace": "<script>console.log(1)</script></head>",
      "scope": "global"
    }
  ],
  "headers": {
    "set": {
      "x-powered-by": "",
      "x-server-balancer_addr": "$balancer_ip:$balancer_port"
    }
  },
  "vars": [
    [
      "status",
      "==",
      200
    ],
    [
      "content-type",
      "==",
      "text/html; charset=utf-8"
    ]
  ]
}
```

### Environment

- APISIX version (run `apisix version`): apache/apisix:2.99.0-debian
- APISIX Dashboard version, if relevant: apache/apisix-dashboard:2.13-alpine

soulbird:
same as #7852

mscb402:
Please read document https://apisix.apache.org/docs/apisix/terminology/plugin/#plugin-common-configuration 
Currently apisix only support request parameters, so your configuration is not right. You should remove condition on vars
```
"vars": [
    [
      "status",
      "==",
      200
    ],
    [
      "content-type",
      "==",
      "text/html; charset=utf-8"
    ]
  ]
```
to 
```
"var": []
```

soulbird:
```
<head>
    <title>test</title>
    <meta charset="utf-8">
</head>
```
Can you give a full upstream response? Contains response headers and a response body.

soulbird:
> t request parameters,

The response-rewrite plugin has already received the response when executing the `vars` expression, so it can contain some variables in the response, you can check this test case: https://github.com/apache/apisix/blob/master/ t/plugin/response-rewrite2.t#L236

[...]

PR #8162, opened 2022-10-25
fix: meta.filter in plugin doesn't work in response-rewrite plugin

### LINKED ISSUE #7852: bug: meta.filter in plugin doesn't work when we use some variable
### Current Behavior

When I use meta.filter in the plugin like this:

```
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '{
  "plugins": {
    "file-logger": {
         "_meta": {
            "filter": [
                ["upstream_status", "~=", 200]
            ]
        },
      "path": "logs/file.log"
    }
  },
  "upstream": {
    "type": "roundrobin",
    "nodes": {
      "httpbin.org": 1
    }
  },
  "uri": "/*"
}'
```

What I expect is that when the upstream_status is not equal to 200, then we will write a log into the file.
But now, the plugin will be executed even if the upstream_status is 200.

Reason:
Because when we do the plugin filter, we can't get the upstream_status (because we hadn't sent a request to upstream), so the value of **`ngx.var.upstream_status` is nil**, and the filter is always matched.

https://github.com/apache/apisix/blob/47187faccd96d02773fb1c71980c8ca0ea838d6a/apisix/plugin.lua#L421-L458

### Expected Behavior

The meta.filter can work even if the plugin is run at the log phase.

### Error Logs

_No response_

### Steps to Reproduce

1. Create route

```
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '{
  "plugins": {
    "file-logger": {
         "_meta": {
            "filter": [
                ["upstream_status", "~=", 200]
            ]
        },
      "path": "logs/file.log"
    }
  },
  "upstream": {
    "type": "roundrobin",
    "nodes": {
      "httpbin.org": 1
    }
  },
  "uri": "/*"
}'
```

2. Send two requests

```
curl localhost:9080/status/200 -v
```

```
curl localhost:9080/status/400 -v
```

3. We can see both requests are logged.
image

### Environment

soulbird:
### Description

Fixes #7852

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -415,7 +415,8 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
         return true
     end
 
-    local match_cache_key = ctx.conf_type .. plugin_name .. "meta_filter_matched"
+    local match_cache_key =
+        ctx.conf_type .. "#" .. plugin_name .. "#meta_filter_matched"
     if ctx[match_cache_key] ~= nil then
         return ctx[match_cache_key]
     end
@@ -924,8 +925,6 @@ function _M.run_plugin(phase, plugins, api_ctx)
     if not plugins or #plugins == 0 then
         return api_ctx
     end
-
-    --local matched = meta_filter(ctx, name, plugin_conf)
 
     if phase ~= "log"
         and phase ~= "header_filter"
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -415,6 +415,11 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
         return true
     end
 
+    local match_cache_key = ctx.conf_type .. plugin_name .. "meta_filter_matched"
+    if ctx[match_cache_key] ~= nil then
+        return ctx[match_cache_key]
+    end
+
     local ex, ok, err
     if ctx then
         ex, err = expr_lrucache(plugin_name .. ctx.conf_type .. ctx.conf_id,
@@ -433,6 +438,8 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
                          " plugin_name: ", plugin_name)
         return true
     end
+
+    ctx[match_cache_key] = ok
     return ok
 end
 
@@ -457,9 +464,9 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
             goto continue
         end
 
-        local matched = meta_filter(ctx, name, plugin_conf)
+        --local matched = meta_filter(ctx, name, plugin_conf)
         local disable = check_disable(plugin_conf)
-        if not disable and matched then
+        if not disable --[[and matched]] then
             if plugin_obj.run_policy == "prefer_route" and route_plugin_conf ~= nil then
                 local plugin_conf_in_route = route_plugin_conf[name]
                 local disable_in_route = check_disable(plugin_conf_in_route)
@@ -920,6 +927,8 @@ function _M.run_plugin(phase, plugins, api_ctx)
         return api_ctx
     end
 
+    --local matched = meta_filter(ctx, name, plugin_conf)
+
     if phase ~= "log"
         and phase ~= "header_filter"
         and phase ~= "body_filter"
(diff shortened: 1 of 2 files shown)


apisix/plugin.lua
```diff
@@ -415,6 +415,11 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
         return true
     end
 
+    local match_cache_key = ctx.conf_type .. plugin_name .. "meta_filter_matched"
```
tokers, line 418: Use some separator like `#` before `meta_filter_matched` ?

apisix/plugin.lua
```diff
@@ -920,6 +925,8 @@ function _M.run_plugin(phase, plugins, api_ctx)
         return api_ctx
     end
 
+    --local matched = meta_filter(ctx, name, plugin_conf)
```
tokers, line 928: Remove this?

t/plugin/plugin.t
```diff
@@ -525,3 +525,170 @@ passed
...
+--- response_headers
+test-header: error
+--- error_code: 502
+
+
+
+=== TEST 21: upstream_status = 200, disable response-rewrite plugin
+--- request
+GET /hello
+
```
tokers, line 588: Should check if the `x-test-upstream-status` doesn't exist.

apisix/plugin.lua
```diff
@@ -433,6 +438,8 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
                          " plugin_name: ", plugin_name)
         return true
     end
+
+    ctx[match_cache_key] = ok
```
starsz, line 442: I think we can't cache the result of the filter. Since we have a different phase for every plugin like "rewrite", "access" e.g.

t/plugin/plugin.t
soulbird: It doesn't make sense to check this, it's better to check if `test-header` exists.

apisix/plugin.lua
soulbird: `filter` works better at the plugin level and is executed at each phase, which will increase the difficulty for users to use.

apisix/plugin.lua
```diff
@@ -415,6 +415,12 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
         return true
     end
 
+    local match_cache_key =
+        ctx.conf_type .. "#" .. plugin_name .. "#meta_filter_matched"
```
spacewander, line 419: Does this key work with multiple global_rules? They have the same conf_type.

soulbird: Do you mean the scene where the `con_type` is the same and the `plugin_name` is different? I think my test case already contains. The `conf_type` is `route`.

spacewander: @soulbird 
Multiple global rules will share the same conf_type
https://github.com/apache/apisix/blob/e2d0f0b16bb157412684a38b129d12acb4628d39/apisix/plugin.lua#L1003

Will this affect the cache of eval result?

soulbird: There is indeed a problem, I adjusted `match_cache_key`, added some test case：https://github.com/apache/apisix/pull/8162/files#diff-29e2373df60fd8fc082317d17e636b6fa17c60459c1a691c5838f36b31cc486aR532-R567

t/plugin/plugin.t
```diff
@@ -28,6 +28,7 @@ add_block_preprocessor(sub {
     $block;
 });
 
+repeat_each(1);
```
spacewander, line 31: https://github.com/apache/apisix/blob/e523132f35fe2eb7e1d06990d037f18f89c9dcba/t/APISIX.pm#L23
We already set it by default?

spacewander (approved):

spacewander:
@soulbird 
> Fixes https://github.com/apache/apisix/issues/7852 https://github.com/apache/apisix/issues/8087

Need to use double fixes

soulbird:
> @soulbird
> 
> > Fixes #7852 #8087
> 
> Need to use double fixes

My mistake, these two don't seem to be the same problem.

t/plugin/plugin.t
```diff
@@ -520,8 +520,220 @@ passed
...
+=== TEST 25: proxy-rewrite plugin will set $http_foo_age, response-rewrite plugin return 403
+--- request
+GET /hello?age=18
+--- error_code: 403
+
+
+
+=== TEST 26: response-rewrite plugin disable, return 200
+--- request
+GET /hello
```
starsz, line 689: Missing any check ？

soulbird: By default, the 200 status code will be checked, so there is no need to repeatedly write.

starsz (approved):

PR #8233, opened 2022-11-02
chore(fmt): stylua

fundon:
### Description

Use [`stylua`](https://github.com/JohnnyMorganz/StyLua) to format codes.

```shell
stylua -g '*.lua' -- .
```

Some projects use it.

* [neovim](https://github.com/neovim/neovim)

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ t/lib/dubbo-backend/dubbo-backend-provider/target/
 !.markdownlint.yml
 !.licenserc.yaml
 !.ignore_words
+!.stylua.toml
 conf/apisix.yaml
 conf/apisix-*.yaml
 conf/config-*.yaml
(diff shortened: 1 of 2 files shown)


tokers:
Those changes will break the git blame result. I'm afraid the benefits are small.

fundon:
I think it is possible to unify the style in a progressive way.

When a particular developer makes changes to a file, the file can be formatted last.

```shell
stylua a.lua
```

spacewander (changes requested):
I am sorry that we can't accept a change which will break the git blame result

spacewander:
Enforcing a code style in Lua code is not a good idea as there is no standard in the Lua world. And we don't want to force every developer to use an X formatter.

fundon:
我想主要问题还是会破坏 git blame 结果，对提交者非常不友好，另外是历史遗留问题，语言没内置 fmt 工具。

但如果有一定的代码规范的约定的话，对项目还是有一点点助理。:)

PR #8254, opened 2022-11-06
test: merge extra_init_by_lua_start and extra_init_by_lua directive in test cases.

### LINKED ISSUE #7990: Merge `extra_init_by_lua_start ` and `extra_init_by_lua ` directive in test cases
### Discussed in https://github.com/apache/apisix/discussions/7988

<div type='discussions-op-text'>

<sup>Originally posted by **monkeyDluffy6017** September 26, 2022</sup>
Maybe we should merge `extra_init_by_lua_start ` and `extra_init_by_lua ` directives in test cases, they are just different on position.</div>

mikechengwei:
### Description

Fixes #7990.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ENV_NGINX_EXEC         := $(shell command -v openresty 2>/dev/null || command -v
 ENV_OPENSSL_PREFIX     ?= $(addprefix $(ENV_NGINX_PREFIX), openssl)
 ENV_LUAROCKS           ?= luarocks
 ## These variables can be injected by luarocks
-ENV_INST_PREFIX        ?= /usr
+ENV_INST_PREFIX        ?= /Users/mikechengwei/apisix
 ENV_INST_LUADIR        ?= $(ENV_INST_PREFIX)/share/lua/5.1
 ENV_INST_BINDIR        ?= $(ENV_INST_PREFIX)/bin
 ENV_HOMEBREW_PREFIX    ?= /usr/local
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ENV_NGINX_EXEC         := $(shell command -v openresty 2>/dev/null || command -v
 ENV_OPENSSL_PREFIX     ?= $(addprefix $(ENV_NGINX_PREFIX), openssl)
 ENV_LUAROCKS           ?= luarocks
 ## These variables can be injected by luarocks
-ENV_INST_PREFIX        ?= /Users/mikechengwei/apisix
+ENV_INST_PREFIX        ?= /usr
 ENV_INST_LUADIR        ?= $(ENV_INST_PREFIX)/share/lua/5.1
 ENV_INST_BINDIR        ?= $(ENV_INST_PREFIX)/bin
 ENV_HOMEBREW_PREFIX    ?= /usr/local
--- a/t/APISIX.pm
+++ b/t/APISIX.pm
@@ -19,6 +19,8 @@ package t::APISIX;
 use lib 'lib';
 use Cwd qw(cwd);
 use Test::Nginx::Socket::Lua::Stream -Base;
+use File::Spec::Functions qw(rel2abs);
+use strict;
 
 repeat_each(1);
 log_level('info');
@@ -504,7 +506,8 @@ _EOC_
 
     require "resty.core"
 _EOC_
-    if (defined $block->extra_init_before_apisix_init) {
+    my $calling_sub = (caller(1))[1];
+    if (index($calling_sub, "tars") != -1) {
         $init_by_lua_block .= <<_EOC_;
     $extra_init_by_lua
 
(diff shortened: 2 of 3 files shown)


tokers (approved):

t/APISIX.pm
```diff
@@ -495,8 +495,6 @@ _EOC_
 
     # The new directive is introduced here to modify the schema
     # before apisix validate in require("apisix")
-    # Todo: merge extra_init_by_lua_start and extra_init_by_lua
```
spacewander, line 498: We need to change extra_init_by_lua so we don't need a separate extra_init_by_lua_start.
Otherwise the CI will fail: https://github.com/apache/apisix/actions/runs/3404934882/

mikechengwei: Do you mean the `extra_init_by_lua` function should include the extra_init_by_lua_start module?
This commit is my one attempt.

spacewander (commented):
Please make the CI pass first.

mikechengwei:
> Please make the CI pass first.

plz help re-run ci.

t/APISIX.pm
```diff
@@ -506,8 +503,10 @@ _EOC_
     end
 
     require "resty.core"
-
-    $extra_init_by_lua_start
+_EOC_
+    if (index($extra_init_by_lua, "apisix.discovery.tars.schema") != -1) {
```
spacewander, line 507: Switching the extra_init_by_lua position according to its content is not a good idea. If we hit the problem with some other module like `apisix.blah.blah`, it is too hack to extend the checklist.

We should dig deeper and solve the real problem.

mikechengwei: Ok,let me think sometime.

mikechengwei: I find tars test $extra_init_by_lua need to execute before apisix init, so i add a '$extra_init_before_apisix_init' token identify it. WDYT?

mikechengwei:
@spacewander plz help re-run workflow.

Makefile
```diff
@@ -42,7 +42,7 @@ ENV_NGINX_EXEC         := $(shell command -v openresty 2>/dev/null || command -v
 ENV_OPENSSL_PREFIX     ?= $(addprefix $(ENV_NGINX_PREFIX), openssl)
 ENV_LUAROCKS           ?= luarocks
 ## These variables can be injected by luarocks
-ENV_INST_PREFIX        ?= /usr
+ENV_INST_PREFIX        ?= /Users/mikechengwei/apisix
```
tzssangglass, line 45: don't use your private path

tzssangglass (changes requested):

Makefile
mikechengwei: thanks.

t/tars/discovery/tars.t
```diff
@@ -63,7 +63,8 @@ _EOC_
         schema.properties.full_fetch_interval.default = 3
 _EOC_
 
-    $block->set_value("extra_init_by_lua_start", $extra_init_by_lua_start);
+    $block->set_value("extra_init_by_lua", $extra_init_by_lua);
+    $block->set_value("extra_init_before_apisix_init");
```
spacewander, line 67: Could we modify the content in `$init_by_lua_block` (for example, reorder the statements) to make sure `extra_init_by_lua` work in this file?

spacewander: So that we don't need to introduce another flag like extra_init_by_lua_start or extra_init_before_apisix_init

mikechengwei: ok,let me try.

mikechengwei:
Now I use `caller` to get the calling file to judge the execution order, WDYT? @spacewander

t/APISIX.pm
```diff
@@ -506,8 +503,11 @@ _EOC_
     end
 
     require "resty.core"
-
-    $extra_init_by_lua_start
+_EOC_
+    my $calling_sub = (caller(1))[1];
+    if (index($calling_sub, "tars") != -1) {
```
tzssangglass, line 508: It looks like we didn't actually remove the `extra_init_by_lua_start` mount point. It's just determining if it's a tars module in the APISIX.pm file and mounting it.

If there are other test cases that we want to mount in this location in the future, we need to add it in APISIX.pm?

mikechengwei: IIUC the code, tars module `extra_init_by_lua` code need to load before APISIX init ,other modules `extra_init_by_lua` code  need to load after APISIX init.If we must remove `extra_init_by_lua_start`, we need to determine the execution order of `extra_init_by_lua` in a way similar to judging the module name.This way of displaying declarations in APISIX.pm may be more explicit.

spacewander: Actually, I don't think this is better than the previously rejected solution in https://github.com/apache/apisix/pull/8254#discussion_r1017453415...

PR #8256, opened 2022-11-07
fix: APISIX extended variables cannot be used in `_meta.filter`

### LINKED ISSUE #8251: bug: built-in variables are not supported in `_meta.filter`
### Current Behavior

The filters in the plugin support only Nginx variables, not APISIX's built-in variables.

For example, I can use the Nginx variable `http_xxx`, but I can't use APISIX custom variables like `arg_xxxx` or `graphql_xxxx`.

### Expected Behavior

It can support all APISIX variable syntax, including `cookie_xxxx`, `arg_xxxx`, `post_arg_xxxx`, `graphql_xxxx`, etc.

### Error Logs

_No response_

### Steps to Reproduce

1. Set the filter in the plugin to use the above unsupported syntax.

-------------

**The reason is that we didn't pass in our ctx for that expr executor, and the expr uses its default `ngx.var`, which wouldn't support extended syntax.**

https://github.com/apache/apisix/blob/master/apisix/plugin.lua#L437
https://github.com/api7/lua-resty-expr/blob/main/lib/resty/expr/v1.lua#L376

We just need to change it to `ok, err = ex:eval(ctx.var)` and it will support those extended syntaxes.

### Environment

- APISIX version (run `apisix version`): 3.0.0

soulbird:
### Description

Fixes #8251 

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -434,7 +434,7 @@ local function meta_filter(ctx, plugin_name, plugin_conf)
                          " plugin_name: ", plugin_name)
         return true
     end
-    ok, err = ex:eval()
+    ok, err = ex:eval(ctx.var)
     if err then
         core.log.warn("failed to run the 'vars' expression: ", err,
                          " plugin_name: ", plugin_name)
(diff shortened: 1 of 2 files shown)


bzp2010 (approved):
Maybe there is some typo in the PR title. 🤔

PR #8270, opened 2022-11-07
feat: add method that parse ip from /etc/hosts in high priority

### LINKED ISSUE #5233: request help: no host allowed while connecting to upstream
### Issue description

从nacos上发现服务.
- nacos上的服务ip为主机名`uaa-service`, 一个为ip
- 当负载均衡运行时报 `failed to set server peer [uaa-service:20101] err: no host allowed while connecting to upstream`
- 造成当负载到为主机名的时候返回网关错误. 负载到ip的时候则正常.

### Environment

- OS (cmd: `uname -a`):

TenYearsIn:
### Description

Fixes #5233 

### Checklist


apisix/core/resolver.lua
```diff
@@ -19,15 +19,67 @@
...
+
+
+local function init_hosts_ip()
+    if not DEFAULT_HOSTS then
+        return
+    end
+
+    local file = open(DEFAULT_HOSTS)
+    if not file then
+        return
```
tokers, line 56: It's better to print the error reason here.

soulbird:
Need test cases also.

apisix/core/resolver.lua
```diff
@@ -19,15 +19,67 @@
...
+
+    if name:find(":") then
+        return "ipv6"
+    end
+
+    return "hostname"
+end
+
+
+local function init_hosts_ip()
```
spacewander, line 49: Could we use the method from https://github.com/Kong/lua-resty-dns-client/blob/fb1e686a8857c7492fdd30ca228d200afbde19aa/src/resty/dns/client.lua#L493

TenYearsIn: Yeah,  I will replace it

TenYearsIn:
please approve running workflows @spacewander

TenYearsIn:
I analyze the error test cases reported by the CI, but I find that It has nothing to do with the code I commit, what I should do

spacewander:
The tests fail because there is no "dns resolver domain: xxx to yyy" in the error.log, as resolving host in /etc/hosts doesn't go through the DNS client. Maybe you can add a similar log to let test pass.

TenYearsIn:
> The tests fail because there is no "dns resolver domain: xxx to yyy" in the error.log, as resolving host in /etc/hosts doesn't go through the DNS client. Maybe you can add a similar log to let test pass.

Thx,  How can i trigger ci locally.  I want to know the cause .

spacewander:
> > The tests fail because there is no "dns resolver domain: xxx to yyy" in the error.log, as resolving host in /etc/hosts doesn't go through the DNS client. Maybe you can add a similar log to let test pass.
> 
> Thx, How can i trigger ci locally. I want to know the cause .

You can submit the PR to your fork, where you have the right to rerun it if you want.

TenYearsIn:
I analyze the failed test cases reported by the CI And I Find out that some test cases ignore can't match this feature. So I modify some test-case code . Please review my thoughts @spacewander

apisix/core/resolver.lua
```diff
@@ -42,6 +58,21 @@ end
 -- @usage
 -- local ip, err = core.resolver.parse_domain("apache.org") -- "198.18.10.114"
 function _M.parse_domain(host)
+    local rev = HOSTS_IP_MATCH_CACHE[host]
+    if rev then
+        -- use ipv4 in high priority
+        local ip = rev["ipv4"]
+        if not ip then
+            ip = rev["ipv6"]
```
spacewander, line 66: We need to consider this setting:
https://github.com/apache/apisix/blob/2f4a4dba2c0cbc426ae99133c2546112ea71dba8/conf/config-default.yaml#L35

t/core/resolver.t
```diff
@@ -0,0 +1,90 @@
...
+            local ip_info, err = resolver.parse_domain(domain)
+            if not ip_info then
+                core.log.error("failed to parse domain: ", domain, ", error: ",err)
+                return
+            end
+            ngx.say("ip_info: ", require("toolkit.json").encode(ip_info))
+        }
+    }
+--- request
+GET /t
```
spacewander, line 44: We can set the common sections at the top of this file like
https://github.com/apache/apisix/blob/2f4a4dba2c0cbc426ae99133c2546112ea71dba8/t/core/ctx2.t#L23

So we don't need to repeat it in each test.

t/core/resolver.t
```diff
@@ -0,0 +1,90 @@
...
+--- response_body
+ip_info: "151.101.2.132"
```
spacewander, line 70: Better to use injected code to check it.
The machine running apisix.apache.org is not owned by us and its IP may change.

t/node/upstream-node-dns.t
```diff
@@ -67,7 +67,7 @@ passed
 
     local utils = require("apisix.core.utils")
     utils.dns_parse = function (domain)  -- mock: DNS parser
-        if domain == "test.com" then
+        if domain == "test2.com" then
```
spacewander, line 70: Look like this test now will resolve with `test.com` and return "127.0.0.2" always.
Maybe we can change `test.com` to `test1.com` so we can bypass the /etc/hosts in the CI env.

spacewander: I mean replacing `test.com` in the whole test file.

apisix/core/resolver.lua
TenYearsIn: I thought of this setting before, But I think when the default value of this setting is true , We also use ipv4 .  So  I think the  ip's mapping that in /etc/hosts decides, but I improve ipv4's priority

t/core/resolver.t
TenYearsIn: OK,  Thanks !

apisix/core/resolver.lua
```diff
@@ -19,15 +19,31 @@
...
 
 local _M = {}
 
 
+local function init_hosts_ip()
+    local hosts, err = dns_utils.parseHosts()
+    if not hosts then
+        return hosts, err
+    end
+    HOSTS_IP_MATCH_CACHE = hosts
```
tzssangglass, line 39: When will `HOSTS_IP_MATCH_CACHE` be cleaned? APISIX will not update it after it is read at startup. Unless APISIX is restarted.

TenYearsIn: This kind of configuration is basically unchanged, so I didn't consider reloading logic at that time

apisix/core/resolver.lua
spacewander: When the users disable ipv6, we should not return the IPv6 address in /etc/hosts to them.

t/core/resolver.t
```diff
@@ -0,0 +1,97 @@
...
+no_long_string();
+no_root_location();
+log_level("info");
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if (!$block->request) {
+        $block->set_value("request", "GET /t");
+    }
```
spacewander, line 29: Let's enable the no_error_log by default too:
https://github.com/apache/apisix/blob/ae400b95b91240fb867c3cedc9fa65a6b5593f36/t/core/ctx2.t#L34

TenYearsIn: Please Review ~ @spacewander

spacewander (approved):

tzssangglass (approved):

PR #8298, opened 2022-11-10
feat: support openssl fips

kingluo:
### Description

Test apisix with openssl 3.0 (enable-fips)

### Checklist


.github/workflows/fips.yml
```diff
@@ -0,0 +1,130 @@
...
+      fail-fast: false
+      matrix:
+        platform:
+          - ubuntu-20.04
+        os_name:
+          - linux_openresty
+        test_dir:
+          - t/plugin/jwt-auth2.t t/plugin/hmac-auth2.t t/plugin/redirect.t t/plugin/http-logger.t t/plugin/clickhouse-logger.t t/plugin/openid-connect.t t/plugin/jwt-auth.t t/plugin/google-cloud-logging.t t/plugin/hmac-auth.t t/plugin/prometheus.t t/plugin/authz-keycloak.t t/plugin/batch-requests.t t/plugin/elasticsearch-logger.t
+          - t/admin/ssl* t/admin/schema.t t/admin/upstream.t t/config-center-yaml/ssl.t t/core/etcd-mtls.t t/core/config_etcd.t t/deployment/conf_server.t t/misc/patch.t
+          - t/node/grpc-proxy-unary.t t/node/upstream-keepalive-pool.t t/node/upstream-websocket.t t/node/client-mtls.t t/node/upstream-mtls.t t/pubsub/kafka.t t/router/radixtree-sni2.t t/router/multi-ssl-certs.t t/router/radixtree-sni.t t/stream-node/mtls.t t/stream-node/tls.t t/stream-node/upstream-tls.t t/stream-node/sni.t
```
spacewander, line 30: I wonder why we only run a subset of test, and is there a standard to determine what tests need to be run?

t/plugin/google-cloud-logging.t
```diff
@@ -313,6 +313,8 @@ qr/\{\"error\"\:\"[\w+\s+]*\"\}/
 --- error_log
 Batch Processor[google-cloud-logging] failed to process entries
 Batch Processor[google-cloud-logging] exceeded the max_retry_count
+--- skip_openssl
```
spacewander, line 316: We need to add comment about why this test is skipped under OpenSSL 3.0. If possible, we can modify the test to make it pass under different OpenSSL versions.

.github/workflows/fips.yml
```diff
@@ -0,0 +1,130 @@
...
+          path: ~/openssl-3.0
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ matrix.os_name }}
+
+      - name: Toggle openssl compile
+        id: test_ssl_env
+        shell: bash
+        if: steps.cache-openssl.outputs.cache-hit != 'true'
+        run: |
+          echo "##[set-output name=openssl3;]$(echo 'yes')"
+          echo "##[set-output name=openssl3_prefix;]$HOME"
```
soulbird, line 67: See: https://github.com/apache/apisix/pull/8259

.github/workflows/fips.yml
kingluo: Because only these tests involves ssl, and according to @membphis, we only need to cover a subset of test files to minimize the test duration.

t/plugin/google-cloud-logging.t
kingluo: openssl 3.0 with fips enabled requires keys larger than 2048 bits. All tests skipped for openssl 3.0 use small key.

.github/workflows/fips.yml
spacewander: @membphis 
Could you confirm this?
What if we need to move the test files, or add a new test that also involves SSL? Do we need to sync this list too?

t/plugin/google-cloud-logging.t
kingluo: Those tests just aims for small keys.

spacewander: Are `test route (no access to this scopes)` in google-cloud-logging.t and `use lifetime_grace_period default value` in jwt-auth2.t (and other similar tests) tests for small keys?

spacewander: We can give a new key in those tests.

t/admin/ssl4.t
```diff
@@ -55,8 +56,23 @@ location /t {
 
             local sess, err = sock:sslhandshake(nil, "www.test.com", true)
             if not sess then
-                ngx.say("failed to do SSL handshake: ", err)
-                return
+                sock = ngx.socket.tcp()
```
spacewander, line 59: Why do we need a retry for FIPS?
BTW, we can exact the connect operation into a function, instead of repeating it.

kingluo: In openssl 3.0, decrypt error would be saved in global, which is different to openssl 1.1.
The test case provides two decrypt keys, the first is wrong one, it test the retries and success finally.

https://github.com/apache/apisix/blob/master/t/admin/ssl4.t#L238

```
=== TEST 6: client request successfully, use the two key_encrypt_salt to decrypt in turn
--- yaml_config
apisix:
    node_listen: 1984
    ssl:
        key_encrypt_salt:
            - edd1c9f0985e76a2
            - edd1c9f0985e76a1
```

Then there comes an awkward problem.

The cosocket tries to do ssl hanshake:

https://github.dev/apache/apisix/blob/master/t/admin/ssl4.t#L56

```
            local sess, err = sock:sslhandshake(nil, "www.test.com", true)
            if not sess then
                ngx.say("failed to do SSL handshake: ", err)
                return
            end
```

In the server side of the same worker process, it failed to decrypt using the first key, which would save the error in the `SSL` object.

https://github.com/apache/apisix/blob/master/apisix/ssl.lua#L173

```lua
local function parse_pem_priv_key(sni, pkey)
    core.log.debug("parsing priv key for sni: ", sni)

    local parsed, err = ngx_ssl.parse_pem_priv_key(aes_decrypt_pkey(pkey))
    return parsed, err
end


function _M.fetch_pkey(sni, pkey)
    local parsed_pkey, err = pkey_cache(pkey, nil, parse_pem_priv_key, sni, pkey)
    if not parsed_pkey then
        return false, err
    end

    return parsed_pkey
end
```

Then when the client side handshake returns, it would get error, but in fact, the handshake is successful, it jsut gets the decrypt error saved before.

`src/event/ngx_event_openssl.c`

```c
n = SSL_do_handshake(c->ssl->connection);
...
sslerr = SSL_get_error(c->ssl->connection, n);
```

When you do second handshake, because the server side reused the decrypted key in the cache, so no problem then.

All in all, it's triggered by the combination of cosocket and server reside in the same worker process, openssl changes and nginx ssl handshake handling.

.github/workflows/fips.yml
membphis: I think we can reduce the test cases and reduce the difficulty of maintenance.
In addition, some notes need to be added.

t/plugin/authz-keycloak.t
```diff
@@ -314,7 +314,7 @@ GET /t
 --- response_body
 false
 --- error_log
-Error while sending authz request to https://127.0.0.1:8443/auth/realms/University/protocol/openid-connect/token: 18: self signed certificate
+Error while sending authz request to https://127.0.0.1:8443/auth/realms/University/protocol/openid-connect/token: 18
```
moonming, line 317: why change this?

t/plugin/jwt-auth.t
```diff
@@ -572,6 +572,8 @@ property "key" is required
     }
 --- response_body
 passed
+--- skip_eval
+1: $ENV{OPENSSL_FIPS} eq 'yes'
```
moonming, line 576: Why add this?

t/pubsub/kafka.t
```diff
@@ -330,8 +330,8 @@ all brokers failed in fetch topic metadata
             test_pubsub:close_ws()
         }
     }
---- error_log
-self signed certificate
+--- error_log eval
+qr/self[- ]signed certificate/
```
moonming, line 334: I don't understand why the change

t/router/multi-ssl-certs.t
```diff
@@ -239,9 +239,9 @@ location /t {
 }
 --- request
 GET /t
---- response_body
+--- response_body_like
 connected: 1
-failed to do SSL handshake: 18: self signed certificate
+failed to do SSL handshake: 18: self[- ]signed certificate
```
moonming, line 244: ditto

t/plugin/authz-keycloak.t
kingluo: Openssl 3.0 changes the error log content, so we must adapt it.

t/plugin/jwt-auth.t
kingluo: Openssl 3.0 with FIPS enabled does not support keys smaller than 2048bit.

t/pubsub/kafka.t
kingluo: ditto

t/router/multi-ssl-certs.t
kingluo: ditto

t/plugin/authz-keycloak.t
kingluo: @moonming I already manged to run and pass all tests related to SSL. But according to @membphis , I need to limit the test cases.

moonming: > 

What is the error message of OpenSSL 3.0?

kingluo: In most cases, self assgined cert reject would be "18: self-signed certificate"

moonming: Understood. So we can be compatible with it not removed it.

.github/workflows/fips.yml
kingluo: @moonming In fact, all tests related to SSL usage are included in this list.  So the question is, whether we need to slim this list?

t/admin/ssl4.t
kingluo: > we can exact the connect operation into a function, instead of repeating it.

It's trivial and unnecessary to do this. Only connect and handshake functions are called twice on demand, and we need to control the error response comes from the second try.

spacewander: I see.

.github/workflows/fips.yml
membphis: I think you have replied in your comments, all right?

https://github.com/apache/apisix/pull/8298/files#diff-5d1da2612215b3509570907291b3f68aa763efd84b326277239ca6312bd73155R28

t/plugin/authz-keycloak.t
membphis: agree with @moonming

membphis:
where is the doc?
How can APISIX users use FIPS?

.github/workflows/fips.yml
kingluo: > I think you have replied in your comments, all right?
> 
> https://github.com/apache/apisix/pull/8298/files#diff-5d1da2612215b3509570907291b3f68aa763efd84b326277239ca6312bd73155R28

Yes, I did. Please read the comment in fips.yml.

docs/en/latest/support-fips-in-apisix.md
```diff
@@ -0,0 +1,54 @@
...
+
+export cc_opt="-I$OPENSSL3_PREFIX/openssl-3.0/include"
+export ld_opt="-L$OPENSSL3_PREFIX/openssl-3.0/lib64 -Wl,-rpath,$OPENSSL3_PREFIX/openssl-3.0/lib64"
+
+wget https://raw.githubusercontent.com/api7/apisix-build-tools/master/build-apisix-base.sh
+chmod +x build-apisix-base.sh
+./build-apisix-base.sh latest
+```
+
+The apisix-base would be installed into `/usr/local/openresty-debug`.
```
soulbird, line 54: Are these operation commands included in the `install-linux-openresty.sh` script? I think this needs to be more concise, a lot of command line looks unfriendly

kingluo: Currently, in my opinion, because we don't plan to touch build-tool,  it's the only way to reuse the install script. If you need more clean way, please give solution directly.
If in later future, we could apply it into build-tool, then no need to do tricy and temporary stuff here.

docs/en/latest/support-fips-in-apisix.md
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+OpenSSL 3.0 [supports](https://www.openssl.org/blog/blog/2022/08/24/FIPS-validation-certificate-issued/)
+[FIPS](https://en.wikipedia.org/wiki/FIPS_140-2) mode. So to support
+FIPS in APISIX, we just need to compile apisix-base with OpenSSL 3.0.
+
+## Compilation
```
spacewander, line 28: This doc is named "Support FIPS in APISIX", but most of its content is only about compiling APISIX-Base.

@juzhiyuan @navendu-pottekkat 
Could you help to review this doc as the role of APISIX user? Thanks!

navendu-pottekkat: The title is correct. Even if we are compiling APISIX, the end goal is to support FIPS. So it is correct. Please also add YAML front matter after the title. Refer to other documentation pages.

docs/en/latest/support-fips-in-apisix.md
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+OpenSSL 3.0 [supports](https://www.openssl.org/blog/blog/2022/08/24/FIPS-validation-certificate-issued/)
+[FIPS](https://en.wikipedia.org/wiki/FIPS_140-2) mode. So to support
+FIPS in APISIX, we just need to compile apisix-base with OpenSSL 3.0.
```
navendu-pottekkat, line 26:
```suggestion
OpenSSL 3.0 [supports](https://www.openssl.org/blog/blog/2022/08/24/FIPS-validation-certificate-issued/) [FIPS](https://en.wikipedia.org/wiki/FIPS_140-2) mode. To support FIPS in APISIX, you can compile apisix-base with OpenSSL 3.0.
```

docs/en/latest/support-fips-in-apisix.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Compilation
+
+```bash
+# should be as root user
+
```
navendu-pottekkat, line 32: This comment could be replaced by text before the code block:

"To compile apisix-base with OpenSSL 3.0, run the commands below as root user:"

docs/en/latest/support-fips-in-apisix.md
navendu-pottekkat, line 54:
```suggestion
This will install apisix-base to `/usr/local/openresty-debug`.
```

docs/en/latest/support-fips-in-apisix.md
kingluo: > The title is correct. Even if we are compiling APISIX, the end goal is to support FIPS. So it is correct. Please also add YAML front matter after the title. Refer to other documentation pages.

Most docs only contains title. So what extra fields I need to add? Could you give a doc for reference?

navendu-pottekkat: See https://github.com/apache/apisix/blob/ae400b95b91240fb867c3cedc9fa65a6b5593f36/docs/en/latest/building-apisix.md?plain=1#L4

I'm trying to add these to all pages. So if we do it from the start, it would be easier.

.github/workflows/fips.yml
```diff
@@ -0,0 +1,135 @@
+name: CI FIPS
+
+on:
+  push:
+    branches: [master]
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+  pull_request:
```
soulbird, line 9: I don't think we need to run this CI every PR. Can be triggered when a new version is released according to a specific PR title, such as:  `feat:release xxx`

kingluo: @spacewander @membphis So do I need to change it according to @soulbird ?

navendu-pottekkat (commented):
LGTM on docs.

PR #8336, opened 2022-11-16
feat: proxy-rewrite support config add set and remove header

### LINKED ISSUE #8239: feat: Support adding header in proxy-rewrite plugin
### Description

Currently, proxy-rewrite plugin only supports set / delete headers. The limitation makes some operations impossible. This feature is required to implement [Gateway API HTTPHeaderFilter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderFilter).

* Similar to #7794 

## Expect
```json
{
    "proxy-rewrite": {
        "headers": {
            "add": {
                "X-Server-id": 3,
                "X-Server-status": "on",
                "X-Server-balancer_addr": "$balancer_ip:$balancer_port"
            }
        }
    }
}
```
## Actual
```json
{
    "proxy-rewrite": {
        "headers": {
            "X-Api-Version": "v1",
            "X-Api-Engine": "apisix",
            "X-Api-useless": ""
        }
    }
}
```

mscb402:
### Description

Fixes #8239 

### Checklist


apisix/plugins/proxy-rewrite.lua
```diff
@@ -163,6 +163,8 @@ local function check_set_headers(headers)
         if #field == 0 then
             return false, 'invalid field length in header'
         end
+
+        core.log.info("header field: ", field)
```
tokers, line 167: Why not also show the value?

apisix/plugins/proxy-rewrite.lua
```diff
@@ -70,8 +74,58 @@ local schema = {
         },
         headers = {
             description = "new headers for request",
-            type = "object",
-            minProperties = 1,
+            anyOf = {
```
tokers, line 77: We should use `oneOf`?

docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -39,7 +39,10 @@ The `proxy-rewrite` Plugin rewrites Upstream proxy information such as `scheme`,
...
+| headers                     | object        | False    |         |                                                                                                                                   |                   |
+| headers.add     | object   | false     |        |                 | Append the new headers. The format is `{"name: value",...}`. The values in the header can contain Nginx variables like $remote_addr and $balancer_ip.                                                                                              |
+| headers.set     | object  | false     |        |                 | Rewriting the headers. The format is  `{"name": "value", ...}`. The values in the header can contain Nginx variables like $remote_addr and $balancer_ip.                                                                                                |
```
tzssangglass, line 44: I think use `headers.rewrite` is more clear. 

What happens if `headers.set` is set, but the actual request does not have a header name that needs to be rewritten?

apisix/plugins/proxy-rewrite.lua
mscb402: Because the old version has this line, some test cases will check this output.

apisix/plugins/proxy-rewrite.lua
mscb402: No, we should use anyOf to maintain downward compatibility. The response-rewrite plugin has the same schema.

docs/en/latest/plugins/proxy-rewrite.md
mscb402: This `headers.set` is the same as the response-rewrite plugin. When the header does not exist will add it.

tokers, line 44: I think overwrite will be better than rewriting.

docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -39,7 +39,10 @@ The `proxy-rewrite` Plugin rewrites Upstream proxy information such as `scheme`,
...
+| headers.remove  | array   | false     |        |                 | Remove the headers. The format is `["name", ...]`.
```
tokers, line 45: Should also tell users the priority of set, remove, and add.

t/plugin/proxy-rewrite3.t
```diff
@@ -337,3 +337,94 @@ X-Forwarded-Host: apisix.ai
...
+
+
+=
+== TEST 18: remove header success
+--- request
+GET /echo HTTP/1.1
+--- response_headers
+test2: 2233
+--- no_error_log
+[error]
```
tokers, line 430: Missing cases:

1. The set is not used
2. Add a case which add, set the same header, for testing the priority.

docs/en/latest/plugins/proxy-rewrite.md
mscb402: yes, look like overwrite is better

t/plugin/proxy-rewrite3.t
mscb402: Sorry for the wrong test name, this test case should be named `set header success`. I will fix that and add a new test case for testing the priority.

mscb402: done

docs/en/latest/plugins/proxy-rewrite.md
mscb402: donr

docs/en/latest/plugins/proxy-rewrite.md
mscb402: done

apisix/plugins/proxy-rewrite.lua
```diff
@@ -150,13 +220,69 @@ do
...
         end
+
+        return {
+            add = add,
+            set = set,
+            remove = hdr_conf.remove or {},
+        }
     end
 
+    local function addHeader(ctx, header_name, header_value)
```
tzssangglass, line 252:
```suggestion
    local function add_header(ctx, header_name, header_value)
```

apisix/plugins/proxy-rewrite.lua
```diff
@@ -150,13 +220,69 @@ do
...
+        local oldHeader = core.request.header(ctx, header_name)
```
tzssangglass, line 253:
```suggestion
        local old_header = core.request.header(ctx, header_name)
```

apisix/plugins/proxy-rewrite.lua
mscb402: doen

apisix/plugins/proxy-rewrite.lua
mscb402: done

t/plugin/proxy-rewrite3.t
```diff
@@ -337,3 +337,145 @@ X-Forwarded-Host: apisix.ai
...
+
+
+=== TEST 15: add exist header in muti-header
+--- request
+GET /echo HTTP/1.1
+--- more_headers
+test: sssss
+test: bbb
+--- response_headers
+test: sssss, bbb, 123
```
tzssangglass, line 393: why it is 

```
test: sssss
test: bbb
test: 123
```

mscb402: This is for test header combination. First combine two existing headers `test: sssss` and `test: bbb`. And then add a new header `test: 123` from proxy_rewrite. Finally, it will combine all headers in one.

apisix/plugins/proxy-rewrite.lua
tokers: `anyOf` means you can configure both the old way and the new way at the same time, which is unnecessary. While `oneOf` means you can configure either the old way or the new way.

apisix/plugins/proxy-rewrite.lua
tokers: OK, ignore it.

t/plugin/proxy-rewrite3.t
```diff
@@ -337,3 +337,145 @@ X-Forwarded-Host: apisix.ai
...
+--- no_error_log
+[error]
+
+
+
+=== TEST 20: set and test priority test
+--- request
+GET /echo HTTP/1.1
+--- response_headers
+test: test_in_set
```
tokers, line 479: As per the doc you wrote, the `set` operation takes the precedence than `add`, in such a case, the `test` header should have two values.

t/plugin/proxy-rewrite3.t
tzssangglass: Is it necessary to merge header with the same name? I'm not sure. cc @spacewander

t/plugin/proxy-rewrite3.t
mscb402: Maybe we have different definitions of priorities. In a document, a priority represents an area of influence. The `remove` have a high priority. If you put in `remove` it will remove this header whatever you config in `set` or `add`. If you config `set` and `add` together, only `set` will process because `set` has high priority than `add`. 
So, if in your different definitions of priorities. The priorities should be  `add` > `set` > `remove`

apisix/plugins/proxy-rewrite.lua
```diff
@@ -70,8 +74,58 @@ local schema = {
...
+                                    }
+                                }
+                            },
+                        },
+                        remove = {
+                            type = "array",
+                            minItems = 1,
+                            items = {
+                                type = "string",
+                                -- "Set-Cookie"
```
spacewander, line 121: Set-Cookie is a response header. Better to use another example.

t/plugin/proxy-rewrite3.t
spacewander: We need to use https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/req.md#add_header. Merging multiple request headers with `,` doesn't obey the standard.

t/plugin/proxy-rewrite3.t
```diff
@@ -337,3 +337,145 @@ X-Forwarded-Host: apisix.ai
...
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
```
spacewander, line 467: Please remove duplicate sections which are set in https://github.com/apache/apisix/blob/ae400b95b91240fb867c3cedc9fa65a6b5593f36/t/plugin/proxy-rewrite3.t#L28

t/plugin/proxy-rewrite3.t
```diff
@@ -337,3 +337,145 @@ X-Forwarded-Host: apisix.ai
...
+--- response_body
+passed
+--- no_error_log
```
spacewander, line 470: ditto

t/plugin/proxy-rewrite3.t
mscb402: done

t/plugin/proxy-rewrite3.t
mscb402: done

t/plugin/proxy-rewrite3.t
mscb402: done @spacewander

t/plugin/proxy-rewrite3.t
tokers: I just reviewed the code and now I know that `add` runs before than `set`. What we want to is writing the running priority between these operations. Then users can evaluate if this is expected.

mscb402: document updated

apisix/core/request.lua
```diff
@@ -22,6 +22,12 @@
 local lfs = require("lfs")
 local log = require("apisix.core.log")
 local io = require("apisix.core.io")
+local ngx_req
```
spacewander, line 25: We don't need a ngx_req as a module variable

t/plugin/proxy-rewrite3.t
```diff
@@ -328,12 +379,95 @@ passed
...
 --- request
 GET /echo HTTP/1.1
 --- more_headers
-X-Forwarded-Host: apisix.ai
+test: sssss
+test: bbb
 --- response_headers
-X-Forwarded-Host: test.com
+test: sssss, bbb, 123
+--- no_error_log
```
spacewander, line 390: Need to deal with https://github.com/apache/apisix/pull/8336#discussion_r1026081925 in other test

t/plugin/proxy-rewrite3.t
```diff
@@ -313,6 +313,57 @@ ngx.var.request_uri: /print_uri_detailed
...
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 13: rewrite X-Forwarded-Host
```
spacewander, line 327: This test isn't relative to this feature?

docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -39,9 +39,18 @@ The `proxy-rewrite` Plugin rewrites Upstream proxy information such as `scheme`,
...
 | use_real_request_uri_unsafe | boolean       | False    | false   |                                                                                                                                        | Use real_request_uri (original $request_uri in nginx) to bypass URI normalization. **Enabling this is considered unsafe as it bypasses all URI normalization steps**.                                                                                                                                                                                                                                                                                                     |
 
+## Header Priority
+
+Header Header configurations are executed according to the following priorities:
```
spacewander, line 50:
```suggestion
Header configurations are executed according to the following priorities:
```

apisix/core/request.lua
```diff
@@ -138,6 +145,15 @@ function _M.set_header(ctx, header_name, header_value)
...
 end
 
+function _M.add_header(header_name, header_value)
+    local err
+    header_name, err = _validate_header_name(header_name)
+    if err then
+        error(err)
+    end
+
+    req_add_header(header_name,header_value)
```
spacewander, line 155: Please ensure a space is after the ','. Let's deal with similar places in this PR.

t/plugin/proxy-rewrite3.t
mscb402: Nope, but this change is to remove duplicate sections from your advice.

image

apisix/core/request.lua
mscb402: done

apisix/core/request.lua
mscb402: removed

t/plugin/proxy-rewrite3.t
mscb402: done

apisix/core/request.lua
```diff
@@ -39,6 +44,7 @@ local req_get_uri_args = ngx.req.get_uri_args
 local req_set_uri_args = ngx.req.set_uri_args
 
 
+
```
tzssangglass, line 47: delete this line

mscb402: done

apisix/plugins/proxy-rewrite.lua
mscb402: If replace `anyOf` with `oneOf`, the schema can't work correctly.

mscb402: Done

tokers (approved):

spacewander (approved):

PR #8358, opened 2022-11-18
refactor: use pl.stringx.rfind to replace the get_last_index func

monkeyDluffy6017:
### Description

We use the `get_last_index` function to get the last index of substring, but the function is not correct now.https://github.com/apache/apisix/blob/3cac54c84661fcb60416afe0234dcb0115a77370/apisix/plugins/log-rotate.lua#L132

### Checklist


tzssangglass:
can we add test cases to cover this?

monkeyDluffy6017:
@tzssangglass  I can test the function with unit test, how can i export the local function `get_last_index`?

tzssangglass:
> I can test the function with unit test, how can i export the local function `get_last_index`?

core.log.debug?

In fact, without testing, and with all the current test cases passing, I don't know where the original error was.

monkeyDluffy6017:
How do I write a test case for this? The original function is faulty, but the final function happens to be fine
@spacewander

spacewander:
> How do I write a test case for this? The original function is faulty, but the result happens to be fine @spacewander

We can provide some fake access log to see if the function can filter out the right one from the wrong ones?

monkeyDluffy6017:
> get_last_index

Can I export the `get_last_index` function and do unit test?

spacewander:
> Can I export the `get_last_index` function and do unit test?

Is this the only way?

monkeyDluffy6017:
> Is this the only way?

It doesn't matter if we don't change it, the log-rotate plugin is now working fine, but someone should be careful if they are going to use the `get_last_index` function later

spacewander:
> > Is this the only way?
> 
> It doesn't matter if we don't change it, the log-rotate plugin is now working fine, but someone should be careful if they are going to use the `get_last_index` function later

We can put it in the core.utils so it's able to test the function?

monkeyDluffy6017:
@spacewander done

apisix/core/utils.lua
```diff
@@ -335,4 +337,19 @@ end
 _M.resolve_var = resolve_var
 
 
+function _M.get_last_index(str, key)
+    local str_rev = str_reverse(str)
+    local key_rev = str_reverse(key)
+    local idx, _ = str_find(str_rev, key_rev)
```
spacewander, line 343: We can use plain string find here?

monkeyDluffy6017: Done

apisix/core/utils.lua
```diff
@@ -335,4 +337,19 @@ end
 _M.resolve_var = resolve_var
 
 
+function _M.get_last_index(str, key)
+    local str_rev = str_reverse(str)
+    local key_rev = str_reverse(key)
+    local idx, _ = str_find(str_rev, key_rev, 1, true)
```
spacewander, line 343: We can use core_str.find for the plain str find

spacewander:

spacewander (dismissed):

kingluo (dismissed):

soulbird (dismissed):

apisix/core/utils.lua
```diff
@@ -335,6 +336,21 @@ end
 _M.resolve_var = resolve_var
 
 
+function _M.get_last_index(str, key)
+    local str_rev = str_reverse(str)
+    local key_rev = str_reverse(key)
+    local idx, _ = core_str.find(str_rev, key_rev)
```
soulbird, line 342: `local idx = core_str.find(str_rev, key_rev)`

monkeyDluffy6017: Done

apisix/core/utils.lua
```diff
@@ -335,6 +336,21 @@ end
 _M.resolve_var = resolve_var
 
 
+function _M.get_last_index(str, key)
```
kingluo, line 339: for such a basic string operation, I suggest reusing the `pl.stringx` from Penlight library. After all, we already import this lib in our code base. Never reinvent the wheel unless you have to.

https://lunarmodules.github.io/Penlight/libraries/pl.stringx.html#rfind

Example:

```
> print(require("pl.stringx").rfind("apisix__1928", "__"))
7
```

monkeyDluffy6017: Done

kingluo (approved):

Issue #8164, opened 2022-10-25
feat: As a user, I want to add type for every plugin, so that we can get the type info of every plugin in the schema


starsz:
### Description

Hello, I find that we have a type field in the plugin.

```
local _M = {
    version = 0.1,
    priority = 2500,
    type = 'auth',
    name = plugin_name,
    schema = schema,
    consumer_schema = consumer_schema,
}

```
refer: https://github.com/apache/apisix/blob/master/apisix/plugins/key-auth.lua#L57.

But only the auth plugin uses the type. For example: 
The file-logger plugin doesn't have.
refer: https://github.com/apache/apisix/blob/master/apisix/plugins/file-logger.lua#L47-L53.

So I really recommend that we can add the type filed to every plugin.
We can have "auth", "security", "serverless", "transformation" e.g.

spacewander:
The `auth` type is used here: https://github.com/apache/apisix/blob/e2d0f0b16bb157412684a38b129d12acb4628d39/apisix/consumer.lua#L46. (BTW, the `auth` is a bad name to me. It would be better to be named as `consumer`. We keep this name just because we don't have a chance to rename it.) 

The `auth` type is also used in the dashboard to make sure a plugin can be bound with the consumer.

So what is the use case for `"security"` and the other types, and how to classify them?

starsz:
> So what is the use case for "security" and the other types, and how to classify them?

IMO, I think we can follow the docs of APISIX, https://apisix.apache.org/docs/apisix/installation-guide/
<img width="276" alt="image" src="https://user-images.githubusercontent.com/25628854/197705116-38ff7072-ea3c-4350-850b-dbc822e432d9.png">


The type also can be used in the dashboard, for example 
<img width="2248" alt="image" src="https://user-images.githubusercontent.com/25628854/197706260-8702aa1f-12d0-4805-b52e-e29473312d60.png">
The front-end can use the type field to list and filter the plugin.


> how to classify them

We can define the plugin type list as constant in the APISIX core first.

pixeldin:
@starsz, I would love to try it out, can you assign it to me?

pixeldin:
> "the auth is a bad name to me. It would be better to be named as consumer."

What about my thought is:
1. Stay and keep the `type` field in the original plugin, as reserved for logical judgment for the consumer, we can rename the type value of "auth" to "consumer" or whatever appropriate context.
[...]

PR #8362, opened 2022-11-20
feat: adding category for plugin to classify

pixeldin:
### Description

Provides a `category` field for existing plugins to distinguish and classify different plugin as requested in https://github.com/apache/apisix/issues/8164.
Which are refer to the docs of [apisix plugin list](https://apisix.apache.org/docs/apisix/next/plugins/batch-requests/) and
 option values could be **General/Authentication/Security/Serverless**, etc...

### Checklist


starsz (commented):
Can we add it into schema ?

pixeldin:
> Can we add it into schema ?

Ok, I will try that.

PR #8374, opened 2022-11-22
fix: passive healthy code review.

840963657:
### Description

Fixes # (issue)

### Checklist


840963657:
when resp_status == status, there is no need to perform other loop operations.

tzssangglass (approved):

apisix/init.lua
```diff
@@ -724,6 +724,7 @@ local function healthcheck_passive(api_ctx)
                                            port or api_ctx.balancer_port,
                                            host,
                                            resp_status)
+                return
```
spacewander, line 727: We should use break here?

840963657: In the second loop, I think break and return have the same effect

tzssangglass: we should use break here.

spacewander: Sorry, I can't get your point. What does "the second loop' mean?

840963657: Loop of the 727 line, the function ends when the loop terminates, so think break and return have the same effect
image

tzssangglass: in L730, we should use `break`?

Issue #8319, opened 2022-11-14
Proposal: APISIX Supports Secrets Management


soulbird:
# Background

Secrets refer to any sensitive information required during the running process of APISIX, which may be part of the core configuration (such as the etcd's password) or some sensitive information in the plugin. Common types of Secrets in APISIX include:

- username, the password for some components (etcd, Redis, Kafka, etc.)
- the private key of the certificate
- API key
- Sensitive plugin configuration fields, typically used for authentication, hashing, signing, or encryption

 Secrets Management refers to allowing users to store Secrets in APISIX through some secrets management services (vault, etc.), and read them according to the key when using them to ensure that **Secrets do not exist in plaintext in the entire platform**.

Currently, APISIX only supports the use of vault to manage secrets in the `jwt-auth` plugin. We plan to use more secrets management services in more plugins.

 # Scheme

## Overall Design

In order to conveniently use the Secrets Management capability in various plugins of APISIX to protect secrets or other sensitive data, we will implement the following steps:

1. Add the environment variable module to support referencing environment variables in plugins and other APISIX resources in a specific way
2. Design a general KMS component, compatible with different secrets management services (vault, aws, etc.), and support referring to the KMS component in a specific way in the plugin
3. Add a new KMS resource in APISIX to configure and manage KMS components
4. Use a "non-intrusive" way to reference KMS components in plugins
5. We will first use secrets Management in the authentication class plugin

It works as follows:
![image](https://user-images.githubusercontent.com/11553520/201565839-c3746894-c555-4f58-8baf-28739f82c5f0.png)

### Environment variable

APISIX loads all environment variables at startup, so there is no need to add them through custom resources. Environment variables can be referenced in the following ways:
```
$ENV://$env_name/$sub_key
```

- env_name: environment variable name
- sub_key: get the value of a property when the value of the environment variable is a JSON string

 If the value of the environment variable is of type string, such as:

```
export JACK_JWT_KEY=abc
```

It can be referenced as follows:
```
$ENV://JACK_JWT_KEY
```

If the value of the environment variable is a JSON string like:
```
export JACK={"jwt-key":"abc","openid-key": "def"}
```

It can be referenced as follows:
```
# Get the jwt-key of the environment variable JACK
$ENV://JACK/jwt-key
# Get the openid-key of the environment variable JACK
$ENV://JACK/openid-key
```

#### Example: use in jwt-auth plugin

First, create environment variables before the APISIX instance starts
```
[...]

PR #8390, opened 2022-11-24
feat: support store secret in env for auth plugin

soulbird:
### Description

As part of https://github.com/apache/apisix/issues/8319:

Step 1: Support the use of environment variables to manage keys. Information in the consumer configuration is more private, so we first used environment variables in the consumer to manage sensitive information.

Currently only test cases have been added for the key-auth plugin, the test cases for other plugins will be completed in a separate PR.

### Checklist

### DIFF
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -14,18 +14,19 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
-local ffi = require "ffi"
+local ffi         = require "ffi"
 
-local json   = require("apisix.core.json")
-local log    = require("apisix.core.log")
-local string = require("apisix.core.string")
+local json        = require("apisix.core.json")
+local log         = require("apisix.core.log")
+local string      = require("apisix.core.string")
 
-local os = os
-local type = type
-local upper = string.upper
-local find = string.find
-local sub = string.sub
-local str = ffi.string
+local os          = os
+local type        = type
+local upper       = string.upper
+local find        = string.find
+local sub         = string.sub
+local str         = ffi.string
+local C           = ffi.C
 
 local _M = {}
 
@@ -35,6 +36,7 @@ local apisix_env_vars = {}
 
 ffi.cdef [[
   extern char **environ;
+  int memcmp(const void *s1, const void *s2, size_t n);
 ]]
 
 
@@ -59,9 +61,10 @@ end
 
 
 local function is_env_ref(ref)
-    -- Avoid the error caused by has_prefix to cause a crash,
-    -- so judge the type of ref.
-    return type(ref) == "string" and string.has_prefix(upper(ref), ENV_PREFIX)
+    -- We will not use string.has_prefix,
+    -- to avoid the error caused by has_prefix to cause a crash.
+    return type(ref) == "string" and #ref > 7 and
+        0 == C.memcmp(upper(ref), ENV_PREFIX, 7)
 end
 
 
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -59,7 +59,9 @@ end
 
 
 local function is_env_ref(ref)
-    return type(ref) == "string" and sub(upper(ref), 1, 7) == ENV_PREFIX
+    -- Avoid the error caused by has_prefix to cause a crash,
+    -- so judge the type of ref.
+    return type(ref) == "string" and string.has_prefix(upper(ref), ENV_PREFIX)
 end
 
 
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -26,7 +26,6 @@ local upper       = string.upper
 local find        = string.find
 local sub         = string.sub
 local str         = ffi.string
-local C           = ffi.C
 
 local _M = {}
 
@@ -36,7 +35,6 @@ local apisix_env_vars = {}
 
 ffi.cdef [[
   extern char **environ;
-  int memcmp(const void *s1, const void *s2, size_t n);
 ]]
 
 
@@ -61,15 +59,13 @@ end
 
 
 local function is_env_ref(ref)
-    -- We will not use string.has_prefix,
-    -- to avoid the error caused by has_prefix to cause a crash.
-    return type(ref) == "string" and #ref > 7 and
-        0 == C.memcmp(upper(ref), ENV_PREFIX, 7)
+    -- Avoid the error caused by has_prefix to cause a crash.
+    return type(ref) == "string" and string.has_prefix(upper(ref), ENV_PREFIX)
 end
 
 
 local function parse_ref(ref)
-    local path = sub(ref, 8)
+    local path = sub(ref, #ENV_PREFIX + 1)
     local idx = find(path, "/")
     if not idx then
         return {key = path, sub_key = ""}
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -69,12 +69,15 @@ local function parse_ref(ref)
     local path = sub(ref, 8)
     local idx = find(path, "/")
     if not idx then
-        return path, ""
+        return {key = path, sub_key = ""}
     end
     local key = sub(path, 1, idx - 1)
     local sub_key = sub(path, idx + 1)
 
-    return key, sub_key
+    return {
+      key = key,
+      sub_key = sub_key
+    }
 end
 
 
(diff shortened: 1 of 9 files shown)


apisix/core/env.lua
```diff
@@ -0,0 +1,97 @@
...
+  local e = ffi.C.environ
+  if not e then
+    log.warn("could not access environment variables")
+    return
+  end
+
+  local i = 0
+  while e[i] ~= nil do
+    local var = str(e[i])
+    local p = find(var, "=")
```
spacewander, line 50: We can use core.string.find, which is a plaintext find.

apisix/core/env.lua
```diff
@@ -0,0 +1,97 @@
...
+function _M.get(ref)
+    if not is_env_ref(ref) then
+        return nil
+    end
+
+    local key, sub_key = parse_ref(ref)
+    local main_value = apisix_env_vars[key] or os.getenv(key)
+    if main_value and sub_key ~= "" then
+        local vt = json.decode(main_value)
+        if not vt then
```
spacewander, line 87: Let's log down the error when the decode is failed.

apisix/core/utils.lua
```diff
@@ -329,4 +332,57 @@ end
...
+                retrieve_ref(v)
+            end
+        end
+        return refs
+    end
+
+    local function retrieve(refs)
+        log.info(string.format("retrieve secrets refs: %p", refs))
+
+        local new_refs = table.clone(refs)
```
spacewander, line 357: Do we need to use deepcopy for the later recursive call?

soulbird: Actually I'm not sure, but it seems safer to use deepcopy.

spacewander (dismissed):

apisix/core/env.lua
```diff
@@ -0,0 +1,99 @@
...
+  local e = ffi.C.environ
+  if not e then
+    log.warn("could not access environment variables")
+    return
+  end
+
+  local i = 0
+  while e[i] ~= nil do
+    local var = str(e[i])
+    local p = find(var, "=")
```
membphis, line 51: we should use `plain` model, all right?

membphis: which is better performance

apisix/core/env.lua
```diff
@@ -0,0 +1,99 @@
...
+        apisix_env_vars[sub(var, 1, p - 1)] = sub(var, p + 1)
+    end
+
+    i = i + 1
+  end
+end
+
+
+local function is_env_ref(ref)
+    return type(ref) == "string" and sub(upper(ref), 1, 7) == ENV_PREFIX
```
membphis, line 62: we can use `C.memcmp` which is better performance

apisix/core/env.lua
```diff
@@ -0,0 +1,99 @@
...
+end
+
+
+local function parse_ref(ref)
+    local path = sub(ref, 8)
+    local idx = find(path, "/")
```
membphis, line 68: use `plain` model

apisix/core/env.lua
```diff
@@ -0,0 +1,99 @@
...
+    return key, sub_key
+end
+
+
+function _M.get(ref)
+    if not is_env_ref(ref) then
+        return nil
+    end
+
+    local key, sub_key = parse_ref(ref)
```
membphis, line 84: I think we should use `array` object

apisix/core/utils.lua
```diff
@@ -329,4 +332,57 @@ end
...
+                refs[k] = env.get(v) or v
+            elseif typ == "table" then
+                retrieve_ref(v)
+            end
+        end
+        return refs
+    end
+
+    local function retrieve(refs)
+        log.info(string.format("retrieve secrets refs: %p", refs))
```
membphis, line 355: is it maybe in `hot` code path? if yes, we need to cache `string.format`

apisix/core/env.lua
soulbird: Already use `local string = require("apisix.core.string")`, it's `plain` mode.

apisix/core/env.lua
soulbird: ditto

apisix/core/env.lua
membphis: @soulbird you are right, ignore my comments

apisix/core/utils.lua
soulbird: Not a hot path, but it would be better to use cache.

apisix/core/env.lua
```diff
@@ -0,0 +1,107 @@
...
+end
+
+
+function _M.get(ref)
+    if not is_env_ref(ref) then
+        return nil
+    end
+
+    local opts = parse_ref(ref)
+    local main_value = apisix_env_vars[opts.key] or os.getenv(opts.key)
```
tokers, line 93: Why still try `os.getenv` here?

apisix/core/env.lua
```diff
@@ -0,0 +1,107 @@
...
+    if main_value and opts.sub_key ~= "" then
```
tokers, line 94: Where is the convention about the value format (a JSON string), missing docs will make users confusing.

tokers: Also, looks like we only support one level index, that also needs to be written.

apisix/core/env.lua
```diff
@@ -0,0 +1,107 @@
...
+    i = i + 1
+  end
+end
+
+
+local function is_env_ref(ref)
+    -- We will not use string.has_prefix,
+    -- to avoid the error caused by has_prefix to cause a crash.
+    return type(ref) == "string" and #ref > 7 and
+        0 == C.memcmp(upper(ref), ENV_PREFIX, 7)
```
spacewander, line 67: `type(ref) == "string" and string.has_prefix(upper(ref), ENV_PREFIX)`, will this way cause a crash?

apisix/core/env.lua
```diff
@@ -0,0 +1,107 @@
...
+local apisix_env_vars = {}
+
+ffi.cdef [[
+  extern char **environ;
+  int memcmp(const void *s1, const void *s2, size_t n);
+]]
+
+
+function _M.init()
+  local e = ffi.C.environ
```
spacewander, line 44:
```suggestion
  local e = C.environ
```

apisix/core/env.lua
```diff
@@ -0,0 +1,107 @@
...
+local function is_env_ref(ref)
+    -- We will not use string.has_prefix,
+    -- to avoid the error caused by has_prefix to cause a crash.
+    return type(ref) == "string" and #ref > 7 and
+        0 == C.memcmp(upper(ref), ENV_PREFIX, 7)
+end
+
+
+local function parse_ref(ref)
+    local path = sub(ref, 8)
```
spacewander, line 72:
```suggestion
    local path = sub(ref, #ENV_PREFIX + 1)
```
can remove the magic number.

t/core/env.t
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+nil
+
+
+
+=== TEST 4: worong format: worong type
```
tokers, line 80:
```suggestion
=== TEST 4: wrong format: wrong type
```

apisix/core/env.lua
soulbird: No, but `type(ref) == "string"` will be executed one more time.

apisix/core/env.lua
soulbird: In order to be compatible with `env foo=bar;` this Nginx configuration.

apisix/core/env.lua
soulbird: The docs will be completed at the end, I'll notice you mention this.

tokers: Got it.

apisix/core/env.lua
tokers: Then why now also handle them when init the `apisix_env_vars`.

apisix/core/env.lua
spacewander: IMHO, check type twice is better than reinventing a function. The type check is cheap.

soulbird: OK

apisix/core/env.lua
soulbird: Means `Then why not also`?
Because I didn't find a way to iterate through them, only by a specific name.

spacewander (approved):

apisix/core/env.lua
tokers: ok

t/core/env.t
tokers: Forget to change this?

tokers (approved):

tzssangglass (approved):

spacewander (approved):

PR #8394, opened 2022-11-24
feat(admin): add kms admin api

kingluo:
### Description

As part of https://github.com/apache/apisix/issues/8319:

Support adding KMS resources through Admin API.

### Checklist

### DIFF
--- a/apisix/admin/kms.lua
+++ b/apisix/admin/kms.lua
@@ -58,14 +58,14 @@ end
 function _M.put(id, conf, sub_path)
     local uri_segs = core.utils.split_uri(sub_path)
     if #uri_segs ~= 1 then
-        return 400, "no kms id in uri"
+        return 400, {error_msg = "no kms id in uri"}
     end
     local typ = id
     id = uri_segs[1]
 
     local ok, err = check_conf(id, conf, true, typ)
     if not ok then
-        return 400, err
+        return 400, {error_msg = err}
     end
 
     local key = "/kms/" .. typ .. "/" .. id
@@ -115,7 +115,7 @@ end
 function _M.delete(id, conf, sub_path)
     local uri_segs = core.utils.split_uri(sub_path)
     if #uri_segs ~= 1 then
-        return 400, "no kms id in uri"
+        return 400, {error_msg = "no kms id in uri"}
     end
     local typ = id
     id = uri_segs[1]
@@ -139,7 +139,7 @@ end
 function _M.patch(id, conf, sub_path)
     local uri_segs = core.utils.split_uri(sub_path)
     if #uri_segs < 2 then
-        return 400, "no kms id and/or sub path in uri"
+        return 400, {error_msg = "no kms id and/or sub path in uri"}
     end
     local typ = id
     id = uri_segs[1]
@@ -191,12 +191,12 @@ function _M.patch(id, conf, sub_path)
 
     local ok, err = check_conf(id, node_value, true, typ)
     if not ok then
-        return 400, err
+        return 400, {error_msg = err}
     end
 
     local res, err = core.etcd.atomic_set(key, node_value, nil, modified_index)
     if not res then
-        core.log.error("failed to set new consumer group[", key, "]: ", err)
+        core.log.error("failed to set new kms[", key, "]: ", err)
         return 503, {error_msg = err}
     end
 
--- a/apisix/admin/kms.lua
+++ b/apisix/admin/kms.lua
@@ -55,13 +55,22 @@ local function check_conf(id, conf, need_id, typ)
 end
 
 
-function _M.put(id, conf, sub_path)
+local function split_typ_and_id(id, sub_path)
     local uri_segs = core.utils.split_uri(sub_path)
-    if #uri_segs ~= 1 then
+    local typ = id
+    local id = nil
+    if #uri_segs > 0 then
+        id = uri_segs[1]
+    end
+    return typ, id
+end
+
+
+function _M.put(id, conf, sub_path)
+    local typ, id = split_typ_and_id(id, sub_path)
+    if not id then
         return 400, {error_msg = "no kms id in uri"}
     end
-    local typ = id
-    id = uri_segs[1]
 
     local ok, err = check_conf(typ .. "/" .. id, conf, true, typ)
     if not ok then
@@ -86,12 +95,7 @@ end
 
 
 function _M.get(id, conf, sub_path)
-    local uri_segs = core.utils.split_uri(sub_path)
-    local typ = id
-    id = nil
-    if #uri_segs > 0 then
-        id = uri_segs[1]
-    end
+    local typ, id = split_typ_and_id(id, sub_path)
 
     local key = "/kms/"
     if id then
@@ -111,12 +115,10 @@ end
 
 
 function _M.delete(id, conf, sub_path)
-    local uri_segs = core.utils.split_uri(sub_path)
-    if #uri_segs ~= 1 then
+    local typ, id = split_typ_and_id(id, sub_path)
+    if not id then
         return 400, {error_msg = "no kms id in uri"}
     end
-    local typ = id
-    id = uri_segs[1]
 
     if not id then
         return 400, {error_msg = "missing kms id"}
--- a/apisix/admin/kms.lua
+++ b/apisix/admin/kms.lua
@@ -46,7 +46,7 @@ local function check_conf(id, conf, need_id, typ)
     conf.id = id
 
     core.log.info("conf: ", core.json.delay_encode(conf))
-    local ok, err = core.schema.check(core.schema[typ], conf)
+    local ok, err = core.schema.check(core.schema["kms_" .. typ], conf)
     if not ok then
         return nil, {error_msg = "invalid configuration: " .. err}
     end
--- a/apisix/admin/kms.lua
+++ b/apisix/admin/kms.lua
@@ -63,7 +63,7 @@ function _M.put(id, conf, sub_path)
     local typ = id
     id = uri_segs[1]
 
-    local ok, err = check_conf(id, conf, true, typ)
+    local ok, err = check_conf(typ .. "/" .. id, conf, true, typ)
     if not ok then
         return 400, {error_msg = err}
     end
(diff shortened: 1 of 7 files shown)


apisix/schema_def.lua
```diff
@@ -692,6 +692,23 @@ _M.service = {
 }
 
 
+_M.vault = {
+    type = "object",
+    properties = {
+        uri = {
+            type = "string",
```
soulbird, line 699: Should use `uri_def`.

apisix/schema_def.lua
```diff
@@ -692,6 +692,23 @@ _M.service = {
 }
 
 
+_M.vault = {
```
soulbird, line 695: Use `kms_vault`.

t/admin/kms.t
```diff
@@ -0,0 +1,192 @@
...
+            assert(prev_update_time ~= update_time, "update_time should be changed")
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 5: delete consumer-group
+--- config
```
soulbird, line 181: typo?

t/admin/kms.t
```diff
@@ -0,0 +1,192 @@
...
+log_level("info");
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if (!$block->request) {
+        $block->set_value("request", "GET /t");
+    }
+
+    if ((!defined $block->error_log) && (!defined $block->no_error_log)) {
```
spacewander, line 32: We don't need this now?

apisix/admin/kms.lua
```diff
@@ -0,0 +1,207 @@
...
+        return 503, {error_msg = err}
+    end
+
+    utils.fix_count(res.body, id)
+    return res.status, res.body
+end
+
+
+function _M.delete(id, conf, sub_path)
+    local uri_segs = core.utils.split_uri(sub_path)
```
spacewander, line 116: Can we refactor the code to get type & id from sub_path in a function?

apisix/admin/kms.lua
```diff
@@ -0,0 +1,207 @@
...
+    core.log.info("new conf: ", core.json.delay_encode(node_value, true))
+
+    local ok, err = check_conf(id, node_value, true, typ)
+    if not ok then
+        return 400, err
+    end
+
+    local res, err = core.etcd.atomic_set(key, node_value, nil, modified_index)
+    if not res then
+        core.log.error("failed to set new consumer group[", key, "]: ", err)
```
spacewander, line 199: Typo?

apisix/admin/kms.lua
```diff
@@ -0,0 +1,207 @@
...
+    end
+
+    return true
+end
+
+
+function _M.put(id, conf, sub_path)
+    local uri_segs = core.utils.split_uri(sub_path)
+    if #uri_segs ~= 1 then
+        return 400, "no kms id in uri"
```
spacewander, line 61: Better to keep the style of error message. We should use `{error_msg = xxx}` like other places

apisix/admin/kms.lua
kingluo: PATCH requires a different number of segments than GET and PUT, then it makes function wrap a bit complicated.
And the goal is to replace `id` with real id, and the original id is actually kms type, which is different from other admin resources. And the replacement happens after the invocation, so I think it's trivial to wrap them into function.

apisix/constants.lua
```diff
@@ -33,6 +33,7 @@ return {
         ["/protos"] = true,
         ["/plugin_configs"] = true,
         ["/consumer_groups"] = true,
+        ["/kms/vault"] = true,
```
soulbird, line 36: `["/kms"] = true` is enough. We need to load all kms configs on init_worker, not just vault.

t/admin/kms.t
```diff
@@ -0,0 +1,188 @@
...
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/kms/vault/test1',
+                 ngx.HTTP_DELETE
+            )
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
```
soulbird, line 188: Add test case for schema.

apisix/constants.lua
kingluo: Add both `/kms` and `/kms/vault` makes it work.
Because we need to get all resources both for `/kms` and `/kms/vault` prefix.

apisix/admin/kms.lua
spacewander: But we can share the same code for GET and PUT and DELETE right? This is enough to simplify the implementation instead of copying the code three times.

apisix/admin/kms.lua
```diff
@@ -0,0 +1,207 @@
...
+
+function _M.put(id, conf, sub_path)
+    local typ, id = split_typ_and_id(id, sub_path)
+    if not id then
+        return 400, {error_msg = "no kms id in uri"}
+    end
+
+    local ok, err = check_conf(typ .. "/" .. id, conf, true, typ)
+    if not ok then
+        return 400, {error_msg = err}
```
spacewander, line 77: The err from check_conf already contains error_msg field.

apisix/admin/kms.lua
```diff
@@ -0,0 +1,207 @@
...
+end
+
+
+function _M.delete(id, conf, sub_path)
+    local typ, id = split_typ_and_id(id, sub_path)
+    if not id then
+        return 400, {error_msg = "no kms id in uri"}
+    end
+
+    if not id then
```
spacewander, line 123: Why check the id twice?

docs/en/latest/admin-api.md
```diff
@@ -1119,3 +1120,65 @@ Route used in the [Stream Proxy](./stream-proxy.md).
 To learn more about filtering in stream proxies, check [this](./stream-proxy.md#more-route-match-options) document.
 
 [Back to TOC](#table-of-contents)
+
+## kms
+
+**API**: /apisix/admin/kms/{secretmanager}/{id}
+
+kms means `Secrets Management`, which could use any secret manager supported, e.g. `vault`.
```
spacewander, line 1128: I am curious why kms means `Secrets Management`. `kms` is not the abbreviation of Secrets Management.

docs/en/latest/admin-api.md
```diff
@@ -1119,3 +1120,65 @@ Route used in the [Stream Proxy](./stream-proxy.md).
...
+
+| Parameter   | Required | Type        | Description                                                                                                        | Example                                          |
+| ----------- | -------- | ----------- | ------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------ |
+| uri    | True     | URI        | URI of the vault server.                                                                                              |                                                  |
+| prefix    | True    | string        | key prefix
+| token     | True    | string      | vault token. |                                                  |
+| desc        | False    | Auxiliary   | Description of usage scenarios.                                                                                    | kms xxxx                                    |
+| labels      | False    | Match Rules | Attributes of the kms specified as key-value pairs.                                                           | {"version":"v2","build":"16","env":"production"} |
```
spacewander, line 1151: Does kms have these fields?

docs/en/latest/admin-api.md
kingluo: @soulbird Please help here.

soulbird: There is currently no good acronym to summarize `secrets management`, kms seems relatively appropriate. I think we can weaken kms first without explaining it. So is it better to change it to something like this:
With the kms resource, you can use any secrets management service already supported in APISIX. @spacewander

docs/en/latest/admin-api.md
```diff
@@ -1119,3 +1120,62 @@ Route used in the [Stream Proxy](./stream-proxy.md).
...
+### Request Body Parameters
+
+When `{secretmanager}` is `vault`:
+
+| Parameter   | Required | Type        | Description                                                                                                        | Example                                          |
+| ----------- | -------- | ----------- | ------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------ |
+| uri    | True     | URI        | URI of the vault server.                                                                                              |                                                  |
+| prefix    | True    | string        | key prefix
+| token     | True    | string      | vault token. |                                                  |
+| desc        | False    | Auxiliary   | Description of usage scenarios.                                                                                    | kms xxxx                                    |
```
spacewander, line 1150: The kms_vault doesn't have desc field

t/admin/kms.t
```diff
@@ -0,0 +1,221 @@
...
+                    },
+                    "key": "/apisix/kms/vault/test1"
+                }]]
+                )
+
+            ngx.status = code
+            ngx.say(body)
+
+            local res = assert(etcd.get('/kms/vault/test1'))
+            local create_time = res.body.node.value.create_time
```
spacewander, line 64: Look like we need to add the xxx_time fields, otherwise, the test cases need to be changed.

spacewander: My bad. These timestamps are automatically injected in the Admin API.

spacewander (approved):

soulbird (approved):

tzssangglass (approved):

apisix/admin/kms.lua
```diff
@@ -0,0 +1,203 @@
...
+    if not res then
+        core.log.error("failed to put kms [", key, "]: ", err)
+        return 503, {error_msg = err}
+    end
+
+    return res.status, res.body
+end
+
+
+function _M.get(id, conf, sub_path)
```
shreemaan-abhishek, line 97: @kingluo, @soulbird `conf` is unused.

apisix/admin/kms.lua
```diff
@@ -0,0 +1,203 @@
...
+    end
+
+    local res, err = core.etcd.get(key, not id)
+    if not res then
+        core.log.error("failed to get kms [", key, "]: ", err)
+        return 503, {error_msg = err}
+    end
+
+    utils.fix_count(res.body, id)
+    return res.status, res.body
```
shreemaan-abhishek, line 113: What do you think about `return res, nil`? This way the second return variable is always of type `error` and first one will always be `response`

shreemaan-abhishek (commented):
I was trying to use this method and found these difficulties. I am new to Lua so my suggestions may not be of use.

apisix/admin/kms.lua
soulbird: This is to be consistent with the operation functions of other resources, and they will be called uniformly here https://github.com/apache/apisix/blob/master/apisix/admin/init.lua#L201
In fact, we are refactoring this part, you are welcome to participate in the review https://github.com/apache/apisix/pull/8611

PR #8403, opened 2022-11-25
feat: support global data encryption of secret information

### LINKED ISSUE #8407: Proposal: APISIX Supports Global Data Encryption
# Background
Nowadays, if a plugin needs to store sensitive information, it will design a separate set of encryption mechanism, which causes the problem of repeated development and cumbersome management. Now we consider introducing Data Encryption function, hereinafter referred to as DE, to encrypt the specified information, and any module/plugin that needs to encrypt sensitive information only needs to specify the fields to be encrypted on display, so that transparent encryption/decryption can be performed when saving/reading the information.

# Benefits
Unify the encryption mechanism of APISIX to avoid duplication of development while improving product competitiveness

# Goals 

APISIX provides DE functionality to support the use of this capability in plugins and elsewhere to protect sensitive information.

## The user scenarios are as follows

1. Turn on the DE function in the configuration file

```YAML
apisix:
    enable_global_data_encryption: true
```

2. set encrypted fields

```Ada
local schema = {
    type = "object",
    properties = {
        username = { type = "string" },
        password = { type = "string", encrypted = true },   <==== Here
    },
    required = {"username", "password"},    
}
```

3. Set the plugin

```Shell
curl http://127.0.0.1:9180/apisix/admin/consumers -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "username": "foo",
    "plugins": {
        "basic-auth": {
            "username": "foo",
            "password": "bar"       <= Sensitive Information
        }
    }
}'
```

4. The plugin data read directly from ETCD is encrypted

```C#
etcdctl get /apisix/consumers/foo
{
    "username":"foo",
    "update_time":1668584767,
    "create_time":1668584767,
    "plugins":{
        "basic-auth":{
            "username":"foo",
            "password":"da61c0083b6d19ef3db2490d0da96a71572da0fa"  <= encrypted
        }
    }
}
```

5. The data surface is read normally

```Bash
curl http://127.0.0.1:9180/apisix/admin/consumers -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1'
{
  "list": [
    {
      "createdIndex": 8627,
      "modifiedIndex": 8627,
      "value": {
        "username": "foo",
        "update_time": 1668584767,
        "create_time": 1668584767,
        "plugins": {
          "basic-auth": {
            "username": "foo",
            "password": "bar"         <= Unencrypted
          }
        }
      },
      "key": "/apisix/consumers/foo"
    }
  ],
  "total": 1
}
```

# Detailed Design

1. Add a field in the configuration file to enable data encryption

```YAML
apisix:
    data_encryption:                # use `encrypted = {fields}` in plugin schema to enable encryption
        enable: false               # If not set, the default value is `false`.
        keyring:
            - edd1c9f0985e76a2      # If not set, will save origin value into etcd.
            - qeddd145sfvddff3      # If set this, the keyring should be an array whose elements are string, and the size is also 16, and it will encrypt fields with AES-128-CBC
                                    # !!! So do not change it after encryption, it can't decrypt the fields have be saved if you change !!
                                    # Only use the first key to encrypt, and decrypt in the order of the array.
```

2. Specify the fields to be encrypted in the schema of the plugin

```Ada
local consumer_schema = {
    type = "object",
    title = "work with consumer object",
    properties = {
        username = { type = "string" },
        password = { type = "string", encrypted = true },    <===== here
    },
    required = {"username", "password"},
}

local schema = {

}
```

3. If data encryption is enabled, the corresponding fields of the following plugins will be encrypted

```Bash
jwt-auth：secret，private_key
basic-auth：password
authz-keycloak：client_secret 
authz-casdoor：client_secret
openid-connect：client_secret
hmac-auth: secret_key
csrf：key
rocketmq-logger：secret_key
clickhouse-logger：password
error-log-logger：clickhouse.password
sls-logger：access_key_secret
google-cloud-logging：auth_config.private_key
elasticsearch-logger：auth.password
tencent-cloud-cls：secret_key
kafka-proxy：sasl.password
```

4. Encryption and Decryption

- Commonly used algorithms AES, ChaCha20, we use AES128CBC encryption algorithm, the logic is the same as ssl encryption and decryption, that is, only use the first key to encrypt, if it fails to take turns to use all keys to decrypt, consider reusing the ssl encryption and decryption function  （https://github.com/apache/apisix/blob/master/apisix/ssl.lua）

- Currently, only those distributed through the apisix admin interface can be encrypted. Many users have developed their own admin and directly manipulated storage media such as etcd, so they need to implement the encryption function themselves.

- After the encryption function is enabled, the data that was not encrypted needs to be sent down again through the interface in order to be encrypted, is a refresh interface provided? Consider not providing it for now

- Combining the above, the original string is used directly after decryption failure

- Admin now only uses etcd, considering that the first version only supports etcd and only supports encryption of plugin parameters

- Encryption：

Add encryption and decryption to check_single_plugin_schema, note that there are 2 different types of schema within the plugin, namely: schema and consumer_schema

```Lua
local function check_single_plugin_schema(name, plugin_conf, schema_type, skip_disabled_plugin, encrypt_or_decrypt)
    ...
    if plugin_obj.check_schema then
        ...
    end

    if schema_type == core.schema.TYPE_CONSUMER then
        ok, err = encrypt(plugin_conf, plugin_obj.consumer_schema, encrypt_or_decrypt)
    else
        ok, err = encrypt(plugin_conf, plugin_obj.schema, encrypt_or_decrypt)
    end

    return true
end
```

The encryption and decryption function iterates through the corresponding schema, and if it finds the encrypted = true field, it encrypts and decrypts the field in the corresponding configuration

```Lua
local function encrypt(plugin_conf, schema, encrypt_or_decrypt)
    if schema.type == "object" then
        for name, obj in pairs(schema.properties) do
            encrypt(plugin_conf[name], obj, encrypt_or_decrypt)
        end
    elseif schema.type = "array" then
        ...
    end
    
    if schema.encrypted then
        aes128(plugin_conf)
    end
end
```

Add corresponding options in check_schema and stream_check_schema, now directly called in admin, no need to modify, encrypt_or_decrypt is nil by default, i.e. encrypt

```Lua
local function check_schema(plugins_conf, schema_type, skip_disabled_plugin, encrypt_or_decrypt)
    for name, plugin_conf in pairs(plugins_conf) do
        local ok, err = check_single_plugin_schema(name, plugin_conf,
            schema_type, skip_disabled_plugin, encrypt_or_decrypt)
        if not ok then
            return false, err
        end
    end

    return true
end
```

plugin_checker and stream_plugin_checker are used for checksumming when reading data from etcd, and are decrypted by default

```Ada
function _M.plugin_checker(item, schema_type)
    if item.plugins then
        return check_schema(item.plugins, schema_type, true, decrypt) <=== decrypt
    end

    return true
end
```

tzssangglass:
### Description

Fixes #8407

### Checklist


apisix/consumer.lua
```diff
@@ -32,6 +33,23 @@ local lrucache = core.lrucache.new({
...
+    if not consumer_schema then
+        return
+    end
+
+    for key, props in pairs(consumer_schema.properties) do
+        if props.type == "string" and props.encrypted then
+            local encrypted = apisix_ssl.aes_decrypt_pkey(conf[key], "global_data_encrypt")
+            conf[key] = encrypted
+        end
+    end
```
monkeyDluffy6017, line 49: The schema may have nested objects

tzssangglass: I know that since this PR does not include this case, we can optimize this point in the next PR.

docs/zh/latest/plugin-develop.md
```diff
@@ -273,6 +273,37 @@ function _M.check_schema(conf, schema_type)
...
+
+```lua
+password = { type = "string", encrypted = true },
+```
+
+通过在 `schema` 中指定 `encrypted = true`，可以将参数加密存储。APISIX 将提供以下功能：
+
+- 通过 `Admin API` 来新增和更新资源时，对于 `encrypted = true` 的参数，APISIX 会自动加密存储在 etcd 中
+- 通过 `Admin API` 来获取资源时，对于 `encrypted = true` 的参数，APISIX 会自动解密返回
+- APISIX 在使用 `encrypted = true` 的参数时，会自动解密
```
monkeyDluffy6017, line 288: The two line means the same thing?

tzssangglass: essentially, it's the same

apisix/plugin.lua
```diff
@@ -901,7 +961,15 @@ _M.stream_check_schema = stream_check_schema
 
 function _M.plugin_checker(item, schema_type)
     if item.plugins then
-        return check_schema(item.plugins, schema_type, true)
+        local ok, err = check_schema(item.plugins, schema_type, true)
+
+        if ok then
```
monkeyDluffy6017, line 966: Add judge `data_encryption == true` here ? if `data_encryption == false`, no need to iterate the plugin loop

tzssangglass: would check `data_encryption == true` in `decrypt_conf` function, ref: https://github.com/apache/apisix/pull/8403/files/6fd1700dd37fc68d4b057d302229c06bf342eb24#diff-3936316fb5318f20b79ffe05648cf3d6b8c61e6987b51b52e2571c8a8e73a68bR867-R870

monkeyDluffy6017: But still need to iterate the loop below, we can skip the iteration
```
            for name, conf in pairs(item.plugins) do
                decrypt_conf(name, conf, schema_type)
            end
```

docs/zh/latest/plugin-develop.md
monkeyDluffy6017: can we delete one of them?

apisix/admin/routes.lua
```diff
@@ -131,6 +133,9 @@ local function check_conf(id, conf, need_id)
         if not ok then
             return nil, {error_msg = err}
         end
+        for name, conf in pairs(conf.plugins) do
```
monkeyDluffy6017, line 136: can we do this at lower level? it's not convenient to add this to every admin API,

conf/config-default.yaml
```diff
@@ -123,6 +123,13 @@ apisix:
   #  ip: 127.0.0.1
   #  port: 9090
   disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
+  data_encryption:                # use `encrypted = {fields}` in plugin schema to enable encryption
+    enable: false                 # if not set, the default value is `false`.
+    keyring:
+      - edd1c9f0985e76a2          # If not set, will save origin value into etcd.
```
membphis, line 129: we should use a different key

t/plugin/clickhouse-logger.t
```diff
@@ -236,3 +236,235 @@ clickhouse headers: x-clickhouse-key:a
...
+--- error_log
+clickhouse body: INSERT INTO t FORMAT JSONEachRow
+clickhouse headers: x-clickhouse-key:abc123
+clickhouse headers: x-clickhouse-user:default
+clickhouse headers: x-clickhouse-database:default
+--- wait: 5
+
+
+
+=== TEST 9: POST and get lis
```
tzssangglass, line 327:
```suggestion
=== TEST 9: POST and get list
```

apisix/plugin.lua
```diff
@@ -849,6 +850,65 @@ check_plugin_metadata = function(item)
...
+    if not enable then
+        return conf
+    end
+
+    local schema = check_enable_and_get_plugin_schema(name, schema_type)
+    if not schema then
+        return
+    end
+
+    for key, props in pairs(schema.properties) do
```
soulbird, line 903: Don't we consider the case of configuration nesting here?

apisix/plugins/basic-auth.lua
```diff
@@ -39,7 +38,7 @@ local consumer_schema = {
     title = "work with consumer object",
     properties = {
         username = { type = "string" },
-        password = { type = "string" },
+        password = { type = "string", encrypted = true },
```
soulbird, line 41: We need to adjust the documentation of the corresponding plugin for this point.

t/node/consumer-group.t
```diff
@@ -310,3 +310,140 @@ world
...
+        enable: true
+        keyring:
+            - edd1c9f0985e76a2
+--- config
+    location /t {
+        content_by_lua_block {
+            local json = require("toolkit.json")
+            local t = require("lib.test_admin").test
+            local etcd = require("apisix.core.etcd")
+            local code, body = t('/apisix/admin/consumer_groups/company_a',
```
soulbird, line 329: Why use consumer_groups?

t/node/consumer-group.t
```diff
@@ -310,3 +310,140 @@ world
...
+                ngx.log(ngx.ERR, err)
+                return
+            end
+            ngx.sleep(0.1)
+
+            local http = require "resty.http"
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port
+                        .. "/hello"
+            local ress = {}
+            for i = 1, 3 do
```
soulbird, line 431: Why not use `pipelined_requests`?

apisix/plugin.lua
tzssangglass: updated

conf/config-default.yaml
tzssangglass: updated

apisix/plugin.lua
tzssangglass: I know that since this PR does not include this case, we can optimize this point in the next PR.

apisix/plugins/basic-auth.lua
tzssangglass: added

t/node/consumer-group.t
tzssangglass: make sure that data encryption works well with consumer groups

t/node/consumer-group.t
tzssangglass: I need to pass auth headers

docs/en/latest/plugin-develop.md
```diff
@@ -293,6 +293,38 @@ function _M.check_schema(conf, schema_type)
 end
 ```
 
+### encrypted storage fields
+
+Specify the parameters to be stored encrypted(Requires APISIX version > 3.0.1)
```
spacewander, line 298:
```suggestion
Specify the parameters to be stored encrypted. (Requires APISIX version >= 3.1.0)
```

docs/zh/latest/plugin-develop.md
```diff
@@ -273,6 +273,38 @@ function _M.check_schema(conf, schema_type)
 end
 ```
 
+### 加密存储字段
+
+指定参数需要被加密存储（需要 APISIX 版本大于 3.0.1）
```
spacewander, line 278:
```suggestion
指定参数需要被加密存储（需要 APISIX 版本不小于 3.1）
```

apisix/admin/utils.lua
```diff
@@ -78,4 +80,29 @@ function _M.fix_count(body, id)
 end
 
 
+function _M.decrypt_params(decrypt_func, body, schema_type)
+    -- list
+    if body.list and #body.list > 0 then
+        for _, route in ipairs(body.list) do
+            if route.value and route.value.plugins
+               and core.table.nkeys(route.value.plugins) > 0 then
```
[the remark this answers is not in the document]
tzssangglass: done

apisix/plugin.lua
```diff
@@ -849,6 +850,71 @@ check_plugin_metadata = function(item)
...
+        enable_data_encryption =
+            core.table.try_read_attr(local_conf, "apisix", "data_encryption", "enable")
+        _M.enable_data_encryption = enable_data_encryption
+    end
+
+    return enable_data_encryption
+end
+
+
+local function get_plugin_schema(name, schema_type)
```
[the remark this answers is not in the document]
tzssangglass: done

apisix/plugin.lua
```diff
@@ -849,6 +850,71 @@ check_plugin_metadata = function(item)
...
+    if not enable_gde() then
+        return false
```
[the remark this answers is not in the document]
tzssangglass: done

apisix/ssl.lua
```diff
@@ -173,7 +218,11 @@ end
 local function parse_pem_priv_key(sni, pkey)
     core.log.debug("parsing priv key for sni: ", sni)
 
-    local parsed, err = ngx_ssl.parse_pem_priv_key(aes_decrypt_pkey(pkey))
+    local key, err = aes_decrypt_pkey(pkey)
+    if not key then
+        core.log.error(err)
```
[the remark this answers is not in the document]
tzssangglass: done

docs/en/latest/plugin-develop.md
tzssangglass: done

docs/zh/latest/plugin-develop.md
tzssangglass: done

t/node/data_encrypt.t
```diff
@@ -0,0 +1,581 @@
...
+log_level("info");
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if (!$block->request) {
+        $block->set_value("request", "GET /t");
+    }
+
+    if (!$block->no_error_log) {
```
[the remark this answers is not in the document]
tzssangglass: done

t/node/data_encrypt.t
```diff
@@ -0,0 +1,581 @@
...
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
```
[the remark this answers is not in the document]
tzssangglass: done

t/node/data_encrypt.t
```diff
@@ -0,0 +1,581 @@
...
+
+            local pwds = {}
+            table.insert(pwds, res.list[1].value.plugins["basic-auth"].password)
+            table.insert(pwds, res.list[2].value.plugins["basic-auth"].password)
+
+            ngx.say(json.encode(pwds))
+        }
+    }
+--- request
+GET /t
```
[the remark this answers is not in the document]
tzssangglass: done

apisix/plugin.lua
monkeyDluffy6017: It's better to notice this point at doc ?

tzssangglass: I don't think it's needed and will finish it soon.

monkeyDluffy6017: It's a bit complicated, like 'anyof', 'oneof' need to be considered.

t/node/data_encrypt.t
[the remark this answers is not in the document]
spacewander: @tzssangglass 
Err. The `GET /t` here is not removed yet

apisix/plugin.lua
```diff
@@ -849,6 +850,71 @@ check_plugin_metadata = function(item)
 end
 
 
+local enable_data_encryption
+local function enable_gde()
+    if enable_data_encryption == nil then
+        enable_data_encryption =
+            core.table.try_read_attr(local_conf, "apisix", "data_encryption", "enable")
+        _M.enable_data_encryption = enable_data_encryption
```
spacewander, line 858: The cache won't take effect when `core.table.try_read_attr(local_conf, "apisix", "data_encryption", "enable")` is nil. is this by design?

apisix/admin/utils.lua
```diff
@@ -78,4 +80,28 @@ function _M.fix_count(body, id)
 end
 
 
+function _M.decrypt_params(decrypt_func, body, schema_type)
+    -- list
+    if body.list and #body.list > 0 then
```
spacewander, line 85: For the same reason, we can skip the `and #body.list > 0` too.

apisix/plugin.lua
tzssangglass: Anyof or oneof only restricts the existence of items, if an item has `encrypted = true` but does not exist in conf, it will not be encrypted.

apisix/plugin.lua
soulbird: +1

apisix/admin/utils.lua
tzssangglass: done

t/node/data_encrypt2.t
```diff
@@ -0,0 +1,731 @@
...
+                local headers = ngx.req.get_headers()
+                ngx.log(ngx.WARN, "clickhouse body: ", data)
+                for k, v in pairs(headers) do
+                    ngx.log(ngx.WARN, "clickhouse headers: " .. k .. ":" .. v)
+                end
+                ngx.say("ok")
+            }
+        }
+    }
+_EOC_
```
soulbird, line 60: Can it be moved into t/lib? It is the same as in clickhouse-logger.t ?

apisix/plugin.lua
tzssangglass: from #8407

```
apisix:
    data_encryption: 
        enable: false               # If not set, the default value is `false`.
```

if `apisix.data_encryption.enable` is nil, it is the same as not set `apisix.data_encryption.enable`, then `_M.enable_data_encryption` would be `false`. 

`_M.enable_data_encryption` is used by `if method == "get" and plugin.enable_data_encryption then` in `apisix/admin/init.lua`, it won't into `utils.decrypt_params`

t/node/data_encrypt2.t
tzssangglass: done

soulbird (approved):

apisix/plugin.lua
spacewander: @tzssangglass 
`enable_data_encryption` would be nil in this case? So `enable_data_encryption == nil` will always get a true result and this function doesn't provide a cache if the user doesn't provide a configuration.

tzssangglass: 1. if the user doesn't config `apisix.data_encryption.enable`, then `core.table.try_read_attr(local_conf, "apisix", "data_encryption", "enable")` would get the default value `false` from `conf/config-default.yaml`
2. `enable_data_encryption` would be set to `false` as module var.
3. in the next time, `enable_data_encryption == nil` would get false

spacewander (approved):

PR #8404, opened 2022-11-25
fix: the x-forwarded-* header will be influenced by ai plugin

monkeyDluffy6017:
### Description

Because the ai plugin will replace the `handle_upstream` function, and the  `x-forwarded-*` header is set in `handle_upstream` function, so the  `x-forwarded-*` header from downstream will be ignored.
Now move the function that set `x-forwarded-*` header outside the `handle_upstream` function.

### Checklist


tokers (approved):

t/plugin/proxy-rewrite3.t
```diff
@@ -455,3 +454,51 @@ passed
 GET /echo HTTP/1.1
 --- response_headers
 test: test_in_set
+
+
+
+=== TEST 21: set route (test if X-Forwarded-Port can be set before proxy)
```
spacewander, line 460: It seems the newly added test is not relative to the ai module? Can it pass without your modification?

monkeyDluffy6017: `var_x_forwarded_port` is set in `set_upstream_headers`, `set_upstream_headers` is called in `handle_upstream`, and `handle_upstream` is replaced in the ai module.
It will fail in the ci env if the ai module is triggered.

t/plugin/proxy-rewrite3.t
```diff
@@ -455,3 +454,51 @@ passed
...
+--- config
+    location /t {
```
spacewander, line 462: I run the test locally and it can pass even without this bugfix.

spacewander: I am sorry, I can't get your point. This test contains route with plugins, it won't go through the ai module whether in the ci env or not.

monkeyDluffy6017: You could check this, the test case failed when the ai module is triggered in the ci env, but works fine locally. when i remove the ai module, the ci test passed.
https://github.com/apache/apisix/actions/runs/3421218504/jobs/5697063260
I think it maybe satisfy the condition in ci env. 
https://github.com/apache/apisix/blob/master/apisix/plugins/ai.lua#L277

spacewander: @monkeyDluffy6017 
Interesting. The test case here uses `plugins`, which should disable the ai module:
https://github.com/apache/apisix/blob/bd6cbef5e53e7127d6393c8013d66c60a3c1677e/apisix/plugins/ai.lua#L260

So the test doesn't pass through the ai module in local environment.

IMHO, the ci env won't bring any difference as they are the same ai plugin.
Could you point out the reason why the same code acts differently in the ci env?

monkeyDluffy6017: ok, i'll figure it out

monkeyDluffy6017: @spacewander i can not reproduce the problem any more, even fallback to the old version, I will close this PR

PR #8438, opened 2022-12-01
ci: use fixed os version of ubuntu

tzssangglass:
### Description

Fixes # (issue)
ref: https://github.com/actions/runner-images/issues/6399

### Checklist

### DIFF
--- a/.github/workflows/centos7-ci.yml
+++ b/.github/workflows/centos7-ci.yml
@@ -22,7 +22,7 @@ permissions:
 jobs:
   test_apisix:
     name: run ci on centos7
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     timeout-minutes: 90
     strategy:
       fail-fast: false


tao12345666333 (approved):

soulbird (approved):

spacewander (commented):
https://github.com/apache/apisix/blob/0ed451a32e9b7ba9ee4750b6a02289ed26c8ae2b/.github/workflows/fuzzing-ci.yaml#L25

We should use "find & replace" to update all the workflows?

spacewander:
Anyway, if the policy is to fix the CI and upgrade to 22.04 in the future, we can merge this PR now.

tzssangglass:
> We should use "find & replace" to update all the workflows?

Since Dec 1, the default version of `ubuntu-latest` is 22.04. For workflows that are running on `ubuntu-latest` and work well, I think we don't need to downgrade back to 20.04.

> if the policy is to fix the CI and upgrade to 22.04 in the future, we can merge this PR now.

+1

tokers (approved):

spacewander (approved):

PR #8448, opened 2022-12-02
change: rename kms to secret

soulbird:
### Description

Fixes # https://github.com/apache/apisix/issues/8319#issuecomment-1334787833

### Checklist


t/admin/secrets.t
```diff
@@ -53,14 +53,14 @@ __DATA__
                         "prefix" : "apisix",
                         "token" : "apisix"
                     },
-                    "key": "/apisix/kms/vault/test1"
+                    "key": "/apisix/secrets/vault/test1"
```
tzssangglass, line 56: why we use `secrets` in admin API and use `secret` in etcd, according to the current style of APISIX, it should be standardised with `secrets`.

spacewander (approved):

shreemaan-abhishek:
In some places `kms` hasn't yet been updated to `secret`. This has been fixed in: #8697

PR #8453, opened 2022-12-03
chore: update the default HTTP router

### LINKED ISSUE #8354: Change the default HTTP router from radixtree_uri to radixtree_host_uri
### Description

Hi, community, 

We talked about changing the default HTTP router from `radixtree_uri` to `radixtree_host_uri` in the [mailing list]
(https://lists.apache.org/thread/6kkrdk2bkxmccwho3lk5jz5lrsvbsxgt).

I'll show the content here again:

> Let me just explain the differences between them. The radixtree_uri 
router matches API requests with the existing route tree by URI path, 
While the radixtree_host_uri matches the HTTP host and then matches 
the URI path. 
>
> So when Apache APISIX uses radixtree_uri as the default HTTP router, 
there is a confusing scenario (even we can say it's a bug): Let's say 
we have two routes, the first one requires the host matches 
*.[example.com](http://example.com/), and the URI path is /anything, And the second one needs 
the host matches [foo.example.com](http://foo.example.com/) exactly, and the URI path is also 
/anything. In such a case, if we send an API request, we may hit the 
first route, which is counterintuitive. 
>
> We also have a lot of voices from the community that users have been 
in trouble since they think Apache APISIX will consider the HTTP host 
match by default. However, it's not the case. So I'm here to propose 
changing the default HTTP router to radixtree_host_uri. 
>
> I know this is a broken change, so we may have to do this in 3.x 
releases and may not back it port to the 2.15 LTS version. 
>
> By the way, the radixtree_uri may have better performance than 
radixtree_host_uri, so if we really change the default HTTP router, we 
also need to give users a hint that: if they want better performance 
or you are making a stress test, you may change the HTTP router back 
to the radixtree_uri.

I think to implement the change, we need to close the tasks below:

- [ ] Coding for changing the default URI, making all the test cases passed
- [ ] Write docs for claiming the broken change
- [ ] Write docs for giving users a hint about the performance difference between `radixtree_host_uri` and `radixtree_uri`, give them some advice about the performance tuning

XFarooqi:
### Description

Updated the default HTTP router from radixtree_uri to radixtree_host_uri to remove the confusion that which route will hit first. Let's say, we have two routes, the first one requires the host matches *.example.com, and the URI path is /anything, And the second one needs the host matches foo.example.com exactly, and the URI path is also /anything. In such a case, if we send an API request, we may hit the first route, which is counterintuitive.

This is also helpful as most of the users have been in confusion since they think Apache APISIX will consider the HTTP host match by default. Also updated the documentation reflecting the result of these changes. i.e, Comparatively low performance can be expected in radixtree_host_uri

Fixes #8354

### Checklist

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -75,7 +75,7 @@ apisix:
   # Turn this option on if you want to be compatible with servlet when matching URI path.
   normalize_uri_like_servlet: false
   router:
-    http: radixtree_uri         # radixtree_uri: match route by uri(base on radixtree)
+    http: radixtree_host_uri      # radixtree_uri: match route by uri(base on radixtree)
                                   # radixtree_host_uri: match route by host + uri(base on radixtree)
                                   # radixtree_uri_with_parameter: like radixtree_uri but match uri with parameters,
                                   #   see https://github.com/api7/lua-resty-radixtree/#parameters-in-path for
--- a/docs/en/latest/terminology/router.md
+++ b/docs/en/latest/terminology/router.md
@@ -44,7 +44,8 @@ A Router can have the following configurations:
     - `match priority`: First try an absolute match, if it didn't match, try prefix matching.
     - `Any filter attribute`: This allows you to specify any Nginx built-in variable as a filter, such as URL request parameters, request headers, and cookies.
   - `radixtree_uri_with_parameter`: Like `radixtree_uri` but also supports parameter match.
-  - `radixtree_host_uri`: Matches both host and URI of the request. Use `host + uri` as the primary index (based on the `radixtree` engine).
+  - `radixtree_host_uri`: Matches both host and URI of the request. Use `host + uri` as the primary index (based on the `radixtree` engine). It may lag in performance as compared to redixtree_uri.
+For better performance or stress test, you can change the HTTP router back to the radixtree_uri.
 
 - `apisix.router.ssl`: SSL loads the matching route.
   - `radixtree_sni`: (Default) Use `SNI` (Server Name Indication) as the primary index (based on the radixtree engine).


XFarooqi:
@tokers  
Can you please help me with these failing tests, maybe some guidance?

tokers:
I think some of them failed due to the router was changed and must add more match conditions for routes in these cases.

XFarooqi:
Can you please help me more about the match conditions, I have no good idea as I am very new to this repo. 
But I am ready to complete the task.

tokers:
> 

Please read more about the [what is route](https://apisix.apache.org/docs/apisix/terminology/route/).

membphis:
Because you modified the default router, it will cause some test cases to fail.

So you also need to change the test case at the same time, specify `radixtree_uri` for the failure case.

Boburmirzo:
> Because you modified the default router, it will cause some test cases to fail.
> 
> So you also need to change the test case at the same time, specify `radixtree_uri` for the failure case.

@membphis Could we give to Abdul useful links to any guidelines or tutorials on how to test things on Apache APISIX?

I think it is quite hard and risky to modify existing tests without any knowledge. If we do not have any learning materials, we should prepare for new joiners.

spacewander:
> > Because you modified the default router, it will cause some test cases to fail.
> > So you also need to change the test case at the same time, specify `radixtree_uri` for the failure case.
> 
> @membphis Could we give to Abdul useful links to any guidelines or tutorials on how to test things on Apache APISIX?
> 
> I think it is quite hard and risky to modify existing tests without any knowledge. If we do not have any learning materials, we should prepare for new joiners.

Here it is: https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md

XFarooqi:
> > > Because you modified the default router, it will cause some test cases to fail.
> > > So you also need to change the test case at the same time, specify `radixtree_uri` for the failure case.
> > 
> > 
> > @membphis Could we give to Abdul useful links to any guidelines or tutorials on how to test things on Apache APISIX?
> > I think it is quite hard and risky to modify existing tests without any knowledge. If we do not have any learning materials, we should prepare for new joiners.
> 
> Here it is: https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md

Thank you @spacewander for providing me with the link. I respect your time.

Because I have no prior experience if writing test cases, I am having the following issues in writing test cases:
1. APISIX uses test-nginx framework for testing and its documentation itself is complex to understand.
2. Reading the documentation, I understand how to write test cases, etc but I am not able to find out what changes should I have to make in failing test cases.
3. In the summary of failing test cases, there is some test that is shown as failed but when I see that file then it doesn't include those test cases. For example in `t/plugin/ai.`, Failed Tests are  [3, 9, 12, 15, 19, 22, 25, 28, 41, 45-46] but when I check this file it has only 12 test cases.
4.  In the log description of **Failed Test 4 of `'t/plugin/ai4.t`** there are so many things that I am totally unable to understand.
Screenshot 2023-01-29 at 1 06 54 PM


I genuinely want to contribute to APISIX and solve this issue and I will highly appreciate it if you can guide me with step-by-step procedures for the issue. Because it looks like I am running in a loop where I am learning everything individually but not able to connect them with others.

spacewander:
The questions you asked are more relative to the test-nginx framework. Maybe you can get some help from their side and spend more time learning about it.

PR #7401, opened 2022-07-06
feat(proxy-rewrite): create use_real_request_uri_unsafe option

ilteriseroglu-ty:
### Description

This PR fixes #7386.

As per the linked issue:

> During our tests we have noticed that APISIX always uses the normalized/decoded URI coming from nginx (as per its design, which isn't an issue), and this creates an edge case where a backend that _requires_ URL Encoded URIs to not work properly (i.e. GitLab API calls that reference paths inside a project use `%2F` instead of `/`).

This PR implements an option named `use_real_request_uri_unsafe` to utilize `real_request_uri` (which is originally `request_uri` in nginx; its the request uri that's _not_ normalized) and bypass all encoding calls inside.

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #8564, opened 2022-12-23
fix(proxy-rewrite): fix url normalization bypass

ilteriseroglu-ty:
### Description

This PR fixes a bug where ctx.var.upstream_uri is not set when use_real_request_uri_unsafe is enabled.

It seems that somewhere in the runtime chain (LuaJIT?) a memory jump occurs and ctx.var.upstream_uri _does_ get set even though it definitely **shouldn't**. Hence why this got unnoticed in the CI tests of #7401.

### Checklist

### DIFF
--- a/apisix/plugins/proxy-rewrite.lua
+++ b/apisix/plugins/proxy-rewrite.lua
@@ -297,6 +297,8 @@ do
         else
             ctx.var.upstream_uri = upstream_uri
         end
+    else
+        ctx.var.upstream_uri = upstream_uri
     end
 
     if conf.headers then


apisix/plugins/proxy-rewrite.lua
```diff
@@ -297,6 +297,8 @@ do
         else
             ctx.var.upstream_uri = upstream_uri
         end
+    else
+        ctx.var.upstream_uri = upstream_uri
```
spacewander, line 301: If `ctx.var.upstream_uri` is empty, Nginx will proxy the request with the client's path. Why do we need to set the uri here?

ilteriseroglu-ty: To be quite honest, I don't know. _But_, a route to a GitLab API backend running at couple thousand requests/s has almost 50 percent error rate without this change because the request uri gets normalized _somewhere_ in the execution chain, even though it shouldn't be because `use_real_request_uri_unsafe` is set.

For instance `/api/v4/projects/12345/repository/files/technology%2Ftest.yaml?ref=stage` will be normalized down to `/api/v4/projects/12345/repository/files/technology/test.yaml?ref=stage` without this change, which shouldn't happen as this route has `use_real_request_uri_unsafe` enabled (which we did confirm).

tokers: Do you have reproducible steps for this?

spacewander: > To be quite honest, I don't know. _But_, a route to a GitLab API backend running at couple thousand requests/s has almost 50 percent error rate without this change because the request uri gets normalized _somewhere_ in the execution chain, even though it shouldn't be because `use_real_request_uri_unsafe` is set.
> 
> For instance `/api/v4/projects/12345/repository/files/technology%2Ftest.yaml?ref=stage` will be normalized down to `/api/v4/projects/12345/repository/files/technology/test.yaml?ref=stage` without this change, which shouldn't happen as this route has `use_real_request_uri_unsafe` enabled (which we did confirm).

Interesting. AFAIK, the "missing" assignment here is intended.
Because `upstream_uri` here is equal to `ctx.var.uri`:
https://github.com/ilteriseroglu-ty/apisix/blob/f6d05266852e8a46b8ef060e249c05c577a926f0/apisix/plugins/proxy-rewrite.lua#L263

While `ctx.var.uri` is normalized (see http://nginx.org/en/docs/http/ngx_http_core_module.html#var_uri), assigning it to `ctx.var.upstream_uri` will break the `use_real_request_uri_unsafe`.

Before use_real_request_uri_unsafe is added, we always assign `upstream_uri` to `ctx.var.upstream_uri`:
https://github.com/apache/apisix/blob/964a92ee773264999eb67d0cd5539baf8961001b/apisix/plugins/proxy-rewrite.lua#L171-L179

That's why we don't assign it anymore when use_real_request_uri_unsafe is set.

ilteriseroglu-ty: If `use_real_request_uri_unsafe` gets set, the local `upstream_uri` becomes `ctx.var.real_request_uri`:

https://github.com/apache/apisix/blob/0f1391693bf98a38c7716b1091f40181de8690bb/apisix/plugins/proxy-rewrite.lua#L263-L265

`ctx.var.real_request_uri` is used instead of `ctx.var.request_uri` as the nginx native `request_uri` is rewritten in init.lua to use the normalized version instead:

https://github.com/apache/apisix/blob/5822eca9a28ff7b92dc7fc0431281eb62dd60ba8/apisix/init.lua#L509-L513

Normally, everything should work *without* this PR (as I authored the original `use_real_request_uri_unsafe` change) but, after examining an issue where this didn't work half the time when the system was under load; I had to set `ctx.var.upstream_uri` to `upstream_uri`, which is set to `ctx.var.real_request_uri` because `use_real_request_uri_unsafe` is set.

I am not sure whether this is an issue under LuaJIT or something else.

spacewander: I see. Can you add a test case to cover this? You can learn how to write test by reading https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md

PR #8567, opened 2022-12-24
feat: consumer-restriction plugin support consumer_group_id type

### LINKED ISSUE #8563: feat: As a user, I want to configure consumer_group_id in consumer-restriction plugin, so that we can limit consumers by group_id
### Description

Hello, I found that we don't support configuring `consumer_group_id` in [consumer-restriction](https://apisix.apache.org/docs/apisix/plugins/consumer-restriction/) plugins.

I think we can support configuring `consumer_group_id` in this plugin so that we can limit consumers by group id; it will be easier for users.

refer: 
https://apisix.apache.org/docs/apisix/plugins/consumer-restriction/
https://apisix.apache.org/docs/apisix/admin-api/#consumer-group

ronething:
### Description

Fixes #8563

consumer-restriction plugin add `consumer_group_id` type

### Checklist


starsz (approved):

t/plugin/consumer-restriction.t
```diff
@@ -1436,3 +1436,254 @@ passed
 GET /t
 --- response_body
 passed
+
+
+
+=== TEST 54: create consumer group(group1)
```
spacewander, line 1442: Could you move the new tests to a new file?

> When the test file is too large, for example > 800 lines, you should split it to a new file. Please take a look at t/plugin/limit-conn.t and t/plugin/limit-conn2.t.

https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#check-code-style-and-test-case-style

ronething: done.

starsz (approved):

spacewander (approved):

PR #8578, opened 2022-12-28
feat: limit count plugin support X-RateLimit-Reset

### LINKED ISSUE #8381: feat: As a user of limit-count, I want to communicate the X-RateLimit-Reset header so my api consumers know when their next allocation window starts, this should be communicated on every response even the rejected response.
### Description

- limit-count plugin does not communicate the `X-RateLimit-Reset` header only `X-RateLimit-Remaining` and `X-RateLimit-Limit`
- There are no `X-RateLimit-*` headers communicated on a rejected response, here I believe it can be improved to include the `X-RateLimit-Reset` header as per draft [RFC](https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-reset). This header would also be helpful on response of non-rejected(successful) requests 

The improvement of this approach would be that consumers of an API know when they can attempt another API call without having to "poll" for sucessful responses with Retries.

mscb402:
### Description

limit count plugin support returns a new header `X-RateLimit-Reset` and return the header when Rejected

Fixes #8381 

### Checklist


apisix/plugins/limit-count/init.lua
```diff
@@ -300,9 +327,28 @@ function _M.rate_limit(conf, ctx)
         return 500, {error_msg = "failed to limit count"}
     end
 
+    local reset
+    if remaining == conf.count - 1 then
+        -- set an end time
+        local end_time = ngx_time() + conf.time_window
+        -- save to lrucache by key
+        reset = conf.time_window
+        local end_time_obj = get_end_time(conf,key)
```
spacewander, line 336: By default, the client IP will be used as key, which means there will be a great number of keys per conf. The size of lrucache is limited.

There come two solutions:
1. store the end time info in the actual backend like redis cluster.
2. write a new cache with dynamical expire support to store the end time info, but this way is not memory-effective.

Personally, I prefer solution 1.

docs/en/latest/plugins/limit-count.md
```diff
@@ -254,7 +254,7 @@ curl -i http://127.0.0.1:9180/apisix/admin/routes/1 \
 
 ## Example usage
 
-The above configuration limits to 2 requests in 60 seconds. The first two requests will work and the response headers will contain the headers `X-RateLimit-Limit` and `X-RateLimit-Remaining`:
+The above configuration limits to 2 requests in 60 seconds. The first two requests will work and the response headers will contain the headers `X-RateLimit-Limit` and `X-RateLimit-Remaining` and `X-RateLimit-Reset`:
```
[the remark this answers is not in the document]
mscb402: fixed

apisix/plugins/limit-count/init.lua
```diff
@@ -242,6 +245,15 @@ local function gen_limit_obj(conf, ctx)
     return core.lrucache.plugin_ctx(lrucache, ctx, extra_key, create_limit_obj, conf)
 end
 
+local function get_end_time(conf,key)
+    local create = function()
```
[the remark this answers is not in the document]
mscb402: fixed

apisix/plugins/limit-count/init.lua
```diff
@@ -242,6 +245,15 @@ local function gen_limit_obj(conf, ctx)
...
+        return {
+            endtime=0,
+        }
+    end
+
+    return resetlru(key,"",create,conf)
```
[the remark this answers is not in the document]
mscb402: done

apisix/plugins/limit-count/init.lua
mscb402: solution 1 does not work either. Because by default limit count will not use Redis as the backend. So we can't save it to Redis.

mscb402: By default, There use `resty.limit.count` this package for limit traffic. And I read the code of Resty Limit Count package using `ngx.shared.DICT` to save data. Can I use this?

mscb402: I'm using `ngx.shared` with expire time instead.

spacewander: The `ngx.shared` only works with the default backend which also uses `ngx.shared`. If we use other backends, only one instance will execute `remaining == conf.count - 1` and the end time can only be found in that instance.

mscb402: Fixed, you can review it again.

apisix/plugins/limit-count/limit-count-local.lua
```diff
@@ -0,0 +1,64 @@
...
+local assert = assert
+local setmetatable = setmetatable
+
+local _M = {}
+
+local mt = {
+    __index = _M
+}
+
+function _M.set_endtime(self,key,time_window)
```
spacewander, line 29: Look like we can merge `set_endtime` into the `incoming` method, especially after implementing `read_reset` via eval?

apisix/cli/ngx_tpl.lua
```diff
@@ -276,6 +276,7 @@ http {
     {% if enabled_plugins["limit-count"] then %}
     lua_shared_dict plugin-limit-count {* http.lua_shared_dict["plugin-limit-count"] *};
     lua_shared_dict plugin-limit-count-redis-cluster-slot-lock {* http.lua_shared_dict["plugin-limit-count-redis-cluster-slot-lock"] *};
+    lua_shared_dict plugin-limit-count-reset-header {* http.lua_shared_dict["plugin-limit-count-reset-header"] *};
```
spacewander, line 279: Maybe we can let the size of plugin-limit-count-reset-header be equal to plugin-limit-count, so that people don't need to configure it twice? As these two shdict store similar data.

mscb402: I removed `plugin-limit-count-reset-header` in config

apisix/plugins/limit-count/limit-count-local.lua
mscb402: ok, fixed

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -84,6 +84,23 @@ function _M.new(plugin_name, limit, window, conf)
...
 end
 
+function _M.set_endtime(self,key,time_window)
+    return time_window
+end
+
+function _M.read_reset(self, key)
+    local red = self.red_cli
+    key = self.plugin_name .. tostring(key)
+    local ttl, err = red:ttl(key)
```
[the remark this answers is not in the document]
mscb402: fixed

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -85,27 +70,80 @@ function _M.incoming(self, key)
...
+    local self = {
+        limit = limit,
+        window = window,
+        conf = conf,
+        plugin_name = plugin_name,
+    }
+    return setmetatable(self, mt)
+end
+
+function _M.set_endtime(self,key,time_window)
```
spacewander, line 88: We don't need to expose these methods now?

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -91,16 +92,18 @@ function _M.incoming(self, key)
     local window = self.window
     key = self.plugin_name .. tostring(key)
 
-    local remaining, err = red:eval(script, 1, key, limit, window)
+    local res, err = red:eval(script, 1, key, limit, window)
+    local remaining = res[1]
```
spacewander, line 96: Please use the res after checking the err

t/cli/test_main.sh
```diff
@@ -807,6 +807,7 @@ nginx_config:
       internal-status: 20m
       plugin-limit-req: 20m
       plugin-limit-count: 20m
+      plugin-limit-count-reset-header: 20m
```
spacewander, line 810: We don't need this setting, right?

apisix/plugins/limit-count/limit-count-local.lua
```diff
@@ -0,0 +1,76 @@
...
+    local self = {
+        limit_count = limit_local_new(plugin_name, limit, window, conf),
+        dict = ngx.shared["plugin-limit-count-reset-header"]
+    }
+
+    return setmetatable(self, mt)
+end
+
+function _M.incoming(self, key, commit, conf)
+    local delay, remaining =  self.limit_count:incoming(key, commit)
```
spacewander, line 61:
```suggestion
    local delay, remaining = self.limit_count:incoming(key, commit)
```

t/cli/test_main.sh
mscb402: Yes, I forget to remove

mscb402: Now I'm removed this. Pls review again

apisix/plugins/limit-count/limit-count-local.lua
mscb402: Done

apisix/plugins/limit-count/limit-count-redis.lua
mscb402: yes, removed

apisix/plugins/limit-count/limit-count-redis-cluster.lua
mscb402: fixed

apisix/plugins/limit-count/init.lua
```diff
@@ -302,7 +309,8 @@ function _M.rate_limit(conf, ctx)
 
     if conf.show_limit_quota_header then
         core.response.set_header("X-RateLimit-Limit", conf.count,
-            "X-RateLimit-Remaining", remaining)
+                "X-RateLimit-Remaining", remaining,
```
spacewander, line 312: Why change 4-spaces indentation to 8-spaces here?

apisix/plugins/limit-count/init.lua
```diff
@@ -283,10 +283,17 @@ function _M.rate_limit(conf, ctx)
...
 
-    local delay, remaining = lim:incoming(key, true)
+    local delay, remaining, reset = lim:incoming(key, true, conf)
     if not delay then
         local err = remaining
         if err == "rejected" then
+            -- show count limit header when rejected
+            if conf.show_limit_quota_header then
+                core.response.set_header("X-RateLimit-Limit", conf.count,
+                        "X-RateLimit-Remaining", 0,
```
spacewander, line 293: Ditto

apisix/plugins/limit-count/limit-count-local.lua
```diff
@@ -0,0 +1,76 @@
...
+
+local mt = {
+    __index = _M
+}
+
+local function set_endtime(self, key, time_window)
+    -- set an end time
+    local end_time = ngx_time() + time_window
+    -- save to dict by key
+    self.dict:set(key, end_time, time_window)
```
spacewander, line 33: Better to check the err returned from `set`.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -85,27 +70,52 @@ function _M.incoming(self, key)
...
+        conf = conf,
+        plugin_name = plugin_name,
+    }
+    return setmetatable(self, mt)
+end
+
+function _M.incoming(self, key)
+    local conf = self.conf
+    local red, err = redis_cli(conf)
+    if err then
```
spacewander, line 91:
```suggestion
    if not red then
```

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -85,27 +70,52 @@ function _M.incoming(self, key)
...
+        return red, err
```
spacewander, line 92:
```suggestion
        return nil, err, 0
```
Please follow the code style: https://github.com/apache/apisix/blob/master/CODE_STYLE.md#error-handling

mscb402: fixed

apisix/plugins/limit-count/limit-count-redis.lua
mscb402: fixed

apisix/plugins/limit-count/init.lua
mscb402: fixed

apisix/plugins/limit-count/init.lua
mscb402: fixed

apisix/plugins/limit-count/limit-count-local.lua
mscb402: fixed

spacewander (approved):

tokers (approved):

Issue #8511, opened 2022-12-13
request help: the format of response body that returned by plugin should match Content-Type


tzssangglass:
**Step 1**
Create a route  with consumer restriction
```
curl --location --request PUT ".../apisix/admin/routes/content-type" \
--header "X-API-KEY: edd1c9f034335f136f87ad84b625c8f1" \
--header "Content-Type: application/json" \
--data-raw "{
    \"methods\": [
        \"GET\"
    ],
    \"plugins\": {
        \"consumer-restriction\": {
            \"whitelist\": [
                \"nobody\"
            ]
        }
    },
    \"uri\": \"/content-type/get\"
}"
```
**Step 2** Try to consume the route without being identified to raise an error
I said i want application/json response :
```
curl --location --request GET ".../content-type/get" \
--header "Accept: application/json"
```

Response is :
```
# http code
401
# body 
{"message":"Missing authentication or identity verification."}
# In headers : 
Content-Type:  text/plain; charset=utf-8
```
The Content-Type header should be : application/json as it is in fact a Json in response.

This issue is not on "consumer-restriction"plugin only, it also happen for the key-auth plugin when you provide a wrong key, and i guess it is a .general plugin error behavior.
I  think it should exist a general common way to respond error from plugin to allow manage right content-type depending the response format.
For example in case of a request like : 
```
curl --location --request GET "..../content-type/get" \
--header "Accept: application/xml"
```
i expect a response like 

```
# body 
<message>Missing authentication or identity verification.</message>
# In headers : 
Content-Type:  application/xml; charset=utf-8
```
---
In case of  unknow accept or not managed, here we go for text/plain :
```
--header "Accept: */*"
````
Response should be something like :
```
# body 
Missing authentication or identity verification.
# In headers : 
Content-Type:  text/plain; charset=utf-8
```

_Originally posted by @MekelCon in https://github.com/apache/apisix/discussions/8504#discussioncomment-4386548_

tzssangglass:
@MekelCon thanks your feedback, I have confirmed this.

maybe this is a good first issue. cc @spacewander

tzssangglass:
> Maybe adding a field `error_message_type` in the `plugin._meta`.

maybe fix in `core.response.exit()`?

spacewander:
`core.response.exit()` +1, use the right type according to the client's Accept header if possible.

MekelCon:
I can try to do it i think

spacewander:
@MekelCon 
Assign to you

[...]

PR #8587, opened 2022-12-29
fix: issue 8511 The format of response body that returned by plugin should match Content-Type

MekelCon:
### Description
In the response.lua, improve management of the content type of the response.
If the response is a json, and if the original request accept it, we set the content-type to "application/json".
In other case we keep the old response type (text/plain)

Fixes [8511](https://github.com/apache/apisix/issues/8511)

### Checklist


spacewander:

MekelCon:

Should be done now !

apisix/core/response.lua
```diff
@@ -80,7 +115,20 @@ function resp_exit(code, ...)
     end
 
     if idx > 0 then
-        ngx_print(concat_tab(t, "", 1, idx))
+        local response = concat_tab(t, "", 1, idx)
+        local _, err = decode_json(response)
+        if not err then
+            local content_type = ngx_header["Content-type"]
+            local accept_header = ngx.req.get_headers()["Accept"]
```
spacewander, line 122: We can use https://github.com/apache/apisix/blob/c0c906eb4d62886ffa4b0938b017dc4810cfe639/apisix/core/request.lua#L104 which gets the header from cache

apisix/core/response.lua
```diff
@@ -80,7 +115,20 @@ function resp_exit(code, ...)
     end
 
     if idx > 0 then
-        ngx_print(concat_tab(t, "", 1, idx))
+        local response = concat_tab(t, "", 1, idx)
+        local _, err = decode_json(response)
```
spacewander, line 119: It would be more effective to check the header first before decoding, so we can save lots of unnecessary decoding.

MekelCon: Fiexed by @MekelCon
[perf: decode json later](https://github.com/apache/apisix/pull/8587/commits/b4b60d28eb1ec0b1c0078633219d846e7be839a6)

apisix/core/response.lua
MekelCon: fixed by [get Accept header from cache](https://github.com/apache/apisix/pull/8587/commits/5e439faa537a97d825699059cc71e3b17df78aa9)

t/core/response.t
```diff
@@ -199,3 +199,39 @@ aaa:
...
+GET /t
+--- error_code: 201
+--- response_body
+done
+--- response_headers
+Content-Type: text/plain
+
+
+
+=== TEST 10: exit with table then content-type is application/json
```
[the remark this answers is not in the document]
MekelCon: Fixed by [Add test : request not accepting application/json](https://github.com/apache/apisix/pull/8587/commits/e8d205a7b9d2467a72e38ee5b8e0b29b11bab23e)

MekelCon:
@spacewander I saw the CI failed due to this line, 121 :
`local accept_header = core_request.header(ctx, "Accept")`
Because ctx is 'not defined'
Wich is true, but it works, i am a beginner in .lua so maybe there is something i am missing.

To have the correct syntax should we add the `ctx `parameter to the method resp_exit ?

spacewander:
> @spacewander I saw the CI failed due to this line, 121 : `local accept_header = core_request.header(ctx, "Accept")` Because `ctx` is 'not defined' Wich is true, but it works, i am a beginner in .lua so maybe there is something i am missing.
> 
> To have the correct syntax should we add the `ctx `parameter to the method resp_exit ?

I think a bit it seems that there is no better way to do it. It's OK to add the parameter.

MekelCon:
> > @spacewander I saw the CI failed due to this line, 121 : `local accept_header = core_request.header(ctx, "Accept")` Because `ctx` is 'not defined' Wich is true, but it works, i am a beginner in .lua so maybe there is something i am missing.
> > To have the correct syntax should we add the `ctx `parameter to the method resp_exit ?
> 
> I think a bit it seems that there is no better way to do it. It's OK to add the parameter.


@spacewander , 
so i checked quickly, 
i see at least 93 call to `response.exit(.......)`
SO before doing the update i want to be sure of what need to done.

As today we use this function with this definition : 
`function resp_exit(code, ...)` 

i must update it to :
`function resp_exit(ctx, code, ...)`

And then it means check all the usage to add the parameter ? For example
`return core.response.exit(502)`
will become
`return core.response.exit(ctx, 502)`

spacewander:
> > > @spacewander I saw the CI failed due to this line, 121 : `local accept_header = core_request.header(ctx, "Accept")` Because `ctx` is 'not defined' Wich is true, but it works, i am a beginner in .lua so maybe there is something i am missing.
> > > To have the correct syntax should we add the `ctx `parameter to the method resp_exit ?
> > 
> > 
> > I think a bit it seems that there is no better way to do it. It's OK to add the parameter.
> 
> @spacewander , so i checked quickly, i see at least 93 call to `response.exit(.......)` SO before doing the update i want to be sure of what need to be done.
> 
> As today we use this function with this definition : `function resp_exit(code, ...)`
> 
> i must update it to : `function resp_exit(ctx, code, ...)`
> 
> And then it means check all the usage to add the parameter ? For example `return core.response.exit(502)` will become `return core.response.exit(ctx, 502)`

Yes, you are right

MekelCon:
> > > > @spacewander I saw the CI failed due to this line, 121 : `local accept_header = core_request.header(ctx, "Accept")` Because `ctx` is 'not defined' Wich is true, but it works, i am a beginner in .lua so maybe there is something i am missing.
> > > > To have the correct syntax should we add the `ctx `parameter to the method resp_exit ?
> > > 
> > > 
> > > I think a bit it seems that there is no better way to do it. It's OK to add the parameter.
> > 
> > 
> > @spacewander , so i checked quickly, i see at least 93 call to `response.exit(.......)` SO before doing the update i want to be sure of what need to be done.
> > As today we use this function with this definition : `function resp_exit(code, ...)`
> > i must update it to : `function resp_exit(ctx, code, ...)`
> > And then it means check all the usage to add the parameter ? For example `return core.response.exit(502)` will become `return core.response.exit(ctx, 502)`
> 
> Yes, you are right

1st try done, i must say i am not very confident about what i did :/
At least i can confirm Apisix still run locally, and my test are passing.

apisix/init.lua
```diff
@@ -372,7 +372,7 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
         local upstream = apisix_upstream.get_by_id(up_id)
         if not upstream then
             if is_http then
-                return core.response.exit(502)
+                return core.response.exit(core.ctx, 502)
```
spacewander, line 375: Sorry, what does the core.ctx here mean? It should be `ctx` (or sometimes called `api_ctx`).

soulbird:
Hi, @MekelCon Will you continue to complete this PR ?

MekelCon:
> Hi, @MekelCon Will you continue to complete this PR ?

Yes i will !

PR #7517, opened 2022-07-21
feat(clickhouse-logger): support multi clickhouse endpoints

zhendongcmss:
### Description

When the apisix load is heavy, there are many access logs, and the endpoint load of a single clickhouse is too heavy. The clickhouse-alogger plugin supports multiple endpoints to share the load.

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #8604, opened 2023-01-04
feat(elasticsearch-logger): support multi elasticsearch endpoints

xiaoxuanzi:
### Description

When the apisix load is heavy, there are many logs, and the endpoint load of a single elasticsearch is too heavy. The elasticsearch-logger plugin supports multiple endpoints to share the load.

Fixes  (#8431)

### Checklist

### DIFF
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -23,18 +23,35 @@ local plugin          = require("apisix.plugin")
 
 local ngx             = ngx
 local str_format      = core.string.format
+local math_random     = math.random
+local type            = type
 
 local plugin_name = "elasticsearch-logger"
 local batch_processor_manager = bp_manager_mod.new(plugin_name)
 
 
-local schema = {
-    type = "object",
-    properties = {
-        endpoint_addr = {
+local endpoint_schema = {
+    anyOf = {
+        {
+            -- deprecated, use "array" instead
             type = "string",
             pattern = "[^/]$",
         },
+        {
+            type = "array",
+            minItems = 1,
+            items = {
+                type = "string",
+                pattern = "[^/]$",
+            },
+        }
+    }
+}
+
+local schema = {
+    type = "object",
+    properties = {
+        endpoint_addr = endpoint_schema,
         field = {
             type = "object",
             properties = {
@@ -68,7 +85,7 @@ local schema = {
         }
     },
     encrypt_fields = {"auth.password"},
-    required = { "endpoint_addr", "field" },
+    required = {"endpoint_addr", "field"},
 }
 
 
@@ -127,7 +144,13 @@ local function send_to_elasticsearch(conf, entries)
         return false, str_format("create http error: %s", err)
     end
 
-    local uri = conf.endpoint_addr .. "/_bulk"
+    local selected_endpoint_addr
+    if type(conf.endpoint_addr) == "string" then
+        selected_endpoint_addr = conf.endpoint_addr
+    else
+        selected_endpoint_addr = conf.endpoint_addr[math_random(#conf.endpoint_addr)]
+    end
+    local uri = selected_endpoint_addr .. "/_bulk"
     local body = core.table.concat(entries, "")
     local headers = {["Content-Type"] = "application/x-ndjson"}
     if conf.auth then
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -24,34 +24,27 @@ local plugin          = require("apisix.plugin")
 local ngx             = ngx
 local str_format      = core.string.format
 local math_random     = math.random
-local type            = type
 
 local plugin_name = "elasticsearch-logger"
 local batch_processor_manager = bp_manager_mod.new(plugin_name)
 
 
-local endpoint_schema = {
-    anyOf = {
-        {
-            -- deprecated, use "array" instead
+local schema = {
+    type = "object",
+    properties = {
+        -- deprecated, use "endpoint_addrs" instead
+        endpoint_addr = {
             type = "string",
             pattern = "[^/]$",
         },
-        {
+        endpoint_addrs = {
             type = "array",
             minItems = 1,
             items = {
                 type = "string",
                 pattern = "[^/]$",
             },
-        }
-    }
-}
-
-local schema = {
-    type = "object",
-    properties = {
-        endpoint_addr = endpoint_schema,
+        },
         field = {
             type = "object",
             properties = {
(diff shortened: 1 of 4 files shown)


apisix/plugins/elasticsearch-logger.lua
```diff
@@ -31,10 +32,19 @@ local batch_processor_manager = bp_manager_mod.new(plugin_name)
 local schema = {
     type = "object",
     properties = {
+        -- deprecated, use "endpoint_addrs" instead
         endpoint_addr = {
             type = "string",
             pattern = "[^/]$",
         },
+        endpoint_addrs = {
```
soulbird, line 40: Extending endpoint_addr in a compatible way, and adding a new configuration is not a good choice

xiaoxuanzi: Thank you for your reply.

What is your suggestion?
Can you give an example?
I refer to PR #7517

> feat(clickhouse-logger): support multi clickhouse endpoints

soulbird: Looking at this, it also indicates the need to unify into one. You can refer to the definition of nodes_schema to allow endpoint_addrs to be set to multiple types: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L289

xiaoxuanzi: > Looking at this, it also indicates the need to unify into one. You can refer to the definition of nodes_schema to allow endpoint_addrs to be set to multiple types: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L289

Thanks for the helpful advice!

xiaoxuanzi: Hi @soulbird. 
I fixed the code according to your suggestion. please review it.

soulbird (dismissed):

docs/en/latest/plugins/elasticsearch-logger.md
```diff
@@ -53,6 +53,8 @@ This Plugin supports using batch processors to aggregate and process entries (lo
 
 ## Enabling the Plugin
 
+If multiple endpoints are configured, they will be written randomly.
```
spacewander, line 56: This should be put into the attribute table

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -23,18 +23,35 @@ local plugin          = require("apisix.plugin")
...
+                pattern = "[^/]$",
+            },
+        }
+    }
+}
+
+local schema = {
+    type = "object",
+    properties = {
+        endpoint_addr = endpoint_schema,
```
spacewander, line 54: Better to use `endpoint_addrs` with the new type.
Using different types in the same field brings us lots of trouble when we need to work with static-type language, like writing a Go client.

docs/en/latest/plugins/elasticsearch-logger.md
```diff
@@ -37,7 +37,7 @@ When the Plugin is enabled, APISIX will serialize the request context informatio
 
 | Name          | Type    | Required | Default                     | Description                                                  |
 | ------------- | ------- | -------- | --------------------------- | ------------------------------------------------------------ |
-| endpoint_addr | string  | True     |                             | Elasticsearch API.                                            |
+| endpoint_addr | string/array  | True     |                             | Elasticsearch API.                                            |
```
spacewander, line 40: Let's add a deprecated mark like https://github.com/apache/apisix/blob/release/2.15/docs/en/latest/plugins/mqtt-proxy.md#attributes

t/plugin/elasticsearch-logger.t
```diff
@@ -515,3 +515,39 @@ apisix:
...
+            })
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
```
spacewander, line 553: We need to add a test to show that different endpoints will be chosen randomly.

apisix/plugins/elasticsearch-logger.lua
xiaoxuanzi: Hi @spacewander ,  thank you for your reply. 
In my opinion, to use **endpoint_addrs** will cause trouble, this change is not backward compatible.

t/plugin/elasticsearch-logger.t
xiaoxuanzi: I agree with you, but I don't know how.
Can you give an example?
I take  PR #7517 as reference.
 [=== TEST 5: add plugin on routes using multi clickhouse-logger](https://github.com/apache/apisix/blob/9c899b87648039e2c6c73b087d627aea2df90f55/t/plugin/clickhouse-logger.t#L175)

apisix/plugins/elasticsearch-logger.lua
soulbird: Keep endpoint_addr, add a new field endpoint_addrs to support both string and array better

spacewander: > this change is not backward compatible

I am confused. People can still use the old endpoint_addr configuration.

> Keep endpoint_addr, add a new field endpoint_addrs to support both string and array better

It is not a good idea to have a field of multiple types. The upstream's nodes field is an example. People keep complaining that this field doesn't have a consistent type.

t/plugin/elasticsearch-logger.t
spacewander: What about hitting the path multiple times and triggering multiple sending?
Like https://github.com/apache/apisix/blob/master/docs/en/latest/internal/testing-framework.md#send-request
Then we can check if all the endpoints are used via error log.

apisix/plugins/elasticsearch-logger.lua
soulbird: Understood. So we need to keep both endpoint_addr and endpoint_addrs, where endpoint_addrs is only specified as array type, is that right?

spacewander: Yes

t/plugin/elasticsearch-logger.t
xiaoxuanzi: Hi @spacewander.
I fixed the code according to your suggestion. please review it.

t/plugin/elasticsearch-logger.t
```diff
@@ -551,3 +551,32 @@ PTQvJEaPcNOXcOHeErC0XQ==
...
+                local code, body = t('/hello', ngx.HTTP_GET)
+                ngx.say("code: ", code)
+            end
+
+        }
+    }
+--- response_body
+code: 200
+code: 200
+code: 200
```
spacewander, line 572: We can count the code like https://github.com/apache/apisix/blob/22be6d08cef7de16c3f8a303e75c1872378c3dee/t/node/healthcheck2.t#L139

xiaoxuanzi: Done.

spacewander (approved):

PR #8611, opened 2023-01-05
refactor(admin): refactor resource routes

### LINKED ISSUE #8569: feat: As a user, I want to refactor the resources under apisix.admin, so that we don't need to repeat
### Description

CC @An-DJ

An-DJ:
### Description

Fixes #8569 

### Checklist

### DIFF
--- a/apisix/admin/resource.lua
+++ b/apisix/admin/resource.lua
@@ -31,7 +31,7 @@ local mt = {
 }
 
 
-local function get(self, id)
+function _M:get(id)
     local key = "/" .. self.name
     if id then
         key = key .. "/" .. id
@@ -39,7 +39,7 @@ local function get(self, id)
 
     local res, err = core.etcd.get(key, not id)
     if not res then
-        core.log.error("failed to get " .. self.kind .. "[", key, "] from etcd: ", err)
+        core.log.error("failed to get ", self.kind, "[", key, "] from etcd: ", err)
         return 503, {error_msg = err}
     end
 
@@ -48,8 +48,8 @@ local function get(self, id)
 end
 
 
-local function post(self, check_conf, id, conf, sub_path, args)
-    local id, err = check_conf(id, conf, false)
+function _M:post(id, conf, sub_path, args)
+    local id, err = self.check_conf(id, conf, false)
     if not id then
         return 400, err
     end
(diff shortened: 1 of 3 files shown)


apisix/admin/routes.lua
```diff
@@ -168,143 +162,4 @@ local function check_conf(id, conf, need_id)
...
-        core.log.error("failed to set new route[", key, "] to etcd: ", err)
-        return 503, {error_msg = err}
-    end
-
-    return res.status, res.body
-end
-
-
-return _M
+return resource.new("routes", "route", check_conf)
```
spacewander, line 165: I think about it a bit.
Maybe it would be better to pass an opt like:
```
{
        name = name,
        kind = kind,
        check_conf = check_conf
    }
```

so people don't need to remember the order.

apisix/admin/routes.lua
```diff
@@ -17,19 +17,13 @@
...
 local loadstring = loadstring
 
 
-local _M = {
-    version = 0.2,
-    need_v3_filter = true,
-}
-
-
 local function check_conf(id, conf, need_id)
```
spacewander, line 27: Look like we can also share the id check logic? It is the same in other resources.

apisix/admin/routes.lua
```diff
@@ -168,143 +134,9 @@ local function check_conf(id, conf, need_id)
...
-    return res.status, res.body
-end
-
-
-return _M
+return resource.new({
+    name = "routes",
+    kind = "route",
+    schema = core.schema.route,
+    check_conf_self = check_conf
```
spacewander, line 141: check_conf_self is not a good name. It would be better to use a valid English verb phrase as the method name.

apisix/admin/resource.lua
```diff
@@ -0,0 +1,217 @@
...
+        core.log.error("failed to set new ", self.kind, "[", key, "] to etcd: ", err)
+        return 503, {error_msg = err}
+    end
+
+    return res.status, res.body
+end
+
+
+function _M.new(opt)
+    return setmetatable({
```
spacewander, line 208:
```suggestion
    return setmetatable(opt
```
can be simpler

apisix/admin/routes.lua
```diff
@@ -17,39 +17,13 @@
...
-    end
-
-    if need_id and conf.id and tostring(conf.id) ~= tostring(id) then
-        return nil, {error_msg = "wrong route id"}
-    end
-
-    conf.id = id
-
     core.log.info("schema: ", core.json.delay_encode(core.schema.route))
     core.log.info("conf  : ", core.json.delay_encode(conf))
```
spacewander, line 28: We can move this log to the common part?

spacewander (approved):

soulbird (approved):

PR #8620, opened 2023-01-05
fix: Compression in log-rotate plugin exceeds the default timeout of shell.run

### LINKED ISSUE #8619: bug: Compression in log-rotate plugin exceeds the default timeout of shell.run
### Current Behavior

When rotating log files, the log-rotate plugin uses shell.run() function which has a default timeout of 10 seconds. When rotating big log files this timeout is exceeded thus the tar command is being killed with SIGTERM. This causes the rotated files to be incomplete, and also the log files doesn't get removed.

image

### Expected Behavior

Logs should be rotated completely unless it takes more time than log rotate interval. Therefor the timeout of the shell.run() function should be the same as the interval value.

### Error Logs

_No response_

### Steps to Reproduce

Generate a log file (1GB or more) which takes more than 10 seconds to compress using `tar -cvf`
Let the log rotate plugin work

### Environment

- APISIX 2.15 but should be the same on v3

alptugay:
- fix Compression in log-rotate plugin exceeds the default timeout of shell.run

### Description

This issue fixes the bug: "Compression in log-rotate plugin exceeds the default timeout of shell.run" #8619
The interval parameter is passed to shell.run() function

Fixes #8619

### Checklist


apisix/plugins/log-rotate.lua
```diff
@@ -277,6 +277,7 @@ local function rotate()
         max_size = attr.max_size or max_size
         enable_compression = attr.enable_compression or enable_compression
     end
+    local timeout = interval * 100 -- Timeout for compression is interval in milliseconds
```
membphis, line 280: We'd better add a new configuration item to control the timeout parameter.

image

alptugay: I don't agree that timeout should be a user defined parameter. Because, setting a timeout greater than interval may cause undefined behaviour (eg. second rotation starts but first compression is still running), and setting the value too low also causes the mentioned undesired behaviour (so doesn't fix the issue in the first place). So, this value should be automatically set based on interval and shoud not be user defined imho.

shreemaan-abhishek: @alptugay, I agree with you, however setting a fixed timeout is inappropriate. What if the compression starts just before the next rotation?

The correct way would be to provide a timeout such that timeout occurs before the next log-rotation. And if the compression still gets a timeout, we can reschedule the compression later using ngx_timer. WDYT?

alptugay: > What if the compression starts just before the next rotation?
Rotation is done in intervals. Execution step is like below in simple terms:
1) Rename current log file to logfile.date
2) Send signal to process so new file opens 
3) compress logfile.date with a timeout that is not bigger than the interval,
So I don't think step 3 will start just before next rotation. 

But your second argument is valid, compression can still get a timeout if it can't be finished in the specified interval. We can schedule it with ngx_timer but this means those timers will add up. Assume the interval is 5 minutes and compression takes 10 minutes, after 1 hour we will have done 12 intervals jobs but only 6 compression will be completed and 6 will be waiting. So this will add up and probably make the system unresponsive

shreemaan-abhishek: Makes sense. LGTM.

shreemaan-abhishek (approved):
@lingsamuel, @monkeyDluffy6017, this PR LGTM. Please review and approve.

apisix/plugins/log-rotate.lua
monkeyDluffy6017, line 280: Is it typo?
```suggestion
    local timeout = interval * 1000 -- Timeout for compression is interval in milliseconds
```

monkeyDluffy6017: I suggest adding a configuration to control the timeout parameter, the default timeout is 10s, it's enough for most people, if you use interval as timeout, it could be 1min or 1 hour, maybe someone don't need such long time.

Sn0rt (approved):

apisix/plugins/log-rotate.lua
Sn0rt: @alptugay PTAL

moonming:
@shreemaan-abhishek  please fix https://github.com/apache/apisix/pull/8620#discussion_r1271983717 if it's a typo, then merge this PR, thanks

monkeyDluffy6017:
@shreemaan-abhishek Pay attention to this comment: https://github.com/apache/apisix/pull/8620#discussion_r1271996324

shreemaan-abhishek:
@alptugay can I take over this PR? You can invite me to collaborate on your forked repo.

Revolyssup:
@alptugay are you still on it?

Revolyssup:
I will take over from here.

Revolyssup:
@alptugay Can you give me permission to push on your master branch?

monkeyDluffy6017:
@Revolyssup Could you add this attribute in 'config-default.yaml' and add some comments

Revolyssup:
@monkeyDluffy6017 Added in the config-default.yaml

monkeyDluffy6017 (approved):

AlinsRan (approved):

PR #8640, opened 2023-01-09
feat: stream subsystem support kubernetes service discovery

### LINKED ISSUE #7779: feat: As a user, I want to support service discovery in the stream subsystem
### Description

As a user, I want to support service discovery in the stream subsystem, so that we can use it to get L4 upstream dynamically.

There is already an unfinished closed PR to add this feature: https://github.com/apache/apisix/pull/4488.
However, this PR is closed last year and the code has been changed a lot since that. So we can borrow some code from that PR but can't continue based on it.

ronething:
### Description

Fixes #7779 

because of kubernetes service discovery use lua_shared_dict, so i add `discovery_shared_dicts` with `stream` suffix to stream subsystem in `ngx_tpl.lua`.

for test case, i use `extra_stream_config` to get nodes and return response.

because of openresty version 1.19.3.2 do not implemented ngx.process API for stream subsystem, so ci failed. see https://github.com/apache/apisix/actions/runs/3874260736/jobs/6605244559

i change openresty version from 1.19.3.2 to 1.19.9.1 to solve this problem, refer: https://openresty.org/en/changelog-1019009.html

### Checklist


apisix/discovery/kubernetes/init.lua
```diff
@@ -434,6 +442,9 @@ local function multiple_mode_init(confs)
         end
 
         local endpoint_dict = ngx.shared["kubernetes-" .. id]
+        if not is_http then
+            endpoint_dict = ngx.shared["kubernetes-" .. id .. "-stream"]
+        end
```
spacewander, line 447: It would be better to refactor these three changes into a function, instead of three if-else blocks.

ci/linux_openresty_1_19_runner.sh
```diff
@@ -17,5 +17,5 @@
 #
 
 
-export OPENRESTY_VERSION=1.19.3.2
+export OPENRESTY_VERSION=1.19.9.1
```
spacewander, line 20: We have promised to support the latest three versions of OpenResty, so we still need to cover 1.19.3.2.
What about updating the test case and only running it when the version is matched, like https://github.com/apache/apisix/blob/release/2.14/t/plugin/ext-plugin/sanity-openresty-1-19.t?

It would be better to refuse to start when the OpenResty is too old and k8s discovery is used in the stream.

ronething: > We have promised to support the latest three versions of OpenResty, so we still need to cover 1.19.3.2.
What about updating the test case and only running it when the version is matched, like https://github.com/apache/apisix/blob/release/2.14/t/plugin/ext-plugin/sanity-openresty-1-19.t?

ok, i will try it later.

> It would be better to refuse to start when the OpenResty is too old and k8s discovery is used in the stream.

Is it a good choice if we add a boolean field like `enable_stream`(default is `false`) into `apisix/discovery/kubernetes/schema.lua` as a property and check OpenResty version if `enable_stream` is `true`? Can you give me some hints? thanks.

apisix/discovery/kubernetes/init.lua
```diff
@@ -331,9 +332,18 @@ local function start_fetch(handle)
     ngx.timer.at(0, timer_runner)
 end
 
+local function get_endpoint_dict(shm)
```
spacewander, line 335: We can pass the id as the argument, and put the constant `kubernetes` in this function? There is no need to repeat it.

ci/linux_openresty_1_19_runner.sh
spacewander: something like https://github.com/apache/apisix/blob/84535aff0e5bca575202df45833197958b93fe40/apisix/upstream.lua#L425

ronething: @spacewander 

- first, it will cause err like below when start apisix with openresty 1.19.3.2, because `local discovery = require("apisix.discovery.init").discovery` in `apisix/upstream.lua:19` will import `discovery/kubernetes/init.lua`.

```
$ ./bin/apisix start
/usr/local/openresty-debug//luajit/bin/luajit ./apisix/cli/apisix.lua start
nginx: [error] init_by_lua error: /usr/local/openresty-debug/lualib/ngx/process.lua:5: unsupported subsystem: stream
stack traceback:
        [C]: in function 'error'
        /usr/local/openresty-debug/lualib/resty/core/base.lua:223: in function 'allows_subsystem'
        /usr/local/openresty-debug/lualib/ngx/process.lua:5: in main chunk
        [C]: in function 'require'
        ...ng/Documents/apisix/apisix/discovery/kubernetes/init.lua:29: in main chunk
        [C]: in function 'require'
        /home/ronething/Documents/apisix/apisix/discovery/init.lua:28: in main chunk
        [C]: in function 'require'
        /home/ronething/Documents/apisix/apisix/upstream.lua:19: in main chunk
        [C]: in function 'require'
        /home/ronething/Documents/apisix/apisix/http/service.lua:18: in main chunk
        [C]: in function 'require'
        /home/ronething/Documents/apisix/apisix/init.lua:35: in main chunk
        [C]: in function 'require'
        init_by_lua:3: in main chunk
```

- then, something like `if conf.pass_host == "node" and conf.nodes and ` in `check_upstream_conf` function may only print error log when check upstream if `in_dp` is true, or cause err when post/put upstream or other resource which need to check upstream_conf both in http and stream subsystem. But we need to refuse to start apisix when the OpenResty is too old and k8s discovery is used in the stream?

so maybe we should consider other solutions? if something i was wrong, please tell me, thanks.

spacewander: > But we need to refuse to start apisix when the OpenResty is too old and k8s discovery is used in the stream?

Yes. Unlike the upstream conf which is a runtime conf, the wrong discovery conf in the yaml can't be corrected without restarting APISIX.

And we need to provide a clear error message so people don't need to debug it.

apisix/discovery/kubernetes/init.lua
```diff
@@ -520,6 +520,12 @@ end
 
 
 function _M.init_worker()
+    if not support_process then
+        core.log.error("kubernetes discovery not support in subsystem: "
+                       .. ngx.config.subsystem
```
spacewander, line 525: We can use ',' here

ronething: @spacewander do you mean something like below? `.. ngx.config.subsystem .. ", "`

```lua
if not support_process then
    core.log.error("kubernetes discovery not support in subsystem: "
                   .. ngx.config.subsystem .. ", "
                   .. "please check if your openresty version >= 1.19.9.1 or not")
    return
end
```

spacewander: Err. What I mean is that we can use ',' instead of '..'.

ronething: ok, i see. I will change it later.

spacewander (approved):

ronething:
@tokers @soulbird Could you please take a look, thanks.

soulbird (approved):

starsz (approved):

PR #8651, opened 2023-01-10
fix: fix fetch all service info from consul

Fabriceli:
### Description

Fixes # (issue)
As I mentioned in the email: https://lists.apache.org/thread/x0ksxx003tgqy4833vmsyq1mwwyc6wz7

### Checklist


t/discovery/consul_dump.t
```diff
@@ -75,9 +75,11 @@ discovery:
   consul:
     servers:
       - "http://127.0.0.1:8500"
+    skip_services:
+      - "consul"
     dump:
       path: "consul.dump"
-      load_on_init: true
+      load_on_init: false
```
spacewander, line 82: Why change this value?

t/discovery/consul_dump.t
```diff
@@ -161,6 +165,8 @@ discovery:
   consul:
     servers:
       - "http://127.0.0.1:38500"
+    skip_services:
+      - "consul"
```
spacewander, line 169: Do we need to add it as the default value?

apisix/discovery/consul/init.lua
```diff
@@ -237,53 +208,80 @@ function _M.connect(premature, consul_server, retry_delay)
...
     if consul_server.index ~= watch_result.headers['X-Consul-Index'] then
+        local up_services = core.table.new(0, #watch_result.body)
+        local consul_client_svc = resty_consul:new({
+            host = consul_server.host,
+            port = consul_server.port,
+            connect_timeout = consul_server.connect_timeout,
+            read_timeout = consul_server.read_timeout,
+        })
+        for service_name, _ in pairs(watch_result.body) do
+            -- check is skip service
```
spacewander, line 223:
```suggestion
            -- check if the service_name is 'skip service'
```

apisix/discovery/consul/init.lua
```diff
@@ -237,53 +208,80 @@ function _M.connect(premature, consul_server, retry_delay)
     log.info("connect consul: ", consul_server.consul_server_url,
             ", watch_result status: ", watch_result.status,
             ", watch_result.headers.index: ", watch_result.headers['X-Consul-Index'],
-            ", consul_server.index: ", consul_server.index,
-            ", consul_server: ", json_delay_encode(consul_server, true))
+            ", consul_server.index: ", consul_server.index)
```
spacewander, line 211: Would you get rid of the 8 spaces indentation in the new code?

apisix/discovery/consul/init.lua
```diff
@@ -237,53 +208,80 @@ function _M.connect(premature, consul_server, retry_delay)
...
 
     -- if current index different last index then update service
     if consul_server.index ~= watch_result.headers['X-Consul-Index'] then
+        local up_services = core.table.new(0, #watch_result.body)
+        local consul_client_svc = resty_consul:new({
+            host = consul_server.host,
+            port = consul_server.port,
+            connect_timeout = consul_server.connect_timeout,
+            read_timeout = consul_server.read_timeout,
```
spacewander, line 220: Missing default_args?

t/discovery/consul_dump.t
Fabriceli: for the first time to start apisix, it will print some error log. Just for better reading log

apisix/discovery/consul/init.lua
Fabriceli: it is not needed, because `default_args` which has `index` arg, that will make **List Nodes for Service** got the wrong result.

> [Consul API Blocking Queries](https://developer.hashicorp.com/consul/api-docs/features/blocking)

apisix/discovery/consul/init.lua
Fabriceli: Deleted by mistake, Undo it

apisix/discovery/consul/init.lua
Fabriceli: done

t/discovery/consul_dump.t
Fabriceli: the `consul` will register itself

spacewander: So the `consul` will be always registered automatically? If this is true, I think it would be better to skip by default. Otherwise, the users need to do this themselves.

Fabriceli: Yes, consul will be registered by default. I dont think it would be better if we skip it by default. We can give users to choice whether to skip or not, that can be closed-loop control. Of course, we should mention this info in Consul doc

spacewander: Is it a good idea? If consul is not skipped by default, people will need to repeat the same configuration each time they need to use consul discovery, similar to the `proxy_http_version 1.1` in the Nginx conf.

Fabriceli: Does the following scenario exist?
User A want to set up Consul route in APISIX with the Consul discovery plugin

spacewander: If you insist that people won't need to handle this case, it's OK.
But we need to write it down in the doc at least.

Fabriceli: OK, I add the `consul` as default skip service, and write it down in the doc

apisix/discovery/consul/init.lua
```diff
@@ -242,48 +215,76 @@ function _M.connect(premature, consul_server, retry_delay)
...
+                goto CONTINUE
+            end
+            -- get node from service
+            local svc_url = consul_server.consul_sub_url .. "/" .. service_name
+            local result, err = consul_client_svc:get(svc_url)
+            local error_info = (err ~= nil and err) or
+                    ((result ~= nil and result.status ~= 200) and result.status)
+            if error_info then
+                log.error("connect consul: ", consul_server.consul_server_url,
+                        " by svc url: ", svc_url, ", with error: ", error_info)
```
spacewander, line 237: Would you like to fix the indent when adding them as new code? 
Thanks!

Fabriceli: done

Fabriceli:
@spacewander Please take a look at this PR, thanks

spacewander:
https://github.com/apache/apisix/pull/8651#discussion_r1073064191
It seems there are still indentation issues. Some places use 8-spaces indentation instead of 4.

Fabriceli:
> [#8651 (comment)](https://github.com/apache/apisix/pull/8651#discussion_r1073064191) It seems there are still indentation issues. Some places use 8-spaces indentation instead of 4.

I checked the code and `make lint` cant get WARN and ERROR, could you give some hint about this? I break a long code line to multiple lines, it will be made like follow by default:
```lua
log.error("connect consul: ", consul_server.consul_server_url,
        ", by service url: ", svc_url, ", with error: ", error_info)
```

spacewander:
> > [#8651 (comment)](https://github.com/apache/apisix/pull/8651#discussion_r1073064191) It seems there are still indentation issues. Some places use 8-spaces indentation instead of 4.
> 
> I checked the code and `make lint` cant get WARN and ERROR, could you give some hint about this? I break a long code line to multiple lines, it will be made like follow by default:
> 
> ```lua
> log.error("connect consul: ", consul_server.consul_server_url,
>         ", by service url: ", svc_url, ", with error: ", error_info)
> ```

The style guide tells us that we need to use 4-spaces indentation: https://github.com/apache/apisix/blob/master/CODE_STYLE.md#indentation

Some of the newly added code use 8-spaces indentation.

Another acceptable indentation is aligning the '(', like:
```
log.error("connect consul: ", consul_server.consul_server_url,
          ", by service url: ", svc_url, ", with error: ", error_info)
```

It is widely used in the APISIX source code.

I have tried several lints/fmtters. Unfortunately, they can't handle the second case. Therefore, so far we don't have an automatical mechanism to detect the indentation issue.

Fabriceli:
> > > [#8651 (comment)](https://github.com/apache/apisix/pull/8651#discussion_r1073064191) It seems there are still indentation issues. Some places use 8-spaces indentation instead of 4.
> > 
> > 
> > I checked the code and `make lint` cant get WARN and ERROR, could you give some hint about this? I break a long code line to multiple lines, it will be made like follow by default:
> > ```lua
> > log.error("connect consul: ", consul_server.consul_server_url,
> >         ", by service url: ", svc_url, ", with error: ", error_info)
> > ```
> 
> The style guide tells us that we need to use 4-spaces indentation: https://github.com/apache/apisix/blob/master/CODE_STYLE.md#indentation
> 
> Some of the newly added code use 8-spaces indentation.
> 
> Another acceptable indentation is aligning the '(', like:
> 
> ```
> log.error("connect consul: ", consul_server.consul_server_url,
>           ", by service url: ", svc_url, ", with error: ", error_info)
> ```
> 
> It is widely used in the APISIX source code.
> 
> I have tried several lints/fmtters. Unfortunately, they can't handle the second case. Therefore, so far we don't have an automatical mechanism to detect the indentation issue.

Done, please re-check it, thanks

spacewander (approved):

Fabriceli:
@tzssangglass could you have a look at this PR, thanks

Fabriceli:
@spacewander Could you request another committer to do the code review for this PR, thanks

soulbird (approved):

PR #8661, opened 2023-01-12
refactor(admin): stream_routes/upstreams/protos/services/global_rules/consumer_groups/plugin_configs

### LINKED ISSUE #8569: feat: As a user, I want to refactor the resources under apisix.admin, so that we don't need to repeat

An-DJ:
### Description

Fixes #8569 

- [X] stream_routes
- [X] upstreams
- [X] protos
- [X] services
- [X] global_rules
- [X] consumer_groups
- [X] plugin_configs
### Checklist

### DIFF
--- a/apisix/admin/consumer_group.lua
+++ b/apisix/admin/consumer_group.lua
@@ -16,40 +16,15 @@
 --
 local core = require("apisix.core")
 local consumers = require("apisix.consumer").consumers
-local utils = require("apisix.admin.utils")
+local resource = require("apisix.admin.resource")
 local schema_plugin = require("apisix.admin.plugins").check_schema
 local type = type
 local tostring = tostring
 local ipairs = ipairs
 
 
-local _M = {
-    need_v3_filter = true,
-}
-
-
-local function check_conf(id, conf, need_id)
-    if not conf then
-        return nil, {error_msg = "missing configurations"}
-    end
-
-    id = id or conf.id
-    if need_id and not id then
-        return nil, {error_msg = "missing id"}
-    end
-
-    if not need_id and id then
-        return nil, {error_msg = "wrong id, do not need it"}
-    end
-
-    if need_id and conf.id and tostring(conf.id) ~= tostring(id) then
-        return nil, {error_msg = "wrong id"}
-    end
-
-    conf.id = id
-
-    core.log.info("conf: ", core.json.delay_encode(conf))
-    local ok, err = core.schema.check(core.schema.consumer_group, conf)
+local function check_conf(id, conf, need_id, schema)
+    local ok, err = core.schema.check(schema, conf)
     if not ok then
         return nil, {error_msg = "invalid configuration: " .. err}
     end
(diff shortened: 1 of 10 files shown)


apisix/admin/consumer_group.lua
```diff
@@ -59,54 +34,11 @@ local function check_conf(id, conf, need_id)
...
-    end
-    local res, err = core.etcd.get(key, not id)
-    if not res then
-        core.log.error("failed to get consumer group[", key, "]: ", err)
-        return 503, {error_msg = err}
-    end
-
-    utils.fix_count(res.body, id)
-    return res.status, res.body
+    return need_id and id or true
```
spacewander, line 37: Look like we can move this line to resource.lua?

apisix/admin/routes.lua
```diff
@@ -143,5 +143,6 @@ return resource.new({
     name = "routes",
     kind = "route",
     schema = core.schema.route,
-    checker = check_conf
+    checker = check_conf,
+    unsupported_methods = {}
```
spacewander, line 147: We can leave this nil as this resource doesn't have unsupported methods.

apisix/admin/services.lua
```diff
@@ -189,75 +99,15 @@ function _M.delete(id)
...
 end
 
 
-return _M
+return resource.new({
+    name = "services",
+    kind = "service",
+    schema = core.schema.service,
+    checker = check_conf,
+    unsupported_methods = {},
```
spacewander, line 111: Ditto

apisix/admin/resource.lua
```diff
@@ -57,7 +57,13 @@ function _M:check_conf(id, conf, need_id)
     core.log.info("conf  : ", core.json.delay_encode(conf))
 
     -- check the resource own rules
-    return self.checker(id, conf, need_id, self.schema)
+    local ok, err = self.checker(id, conf, need_id, self.schema)
+
+    if err then
```
spacewander, line 62: Code style, please use `if not ok then`

spacewander (approved):

apisix/admin/init.lua
```diff
@@ -199,7 +199,17 @@ local function run()
     end
 
     local code, data
-    if seg_res == "routes" then
+    local refactored_resources = {
```
soulbird, line 202: directly named as `resources`?

apisix/admin/resource.lua
```diff
@@ -57,11 +57,21 @@ function _M:check_conf(id, conf, need_id)
...
+    if not ok then
+        return ok, err
+    else
+        return need_id and id or true
+    end
 end
 
 
 function _M:get(id)
+    if self.unsupported_methods and core.table.array_find(self.unsupported_methods, "get") then
```
soulbird, line 71: Is it enough to use `if core.table.array_find(self.unsupported_methods, "get") then` ?

apisix/admin/init.lua
An-DJ: There has been a global variable named `resources` 😭. Maybe we can rename that variable after all resources are refactored.

apisix/admin/resource.lua
An-DJ: @soulbird Done.

apisix/admin/init.lua
soulbird: ok

soulbird (approved):

spacewander (approved):

PR #8672, opened 2023-01-12
ci: kafka should register port in the zookeeper same as exposed

starsz:
### Description

This PR is used to fix the CI of `t/plugin/kafka-logger.t`.
1. Since the kafka will register host:port in the zookeeper, so we should keep the same port as exposed, else
we will get the error log like this:

image

2. Add `no-error-log` flag to check to ensure the access_log is sent to `kafka` successfully.

### Checklist


ronething:
https://github.com/apache/apisix/actions/runs/3935628692/jobs/6731487434, this is the failed job link which cause something like `connection refused`.

nic-chen (approved):
LGTM

tokers (approved):

spacewander (approved):

PR #8681, opened 2023-01-13
feat(file-logger): support standard output for file logger plugins

### LINKED ISSUE #6797: bug: cannot use file-logger to log to stdout
### Current Behavior

When using the file-logger plugin, it is possible to send the logging to stdout by defining /dev/stdout. This in order to use the docker output and relais this in kubernetes to a ELK stack. Though an error is thrown indicating a permission denied for the current user.

### Expected Behavior

Be able to use /dev/stdout as a destination with the file-logger plugin.

### Error Logs

failed to open file: /dev/stdout, error info: /dev/stdout: Permission denied while logging request

### Steps to Reproduce

Run apisix docker with apisix.yaml:
```
routes:
  - uri: /*
    upstream_id: 1
    plugins:
      file-logger:
        path: /dev/stdout

upstreams:
  - id: 1
    nodes:
      "host.docker.internal:3000": 1
    type: roundrobin

plugins:
  - name: file-logger
#END
```

### Environment

- APISIX Docker version: apache/apisix:2.13.0-alpine
- Host system: Macbook Pro Intel with docker desktop for mac

e1ijah1:
### Description

support standard output for file logger plugins

Fixes #6797 

### Checklist

### DIFF
--- a/apisix/cli/env.lua
+++ b/apisix/cli/env.lua
@@ -20,6 +20,7 @@ local util = require("apisix.cli.util")
 
 local pcall = pcall
 local error = error
+local getenv = os.getenv
 local exit = os.exit
 local stderr = io.stderr
 local str_find = string.find
@@ -89,6 +90,12 @@ return function (apisix_home, pkg_cpath_org, pkg_path_org)
     local openresty_args = openresty_path_abs .. [[ -p ]] .. apisix_home .. [[ -c ]]
                            .. apisix_home .. [[/conf/nginx.conf]]
 
+    local apisix_daemon = true
+    if getenv("APISIX_DAEMON_OFF") then
+        apisix_daemon  = false
+        openresty_args = openresty_args .. [[ -g ]] .. "'daemon off;'"
+    end
+
     local or_info, err = util.execute_cmd("openresty -V 2>&1")
     if not or_info then
         error("failed to exec cmd \'openresty -V 2>&1\', err: " .. err)
@@ -104,6 +111,7 @@ return function (apisix_home, pkg_cpath_org, pkg_path_org)
     return {
         apisix_home = apisix_home,
         is_root_path = is_root_path,
+        apisix_daemon = apisix_daemon,
         openresty_args = openresty_args,
         openresty_info = or_info,
         use_apisix_base = use_apisix_base,
--- a/apisix/cli/env.lua
+++ b/apisix/cli/env.lua
@@ -20,7 +20,6 @@ local util = require("apisix.cli.util")
 
 local pcall = pcall
 local error = error
-local getenv = os.getenv
 local exit = os.exit
 local stderr = io.stderr
 local str_find = string.find
@@ -90,12 +89,6 @@ return function (apisix_home, pkg_cpath_org, pkg_path_org)
     local openresty_args = openresty_path_abs .. [[ -p ]] .. apisix_home .. [[ -c ]]
                            .. apisix_home .. [[/conf/nginx.conf]]
 
-    local apisix_daemon = true
-    if getenv("APISIX_DAEMON_OFF") then
-        apisix_daemon  = false
-        openresty_args = openresty_args .. [[ -g ]] .. "'daemon off;'"
-    end
-
     local or_info, err = util.execute_cmd("openresty -V 2>&1")
     if not or_info then
         error("failed to exec cmd \'openresty -V 2>&1\', err: " .. err)
@@ -111,7 +104,6 @@ return function (apisix_home, pkg_cpath_org, pkg_path_org)
     return {
         apisix_home = apisix_home,
         is_root_path = is_root_path,
-        apisix_daemon = apisix_daemon,
         openresty_args = openresty_args,
         openresty_info = or_info,
         use_apisix_base = use_apisix_base,
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -825,7 +825,10 @@ local function start(env, ...)
         init_etcd(env, args)
     end
 
-    util.execute_cmd(env.openresty_args, not env.apisix_daemon)
+    local data, err = util.execute_cmd(env.openresty_args)
+    if not err and data and data ~= "" then
+        print(data)
+    end
 end
 
 
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -825,7 +825,7 @@ local function start(env, ...)
         init_etcd(env, args)
     end
 
-    util.execute_cmd(env.openresty_args)
+    util.execute_cmd(env.openresty_args, not env.apisix_daemon)
 end
 
 
--- a/apisix/cli/util.lua
+++ b/apisix/cli/util.lua
@@ -20,8 +20,8 @@ local pcall = pcall
 local open = io.open
 local popen = io.popen
 local exit = os.exit
-local stdout = io.stdout
 local stderr = io.stderr
+local stdout = io.stdout
 local str_format = string.format
 local tonumber = tonumber
 
@@ -30,32 +30,19 @@ local _M = {}
 
 -- Note: The `execute_cmd` return value will have a line break at the end,
 -- it is recommended to use the `trim` function to handle the return value.
-local function execute_cmd(cmd, print_output)
+local function execute_cmd(cmd)
     local t, err = popen(cmd)
     if not t then
         return nil, "failed to execute command: "
                     .. cmd .. ", error info: " .. err
     end
 
-    local data, c
-    if print_output then
-        stdout:setvbuf("no")
-        data = ""
-        repeat
-            c = t:read(1)
-            if c then
-                stdout:write(c)
-                data = data .. c
-            end
-        until not c
-        t:close()
-    else
-        data, err = t:read("*all")
-        t:close()
-        if not data then
-            return nil, "failed to read execution result of: "
-                        .. cmd .. ", error info: " .. err
-        end
+    local data, err = t:read("*all")
+    t:close()
+
+    if not data then
+        return nil, "failed to read execution result of: "
+                    .. cmd .. ", error info: " .. err
     end
 
     return data
--- a/apisix/cli/util.lua
+++ b/apisix/cli/util.lua
@@ -20,6 +20,7 @@ local pcall = pcall
 local open = io.open
 local popen = io.popen
 local exit = os.exit
+local stdout = io.stdout
 local stderr = io.stderr
 local str_format = string.format
 local tonumber = tonumber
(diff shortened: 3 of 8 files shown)


apisix/plugins/file-logger.lua
```diff
@@ -30,6 +31,7 @@ local schema = {
         path = {
             type = "string"
         },
+        log_to_std = {type = "boolean", default = false},
```
tokers, line 34:
```suggestion
        log_to_stdout = {type = "boolean", default = false},
```

`std` is too short.

apisix/plugins/file-logger.lua
```diff
@@ -136,6 +138,20 @@ local function write_file_data(conf, log_message)
             core.log.error("failed to write file: ", conf.path, ", error info: ", err)
         end
 
+        if conf.log_to_std then
+            local std_out, err = io_open(std_out_file, "w")
```
tokers, line 142: File open/close should be handled unified, which means you should open the stdout device in the `else` block of line #125.

e1ijah1: Before opening the file, we must check whether log to stdout is enabled. I am concerned that moving the file open logic above will complicate error handling.

e1ijah1: fixed

apisix/plugins/file-logger.lua
```diff
@@ -138,10 +138,10 @@ local function write_file_data(conf, log_message)
             core.log.error("failed to write file: ", conf.path, ", error info: ", err)
         end
 
-        if conf.log_to_std then
-            local std_out, err = io_open(std_out_file, "w")
-            std_out:setvbuf("no")
+        if conf.log_to_stdout then
```
tokers: This check should be in advance because here you open the `conf.path` and then close it finally. It's meaningless.

e1ijah1: I already modified this part of the code yesterday, please check the latest version.

apisix/plugins/file-logger.lua
```diff
@@ -14,23 +14,23 @@
...
+local std_out_file = "/dev/stdout"
 
 local schema = {
     type = "object",
     properties = {
         path = {
             type = "string"
         },
-        include_resp_body = {type = "boolean", default = false},
+        log_to_stdout = { type = "boolean", default = false },
```
spacewander, line 32: Would it be better to use a special value to indicate the stdout like Nginx (https://nginx.org/en/docs/ngx_core_module.html#error_log)?
So we don't need an extra configuration to shadow the path.

apisix/plugins/file-logger.lua
```diff
@@ -14,23 +14,23 @@
...
+local log_util = require("apisix.utils.log-util")
+local core = require("apisix.core")
+local ngx = ngx
+local io_open = io.open
 local is_apisix_or, process = pcall(require, "resty.apisix.process")
 
-
 local plugin_name = "file-logger"
-
+local std_out_file = "/dev/stdout"
```
spacewander, line 24: We should use Lua's writing to stdout feature, instead of Linux's special file system (which is broken in other OS).

spacewander (commented):
Can you add a test in https://github.com/apache/apisix/tree/master/t/cli?

apisix/plugins/file-logger.lua
e1ijah1: Thanks for catching that, I've already updated it.

e1ijah1:
> Can you add a test in https://github.com/apache/apisix/tree/master/t/cli?

I'm not quite sure how to capture the standard output in E2E testing. Could you please give me some hints if you're available?

spacewander:
> > Can you add a test in [`master`/t/cli](https://github.com/apache/apisix/tree/master/t/cli?rgh-link-date=2023-01-16T09%3A58%3A02Z)?
> 
> I'm not quite sure how to capture the standard output in E2E testing. Could you please give me some hints if you're available? @spacewander

You can add a custom configuration to let APISIX don't run as a daemon. See https://github.com/apache/apisix/blob/master/docs/en/latest/customize-nginx-configuration.md

apisix/plugins/file-logger.lua
```diff
@@ -14,23 +14,23 @@
...
-local ngx          =   ngx
-local io_open      =   io.open
+local log_util = require("apisix.utils.log-util")
+local core = require("apisix.core")
+local ngx = ngx
+local io_open = io.open
+local io_output = io.output
 local is_apisix_or, process = pcall(require, "resty.apisix.process")
 
-
```
spacewander, line 23: Please don't remove the blank line

apisix/plugins/file-logger.lua
```diff
@@ -14,23 +14,23 @@
...
+local std_out_file = "stdout"
 
 local schema = {
     type = "object",
     properties = {
         path = {
             type = "string"
         },
-        include_resp_body = {type = "boolean", default = false},
+        include_resp_body = { type = "boolean", default = false },
```
spacewander, line 33: Please don't introduce a new style.

docs/en/latest/plugins/file-logger.md
```diff
@@ -43,10 +43,10 @@ The `file-logger` Plugin is used to push log streams to a specific location.
...
-| path | string | True     | Log file path. |
-| include_resp_body      | boolean | False     | When set to `true` includes the response body in the log file.                                                                                                                                                                |
+| Name                   | Type    | Required | Description                                                                                                                                                                                                                        |
+|------------------------|---------| -------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| path                   | string  | True     | Log file path. The special value `stdout` selects the standard output file.                                                                                                                                                        |
```
spacewander, line 48: The stdout isn't always a file

spacewander (changes requested):

apisix/plugins/file-logger.lua
e1ijah1: Sorry, this was auto-formatted by the IDE.

apisix/plugins/file-logger.lua
```diff
@@ -39,7 +41,7 @@ local schema = {
             }
         }
     },
-    required = {"path"}
+    required = { "path" }
```
spacewander, line 44: The blank line and other style issue are not addressed totally

apisix/plugins/file-logger.lua
```diff
@@ -108,19 +115,23 @@ if is_apisix_or then
             end
         end
 
+
```
spacewander, line 118: Ditto

t/cli/plugin/test_file_logger_std_output.sh
```diff
@@ -0,0 +1,75 @@
...
+
+log_file=logs/stdout.log
+
+if [ -f "$log_file" ]; then
+    rm "$log_file"
+fi
+
+make init
+
+stdbuf -o0 /usr/local/openresty/bin/openresty -p . -g 'daemon off;' > "$log_file" 2>&1 &
```
spacewander, line 30: Would you please use the suggestion in https://github.com/apache/apisix/pull/8681#issuecomment-1384864654?
We are testing APISIX - better to avoid running bare openresty.

t/cli/plugin/test_file_logger_std_output.sh
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+. ./t/cli/common.sh
+
+log_file=logs/stdout.log
+
+if [ -f "$log_file" ]; then
+    rm "$log_file"
```
spacewander, line 25: We can avoid using real file, like https://github.com/apache/apisix/blob/4131fbf641c02eb4209ffa9cddbe0bea247e990b/t/cli/test_validate_config.sh#L30

spacewander (changes requested):

t/cli/plugin/test_file_logger_std_output.sh
e1ijah1: I ran 'make run' without daemon mode, but it didn't print out standard output.

t/cli/plugin/test_file_logger_std_output.sh
spacewander: Ping @e1ijah1

t/cli/plugin/test_file_logger_std_output.sh
spacewander: We need to run APISIX directly.

t/cli/plugin/test_file_logger_std_output.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+# setup upstream server
+echo '
+apisix:
+    node_listen: 9080
```
spacewander, line 31: We don't need to specify listen port here

e1ijah1:
@spacewander It looks like the connection to the ETCD in one environment is down, causing CI to fail. Would you mind restarting the CI and reviewing my code?

apisix/cli/env.lua
```diff
@@ -89,6 +90,12 @@ return function (apisix_home, pkg_cpath_org, pkg_path_org)
     local openresty_args = openresty_path_abs .. [[ -p ]] .. apisix_home .. [[ -c ]]
                            .. apisix_home .. [[/conf/nginx.conf]]
 
+    local apisix_daemon = true
```
spacewander, line 93: As I mentioned many times in
https://github.com/apache/apisix/pull/8681#issuecomment-1384864654
https://github.com/apache/apisix/pull/8681#discussion_r1073065444

Could we add a custom configuration to let APISIX don't run as a daemon, just in the test?

apisix/cli/util.lua
```diff
@@ -38,10 +38,9 @@ local function execute_cmd(cmd)
 
     local data, err = t:read("*all")
     t:close()
-
     if not data then
         return nil, "failed to read execution result of: "
-                    .. cmd .. ", error info: " .. err
```
spacewander, line 44: Please don't remove blank line and break the indentation

t/cli/plugin/test_file_logger_std_output.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
spacewander, line 1: Is this test run in the CI?
https://github.com/apache/apisix/blob/d019da018ca8c06bf80e363a0512a30808da1b8e/ci/linux_apisix_current_luarocks_runner.sh#L73

apisix/cli/ops.lua
```diff
@@ -825,7 +825,10 @@ local function start(env, ...)
         init_etcd(env, args)
     end
 
-    util.execute_cmd(env.openresty_args)
+    local data, err = util.execute_cmd(env.openresty_args)
+    if not err and data and data ~= "" then
+        print(data)
```
spacewander, line 830: It is useless to print the access log only after the Nginx exited.

t/cli/plugin/test_file_logger_std_output.sh
```diff
@@ -0,0 +1,82 @@
...
+
```
spacewander, line 2: Could we remove the intermediate directory, so this file can be tested without modifying the ci script?

PR #8693, opened 2023-01-16
feat: support send error-log to kafka brokers

### LINKED ISSUE #8678: feat: As a user, I want to support send error-log to kafka, so that we can get error logs from kafka
### Description

Hi, folks.

Now I found that we only support sending error log to `tcp server` , `skywalking`, `clickhouse ` .
I want to support sending error log to kafka, so that we can collect `access log` and `error log` via kafka.

refer: https://apisix.apache.org/docs/apisix/2.13/plugins/error-log-logger/

ronething:
### Description

Fixes #8678 

support send error-log to kafka brokers.

- for producer via lrucache, i use `plugin_name .. "#kafka"` as key, and `metadata.modifiedIndex` as version.

```
        -- reuse producer via lrucache to avoid unbalanced partitions of messages in kafka
        prod, err = lrucache(plugin_name .. "#kafka", metadata.modifiedIndex,
                             create_producer, config.kafka.brokers, broker_config, config.kafka.cluster_name)
```

### Checklist


spacewander (changes requested):

apisix/plugins/error-log-logger.lua
```diff
@@ -272,6 +330,55 @@ local function send_to_clickhouse(log_message)
...
+    core.log.info("create new kafka producer instance")
+    return producer:new(broker_list, broker_config, cluster_name)
+end
+
+
+local function send_to_kafka(log_message)
+    core.log.info("sending a batch logs to kafka brokers: ", core.json.encode(config.kafka.brokers))
+
+    local broker_config = {}
+    broker_config["request_timeout"] = config.timeout * 1000
```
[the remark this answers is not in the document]
ronething: it come from https://github.com/apache/apisix/blob/d85295303c2b07253ec4811fcf3071a8fa913ca6/apisix/plugins/error-log-logger.lua#L72
https://github.com/apache/apisix/blob/d85295303c2b07253ec4811fcf3071a8fa913ca6/apisix/plugins/error-log-logger.lua#L168 
and set by 
https://github.com/apache/apisix/blob/d85295303c2b07253ec4811fcf3071a8fa913ca6/apisix/plugins/error-log-logger.lua#L305

ronething:
i will change code and update PR later.

apisix/plugins/error-log-logger.lua
```diff
@@ -272,6 +333,52 @@ local function send_to_clickhouse(log_message)
...
+local function send_to_kafka(log_message)
+    core.log.info("sending a batch logs to kafka brokers: ",
+                   core.json.delay_encode(config.kafka.brokers))
+
+    local broker_config = {}
+    broker_config["request_timeout"] = config.timeout * 1000
+    broker_config["producer_type"] = config.kafka.producer_type
+    broker_config["required_acks"] = config.kafka.required_acks
+
+    local metadata = plugin.plugin_metadata(plugin_name)
```
spacewander, line 351: It seems using the config passed from `process` here might create a race:

Consider we have c1(config) and m1 (modifiedIndex) in `process`, and c2/m2 in `send`. It looks like we might use m2 as key and c1 as value in the cache below.

ronething: >  It looks like we might use m2 as key and c1 as value in the cache below.

`process` and `send` only read data from plugin_metadata, this case maybe not occur?

ronething: > It looks like we might use m2 as key and c1 as value in the cache below.

if you mean that we need to clone `config` before we use it like below? then we can use m2 and c2 in `send`.

`local conf = core.table.clone(config)`

spacewander: A clone of c1 still has a c1's value.
Maybe we can get the c2 in `send` like what we have done in `process`.

ronething: ok, i will change the code and update the PR later.

ronething: @spacewander Could you please take a look, thanks.

spacewander (approved):

ronething:
@soulbird Could you please take a look, thanks.

soulbird (approved):

PR #8711, opened 2023-01-19
feat: file logger plugin support response body in variable

### LINKED ISSUE #8705: feat: As a user, I want file-logger plugin support record response body in a custom format, so that make the log format more flexible
### Description

As a user, I would like the file-logger plugin to support logging the response body in a custom format to make the logging format more flexible. 
Currently the file-logger plugin supports logging the response body, but not logging the response in the way the user wants. 
So if the user configures the log format in meta. There will be no way to log the response body. Only clear log format in meta will be enabled.
So I want to add a new var `resp_body` to let the user put it in a log format which let the user use the response in the way there want. 
What do you think about this? If you think this is ok, pls assign me to add this feature.

mscb402:
### Description

We need to add a new var resp_body to let the user put it in a log format which let the user use the response in the way there want.

Fixes #8705

### Checklist


t/plugin/file-logger3.t
```diff
@@ -0,0 +1,124 @@
+#
```
spacewander, line 1: Could we move these tests under https://github.com/apache/apisix/blob/master/t/plugin/file-logger2.t?

apisix/plugins/file-logger.lua
```diff
@@ -59,6 +59,9 @@ local _M = {
     metadata_schema = metadata_schema
 }
 
+core.ctx.register_var("resp_body", function(ctx)
```
spacewander, line 62: Would be better to put it in https://github.com/apache/apisix/blob/master/apisix/core/ctx.lua so other loggers can use it too.

docs/en/latest/plugins/file-logger.md
```diff
@@ -43,10 +43,10 @@ The `file-logger` Plugin is used to push log streams to a specific location.
...
-| ---- | ------ | -------- | ------------- |
+| Name | Type   | Required | Description                                                                                                                                                                                                                        |
+| ---- | ------ | -------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | path | string | True     | Log file path. |
-| include_resp_body      | boolean | False     | When set to `true` includes the response body in the log file.                                                                                                                                                                |
+| include_resp_body      | boolean | False     | When set to `true` includes the response body in the log file. If format is set, you can use the variable $resp_body to get the response body.                                                                                     |
```
spacewander, line 49: Would be better to put it in https://github.com/apache/apisix/blob/master/docs/en/latest/apisix-variable.md, instead of modifying each logger's doc.
Note that we need to mean that this variable is only for logger and requires the logger to have a special configuration.

apisix/plugins/file-logger.lua
mscb402: I don't think this var should be put into the `ctx.lua` file. Because the value of resp_body comes from `ctx.resp_body`, which is set by `file-logger` when include_resp_body = true.

t/plugin/file-logger3.t
mscb402: Move to file-logger2

apisix/plugins/file-logger.lua
tokers: Not only one logger plugin will support recording the response body. `file-logger.lua` is not an appropriate place.

docs/en/latest/plugins/file-logger.md
tokers: +1.

apisix/core/ctx.lua
```diff
@@ -213,6 +213,10 @@ do
         route_name = true,
         service_id = true,
         service_name = true,
+        resp_body = function(ctx)
```
spacewander, line 216: https://github.com/apache/apisix/blob/c8d5afb5b02a800bb3f7090bc5adb1d26e281d53/apisix/core/ctx.lua#L206

apisix/core/ctx.lua
```diff
@@ -213,6 +213,10 @@ do
...
+            -- only work when file-logger set include_resp_body = true
```
spacewander, line 217: The comment is incorrect. See https://github.com/apache/apisix/pull/8711#discussion_r1089879862

apisix/core/ctx.lua
mscb402: done

apisix/core/ctx.lua
mscb402: done

docs/en/latest/apisix-variable.md
```diff
@@ -49,5 +49,6 @@ additional variables.
 | service_name        | core       | Name of Service.                                                                    |                |
 | redis_cmd_line      | Redis      | The content of Redis command.                                                       |                |
 | rpc_time            | xRPC       | Time spent at the rpc request level.                                                |                |
+| resp_body           | core       | This variable is only for logger and requires the logger to have a special configuration. |                |
```
tokers, line 52: Which special configuration? I'm afraid that such a description is not useful for users to use this variable.

mscb402: I was update it

apisix/core/ctx.lua
```diff
@@ -209,10 +209,14 @@ do
...
         consumer_group_id = true,
         consumer_name = true,
+        resp_body = function(ctx)
+            -- only for logger and requires the logger to have a special configuration
+            return ctx.resp_body or ''
+        end,
         route_id = true,
         route_name = true,
         service_id = true,
-        service_name = true,
```
spacewander, line 215: Why remove the comma?

docs/zh/latest/apisix-variable.md
```diff
@@ -48,5 +48,6 @@ APISIX 除了支持 [NGINX 变量](http://nginx.org/en/docs/varindex.html)外，
 | service_name        | core       | APISIX 服务的名称。                                                               |                   |
 | redis_cmd_line      | Redis      | Redis 命令的内容。                                                                |                   |
 | rpc_time            | xRPC       | 在 RPC 请求级别所花费的时间。                                                      |                   |
+| resp_body           | core       | 此变量仅用于 Logger，并要求对应的 Logger 需要具备某些配置。                            |                  |
```
spacewander, line 51: Please put it in alphabetical order

docs/en/latest/apisix-variable.md
```diff
@@ -49,5 +49,6 @@ additional variables.
 | service_name        | core       | Name of Service.                                                                    |                |
 | redis_cmd_line      | Redis      | The content of Redis command.                                                       |                |
 | rpc_time            | xRPC       | Time spent at the rpc request level.                                                |                |
+| resp_body           | core       | This variable is only for logger and requires the logger plugin to have a special configuration like `include_resp_body`. |                |
```
spacewander, line 52: We should explain what this variable is, and what decides its value

apisix/core/ctx.lua
mscb402: I readded it

docs/en/latest/apisix-variable.md
mscb402: done

docs/zh/latest/apisix-variable.md
mscb402: done

docs/en/latest/apisix-variable.md
```diff
@@ -48,6 +48,7 @@ additional variables.
 | service_id          | core       | Id of Service.                                                                      |                |
 | service_name        | core       | Name of Service.                                                                    |                |
 | redis_cmd_line      | Redis      | The content of Redis command.                                                       |                |
+| resp_body           | core       | In the logger plugin, if some of the plugins support logging of response content, for example by configuring `include_resp_body: true`, then this variable can be used in the log format. |                |
```
spacewander, line 51:
```suggestion
| resp_body           | core       | In the logger plugin, if some of the plugins support logging of response body, for example by configuring `include_resp_body: true`, then this variable can be used in the log format. |                |
```

docs/zh/latest/apisix-variable.md
```diff
@@ -47,6 +47,8 @@ APISIX 除了支持 [NGINX 变量](http://nginx.org/en/docs/varindex.html)外，
 | service_id          | core       | APISIX 服务的 ID。                                                                |                   |
 | service_name        | core       | APISIX 服务的名称。                                                               |                   |
 | redis_cmd_line      | Redis      | Redis 命令的内容。                                                                |                   |
+| resp_body           | core       | 在 logger 插件中，如果部分插件支持记录响应内容，比如配置 `include_resp_body: true`，那可以在 log format 中使用这个变量。|                   |
 | rpc_time            | xRPC       | 在 RPC 请求级别所花费的时间。                                                      |                   |
+| resp_body           | core       | 此变量仅用于 Logger，并要求对应的 Logger 需要具备某些配置。                            |                  |
```
soulbird, line 52: repeated?

mscb402: removed

soulbird (dismissed):

soulbird (approved):

tokers (approved):

spacewander (approved):

PR #8718, opened 2023-01-22
fix(proxy-mirror): use with uri rewrite

### LINKED ISSUE #8167: bug: proxy-rewrite and proxy-mirror do not work together
### Current Behavior

The route configuration (omitted some irrelevant content):
```
{
    "uri": "/v1/hello/*",
    "plugins": {
        "proxy-mirror": {
            "sample_ratio": 1,
            "host": "http://mirror.api:8082"
        },
        "proxy-rewrite": {
            "regex_uri": [
                "/v1/(.*)",
                "/$1"
            ],
            "_meta": {
                "priority": 2000
            }
        }
    }
}
```

The client uses this URI to request:
```
GET /v1/hello
```

Requests to the upstream will be rewritten as:
```
GET /hello
```

But the request to the mirror address is still:
```
GET /v1/hello
```

**So the URI rewritten in `proxy-rewrite` does not actually take effect for `proxy-mirror`. (I have overridden the proxy-rewrite priority via `_meta.priority` and it will be executed first)**

### Expected Behavior

The modified URI in `proxy-rewrite` will take effect for `proxy-mirror`.

### Error Logs

_No response_

### Steps to Reproduce

1. Create a route by referring to the route configuration in the current behavior.
2. Request API
3. Check mirror service log

Explanation of reasons:
`proxy-rewrite` gets the URI from `ctx.var.uri`, writes it to the `upstream_uri` variable, and subsequently uses that variable instead of the original `uri` when forwarding upstream. `proxy-mirror` does not take this into account and uses the original `uri` variable directly. I think it should behave in such a way that the `uri` is used first and the `upstream_uri` is used when the value does not match the `upstream_uri`. If necessary we also need to modify the `proxy-rewrite` priority so that it takes precedence over `proxy-mirror`.

### Environment

bzp2010:
### Description

APISIX started to provide `_meta.priority` plugin field in 2.13 to help users rewrite the hard-coded plugin priority at runtime, so users can arbitrarily combine various plug-ins and use different order of precedence.

When trying to use a combination of proxy-rewrite and proxy-mirror, a logical error occurs. In the default configuration, mirror takes precedence over rewrite, and when the user rewrites the uri, it does not take effect in mirror, which is not the user's expectation, so the user can rewrite the priority.

However, when you rewrite the priority you will find that the uri rewrite still does not take effect, because the mirror plugin uses the uri field in the nginx variable directly, ignoring the upstream_uri variable that other APISIX plugins can rewrite, which is the key to uri rewriting.

We can fix this behavior by using logic such as:
- If `upstream_uri ! = nil` and `upstream_uri ! = ""` (which means that other plugins or APISIX itself has rewritten the uri), mirror uses the `upstream_uri` value as the new uri, while the existing prefix settings are not affected.
- If `upstream_uri` is empty, it follows the same logic as before and uses `var.uri` to splice args.

With this logic, any old usage is not affected (as you can see, I did not modify any of the old test cases), and this modification only takes effect when the user modifies the priority, modifies the uri in advance, or modifies the uri through other custom plugins, with minimal impact.

Fixes #8167 
Mail list: https://lists.apache.org/thread/8y3kxrwxl9yh9f049660kbc1ob8vhqkk

_BTW, I noticed that no one has raised any issues in the formal discussion channels (mail list), that thread has been there for three months. So I will assume that no one has objected to that.
And as I showed in the description above, this solution has minimal impact and does not cause any disruptive behavior._

### Checklist


t/plugin/proxy-mirror.t
```diff
@@ -838,3 +838,66 @@ GET /hello?a=1
 hello world
 --- error_log
 uri: /a/hello?a=1
+
+
+
+=== TEST 30: use proxy-rewrite to change uri before mirror
```
spacewander, line 844: Let's move it to a new file.

> When the test file is too large, for example > 800 lines, you should split it to a new file. Please take a look at t/plugin/limit-conn.t and t/plugin/limit-conn2.t.

https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#check-code-style-and-test-case-style

spacewander (approved):

soulbird (approved):

t/plugin/proxy-mirror.t
bzp2010: moved

spacewander (approved):

Issue #8670, opened 2023-01-12
feat: For clickhouse logger to support include req and response body option


gaoxingliang:
### Description

I want to track all the api requests and response to the clickhouse database.
And I found the clickhouse logger doesn't support this.
https://apisix.apache.org/docs/apisix/plugins/clickhouse-logger/
I think we only need to add two / three lines at:
https://github.com/apache/apisix/blob/master/apisix/plugins/clickhouse-logger.lua#L36
and adding:
```
        include_req_body = {type = "boolean", default = false},
        include_resp_body = {type = "boolean", default = false},
```
and change related doc stuff. (I don't know the lua script before. Just reading the code.)

An-DJ:
I found that other Loggers (such as [kafka-logger](https://apisix.apache.org/docs/apisix/plugins/kafka-logger/)) contain not only `include_req_body` and `include_resp_body`, but also `include_req_body_expr` and `include_resp_body_expr` for filtering.

It seems like a meaningful use scenario for clickhouse-logger.

CC @spacewander Should the implementation of their expr be added in the meantime?

spacewander:
LGTM

pixeldin:
This seems to be a familiar feature implemented in `file-logger`, maybe I can try it, can you assign it to me?

PR #8722, opened 2023-01-24
feat: Support clickhouse-logger plugin log option of request/response body

pixeldin:
### Description

Providing some optional setting items for clickhouse-logger to control the recording log content as requested in https://github.com/apache/apisix/issues/8670.

I refer some options of the [http-logger](https://apisix.apache.org/docs/apisix/next/plugins/http-logger/):
`include_req_body`, `include_req_body_expr`, `include_resp_body` and `include_resp_body_expr`

Fixes #([8670](https://github.com/apache/apisix/issues/8670))

### Checklist


t/plugin/clickhouse-logger2.t
```diff
@@ -0,0 +1,244 @@
...
+--- response_body
+passed
+
+
+
+=== TEST 2: collect response log
+--- request
+GET /hello
+--- error_log eval
+qr/\{.*"body":"hello world\\n"/
```
spacewander, line 81: Could you also include the `clickhouse body: ` prefix? So people can know this output is from the dummy upstream, not from somewhere else.

t/plugin/clickhouse-logger2.t
```diff
@@ -0,0 +1,244 @@
...
+--- response_body
+passed
+
+
+
+=== TEST 4: skip collect response log for condition
+--- request
+GET /hello?foo=unknown
+--- no_error_log
+"response":{"body":"hello world\n"
```
spacewander, line 137: Ditto

gaoxingliang:
@pixeldin Thanks for your work. I will upgrade to this version when it's released.

gaoxingliang:
Hi @pixeldin how can I avoid recording full response body if it's too long? (Can I truncate the response to max 1024 bytes?)

t/plugin/clickhouse-logger2.t
pixeldin: lgtm, I will cover it.

t/plugin/clickhouse-logger2.t
pixeldin: ok, I will cover it.

pixeldin:
> Hi @pixeldin how can I avoid recording full response body if it's too long? (Can I truncate the response to max 1024 bytes?)

IMO, truncating body by length may lose some data unexpected. What about trying to extract the target response body you concerned in upstream or middle layer at first?

gaoxingliang:
> > Hi @pixeldin how can I avoid recording full response body if it's too long? (Can I truncate the response to max 1024 bytes?)
> 
> IMO, truncating body by length may lose some data unexpected. What about trying to extract the target response body you concerned in upstream or middle layer at first?

I'm trying to record the request/response into clickhouse for audit purpose. while the json response maybe too large.  I think if we have some normal / quick config, then I don't need to write code to do that. (For now, I set a column `response` to store all response related data.) I don't think it's a good idea if the body in response is too large when doing olap.

t/plugin/clickhouse-logger2.t
```diff
@@ -0,0 +1,245 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+use t::APISIX 'no_plan';
+
+log_level("info");
+#repeat_each(1);
```
spacewander, line 21: We can remove it?

spacewander (dismissed):

pixeldin:
> > > Hi @pixeldin how can I avoid recording full response body if it's too long? (Can I truncate the response to max 1024 bytes?)
> > 
> > 
> > IMO, truncating body by length may lose some data unexpected. What about trying to extract the target response body you concerned in upstream or middle layer at first?
> 
> I'm trying to record the request/response into clickhouse for audit purpose. while the json response maybe too large. I think if we have some normal / quick config, then I don't need to write code to do that. (For now, I set a column `response` to store all response related data.) I don't think it's a good idea if the body in response is too large when doing olap. I also checked the operators here - https://github.com/api7/lua-resty-expr#operator-list, It doesn't provide a operator like `length`

I know your mean. As I know, there is no suggestion for me about the body length limit in this scenario. Maybe you can start and create a new issue for this.

spacewander (approved):

soulbird:
Please resolve the conflicts @pixeldin

spacewander:
Let's merge master to solve the conflict.

soulbird (approved):

spacewander (approved):

PR #8735, opened 2023-01-29
feat(ci): implement image caching to reduce ci build time.

shreemaan-abhishek:
### Description

We can use image caching to reduce the CI build time. This has been implemented in this PR. Observed results show reduce in ci time of 9-11 minutes.

cc: @soulbird 

### Checklist


.github/workflows/build.yml
```diff
@@ -99,29 +99,92 @@ jobs:
...
-          # launch deps env
-          make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
+      - name: Cache images
+        id: cache-images
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-apisix-docker-images
+        with:
+          path: docker-images-backup
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ matrix.os_name }}-${{ steps.test_env.outputs.type }}-${{ hashFiles('./ci/pod/*.yml') }}
```
soulbird, line 109: What is cached are docker images from docker-compose-${type}.yml. So, why calculate the hash value of *.yml？

.github/workflows/build.yml
```diff
@@ -99,29 +99,92 @@ jobs:
...
-          make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
-          sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
+          if [[ ${{ steps.test_env.outputs.type }} == 'first' ]]; then
+            # launch deps env
+            echo "start first service ..., time: $(date)"
+            make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
+            echo "start first service done, time: $(date)"
+          fi
+          if [[ ${{ steps.test_env.outputs.type }} == 'plugin' ]]; then
+            echo "start plugin service ..., time: $(date)"
```
soulbird, line 142: Wouldn't it be better to encapsulate these scripts into a shell file? for example
```shell
# init-plugin-test-service.sh
before() {
           sudo wget https://github.com/jacekkow/keycloak-protocol-cas/releases/download/18.0.2/keycloak-protocol-cas-18.0.2.jar -O /opt/keycloak-protocol-cas-18.0.2.jar
            bash ci/pod/openfunction/build-function-image.sh
}

after() {
            sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
            echo "start plugin service done, time: $(date)"
}
```

'first' and 'last' are similar.

.github/workflows/build.yml
shreemaan-abhishek: So that, if any of the docker compose file changes, it is very likely that images might have been updated(version update)/added/removed. So changes to docker compose file invalidates the cache.

soulbird: What I want to say is that you should only calculate the hash value of `docker-compose-${type}.yml `, what is the result of `${{ hashFiles('./ci/pod/*.yml') }}` ?

shreemaan-abhishek: Currently, all docker images are being saved after extracting the image names from all docker compose files, with a key containing the combined hash of all docker compose files.

image


So if we want to use hash value of  only `docker-compose-${type}.yml`, then I think we should only save images present in `docker-compose-${type}.yml` only, not from all the docker compse files. WDYT?

soulbird: You may not understand github action matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs
It has a "fission effect".
'first', 'plugin', and 'last' will be several independent CI tasks that do not affect each other

shreemaan-abhishek: > It has a "fission effect".
Oh yes! What I meant to say is using hash value of only `docker-compose-${type}.yml` will lead to multiple cache files containing the same docker images.

So if we change `${{ hashFiles('./ci/pod/*.yml') }}` to `${{ hashFiles('.ci/pod/docker-compose-${type}.yml') }}`. Then we should also update the part that is saving the docker images, i.e save images mentioned in the `.ci/pod/docker-compose-${type}.yml` file only (instead of saving images mentioned in all the docker compose files).

shreemaan-abhishek: My question is which path should we prefer? Saving all docker images mentioned in all docker compose files in a single cache file.

Or saving docker images mentioned in each docker compose file (first or last or plugin... ..) in a different cache file.

I think the second one sounds more efficient, but I am waiting for your opinion.

.github/workflows/build.yml
```diff
@@ -99,29 +99,92 @@ jobs:
...
+              sudo wget https://github.com/jacekkow/keycloak-protocol-cas/releases/download/18.0.2/keycloak-protocol-cas-18.0.2.jar -O /opt/keycloak-protocol-cas-18.0.2.jar
+            fi
+            make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
+            if [[ ${{ steps.test_env.outputs.type }} != 'first' ]]; then
+              sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
+            fi
+            echo "lanuch saved images done, time: $(date)"
+          fi
+      - if: ${{ steps.cache-images.outputs.cache-hit != 'true' }}
+        name: Linux lanuch services
```
shreemaan-abhishek, line 133: rename lanuch to launch

.github/workflows/build.yml
soulbird: The second method should be used, this method has as few cached docker images as possible. If we update the images in docker-compose.first, should the cache still be valid when executing plugin tasks? So the hash value only needs to be calculated for a certain docker compose file, not all

.github/workflows/build.yml
soulbird: Try to adjust it like this, the current shell judgment logic is not clear enough. And there are some repeated commands, such as: `sudo keytool -genkeypair -keyalg RSA -dname "CN=127.0.0.1" -alias 127.0.0.1 -keystore ./ci/pod/kafka/kafka-server/selfsigned.jks -validity 365 -keysize 2048 -storepass changeit`

run: |
    sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh before
    make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
    sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after

    bash ./ci/backup-docker-images.sh ${{ steps.test_env.outputs.type }}

ci/init-plugin-test-service.sh
```diff
@@ -16,30 +16,44 @@
...
+  # wait for keycloak ready
+  bash -c 'while true; do curl -s localhost:8080 &>/dev/null; ret=$?; [[ $ret -eq 0 ]] && break; sleep 3; done'
+  docker cp ci/kcadm_configure_cas.sh apisix_keycloak_new:/tmp/
+  docker exec apisix_keycloak_new bash /tmp/kcadm_configure_cas.sh
+}
+
+before() {
+  # download keycloak cas provider
+  sudo wget https://github.com/jacekkow/keycloak-protocol-cas/releases/download/18.0.2/keycloak-protocol-cas-18.0.2.jar -O /opt/keycloak-protocol-cas-18.0.2.jar
+  ./ci/pod/openfunction/build-function-image.sh
```
soulbird, line 52: This script builds the docker image, if the cache hits then it should not be called

.github/workflows/centos7-ci.yml
```diff
@@ -116,15 +116,15 @@ jobs:
...
         ./ci/pod/openfunction/build-function-image.sh
         make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
-        ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
+        ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after
 
     - name: Start CI env (LAST_TEST)
       if: steps.test_env.outputs.type == 'last'
       run: |
         # generating SSL certificates for Kafka
         keytool -genkeypair -keyalg RSA -dname "CN=127.0.0.1" -alias 127.0.0.1 -keystore ./ci/pod/kafka/kafka-server/selfsigned.jks -validity 365 -keysize 2048 -storepass changeit
```
soulbird, line 125: already in ci/init-last-test-service.sh before ？

.github/workflows/fips.yml
```diff
@@ -103,15 +103,15 @@ jobs:
...
           ./ci/pod/openfunction/build-function-image.sh
           make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
-          sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
+          sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after
 
       - name: Start CI env (LAST_TEST)
         if: steps.test_env.outputs.type == 'last'
         run: |
           # generating SSL certificates for Kafka
           sudo keytool -genkeypair -keyalg RSA -dname "CN=127.0.0.1" -alias 127.0.0.1 -keystore ./ci/pod/kafka/kafka-server/selfsigned.jks -validity 365 -keysize 2048 -storepass changeit
```
soulbird, line 112: ditto

ci/init-last-test-service.sh
```diff
@@ -16,15 +16,27 @@
...
+  # create messages for test-consumer
+  for i in `seq 30`
+  do
+      docker exec -i apache-apisix_kafka-server1_1 bash -c "echo "testmsg$i" | /opt/bitnami/kafka/bin/kafka-console-producer.sh --bootstrap-server 127.0.0.1:9092 --topic test-consumer"
+      echo "Produces messages to the test-consumer topic, msg: testmsg$i"
+  done
+  echo "Kafka service initialization completed"
+}
+
+if [[ $1 == 'after' ]]; then
```
soulbird, line 38: Let's use `case xxx in`

shreemaan-abhishek: Brilliant idea!

.github/workflows/build.yml
```diff
@@ -99,29 +99,38 @@ jobs:
...
-          sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh
-
-      - name: Start CI env (LAST_TEST)
-        if: steps.test_env.outputs.type == 'last'
+          if [[ -f docker-images-backup/apisix-images.tar ]]; then
+            [[ ${{ steps.test_env.outputs.type }} != first ]] && sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh before
+            docker load --input docker-images-backup/apisix-images.tar
+            make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
+            echo "loaded docker images"
+            echo test_type:
```
soulbird, line 119: remove this log

.github/workflows/build.yml
```diff
@@ -99,29 +99,38 @@ jobs:
...
+            [[ ${{ steps.test_env.outputs.type }} != first ]] && sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after && echo "executed"
+          fi
+          echo "exited if"
```
soulbird, line 122: ditto

.github/workflows/build.yml
```diff
@@ -158,3 +164,12 @@ jobs:
         env:
           TEST_FILE_SUB_DIR: ${{ matrix.test_dir }}
         run: sudo -E ./ci/${{ matrix.os_name }}_runner.sh script
+
+      - if: ${{ steps.cache-images.outputs.cache-hit != 'true' }}
+        name: Save docker images
+        run: |
+          # free some space to avoid disk full
+          sudo rm -rf /usr/local/lib/android /usr/share/dotnet /usr/share/swift
```
soulbird, line 172: Before that, all CIs were working fine. So I think your modification caused the lack of disk space, but your modification is obviously irrelevant to the content you submitted this time, so I cannot accept it.

soulbird (approved):

ci/backup-docker-images.sh
```diff
@@ -0,0 +1,46 @@
...
+openwhisk_tag="openwhisk/action-nodejs-v14:nightly openwhisk/standalone:nightly"
+echo
+echo special_tag: $special_tag
+echo openwhisk_tag: $openwhisk_tag
+echo
+all_tags="${special_tag} ${openwhisk_tag}"
+to_pull=""
+for tag in $all_tags
+do
+if ! ( docker inspect $tag &> /dev/null )
```
spacewander, line 35: Bad indentation

ci/free_disk_space.sh
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+
+echo "=============================================================================="
+echo "Freeing up disk space on CI system"
+echo "=============================================================================="
+
+df -h
+echo "Removing unnecessary large directories"
+sudo rm -rf /usr/local/lib/android /usr/share/dotnet /usr/share/swift
```
spacewander, line 25: Why do we need to remove these in CI for a server?

shreemaan-abhishek: We were getting `disk storage full` errors when saving docker images, so we are deleting large directories that we don't need to free up space. This is an implementation inspired by Flink. https://github.com/apache/flink/blob/02d30ace69dc18555a5085eccf70ee884e73a16e/tools/azure-pipelines/free_disk_space.sh

spacewander: Does flink's Azure pipelines use the same image as GitHub Action? They seem to be two different CI env.
Maybe we need to use `du` to check the actual disk usage.
And let's add a comment for it.

ci/backup-docker-images.sh
```diff
@@ -0,0 +1,46 @@
...
+openwhisk_tag="openwhisk/action-nodejs-v14:nightly openwhisk/standalone:nightly"
+echo
+echo special_tag: $special_tag
+echo openwhisk_tag: $openwhisk_tag
+echo
+all_tags="${special_tag} ${openwhisk_tag}"
+to_pull=""
+for tag in $all_tags
+do
+  if ! ( docker inspect $tag &> /dev/null )
```
spacewander, line 35: Please use 4-space indentation like other shell files: https://github.com/apache/apisix/blob/master/ci/tars-ci.sh

ci/init-last-test-service.sh
```diff
@@ -16,15 +16,30 @@
...
-do
-    docker exec -i apache-apisix_kafka-server1_1 bash -c "echo "testmsg$i" | /opt/bitnami/kafka/bin/kafka-console-producer.sh --bootstrap-server 127.0.0.1:9092 --topic test-consumer"
-    echo "Produces messages to the test-consumer topic, msg: testmsg$i"
-done
-echo "Kafka service initialization completed"
+after() {
+  docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 1 --topic test2
+  docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 3 --topic test3
+  docker exec -i apache-apisix_kafka-server2_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server2:2181 --replication-factor 1 --partitions 1 --topic test4
+  docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 1 --topic test-consumer
```
spacewander, line 28: Ditto

ci/init-plugin-test-service.sh
```diff
@@ -16,30 +16,46 @@
 # limitations under the License.
 #
 
-docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 1 --topic test2
-docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 3 --topic test3
-docker exec -i apache-apisix_kafka-server2_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server2:2181 --replication-factor 1 --partitions 1 --topic test4
+after() {
+  docker exec -i apache-apisix_kafka-server1_1 /opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper zookeeper-server1:2181 --replication-factor 1 --partitions 1 --topic test2
```
spacewander, line 20: Ditto

ci/free_disk_space.sh
shreemaan-abhishek: Not sure if they use the same Github Action but I manually _ssh-ed_ into the action runner to find out large directories.

shreemaan-abhishek: > And let's add a comment for it.

I'm sorry I did not understand. Comment for what? 😅

shreemaan-abhishek: > Maybe we need to use du to check the actual disk usage.

I think `df -h` gives us a better result.

spacewander: > > And let's add a comment for it.
> 
> I'm sorry I did not understand. Comment for what? 😅

Comment on why we decided to delete those directories. For example, we deleted the top three unnecessary large directories and they are found by command xxx...

shreemaan-abhishek: Understood! Just pushed my changes.

spacewander: Where can I find the new comment?

shreemaan-abhishek: [Here](https://github.com/apache/apisix/pull/8735/files#diff-54747011ef6df77ae48a9abd7cf3c62ee69410f22fa6a41cbf04e4852a6f6a1cR19-R22), in this [commit](https://github.com/apache/apisix/pull/8735/commits/e4d38d321ef46552e3121b65186c012094b36d87).

spacewander: Got it. Thanks!

spacewander (approved):

soulbird (approved):

PR #8786, opened 2023-02-03
feat(batch-processor): enhance batch processor: support entries_max_size

xiaoxuanzi:
### Description

Enhance batch processor: support entries_max_size

In production, the log has reached batch_max_size, and the memory may be very large.
This feature solves this problem.
When the size of the log is reached entries_max_size, it is pushed to the HTTP/HTTPS service.

Fixes # (issue)

### Checklist


apisix/utils/batch-processor.lua
```diff
@@ -200,10 +202,19 @@ function batch_processor:push(entry)
     end
     self.last_entry_t = now()
 
+    local entry_size = #entry
```
spacewander, line 205: The entry here is usually an object. Is there any real-world use case for this feature?

xiaoxuanzi: Yes, sometimes we want to include request body or response body in the log.
Request body or response body may be large size.

spacewander (commented):
> Request body or response body may be large size.

I wonder whether it is good to solve this problem by adding a new size counter.

First of all, we can always reduce the batch size to solve this problem, without adding any code/complexity.

Secondly, if some requests have a big body, it would be better not to record them. Record body in MBs is a strange demand. We can do it via:
* condition checks Content-Length
* limit the body size in the config (planning but not yet supported)

Thirdly, how to serialize the entry is decided by the sender. Serializing it to JSON is not accurate, it may be in pb format or something else. Anyway, it is strange to let the processor do the sender's job.

xiaoxuanzi:
Hi @spacewander.
I got it.
I really appreciate your precious advice.
Thanks.

Issue #8804, opened 2023-02-07
chore: Supplement the schema in deployment 


tao12345666333:
### Description

Apache APISIX v3 adds a decoupled deployment mode. ref: https://apisix.apache.org/docs/apisix/deployment-modes/#decoupled

The current schema is not perfect. 

When the role is `control_plane`, it should contain the configuration of the `admin` field, which is consistent with traditional.

https://github.com/apache/apisix/blob/5984321c34d3cdc90573ebbecfd17ba9082fb3ba/apisix/cli/schema.lua#L375-L412

xref: https://github.com/apache/apisix-helm-chart/issues/459

jiangfucheng:
Hi, could assign it me, i want to try it.

rukundob451:
@tao12345666333 was the issue worked on?

tokers:
@jiangfucheng Did you start it? I can assign it to you.

jiangfucheng:
> @jiangfucheng Did you start it? I can assign it to you.

I was submit a pr https://github.com/apache/apisix/pull/8809, but still has some problem, could you guide me slove it? I can't reproduce it in my local machine.

PR #8809, opened 2023-02-07
fix: add admin schema to control_plane config

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/8804

Maybe just add the schema of admin to conrol_plane config definition is enough?

### Checklist

### DIFF
--- a/apisix/cli/schema.lua
+++ b/apisix/cli/schema.lua
@@ -375,6 +375,7 @@ local deployment_schema = {
     control_plane = {
         properties = {
             etcd = etcd_schema,
+            admin = admin_schema,
             role_control_plane = {
                 properties = {
                     config_provider = {
--- a/t/cli/test_deployment_control_plane.sh
+++ b/t/cli/test_deployment_control_plane.sh
@@ -53,7 +53,7 @@ deployment:
             cert: t/certs/mtls_server.crt
             cert_key: t/certs/mtls_server.key
     admin:
-      https_admin: "abc"
+        https_admin: "abc"
     etcd:
         prefix: "/apisix"
         host:
--- a/t/cli/test_deployment_control_plane.sh
+++ b/t/cli/test_deployment_control_plane.sh
@@ -52,6 +52,36 @@ deployment:
             listen: admin.apisix.dev:12345
             cert: t/certs/mtls_server.crt
             cert_key: t/certs/mtls_server.key
+    admin:
+      https_admin: "abc"
+    etcd:
+        prefix: "/apisix"
+        host:
+            - http://127.0.0.1:2379
+    certs:
+        trusted_ca_cert: t/certs/mtls_ca.crt
+' > conf/config.yaml
+
+out=$(make init 2>&1 || true)
+if ! echo "$out" | grep 'property "https_admin" validation failed: wrong type: expected boolean, got string'; then
+    echo "failed: should check deployment schema during init"
+    exit 1
+fi
+
+echo "passed: should check deployment schema during init"
+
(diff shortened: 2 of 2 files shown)


spacewander (commented):
Can you add a test to validate the admin schema in https://github.com/apache/apisix/blob/master/t/cli/test_deployment_control_plane.sh?
So we won't miss schema next time.

jiangfucheng:
> Can you add a test to validate the admin schema in https://github.com/apache/apisix/blob/master/t/cli/test_deployment_control_plane.sh? So we won't miss schema next time.

I added test case for this pr. I wanted to through `schema.validate()` method to validate whether config is right, but admin field already in default-config.yaml, `schema.validate()` can't return any error, so i deleted admin field in config-default.yaml.

spacewander (commented):
You can provide an incorrect type in
https://github.com/apache/apisix/blob/e021266129eff537ea6b59cca0061385663d3a69/apisix/cli/schema.lua#L338

The basic rule: try your best to avoid touching the functional code when you write test

jiangfucheng:
@spacewander I can't reproduce the failed CI, could you guide me how to fix it?

shreemaan-abhishek:
@jiangfucheng You can make an empty commit and push it to re-run the CI. This might be some error that does not occur consistently.

t/cli/test_deployment_control_plane.sh
```diff
@@ -40,6 +40,37 @@ fi
...
+    etcd:
+        prefix: "/apisix"
+        host:
+            - http://127.0.0.1:2379
+    certs:
+        trusted_ca_cert: t/certs/mtls_ca.crt
+' > conf/config.yaml
+
+out=$(make init 2>&1 || true)
+if ! echo "$out" | grep 'expect: number, but got: string'; then
```
spacewander, line 67: The err is from there: https://github.com/apache/apisix/blob/0bc65ea9acd726f79f80ae0abd8f50b7eb172e3d/apisix/cli/file.lua#L180
It is detected before the schema check.
We should find another way to validate the schema check.

apisix/cli/schema.lua
```diff
@@ -409,7 +410,7 @@ local deployment_schema = {
                 default = {},
             },
         },
-        required = {"etcd", "role_control_plane"}
+        required = {"etcd", "admin", "role_control_plane"}
```
spacewander, line 413: The admin is not required as we always have a default value.

t/cli/test_deployment_control_plane.sh
jiangfucheng: I changed a special field, it can be merged, but can't pass config validation, although this way is not perfect。🤣

t/cli/test_deployment_control_plane.sh
```diff
@@ -40,6 +40,36 @@ fi
...
+deployment:
+    role: control_plane
+    role_control_plane:
+        config_provider: etcd
+        conf_server:
+            listen: admin.apisix.dev:12345
+            cert: t/certs/mtls_server.crt
+            cert_key: t/certs/mtls_server.key
+    admin:
+      https_admin: "abc"
```
spacewander, line 56:
```suggestion
        https_admin: "abc"
```

spacewander (dismissed):

t/cli/test_deployment_control_plane.sh
jiangfucheng: Fixed

spacewander (approved):

soulbird (approved):

PR #8819, opened 2023-02-09
chore(test): split the access log of apisix and test server

soulbird:
### Description

1. Record the logs of apisix and test server separately for better observation
2. Use the defined log format

### Checklist

### DIFF
--- a/t/plugin/ext-plugin/http-req-call.t
+++ b/t/plugin/ext-plugin/http-req-call.t
@@ -515,7 +515,7 @@ GET /hello
         }
     }
 --- error_log
-undefined path in test server, uri: /plugin_proxy_rewrite_args?a=2
+undefined path in test server, uri: /plugin_proxy_rewrite_args%3Fa=2
 --- error_code: 404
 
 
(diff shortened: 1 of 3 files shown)


spacewander:

t/plugin/ext-plugin/http-req-call.t
```diff
@@ -514,8 +514,8 @@ GET /hello
             ext.go({rewrite_bad_path = true})
         }
     }
---- access_log
-GET /plugin_proxy_rewrite_args%3Fa=2
+--- error_log
+undefined path in test server, uri: /plugin_proxy_rewrite_args?a=2
```
spacewander, line 518: IMHO, we can use `request_uri` so we know it is a "bad" path.

soulbird: The access log of the test server has been placed in another file. --- access_log It seems that the access_log file path cannot be specified.

spacewander: I mean using it in the error log. Like:
```
ngx.log(ngx.WARN, "undefined path in test server, uri: ", ngx.var.request_uri)
```

spacewander (approved):

monkeyDluffy6017 (approved):

PR #8762, opened 2023-02-01
chore(kafka-logger): support configuration `meta_refresh_interval` parameter

nic-chen:
### Description

Fixes #6033

please refer to:
https://github.com/doujiang24/lua-resty-kafka/issues/70

[ci on centos7](https://github.com/apache/apisix/actions/runs/4061647503/jobs/6991872804#logs) is run in container, so we couldn't restart kafka broker in the test.

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #8821, opened 2023-02-09
chore(error-log-logger): add kafka meta_refresh_interval

ronething:
### Description

Fixes # (issue)

Inspired by PR: https://github.com/apache/apisix/pull/8762

### Checklist


docs/en/latest/plugins/error-log-logger.md
```diff
@@ -60,6 +60,7 @@ It might take some time to receive the log data. It will be automatically sent a
...
 | kafka.key                              | string  | False    |                |                       | Key used for allocating partitions for messages.                                                                                                                                                                                                                                                                                                 |
 | kafka.cluster_name           | integer | False    | 1              | [0,...]               | Name of the cluster. Used when there are two or more Kafka clusters. Only works if the `producer_type` attribute is set to `async`.                                                                                                                                                                                                              |
+| kafka.meta_refresh_interval | integer | optional    | 30              | [1,...]               | `refresh_interval` parameter in [lua-resty-kafka](https://github.com/doujiang24/lua-resty-kafka) specifies the time to auto refresh the metadata, in seconds.|
```
spacewander, line 63:
```suggestion
| kafka.meta_refresh_interval | integer | False    | 30              | [1,...]               | `refresh_interval` parameter in [lua-resty-kafka](https://github.com/doujiang24/lua-resty-kafka) specifies the time to auto refresh the metadata, in seconds.|
```

t/plugin/error-log-logger-kafka.t
```diff
@@ -87,7 +87,8 @@ done
                             "host": "127.0.0.1",
                             "port": 9092
                         }],
-                        "kafka_topic": "test2"
+                        "kafka_topic": "test2",
+                        "meta_refresh_interval": 1
```
spacewander, line 91: Can we add a mock test to ensure this config is passed to broker_config?

ronething: @spacewander Can we check by the following steps:

1. log the `broker_config` info in `send_to_kafka` function
2. check if the error log include `"refresh_interval":1000` or not?

docs/en/latest/plugins/error-log-logger.md
ronething: it seems in this disscussion https://github.com/apache/apisix/pull/8762#discussion_r1093966576, the value is `optional`?

t/plugin/error-log-logger-kafka.t
spacewander: We can verify it without adding any code to the plugin.

https://github.com/apache/apisix/blob/0bc65ea9acd726f79f80ae0abd8f50b7eb172e3d/t/plugin/sls-logger.t#L373
Like this, we can inject some code while creating the producer.

docs/en/latest/plugins/error-log-logger.md
tokers: Let's keep with the same style in this docs.

If we need to change the value from `True/False` to `optional`, let's update all of them.

t/plugin/error-log-logger-kafka.t
ronething: ok, i see. But func `create_producer` is a local func, how can we inject it by `extra_init_by_lua`?

docs/en/latest/plugins/error-log-logger.md
ronething: > Let's keep with the same style in this docs.

i will change it later.

t/plugin/error-log-logger-kafka.t
spacewander: We can inject https://github.com/apache/apisix/blob/0bc65ea9acd726f79f80ae0abd8f50b7eb172e3d/apisix/plugins/error-log-logger.lua#L351

ronething: @spacewander updated, please take a look, thanks.

spacewander (approved):

soulbird (approved):

Issue #8811, opened 2023-02-07
bug: Lots of "request.lua:120: set_header(): DEPRECATED" warnings in error.log


macdoor:
### Current Behavior

Every request produces "request.lua:120: set_header(): DEPRECATED" warning in error.log. It makes error.log useless

### Expected Behavior

The deprecated set_header call needs to be changed

### Error Logs

2023/02/08 06:20:11 [warn] 9883#9883: *152597 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.11.69.11, server: _, request: "POST /g/gw/interface/performance/v2.0/data HTTP/1.1", host: "36.133.195.59:18081"
2023/02/08 06:20:11 [warn] 9883#9883: *153292 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.11.69.8, server: _, request: "POST /g/gw/interface/performance/v2.0/data HTTP/1.1", host: "36.133.195.59:18081"
2023/02/08 06:20:11 [warn] 9883#9883: *153292 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.11.69.8, server: _, request: "POST /g/gw/interface/performance/v2.0/data HTTP/1.1", host: "36.133.195.59:18081"
2023/02/08 06:20:11 [warn] 9884#9884: *153311 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.11.69.11, server: _, request: "POST /g/gw/interface/performance/v2.0/data HTTP/1.1", host: "36.133.195.59:18081"
2023/02/08 06:20:11 [warn] 9884#9884: *153311 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.11.69.11, server: _, request: "POST /g/gw/interface/performance/v2.0/data HTTP/1.1", host: "36.133.195.59:18081"

### Steps to Reproduce

1. yum install apisix
2. create conf/config.yaml
```
deployment:
  role: traditional
  role_traditional:
    config_provider: etcd
  etcd:
    host:
      - http://localhost:2379

  admin:
    admin_key:
      - name: admin
        key: edd1c9f034335f136f87ad84b625c8f1  # using fixed API token has security risk, please update it when you deploy to production environment
        role: admin

apisix:
[...]

PR #8824, opened 2023-02-09
fix: invalidate cache in core.request.add_haeder and fix some calls

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/8811

### Checklist


apisix/plugins/proxy-rewrite.lua
```diff
@@ -311,18 +311,18 @@ do
         for i = 1, field_cnt, 2 do
             local val = core.utils.resolve_var(hdr_op.add[i + 1], ctx.var)
             local header = hdr_op.add[i]
-            core.request.add_header(header, val)
+            core.request.add_header(ctx, header, val)
```
tokers, line 314: We may add a case to ensure the warning message won't be generated anymore.

jiangfucheng: Done

jiangfucheng:
image
This test can pass on my local machine😳.

t/core/request.t
```diff
@@ -439,3 +439,59 @@ while ($i <= 101) {
...
+            core.request.add_header(ctx, "test_header", "test")
+            local h = core.request.header(ctx, "test_header")
+            ngx.say(ctx.headers.test_header)
+            core.request.add_header(ctx, "test_header", "abc")
+            ngx.say(ctx.headers.test_header)
+        }
+    }
+--- response_body
+test
+nil
```
soulbird, line 478: Should be `abc` ？

jiangfucheng: `add_header()` method will invalidate cache in `ctx`, `header()` method will add cache to `ctx`, so first value is "test" and the second value is "nil"

jiangfucheng:
@soulbird @spacewander @tokers PTAL, Thanks.

t/core/request.t
soulbird: Why does `add_header()` invalidate the cache?

t/plugin/proxy-rewrite3.t
```diff
@@ -455,3 +455,12 @@ passed
 GET /echo HTTP/1.1
 --- response_headers
 test: test_in_set
+
+
+
+=== TEST 21: deprecated calls test
```
tokers, line 461: We can merge this case to `TEST 20`?

t/core/request.t
jiangfucheng: This is the feature of current PR. `header(ctx, header_name)` method maybe get wrong values from ctx cache if the `add_header` method does not invalidate the cache.

t/plugin/proxy-rewrite3.t
jiangfucheng: I think each test should foucus on one function point, even if their trigger conditions are same.

apisix/core/request.lua
```diff
@@ -143,14 +143,28 @@ function _M.set_header(ctx, header_name, header_value)
...
         error(err)
     end
 
     req_add_header(header_name, header_value)
+
+    if not ctx then
+        ctx = ngx.ctx.api_ctx
+    end
+
+    ctx.headers = {}
```
monkeyDluffy6017, line 167: The invalidate cache operatio should follow the `set_header` function

soulbird (approved):

apisix/core/request.lua
```diff
@@ -143,14 +143,35 @@ function _M.set_header(ctx, header_name, header_value)
     end
 end
 
-function _M.add_header(header_name, header_value)
+function _M.add_header(ctx, header_name, header_value)
```
monkeyDluffy6017, line 146: The function `add_header` and `set_header` are almost identical, maybe we could extract a common function

spacewander: The names follow the corresponding API in OpenResty.

apisix/core/request.lua
```diff
@@ -143,14 +143,35 @@ function _M.set_header(ctx, header_name, header_value)
...
+        changed = a6_request.is_request_header_set()
+    end
+
+
     req_add_header(header_name, header_value)
+
+    if is_apisix_or and not changed then
+        a6_request.clear_request_header()
+        if ctx and ctx.headers then
+            ctx.headers[header_name] = header_value
```
spacewander, line 172: We should append instead of overwriting it?

jiangfucheng: Fixed

apisix/core/request.lua
```diff
@@ -131,26 +128,36 @@ function _M.set_header(ctx, header_name, header_value)
...
     if is_apisix_or and not changed then
         -- if the headers are not changed before,
         -- we can only update part of the cache instead of invalidating the whole
         a6_request.clear_request_header()
         if ctx and ctx.headers then
-            ctx.headers[header_name] = header_value
+            if override or not ctx.headers[header_name] then
+                ctx.headers[header_name] = header_value
+            else
+                ctx.headers[header_name] = ctx.headers[header_name] .. header_value
```
monkeyDluffy6017, line 141: Do you have test cases to cover this?

jiangfucheng: The Test 15 in `request.t` can cover this.

monkeyDluffy6017: I think it is not corrent.
request headers like this:
```
 Foo: foo
 Foo: bar
 Foo: baz
```
will be a lua table like this:
```
 {"foo", "bar", "baz"}
```

jiangfucheng: Yes, I misunderstood the API of `add_header`, I will fix it later.

jiangfucheng:
@spacewander @monkeyDluffy6017 @soulbird Please help me review this PR, thanks❤️.

apisix/core/request.lua
```diff
@@ -131,26 +129,41 @@ function _M.set_header(ctx, header_name, header_value)
...
 end
 
-function _M.add_header(header_name, header_value)
-    local err
-    header_name, err = _validate_header_name(header_name)
-    if err then
-        error(err)
+function _M.set_header(ctx, header_name, header_value)
+    if type(ctx) == "string" then
+        log.warn("DEPRECATED: use set_header(ctx, header_name, header_value) instead")
```
spacewander, line 155: We can move the DEPRECATED log into the common function, and show a different log according to the override parameter.

apisix/core/request.lua
```diff
@@ -131,26 +129,41 @@ function _M.set_header(ctx, header_name, header_value)
         changed = a6_request.is_request_header_set()
     end
 
-    ngx.req.set_header(header_name, header_value)
+    update_func(header_name, header_value)
```
spacewander, line 132: We can use override instead of passing an extra function.

t/core/request.t
```diff
@@ -439,3 +439,47 @@ while ($i <= 101) {
...
+        content_by_lua_block {
+            local core = require("apisix.core")
+            ngx.ctx.api_ctx = {}
+            local ctx = ngx.ctx.api_ctx
+            core.request.add_header(ctx, "test_header", "test")
+            local h = core.request.header(ctx, "test_header")
+            ngx.say(h)
+            core.request.add_header(ctx, "test_header", "t2")
+            local h2 = core.request.header(ctx, "test_header")
+            ngx.say("[" .. table.concat(h2, ", ") .. "]")
```
monkeyDluffy6017, line 457: is it better to use `core.json.encode` instead?

jiangfucheng: done

apisix/core/request.lua
jiangfucheng: Fixed.

apisix/core/request.lua
jiangfucheng: Fixed.

t/core/request.t
```diff
@@ -439,3 +439,47 @@ while ($i <= 101) {
...
 --- response_body
 101
+
+
+
+=== TEST 15: add header
+--- config
+    location /t {
+        content_by_lua_block {
+            local core = require("apisix.core")
```
soulbird, line 449: It is better to use `local json = require("toolkit.json")`, because it will sort the keys of the table

jiangfucheng: Thanks, fixed.

t/plugin/proxy-rewrite3.t
```diff
@@ -619,3 +622,50 @@ GET /test/plugin/proxy/rewrite HTTP/1.1
     }
 --- response_body
 /plugin_proxy_rewrite?a=c
+
+
+
+=== TEST 27: set route
```
soulbird, line 628: Is it a duplicate of test 19?

jiangfucheng: Just want prepare configuration for Test 28.

soulbird: Aren't Test 28 and Test 20 duplicates?

jiangfucheng: Ok, I will merge them. At first, I just want one test focus on one function pont.

jiangfucheng: Fixed.

soulbird (dismissed):

apisix/core/request.lua
```diff
@@ -131,26 +135,37 @@ function _M.set_header(ctx, header_name, header_value)
         changed = a6_request.is_request_header_set()
     end
 
-    ngx.req.set_header(header_name, header_value)
+    if override then
+        ngx.req.set_header(header_name, header_value)
```
spacewander, line 139: For symmetry beauty, let's also localize this function too.

jiangfucheng: Fixd.

spacewander (approved):

monkeyDluffy6017 (approved):

soulbird (approved):

PR #8855, opened 2023-02-14
fix(admin): fix wrong http code for patch method

An-DJ:
### Description

The default http code of patch method to admin.resources is `201`. It should be corrected to `200`.

### Checklist

### DIFF
--- a/apisix/core/etcd.lua
+++ b/apisix/core/etcd.lua
@@ -507,7 +507,6 @@ function _M.atomic_set(key, value, ttl, mod_revision)
         key = key,
         value = value,
     }
-    res.status = 200
 
     return res, nil
 end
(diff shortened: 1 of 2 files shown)


soulbird (approved):

tokers (dismissed):

apisix/core/etcd.lua
```diff
@@ -507,7 +507,7 @@ function _M.atomic_set(key, value, ttl, mod_revision)
         key = key,
         value = value,
     }
-    res.status = 201
+    res.status = 200
```
spacewander, line 510: Is the default status already 200?

An-DJ: IMHO yes. Only the PATCH method uses this function now.

spacewander: So we don't need to reassign it at all

An-DJ: The res.status is assigned here and is only called by `PATCH`. That seems reasonable, doesn't it?

spacewander (approved):

soulbird (approved):

PR #8866, opened 2023-02-15
feat: use env var instead of plain text for vault token

shreemaan-abhishek:
### Description

Currently, we use plain text to add `vault` as a secret resource. Like so:

```bash
curl http://127.0.0.1:9180/apisix/admin/secret/vault/1 \
{
    "uri": "https://127.0.0.1:8200"，
    "prefix": "apisix",
    "token": "root"
}'
```

This PR implements a feature that would allow the use of environment variables in the following format: `$ENV://NAME_OF_ENV_VAR`.

This is the final outcome:

```bash
curl http://127.0.0.1:9180/apisix/admin/secret/vault/1 \
{
    "uri": "https://127.0.0.1:8200"，
    "prefix": "apisix",
    "token": "$ENV://VAULT_TOKEN"
}'
```

A part of #8319 

cc: @soulbird 

### Checklist


t/config-center-yaml/secret.t
```diff
@@ -61,7 +61,7 @@ __DATA__
 secrets:
   - id: vault/1
     prefix: kv/apisix
-    token: root
+    token: "$ENV://VAULT_TOKEN"
```
soulbird, line 64: So we can no longer use such a configuration?
```
token: root
```

shreemaan-abhishek: Yes, we won't be able to use such a configuration.

ci/centos7-ci.sh
```diff
@@ -84,6 +84,7 @@ run_case() {
     export_or_prefix
     make init
     set_coredns
+    export VAULT_TOKEN="root"
```
soulbird, line 87: refer: https://github.com/apache/apisix/blob/master/t/core/env.t#L17-L20

t/config-center-yaml/secret.t
```diff
@@ -388,3 +392,131 @@ env secret=apisix;
...
+            local values = secret.secrets()
+            ngx.say(#values)
+        }
+    }
+--- request
+GET /t
+--- response_body
+0
+--- error_log
+property "uri" validation failed: failed to match pattern "^[^\\/]+:\\/\\/([\\da-zA-Z.-]+|\\[[\\da-fA-F:]+\\])(:\\d+)?"
```
soulbird, line 419: Why add this test case? It doesn't look like it has anything to do with your modification？

t/config-center-yaml/secret.t
```diff
@@ -388,3 +392,131 @@ env secret=apisix;
...
+            local values = secret.secrets()
+            ngx.say(#values)
+        }
+    }
+--- request
+GET /t
+--- response_body
+0
+--- error_log
+secret manager not exits
```
soulbird, line 444: ditto

t/config-center-yaml/secret.t
```diff
@@ -388,3 +392,131 @@ env secret=apisix;
...
+    }
+--- request
+GET /t
+--- response_body
+len: 1
+id: vault/1
+prefix: kv/apisix
+token: $ENV://VAULT_TOKEN
+uri: http://127.0.0.1:8200
+
```
soulbird, line 477: There is no need to add this test

t/config-center-yaml/secret.t
```diff
@@ -388,3 +392,131 @@ env secret=apisix;
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+value
+
+
+
+=== TEST 21: secret.fetch_by_uri, no sub key value with the token in an env var
```
soulbird, line 503: This test case already exists

t/plugin/basic-auth.t
```diff
@@ -540,3 +544,73 @@ GET /echo
...
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
```
soulbird, line 616: You only configured the route, you also need to verify whether the route is effective

t/config-center-yaml/secret.t
shreemaan-abhishek: I just duplicated all the test cases containing the token with just one change: replace the token with the environment variable.

t/config-center-yaml/secret.t
```diff
@@ -14,6 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+BEGIN {
+    $ENV{VAULT_TOKEN} = "root";
```
soulbird, line 18: Why was this added?

soulbird:
IMO, just adding a test case for t/secret/vault.t is enough. The test cases under t/plugin seem redundant, don't you think?

shreemaan-abhishek:
> The test cases under t/plugin seem redundant, don't you think?

I thought the same. But then I thought they would provide better safety from potential nasty bugs in the future.

soulbird:
> > The test cases under t/plugin seem redundant, don't you think?
> 
> I thought the same. But then I thought they would provide better safety from potential nasty bugs in the future.

ok

soulbird (approved):

apisix/secret/vault.lua
```diff
@@ -26,6 +26,7 @@ local norm_path = require("pl.path").normpath
 local sub        = core.string.sub
 local rfind_char = core.string.rfind_char
 
+local env        = require("apisix.core.env")
```
spacewander, line 29: Can't we import it like `core.string.sub`?

spacewander (commented):
Let's update https://github.com/apache/apisix/blob/master/docs/en/latest/terminology/secret.md

apisix/secret/vault.lua
```diff
@@ -53,10 +54,15 @@ local function make_request_to_vault(conf, method, key, data)
     local req_addr = conf.uri .. norm_path("/v1/"
                 .. conf.prefix .. "/" .. key)
 
+    local token, _ = env.fetch_by_uri(conf.token)
```
soulbird, line 57: Just use `core.env.fetch_by_uri`

shreemaan-abhishek: I think it's okay to import using `require` and use because if in the future we need to use the env module again in this file then we won't have to re-import using `require`.

Also, at that time in the future, we will not have to replace `core.env.fetch_by_uri` with `env.fetch_by_uri`. WDYT?

soulbird (dismissed):

apisix/secret/vault.lua
```diff
@@ -26,6 +26,7 @@ local norm_path = require("pl.path").normpath
 local sub        = core.string.sub
 local rfind_char = core.string.rfind_char
```
spacewander, line 28: We can avoid the blank line here? Other files don't add a blank line among localized variables.

spacewander (approved):

soulbird (approved):

PR #8884, opened 2023-02-20
feat: handle_extra_info add set req body

hongdingyi:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -288,7 +288,14 @@ local function handle_extra_info(ctx, input)
 
     local res
     local info_type = req:InfoType()
-    if info_type == extra_info.Var then
+    if info_type == extra_info.SetReqBody then  --4 set_body
+        local info = req:Info()
+        local var_req = extra_info_var.New()
+        var_req:Init(info.bytes, info.pos)
+        local var_name = var_req:Name()
+        core.request.set_body(var_name)
+        res = 0
+    elseif info_type == extra_info.Var then
         local info = req:Info()
         local var_req = extra_info_var.New()
         var_req:Init(info.bytes, info.pos)
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -288,7 +288,7 @@ local function handle_extra_info(ctx, input)
 
     local res
     local info_type = req:InfoType()
-    if info_type == extra_info.SetReqBody then  --4 set_body
+    if info_type == extra_info.RespBody + 1 then  --4 set_body
         local info = req:Info()
         local var_req = extra_info_var.New()
         var_req:Init(info.bytes, info.pos)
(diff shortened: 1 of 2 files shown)


hongdingyi:
fix:handle_extra_info

hongdingyi:
image

hongdingyi:
You should add a type。in apisix/apisix-3.1.0/deps/share/lua/5.1/A6/ExtraInfo

hongdingyi:
add extra info type

apisix/plugins/ext-plugin/init.lua
```diff
@@ -288,7 +288,14 @@ local function handle_extra_info(ctx, input)
 
     local res
     local info_type = req:InfoType()
-    if info_type == extra_info.Var then
+    if info_type == extra_info.RespBody + 1 then  --4 set_body
```
spacewander, line 291: Could you add the new field in https://github.com/api7/ext-plugin-proto?
You can read https://github.com/apache/apisix/blob/master/docs/en/latest/internal/plugin-runner.md for the implemented detail.

spacewander (commented):
Let's add test in https://github.com/apache/apisix/blob/master/t/plugin/ext-plugin/request-body.t

rubikplanet:
feature required

moonming:
Is this the same as https://github.com/apache/apisix/pull/9990? @monkeyDluffy6017

monkeyDluffy6017:
@moonming It's two different things

moonming:
> @moonming It's two different things

So what problems and scenarios does this PR solve? @monkeyDluffy6017

shreemaan-abhishek:
@hongdingyi what problem does this PR solve?

Revolyssup:
This PR requires to add a field in this file https://github.com/api7/ext-plugin-proto/blob/4af76c210e00ead699489cfeb055d8b9bb7596ed/ext-plugin.fbs#L121

@moonming @shreemaan-abhishek @monkeyDluffy6017 
As far as I can see, this PR and https://github.com/apache/apisix/pull/9990 are kind of similar. Both set the request's body but at different stage. 
The one that is merged sets the request body by signalling it in the response of the RPC. After that the request is sent to the upstream.
This PR on the other hand allows to set the request body using handle_extra_info function which facilitates communication between plugin runner and APISIX while the plugin runner is still processing the request. Here after the request is set, the request won't be sent to upstream immediately as the Reply isn't yet sent by the plugin runner. The plugin runner can do some more things.

Conclusion:
According to me, this PR is not required because after the request is set to something, why should plugin runner do anything else on it. It doesn't seem like something that should happen with handle_extra_info. This can be achieved by the reply of the HTTPReq RPC call. 

@hongdingyi In case I am missing some scenario where this actually will be required and just setting the request body using the Reply of the RPC wouldn't justify the use case, please let me know.

shreemaan-abhishek:
let's wait for @hongdingyi's input. It seems the github actions were not triggered properly so I just made them re-run.

monkeyDluffy6017:
@hongdingyi are you still working on the pr?

Revolyssup:
@hongdingyi Can you fix the lint check?

Revolyssup:
closed because reason not provided

PR #8904, opened 2023-02-21
feat: add consumer name to header

mscb402:
### Description

Some users in Slack hope that APISIX can return some information about consumers like Kong https://docs.konghq.com/hub/kong-inc/key-auth/#upstream-headers, so I submit this PR to add Consumer Name and Consumer Group ID to the header. So a user could know which consumer was authenticated.

### Checklist


apisix/consumer.lua
```diff
@@ -88,6 +88,11 @@ function _M.attach_consumer(ctx, consumer, conf)
     ctx.consumer_ver = conf.conf_version
 end
 
+-- attach chosen consumer to the request headers, used in auth plugin
+function _M.attach_consumer_to_request(ctx, req, consumer)
+    req.set_header(ctx, "X-Consumer-Name", consumer.consumer_name)
+    req.set_header(ctx, "X-Consumer-Group-ID", consumer.group_id)
```
spacewander, line 94: Can't the user use proxy-rewrite to add them?
https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/proxy-rewrite.md
We support using variables in the header and we have these variables: https://github.com/apache/apisix/blob/master/docs/en/latest/apisix-variable.md

mscb402: Yes. Of course, users can use proxy-rewrite to implement it, but if APISIX supports adding the name directly to the header, the difference is more convenient.

mscb402: Not all users think that they can go through proxy-rewrite to add the consumer name to the header. This method is too tricky

spacewander (changes requested):
Most of the upstream don't need to know how the gateway authenticates.  For the few users who require this, they can configure their rules.
There is no need to bloat the APISIX.

PR #8911, opened 2023-02-22
refactor(admin): secrets and ssls

### LINKED ISSUE #8569: feat: As a user, I want to refactor the resources under apisix.admin, so that we don't need to repeat

An-DJ:
### Description

**Continue to review this work after #8820 is merged.**

Refactor secrets and ssls.

All admin.resources is refactored.

Close #8569 

### Checklist


apisix/admin/resource.lua
```diff
@@ -31,49 +32,89 @@ local mt = {
...
 
-    core.log.info("schema: ", core.json.delay_encode(self.schema))
     core.log.info("conf  : ", core.json.delay_encode(conf))
 
     -- check the resource own rules
-    local ok, err = self.checker(id, conf, need_id, self.schema)
+    if self.schema then
+        typ = self.schema
+    else
+        core.log.info("schema: ", core.json.delay_encode(self.schema))
```
spacewander, line 85: Why encode a nil value and log it?

apisix/admin/resource.lua
```diff
@@ -31,49 +32,89 @@ local mt = {
...
+    end
+
+    local ok, err = self.checker(id, conf, need_id, typ)
```
spacewander, line 88: Better to put typ as the fifth argument, so the fourth argument is always the schema in every checker.

apisix/admin/resource.lua
```diff
@@ -31,49 +32,89 @@ local mt = {
...
+    end
 
-    core.log.info("schema: ", core.json.delay_encode(self.schema))
     core.log.info("conf  : ", core.json.delay_encode(conf))
 
     -- check the resource own rules
-    local ok, err = self.checker(id, conf, need_id, self.schema)
+    if self.schema then
+        typ = self.schema
+    else
```
spacewander, line 84: It would be better to check if the name == "secrets". As this is the only case when schema is nil, and we already have many `name == "secrets"` checks.

apisix/admin/resource.lua
```diff
@@ -148,23 +233,36 @@ function _M:put(id, conf, sub_path, args)
 end
 
 
-function _M:delete(id)
+function _M:delete(id, conf, sub_path)
```
spacewander, line 236: Why do we need to pass `conf` in this method?

apisix/admin/resource.lua
```diff
@@ -121,16 +178,44 @@ function _M:put(id, conf, sub_path, args)
...
-    if not id then
-        return 400, err
+    local key = "/" .. self.name
+    local typ = nil
+    if self.name == "secrets" then
+        typ, id = split_typ_and_id(id, sub_path)
     end
 
-    local key = "/" .. self.name .. "/" .. id
+    if self.name == "secrets" then
```
spacewander, line 187: We can merge this branch with the one above

apisix/admin/resource.lua
An-DJ: @spacewander We should Keep all methods in the same form.

https://github.com/apache/apisix/blob/master/apisix/admin/init.lua#L220

The meaningful first param is id, the second one is conf, and the third one is the sub_path.

Should we?

spacewander: Let's add a comment that explains why we need a placeholder `conf` here.

An-DJ: Done.

spacewander (approved):

soulbird (approved):

PR #9010, opened 2023-03-06
fix: Non wildcard origin in CORS should sent Vary header

boekkooi-lengoo:
### Description

When CORS requirements are more complicated than setting `Access-Control-Allow-Origin` to `*` then we set the `Vary` to `Origin`. This avoids caching the wrong response.

Last week I was debugging an issue where requests where failing with CORS errors. 
After some digging it became clear that when the first request to the endpoint did not match the allowed origins the request was cached and this cached resolve was then used later on for Origins that should be allowed. This turned out to be because the `Vary` header did not contain `Origin`. 

Secondly I also resolved an issue where I was unable to use only `allow_origins_by_regex` in standalone mode by allowing `allow_origins` to be an empty string. 
*I'm not sure what `metadata_schema` is used for so this maybe wrong.*

### Checklist

### DIFF
--- a/apisix/plugins/cors.lua
+++ b/apisix/plugins/cors.lua
@@ -190,10 +190,6 @@ local function set_cors_headers(conf, ctx)
     end
 
     core.response.set_header("Access-Control-Allow-Origin", ctx.cors_allow_origins)
-    if ctx.cors_allow_origins ~= "*" then
-        core.response.add_header("Vary", "Origin")
-    end
-
     core.response.set_header("Access-Control-Allow-Methods", allow_methods)
     core.response.set_header("Access-Control-Max-Age", conf.max_age)
     core.response.set_header("Access-Control-Expose-Headers", conf.expose_headers)
@@ -307,6 +303,9 @@ function _M.header_filter(conf, ctx)
         allow_origins = process_with_allow_origins_by_metadata(
                 conf.allow_origins_by_metadata, ctx, req_origin
         )
+    end
+    if conf.allow_origins ~= "*" then
+        core.response.add_header("Vary", "Origin")
     end
     if allow_origins then
         ctx.cors_allow_origins = allow_origins
--- a/t/plugin/cors3.t
+++ b/t/plugin/cors3.t
@@ -163,7 +163,7 @@ Origin: http://foo.example.org
 hello world
 --- response_headers
 Access-Control-Allow-Origin:
-Vary:
+Vary: Origin
 Access-Control-Allow-Methods:
 Access-Control-Allow-Headers:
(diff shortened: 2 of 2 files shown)


navendu-pottekkat:
@boekkooi-lengoo Thanks for opening the PR. Tagging @spacewander for review.

apisix/plugins/cors.lua
```diff
@@ -37,7 +38,7 @@ local metadata_schema = {
             type = "object",
             additionalProperties = {
                 type = "string",
-                pattern = origins_pattern
+                pattern = origins_metadata_pattern
```
spacewander, line 41: Why change this pattern?

apisix/plugins/cors.lua
```diff
@@ -178,6 +179,9 @@ function _M.check_schema(conf, schema_type)
             end
         end
     end
+    if conf.allow_origins == "" and (not conf.allow_origins_by_regex or next(conf.allow_origins_by_regex) == nil) then
```
spacewander, line 182: As allow_origins_by_regex is an array, we can use `#` to check if it's empty

apisix/plugins/cors.lua
boekkooi-lengoo: Because else the tests for the `metadata_schema` started failing and I don't understand what `metadata_schema` is used for so I wanted to save and not sorry.

However if you are fine saying that an empty string is a valid origin we can avoid this and I can fix the text.

apisix/plugins/cors.lua
boekkooi-lengoo: Thanks @spacewander I knew there should be a better way but my lua foo is not very good :slightly_smiling_face: 

I will update this once we are clear on the `metadata_schema` item.

apisix/plugins/cors.lua
spacewander: an empty string is not a valid origin. We can keep the same regex in metadata_schema and the others.

boekkooi-lengoo: I think this is incorrect as the allow_origins can be empty or unspecified if a regex is allow_origins_by_regex is specified.

However I didn't find any way to make `allow_origins` optional as it has a default value of `*`.

spacewander: I notice there is a fix recently: https://github.com/apache/apisix/pull/9028

boekkooi-lengoo:
@spacewander I have update the PR to only include the Vary header fix now.

spacewander (approved):

leslie-tsang (approved):

PR #9012, opened 2023-03-06
change: move server-info plugin into core

kingluo:
### Description

Add `GET /apisix/admin/data_planes` admin API, so that we could know how many data planes.

Example:

```bash
curl http://127.0.0.1:9180/apisix/admin/data_planes \
-H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X GET -s | jq
[
  {
    "value": {
      "etcd_version": "3.5.0",
      "hostname": "bar",
      "version": "3.2.0",
      "id": "de900110-a4b2-456a-9112-f3f76578657c",
      "boot_time": 1678095922
    },
    "key": "/apisix/data_plane/server_info/de900110-a4b2-456a-9112-f3f76578657c",
    "create_revision": "2687091",
    "version": "1",
    "mod_revision": "2687091"
  },
  {
    "value": {
      "etcd_version": "3.5.0",
      "hostname": "foo",
      "version": "3.2.0",
      "id": "e16b5da2-1b5c-4cd3-866c-7bcea7b32ffe",
      "boot_time": 1678095822
    },
    "create_revision": "2687090",
    "key": "/apisix/data_plane/server_info/e16b5da2-1b5c-4cd3-866c-7bcea7b32ffe",
    "lease": "7587868007797608997",
    "version": "1",
    "mod_revision": "2687090"
  }
]


curl http://127.0.0.1:9180/apisix/admin/data_planes/e16b5da2-1b5c-4cd3-866c-7bcea7b32ffe \
-H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X GET -s | jq
{
  "value": {
    "etcd_version": "3.5.0",
    "hostname": "foo",
    "version": "3.2.0",
    "id": "e16b5da2-1b5c-4cd3-866c-7bcea7b32ffe",
    "boot_time": 1678095822
  },
  "create_revision": "2687090",
  "key": "/apisix/data_plane/server_info/e16b5da2-1b5c-4cd3-866c-7bcea7b32ffe",
  "lease": "7587868007797608997",
  "version": "1",
  "mod_revision": "2687090"
}
```

### Checklist

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -237,6 +237,25 @@ local function run()
 end
 
 
+local function get_data_planes()
+    local key = "/data_plane/server_info"
+    local is_dir = true
+    local uri_segs = core.utils.split_uri(ngx.var.uri)
+    local id = uri_segs[5]
+    if id then
+        key = key .. "/" .. id
+        is_dir = false
+    end
+
+    local etcd_cli, prefix = core.etcd.get_etcd_cli()
+    key = prefix .. key
+    local res = etcd_cli:readdir(key)
+    if res.status == 200 then
+        core.response.exit(200, is_dir and res.body.kvs or res.body.kvs[1])
+    end
+end
+
+
 local function get_plugins_list()
     local api_ctx = {}
     core.ctx.set_vars_meta(api_ctx)
@@ -379,6 +398,16 @@ local uri_route = {
         paths = reload_event,
         methods = {"PUT"},
         handler = post_reload_plugins,
+    },
+    {
+        paths = [[/apisix/admin/data_planes]],
+        methods = {"GET"},
+        handler = get_data_planes,
+    },
+    {
+        paths = [[/apisix/admin/data_planes/*]],
+        methods = {"GET"},
+        handler = get_data_planes,
     },
 }
 
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -249,10 +249,25 @@ local function get_data_planes()
 
     local etcd_cli, prefix = core.etcd.get_etcd_cli()
     key = prefix .. key
-    local res = etcd_cli:readdir(key)
-    if res.status == 200 then
-        core.response.exit(200, is_dir and res.body.kvs or res.body.kvs[1])
+    local res, err = etcd_cli:readdir(key)
+    if err then
+        core.log.warn("failed to do etcd readdir: ", err)
+        core.response.exit(502, {error_msg =
+            "failed to do etcd readdir: " .. err})
     end
+
+    if res.status ~= 200 then
+        core.log.warn("failed to do etcd readdir: status=", res.status)
+        core.response.exit(502, {error_msg =
+            "failed to do etcd readdir: status=" .. res.status})
+    end
+
+    if res.body.kvs == nil then
+        core.log.warn("empty server-info")
+        core.response.exit(502, {error_msg = "empty server-info"})
+    end
+
+    core.response.exit(200, is_dir and res.body.kvs or res.body.kvs[1])
 end
 
 
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -252,19 +252,19 @@ local function get_data_planes()
     local res, err = etcd_cli:readdir(key)
     if err then
         core.log.warn("failed to do etcd readdir: ", err)
-        core.response.exit(502, {error_msg =
+        core.response.exit(503, {error_msg =
             "failed to do etcd readdir: " .. err})
     end
 
     if res.status ~= 200 then
         core.log.warn("failed to do etcd readdir: status=", res.status)
-        core.response.exit(502, {error_msg =
+        core.response.exit(503, {error_msg =
             "failed to do etcd readdir: status=" .. res.status})
     end
 
     if res.body.kvs == nil then
         core.log.warn("empty server-info")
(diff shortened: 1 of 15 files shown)


docs/en/latest/get-data-planes-from-etcd.md
```diff
@@ -0,0 +1,97 @@
...
+#
+-->
+
+To query all data planes information from etcd, we move `server-info` plugin into the APISIX core, and use it to report data plane information.
+
+The data plane information keeps the same format of `server-info` plugin output:
+
+https://apisix.apache.org/en/docs/apisix/plugins/server-info/
+
+## Admin API to query data planes
```
spacewander, line 35: We should move this into Admin API doc: https://github.com/apache/apisix/blob/master/docs/en/latest/admin-api.md

apisix/plugin.lua
```diff
@@ -746,6 +746,10 @@ function _M.init_worker()
     -- see https://github.com/apache/apisix/issues/3286
     _M.load()
 
+    if is_http then
+        require("apisix.plugins.server-info").init_worker()
```
spacewander, line 750: `require("apisix.plugins.xxx")` will make the module import complex. We should use a separate file to hold the functons.

apisix/plugins/server-info.lua
```diff
@@ -307,8 +307,8 @@ function _M.init()
 end
 
 
-function _M.destroy()
-    timers.unregister_timer("plugin#server-info", true)
+function _M.init()
+    core.log.warn("server-info plugin was moved into APISIX core, no need to enable it explicitly")
```
spacewander, line 311: We should update server-info plugin's doc too.

apisix/admin/init.lua
```diff
@@ -237,6 +237,40 @@ local function run()
...
+        key = key .. "/" .. id
+        is_dir = false
+    end
+
+    local etcd_cli, prefix = core.etcd.get_etcd_cli()
+    key = prefix .. key
+    local res, err = etcd_cli:readdir(key)
+    if err then
+        core.log.warn("failed to do etcd readdir: ", err)
+        core.response.exit(502, {error_msg =
```
spacewander, line 255: Like https://github.com/apache/apisix/blob/09cc4946bc051452960975b4dd7436a2497df346/apisix/admin/resource.lua#L122, we usually return 503 for this case

apisix/plugin.lua
kingluo: Do you mean moving `server-info.lua` to elsewhere? But we should keep compatibility so that users could still use this plugin like before but with warning logs.
If only for `init_worker()` function, then because it's tightly coupled with symbols in `server-info.lua`, so it seems no clean way to separate it.
And I think it's not so necessary. `apisix/plugin.lua` is exactly used to load a specific plugin. In fact, in the same function, it also requires Prometheus plugin with a similar path:
https://github.com/apache/apisix/blob/09cc4946bc051452960975b4dd7436a2497df346/apisix/plugin.lua#L740

spacewander: Yes, but we don't require Prometheus plugin. The required module is "apisix.plugins.prometheus.exporter", not the Prometheus plugin itself. We need to do the same with server-info.

kingluo: @spacewander Split server-info.lua into server-info/common.lua and server-info/init_worker.lua, have a check please.

docs/en/latest/get-data-planes-from-etcd.md
```diff
@@ -0,0 +1,103 @@
...
+| hostname     | string  | Hostname of the machine/pod APISIX is deployed to.                                                                     |
+
+## Admin API to query data planes
+
+* `GET /apisix/admin/data_planes`
+
+It returns a JSON array which contains all alive data planes.
+
+```bash
+curl http://127.0.0.1:9180/apisix/admin/data_planes \
```
spacewander, line 48: We can simply drop a link to the admin-api.md here?

docs/en/latest/get-data-planes-from-etcd.md
```diff
@@ -0,0 +1,45 @@
...
+| id           | string  | APISIX instance ID.                                                                                                    |
+| etcd_version | string  | Version of the etcd cluster used by APISIX. Will be `unknown` if the network to etcd is partitioned.                   |
+| version      | string  | Version of APISIX instance.                                                                                            |
+| hostname     | string  | Hostname of the machine/pod APISIX is deployed to.                                                                     |
+
+## Admin API to query data planes
+
+Refer to this link for detail:
+
+https://apisix.apache.org/en/docs/apisix/admin-api/#query-data-planes
```
spacewander, line 45: In most other cases, we use relative links like
https://github.com/apache/apisix/blob/7e292eadd36e738f84647634b0c351bc268ea529/docs/en/latest/terminology/plugin.md?plain=1#L35
https://github.com/apache/apisix/blob/7e292eadd36e738f84647634b0c351bc268ea529/docs/en/latest/terminology/route.md?plain=1#L82

@navendu-pottekkat 
Are the relative links preferred nowadays?

navendu-pottekkat: @spacewander Yes. It would be okay as long as the relative path mentioned is in the same repo. This would mean that links work on GitHub as well as the built website.

spacewander (approved):

Issue #8995, opened 2023-03-02
bug:  The gzip plugin does not work in special scenarios


zuiyangqingzhou:
### Current Behavior

0. Openresty as the upstream, the configuration file is as follows
```
pid logs/nginx.pid;
events {
    worker_connections 1024;
}

http {
    server {
        listen 8866;
        location / {
	    content_by_lua '
	    	 ngx.say("hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world. hello, world.")
	    ';
        }
    }
}
```
1. enable the proxy-rewrite plugin，config as follow
```
 "gzip": {
      "buffers": {
        "number": 64,
        "size": 4096
      },
      "comp_level": 5,
      "disable": false,
      "min_length": 200,
      "types": "*",
      "vary": true
    }
```
2. send a request without a via header
```
curl http://127.0.0.1:9080/test_gzip -H 'accept-encoding: gzip' -i
```
- output
```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Date: Thu, 02 Mar 2023 12:33:53 GMT
Server: Openresty
Vary: Accept-Encoding
Content-Encoding: gzip

Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
```
3. send a request with a via header
```
curl http://127.0.0.1:9080/test_gzip -H 'accept-encoding: gzip' -i -H 'via: aaa'
```
- output
```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Date: Thu, 02 Mar 2023 12:36:11 GMT
Server: Openresty
Vary: Accept-Encoding

[...]

PR #9024, opened 2023-03-07
feat(gzip): add via control option in gzip plugin

zuiyangqingzhou:
### Description

Fixes https://github.com/apache/apisix/issues/8995

### Checklist

### DIFF
--- a/apisix/plugins/gzip.lua
+++ b/apisix/plugins/gzip.lua
@@ -79,6 +79,10 @@ local schema = {
         },
         vary = {
             type = "boolean",
+        },
+        disable_via = {
+            type = "boolean",
+            default = false,
         }
     },
 }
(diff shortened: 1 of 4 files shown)


apisix/plugins/gzip.lua
```diff
@@ -100,6 +104,14 @@ function _M.check_schema(conf)
 end
 
 
+function _M.rewrite(conf, ctx)
+    if conf.disable_via == true then
+        -- According to nginx source code, via header will cause gzip not work
+        core.request.set_header(ctx, "via", nil)
```
spacewander, line 110: If this can satisfy your requirement, why not just use proxy-rewrite plugin to do so?

zuiyangqingzhou: I just think that this is a control that the gzip plugin lacks, and there is no need to introduce additional plugins. (of course, proxy-rewrite can also be implemented)

PR #9022, opened 2023-03-07
feat: add ldap-auth-advanced plugin

bzp2010:
### Description

According to the Proposal in #8958, this PR implements a new LDAP authentication plugin that is more standardized and flexible.

New #8958

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #9037, opened 2023-03-08
feat: bump lua-resty-ldap version for ldap-auth

bzp2010:
### Description

A new `lua-resty-ldap` has been released that provides a completely different API and more new features, we need to upgrade it to exclude API incompatibilities and prepare for another advanced LDAP plugin.

This upgrade is completely non-destructive to the user, and theoretically any test cases do not need to be modified to prove compatibility.

Head of #9022

### Checklist

### DIFF
--- a/apisix/plugins/ldap-auth.lua
+++ b/apisix/plugins/ldap-auth.lua
@@ -19,6 +19,7 @@ local ngx = ngx
 local ngx_re = require("ngx.re")
 local consumer_mod = require("apisix.consumer")
 local ldap = require("resty.ldap")
+local ok, ldap_cli = pcall(require, "resty.ldap.client")
 
 local schema = {
     type = "object",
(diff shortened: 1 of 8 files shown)


apisix/plugins/ldap-auth.lua
```diff
@@ -100,6 +100,11 @@ local function extract_auth_header(authorization)
 end
 
 function _M.rewrite(conf, ctx)
+    if not ok then -- ensure rasn library loaded
```
spacewander, line 103: For code style, we return 501 in the other similar cases:
https://github.com/apache/apisix/blob/7e292eadd36e738f84647634b0c351bc268ea529/apisix/plugins/gzip.lua#L104-L108

ci/common.sh
```diff
@@ -56,6 +56,19 @@ rerun_flaky_tests() {
     FLUSH_ETCD=1 prove --timer -I./test-nginx/lib -I./ $(echo "$tests" | xargs)
 }
 
+install_curl () {
+    wget https://curl.se/download/curl-7.88.0.tar.gz
```
spacewander, line 60: What about exacting the version out like the other install_xxx function?

bzp2010: Yes, it makes sense, fixed.

apisix/plugins/ldap-auth.lua
bzp2010: changed

spacewander (approved):

soulbird (approved):

PR #9043, opened 2023-03-09
feat: add nacos metadata suport, for customize  balancer

### LINKED ISSUE #5706: request help: How to route according to the configuration of nacos metadata?
### Issue description

I use nacos discovery, and for each microservice I register some additional metadata for finer-grained control

I know that the traffic-split plugin can match routes at a certain level of granularity
As far as I know, this plugin can't match upstreams based on metadata level

So, I would like to ask how to do this

Tks

### Environment

- apisix 2.10.2
- Centos7
- openresty/1.19.3.2
- etcd 3.5.0

charlesvhe:
### Description
In order to customize the nacos-based balancer, needs nacos instance metadata. So metadata needs to be added to the node information of apisix nacos discovery

Fixes #5706 request help: How to route according to the configuration of nacos metadata?

### Checklist

- [ X ] I have explained the need for this PR and the problem it solves
- [ X ] I have explained the changes or the new features added to this PR
- [ X ] I have added tests corresponding to this change
- [ X ] I have updated the documentation to reflect this change
- [ X ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

### DIFF
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -315,6 +315,7 @@ local function fetch_full_registry(premature)
                 host = host.ip,
                 port = host.port,
                 weight = host.weight or default_weight,
+                metadata = host.metadata or {},
             }
 
             -- docs: https://github.com/yidongnan/grpc-spring-boot-starter/pull/496


apisix/discovery/nacos/init.lua
```diff
@@ -315,6 +315,7 @@ local function fetch_full_registry(premature)
                 host = host.ip,
                 port = host.port,
                 weight = host.weight or default_weight,
+                metadata = host.metadata or {},
```
monkeyDluffy6017, line 318: Why the default value is an empty table, is nil more appropriate?

monkeyDluffy6017:
We need a test case in apisix/t/discovery/nacos.t

apisix/discovery/nacos/init.lua
charlesvhe: In fact, nacos currently returns an empty json object when metadata is not configured, which corresponds to an empty table in lua. In order to keep this feature stable, this is a failback mechanism. Prevent future destructive changes to existing apisix code if metadata does not return or null is returned by nacos next release

charlesvhe:
> 

I have read the existing test case and currently no component uses this metadata for testing. I will close this PR and submit another PR later for  a node_label_balancer component together

PR #9047, opened 2023-03-10
change: change the default router from radixtree uri to radixtree hos…

### LINKED ISSUE #8354: Change the default HTTP router from radixtree_uri to radixtree_host_uri

monkeyDluffy6017:
### Description

Update the default HTTP router from radixtree_uri to radixtree_host_uri .

> when Apache APISIX uses radixtree_uri as the default HTTP router,
there is a confusing scenario (even we can say it's a bug): Let's say
we have two routes, the first one requires the host matches
*.[example.com](http://example.com/), and the URI path is /anything, And the second one needs
the host matches [foo.example.com](http://foo.example.com/) exactly, and the URI path is also
/anything. In such a case, if we send an API request, we may hit the
first route, which is counterintuitive.
We also have a lot of voices from the community that users have been
in trouble since they think Apache APISIX will consider the HTTP host
match by default. However, it's not the case. So I'm here to propose
changing the default HTTP router to radixtree_host_uri.

Fixes #8354

### Checklist


docs/en/latest/terminology/router.md
```diff
@@ -38,13 +38,19 @@ A Router can have the following configurations:
...
     - `Prefix match`: Match with the given prefix. Use `*` to represent the given `uri` for prefix matching. For example, `/foo*` can match with `/foo/`, `/foo/a` and `/foo/b`.
     - `match priority`: First try an absolute match, if it didn't match, try prefix matching.
     - `Any filter attribute`: This allows you to specify any Nginx built-in variable as a filter, such as URL request parameters, request headers, and cookies.
   - `radixtree_uri_with_parameter`: Like `radixtree_uri` but also supports parameter match.
-  - `radixtree_host_uri`: Matches both host and URI of the request. Use `host + uri` as the primary index (based on the `radixtree` engine).
+  - `radixtree_host_uri`: (Default) Matches both host and URI of the request. Use `host + uri` as the primary index (based on the `radixtree` engine).
+
+:::note
+
+In version 3.2 and earlier, APISIX used `radixtree_uri` as the default Router. `radixtree_uri` has better performance than `radixtree_host_uri`, so if you have higher performance requirements and can live with the fact that `radixtree_uri` only matches uri, consider continuing to use `radixtree_uri` as the default Router.
```
spacewander, line 51: Only use the `uri` as the primary index, not only matches uri

spacewander (commented):
We need to change the route setting in https://github.com/apache/apisix/blob/master/t/router/radixtree-uri-host.t and other similar tests too.

monkeyDluffy6017:
> 

@spacewander do you mean remove the config like https://github.com/apache/apisix/blob/master/t/router/radixtree-host-uri.t#L29?

docs/en/latest/terminology/router.md
monkeyDluffy6017: Done

t/router/radixtree-method.t
```diff
@@ -16,9 +16,20 @@
 #
 use t::APISIX 'no_plan';
 
+our $yaml_config = <<_EOC_;
+apisix:
+    node_listen: 1984
+    router:
+        http: 'radixtree_uri'
```
spacewander, line 23: Can't this test pass with the default router?

spacewander (approved):

t/router/radixtree-method.t
monkeyDluffy6017: It can pass, should we use the default router here?

soulbird (approved):

moonming (approved):

t/router/radixtree-method.t
spacewander: Yes, it would be better. Adding special configuration here makes it like a special router is required.

monkeyDluffy6017: I was wondering if it would be better to specify the routing engine for these test cases on router, so that the default engine is changed and the test cases are not affected

spacewander: It's fine if you want to keep it.

PR #9116, opened 2023-03-20
feat: Add feature to check for stream route references when deleting and implement protocol matching

ro4i7:
### Description

Implemented a feature to check if a stream route is referenced by another stream route when deleting. Also, add a check to match the current stream route's protocol with the superior stream route's protocol. Log an info message if matched; otherwise, log an error message. Log an error message if there's an error while fetching the superior stream route.

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/stream/router/ip_port.lua
+++ b/apisix/stream/router/ip_port.lua
@@ -78,9 +78,30 @@ do
 
             local route = item.value
             if route.protocol and route.protocol.superior_id then
-                -- subordinate route won't be matched in the entry
-                -- TODO: check the subordinate relationship in the Admin API
-                goto CONTINUE
+                local admin_api = require("apisix.admin").check_http_admin()
+                if not admin_api then
+                    ngx.log(ngx.ERR, "failed to get http admin node")
+                    goto CONTINUE
+                end
+
+                local route_id = route.id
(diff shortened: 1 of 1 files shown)


monkeyDluffy6017:
hi @ro4i7, please make the ci pass

apisix/stream/router/ip_port.lua
```diff
@@ -78,9 +78,30 @@ do
 
             local route = item.value
             if route.protocol and route.protocol.superior_id then
-                -- subordinate route won't be matched in the entry
-                -- TODO: check the subordinate relationship in the Admin API
-                goto CONTINUE
+                local admin_api = require("apisix.admin").check_http_admin()
+                if not admin_api then
+                    ngx.log(ngx.ERR, "failed to get http admin node")
```
starsz, line 83: Can we use "core.log" instead?

ro4i7: > Can we use "core.log" instead?

@starsz Sorry, I have been out of the station, so replying late. Yeah definitely, I have updated it according to it, please check.

starsz (commented):
Can you fix the CI and add some testcase to cover it ?

monkeyDluffy6017:
I will close this PR, the functions involved in this PR don't even exist

PR #9127, opened 2023-03-21
openid-connect plugin - check token request against granted scopes 

whatvn:
### Description

Current openid connect plugin does not check for access token scope permission. For example, a client-id clientX has been granted for scope-a, apisix with openid connect has 2 routes set up with scope-a and scope-b for different client-id. 

ClientX then use client-id and client-secret to get access-token to request for service defined with scope-a, this access token can still be used to access service defined with scope-b since this plugin does not check for scope when doing oidc introspect. 

This fixes issue mentioned above by also checking allowed scope of access token with granted scope in oidc 

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -19,6 +19,7 @@ local core    = require("apisix.core")
 local ngx_re  = require("ngx.re")
 local openidc = require("resty.openidc")
 local random  = require("resty.random")
+local cjson = require "cjson"
 local string  = string
 local ngx     = ngx
 
@@ -304,7 +305,7 @@ function _M.rewrite(plugin_conf, ctx)
         conf.ssl_verify = "no"
     end
 
-    local response, err, session, _
+    local response, err, session, result, _
 
     if conf.bearer_only or conf.introspection_endpoint or conf.public_key then
         -- An introspection endpoint or a public key has been configured. Try to
@@ -312,7 +313,9 @@ function _M.rewrite(plugin_conf, ctx)
         -- request header. Otherwise, return a nil response. See below for
         -- handling of the case where the access token is stored in a session cookie.
         local access_token, userinfo
+        core.log.debug("conf to introspect:", cjson.encode(conf))
         response, err, access_token, userinfo = introspect(ctx, conf)
+        core.log.debug("introspect response:", cjson.encode(response))
 
         if err then
             -- Error while validating token or invalid token.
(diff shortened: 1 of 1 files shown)


whatvn:
@starsz @spacewander  please take a look, thank!

starsz:
> @starsz @spacewander please take a look, thank!

Hi, @whatvn.I don't see any doc to introduce that we can add scope in the request body of the introspection request.
But I can see that we can get the scope from the introspection response, and it's optional.

Can you give me some information about this?  Thanks very much.

whatvn:
@starsz according to rfc7662 https://www.rfc-editor.org/rfc/rfc7662#section-2.1
```
The introspection endpoint MAY accept other OPTIONAL parameters to
   provide further context to the query.  For instance, an authorization
   server may desire to know the IP address of the client accessing the
   protected resource to determine if the correct client is likely to be
   presenting the token.  The definition of this or any other parameters
   are outside the scope of this specification, to be defined by service
   documentation or extensions to this specification.  If the
   authorization server is unable to determine the state of the token
   without additional information, it SHOULD return an introspection
   response indicating the token is not active as described in
   [Section 2.2](https://www.rfc-editor.org/rfc/rfc7662#section-2.2).
```

Client can also provide additional parameter in order to let authorization server to optimize search query. That's why in oidc resty module which we are using, it allows us to add **scope** to introspect the token. 

With your question, I think I may also add another check, to check in case authorization (other than Ory Hydra that I am testing) does not support for **scope** in request, we still can be able to check granted scope. What do you think?

starsz:
> https://www.rfc-editor.org/rfc/rfc7662#section-2.1

Thanks for providing the RFC docs. I agree with you.
But maybe it's a broken change, so we should add an option to control this changes.

whatvn:
@starsz I add options to control this change, also update doc (en) to clarify new options and behaviour. Please take a look, thank!

starsz:
> @starsz I add options to control this change, also update doc (en) to clarify new options and behaviour. Please take a look, thank!

Hi, @whatvn Can you add test case of this?

whatvn:
@starsz let me try again, I tried to run current test case on my dev machine followed document to setup test but it failed everytime.

starsz:
> @starsz let me try again, I tried to run current test cases on my dev machine followed document to setup test but it failed everytime.

What's the error that your dev machine report?

whatvn:
@starsz please find output of command in attached file. 
```bash
prove -I. -I../test-nginx/inc -I../test-nginx/lib -r t/plugin/openid-connect.t  >t.log 2>&1
```
I modified TEST 4 to get HTTP response code instead of just "pass" the test, it started failing at TEST 4, couldn't install simple rule and return 500 instead of 200 

[t.log](https://github.com/apache/apisix/files/11109461/t.log)

whatvn:
@starsz I built and run test again on different clean dev machine, still got this error. Any idea?

whatvn:
@starsz I run the test on master, not my branch. Just merge master again with updates and test is still fail

whatvn:
@starsz question, how do we setup keycloak to perform the test? Do we have document on how to setup test for openid-connect? Thank

starsz:
> @starsz question, how do we setup keycloak to perform the test? Do we have document on how to setup test for openid-connect? Thank

Hi, @whatvn. As you can see in the workflow:

[`https://github.com/apache/apisix/blob/master/ci/pod/docker-compose.plugin.yml`](https://github.com/apache/apisix/blob/master/.github/workflows/centos7-ci.yml#L119-L139)

APISIX will start keycloak by docker-compose.
The docker-compose file is here: https://github.com/apache/apisix/blob/master/ci/pod/docker-compose.plugin.yml

> Do we have document on how to setup test for openid-connect.

We have a blog of [Getting Started with APISIX Test](https://apisix.apache.org/blog/2022/06/27/getting-start-with-apisix-test-cases/).

PR #9147, opened 2023-03-22
feat(cli): support bypassing Admin API Auth by configuration

### LINKED ISSUE #9071: feat: As a user, I want to allow external ips in `allow_admin` without adminKey in debug mode, so that users can request containered APISIX Admin API from the host for test
### Description

As a user, I want to allow external ips in `allow_admin` without adminKey in debug mode, so that users can request containered APISIX Admin API from the host for test.

To simplify the operating steps in the "Getting Started" document, we want to make empty adminKey.

However, the APISIX now does not allow the requests without adminKey to visit Admin API except `127.0.0.0/24`, 
https://github.com/apache/apisix/blob/master/apisix/cli/ops.lua#L191
so that users cannot visit containered APISIX Admin API from the host.

Such a strategy is made for safety, but we can support it in debug mode, so that users can try the APISIX easily with "Getting Started".

What I want to do is allow the external custom ips in `allow_admin` with empty `admin_key` in debug mode, like below:
```
deployment:
  admin:
    allow_admin:
      - 0.0.0.0/0
    admin_key:
    # empty
```

An-DJ:
### Description

APISIX now does not allow the requests with empty adminKey to access Admin API except 127.0.0.0/24.

The user can allow this by the configuration below.

```
deployment:
  admin:
    admin_key_required: false    # Default value is true
```
Fixes #9071 

image

### Checklist

### DIFF
--- a/t/admin/api.t
+++ b/t/admin/api.t
@@ -172,7 +172,20 @@ GET /apisix/admin/routes
 
 
 
-=== TEST 11: Access without api key, but admin_key_required=true
+=== TEST 11: Access with wrong api key, and admin_key_required=true
+--- yaml_config
+deployment:
+  admin:
+    admin_key_required: true
+--- more_headers
+X-API-KEY: wrong-key
+--- request
+GET /apisix/admin/routes
+--- error_code: 401
+
+
+
+=== TEST 12: Access without api key, and admin_key_required=true
 --- yaml_config
 deployment:
   admin:
@@ -183,7 +196,7 @@ GET /apisix/admin/routes
 
 
 
-=== TEST 12: Access with api key, but admin_key_required=false
+=== TEST 13: Access with api key, but admin_key_required=false
 --- yaml_config
 deployment:
   admin:
@@ -193,10 +206,27 @@ X-API-KEY: edd1c9f034335f136f87ad84b625c8f1
 --- request
 GET /apisix/admin/routes
 --- error_code: 200
+--- error_log
+Admin key is bypassed!
 
 
 
-=== TEST 13: Access without api key, but admin_key_required=false
+=== TEST 14: Access with wrong api key, but admin_key_required=false
+--- yaml_config
+deployment:
+  admin:
+    admin_key_required: false
+--- more_headers
+X-API-KEY: wrong-key
+--- request
+GET /apisix/admin/routes
+--- error_code: 200
+--- error_log
+Admin key is bypassed!
+
+
+
+=== TEST 15: Access without api key, but admin_key_required=false
 --- yaml_config
 deployment:
   admin:


apisix/cli/ngx_tpl.lua
```diff
@@ -51,6 +51,8 @@ worker_shutdown_timeout {* worker_shutdown_timeout *};
 env APISIX_PROFILE;
 env PATH; # for searching external plugin runner's binary
 
+env APISIX_ALLOW_NONE_AUTHENTICATION; # allow any IP with empty admin_key
```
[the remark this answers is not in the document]
An-DJ: IMO we just bypass the configuration check here.

That is, if the user set `APISIX_ALLOW_NONE_AUTHENTICATION=true` but still configure the adminKey, the key also works.

So this env var just `allow` none auth, not `bypass`.

monkeyDluffy6017:
If this option is turned on, can we provide some hints, otherwise there may be security risks

leslie-tsang:
> If this option is turned on, can we provide some hints, otherwise there may be security risks Do we need to add this inspection to the inspection script? @leslie-tsang

Yes for sure, need to log it in error log as well.

moonming (changes requested):

monkeyDluffy6017:
Please make the CI pass

apisix/cli/ngx_tpl.lua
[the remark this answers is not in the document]
An-DJ: Replaced by APISIX_BYPASS_ADMIN_API_AUTH.

apisix/admin/init.lua
```diff
@@ -66,6 +67,12 @@ local router
 
 
 local function check_token(ctx)
+    -- check if APISIX_BYPASS_ADMIN_API_AUTH=true
+    local none_auth = getenv("APISIX_BYPASS_ADMIN_API_AUTH")
```
moonming, line 71:
```suggestion
    local bypass_admin_api_auth = getenv("APISIX_BYPASS_ADMIN_API_AUTH")
```

apisix/cli/ops.lua
```diff
@@ -190,6 +190,15 @@ local function init(env)
...
 
+    -- check if APISIX_BYPASS_ADMIN_API_AUTH=true
+    local allow_none_auth = getenv("APISIX_BYPASS_ADMIN_API_AUTH")
+    if allow_none_auth == "true" then
+        checked_admin_key = true
+        print("Warning! AdminKey is bypassed because of APISIX_BYPASS_ADMIN_API_AUTH=true.",
+            "If you are deploying APISIX in a production environment,",
+            "please disable it and set a secure password for the adminKey!")
+    end
+
```
moonming, line 201: Why are there duplicate codes?

t/admin/token.t
```diff
@@ -177,3 +177,21 @@ PUT /apisix/admin/plugins/reload?api_key=4054f7cf07e344346cd3f287985e76a2
...
+=== TEST 10: access without api key
+--- request
+GET /apisix/admin/routes
+--- error_code: 401
+
+
+
+=== TEST 11: access without api key, but APISIX_BYPASS_ADMIN_API_AUTH=true
+--- main_config
+env APISIX_BYPASS_ADMIN_API_AUTH=true;
```
moonming, line 192: need to add test cases for `APISIX_BYPASS_ADMIN_API_AUTH` is not `true`

moonming: and if we set `APISIX_BYPASS_ADMIN_API_AUTH` to `abc`, what will happen?

moonming: do we have a schema check for these configures?

apisix/cli/ops.lua
An-DJ: Do you mean that hint appears in both stdout and log?

1. `opt.lua` checks if allow_admin in config.yaml is correct before APISIX starts. We need to bypass the adminKey check here.
2. `init.lua` checks the adminKey when accessing the Admin API. We need to bypass check_token, too.

That's why I have implemented bypass-logic in two different stage.(before and after the start).

t/admin/token.t
An-DJ: > and if we set `APISIX_BYPASS_ADMIN_API_AUTH` to `abc`, what will happen?

All values that are not true will be considered false.

apisix/cli/ops.lua
An-DJ: In other words, opt.lua does a static check on the configuration file (if it is not 127.0.0.0/24 and the adminKey is empty, APISIX will not start). 

We also have to bypass the checks here.

t/admin/token.t
moonming: it's too hacky. We need a configure checker(if apisix not has) to check schema and values.

apisix/cli/ops.lua
moonming: If there is repeated code, then it needs to be abstracted into a function

t/admin/token.t
An-DJ: > need to add test cases for `APISIX_BYPASS_ADMIN_API_AUTH` is not `true`

Done.

apisix/cli/ngx_tpl.lua
```diff
@@ -51,6 +51,8 @@ worker_shutdown_timeout {* worker_shutdown_timeout *};
 env APISIX_PROFILE;
 env PATH; # for searching external plugin runner's binary
 
+env APISIX_BYPASS_ADMIN_API_AUTH; # bypass admin api auth
```
moonming, line 54:
```suggestion
env ADMIN_API_TOKEN_REQUIRED = true; # token is required to access the admin API
```

t/admin/token.t
moonming: > This variable is currently a reserved environment variable. It is not a configuration item in the configuration file (config.yaml).

can we move it to config.yaml? near by the token?

apisix/cli/ngx_tpl.lua
An-DJ: Sounds weird...

I prefer the variable names in bitnami/etcd that bypass authentication:

image

`ALLOW_NONE_AUTHENTICATION` looks more natural. 

If we use `ADMIN_API_TOKEN_REQUIRED`, users should bypass authentication by specifying `false`.

IMHO, `false` is not a good value to make users feel positive.

An-DJ: Ref: https://hub.docker.com/r/bitnami/etcd/

moonming: I don't think etcd has a good name for this configuration.
Straightforward, without explanation, is a good name.

moonming: > IMHO, `false` is not a good value to make users feel positive.

you are right. So we don't suggest the user set it to `false` in prod.

t/admin/token.t
moonming: can we set it on both sides?

conf/config-default.yaml
```diff
@@ -588,6 +588,10 @@ deployment:
   role_traditional:
     config_provider: etcd
   admin:
+    # If admin_key is required, default value is true.
```
monkeyDluffy6017, line 591: Is `admin_key required or not` better?

An-DJ: Good point.

An-DJ: Done.

apisix/cli/ops.lua
```diff
@@ -189,6 +189,13 @@ local function init(env)
        and #allow_admin == 1 and allow_admin[1] == "127.0.0.0/24" then
         checked_admin_key = true
     end
+    -- check if admin_key is required
```
monkeyDluffy6017, line 192: It seems that this logic is not necessary, It's just for start APISIX, but you can not access apisix without admin key, right?
```
    if yaml_conf.apisix.enable_admin and allow_admin
       and #allow_admin == 1 and allow_admin[1] == "127.0.0.0/24" then
        checked_admin_key = true
    end
```

apisix/admin/init.lua
```diff
@@ -395,6 +401,13 @@ function _M.init_worker()
     events.register(reload_plugins, reload_event, "PUT")
 
     if ngx_worker_id() == 0 then
+        -- check if admin_key is required
+        if not local_conf.deployment.admin.admin_key_required then
+            core.log.warn("AdminKey is bypassed.",
+                "If you are deploying APISIX in a production environment,",
+                "please disable it and set a secure password for the adminKey!")
```
monkeyDluffy6017, line 408: Is this better ?
```           
 core.log.warn("Admin key is bypassed! ",
                        "If you are deploying APISIX in a production environment, ",
                        "please disable it and set a secure password for the admin Key!")
```

apisix/cli/ops.lua
```diff
@@ -189,6 +189,13 @@ local function init(env)
...
+    if not yaml_conf.deployment.admin.admin_key_required then
+        checked_admin_key = true
+        print("Warning! AdminKey is bypassed. "
```
monkeyDluffy6017, line 195: ditto

t/cli/test_admin.sh
```diff
@@ -189,6 +189,62 @@ fi
...
+' > conf/config.yaml
+
+make init > output.log 2>&1 | true
+
+if grep -E "ERROR: missing valid Admin API token." output.log > /dev/null; then
+    echo "failed: should not show 'ERROR: missing valid Admin API token.'"
+    exit 1
+fi
+
+if ! grep -E "Warning! AdminKey is bypassed" output.log > /dev/null; then
```
monkeyDluffy6017, line 228: `Admin key` is better?

t/cli/test_admin.sh
```diff
@@ -189,6 +189,62 @@ fi
...
+echo '
+deployment:
+  admin:
+    admin_key_required: invalid-value
+' > conf/config.yaml
+
+make init > output.log 2>&1 | true
+
+if grep -E "path[deployment->admin->admin_key_required] expect: boolean, but got: string" output.log > /dev/null; then
+    echo "failed: should show 'expect: boolean, but got: string'"
```
monkeyDluffy6017, line 242: echo "check admin_key_required value failed: should show 'expect: boolean, but got: string'"

apisix/cli/ops.lua
An-DJ: IMHO These codes does not conflict with the current feature.

This logic is to allow `127.0.0.0/24` authentication-free access to the Admin API by default.

t/cli/test_admin.sh
An-DJ: This is the schema check error msg automatically created... It seems that it cannot be modified by custom.

monkeyDluffy6017: The warning message prefix is needed: `check admin_key_required value failed`, or you will not what happened only with `failed: should show 'expect: boolean, but got: string`

t/cli/test_admin.sh
```diff
@@ -189,6 +189,62 @@ fi
...
 echo "pass: missing admin key and only allow 127.0.0.0/24 to access admin api"
 
+# allow any IP to access admin api with empty admin_key, when admin_key_required=true
+
+git checkout conf/config.yaml
+
+echo '
+deployment:
+  admin:
+    admin_key_required: true
```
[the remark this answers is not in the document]
An-DJ: True is assigned to admin_key_required by default in `config-default.yaml`.

IMHO we do not need to add more test cases to test the missing case.

All the test cases now under the `t/` can validate it, which not configuring the `admin_key_required`.

apisix/cli/ops.lua
monkeyDluffy6017: If this code is useless, would you mind creating a PR to remove it?

An-DJ: > OK, it's just for starting, righting?
Yes. It just a check before starting.

> if set `127.0.0.0/24`, the user cann't access the APISIX? I don't find the code
I mean if set `127.0.0.0/24`, the user **CAN** access the APISIX with empty admin key configuraion. Otherwise the APISIX will start failed (such as `0.0.0.0/0` in allow_admin but empty admin_key)..

An-DJ: It is a default behavior that allow the user access Admin API from local.

t/cli/test_admin.sh
An-DJ: I mean that this msg is automatically created and outputed by the schema check step.

Total output:
image

It seems that we cannot easily custom info...

I think the grep text now has enough to help to know "what happened".

image

monkeyDluffy6017: OK, this is for the output file, i mean what you `echo` to the screen

An-DJ: Got it.

An-DJ: Done.

apisix/cli/ops.lua
monkeyDluffy6017: I checked this, after set `127.0.0.0/24` and remove the admin key:
```
curl http://127.0.0.1:9180/apisix/admin/upstreams/1 -X PUT -d '
{
    "type": "chash",
    "key": "remote_addr",
    "nodes": {
        "127.0.0.1:80": 1,
        "foo.com:80": 2
    }
}'
{"error_msg":"failed to check token"}
```

monkeyDluffy6017: We talk about this in another PR

monkeyDluffy6017 (dismissed):

leslie-tsang (approved):

apisix/admin/init.lua
```diff
@@ -395,6 +401,13 @@ function _M.init_worker()
     events.register(reload_plugins, reload_event, "PUT")
 
     if ngx_worker_id() == 0 then
+        -- check if admin_key is required
+        if local_conf.deployment.admin.admin_key_required == false then
+            core.log.warn("Admin key is bypassed! ",
+                "If you are deploying APISIX in a production environment, ",
+                "please disable it and set a secure password for the admin Key!")
```
moonming, line 408: admin key is not password

apisix/cli/ops.lua
```diff
@@ -189,6 +189,13 @@ local function init(env)
        and #allow_admin == 1 and allow_admin[1] == "127.0.0.0/24" then
         checked_admin_key = true
     end
+    -- check if admin_key is required
+    if yaml_conf.deployment.admin.admin_key_required == false then
+        checked_admin_key = true
+        print("Warning! Admin key is bypassed! "
+                .. "If you are deploying APISIX in a production environment, "
+                .. "please disable it and set a secure password for the admin Key!")
+    end
```
moonming, line 198: duplicate code

conf/config-default.yaml
```diff
@@ -588,6 +588,10 @@ deployment:
   role_traditional:
     config_provider: etcd
   admin:
+    # admin_key required or not. Default value is true.
+    # Bypass the Admin API authentication by modifying this value to false if needed.
```
moonming, line 592:
```suggestion
    # admin API authentication is enabled by default. 
    # Set it false in the production environment will cause a serious security issue.
```

apisix/admin/init.lua
moonming, line 408: Please write clearly what configuration to disable

t/admin/api.t
```diff
@@ -156,3 +156,40 @@ X-API-VERSION: v2
...
+deployment:
+  admin:
+    admin_key_required: true
+--- request
+GET /apisix/admin/routes
+--- error_code: 401
+
+
+
+=== TEST 12: Access without api key, but admin_key_required=false
```
moonming, line 186: no test case for `Access with api key, but admin_key_required=false`

An-DJ: Done.

t/admin/token.t
moonming: Good. we can write this to docs.

apisix/cli/ops.lua
moonming: Please add an issue for the duplicate code and fix it later.

apisix/admin/init.lua
An-DJ: Corrected.

conf/config-default.yaml
An-DJ: Done.

moonming (dismissed):

t/admin/api.t
```diff
@@ -156,3 +156,51 @@ X-API-VERSION: v2
...
+
+
+=== TEST 12: Access without api key, but admin_key_required=true
+--- yaml_config
+deployment:
+  admin:
+    admin_key_required: true
+--- request
+GET /apisix/admin/routes
+--- error_code: 401
```
moonming, line 193: Why are these two test cases the same?

monkeyDluffy6017: It should be `Access with api key, but admin_key_required=false`

An-DJ: Fixed

moonming (approved):

leslie-tsang (approved):

PR #9150, opened 2023-03-23
fix: host_hdr should not be false

kingluo:
### Description

Fixes https://github.com/api7/lua-resty-healthcheck/pull/15#discussion_r1144565155

### Checklist

### DIFF
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -135,7 +135,7 @@ local function create_checker(upstream)
     local host = upstream.checks and upstream.checks.active and upstream.checks.active.host
     local port = upstream.checks and upstream.checks.active and upstream.checks.active.port
     local up_hdr = upstream.pass_host == "rewrite" and upstream.upstream_host
-    local use_node_hdr = upstream.pass_host == "node"
+    local use_node_hdr = upstream.pass_host == "node" or nil
     for _, node in ipairs(upstream.nodes) do
         local host_hdr = up_hdr or (use_node_hdr and node.domain)
         local ok, err = checker:add_target(node.host, port or node.port, host,


apisix/upstream.lua
```diff
@@ -134,10 +134,13 @@ local function create_checker(upstream)
...
-    local up_hdr = upstream.pass_host == "rewrite" and upstream.upstream_host
-    local use_node_hdr = upstream.pass_host == "node"
     for _, node in ipairs(upstream.nodes) do
-        local host_hdr = up_hdr or (use_node_hdr and node.domain)
+        local host_hdr
+        if upstream.pass_host == "rewrite" and upstream.upstream_host then
+            host_hdr = upstream.upstream_host
+        elseif upstream.pass_host == "node" and node.domain then
+            host_hdr = node.domain
+        end
```
monkeyDluffy6017, line 143: The `up_hdr` won't affect the result of `host_hdr`. would this be better?
```
    local use_node_hdr = upstream.pass_host == "node" or nil     <== here
    for _, node in ipairs(upstream.nodes) do
        local host_hdr = up_hdr or (use_node_hdr and node.domain)
```

kingluo: @monkeyDluffy6017 Yes, it's the [same](https://www.lua.org/manual/5.1/manual.html#2.5.3), but it's not as clear as explicit conditional statements.

monkeyDluffy6017: But we don't need to compare `upstream.pass_host == "rewrite"` and `upstream.pass_host == "node"` every time in loop

monkeyDluffy6017 (approved):

leslie-tsang (approved):

Neilblaze (approved):

PR #9176, opened 2023-03-27
docs: updated ssl sni parameter requirement in admin API

kayx23:
SNI is only a required parameter iff `type` is `server`, meaning the SSL object is used in a downstream communication between APISIX and user, where APISIX acts as the server.

https://github.com/apache/apisix/blob/81149cd31567f6a86e33100fa7d09d7550073157/apisix/schema_def.lua#L778-L792


kayx23:
Discussed with @kingluo to understand how SNI works for downstream and upstream segment of traffic a few days ago. Only to find out the doc isn't that accurate so I'm making a small edit but closing a large understanding gap...

kingluo (approved):

moonming (approved):

Issue #9179, opened 2023-03-28
feat: proxy_rewrite plugin support multiple regex pattern matching


fengxsong:
### Description

Currently the `regex_uri` field in plugin `proxy_rewrite` is a slice type and only supports containing exactly 2 parameters, the first one is `what matched`, and the second is `where to go`. IMO, **multiple regex pattern matching** is a very common need, so we don't need to write multiple proxy configs in one single rule.
So I have a proposal if we can support an even number of parameters(compatible with the previous version)?

PR #9194, opened 2023-03-29
feat: proxy_rewrite support miltiple regex pattern matching

fengxsong:
### Description

As explained in #9179, this PR is a preliminary implementation. The following changes are included:
- ~~a new property that named `none_match_abort`, when `none_match_abort=true` and none of regex_uri matched, apisix will return 404 directly, for compatibility it defaults to `false`~~;
- remove the length limit of field `regex_uri`, but it must have a even length;

With this feature, we can now configure `regex_uri` like

```
......
      name: proxy-rewrite
      enable: true
      config:
        regex_uri:
        - "/v1/user/([:A-Za-z0-9_-]+)/follow"
        - "/api/v1/user/$1/follow"
        - "/v1/user/([:A-Za-z0-9_-]+)/test"
        - "/api/v1/user/$1/test"
        # - can add more regex patterns here
......
```

### Checklist

- [x] I have updated the documentation to reflect this change (Not yet, I will update the document when this proposal is accepted)


monkeyDluffy6017:
Hi @fengxsong, Test cases are needed!

fengxsong:
> Hi @fengxsong, Test cases are needed!

done

fengxsong:
ping @monkeyDluffy6017

monkeyDluffy6017:
I will check this later, THX @fengxsong

apisix/plugins/proxy-rewrite.lua
```diff
@@ -263,109 +274,118 @@ do
     end
 
 
-function _M.rewrite(conf, ctx)
-    for _, name in ipairs(upstream_names) do
-        if conf[name] then
-            ctx.var[upstream_vars[name]] = conf[name]
+    function _M.rewrite(conf, ctx)
```
kingluo, line 277: Don't change the original indent of this file in this PR, which should reflect this feature only.
If necessary, submit another PR to fix it.

fengxsong: Files are automatically formatted by vscode editor. I'll revert changes except code change related to this feature only.

fengxsong:
Is the ci failure caused by this code change? https://github.com/apache/apisix/actions/runs/4685205482/jobs/8302219899?pr=9194

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +288,36 @@ function _M.rewrite(conf, ctx)
...
-                                   conf.regex_uri[2], "jo")
-        if not uri then
-            local msg = "failed to substitute the uri " .. ctx.var.uri ..
-                        " (" .. conf.regex_uri[1] .. ") with " ..
-                        conf.regex_uri[2] .. " : " .. err
-            core.log.error(msg)
-            return 500, {message = msg}
+        local error_msg
+        for i = 1, #conf.regex_uri, 2 do
+            local captures, err = re_match(upstream_uri, conf.regex_uri[i], "jo")
```
kingluo, line 293: Why not do substitution directly? If successful, then break.

fengxsong: > Why not do substitution directly? If successful, then break.

variable `ctx.proxy_rewrite_regex_uri_captures` will be used later modifying request headers.

kingluo:
@fengxsong please add doc to describe this feature.

docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -38,7 +38,8 @@ The `proxy-rewrite` Plugin rewrites Upstream proxy information such as `scheme`,
...
+| regex_uri                   | array[string] | False    |         |                                                                                                                                        | New upstream forwarding address. Regular expressions can be used to match the URL from client. If it matches, the URL template is forwarded to the Upstream otherwise, the URL from the client is forwarded if `none_match_abort=false`. When both `uri` and `regex_uri` are configured, `uri` is used first. For example, `["^/iresty/(.*)/(.*)/(.*)", "/$1-$2-$3", ^/theothers/(.*)/(.*)", "/theothers/$1-$2"]`. Here, the first and the third elements are the regular expressions to match, the second and the fourth elements are the URL templates forwarded to the Upstream. Length of this property must be a even number. |
```
kingluo, line 42: It's better to declare it supports multiple patterns for matching and substitution. It will iterate all patterns one by one until a successful match and substitution.

apisix/plugins/proxy-rewrite.lua
```diff
@@ -191,11 +194,17 @@ function _M.check_schema(conf)
     end
 
     if conf.regex_uri and #conf.regex_uri > 0 then
-        local _, _, err = re_sub("/fake_uri", conf.regex_uri[1],
-                                   conf.regex_uri[2], "jo")
-        if err then
-            return false, "invalid regex_uri(" .. conf.regex_uri[1] ..
-                            ", " .. conf.regex_uri[2] .. "): " .. err
+        -- remove this check util 'multipleOf' is supported
```
soulbird, line 197: What‘s means？

t/plugin/proxy-rewrite3.t
```diff
@@ -771,3 +771,127 @@ passed
...
+passed
+
+
+
+=== TEST 34: hit
+--- request
+GET /test/plugin/proxy/rewrite/hello HTTP/1.1
+--- http_config
+    server {
+        listen 8125;
```
soulbird, line 823: There are many URIs for testing in lib.server, usually, it is not necessary to start the server separately

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +288,36 @@ function _M.rewrite(conf, ctx)
...
-            return 500
+        if error_msg ~= nil and #error_msg > 0 then
+            core.log.error(error_msg)
+            return 500, { error_msg = error_msg }
         end
-        ctx.proxy_rewrite_regex_uri_captures = m
 
-        upstream_uri = uri
+        if ctx.proxy_rewrite_regex_uri_captures == nil and conf.none_match_abort then
+            return 404, { error_msg = "404 Route Not Found" }
```
soulbird, line 319: This error message is wrong, but the rewrite was not successful, not that the route did not match successfully

soulbird, line 319: This error message is wrong, but the rewrite was not successful, not that the route did not match successfully

apisix/plugins/proxy-rewrite.lua
fengxsong: I had the wrong perception of `multipleOf` validator. it only take affect on number type but not length of array. I‘ll fix this later.

fengxsong: > What‘s means？

I had the wrong perception of `multipleOf` validator. it only take affect on number type but not length of array.

fengxsong:
> I don't understand why the none_match_abort parameter should be added, the proxy-rewrite plugin should not be able to block the request

There are some scenarios we don't want to expose all of our APIs, for example.

```
      match:
        hosts:
        - echo.example.local
        paths:
        - /v1/user/*
      name: echo
      plugins:
      - config:
          none_match_abort: true
          regex_uri:
          - /v1/user/([:A-Za-z0-9_-]+)/follow
          - /api/v1/user/$1/follow
          - /v1/user/([:A-Za-z0-9_-]+)/comment
          - /api/v1/user/$1/comment
        enable: true
        name: proxy-rewrite
```

we can only expose `/v1/user/([:A-Za-z0-9_-]+)/follow` and `/v1/user/([:A-Za-z0-9_-]+)/comment`, and hide the rest API endpoints.

From the point of view of ability classification, `none_match_abort` shouldn't be here. so this might needs more discussions.

apisix/plugins/proxy-rewrite.lua
```diff
@@ -61,12 +61,15 @@ local schema = {
             type        = "string",
             enum        = schema_method_enum
         },
+        none_match_abort = {
```
monkeyDluffy6017, line 64: It seems that the `none_match_abort` feature has nothing to do with this pr?

fengxsong: > It seems that the `none_match_abort` feature has nothing to do with this pr?

please check https://github.com/apache/apisix/pull/9194#issuecomment-1508145751

monkeyDluffy6017: I don't think uri blocking should be done in this plugin, it can be combined with other plugins, keep this plugin simple

fengxsong: > I don't think uri blocking should be done in this plugin, it can be combined with other plugins, keep this plugin simple

OK. I'll make another commit to remove this field later.

apisix/plugins/proxy-rewrite.lua
```diff
@@ -216,7 +218,6 @@ function _M.check_schema(conf)
     return true
 end
 
-
```
monkeyDluffy6017, line 219: Why do you remove this blank?

fengxsong: Is it necessary to keep it?

apisix/plugins/proxy-rewrite.lua
```diff
@@ -191,11 +188,16 @@ function _M.check_schema(conf)
     end
 
     if conf.regex_uri and #conf.regex_uri > 0 then
-        local _, _, err = re_sub("/fake_uri", conf.regex_uri[1],
-                                   conf.regex_uri[2], "jo")
-        if err then
-            return false, "invalid regex_uri(" .. conf.regex_uri[1] ..
-                            ", " .. conf.regex_uri[2] .. "): " .. err
+        if (#conf.regex_uri % 2 ~= 0) then
+            return false, "length of regex_uri should be a even number"
```
monkeyDluffy6017, line 192: The length of regex_uri should be an even number

apisix/plugins/proxy-rewrite.lua
```diff
@@ -65,8 +65,6 @@ local schema = {
             description = "new uri that substitute from client uri " ..
                           "for upstream, lower priority than uri property",
             type        = "array",
-            maxItems    = 2,
-            minItems    = 2,
```
monkeyDluffy6017, line 69: the limit `minItems` should be retained

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +281,32 @@ function _M.rewrite(conf, ctx)
...
+                end
+                break
+            end
         end
-
-        local m, err = re_match(upstream_uri, conf.regex_uri[1], "jo")
-        if not m and err then
-            core.log.error("match error in proxy-rewrite plugin, please check: ", err)
-            return 500
+        if error_msg ~= nil and #error_msg > 0 then
```
monkeyDluffy6017, line 306: why do you check the error_msg's length? `if error_msg ~= nil` is enough

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +281,32 @@ function _M.rewrite(conf, ctx)
...
-
-        local uri, _, err = re_sub(upstream_uri, conf.regex_uri[1],
-                                   conf.regex_uri[2], "jo")
-        if not uri then
-            local msg = "failed to substitute the uri " .. ctx.var.uri ..
-                        " (" .. conf.regex_uri[1] .. ") with " ..
-                        conf.regex_uri[2] .. " : " .. err
-            core.log.error(msg)
-            return 500, {message = msg}
+        local error_msg
```
monkeyDluffy6017, line 284: Need a blank before this line

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +281,32 @@ function _M.rewrite(conf, ctx)
...
+                local uri, _, err = re_sub(upstream_uri,
+                    conf.regex_uri[i], conf.regex_uri[i + 1], "jo")
+                if uri then
+                    upstream_uri = uri
+                else
+                    error_msg = "failed to substitute the uri " .. ngx.var.uri ..
+                        " (" .. conf.regex_uri[i] .. ") with " ..
+                        conf.regex_uri[i + 1] .. " : " .. err
+                end
+                break
```
monkeyDluffy6017, line 303: Need a blank before this line

apisix/plugins/proxy-rewrite.lua
```diff
@@ -281,25 +281,32 @@ function _M.rewrite(conf, ctx)
...
+        for i = 1, #conf.regex_uri, 2 do
+            local captures, err = re_match(upstream_uri, conf.regex_uri[i], "jo")
+            if err then
+                error_msg = "failed to match the uri " .. ctx.var.uri ..
+                    " (" .. conf.regex_uri[i] .. ") " .. err
+                break
+            end
+            if captures then
+                ctx.proxy_rewrite_regex_uri_captures = captures
+                local uri, _, err = re_sub(upstream_uri,
```
monkeyDluffy6017, line 294: Need a blank before this line

apisix/plugins/proxy-rewrite.lua
monkeyDluffy6017: Need a blank before this line

apisix/plugins/proxy-rewrite.lua
```diff
@@ -330,7 +337,7 @@ function _M.rewrite(conf, ctx)
 
     if conf.headers then
         local hdr_op, err = core.lrucache.plugin_ctx(lrucache, ctx, nil,
-                                    create_header_operation, conf.headers)
+            create_header_operation, conf.headers)
```
monkeyDluffy6017, line 340: Why do you modify this?

apisix/plugins/proxy-rewrite.lua
```diff
@@ -339,7 +346,7 @@ function _M.rewrite(conf, ctx)
         local field_cnt = #hdr_op.add
         for i = 1, field_cnt, 2 do
             local val = core.utils.resolve_var_with_captures(hdr_op.add[i + 1],
-                                            ctx.proxy_rewrite_regex_uri_captures)
+                ctx.proxy_rewrite_regex_uri_captures)
```
monkeyDluffy6017, line 349: ditto

apisix/plugins/proxy-rewrite.lua
```diff
@@ -348,7 +355,7 @@ function _M.rewrite(conf, ctx)
         local field_cnt = #hdr_op.set
         for i = 1, field_cnt, 2 do
             local val = core.utils.resolve_var_with_captures(hdr_op.set[i + 1],
-                                            ctx.proxy_rewrite_regex_uri_captures)
+                ctx.proxy_rewrite_regex_uri_captures)
```
monkeyDluffy6017, line 358: ditto

docs/zh/latest/plugins/proxy-rewrite.md
```diff
@@ -38,7 +38,7 @@ description: 本文介绍了关于 Apache APISIX `proxy-rewrite` 插件的基本
...
-| regex_uri | array[string] | 否    |         |                                                                                                                                        | 转发到上游的新 `uri` 地址。使用正则表达式匹配来自客户端的 `uri`，如果匹配成功，则使用模板替换转发到上游的 `uri`，如果没有匹配成功，则将客户端请求的 `uri` 转发至上游。当同时配置 `uri` 和 `regex_uri` 属性时，优先使用 `uri`。例如：["^/iresty/(.*)/(.*)/(.*)","/$1-$2-$3"] 第一个元素代表匹配来自客户端请求的 `uri` 正则表达式，第二个元素代表匹配成功后转发到上游的 `uri` 模板。但是目前 APISIX 仅支持一个 `regex_uri`，所以 `regex_uri` 数组的长度是 `2`。 |
+| regex_uri | array[string] | 否    |         |                                                                                                                                        | 多个正则表达式与替换模式对。使用正则表达式匹配来自客户端的 `uri`，如果匹配成功，则使用模板替换转发到上游的 `uri`，如果没有匹配成功则将客户端请求的 `uri` 转发至上游。当同时配置 `uri` 和 `regex_uri` 属性时，优先使用 `uri`。当前支持多个正则表达式模式匹配，`apisix`将逐个尝试匹配直到有一个成功匹配和替换。例如：`["^/iresty/(.*)/(.*)/(.*)", "/$1-$2-$3", ^/theothers/(.*)/(.*)", "/theothers/$1-$2"]` 奇数索引的元素代表匹配来自客户端请求的 `uri` 正则表达式，偶数索引的元素代表匹配成功后转发到上游的 `uri` 模板。请注意该值的长度必须为**偶数值**。 |
```
monkeyDluffy6017, line 41: 如果没有匹配成功，则将客户端请求的

t/plugin/proxy-rewrite3.t
monkeyDluffy6017: I think he's following the other test cases in this test case file

docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -38,7 +38,7 @@ The `proxy-rewrite` Plugin rewrites Upstream proxy information such as `scheme`,
...
+| regex_uri                   | array[string] | False    |         |                                                                                                                                        | Multiple regular expression and substitute patterns. Regular expressions can be used to match the URL from client. If it matches, the URL template is forwarded to the upstream. Otherwise, the URL from the client is forwarded. When both `uri` and `regex_uri` are configured, `uri` has a higher priority. It currently supports multiple regular expressions for matching and substitution, apisix will iterate all patterns until a successful match and substitution. For example, `["^/iresty/(.*)/(.*)/(.*)", "/$1-$2-$3", ^/theothers/(.*)/(.*)", "/theothers/$1-$2"]`. Here, the odd index elements are the regular expressions to match, the even index elements are the URL templates forwarded to the Upstream. Length of this property must be a even number. |
```
monkeyDluffy6017, line 41: | regex_uri                   | array[string] | False    |         |                                                                                                                                        | Regular expressions can be used to match the URL from client. If it matches, the URL template is forwarded to the upstream. Otherwise, the URL from the client is forwarded. When both `uri` and `regex_uri` are configured, `uri` has a higher priority. Multiple regular expressions are currently supported for pattern matching, and the plugin will try to match them one by one until they succeed or all fail. For example: `["^/iresty/(. *)/(. *)/(. *)", "/$1-$2-$3", ^/theothers/(. *)/(. *)", "/theothers/$1-$2"]`, the element with the odd index represents the `uri` regular expression that matches the request from the client, and the element with the even index represents the `uri` template that is forwarded upstream upon a successful match. Please note that the length of this value must be an **even number** |

monkeyDluffy6017:
@fengxsong please fix the ci

monkeyDluffy6017 (dismissed):

kingluo (changes requested):
rebase to the latest master, because proxy-rewrite.lua has merged new changes.

monkeyDluffy6017 (approved):

soulbird (approved):

kingluo (approved):

PR #9247, opened 2023-04-06
feat: support store route's cert in secrets manager

soulbird:
### Description

As part of #8319

Support for storing certificates on routers through Secret Manager.
When using this function to configure the certificate, we only verify the validity of the certificate when the certificate is used on the DP side. This is done to avoid the problem that the CP side may not be able to access the backend Secret Manager Service.

### Checklist


kingluo (dismissed):

apisix/schema_def.lua
```diff
@@ -725,8 +725,19 @@ _M.ssl = {
             default = "server",
             enum = {"server", "client"}
         },
-        cert = certificate_scheme,
-        key = private_key_schema,
+        cert = {
+            oneOf = {
+                { type = "string", minLength = 128, maxLength = 64*1024},
+                -- TODO: uniformly define the schema of secret_uri
+                { type = "string", pattern = "^\\$secret://"}
```
monkeyDluffy6017, line 732: Does this could work?
```
        cert = {
            oneOf = {
                certificate_scheme,
                { type = "string", pattern = "^\\$secret://"}

```

apisix/ssl.lua
```diff
@@ -252,9 +253,14 @@ function _M.check_ssl_conf(in_dp, conf)
         end
     end
 
-    local ok, err = validate(conf.cert, conf.key)
-    if not ok then
-        return nil, err
+    -- if the certificate uses a secret reference, we only verify it when using it
+    if not secret.check_secret_uri(conf.cert) and
+        not secret.check_secret_uri(conf.key) then
+
```
monkeyDluffy6017, line 259: what about only one use a secret? should we check this?

t/router/radixtree-sni2.t
```diff
@@ -543,3 +543,85 @@ qr/(fetch|release) table \w+/
...
+                )
+            if code >= 300 then
+                ngx.status = code
+                return ngx.say(body)
+            end
+            -- set ssl
+            local code, body = t('/apisix/admin/ssls/1',
+                ngx.HTTP_PUT,
+                [[{
+                     "cert": "$secret://vault/test1/ssl/test2.com.crt",
```
monkeyDluffy6017, line 581: extra blank

docs/en/latest/admin-api.md
```diff
@@ -1154,8 +1154,8 @@ SSL resource request address: /apisix/admin/ssls/{id}
...
-| cert         | True     | Certificate              | HTTPS certificate.                                                                                             |                                                  |
-| key          | True     | Private key              | HTTPS private key.                                                                                             |                                                  |
+| cert         | True     | Certificate              | HTTPS certificate. This field supports saving the value in Secret Manager using the [APISIX Secret](../terminology/secret.md) resource.                                                                                             |                                                  |
+| key          | True     | Private key              | HTTPS private key. This field supports saving the value in Secret Manager using the [APISIX Secret](../terminology/secret.md) resource.                                                                                             |                                                  |
```
monkeyDluffy6017, line 1158: Does it support `environment variables`, i see you only check the prefix `$secret://`?

apisix/ssl.lua
soulbird: Let me adjust it. If any of the certificate and key use secret, no verification will be done.

t/router/radixtree-sni2.t
```diff
@@ -543,3 +543,85 @@ qr/(fetch|release) table \w+/
...
+
+            ngx.say("passed")
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 17: get cert from vault
```
monkeyDluffy6017, line 599: could we add test cases about using `environment variable`?

soulbird: done

apisix/ssl.lua
```diff
@@ -252,9 +253,14 @@ function _M.check_ssl_conf(in_dp, conf)
...
     end
 
-    local ok, err = validate(conf.cert, conf.key)
-    if not ok then
-        return nil, err
+    -- if the certificate or key uses a secret reference, we only verify it when using it
+    if not secret.check_secret_uri(conf.cert) or
+        not secret.check_secret_uri(conf.key) then
+
+        local ok, err = validate(conf.cert, conf.key)
```
monkeyDluffy6017, line 260: If the cert is using secret_uri and the key doesn't, the validate will failed.
We have two choices:
1. the key and cert should support secret_uri both
2. validate cert and key seperately

Could we add test cases about this?

apisix/ssl.lua
```diff
@@ -252,9 +253,14 @@ function _M.check_ssl_conf(in_dp, conf)
         end
     end
 
-    local ok, err = validate(conf.cert, conf.key)
-    if not ok then
-        return nil, err
+    -- if the certificate or key uses a secret reference, we only verify it when using it
```
monkeyDluffy6017, line 256: where do we validate the cert and key if using secret reference?

monkeyDluffy6017: OK, i know

apisix/ssl/router/radixtree_sni.lua
```diff
@@ -120,16 +121,17 @@ end
 
 -- export the set cert/key process so we can hook it in the other plugins
 function _M.set_cert_and_key(sni, value)
```
monkeyDluffy6017, line 123: the gm will replace the `set_cert_and_key` function, please check the influence

apisix/ssl.lua
soulbird: The check_ssl_conf function is called in two places:
1. cp plane write configuration
2. The dp plane reads the configuration from the configuration center
When a cert or key is referenced by a secret, it is not verified for the following reasons:
1. Do not do too many dynamic checks on the cp plane, which can simplify the network topology
2. The dp plane will verify when using the cert and key
3. 
Therefore, it is only verified when neither the cert nor the key is a secret reference

apisix/ssl/router/radixtree_sni.lua
soulbird: Adjustments have been made, and now it is compatible with gm ssl

apisix/ssl.lua
monkeyDluffy6017: OK, when the key or cert is  a secret reference, validate them later

apisix/ssl.lua
```diff
@@ -252,9 +253,14 @@ function _M.check_ssl_conf(in_dp, conf)
         end
     end
 
-    local ok, err = validate(conf.cert, conf.key)
-    if not ok then
-        return nil, err
+    -- when cert or key is referenced by secret, avoid dynamic verification of it
```
monkeyDluffy6017, line 256: The comment is difficult to understand

apisix/ssl/router/radixtree_sni.lua
monkeyDluffy6017: I don't see the code, No changes required?

monkeyDluffy6017:
please make ci pass

apisix/ssl.lua
monkeyDluffy6017, line 256: can we cache key and cert here with 
```
local new_ssl_value = secret.fetch_secrets(matched_ssl.value) or matched_ssl.value
```
And validate the `new_ssl_value `

monkeyDluffy6017: check like this: 
```
local ok, err = validate(new_ssl_value.cert, new_ssl_value.key )
```

soulbird: new_ssl_value as a whole is cacheable

soulbird: At present, only the MVP version is implemented, and a perfect caching mechanism will be considered in the future to reduce the time-consuming of a single request

monkeyDluffy6017 (approved):

kingluo (approved):

PR #9259, opened 2023-04-07
avoid creating the whole radixtree while changing one route

ranxuxin001:
### Description

This patch improves the performance of apisix 2.9 (also the newest version) when there're more than 10 thousand routes and modifying one route. It only modifies the route data in the routes array or the relative node of radixtree instead of creating the whole radixtree while one route was changed.

After pressure test as users with 1000 concurrent requests using ab tool while changing a route every second, the overhead of cpu time is only 7%-8%. On the other hand, the apisix 2.9 costs near 100% cpu time.

This patch (commit ea64d7d998c8e8f2ee9f4a70d989cd78d9ec1fca) is based on the commit c924eb1cded079ec0d6279b211466b7d84a91ec2 in master branch of apisix.

In my centos, the apisix is installed using apisix-2.9-0.el7.x86_64.rpm. And the patch is made based on it.

Fixes #9140

PR #9265, opened 2023-04-07
feat: incremental radixtree udpate

ranxuxin:
refer to #9259
When update one route every second and more than 10 thousand routes exist, the result of pressure test using the patched apisix indicates below:

cpu usage 7-8%   workload 0.5-1:
image

wrk pressure test result:
image

The result of test pressure using original apisix 2.9 indicates below:

cpu usage 80%-100% :
image

wrk pressure test result:
image


ranxuxin:
I tested apisix-2.9 and the improved version again. 
machine:   8core  2GHz    32G memory,    apisix worker auto
wrk -c1000 -d60s -t8 --latency --timeout=15s  http://www.aaabbb1122.com:9080/123
update one route every 0.5 second. And more than 10 thousand routes exist in apisix.

- apisix 2.9 improved version:

image


the elapsed time of dealing with updating one route:

image



- apisix 2.9 original version:

image




the elapsed time of dealing with updating one route:

image



conclusion:
From the test between apisix 2.9 improved version and original version, the improved version has much better performance during dealing with update route requests. The reason is that the improved version only find and update the updated route in radixtree and lua hash table. By comparison, the original version creates the whole radixtree when updating one route. The influence is amplified when large amount of routes exist in apisix.

ranxuxin:
I tested apisix-2.9 and the improved version again. 
machine:   8core  2GHz    32G memory,    apisix worker auto
wrk -c1000 -d60s -t8 --latency --timeout=15s  http://www.aaabbb1122.com:9080/123
update one route every 0.5 second. And more than 10 thousand routes exist in apisix.

- apisix 2.9 improved version:

image


the elapsed time of dealing with updating one route:

image



- apisix 2.9 original version:

image




the elapsed time of dealing with updating one route:

image



conclusion:
From the test between apisix 2.9 improved version and original version, the improved version has much better performance during dealing with update route requests. The reason is that the improved version only find and update the updated route in radixtree and lua hash table. By comparison, the original version creates the whole radixtree when updating one route. The influence is amplified when large amount of routes exist in apisix.

PR #9291, opened 2023-04-12
feat: support consuming specified cost in the limit count function

shreemaan-abhishek:
### Description

The limit count plugin is based on deducting a constant value ( = 1, currently) from the given `count`. Using a variable and setting a value to it instead of replicating the literal value ( = 1, currently) is more maintainable and extendable.

### Checklist


soulbird:
Can you add some test cases？

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -112,5 +113,5 @@ function _M.incoming(self, key)
     return 0, remaining, ttl
 end
 
-
+_M.script = script
```
soulbird, line 116: Why export this?

shreemaan-abhishek: The redis script in `limit-count-redis.lua` and `limit-count-redis-cluster.lua` are the same so I exported it to avoid duplication.

soulbird, line 116: If you want to export this, it's better to put it in init.lua, don't refer to each other between parallel files. But it doesn't seem appropriate to put it in init.lua, so let's keep it as it is

t/plugin/limit-count4.t
```diff
@@ -171,3 +171,57 @@ passed
...
+
+=== TEST 5: modified limit-count.incoming, cost == 2
+--- config
+    location = /t {
+        content_by_lua_block {
+            local limit_count_local = require "apisix.plugins.limit-count.limit-count-local"
+            local lim = limit_count_local.new("plugin-limit-count", 10, 60)
+            local uri = ngx.var.uri
+            for i = 1, 7 do
+                local delay, err = lim.limit_count:handle_incoming(uri, 2, true)
```
soulbird, line 185: Please add test cases to test the incoming function separately, limit_count.handle_incoming is not a module function of apisix.

shreemaan-abhishek: In fact, the `handle_incoming` is a part of APISIX. I extended the existing function from lua-resty-limit-traffic.

https://github.com/apache/apisix/blob/f294304b61f104db6ef3cff9cc26e9a36d125824/apisix/plugins/limit-count/limit-count-local.lua#L19-L43

soulbird: I know, we hook it in a very hacky way. But essentially it doesn't belong to apisix, if you only test it, some logic in the incoming function will be missed

shreemaan-abhishek: Addressed the issue, please see if the tests are up to the mark.

apisix/plugins/limit-count/limit-count-redis-cluster.lua
soulbird: I mean don't export the script, keep it as it is

shreemaan-abhishek: Ah sorry let me fix it.

shreemaan-abhishek: FYI, we can't test the script without exporting it, so I removed the tests as well.

soulbird (approved):

apisix/plugins/limit-count/limit-count-local.lua
```diff
@@ -14,7 +14,34 @@
...
+local limit_count = require("resty.limit.count")
+
+limit_count.handle_incoming = function (self, key, cost, commit)
+  if cost < 1 then
+    return nil, "cost must be at least 1"
+  end
+  local dict = self.dict
+  local limit = self.limit
+  local window = self.window
+
```
monkeyDluffy6017, line 26: I think we'd better not hack the underlying functions and suggest directly fork the corresponding repositories to make changes

shreemaan-abhishek: Nevertheless, we HAVE to maintain a separate codebase. And in fact, we are just adding a new function which is essentially a modified version of the existing `incoming` function.

shreemaan-abhishek (commented):
LGTM 👌🏼

leslie-tsang (approved):

apisix/plugins/limit-count/init.lua
```diff
@@ -20,7 +20,6 @@ local tab_insert = table.insert
 local ipairs = ipairs
 local pairs = pairs
 
-local plugin_name = "limit-count"
```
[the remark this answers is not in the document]
soulbird: The code here is a public capability of apisix, not a plugin. The plugin name should be defined by the plugin using the capability

PR #9303, opened 2023-04-13
docs: fix admonition in Terminology, Secret

kayx23:
## Before
image
https://apisix.apache.org/docs/apisix/terminology/secret/

## After
See changed file.

### DIFF
--- a/docs/en/latest/terminology/secret.md
+++ b/docs/en/latest/terminology/secret.md
@@ -42,9 +42,9 @@ APISIX currently supports storing secrets in the following ways:
 
 You can use APISIX Secret functions by specifying format variables in the consumer configuration of the following plugins, such as `key-auth`.
 
-::: note
+:::note
 
-If a configuration item is: `key: "$ENV://ABC"`, when the actual value corresponding to $ENV://ABC is not retrieved in APISIX Secret, the value of the key will be "$ENV://ABC" instead of `nil`.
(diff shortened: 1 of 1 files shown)


juzhiyuan (approved):

kayx23:
Thanks. Merge?
> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

kayx23:
Anybody know why the change is not reflected on the website? I'm still seeing the OG version which is not properly formatted: 
image

https://apisix.apache.org/docs/apisix/terminology/secret/

Confused. Does this mean if I spot a problem on version 3.2, I should commit my change to BOTH version 3.2 and master branch?

Neilblaze:
@kayx23 I hope so, since the preview is being generated from v3.2 & not from master branch.

kayx23:
That's actually quite annoying innit?! Have to open two PRs for the same change? I wonder how many people have committed doc improvements but to the wrong branch....

I was told I shouldn't commit to a specific version, say 3.2, because the change won't be carried over unless I commit to master......

Neilblaze:
I think it's better to push the content to v3.2 & run a cron to migrate files (docs) to the master asynchronously, but yeah I also feel the same.

kayx23:
Maybe we should open a discussion

PR #9309, opened 2023-04-14
fix: proxy-rewrite should set ngx.var.uri

kingluo:
### Description

Besides custom `upstream_uri` var, it should set `ngx.var.uri` to keep compatible.

### Checklist

### DIFF
--- a/apisix/plugins/proxy-rewrite.lua
+++ b/apisix/plugins/proxy-rewrite.lua
@@ -22,6 +22,7 @@ local ngx         = ngx
 local type        = type
 local re_sub      = ngx.re.sub
 local re_match    = ngx.re.match
+local req_set_uri = ngx.req.set_uri
 local sub_str     = string.sub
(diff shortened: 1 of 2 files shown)


apisix/plugins/proxy-rewrite.lua
```diff
@@ -317,6 +318,8 @@ function _M.rewrite(conf, ctx)
             upstream_uri = core.utils.uri_safe_encode(upstream_uri)
         end
 
+        req_set_uri(upstream_uri)
```
kingluo, line 321: Note that we don't need to set uri when `use_real_request_uri_unsafe` is true, because in that case, the `upstream_uri` is supposed to take precedence.

leslie-tsang (approved):

Neilblaze (approved):

monkeyDluffy6017 (approved):

PR #9365, opened 2023-04-24
feat(cors): support for the Timing-Allow-Origin header

### LINKED ISSUE #9307: feat: support for the Timing-Allow-Origin header in CORS plugin
### Description

When using the CORS plugin, I'd like to be able to set also the Timing-Allow-Origin header (https://www.w3.org/TR/resource-timing/#sec-timing-allow-origin).

The plugin configuration for this header should be separate from the one dedicated to the Access-Control-Allow-Origin header (so I should be able to set 2 different allowed values for these headers).
I'm not completely sure about the need of this separation in the real world, but a possible use case could be: 
- I set Access-Control-Allow-Origin: *
- I restrict the domains that can see the timings setting Timing-Allow-Origin: https://my-specific-domain.com

The plugin configuration should have 2 fields to configure the Timing-Allow-Origin header, similarly to Access-Control-Allow-Origin: a field for accepting a single value "timing_allow_origin", and a field "timing_allow_origin_by_regex" to match different possible origins.

I think that the new header should be set in response by the plugin only if the user choose to configure it in the CORS plugin configuration, also given the fact that setting the value * (wildcard) may be considered insecure (see discussion https://github.com/w3c/resource-timing/issues/222) and so we can not set it as default.

If you agree on the issue, I would like to try implementing this feature

skimdz86:
### Description

Implements feature described in issue https://github.com/apache/apisix/issues/9307: added support  to the Timing-Allow-Origin header in the CORS plugin.

The new header can be configured with a list of domains or a regex, just as the Access-Control-Allow-Origin.
The new header is returned only if the Access-Control-Allow-Origin is returned.

I added a new parameter to the 2 functions process_with_allow_origins and process_with_allow_origins_by_regex to make these functions "common" and now used both for calculating the Access-Control-Allow-Origin and the Timing-Allow-Origin header; the new "allow_origin_type" parameter was essential to differentiate the cache used to save the elaboration results for the 2 different headers.
 
**Note**: I updated also the markdown documentation (english version) adding the 2 new configuration fields, but unfortunately I don't speak Chinese, so any help for updating also the /docs/zh/latest/plugins/cors.md file will be appreciated.

Fixes #9307  

### Checklist


monkeyDluffy6017:
Hi @skimdz86, thanks for your contribution, Is Timing-Allow-Origin some kind of CORS?

apisix/plugins/cors.lua
```diff
@@ -179,6 +204,23 @@ function _M.check_schema(conf, schema_type)
         end
     end
 
+    if not conf.allow_origins then
+        if conf.timing_allow_origins or conf.timing_allow_origins_by_regex then
+            return false, "you can not set 'timing_allow_origin' " ..
+                          "or 'timing_allow_origin_by_regex' " ..
+                          "when 'allow_origins' is not set"
+        end
+    end
```
monkeyDluffy6017, line 213: if the `allow_origins_by_regex ` has been set , is there a problem here?

apisix/plugins/cors.lua
```diff
@@ -202,9 +244,12 @@ local function set_cors_headers(conf, ctx)
     if conf.allow_credential then
         core.response.set_header("Access-Control-Allow-Credentials", true)
     end
+    if ctx.timing_allow_origin then
```
monkeyDluffy6017, line 247: Is a blank line needed here? as what is to be set is relevant

t/plugin/cors4.t
```diff
@@ -0,0 +1,623 @@
...
+GET /hello HTTP/1.1
+--- more_headers
+Origin: http://testurl.domain.com
+--- response_headers
+Access-Control-Allow-Origin: http://testurl.domain.com
+Access-Control-Allow-Methods: GET,POST
+Access-Control-Allow-Headers: request-h
+Access-Control-Expose-Headers: expose-h
+Access-Control-Max-Age: 10
+Timing-Allow-Origin: http://testurl.domain.com
```
monkeyDluffy6017, line 623: Need a blank at the end of file

apisix/plugins/cors.lua
skimdz86: Actually, I didn't consider the case in which the CORS are managed already by the upstream server; so I agree, this way we can add only the timing  header in the api gateway in addition to the existing CORS headers, leaving freedom to the users in the configuration. So I'll remove this check, but I also have to change the logic in the header_filter function, to allow setting the timing even when allow_origins is not set in the plugin conf.

apisix/plugins/cors.lua
monkeyDluffy6017: Yes, it's my personal thought

skimdz86:
@monkeyDluffy6017  Updated the PR with your suggestions; since we decoupled the management of cors headers and timing header, I also split the set_cors_headers function in 2, adding the set_timing_headers function.

skimdz86:
Hi @monkeyDluffy6017 , any news on this PR?

Thanks

monkeyDluffy6017:
@skimdz86  sorry for this delay, i have no idea if Timing-Allow-Origin belongs to CORS

skimdz86:
> @skimdz86 sorry for this delay, i have no idea if Timing-Allow-Origin belongs to CORS

Ok, so what are the next steps in this case?

t/plugin/cors4.t
```diff
@@ -0,0 +1,623 @@
...
+Access-Control-Allow-Origin: *
+Access-Control-Allow-Methods: GET,POST
+Access-Control-Allow-Headers: request-h
+Access-Control-Expose-Headers: expose-h
+Access-Control-Max-Age: 10
+Access-Control-Allow-Credentials:
+Timing-Allow-Origin:
+
+
+=== TEST 5: set route ( allow_origins same as timing_allow_origins )
```
monkeyDluffy6017, line 166: Three blanks are needed, please check the other place

apisix/plugins/cors.lua
```diff
@@ -292,22 +346,44 @@ function _M.header_filter(conf, ctx)
...
         allow_origins = process_with_allow_origins_by_metadata(
-                conf.allow_origins_by_metadata, ctx, req_origin
+            TYPE_ACCESS_CONTROL_ALLOW_ORIGIN, conf.allow_origins_by_metadata, ctx, req_origin
         )
     end
     if conf.allow_origins ~= "*" then
         core.response.add_header("Vary", "Origin")
     end
-    if allow_origins then
+    if allow_origins then
```
monkeyDluffy6017, line 366: remove redundant indents

apisix/plugins/cors.lua
```diff
@@ -202,9 +236,20 @@ local function set_cors_headers(conf, ctx)
     if conf.allow_credential then
         core.response.set_header("Access-Control-Allow-Credentials", true)
     end
+
+    if ctx.timing_allow_origin then
+        core.response.set_header("Timing-Allow-Origin", ctx.timing_allow_origin)
```
monkeyDluffy6017, line 241: Why do you need this?

monkeyDluffy6017:
I find that these are 2 completely separate functions, and if put in a plugin, the complexity of the configuration items can be very confusing, for example the configuration item allow_methods only works for cors, so would it be possible to consider extracting the common functions and redeveloping a plugin

docs/en/latest/plugins/cors.md
```diff
@@ -42,6 +42,8 @@ The `cors` Plugins lets you enable [CORS](https://developer.mozilla.org/en-US/do
...
 | allow_origins_by_metadata | array   | False    | nil     | Origins to enable CORS referenced from `allow_origins` set in the Plugin metadata. For example, if `"allow_origins": {"EXAMPLE": "https://example.com"}` is set in the Plugin metadata, then `["EXAMPLE"]` can be used to allow CORS on the origin `https://example.com`.                                                                                                                          |
+| timing_allow_origins             | string  | False    | nil     | Origin to allow to access the resource timing information. See [Timing-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Timing-Allow-Origin). Use the `scheme://host:port` format. For example, `https://somedomain.com:8081`. If you have multiple origins, use a `,` to list them. |
+| timing_allow_origins_by_regex    | array   | False    | nil     | Regex to match with origin for enabling access to the resource timing information. For example, `[".*\.test.com"]` can match all subdomain of `test.com`. When set to specified range, only domains in this range will be allowed, no matter what `timing_allow_origins` is. |
```
monkeyDluffy6017, line 46: I find that CORS and Timing-Allow-Origin are 2 completely separate functions, and if we put their configurations in one table, the complexity of the configuration items can be very confusing, for example the configuration item `allow_methods` only works for CORS, so how about split the table into two part, one for cors and the other for Timing-Allow-Origin

monkeyDluffy6017:
I will approve this pr after you resolve the review comments

apisix/plugins/cors.lua
skimdz86: ok, useless, I may have forgotten to remove it when I decoupled the logic of classic cors and timing headers. I'll remove it

docs/en/latest/plugins/cors.md
skimdz86: ok seems a good idea, I'll fix it

skimdz86:
> I will approve this pr after you resolve the review comments

@monkeyDluffy6017 PR updated

monkeyDluffy6017 (approved):

shreemaan-abhishek (dismissed):

shreemaan-abhishek (approved):

monkeyDluffy6017 (approved):

Issue #9223, opened 2023-04-03
bug: response-rewrite : The format validation for headers.add is unreasonable.


840963657:
### Current Behavior

response-rewrite plugin Configuration Information:
"plugins": {
        "response-rewrite": {
            "headers": {
                "add": [
                    "X-Server-test:a"
                ]
            }
        }
    }

error Information
{"error_msg":"failed to check the configuration of plugin response-rewrite err: property \"headers\" validation failed: object matches none of the required"}

### Expected Behavior

The parameter value should be able to contain a single character.

I think this regular expression is incorrect:
response-rewrite.lua:55
                 add = {
                            type = "array",
                            minItems = 1,
                            items = {
                                type = "string",
                                -- "Set-Cookie: <cookie-name>=<cookie-value>; Max-Age=<number>"
                                pattern = "^[^:]+:[^:]+[^/]$"
                            }

### Error Logs

{"error_msg":"failed to check the configuration of plugin response-rewrite err: property \"headers\" validation failed: object matches none of the required"}

### Steps to Reproduce

add response-rewrite plugin;
Configure the headers.add value to be a single character

### Environment

- APISIX version (run `apisix version`):3.2
- Operating system (run `uname -a`):
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

kayx23:
![image](https://user-images.githubusercontent.com/39619599/234445426-153692cd-f29f-40d4-92d4-95c1f4f11277.png)

PR #9372, opened 2023-04-26
fix: response-rewrite plugin can't add only one character

LiteSun:
### Description

fix response-rewrite plugin can't add only one character

Fixes https://github.com/apache/apisix/issues/9223
### Checklist

### DIFF
--- a/t/plugin/response-rewrite.t
+++ b/t/plugin/response-rewrite.t
@@ -696,3 +696,38 @@ passed
 --- request
 GET /hello
 --- response_body
+
+
+
+=== TEST 27: test add header with one word
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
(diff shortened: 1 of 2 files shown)


t/plugin/response-rewrite.t
```diff
@@ -696,3 +696,40 @@ passed
...
+                }]]
+                )
+
+            ngx.say(body)
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
```
soulbird, line 735: Do you need to verify whether the header is valid?

monkeyDluffy6017 (approved):

t/plugin/response-rewrite.t
monkeyDluffy6017: @LiteSun

LiteSun: @soulbird Hi, this PR only fixes the issue of not being able to configure a single character and adds test coverage, which header can be verified to have been covered by the previous tests, not what this PR needs to do. 🤔

soulbird (approved):

Issue #9351, opened 2023-04-21
bug: proxy-rewrite does not work with aws-lambda plugin


lpiob:
### Current Behavior

When configuring an ApisixRoute endpoint, the `proxy-rewrite` allows me to for example remove specific request headers.

This works fine, unless the request is routed to AWS Lambda function through `aws-lambda` plugin.

The destination function get triggered successfully, yet the headers are not removed.

### Expected Behavior

`proxy-rewrite` should work the same regardless of the final endpoint

### Error Logs

Some header manipulation code IS running, as can be seen from apisix logs printed while processing request:
```
apisix-5b49f6b74-wvl4r apisix 2023/04/21 09:08:49 [warn] 48#48: *9023948 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.200.160.150, server: _, request: "POST /lbiegaj/test HTTP/1.1", host: "api.example.com"
apisix-5b49f6b74-wvl4r apisix 2023/04/21 09:08:49 [warn] 48#48: *9023948 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.200.160.150, server: _, request: "POST /lbiegaj/test HTTP/1.1", host: "api.example.com"
apisix-5b49f6b74-wvl4r apisix 2023/04/21 09:08:49 [warn] 48#48: *9023948 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.200.160.150, server: _, request: "POST /lbiegaj/test HTTP/1.1", host: "api.example.com"
apisix-5b49f6b74-wvl4r apisix 2023/04/21 09:08:49 [warn] 48#48: *9023948 [lua] request.lua:120: set_header(): DEPRECATED: use set_header(ctx, header_name, header_value) instead, client: 10.200.160.150, server: _, request: "POST /lbiegaj/test HTTP/1.1", host: "api.example.com"
```

### Steps to Reproduce

1. Create an AWS Lambda with function url. Sample code that reflects the received event:
```export const handler = async(event) => {
    const response = {
        statusCode: 200,
        body: JSON.stringify(event),
    };
    return response;
};
```
2. Configure an ApisixRoute that uses the `proxy-rewrite` and `aws-lambda` plugins.
```
spec:
  http:
    backends:
    - serviceName: lambda
      servicePort: 80
    match:
      methods:
      - POST
[...]

PR #9389, opened 2023-04-28
ci: new test for lambda and proxy-rewrite work together

Sn0rt:
### Description

Fixes https://github.com/apache/apisix/issues/9351

### Checklist


soulbird:
Looks like we need documentation more than test cases

Sn0rt:
> Looks like we need documentation more than test cases

now. this PR is not needed anymore. 
I will implement a new feature that supports users removing the UA by proxy-rewrite. this PR need to keep pending status.

Issue #9391, opened 2023-04-28
feat: As a user, I want the test of tcp-logger is a real case, so that 


Sn0rt:
### Description

Now the ci of tcp-logger is mock, I want this to be a real case.

more detail at `t/plugin/tcp-logger.t`

PR #9392, opened 2023-04-28
ci: tcp logger with real logger server

Sn0rt:
### Description

#9391

the [vector](https://vector.dev/docs/reference/configuration/sources/) is a popular logger service implemented  by  `Rust`

to add the vector as the logger server for receiving the log from the user client.

### Checklist


bzp2010 (commented):
For the content, I see no problem. But I have a question: is there any reason why you chose `vector`?

Sn0rt:
> For the content, I see no problem. But I have a question: is there any reason why you chose `vector`?

Simple, easy to configure, well documented.

There is another reason why I tried to use Fluentd and encountered a strange case. I spent half a day trying to fix it to no avail. There is no problem with how to use `vector `.

bzp2010:
> Simple, easy to configure, well documented.
> 
> There is another reason why I tried to use Fluentd and encountered a strange case. I spent half a day trying to fix it to no avail.

OK, @Sn0rt 

For the first point, that's very good. And for the second point, have you tried fluent bit?

Sn0rt:
> > Simple, easy to configure, well documented.
> > There is another reason why I tried to use Fluentd and encountered a strange case. I spent half a day trying to fix it to no avail.
> 
> OK, @Sn0rt
> 
> For the first point, that's very good. And for the second point, have you tried fluent bit (it is the next generation of fluentd)?

I took a brief look at the document construction of fluent-bit, which is not as good as that of vector. My requirement is to cooperate with multiple reporting log formations. fluent-bit does not have an example available out of the box, but vector provides it.

In this Pr, I expect vector to become the basic component of a ci log plug-in test, and to cooperate with other log plug-ins later.

bzp2010:
@Sn0rt Understand and recognize your position.

But I'm not sure if it's the right strategy to keep using `vector` in the future because essentially the open source project is a vendor-neutral project, the plugin is called `tcp-logger` and customers may use any logging pipeline project that supports TCP, I'm not sure....

bzp2010 (dismissed):

shreemaan-abhishek (commented):
LGTM

Neilblaze (approved):

ci/pod/vector/vector_logs_ca.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
soulbird, line 1: We'd better move certs to `t/certs`

Sn0rt: updated

t/plugin/tcp-logger.t
```diff
@@ -482,3 +482,119 @@ hello world
...
+    }
+--- request
+GET /t
+--- wait: 0.5
+--- response_body
+opentracing
+
+
+
+=== TEST 15: check tcp log
```
monkeyDluffy6017, line 588: The test case 15 should be put before the test case 14

Sn0rt: good.

monkeyDluffy6017 (approved):

soulbird:
We need to adjust the existing test cases. Currently, the log is forwarded to port 5044, which is a simulated log service.

Sn0rt:
> We need to adjust the existing test cases. Currently, the log is forwarded to port 5044, which is a simulated log service.

copy that. I rewrite some of the tcp-logger test cases.

monkeyDluffy6017 (approved):

soulbird (approved):

PR #9396, opened 2023-04-28
feat: route name  recorded  in  access.log

tylitianrui:
### Description
In the log file, there is no flag to indicate which route configuration is working ,when multiple same paths with different priorities exist. It is necessary to record the route name.

### Checklist

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -195,7 +195,7 @@ nginx_config:                     # config for render the template to generate n
   http:
     enable_access_log: true         # enable access log or not, default true
     access_log: logs/access.log
-    access_log_format: "$remote_addr - $remote_user [$time_local] $http_host \"$request\" $status $body_bytes_sent $request_time \"$http_referer\" \"$http_user_agent\" $upstream_addr $upstream_status $upstream_response_time \"$upstream_scheme://$upstream_host$upstream_uri\""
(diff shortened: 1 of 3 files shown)


conf/config-default.yaml
```diff
@@ -195,7 +195,7 @@ nginx_config:                     # config for render the template to generate n
   http:
     enable_access_log: true         # enable access log or not, default true
     access_log: logs/access.log
-    access_log_format: "$remote_addr - $remote_user [$time_local] $http_host \"$request\" $status $body_bytes_sent $request_time \"$http_referer\" \"$http_user_agent\" $upstream_addr $upstream_status $upstream_response_time \"$upstream_scheme://$upstream_host$upstream_uri\""
+    access_log_format: "$remote_addr - $remote_user [$time_local] $http_host \"$request\" $status $body_bytes_sent $route_name $request_time \"$http_referer\" \"$http_user_agent\" $upstream_addr $upstream_status $upstream_response_time \"$upstream_scheme://$upstream_host$upstream_uri\""
     access_log_format_escape: default       # allows setting json or default characters escaping in variables
```
soulbird, line 199: Modifying the default log format will introduce a new break change, and not everyone needs to log route_name, as far as I know, most people use route_id

So I think you can add this variable and then use it in your own config.yaml

monkeyDluffy6017: The `$route_name` is not recommended, because it may be duplicated.
We don't need to consider the apisix-dashboard project, this project is intended for beginners only and is not recommended for online environments.

monkeyDluffy6017:
I will close this PR, please feel free to reopen it

PR #9420, opened 2023-05-06
fix(etcd): reuse cli and enable keepalive

kingluo:
### Description

1. The JWT auth token is stored in the etcd cli table, so reuse the table between calls to avoid repeat auth calls.
2. The path between conf server and etcd should enable keepalive so that we could reuse the same connection to do etcd calls. Besides avoiding connection overhead, the etcd v3 uses [connection based auth](https://etcd.io/docs/v3.5/learning/design-auth-v3/).

### Checklist

### DIFF
--- a/t/core/config_etcd.t
+++ b/t/core/config_etcd.t
@@ -233,44 +233,7 @@ passed
 
 
 
-=== TEST 7: ensure only one auth request per subsystem for all the etcd sync
---- yaml_config
-apisix:
-  node_listen: 1984
-deployment:
-  role: traditional
-  role_traditional:
-    config_provider: etcd
-  etcd:
-    host:
-      - "http://127.0.0.1:1980" -- fake server port
-    timeout: 1
-    user: root                    # root username for etcd
-    password: 5tHkHhYkjr6cQY      # root password for etcd
---- extra_init_by_lua
-local health_check = require("resty.etcd.health_check")
-health_check.get_target_status = function()
-    return true
-end
---- config
-    location /t {
-        content_by_lua_block {
-            ngx.sleep(0.5)
-        }
-    }
---- request
-GET /t
---- grep_error_log eval
-qr/etcd auth failed/
---- grep_error_log_out
-etcd auth failed
-etcd auth failed
-etcd auth failed
-etcd auth failed
-
-
-
-=== TEST 8: ensure add prefix automatically for _M.getkey
+=== TEST 7: ensure add prefix automatically for _M.getkey
 --- config
     location /t {
         content_by_lua_block {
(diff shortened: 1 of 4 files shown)


t/core/config_etcd.t
```diff
@@ -233,44 +233,7 @@ passed
...
-        content_by_lua_block {
-            ngx.sleep(0.5)
-        }
-    }
---- request
-GET /t
---- grep_error_log eval
-qr/etcd auth failed/
---- grep_error_log_out
-etcd auth failed
```
kingluo, line 266: no duplicated auth anymore, due to cli reused.

t/core/etcd.t
```diff
@@ -401,35 +401,3 @@ qr/init_by_lua:\d+: \S+/
...
-  etcd:
-    host:
-      - "http://127.0.0.1:2379"
-    prefix: "/apisix"
---- config
-    location /t {
-        content_by_lua_block {
-            local etcd_lib = require("resty.etcd")
-            -- the mock won't take effect when using gRPC because the connection will be cached
-            etcd_lib.new = function()
```
kingluo, line 422: hijacking `resty.etcd.new` does not work after the cli is created once, because it's reused afterwards.

leslie-tsang (dismissed):

t/core/config_etcd.t
monkeyDluffy6017: Why remove this? how do you prove that the cli is reused?

kingluo: No need to prove reused or not in this test case.
If we need a meaningless test case, we could hook the get_etcd_cli to check if the cli is reused, but we could prove it in reality based on the fact that the performance is improved, which is hard to express via test::nginx. So we just do not write a new test case for this PR.

Back to this test case, it is arbitrary and does not describe what is "subsystem".
It's useless and confusing. If you could explain it, then I will keep it.

leslie-tsang: We need a test case to make sure the fix work as expected. :)
Plz add a new test case to prove it.

leslie-tsang: BTW, it‘s weird to delete a test case in a fix PR without adding any new test case.

kingluo: done

t/core/config_etcd.t
```diff
@@ -266,7 +266,6 @@ qr/etcd auth failed/
 etcd auth failed
 etcd auth failed
 etcd auth failed
-etcd auth failed
```
monkeyDluffy6017, line 269: Does this test case prove that the cli is reused?

kingluo: yes.

monkeyDluffy6017 (approved):

leslie-tsang (approved):

PR #9437, opened 2023-05-08
chore(ci): use real clickhouse server in ci

### LINKED ISSUE #9459: feat: CI should use real clickhouse server for testing.
### Description

Currently, CI is using a mock server to test the clickhouse-logger plugin. We should use a real one.

shreemaan-abhishek:
### Description

Fixes #9459 

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,6 @@ jobs:
           make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
           [[ ${{ steps.test_env.outputs.type }} != first ]] && sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after
           echo "Linux launch services, done."
-
       - name: Start Dubbo Backend
         if: matrix.os_name == 'linux_openresty' && steps.test_env.outputs.type == 'plugin'
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,6 +136,10 @@ jobs:
           make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
           [[ ${{ steps.test_env.outputs.type }} != first ]] && sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after
           echo "Linux launch services, done."
+      - name: Prepare clickhouse
+        run: |
+          echo "CREATE TABLE default.test (   \`host\` String,   \`client_ip\` String,   \`route_id\` String,   \`service_id\` String,   \`@timestamp\` String,    PRIMARY KEY(\`@timestamp\`) ) ENGINE = MergeTree()" | curl 'http://localhost:8123/'
+
       - name: Start Dubbo Backend
         if: matrix.os_name == 'linux_openresty' && steps.test_env.outputs.type == 'plugin'
         run: |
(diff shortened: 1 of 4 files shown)


shreemaan-abhishek:
The reviewers might request to use the real clickhouse server everywhere in the test suite and eliminate the mock server. We can do that but if we remove the mock server we need to create another clickhouse server to test the multiple endpoints feature.

Incrementally adding docker images/containers to ci will make it slower which is not good in the long run. WDYT? 🤔

.github/workflows/build.yml
```diff
@@ -136,6 +136,7 @@ jobs:
           make ci-env-up project_compose_ci=ci/pod/docker-compose.${{ steps.test_env.outputs.type }}.yml
           [[ ${{ steps.test_env.outputs.type }} != first ]] && sudo ./ci/init-${{ steps.test_env.outputs.type }}-test-service.sh after
           echo "Linux launch services, done."
+
```
Sn0rt, line 139: remove this newline (it's not needed

ci/init-plugin-test-service.sh
```diff
@@ -51,6 +51,8 @@ after() {
     # configure keycloak
     docker exec apisix_keycloak bash /tmp/kcadm_configure_cas.sh
     docker exec apisix_keycloak bash /tmp/kcadm_configure_university.sh
+
+    echo "CREATE TABLE default.test (\`host\` String, \`client_ip\` String, \`route_id\` String, \`service_id\` String, \`@timestamp\` String, PRIMARY KEY(\`@timestamp\`)) ENGINE = MergeTree()" | curl 'http://localhost:8123/'
```
Sn0rt, line 55: should we set the logger format? or is this the default struct?

ci/pod/docker-compose.plugin.yml
```diff
@@ -324,10 +324,21 @@ services:
     networks:
       vector_net:
 
+  clickhouse:
+    image: clickhouse/clickhouse-server:23.4.2-alpine
+    container_name: clickhouse
+    ports:
+      - '8123:8123'
+      - '9000:9000'
```
Sn0rt, line 332: why create 9000? I only find the port 8123 port has been used.

shreemaan-abhishek: Just in case we need those ports in the future, we will not have to redeclare it again. That's why I bound all the ports available.

ci/init-plugin-test-service.sh
shreemaan-abhishek: This was given in the documentation, of course we can use a different format as well.

ci/pod/docker-compose.plugin.yml
Sn0rt: one PR only for one thing I think.

Sn0rt (approved):

monkeyDluffy6017 (dismissed):

kingluo:
LGTM

soulbird:
It is better to change the old test cases and use the real clickhouse sever

leslie-tsang (approved):

Sn0rt:
LGTM

Sn0rt (dismissed):

Revolyssup:
@Sn0rt @shreemaan-abhishek Any reason why we are not using vector here?

shreemaan-abhishek:
@Revolyssup, we use vector for collecting logs via syslog. Since Clickhouse is a database, vector does not make sense here.

monkeyDluffy6017:
> It is better to change the old test cases and use the real clickhouse sever

Hi @shreemaan-abhishek, do you notice the review comment?

shreemaan-abhishek:
@monkeyDluffy6017 I added this commit, isn't it sufficient? https://github.com/apache/apisix/pull/9437/commits/ac3af865ec5cbad0d266d5839d3868b51725073d

soulbird (approved):

monkeyDluffy6017 (approved):

Issue #7467, opened 2022-07-15
feat:As a user, I don't think allowlist and denylist of  ua-restriction can be enabled at the same time


840963657:
### Description

I found other restriction plugin has a logic：Either one of whitelist or blacklist attribute must be specified。
But in ua-restriction, both allowlist and denylist can be used on their own.
If User-Agent is neither on the allowlist nor the denylist , the processing logic is the same as that of the allowlist .
So why do we still use the allowlist ？

tokers:
Fair enough. But if we make `allowlist` and `denylist` exclusive, then this is a broken change. We need to evaluate it.

tzssangglass:
What if allowlist and denylist are both on, ua-restriction only uses allowlist, and User-Agents that are not in allowlist are rejected?

spacewander:
Look like if the allowlist is matched, the denylist will be skipped?
The match rule is via regex, so it is possible for a UA to match both allowlist and denylist.

840963657:
Another problem: If I only configure allowlist , User-Agent is not on allowlist will also be released.

this code in apisix/plugins/ua-restriction.lua
`
local MATCH_NONE = 0
local MATCH_ALLOW = 1
local MATCH_DENY = 2
......
if match > MATCH_ALLOW then
        return 403, { message = conf.message }
    end
`

I looked at the code, If User-Agent on allowlist, the value of **match** becomes MATCH_ALLOW.   If User-Agent not on allowlist, the value of **match** is MATCH_NONE. 
So I think the allowlist is meaningless.

spacewander:
So we need an allowlist-only mode. When only allowlist is given, `MATCH_NONE` will be rejected.
@840963657 
Would you like to work on it?

monkeyDluffy6017:
I think the allowlist and denylist should be exclusive, it they exist at the same time, it is very difficult to understand

bzp2010:
I think it could go either way:

- Follow some sort of prioritization, e.g., allowlist over denylist. Match requests against each rule in order, and when any of the rules are satisfied, stop matching and either reject or pass them
- Use exclusion mode to allow only one type of allowlist or denylist to be configured

The second is simpler and easier to understand. I'm not sure there is a need for the first way of this plugin.

monkeyDluffy6017:
@bzp2010 The first one looks meanless, if you set a allowlist, all you want is for the allowlist to pass, why do you set a blacklist at all?

jiangfucheng:
[...]

PR #9448, opened 2023-05-09
feat(ua-restriction): support allowlist-only mode

jiangfucheng:
### Description

Close https://github.com/apache/apisix/issues/7467
### Checklist


Sn0rt:
it sounds good.

if the user only set allow list and without deny list to block others UA which is not in allow list is semantic.

@leslie-tsang pls take a look.

docs/zh/latest/plugins/ua-restriction.md
```diff
@@ -43,7 +43,8 @@ description: 本文介绍了 Apache APISIX ua-restriction 插件的使用方法
 
 :::note
 
-`allowlist` 和 `denylist` 可以同时启用。同时启用时，插件会根据 `User-Agent` 先检查 `allowlist`，再检查 `denylist`。
+1.`allowlist` 和 `denylist` 可以同时启用。同时启用时，插件会根据 `User-Agent` 先检查 `allowlist`，再检查 `denylist`。
+2.如果只设置了 `allowlit` 并且 `User-Agent` 不在 `allowlist` 中，请求将会被拒绝。
```
monkeyDluffy6017, line 47:
```suggestion
2.如果只设置了 `allowlist` 并且 `User-Agent` 不在 `allowlist` 中，请求将会被拒绝。
```

docs/zh/latest/plugins/ua-restriction.md
```diff
@@ -43,7 +43,8 @@ description: 本文介绍了 Apache APISIX ua-restriction 插件的使用方法
 
 :::note
 
-`allowlist` 和 `denylist` 可以同时启用。同时启用时，插件会根据 `User-Agent` 先检查 `allowlist`，再检查 `denylist`。
+1.`allowlist` 和 `denylist` 可以同时启用。同时启用时，插件会根据 `User-Agent` 先检查 `allowlist`，再检查 `denylist`。
```
monkeyDluffy6017, line 46: 1. `allowlist` 和 `denylist` 可以单独启用。
2. 同时启用时，插件会根据 `User-Agent` 先检查 `allowlist`，再检查 `denylist`。

docs/en/latest/plugins/ua-restriction.md
```diff
@@ -43,7 +43,8 @@ A common scenario is to set crawler rules. `User-Agent` is the identity of the c
 
 :::note
 
-Both `allowlist` and `denylist` can be used on their own. If they are used together, the `allowlist` matches before the `denylist`.
+1.Both `allowlist` and `denylist` can be used on their own. If they are used together, the `allowlist` matches before the `denylist`.
+2.If only `allowlist` is set and the `User-Agent` is not in the `allowlist`, the request will be rejected.
```
monkeyDluffy6017, line 47:
```suggestion
1. Both `allowlist` and `denylist` can be used on their own. 
2. If they are used together, the `allowlist` matches before the `denylist`.
3. If only `allowlist` is set and the `User-Agent` is not in the `allowlist`, the request will be rejected.
```

apisix/plugins/ua-restriction.lua
```diff
@@ -142,7 +142,7 @@ function _M.access(conf, ctx)
         match = lrucache_useragent(user_agent, conf, match_user_agent, user_agent, conf)
     end
 
-    if match > MATCH_ALLOW then
+    if match > MATCH_ALLOW or (match == MATCH_NONE and conf.allowlist and not conf.denylist) then
```
monkeyDluffy6017, line 145: What happens if there is more than one UA？
It will depends on the last UA in the array and it looks strange

jiangfucheng:
```
   if type(user_agent) == "table" then
        for _, v in ipairs(user_agent) do
            if type(v) == "string" then
                match = lrucache_useragent(v, conf, match_user_agent, v, conf)
                if match > MATCH_ALLOW then
                    break
                end
            end
        end
    else
        match = lrucache_useragent(user_agent, conf, match_user_agent, user_agent, conf)
    end
```

When more than one UA, this code will traverse all of the UA talbe util `match > MATCH_ALLOW`, so even if `match == MATCH_NONE or match == MATCH_ALLOW` , it will still continue to traverse.

I don't know if I understand correctly, if not, can you give me a example to explain it?

monkeyDluffy6017: It's a bit complicated here, could we refactor this code?

jiangfucheng: @monkeyDluffy6017 PLAT, thanks.

jiangfucheng: @monkeyDluffy6017 PTAL, thanks.

monkeyDluffy6017:
I think what we need is not just a allowlist-only mode, the blacklist and the allowlist should not exist at the same time

jiangfucheng:
> 



> I think what we need is not just a allowlist-only mode, the blacklist and the allowlist should not exist at the same time

You are right, I also think so.  As mentioned in above issue,  I don't change this logic to be compatible with the old logic.

jiangfucheng:
> I think what we need is not just a allowlist-only mode, the blacklist and the allowlist should not exist at the same time

I also think so, I don't change this logic for compatibility with old logic.

monkeyDluffy6017:
Let's discuss in the issue, i think the plugin should be refactored

Sn0rt:
> Let's discuss in the issue, i think the plugin should be refactored

ok. can you raise a document for discuss?

jiangfucheng:
As the discussion of this [issue](https://github.com/apache/apisix/issues/7467), will raise a new PR for it, So this PR will be closed.

PR #9456, opened 2023-05-10
feat(config_etcd): use a single long http connection to watch all resources

kingluo:
### Description

1. only one http connection to watch the prefix for all resources
2. use chunked streaming to receive events in the connection, timeout=50 second

Sequence Diagram:

```
/routes watcher->/routes watcher: first sync_data()
/routes watcher->main watcher: start
/routes watcher->/routes watcher: wait main watcher
/upstreams watcher->/upstreams watcher: first sync_data(), wait main watcher
main watcher->etcd: get current etcd revision and start watch there
main watcher-->/routes watcher: started
main watcher-->/upstreams watcher: started
/routes watcher->etcd: readdir() if need_reloaded
/routes watcher->main watcher: iterate history watch events, found
/upstreams watcher->main watcher: no history event, wait...
etcd-->main watcher: watch event
main watcher->main watcher: save event
main watcher->/upstreams watcher: notify
/upstreams watcher->main watcher: iterate history watch events, found
etcd-->main watcher: timeout
main watcher->etcd: reconnect and restart watch from max received revision
etcd-->main watcher: compacted
main watcher->etcd: reconnect and restart watch from compacted revision
```
sequence_diagram (21)

### Checklist

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -46,6 +46,8 @@ local error        = error
 local rand         = math.random
 local constants    = require("apisix.constants")
 local health_check = require("resty.etcd.health_check")
+local semaphore    = require("ngx.semaphore")
+local tablex       = require("pl.tablex")
 
 
 local is_http = ngx.config.subsystem == "http"
@@ -58,6 +60,7 @@ if not is_http then
 end
 local created_obj  = {}
 local loaded_configuration = {}
+local watch_ctx
 
 
 local _M = {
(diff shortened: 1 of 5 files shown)


t/core/etcd-sync.t
```diff
@@ -22,65 +22,7 @@ run_tests;
...
-                ngx.say("prev_index not update")
-            end
-        }
-    }
---- request
-GET /t
---- response_body
-prev_index updated
---- error_log eval
-qr/(create watch stream for key|cancel watch connection success)/
```
kingluo, line 79: deprecated test. etcd watch is a single long connection now.

ci/linux_openresty_runner.sh
```diff
@@ -18,5 +18,5 @@
 
 
 export OPENRESTY_VERSION=source
-export TEST_CI_USE_GRPC=true
+#export TEST_CI_USE_GRPC=true
```
kingluo, line 21: we should focus on etcd http testing now.

apisix/core/config_etcd.lua
```diff
@@ -75,6 +82,203 @@ local mt = {
...
+    watch_ctx.started = true
+
+    log.warn("main etcd watcher started, revision=", watch_ctx.rev)
+    for _, sema in pairs(watch_ctx.wait_init) do
+        sema:post()
+    end
+    watch_ctx.wait_init = nil
+
+    local opts = {}
+    opts.timeout = 50 -- second
```
kingluo, line 170: The choice of 50 seconds is to make it smaller than the default proxy_read_timeout value, 60 seconds, so that nginx will not print error logs, such as:

```
2023/05/10 23:43:44 [error] 3668019#3668019: *809 upstream timed out (110: Connection timed out) while reading upstream, client: unix:, server: , request: "POST /v3/watch HTTP/1.1", upstream: "http://127.0.0.1:2379/v3/watch", host: "127.0.0.1"
```

apisix/core/config_etcd.lua
```diff
@@ -75,6 +82,203 @@ local mt = {
...
+                end
+            end
+
+            for i = 1,min_idx-1 do
+                watch_ctx.res[i] = false
+            end
+
+            if min_idx > 100 then
+                for k, idx in pairs(watch_ctx.idx) do
+                    watch_ctx.idx[k] = idx-min_idx+1
```
monkeyDluffy6017, line 234:
```suggestion
                    watch_ctx.idx[k] = idx - min_idx + 1
```

apisix/core/config_etcd.lua
```diff
@@ -157,45 +361,69 @@ local function flush_watching_streams(self)
...
     end
 
-    -- in etcd v3, the 1st res of watch is watch info, useless to us.
-    -- try twice to skip create info
-    local res, err = res_func()
-    if not res or not res.result or not res.result.events then
-        res, err = res_func()
-    end
+    ::iterate_events::
+    for i = watch_ctx.idx[key],#watch_ctx.res do
```
monkeyDluffy6017, line 370:
```suggestion
    for i = watch_ctx.idx[key], #watch_ctx.res do
```

apisix/core/config_etcd.lua
```diff
@@ -157,45 +361,69 @@ local function flush_watching_streams(self)
...
+        watch_ctx.idx[key] = i+1
```
monkeyDluffy6017, line 371:
```suggestion
        watch_ctx.idx[key] = i + 1
```

kingluo:
> Why do you call this "long http connection", the http_cli will be recreated every time: https://github.com/api7/lua-resty-etcd/blob/master/lib/resty/etcd/v3.lua#L803?

Please read the code carefully.
Once the http connection for watching is created, it keeps receiving the watch event chunks, unless it is timeout or compacted.
So is it a long or not?

apisix/core/config_etcd.lua
```diff
@@ -75,6 +85,208 @@ local mt = {
...
+        local res_func, err, http_cli = watch_ctx.cli:watchdir(watch_ctx.prefix, opts)
+        if not res_func then
+            log.error("watchdir: ", err)
+            ngx_sleep(3)
+            goto restart_watch
+        end
+
+        ::watch_event::
+        while true do
+            local res, err = res_func()
```
monkeyDluffy6017, line 191: It's very good design here, the server response is a stream, if we don't close the connection here, we could get the response event one by one

apisix/core/config_etcd.lua
```diff
@@ -157,45 +369,71 @@ local function flush_watching_streams(self)
...
+            for _, evt in ipairs(res.result.events) do
+                if evt.kv.key:find(key) == 1 then
+                    insert_tab(res2.result.events, evt)
+                end
+            end
+            if log_level >= NGX_INFO then
+                log.info("http_waitdir: ", inspect(res2))
+            end
+            return res2
         end
```
monkeyDluffy6017, line 414:
```suggestion
        local events = {}
        if tonumber(res.result.header.revision) > self.prev_index then
            for _, evt in ipairs(res.result.events) do
                if evt.kv.key:find(key) == 1 then
                    insert_tab(events, evt)
                end
            end
        end

        if not core.table.isempty(events) then
            local res2 = tablex.deepcopy(res)
            res2.result.events = events
            return res2
        end
```

apisix/core/config_etcd.lua
```diff
@@ -157,45 +369,71 @@ local function flush_watching_streams(self)
...
-    local opts = {}
-    opts.start_revision = modified_index
-    opts.timeout = timeout
-    opts.need_cancel = true
-    local res_func, func_err, http_cli = etcd_cli:watchdir(key, opts)
-    if not res_func then
-        return nil, func_err
+local function http_waitdir(self, etcd_cli, key, modified_index, timeout)
+    if not watch_ctx.idx[key] then
+        watch_ctx.idx[key] = 1
```
monkeyDluffy6017, line 374: could we store the min_idx and assgin here?

apisix/core/config_etcd.lua
kingluo: No need to create a new table (and replace a newly created sub-table) here.

apisix/core/config_etcd.lua
kingluo: It's used to init the idx per resource. The min_idx is the global status for all resources. What do you mean?

kingluo: Each resource must start from 1, because it needs to consume the earliest event!

monkeyDluffy6017: But the resource before the mini_idx have been cleaned?
```
            for i = 1, min_idx - 1 do
                watch_ctx.res[i] = false
            end
```

apisix/core/config_etcd.lua
monkeyDluffy6017:
```suggestion
        local found = false
        local res2
        if tonumber(res.result.header.revision) > self.prev_index then
            for _, evt in ipairs(res.result.events) do
                if evt.kv.key:find(key) == 1 then
                    if not found then 
                        res2 = tablex.deepcopy(res)
                        table.clear(res2.result.events)
                        found = true
                    end

                    insert_tab(res2.result.events, evt)
                end
            end
        end
```

monkeyDluffy6017: we don't need to iterate from the beginning of the `res.result.events` again

apisix/core/config_etcd.lua
kingluo: It will skip the false items. When it starts to match, the queue may be not cleaned.

monkeyDluffy6017: so store the min_idx, and new resource could start from min_idx, because the event before min_idx is false, right ?

kingluo: 1. min_idx maybe 0, which is an internal state of the main watcher and comes from the summary of consumer idx of all resources
2. after the first watch iteration, the consume idx of the resource will be kept in sync with the global min_idx.
3. the sync gap is small, now the max gap is 100, so the overhead should be ignored

apisix/core/config_etcd.lua
kingluo: fix.

apisix/core/config_etcd.lua
```diff
@@ -75,6 +85,208 @@ local mt = {
...
+        log.error("cancel watch failed: ", err)
+    end
+end
+
+
+-- append res to the queue and notify pending watchers
+local function produce_res(res, err)
+    if log_level >= NGX_INFO then
+        log.info("append res: ", inspect(res), ", err: ", inspect(err))
+    end
```
monkeyDluffy6017, line 118: Why not use `json.delay_encode` here?

apisix/core/config_etcd.lua
monkeyDluffy6017: ok

apisix/core/config_etcd.lua
kingluo: the delay stuff has bug: https://github.com/apache/apisix/blob/a943c036987aa2e9a34f05060015ff65b8913345/apisix/core/json.lua#L116-L120
It only uses a singleton table to log, but here I need two vars to log.
And inspect is more informational than json for debugging.

leslie-tsang (approved):

monkeyDluffy6017 (approved):

Issue #9481, opened 2023-05-15
feat: Inform user when call "/plugins/reload" using a wrong HTTP method


piglei:
### Description

When a user calls the "/apisix/admin/plugins/reload" endpoint to reload the plugins, there is a chance that a wrong HTTP method was used instead of "PUT". In the current implementation, when this happens, for example the endpoint was called via GET method, APISIX will try to find any plugins called "reload", which produces the following response:

```json
{
	"error_msg": "failed to load plugin reload"
}
```

To improve the experience, we can consider returning a more user-friendly response when the "plugins/reload" is requested using a wrong method. For example, return this instead:

```json
{
	"error_msg": "please use PUT method to reload the plugins, GET method is not allowed."
}
```

> NOTE: This change also makes "reload" a reserved plugin name (wondering if this concept exists in APISIX), because it can never be retrieved via "GET /plugins/<name>".

---

Related PR: https://github.com/apache/apisix/pull/9482

PR #9482, opened 2023-05-15
feat: Inform user when request /plugins/reload/ using wrong HTTP methods

piglei:
### Description

Inform user when request /plugins/reload/ using wrong HTTP methods. More details in https://github.com/apache/apisix/issues/9481

### Checklist

  - **confirmation required.**

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -104,9 +104,10 @@ local function check_token(ctx)
     return true
 end
 
--- Set the `apictx` variable and check admin api token, when the check fails, an error
--- response will be returned. This is a higher wrapper for `check_token` function.
--- @returns A boolean value indicates whether the check passed.
+-- Set the `apictx` variable and check admin api token, if the check fails, the current
+-- request will be interrupted and an error response will be returned.
+--
+-- NOTE: This is a higher wrapper for `check_token` function.
 local function set_ctx_and_check_token()
     local api_ctx = {}
     core.ctx.set_vars_meta(api_ctx)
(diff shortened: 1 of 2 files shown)


shreemaan-abhishek:
I think it would be better to allow usage of GET method for plugin reload instead of warning user about wrong method usage, WDYT?

piglei:
> I think it would be better to allow usage of GET method for plugin reload instead of warning user about wrong method usage, WDYT?

Adding support for using GET to reload the plugins is okay and convenient. But consider that the "reload" endpoint is in the "/plugins/reload/<:string>" pattern family and the result of calling other endpoints via GET is reading information of the plugins, which has no side-effect at all. So, it's better to only support the PUT method for doing a plugin reload.

apisix/admin/init.lua
```diff
@@ -142,14 +162,8 @@ end
...
 local function run()
-    local api_ctx = {}
-    core.ctx.set_vars_meta(api_ctx)
-    ngx.ctx.api_ctx = api_ctx
-
-    local ok, err = check_token(api_ctx)
-    if not ok then
-        core.log.warn("failed to check token: ", err)
-        core.response.exit(401, {error_msg = "failed to check token"})
+    if not set_ctx_and_check_token() then
```
monkeyDluffy6017, line 165: no return value is needed, the `core.response.exit` will terminate the processing of the current request

piglei: Thank you for pointing this out. I've updated the function to remove the value-returning logic, the function-call statements were updated as well.

PS: After this modification, I'm thinking that if "set_ctx_and_check_token" is still a good name, because the name didn't imply that the request would exit. Maybe "exit_if_token_invalid" is a better name?

monkeyDluffy6017:
@piglei Could you merge the master?

piglei:
> @piglei Could you merge the master?

Synced with apache/master .

monkeyDluffy6017 (approved):

AlinsRan (approved):

AlinsRan (approved):

Neilblaze (approved):

Issue #9171, opened 2023-03-27
bug: `apisix start -c` report `failed to link customized config, error: File exists` error


nic-6443:
### Current Behavior

When starting with a custom configuration file using `start -c`, if it exits abnormally after two starts, an error will occur when trying to start again：`failed to link customized config, error: File exists`

### Expected Behavior

`apisix start -c` can be executed repeatedly and works normally.

### Steps to Reproduce
1. Create a `config.yaml` file.
```
etcd:
  host: 
    - "http://127.0.0.1:22333" # An etcd service that cannot be connected to.
  prefix: "/apisix"
```
2. start apisix three times using this configuration.

First time：
> [root@apisix-test-6c78bd9db9-d5qkh apisix]# apisix start -c ./config.yaml
/usr/local/openresty/luajit/bin/luajit ./apisix/cli/apisix.lua start -c ./config.yaml
Use customized yaml:    ./config.yaml
Warning! Request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused, retry time=1
Warning! Request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused, retry time=2
request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused
all etcd nodes are unavailable

Second time:
> [root@apisix-test-6c78bd9db9-d5qkh apisix]# apisix start -c ./config.yaml
/usr/local/openresty/luajit/bin/luajit ./apisix/cli/apisix.lua start -c ./config.yaml
Use customized yaml:    ./config.yaml
Warning! Request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused, retry time=1
Warning! Request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused, retry time=2
request etcd endpoint 'http://127.0.0.1:22333/version' error, connection refused
all etcd nodes are unavailable

Third time:
> [root@apisix-test-6c78bd9db9-d5qkh apisix]# apisix start -c ./config.yaml
/usr/local/openresty/luajit/bin/luajit ./apisix/cli/apisix.lua start -c ./config.yaml
**failed to link customized config, error: File exists**

### Error reason
https://github.com/apache/apisix/blob/81149cd31567f6a86e33100fa7d09d7550073157/apisix/cli/ops.lua#L806-L817
[...]

PR #9521, opened 2023-05-21
fix: start error bug

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/9171

### Checklist


apisix/cli/ops.lua
```diff
@@ -807,17 +807,19 @@ local function start(env, ...)
...
+            local ok, err1 = lfs.link(customized_yaml, local_conf_path)
+            if not ok then
+                ok, err = os_rename(local_conf_path_bak,  local_conf_path)
+                if not ok then
+                    util.die("failed to recover original config file, error: ", err)
+                end
+                util.die("failed to link customized config, error: ", err1)
             end
-            util.die("failed to link customized config, error: ", err1)
         end
```
monkeyDluffy6017, line 823: Is this an appropriate change? the customized_yaml won't work if there is a `local_conf_path_bak` file

tokers: Indeed I have another option, why do we need a backup file?

jiangfucheng: Once start APISIX use customized config, the original `config.yaml ` will be `rename` to `config.yaml.bak`  and set a hard link from customized config to `config.yaml`, if occur some errors cause start APISIX fail then it will not restore these config files, so we don't need do above opeartion again, just use `config.yaml` as apisix config directly is enough.
Unless the user create `local_conf_path_bak` manually before start apisix. 
If we can detected if have errors when start APISIX, it maybe the best solution, but, I have no idea about this.

monkeyDluffy6017: I think we can fix this bug by https://github.com/apache/apisix/pull/9701#discussion_r1259432308
Do the `cleanup` when abnormal

jiangfucheng: I fetch codes of this PR, and test it, it can't resolve this https://github.com/apache/apisix/issues/9171. You can try it.

jiangfucheng: I think @tokers is right, we should find a way to access customized config file directly, but not rename the customized config file to origin `config.yaml`, it brings too many problems.

I have a simple idea, we can save the custom config file name to a sparate file(like `.config_name`) when we start apisix, so we can access it directly when we need it.

monkeyDluffy6017: Then you will have two config file: config.yaml and the custom_config.yaml, how do you decide which one to choose each time you start APISIX?

monkeyDluffy6017: The pr is not right, what i want to say is that *when starting APISIX fails, recover the config.yaml file*.  
So we can do like this:
```
    if env.deployment_role ~= "data_plane" then
        local err = init_etcd(env, args)
        if err then
            cleanup()
            util.die(err)
        end
    end
```
Now we die in the `init_etcd` with no returns, so we should modify the `init_etcd` function to make sure the it returns the error msg instead of just crashing!

jiangfucheng: > Then you will have two config file: config.yaml and the custom_config.yaml, how do you decide which one to choose each time you start APISIX?

which config will be choose depend on `--config` flag.  After APISIX started, we will read `.config_name` file content to decide which config will be choose.

jiangfucheng: > The pr is not right, what i want to say is that _when starting APISIX fails, recover the config.yaml file_. We can do like this:
> 
> ```
>     if env.deployment_role ~= "data_plane" then
>         local err = init_etcd(env, args)
>         if err then
>             cleanup()
>             util.die(err)
>         end
>     end
> ```
> 
> Now we die in the `init_etcd` with no returns, so we should modify the `init_etcd` function to make sure the it returns the error msg instead of just crashing! The key point is `cleanup()`

I'm not sure if only `init_etcd` will cause crash, if so, I think is ok.

monkeyDluffy6017: > which config will be choose depend on --config flag. After APISIX started, we will read .config_name file content to decide which config will be choose.
Looks good, but you still have to copy the custom config file to the conf directory, right?

jiangfucheng: Emm, In current logic, you are right. Give me some time to test it, I think it can be solved.

monkeyDluffy6017:
I will close this pr because it is fixed by https://github.com/apache/apisix/pull/9846

PR #9534, opened 2023-05-24
feat: add service path_prefix

kingluo:
### Description

If `path_prefix` is specified in service, then all routes which reference this service add that prefix to uris/uri before creating the radix tree.

If `strip_path_prefix` is specified in route, then it removes the `path_prefix` before processing all phases.

### Checklist


t/node/service-path-prefix.t
```diff
@@ -0,0 +1,78 @@
...
+                ]]
+            ))
+            ngx.sleep(0.5)
+
+            local http = require "resty.http"
+            local httpc = http.new()
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello"
+            local res, err = httpc:request_uri(uri)
+            ngx.status = res.status
+            ngx.log(ngx.WARN, require("inspect")(res))
```
monkeyDluffy6017, line 71: Is this for debug?

kingluo: Yes, it does not affect the test correctness.

monkeyDluffy6017: Remove it if you don't need it

apisix/http/route.lua
```diff
@@ -31,6 +31,23 @@ local loadstring = loadstring
 local _M = {}
 
 
+function _M.prefix_uris(route, service)
+    if service and service.value.path_prefix then
+        local uri, uris
+        if route.value.uris then
+            uris = core.table.new(#route.value.uris)
+            for _, uri in ipairs(route.value.uris) do
+                core.table.insert(uris, service.value.path_prefix .. uri)
```
tokers, line 40: A tiny detail: what if the path_prefix is ended with `/`, in such a case, we'll have continuous slashes. E.g., `/api/v1//order`. While this affects the route match?

apisix/init.lua
```diff
@@ -640,6 +641,14 @@ function _M.http_access_phase()
             enable_websocket = service.value.enable_websocket
         end
 
+        if route.value.strip_path_prefix and service.value.path_prefix then
+            local uri = api_ctx.var.uri
+            if uri:find(service.value.path_prefix) == 1 then
```
tokers, line 646: Why not using `core.string.has_prefix`?

docs/en/latest/admin-api.md
```diff
@@ -585,6 +586,7 @@ Service resource request address: /apisix/admin/services/{id}
 | desc             | False    | Auxiliary   | Description of usage scenarios.                                                                                    | service xxxx                                     |
 | labels           | False    | Match Rules | Attributes of the Service specified as key-value pairs.                                                            | {"version":"v2","build":"16","env":"production"} |
 | enable_websocket | False    | Auxiliary   | Enables a websocket. Set to `false` by default.                                                                    |                                                  |
+| path_prefix | False    | String   | Add prefix to uri.                                                                    |                                                  |
```
monkeyDluffy6017, line 589: Need to update the chinese doc

apisix/init.lua
kingluo: lua built-in string.find is more simple and high efficient.

apisix/http/route.lua
kingluo: That depends on the validation requirements.
Not only the trailing slash, the slash, and other invalid chars may exist in any position of the string. So if we need real validation, we need to define its requirement.
That is, is it enough if I just remove the trailing slash if exists?

leslie-tsang:
Need to discuss first. :)

PR #9611, opened 2023-06-06
docs: update `apisix` section in `config-default.yaml`

kayx23:
### Description

Update comments/documentations in `config-default.yaml` as the file is referenced in new docs.

Will make the updates in a series of PRs (by `apisix`, `nginx_config`, and so on) for easier review. 

This PR is for the `apisix` section.

**No functional changes in these PRs.**

### Checklist

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -19,8 +19,8 @@
 #
 
 apisix:
-  # node_listen: 9080               # APISIX listening port, if single
-  node_listen:                      # APISIX listening ports, if multiple
+  # node_listen: 9080               # Single APISIX listening port
+  node_listen:                      # Multiple APISIX listening ports
     - 9080
   #   - port: 9081
   #     enable_http2: true          # If not set, default to `false`
@@ -28,9 +28,9 @@ apisix:
   #     port: 9082
   #     enable_http2: true
   enable_admin: true
-  enable_dev_mode: false            # Set nginx worker_processes to 1 if true
-  enable_reuseport: true            # Enable nginx SO_REUSEPORT switch if set to true.
-  show_upstream_status_in_response_header: false # when true all upstream status write to `X-APISIX-Upstream-Status` otherwise only 5xx code
+  enable_dev_mode: false            # If true, nginx worker_processes will be set to 1
+  enable_reuseport: true            # If true, nginx SO_REUSEPORT option will be enabled
+  show_upstream_status_in_response_header: false   # If true, show upstream HTTP status code in the response header `X-APISIX-Upstream-Status`
   enable_ipv6: true
 
   #proxy_protocol:                  # Proxy Protocol configuration
@@ -40,9 +40,9 @@ apisix:
                                     # receive http request with proxy protocol
   #  listen_https_port: 9182        # The port with proxy protocol for https
   #  enable_tcp_pp: true            # Enable the proxy protocol for tcp proxy, it works for stream_proxy.tcp option
-  #  enable_tcp_pp_to_upstream: true # Enables the proxy protocol to the upstream server
+  #  enable_tcp_pp_to_upstream: true # Enable the proxy protocol to the upstream server
   enable_server_tokens: true        # Whether the APISIX version number should be shown in Server header.
-                                    # It's enabled by default.
+                                    # True by default.
 
   # configurations to load third party code and/or override the builtin one.
   extra_lua_path: ""                # extend lua_package_path to load third party code
@@ -213,18 +213,18 @@ nginx_config:                     # config for render the template to generate n
     #custom_lua_shared_dict:       # add custom shared cache to nginx.conf
     #  ipc_shared_dict: 100m       # custom shared cache, format: `cache-key: cache-size`
 
-    # Enables or disables passing of the server name through TLS Server Name Indication extension (SNI, RFC 6066)
+    # Enable or disable passing of the server name through TLS Server Name Indication extension (SNI, RFC 6066)
     # when establishing a connection with the proxied HTTPS server.
     proxy_ssl_server_name: true
     upstream:
-      keepalive: 320                # Sets the maximum number of idle keepalive connections to upstream servers that are preserved in the cache of each worker process.
+      keepalive: 320                # Set the maximum number of idle keepalive connections to upstream servers that are preserved in the cache of each worker process.
                                     # When this number is exceeded, the least recently used connections are closed.
-      keepalive_requests: 1000      # Sets the maximum number of requests that can be served through one keepalive connection.
+      keepalive_requests: 1000      # Set the maximum number of requests that can be served through one keepalive connection.
                                     # After the maximum number of requests is made, the connection is closed.
-      keepalive_timeout: 60s        # Sets a timeout during which an idle keepalive connection to an upstream server will stay open.
+      keepalive_timeout: 60s        # Set a timeout during which an idle keepalive connection to an upstream server will stay open.
     charset: utf-8                  # Adds the specified charset to the "Content-Type" response header field, see
                                     # http://nginx.org/en/docs/http/ngx_http_charset_module.html#charset
-    variables_hash_max_size: 2048   # Sets the maximum size of the variables hash table.
+    variables_hash_max_size: 2048   # Set the maximum size of the variables hash table.
 
     lua_shared_dict:
       internal-status: 10m
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -28,8 +28,8 @@ apisix:
     #   port: 9082
     #   enable_http2: true
   enable_admin: true           # Admin API
-  enable_dev_mode: false       # If true, nginx worker_processes will be set to 1.
-  enable_reuseport: true       # If true, nginx SO_REUSEPORT option will be enabled.
+  enable_dev_mode: false       # If true, set nginx worker_processes to 1.
+  enable_reuseport: true       # If true, enable nginx SO_REUSEPORT option.
   show_upstream_status_in_response_header: false  # If true, include the upstream HTTP status code in
                                                   # the response header `X-APISIX-Upstream-Status`.
                                                   # If false, show `X-APISIX-Upstream-Status` only if
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -19,8 +19,8 @@
 #
 
 apisix:
-  # node_listen: 9080               # APISIX listening port (single)
-  node_listen:                      # APISIX listening ports (multiple)
+  # node_listen: 9080               # APISIX listening port, if single
+  node_listen:                      # APISIX listening ports, if multiple
     - 9080
   #   - port: 9081
   #     enable_http2: true          # If not set, default to `false`
@@ -28,7 +28,7 @@ apisix:
   #     port: 9082
   #     enable_http2: true
   enable_admin: true
-  enable_dev_mode: false            # Set nginx worker_processes to 1 if set to true
+  enable_dev_mode: false            # Set nginx worker_processes to 1 if true
   enable_reuseport: true            # Enable nginx SO_REUSEPORT switch if set to true.
   show_upstream_status_in_response_header: false # when true all upstream status write to `X-APISIX-Upstream-Status` otherwise only 5xx code
   enable_ipv6: true
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -109,6 +109,7 @@ apisix:
 
     key_encrypt_salt:           # Set the encryption key for AES-128-CBC. It should be a
       - edd1c9f0985e76a2        # hexadecimal string of length 16.
+                                # If not set, APISIX saves the original data into etcd.
                                 # For the convenience of key replacement, key_encrypt_salt
                                 # supports the use of multiple keys. The additional keys are
                                 # used in sequential order for decryption, when the previous
@@ -130,6 +131,7 @@ apisix:
     enable: false
     keyring:                      # Set the encryption key for AES-128-CBC. It should be a
       - qeddd145sfvddff3          # hexadecimal string of length 16.
+                                  # If not set, APISIX saves the original data into etcd.
                                   # For the convenience of key replacement, keyring
                                   # supports the use of multiple keys. The additional keys are
                                   # used in sequential order for decryption, when the previous
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -19,8 +19,8 @@
 #
 
 apisix:
-  # node_listen: 9080          # Use this if APISIX listens on a single port.
-  node_listen:                 # Use this if APISIX listens on multiple ports.
+  # node_listen: 9080          # APISIX listening port.
+  node_listen:                 # APISIX listening ports.
     - 9080
     # - port: 9081
     #   enable_http2: true     # If not set, default to `false`.
@@ -82,6 +82,7 @@ apisix:
   #   udp:                        # UDP proxy listening ports 
   #     - 9200
   #     - "127.0.0.1:9201"
+  
   # dns_resolver:                 # If not set, read from `/etc/resolv.conf`
   #   - 1.1.1.1
   #   - 8.8.8.8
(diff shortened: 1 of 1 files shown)


navendu-pottekkat:
@kayx23 Why is this PR in draft?

conf/config-default.yaml
```diff
@@ -14,24 +14,23 @@
...
+  #   - ip: 127.0.0.2               # If not set, default to `0.0.0.0`
   #     port: 9082
   #     enable_http2: true
   enable_admin: true
-  enable_dev_mode: false            # Sets nginx worker_processes to 1 if set to true
-  enable_reuseport: true            # Enable nginx SO_REUSEPORT switch if set to true.
-  show_upstream_status_in_response_header: false # when true all upstream status write to `X-APISIX-Upstream-Status` otherwise only 5xx code
+  enable_dev_mode: false            # If true, nginx worker_processes will be set to 1
+  enable_reuseport: true            # If true, nginx SO_REUSEPORT option will be enabled
+  show_upstream_status_in_response_header: false   # If true, show upstream HTTP status code in the response header `X-APISIX-Upstream-Status`
```
kayx23, line 33: Need to test

conf/config-default.yaml
```diff
@@ -14,36 +14,34 @@
...
   enable_ipv6: true
-
-  #proxy_protocol:                  # Proxy Protocol configuration
-  #  listen_http_port: 9181         # The port with proxy protocol for http, it differs from node_listen and admin_listen.
-                                    # This port can only receive http request with proxy protocol, but node_listen & admin_listen
-                                    # can only receive http request. If you enable proxy protocol, you must use this port to
-                                    # receive http request with proxy protocol
-  #  listen_https_port: 9182        # The port with proxy protocol for https
-  #  enable_tcp_pp: true            # Enable the proxy protocol for tcp proxy, it works for stream_proxy.tcp option
-  #  enable_tcp_pp_to_upstream: true # Enables the proxy protocol to the upstream server
```
kayx23, line 44: why is `enable_tcp_pp_to_upstream` needed

```
        {% if proxy_protocol and proxy_protocol.enable_tcp_pp_to_upstream then %}
        proxy_protocol on;
        {% end %}
```

conf/config-default.yaml
```diff
@@ -14,53 +14,48 @@
...
-
-  # configurations to load third party code and/or override the builtin one.
-  extra_lua_path: ""                # extend lua_package_path to load third party code
-  extra_lua_cpath: ""               # extend lua_package_cpath to load third party code
-  #lua_module_hook: "my_project.my_hook"  # the hook module which will be used to inject third party code into APISIX
-
-  proxy_cache:                      # Proxy Caching configuration
-    cache_ttl: 10s                  # The default caching time in disk if the upstream does not specify the cache time
-    zones:                          # The parameters of a cache
-      - name: disk_cache_one        # The name of the cache, administrator can specify
```
kayx23, line 56: Need to understand what it means by "administrator can specify which cache to use by name in the Admin API". Don't remember anything about cache in Admin API.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
+    
+  enable_control: true  # Control API
+  # control:
   #  ip: 127.0.0.1
   #  port: 9090
-  disable_sync_configuration_during_start: false  # safe exit. Remove this once the feature is stable
-  data_encryption:                # add `encrypt_fields = { $field },` in plugin schema to enable encryption
-    enable: false                 # if not set, the default value is `false`.
+
+  disable_sync_configuration_during_start: false  # Safe exit. TO BE REMOVED.
```
kayx23, line 123: confirmed by @monkeyDluffy6017 that this is no longer in use.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-  #     enable_http2: true          # If not set, the default value is `false`.
-  #   - ip: 127.0.0.2               # Specific IP, If not set, the default value is `0.0.0.0`.
-  #     port: 9082
-  #     enable_http2: true
+    # - port: 9081
+    #   enable_http2: true     # If not set, default to `false`.
+    # - ip: 127.0.0.2          # If not set, default to `0.0.0.0`/
+    #   port: 9082
+    #   enable_http2: true
   enable_admin: true
```
navendu-pottekkat, line 30: Add a comment here?

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-  enable_dev_mode: false            # Sets nginx worker_processes to 1 if set to true
-  enable_reuseport: true            # Enable nginx SO_REUSEPORT switch if set to true.
-  show_upstream_status_in_response_header: false # when true all upstream status write to `X-APISIX-Upstream-Status` otherwise only 5xx code
+  enable_dev_mode: false       # If true, nginx worker_processes will be set to 1.
+  enable_reuseport: true       # If true, nginx SO_REUSEPORT option will be enabled.
```
navendu-pottekkat, line 32: Not for this, but we should also look into what attributes need an external URL reference. Linking to Nginx, APISIX docs etc. But not needed here because it is trivial.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-        disk_path: /tmp/disk_cache_one  # The path to store the cache data (disk)
-        cache_levels: 1:2           # The hierarchy levels of a cache (disk)
-      #- name: disk_cache_two
+  # proxy_protocol:                    # PROXY Protocol configuration
+  #   listen_http_port: 9181           # APISIX listening port for HTTP traffic with PROXY protocol.
+  #   listen_https_port: 9182          # APISIX listening port for HTTPS traffic with PROXY protocol.
+  #   enable_tcp_pp: true              # Enable the PROXY protocol when stream_proxy.tcp is set.
+  #   enable_tcp_pp_to_upstream: true  # Enable the PROXY protocol.
+  
+  enable_server_tokens: true           # If ture, show APISIX version in the `Server` response header.
```
navendu-pottekkat, line 45:
```suggestion
  enable_server_tokens: true           # If true, show APISIX version in the `Server` response header.
```

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
+  #     - addr: "127.0.0.1:9101"
+  #   udp:                        # UDP proxy listening ports 
+  #     - 9200
+  #     - "127.0.0.1:9201"
+  
+  # dns_resolver:                 # If not set, read from `/etc/resolv.conf`
+  #   - 1.1.1.1
+  #   - 8.8.8.8
+  # dns_resolver_valid: 30        # Override the default TTL of the DNS records.
+  resolver_timeout: 5             # Time that the server will wait for a response from the DNS resolver
```
navendu-pottekkat, line 90: why do we need to split comments into multiple lines?

conf/config-default.yaml
kayx23: Probably better suited for a comprehensive doc on the website about configuration options. Here we don't want to make things too cramped up.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-  #   - port: 9081
-  #     enable_http2: true          # If not set, the default value is `false`.
-  #   - ip: 127.0.0.2               # Specific IP, If not set, the default value is `0.0.0.0`.
-  #     port: 9082
-  #     enable_http2: true
+    # - port: 9081
+    #   enable_http2: true     # If not set, default to `false`.
+    # - ip: 127.0.0.2          # If not set, default to `0.0.0.0`/
+    #   port: 9082
+    #   enable_http2: true
```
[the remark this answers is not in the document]
kayx23: Could you provide some examples where # at the far left is considered a better practice?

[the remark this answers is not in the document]
navendu-pottekkat: I don't know if it is a better practice but autoformatters I use do this. I thought that was how it is done in general.

conf/config-default.yaml
navendu-pottekkat: I'm not sure. Because it would be difficult to keep a dedicated page in the documentation for these.

conf/config-default.yaml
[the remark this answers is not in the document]
kayx23: Won't fixed at this time because the position of `#` depends on the left most line of code in selection. As there are a lot of indented configurations that need to be commented out here, the `#` signs are indented as well. It would take a lot of manual effort to have them all on the far left (which the original file did not do either).

conf/config-default.yaml
kayx23: >  Linking to Nginx, APISIX docs etc. But not needed here because it is trivial.

Could you give an example where it's needed? These two line items seem clear to me that they are nginx associated and I would google for the relevant doc if interested, from a user perspective.

conf/config-default.yaml
kayx23: Style preference. Researching other styles now. 

Other reviewers: Feel free to share your thoughts on this as well.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
+    enable: false
     keyring:
-      - qeddd145sfvddff3          # If not set, will save origin value into etcd.
-                                  # If set this, the keyring should be an array whose elements are string, and the size is also 16, and it will encrypt fields with AES-128-CBC
-                                  # !!! So do not change it after encryption, it can't decrypt the fields have be saved if you change !!
-                                  # Only use the first key to encrypt, and decrypt in the order of the array.
+      - qeddd145sfvddff3          # Salt for other fields encrypted with AES-128-CBC.
+                                  # Each salt value should be a hexadecimal string of length 16.
+                                  # Support multiple salt values for rotation.
+                                  # CAUTION: DO NOT MODIFY SALT VALUE AFTER CONFIGURATIONS ARE WRITTEN INTO ETCD. VALUES CANNOT BE DECRYPTED OTHERWISE.
```
monkeyDluffy6017, line 132: 1. If not set, will save origin value into etcd.
2. Only use the first key to encrypt, and decrypt in the order of the array.
Why do you remove the two sentence? They are used to explain the function

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-  enable_reuseport: true            # Enable nginx SO_REUSEPORT switch if set to true.
-  show_upstream_status_in_response_header: false # when true all upstream status write to `X-APISIX-Upstream-Status` otherwise only 5xx code
+    # - port: 9081
+    #   enable_http2: true     # If not set, default to `false`.
+    # - ip: 127.0.0.2          # If not set, default to `0.0.0.0`/
+    #   port: 9082
+    #   enable_http2: true
+  enable_admin: true           # Admin API
+  enable_dev_mode: false       # If true, nginx worker_processes will be set to 1.
+  enable_reuseport: true       # If true, nginx SO_REUSEPORT option will be enabled.
```
monkeyDluffy6017, line 32: Is the passive voice better than the active voice, but Grammarly often tells me not to use it?

conf/config-default.yaml
kayx23: 1. if the salt is nil, the encryption won't take place either? AES-128-CBC won't work without a salt?
2. They are but not very clear to me what these mean. I looked up how AES-128-CBC and key rotation work to have a high level understanding. I would either leave it for the users to learn more about AES-128-CBC and key rotation on their own or expand the explanation here, which I need help for. Feel free to provide your suggestions.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-      - name: disk_cache_one        # The name of the cache, administrator can specify
-                                    # which cache to use by name in the admin api (disk|memory)
-        memory_size: 50m            # The size of shared memory, it's used to store the cache index for
-                                    # disk strategy, store cache content for memory strategy (disk|memory)
-        disk_size: 1G               # The size of disk, it's used to store the cache data (disk)
-        disk_path: /tmp/disk_cache_one  # The path to store the cache data (disk)
-        cache_levels: 1:2           # The hierarchy levels of a cache (disk)
-      #- name: disk_cache_two
+  # proxy_protocol:                    # PROXY Protocol configuration
+  #   listen_http_port: 9181           # APISIX listening port for HTTP traffic with PROXY protocol.
```
monkeyDluffy6017, line 40: i think the original comment is more clear?
```
# The port with proxy protocol for http, it differs from node_listen and admin_listen.
                                    # This port can only receive http request with proxy protocol, but node_listen & admin_listen
                                    # can only receive http request. If you enable proxy protocol, you must use this port to
                                    # receive http request with proxy protocol
```

conf/config-default.yaml
kayx23: Grammarly isn't right all the time.

conf/config-default.yaml
kayx23: Original comment is assuming people don't understand PROXY protocol imho and its difference from other protocols (i.e. HTTP without PROXY protocol on L4). I kinda expect people looking for this config option understand this.

conf/config-default.yaml
monkeyDluffy6017: 1. yes, the encryption won't happen if without a salt
2. For the convenience of key replacement, we support using multiple keys simultaneously, encrypting only using the first key, and using the keys in sequence to decrypt (if decryption fails).

conf/config-default.yaml
monkeyDluffy6017: ok, I can’t understand the original comment either.

conf/config-default.yaml
monkeyDluffy6017: The most important feature of Grammarly is to let you change passive voice to active voice, i think it shoud make some sense.

conf/config-default.yaml
kayx23: 1. Thanks. Will update accordingly. 
2. How about we say something like: `Support multiple salts for key rotation with AES-128-CBC. Only the first salt value is used for encryption. The additional salt values are used for decryption in sequential order if the previous decryption attempt fails.` 

    Note that I 
    * updated `key replacement` to `key rotation`.
    * differentiated salts and keys, as I believe they are not the same thing in this context.

Let me know what you think.

monkeyDluffy6017: 1. I don't think we need to keep emphasising this algorithm `AES-128-CBC`, key replacement has nothing to do with the algorithm, only one key actually works, we want to replace the key when the key is no longer secure, the rest of the keys are used to ensure that the old encrypted data can still be decrypted properly. so i say `For the convenience of key replacement`
2. the key in this context is some kind of salt, you could search `iv` and `aes`, the key is the `iv`

kayx23: 1. true but like you said "this encryption won't happen without a salt". There are algorithms that encrypt without a salt. Salts are optionally used to make things more secure. 
2. salt, key, and iv are three different things. see https://www.ibm.com/docs/en/imdm/12.0?topic=encryption-generating-aes-keys-password. The hex value here is the Salt. So why does it make sense to document salt with `For the convenience of key replacement, we support using multiple keys simultaneously, encrypting only using the first key, and using the keys in sequence to decrypt (if decryption fails).`

monkeyDluffy6017:
@kayx23 please fix the ci

conf/config-default.yaml
monkeyDluffy6017: It's not `salt`
```
aes:new(ivs, nil, aes.cipher(128, "cbc"), {iv = ivs}))
```
https://github.com/apache/apisix/blob/master/apisix/ssl.lua#LL81C29-L81C29

And the function prototypes
```
function _M.new(self, key, salt, _cipher, _hash, hash_rounds, iv_len, enable_padding)
```
https://github.com/openresty/lua-resty-string/blob/master/lib/resty/aes.lua#L120

As you see, it's used for the `key` and `iv`, not `salt`, the salt is `nil`

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
+  #       tls: true
+  #     - addr: "127.0.0.1:9101"
+  #   udp:                        # Set the UDP proxy listening ports.
+  #     - 9200
+  #     - "127.0.0.1:9201"
+
+  # dns_resolver:                 # If not set, read from `/etc/resolv.conf`
+  #   - 1.1.1.1
+  #   - 8.8.8.8
+  # dns_resolver_valid: 30        # Override the default TTL of the DNS records.
```
monkeyDluffy6017, line 89: The unit is missing?

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
+                                # Support multiple salt values for rotation.
+                                # CAUTION: DO NOT MODIFY SALT VALUE AFTER KEY IS WRITTEN TO ETCD.
+                                # VALUES CANNOT BE DECRYPTED OTHERWISE.
 
-    key_encrypt_salt:             #  If not set, will save origin ssl key into etcd.
-      - edd1c9f0985e76a2          #  If set this, the key_encrypt_salt should be an array whose elements are string, and the size is also 16, and it will encrypt ssl key with AES-128-CBC
-                                  #  !!! So do not change it after saving your ssl, it can't decrypt the ssl keys have be saved if you change !!
-                                  #  Only use the first key to encrypt, and decrypt in the order of the array.
+    # fallback_sni: "my.default.domain"      # Fallback SNI to use if client does not send SNI during
+    #                                        # the handshake.
```
monkeyDluffy6017, line 116: Fallback SNI to be used in case the client does not send SNI during the handshake.

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-                                            #  ref: https://github.com/mozilla/server-side-tls/issues/135
+    ssl_session_tickets: false  # If true, session tickets are used for SSL/TLS connections.
+                                # Disabled by default because it renders Perfect Forward Secrecy (FPS)
+                                # useless. See https://github.com/mozilla/server-side-tls/issues/135.
+
+    key_encrypt_salt:           # Salt for SSL/TLS private key used with AES-128-CBC.
+      - edd1c9f0985e76a2        # Each salt value should be a hexadecimal string of length 16.
+                                # Support multiple salt values for rotation.
+                                # CAUTION: DO NOT MODIFY SALT VALUE AFTER KEY IS WRITTEN TO ETCD.
+                                # VALUES CANNOT BE DECRYPTED OTHERWISE.
```
monkeyDluffy6017, line 113: should keep with keyring

conf/config-default.yaml
kayx23: adding "in seconds"

conf/config-default.yaml
```diff
@@ -14,122 +14,121 @@
...
-    #     port: 9445
-    #     enable_http2: true
-    #ssl_trusted_certificate: /path/to/ca-cert  # Specifies a file path with trusted CA certificates in the PEM format
-                                                # used to verify the certificate when APISIX needs to do SSL/TLS handshaking
-                                                # with external services (e.g. etcd)
-    ssl_protocols: TLSv1.2 TLSv1.3
+      # - ip: 127.0.0.3                          # If not set, default to `0.0.0.0`.
+      #   port: 9445
+      #   enable_http2: true
+    # ssl_trusted_certificate: /path/to/ca-cert  # Path to CA certificates in the PEM format.
```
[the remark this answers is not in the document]
kayx23: updating to `Set the path to CA certificates used to verify client certificates in the PEM format.`

conf/config-default.yaml
kayx23: ## Discussion Summary

### if the salt is required

`the encryption won't happen if without a salt` is false. The encryption can take place without a salt. See the example in [lua-resty-string](https://github.com/openresty/lua-resty-string/tree/master):

```lua
local aes_256_cbc_with_padding = aes:new(key, nil, aes.cipher(256,"cbc"), {iv = string.sub(key, 1, 16)}, nil, nil, enable_padding)
-- the salt is the second argument which is nil
-- function _M.new(self, key, salt, _cipher, _hash, hash_rounds, iv_len, enable_padding)
```

The reason why it was communicated that way was because these hex values in the configs are passed as `iv` not `salt`. `iv` is required for the encryption to take place. 

### if the value is a key or a salt

It is indeed used as the `iv` in code, not the `salt`.

See how `keyring` is read from local_conf `apisix.data_encryption. keyring` and saved to `ivs`.
```lua
local function get_aes_128_cbc_with_iv_gde(local_conf)
    if _aes_128_cbc_with_iv_tbl_gde == nil then
        local ivs = core.table.try_read_attr(local_conf, "apisix", "data_encryption", "keyring")
        _aes_128_cbc_with_iv_tbl_gde = init_iv_tbl(ivs)
    end
    return _aes_128_cbc_with_iv_tbl_gde
end
```
and then ivs is passed into `aes:new(ivs, nil, aes.cipher(128, "cbc"), {iv = ivs})`:
```
local function init_iv_tbl(ivs)
    local _aes_128_cbc_with_iv_tbl = core.table.new(2, 0)
    local type_ivs = type(ivs)

    if type_ivs == "table" then
        for _, iv in ipairs(ivs) do
            local aes_with_iv = assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv}))
            core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
        end
    elseif type_ivs == "string" then
        local aes_with_iv = assert(aes:new(ivs, nil, aes.cipher(128, "cbc"), {iv = ivs}))
        core.table.insert(_aes_128_cbc_with_iv_tbl, aes_with_iv)
    end

    return _aes_128_cbc_with_iv_tbl
end
```

## Conclusion

I will update the comments for these values to "keys" to be consistent with the code as suggested by @monkeyDluffy6017. I don't however understand why the key for AES-128-CBC is not 128bit, mathematically speaking.

conf/config-default.yaml
```diff
@@ -14,122 +14,128 @@
...
+  data_encryption:                # Encrypt fields specified in `encrypt_fields` in plugin schema.
+    enable: false
+    keyring:                      # Set the encryption key for AES-128-CBC. It should be a
+      - qeddd145sfvddff3          # hexadecimal string of length 16.
+                                  # For the convenience of key replacement, keyring
+                                  # supports the use of multiple keys. The additional keys are
+                                  # used in sequential order for decryption, when the previous
+                                  # decryption attempt fails.
+                                  # CAUTION: DO NOT MODIFY SALT VALUE AFTER DATA HAS BEEN WRITTEN
+                                  # TO ETCD. IT CAN RENDER THE ENCRYPTED DATA UNRECOVERABLE.
```
monkeyDluffy6017, line 138: Not metioned this "If not set, it will save the origin value into etcd." ?

monkeyDluffy6017, line 138:
```suggestion
                                  # CAUTION: If you want to change the key, please replace the first key with a new one, the old one can be placed after it, Until all data is encrypted with the new key.
```

monkeyDluffy6017, line 138:
```suggestion
                                  # CAUTION: If you want to change the key, please replace the first key with a new one, the old one can be placed after it, Until all data is encrypted with the new key.
```
```

monkeyDluffy6017 (dismissed):

leslie-tsang (approved):

monkeyDluffy6017:
@kayx23 please fix the lint error

monkeyDluffy6017 (approved):

leslie-tsang (approved):

PR #9662, opened 2023-06-14
chore: add missing `report_interval` option for `skywalking` plugin in `config-default.yaml`

kayx23:
### Description

`report_interval` was [documented in docs](https://apisix.apache.org/docs/apisix/next/plugins/skywalking/#configuring-the-endpoint) but not documented as a configuration option in `config-default.yaml`. 

This PR adds it.

### Test

https://github.com/apache/apisix/blob/7ec3c0f583510ebcdf9c414f83792e1ac247a5b9/t/admin/plugins-reload.t#L294-L299

### Checklist

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -525,6 +525,7 @@ plugin_attr:
     service_name: APISIX
     service_instance_name: APISIX Instance Name
     endpoint_addr: http://127.0.0.1:12800
+    report_interval: 1
   opentelemetry:
     trace_id_source: x-request-id
     resource:
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -525,7 +525,7 @@ plugin_attr:
     service_name: APISIX
     service_instance_name: APISIX Instance Name
     endpoint_addr: http://127.0.0.1:12800
-    report_interval: 1
+    report_interval: 3
   opentelemetry:
     trace_id_source: x-request-id
     resource:


monkeyDluffy6017 (dismissed):

leslie-tsang (commented):
I wonder if it is necessary to upload once every 1 second.

kayx23:
> I wonder if it is necessary to upload once every 1 second.

If this needs to be changed, maybe it could be in a different PR as it's somewhat of a functional change? This PR only adds the configuration option so that it's visible. It is 1 second by default and therefore isn't a functional change.

leslie-tsang:
> > I wonder if it is necessary to upload once every 1 second.
> 
> If this needs to be changed, maybe it could be in a different PR as it's somewhat of a functional change? This PR only adds the configuration option so that it's visible. It is 1 second by default and therefore isn't a functional change.

Cloud u please tell me where this `1s` comes from?

After reviewing the code, I found that it should actually be 3 seconds. Please correct me if I'm wrong.

References:
* https://github.com/apache/apisix/blob/b024f683ef6f5310a180cdb6f792365e4c78f33a/apisix/plugins/skywalking.lua#L44-L46
* https://github.com/apache/apisix/blob/b024f683ef6f5310a180cdb6f792365e4c78f33a/apisix/plugins/skywalking.lua#L139C16-L139C33
* https://github.com/apache/skywalking-nginx-lua/blob/d15c0ce8ac10a903efc1379bf19cbe48641f331b/lib/skywalking/client.lua#L25C41-L25C41

leslie-tsang (approved):

monkeyDluffy6017 (approved):

PR #9663, opened 2023-06-14
fix: bugfix of limit-conn plugin

### LINKED ISSUE #9661: bug: limit-conn report error attribute does not exist
### Current Behavior

When adding stream_routes for the redis protocol, if the limit-conn plugin is used, an error will be reported

### Expected Behavior

created  route successfully

### Error Logs

2023/06/14 17:35:07 [error] 30109#30109: *773 failed to run log_by_lua*: /opt/linyao/apisix/apisix/plugins/limit-conn/init.lua:123: attempt to perform arithmetic on field 'request_time' (a nil value)
stack traceback:
	/opt/linyao/apisix/apisix/plugins/limit-conn/init.lua:123: in function 'phase_func'
	/opt/linyao/apisix/apisix/plugin.lua:1134: in function 'run_plugin'
	/opt/linyao/apisix/apisix/init.lua:1083: in function 'stream_log_phase'
	log_by_lua(nginx.conf:108):2: in main chunk while prereading client data, client: 127.0.0.1, server: 0.0.0.0:9100
2023/06/14 17:35:12 [error] 30113#30113: *2961 stream [lua] init.lua:106: phase_func(): decrease_to record the connection leaving request while prereading client data, client: 127.0.0.1, server: 0.0.0.0:9100
2023/06/14 17:35:12 [error] 30113#30113: *2961 failed to run log_by_lua*: /opt/linyao/apisix/apisix/plugins/limit-conn/init.lua:123: attempt to perform arithmetic on field 'request_time' (a nil value)
stack traceback:
	/opt/linyao/apisix/apisix/plugins/limit-conn/init.lua:123: in function 'phase_func'
	/opt/linyao/apisix/apisix/plugin.lua:1134: in function 'run_plugin'
	/opt/linyao/apisix/apisix/init.lua:1083: in function 'stream_log_phase'
	log_by_lua(nginx.conf:108):2: in main chunk while prereading client data, client: 127.0.0.1, server: 0.0.0.0:9100

### Steps to Reproduce

1. Tcp dynamic proxy uses redis protocol
2. Create a route with the limit-conn plugin as follow
    curl http://127.0.0.1:9180/apisix/admin/stream_routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
   "plugins": {
        "limit-conn": {
            "conn": 5,
            "burst": 0,
            "default_conn_delay": 0.1,
            "rejected_code": 503,
            "key_type": "var",
            "key": "remote_addr"
        }
    },
    "upstream": {
        "type": "none",
        "nodes": {
            "127.0.0.1:6379": 1
        }
    },
    "protocol": {
        "name": "redis"
    }
}
'
3. Check the error log and find the error message

### Environment

APISIX version (run apisix version): master latest
Operating system (run uname -a): centos7
OpenResty / Nginx version (run openresty -V or nginx -V): openresty/1.21.4.1
etcd version, if relevant (run curl http://127.0.0.1:9090/v1/server_info): 3.5.0
APISIX Dashboard version, if relevant:
Plugin runner version, for issues related to plugin runners:
LuaRocks version, for installation issues (run luarocks --version):

wxbty:
### Description
The tcp dynamic proxy protocol does not define the request_time attribute, we need to be compatible with this situation

Fixes #9661

### DIFF
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -109,40 +109,33 @@ function _M.decrease(conf, ctx)
         return
     end
 
-    if not is_http then
-        core.log.warn("The limit-conn plugin is not applicable in stream mode")
-        return
-    end
+    if is_http then
+        for i = 1, #limit_conn, 4 do
+            local lim = limit_conn[i]
+            local key = limit_conn[i + 1]
+            local delay = limit_conn[i + 2]
+            local use_delay = limit_conn[i + 3]
 
-    for i = 1, #limit_conn, 4 do
-        local lim = limit_conn[i]
-        local key = limit_conn[i + 1]
-        local delay = limit_conn[i + 2]
-        local use_delay =  limit_conn[i + 3]
-
-        local latency
-        if not use_delay then
-            if ctx.proxy_passed then
-                latency = ctx.var.upstream_response_time
-            else
-                if ctx.var.request_time ~= nil then
-                    latency = ctx.var.request_time - delay
-                else
+            local latency
+            if not use_delay then
+                if ctx.proxy_passed then
                     latency = ctx.var.upstream_response_time
+                else
+                    if ctx.var.request_time ~= nil then
+                        latency = ctx.var.request_time - delay
+                    else
+                        latency = ctx.var.upstream_response_time
+                    end
                 end
             end
-        end
-        core.log.debug("request latency is ", latency) -- for test
+            core.log.debug("request latency is ", latency) -- for test
 
-        if not latency then
-            return
-        end
-
-        local conn, err = lim:leaving(key, latency)
-        if not conn then
-            core.log.error("failed to record the connection leaving request: ",
-                           err)
-            break
+            local conn, err = lim:leaving(key, latency)
+            if not conn then
+                core.log.error("failed to record the connection leaving request: ",
+                        err)
+                break
+            end
         end
     end
 
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -16,22 +16,26 @@
 --
 local limit_conn_new = require("resty.limit.conn").new
 local core = require("apisix.core")
+local is_http = ngx.config.subsystem == "http"
 local sleep = core.sleep
 local shdict_name = "plugin-limit-conn"
 if ngx.config.subsystem == "stream" then
     shdict_name = shdict_name .. "-stream"
 end
 
+
 local lrucache = core.lrucache.new({
     type = "plugin",
 })
 local _M = {}
 
+
 local function create_limit_obj(conf)
     core.log.info("create new limit-conn plugin instance")
     return limit_conn_new(shdict_name, conf.conn, conf.burst,
-            conf.default_conn_delay)
+                          conf.default_conn_delay)
 end
+
 
 function _M.increase(conf, ctx)
     core.log.info("ver: ", ctx.conf_version)
@@ -98,13 +102,13 @@ function _M.increase(conf, ctx)
     end
 end
 
+
 function _M.decrease(conf, ctx)
     local limit_conn = ctx.limit_conn
     if not limit_conn then
         return
     end
 
-    local is_http = ctx.config.subsystem == "http"
     if not is_http then
         core.log.warn("The limit-conn plugin is not applicable in stream mode")
         return
@@ -114,7 +118,7 @@ function _M.decrease(conf, ctx)
         local lim = limit_conn[i]
         local key = limit_conn[i + 1]
         local delay = limit_conn[i + 2]
-        local use_delay = limit_conn[i + 3]
+        local use_delay =  limit_conn[i + 3]
 
         local latency
         if not use_delay then
@@ -137,7 +141,7 @@ function _M.decrease(conf, ctx)
         local conn, err = lim:leaving(key, latency)
         if not conn then
             core.log.error("failed to record the connection leaving request: ",
-                    err)
+                           err)
             break
         end
     end
@@ -146,5 +150,6 @@ function _M.decrease(conf, ctx)
     ctx.limit_conn = nil
     return
 end
+
 
 return _M
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -121,11 +121,7 @@ function _M.decrease(conf, ctx)
                 if ctx.proxy_passed then
                     latency = ctx.var.upstream_response_time
                 else
-                    if ctx.var.request_time ~= nil then
-                        latency = ctx.var.request_time - delay
-                    else
-                        latency = ctx.var.upstream_response_time
-                    end
+                    latency = ctx.var.request_time - delay
                 end
             end
             core.log.debug("request latency is ", latency) -- for test
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -119,7 +119,11 @@ function _M.decrease(conf, ctx)
             if ctx.proxy_passed then
                 latency = ctx.var.upstream_response_time
             else
-                latency = ctx.var.request_time - delay
+                if ctx.var.request_time ~= nil then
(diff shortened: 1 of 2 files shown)


monkeyDluffy6017:
Hi @wxbty, test cases are needed

wxbty:
> Hi @wxbty, test cases are needed

Because the log_by_lua_block stage is after returning to the client, it will be more difficult for us to use assertion verification in the unit test, which may be the reason why this bug has not been found.
I can only think of verifying through logs.
before modify:
image
after modify:
image

do you have a better suggestion?

moonming:
@monkeyDluffy6017 please review again

apisix/plugins/limit-conn/init.lua
```diff
@@ -119,7 +119,11 @@ function _M.decrease(conf, ctx)
             if ctx.proxy_passed then
                 latency = ctx.var.upstream_response_time
             else
-                latency = ctx.var.request_time - delay
+                if ctx.var.request_time ~= nil then
+                    latency = ctx.var.request_time - delay
+                else
+                    latency = ctx.var.upstream_response_time
```
monkeyDluffy6017, line 125: The stream module doesn't have the `upstream_response_time` variable too

monkeyDluffy6017:
@wxbty Just add a test case that will raise an error before your fix and will not raise an error after

apisix/plugins/limit-conn/init.lua
shreemaan-abhishek: Are you sure, I tested it locally and it worked 🤔

wxbty: These two variables can be nil, but the calculation will throw an exception. In fact, the limit-conn plug-in is invalid in stream mode. Can we return directly here?

wxbty:
> @wxbty Just add a test case that will raise an error before your fix and will not raise an error after

I'm an openresty beginner. I wrote a unit test but it didn't take effect. I don't know why it didn't enter the decrease method of the plugin. Can you help me to take a look? thanks very much 🤔

=== TEST 1: decrease
--- config
    location /t {
        content_by_lua_block {
            ngx.ctx.proxy_passed = false
            local plugin = require("apisix.plugins.limit-conn")
            local p = plugin.log(nil,ngx.ctx)  
            ngx.say("done")
        }
    }
--- request
GET /t
--- response_body
done

This unit test is located in stream-plugin/limit-conn.t, expecting to throw an exception, because its limit_conn log method calls decrease(conf, ctx), but now it passes, and does not print the additional warn log added by the decrease method

wxbty:
> @wxbty Just add a test case that will raise an error before your fix and will not raise an error after

@monkeyDluffy6017   PTAL

monkeyDluffy6017:
I think you could judge with `local is_http = ngx.config.subsystem == "http"`, when the subsystem is `stream`, no need to assign a value to latency.
I can construct test cases to reproduce your problem, you can try again

wxbty:
> I think you could judge with `local is_http = ngx.config.subsystem == "http"`, when the subsystem is `stream`, no need to assign a value to latency. I can construct test cases to reproduce your problem, you can try again

ok

wxbty:
> I think you could judge with `local is_http = ngx.config.subsystem == "http"`, when the subsystem is `stream`, no need to assign a value to latency. I can construct test cases to reproduce your problem, you can try again



> I think you could judge with `local is_http = ngx.config.subsystem == "http"`, when the subsystem is `stream`, no need to assign a value to latency. I can construct test cases to reproduce your problem, you can try again

@monkeyDluffy6017  Already modified according to this, is there any other problem now?

apisix/plugins/limit-conn/init.lua
```diff
@@ -22,20 +22,17 @@ if ngx.config.subsystem == "stream" then
     shdict_name = shdict_name .. "-stream"
 end
 
-
```
monkeyDluffy6017, line 25: Why do you remove this blank,  two blanks are needed between different functions

apisix/plugins/limit-conn/init.lua
```diff
@@ -22,20 +22,17 @@ if ngx.config.subsystem == "stream" then
...
 local lrucache = core.lrucache.new({
     type = "plugin",
 })
 local _M = {}
 
-
```
monkeyDluffy6017, line 31: Ditto

apisix/plugins/limit-conn/init.lua
```diff
@@ -22,20 +22,17 @@ if ngx.config.subsystem == "stream" then
...
 local function create_limit_obj(conf)
     core.log.info("create new limit-conn plugin instance")
     return limit_conn_new(shdict_name, conf.conn, conf.burst,
-                          conf.default_conn_delay)
+            conf.default_conn_delay)
```
monkeyDluffy6017, line 33: Why do you change this?

apisix/plugins/limit-conn/init.lua
```diff
@@ -101,33 +98,46 @@ function _M.increase(conf, ctx)
...
     local limit_conn = ctx.limit_conn
     if not limit_conn then
         return
     end
 
+    local is_http = ctx.config.subsystem == "http"
+    if not is_http then
+        core.log.warn("The limit-conn plugin is not applicable in stream mode")
+        return
+    end
```
monkeyDluffy6017, line 111: Why not put this on the top of the file, it could cache the result and no need to judge every time

wxbty: The modification is completed, thank you for your suggestion

monkeyDluffy6017:
I don't think what you have done is correct, the limit-conn plugin can be used in the stream mode,  we have test cases to cover this, you just need to fix the bug

wxbty:
> I don't think what you have done is correct, the limit-conn plugin can be used in the stream mode, we have test cases to cover this, you just need to fix the bug

At least the log phase is invalid, because the plugin does not define fields such as request_time, upstream_response_time, etc., and cannot calculate the delay. Or how about deleting this line of log?
core.log.warn("The limit-conn plugin is not applicable in stream mode")

wxbty:
> I don't think what you have done is correct, the limit-conn plugin can be used in the stream mode, we have test cases to cover this, you just need to fix the bug

@monkeyDluffy6017  PTAL again

Sn0rt:
LGTM

t/stream-plugin/limit-conn.t
```diff
@@ -334,3 +334,21 @@ GET /test_concurrency
 --- error_log
 The value of the configured key is empty, use client IP instead
 --- stream_enable
+
+
```
monkeyDluffy6017, line 338: Need three blanks here

Sn0rt:
@wxbty  Can I take over this PR fix?

wxbty:
> @wxbty Can I take over this PR fix?

ok

apisix/plugins/limit-conn/init.lua
```diff
@@ -108,27 +109,29 @@ function _M.decrease(conf, ctx)
...
-        local delay = limit_conn[i + 2]
-        local use_delay =  limit_conn[i + 3]
-
-        local latency
-        if not use_delay then
-            if ctx.proxy_passed then
-                latency = ctx.var.upstream_response_time
-            else
-                latency = ctx.var.request_time - delay
+    if is_http then
```
monkeyDluffy6017, line 112: You shouldn't pass the whole following code, the `lim:leaving` is needed to reduce the count, just pass the code below
```
                if ctx.proxy_passed then
                    latency = ctx.var.upstream_response_time
                else
                    latency = ctx.var.request_time - delay
                end
```

t/stream-plugin/limit-conn.t
```diff
@@ -334,3 +334,22 @@ GET /test_concurrency
...
+
+=== TEST 11: decrease
```
Sn0rt, line 340: Can I rewrite the test case? the test case looks not related to the reproducing step

wxbty: You can change it, thank you

monkeyDluffy6017:
It's fixed by https://github.com/apache/apisix/pull/9816

PR #9720, opened 2023-06-25
feat: mock plugin support adding headers

### LINKED ISSUE #9717: feat: plugin mocking support set_headers
### Description

In some conditions, we need to set some headers while mock the API

so, maybe the plugin `mocking` can support `response_headers`

```json
{

    "response_headers": {
       "key": "value",
       .....
    }
}
```

shreemaan-abhishek:
### Description

Fixes #9717

### Checklist

### DIFF
--- a/apisix/plugins/mocking.lua
+++ b/apisix/plugins/mocking.lua
@@ -50,7 +50,7 @@ local schema = {
         -- generate random response by json schema.
         response_schema = { type = "object" },
         with_mock_header = { type = "boolean", default = true },
-        headers = {
+        response_headers = {
             type = "object",
             minProperties = 1,
             patternProperties = {
@@ -227,8 +227,8 @@ function _M.access(conf, ctx)
         ngx.header["x-mock-by"] = "APISIX/" .. core.version.VERSION
     end
 
-    if conf.headers then
-        for key, value in pairs(conf.headers) do
+    if conf.response_headers then
+        for key, value in pairs(conf.response_headers) do
             core.log.warn(key, value, "dibag")
             core.response.add_header(key, value)
         end
--- a/apisix/plugins/mocking.lua
+++ b/apisix/plugins/mocking.lua
@@ -232,6 +232,7 @@ function _M.access(conf, ctx)
             core.response.add_header(key, value)
         end
     end
+
     if conf.delay > 0 then
         ngx.sleep(conf.delay)
     end
--- a/apisix/plugins/mocking.lua
+++ b/apisix/plugins/mocking.lua
@@ -49,7 +49,19 @@ local schema = {
         -- specify response json schema, if response_example is not nil, this conf will be ignore.
         -- generate random response by json schema.
         response_schema = { type = "object" },
-        with_mock_header = { type = "boolean", default = true }
+        with_mock_header = { type = "boolean", default = true },
+        headers = {
+            type = "object",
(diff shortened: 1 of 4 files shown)


shreemaan-abhishek:
Documentation changes are pending.

apisix/plugins/mocking.lua
```diff
@@ -215,6 +227,12 @@ function _M.access(conf, ctx)
         ngx.header["x-mock-by"] = "APISIX/" .. core.version.VERSION
     end
 
+    if conf.headers then
+        for key, value in pairs(conf.headers) do
+            core.log.warn(key, value, "dibag")
+            core.response.add_header(key, value)
```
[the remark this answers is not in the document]
shreemaan-abhishek: Using `core.response` is more efficient.

apisix/plugins/mocking.lua
```diff
@@ -215,6 +227,11 @@ function _M.access(conf, ctx)
         ngx.header["x-mock-by"] = "APISIX/" .. core.version.VERSION
     end
 
+    if conf.response_headers then
+        for key, value in pairs(conf.response_headers) do
+            core.response.add_header(key, value)
+        end
+    end
     if conf.delay > 0 then
```
monkeyDluffy6017, line 235: Need a blank

t/plugin/mocking.t
```diff
@@ -424,3 +424,44 @@ passed
...
+               local t = require("lib.test_admin").test
+               local code, body = t('/apisix/admin/routes/1',
+                    ngx.HTTP_PUT,
+                    [[{
+                           "plugins": {
+                               "mocking": {
+                                   "response_example": "hello world",
+                                   "response_headers": {
+                                        "X-Apisix": "is-cool",
+                                        "X-Really":"yes"
```
monkeyDluffy6017, line 443: What happens if you set 2 identical response headers?  Add a test case

monkeyDluffy6017:
The doc should be updated

t/plugin/mocking.t
shreemaan-abhishek: What do you mean by "identical response headers"?

monkeyDluffy6017:
```
                                   "response_headers": {
                                        "X-Apisix": "is-cool",
                                        "X-Apisix": "yes"
                                    }
                               }
```

shreemaan-abhishek: 1. Only the last one is retained, the first one is lost. However, if someone needs multiple values for a header they can do `"header-name": "val1, val2"`.

2. Why is this test case needed?

monkeyDluffy6017: You could test, I don't think you are right.

monkeyDluffy6017: I said this because you use the function `core.response.add_header` to add the response header, the function will append the value to the same header, but i tested locally, and found that the `conf.response_headers` won't pass the duplicated headers. 
If we pass 
```
                                   "response_headers": {
                                        "X-Apisix": "is-cool",
                                        "X-Apisix": "yes"
                                    }
```
we will get
```
                                   "response_headers": {
                                        "X-Apisix": "yes"
                                    }
```
So it's ok for this.
But i don't see you example

monkeyDluffy6017 (approved):

moonming (approved):

Issue #9761, opened 2023-06-30
help request: consumer-restriction配置consumer_group_id后api返回"Missing authentication or identity verification."


gzlboy:
### Description

使用consumer-restriction+basic-auth插件，consumer-restriction 配置type为consumer_name时接口返回符合预期，但当把consumer-restriction的type改成consumer_group_id后(设置了blacklist)，不是黑名单内的consumer访问api，就会返回：
{"message":"Missing authentication or identity verification."}
如果使用黑名单内的consumer登录，返回是正常的：
{"message":"The consumer_group_id is forbidden."}

请教是什么原因呀？

`{
    "plugins": {
        "consumer-restriction": {
            "type": "consumer_group_id",
            "blacklist": [
                "web1"
            ]
        }
    }
}`
注：web1为group_id

apisix版本：3.3.0

### Environment

- APISIX version (run `apisix version`):
- Operating system (run `uname -a`):
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

Revolyssup:
Can you set the log level to info and search for the log starting with `consumer: ` when you send request for a non blacklist consumer to confirm which consumer is selected?
Also please provide the consumer configuration as well and the requests you are sending.

gzlboy:
dataPlane logs:
![image](https://github.com/apache/apisix/assets/1492060/8b16295b-e5e6-47d1-a91a-a346b6b94b3b)

consumer jack:
![image](https://github.com/apache/apisix/assets/1492060/29eaeddc-2a9a-4b22-a26d-36edaf62c7e0)

consumer nginxUser1:
![image](https://github.com/apache/apisix/assets/1492060/29196aee-1103-4a69-b926-6581aeb273d9)

router config:
`{
    "createdIndex": 72,
    "value": {
        "methods": [
            "GET",
            "POST",
            "PUT",
            "DELETE",
            "PATCH",
            "HEAD",
            "OPTIONS",
            "CONNECT",
            "TRACE",
            "PURGE"
        ],
        "upstream": {
            "nodes": {
                "nginx-service:80": 1
            },
            "type": "roundrobin",
            "timeout": {
                "send": 6,
                "read": 6,
                "connect": 6
            },
[...]

PR #9778, opened 2023-07-04
refactor(consumer-restriction): optimize the error log when the corresponding resource is not fetched

jiangfucheng:
### Description

Fixes  https://github.com/apache/apisix/issues/9761 and fix some test cases

### Checklist

### DIFF
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -128,7 +128,13 @@ function _M.access(conf, ctx)
     local method = ngx.req.get_method()
 
     if not value then
-        return 401, { message = "Missing authentication or identity verification."}
+        local err_msg
+        if conf.type == "consumer_name" then
+            err_msg = "Missing authentication in request."
+        else
+            err_msg = "Please ensure " .. conf.type .. " is present in request"
+        end
+        return 401, { message = err_msg}
     end
     core.log.info("value: ", value)
 
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -128,12 +128,8 @@ function _M.access(conf, ctx)
     local method = ngx.req.get_method()
 
     if not value then
-        local err_msg
-        if conf.type == "consumer_name" then
-            err_msg = "Missing authentication information in request."
-        else
-            err_msg = "Please ensure " .. conf.type .. " is present in request"
-        end
+        local err_msg = "The request is rejected, please check the "
+                        .. conf.type .. " for this request"
         return 401, { message = err_msg}
     end
     core.log.info("value: ", value)
--- a/apisix/plugins/consumer-restriction.lua
+++ b/apisix/plugins/consumer-restriction.lua
@@ -130,7 +130,7 @@ function _M.access(conf, ctx)
     if not value then
         local err_msg
         if conf.type == "consumer_name" then
-            err_msg = "Missing authentication in request."
+            err_msg = "Missing authentication information in request."
         else
             err_msg = "Please ensure " .. conf.type .. " is present in request"
         end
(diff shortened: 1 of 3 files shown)


apisix/plugins/consumer-restriction.lua
```diff
@@ -95,6 +99,10 @@ local function is_include(value, tab)
 end
 
 local function is_method_allowed(allowed_methods, method, user)
+    if not user then
+        return false
+    end
```
monkeyDluffy6017, line 104: I don't think the logic here is correct, if the blacklist exists and we can't get the user info, the request should be forbidden

apisix/plugins/consumer-restriction.lua
```diff
@@ -86,7 +86,11 @@ local fetch_val_funcs = {
 }
 
 local function is_include(value, tab)
-    for k,v in ipairs(tab) do
+    if not value then
+        return false
+    end
```
monkeyDluffy6017, line 91: I don't think the logic here is correct, if the blacklist exists and we can't get the user info, the request should be forbidden

jiangfucheng: 1.For `consumer`, if have no corresponding `consumer_id`, we can assume that the user is unauthorized, it is reasonable for us to reject this request.
2.For `service` and `consumer_group`, they are not necessary, it will make user confused if we reject this request directly when the `service_id` or `consumer_group_id` not exists.

monkeyDluffy6017: I think reject is acceptable, we couldn't let the request go if it can't be verified.
We only need to optimize the error log in the issue.

monkeyDluffy6017: You are right, but it's too complicated to check every scenario, let's make the logic simple, the user who use the plugin should make the assigned type is present in every request.
So i think it's ok to optimize the related error log
```
Missing authentication or identity verification
```

jiangfucheng: Ok, I will update it.

apisix/plugins/consumer-restriction.lua
```diff
@@ -128,7 +128,13 @@ function _M.access(conf, ctx)
     local method = ngx.req.get_method()
 
     if not value then
-        return 401, { message = "Missing authentication or identity verification."}
+        local err_msg
+        if conf.type == "consumer_name" then
+            err_msg = "Missing authentication in request."
```
monkeyDluffy6017, line 133: Is this better?
```suggestion
            err_msg = "The request is missing authentication information."
```

apisix/plugins/consumer-restriction.lua
```diff
@@ -128,7 +128,13 @@ function _M.access(conf, ctx)
     local method = ngx.req.get_method()
 
     if not value then
-        return 401, { message = "Missing authentication or identity verification."}
+        local err_msg
+        if conf.type == "consumer_name" then
+            err_msg = "Missing authentication information in request."
+        else
+            err_msg = "Please ensure " .. conf.type .. " is present in request"
+        end
```
monkeyDluffy6017, line 136: Can the two message be unifed into this form?
"The request is rejected becase the " .. conf.type .. " is missing"

jiangfucheng: The request be rejected when `config.type` is `consumer_name` maybe cause by that the user authentication info not match any `consumer`, but not missing configuration of `consumer`. So, prompt user configuration is missing maybe not very suitable.

monkeyDluffy6017: But the msg "Please ensure service_id is present in request" is not clear too, because the `service_id` is not configured in the request.

monkeyDluffy6017: How about "The request is rejected, please check the "  .. conf.type .. " for this request"

jiangfucheng: I think it's ok, thank you for your patient suggestion.

monkeyDluffy6017 (approved):

leslie-tsang (approved):

PR #9044, opened 2023-03-09
fix: check upstream reference in traffic-split plugin when delete upstream

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/5339

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #9784, opened 2023-07-05
refactor: move global_rules to separate file

jiangfucheng:
### Description

ref: https://github.com/apache/apisix/pull/9044#discussion_r1133063086

### Checklist


monkeyDluffy6017:
I will check this later

apisix/global_rules.lua
```diff
@@ -0,0 +1,56 @@
...
+            .. err)
+    end
+end
+
+
+function _M.global_rules()
+    if not global_rules then
+        return nil, nil
+    end
+    return global_rules.values, global_rules.conf_version
```
monkeyDluffy6017, line 45: Why don't you just return the `global_rules`, then less code needs to be modified

jiangfucheng: I want keep consistent code style with other resources, It allows the caller to use same logic to process all resources when we needed it. (like upstream `delete_checker()` method

monkeyDluffy6017 (approved):

leslie-tsang (approved):

PR #9811, opened 2023-07-11
fix: update_count is reset once updated, cause cache key conflict 

kingluo:
### Description

update_count is reset once updated, which will cause lru cache key conflict, in turn reusing the old route conf.

### Checklist

### DIFF
--- a/apisix/plugin_config.lua
+++ b/apisix/plugin_config.lua
@@ -78,8 +78,7 @@ function _M.merge(route_conf, plugin_config)
         end
     end
 
-    route_conf.update_count = route_conf.update_count + 1
-    route_conf.modifiedIndex = route_conf.orig_modifiedIndex .. "#" .. route_conf.update_count
+    route_conf.modifiedIndex = route_conf.orig_modifiedIndex .. "#" .. plugin_config.modifiedIndex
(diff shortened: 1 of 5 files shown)


monkeyDluffy6017:
No test cases?

moonming (commented):
1. The title of PR needs to be able to see what problem is fixed
2. Did not understand the relevant modification of the test case

moonming:
// As a matter of fact, this bug cannot be triggered in the master branch, so that's why there is no dedicated test case.

So we don't need this PR?

kingluo:
> // As a matter of fact, this bug cannot be triggered in the master branch, so that's why there is no dedicated test case.
> 
> So we don't need this PR?

Yes, and no.

I think it's necessary, As I said before, the original code should not use a volatile counter instead of the etcd revision, which is definitely a bad practice.
But on the other hand, regarding the lack of test cases, yes, this bug is also useless for the master branch.

moonming:
The master branch does not need this fix, and is not sufficiently tested.

monkeyDluffy6017:
@moonming Any more comments?

moonming (approved):

soulbird:
fixed #9822

PR #9841, opened 2023-07-15
change(ua-restriction): allowlist and denylist can't be enabled at the same time

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/7467

### Checklist

### DIFF
--- a/apisix/plugins/ua-restriction.lua
+++ b/apisix/plugins/ua-restriction.lua
@@ -83,7 +83,7 @@ local function check_with_allow_list(user_agents, allowlist)
     end
 
     if type(user_agents) == "table" then
-        for _, v in ipairs(allowlist) do
+        for _, v in ipairs(user_agents) do
             if lrucache_allow(v, allowlist, check, v) then
                 return true
             end
@@ -101,21 +101,21 @@ local function check_with_deny_list(user_agents, denylist)
 
         for _, rule in ipairs(denylist) do
             if re_find(user_agent, rule, "jo") then
-                return  false
+                return false
             end
         end
         return true
     end
 
     if type(user_agents) == "table" then
-        for _, v in ipairs(denylist) do
-            if lrucache_allow(v, denylist, check, v) then
+        for _, v in ipairs(user_agents) do
+            if lrucache_deny(v, denylist, check, v) then
                 return false
             end
         end
         return true
     else
-        return lrucache_allow(user_agents, denylist, check, user_agents)
+        return lrucache_deny(user_agents, denylist, check, user_agents)
     end
 end
 
--- a/apisix/plugins/ua-restriction.lua
+++ b/apisix/plugins/ua-restriction.lua
@@ -54,6 +54,22 @@ local schema = {
             default = "Not allowed"
         },
     },
+    oneOf = {
+        {
+            required = {"allowlist"}
+        },
+        {
+            required = {"denylist"}
+        },
+        {
+            ["not"] = {
+                anyOf = {
+                    {required = {"allowlist"}},
+                    {required = {"denylist"}}
+                }
+            }
+        }
+    }
 }
 
 local plugin_name = "ua-restriction"
@@ -94,10 +110,6 @@ function _M.check_schema(conf)
 
     if not ok then
         return false, err
-    end
-
-    if conf.allowlist and conf.denylist then
-        return false, "allowlist and denylist can't be enabled at the same time."
     end
 
     if conf.allowlist then
--- a/apisix/plugins/ua-restriction.lua
+++ b/apisix/plugins/ua-restriction.lua
@@ -22,10 +22,6 @@ local type = type
 local str_strip = stringx.strip
 local re_find = ngx.re.find
 
-local MATCH_NONE = 0
-local MATCH_ALLOW = 1
-local MATCH_DENY = 2
-
 local lrucache_useragent = core.lrucache.new({ ttl = 300, count = 4096 })
 
 local schema = {
@@ -74,20 +70,22 @@ local function match_user_agent(user_agent, conf)
     if conf.allowlist then
         for _, rule in ipairs(conf.allowlist) do
             if re_find(user_agent, rule, "jo") then
-                return MATCH_ALLOW
+                return true
             end
         end
+        return false
     end
 
     if conf.denylist then
         for _, rule in ipairs(conf.denylist) do
             if re_find(user_agent, rule, "jo") then
-                return MATCH_DENY
+                return false
             end
         end
+        return true
     end
 
-    return MATCH_NONE
+    return true
 end
 
 function _M.check_schema(conf)
@@ -95,6 +93,10 @@ function _M.check_schema(conf)
 
     if not ok then
         return false, err
+    end
+
+    if conf.allowlist and conf.denylist then
+        return false, "allowlist and denylist can't be enabled at the same time."
     end
 
     if conf.allowlist then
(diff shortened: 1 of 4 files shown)


apisix/plugins/ua-restriction.lua
```diff
@@ -97,6 +95,10 @@ function _M.check_schema(conf)
         return false, err
     end
 
+    if conf.allowlist and conf.denylist then
+        return false, "allowlist and denylist can't be enabled at the same time."
+    end
```
jiangfucheng, line 100: I tried using `oneOf` field of jsonschema to validate configuration, but it's seems can't meet us requirement(Perhaps there is a problem with my usage). So, use code to check it directly.

monkeyDluffy6017: oneOf = {
        {required = {"allowlist"}},
        {required = {"denylist"}},
    },

apisix/plugins/ua-restriction.lua
```diff
@@ -128,12 +130,13 @@ function _M.access(conf, ctx)
...
     end
-    local match = MATCH_NONE
+    local match
+
     if type(user_agent) == "table" then
         for _, v in ipairs(user_agent) do
             if type(v) == "string" then
                 match = lrucache_useragent(v, conf, match_user_agent, v, conf)
-                if match > MATCH_ALLOW then
+                if not match then
```
monkeyDluffy6017, line 139: The logic should be:
1. If one of the user agent match the allowlist,  pass, else if it doesn't match, continue, if the whole user agent
don't match the allowlist, deny.
3. If one of the user agent match the denylist, deny, else if it doesn't match, continue, if the whole user agent 
don't match the denylist, pass.

jiangfucheng: My mistake. Fixed.

apisix/plugins/ua-restriction.lua
jiangfucheng: Both `allowlist` and `denylist` can be disabled simultaneously, so I added an additional schema. Thanks.

apisix/plugins/ua-restriction.lua
```diff
@@ -58,6 +54,22 @@ local schema = {
...
+            required = {"denylist"}
+        },
+        {
+            ["not"] = {
+                anyOf = {
+                    {required = {"allowlist"}},
+                    {required = {"denylist"}}
+                }
+            }
+        }
```
monkeyDluffy6017, line 71: Why need this?
If both the allowlist and the denylist don't exist, the plugin seems meanless

jiangfucheng: If both allowlist and denylist don't exist, it only checks if the `User-Agent` header exists.
image

monkeyDluffy6017: No, the `bypass_missing` is designed for the scenario that one of the `allowlist` and `denylist` exists, but the `User-Agent` is missing

apisix/plugins/ua-restriction.lua
```diff
@@ -128,12 +144,13 @@ function _M.access(conf, ctx)
             return 403, { message = conf.message }
         end
     end
-    local match = MATCH_NONE
+    local match
```
monkeyDluffy6017, line 147: Need a blank before this line

apisix/plugins/ua-restriction.lua
```diff
@@ -128,12 +144,13 @@ function _M.access(conf, ctx)
...
+
     if type(user_agent) == "table" then
         for _, v in ipairs(user_agent) do
             if type(v) == "string" then
                 match = lrucache_useragent(v, conf, match_user_agent, v, conf)
-                if match > MATCH_ALLOW then
+                if (conf.denylist and not match) or (conf.allowlist and match) then
```
monkeyDluffy6017, line 153: The logic is very complicated, Why not add a check function for each of denylist and allowlist?
The function only returns `local PASS = 0` and `local DENY = 1`

apisix/plugins/ua-restriction.lua
jiangfucheng: I'm saying this because of these test cases:
https://github.com/apache/apisix/blob/86004d35d551cddc50d49de23b70f7e62248d68a/t/plugin/ua-restriction.t#L416-L498

I thought it was designed that way. If we don't need this function, I will change it.

monkeyDluffy6017: I don't know why we configure like this, no allowlist and no denylist, it is meanless
```
                        "plugins": {
                            "ua-restriction": {
                            }
                        }
```

Sn0rt:
@jiangfucheng  thx your PR, pls take a look at the comments from monkeyDluffy6017

apisix/plugins/ua-restriction.lua
jiangfucheng: Fixed.

apisix/plugins/ua-restriction.lua
jiangfucheng: Updated.

apisix/plugins/ua-restriction.lua
```diff
@@ -81,27 +70,53 @@ local _M = {
...
                 return true
             end
         end
         return false
     end
 
-    if conf.denylist then
-        for _, rule in ipairs(conf.denylist) do
+    if type(user_agents) == "table" then
+        for _, v in ipairs(allowlist) do
```
monkeyDluffy6017, line 86:
```suggestion
        for _, v in ipairs(user_agents) do
```

t/plugin/ua-restriction.t
```diff
@@ -340,136 +340,7 @@ User-Agent:foo/bar
 hello world
 
 
-
-=== TEST 18: bypass_missing test, using default, reset conf(part1)
```
monkeyDluffy6017, line 374: The `bypass_missing` should work when allowlist or denylist exists and no user agent, the test case is needed

apisix/plugins/ua-restriction.lua
```diff
@@ -81,27 +70,53 @@ local _M = {
...
+
+
+local function check_with_deny_list(user_agents, denylist)
+    local check = function (user_agent)
+        user_agent = str_strip(user_agent)
+
+        for _, rule in ipairs(denylist) do
             if re_find(user_agent, rule, "jo") then
-                return false
+                return  false
```
monkeyDluffy6017, line 104:
```suggestion
                return false
```

apisix/plugins/ua-restriction.lua
```diff
@@ -81,27 +70,53 @@ local _M = {
...
             end
         end
         return true
     end
 
-    return true
+    if type(user_agents) == "table" then
+        for _, v in ipairs(denylist) do
```
monkeyDluffy6017, line 111:
```suggestion
        for _, v in ipairs(user_agents) do
```

apisix/plugins/ua-restriction.lua
monkeyDluffy6017: Why does this pass the test cases?

t/plugin/ua-restriction.t
jiangfucheng: Updated.

apisix/plugins/ua-restriction.lua
jiangfucheng: Stupid mistake.

monkeyDluffy6017 (approved):

leslie-tsang (approved):

PR #9895, opened 2023-07-25
docs: update certificate names in decoupled mode and add configurations for different deployment modes to `config-default.yaml`

### LINKED ISSUE #9883: docs: all fields for certs in decoupled deployment modes use `/path/to/ca-cert`
### Current State

All fields for certs in decoupled deployment modes use `/path/to/ca-cert` as placeholders.

https://apisix.apache.org/docs/apisix/next/deployment-modes/#decoupled

For example:
image

This is inaccurate. Other than CA certificate, there are fields for server/client certificate and key.

### Desired State

Update the fields for certificates that are currently all `/path/to/ca-cert` to be more accurate.

kayx23:
### Description

* Update certificate names in decoupled mode
* Add configuration blocks of `role_data_plane` and `role_control_plane` in `config-default.yaml` to better document for different modes of APISIX deployment.

Fixes #9883 

### Checklist

### DIFF
--- a/conf/config-default.yaml
+++ b/conf/config-default.yaml
@@ -611,9 +611,34 @@ plugin_attr:          # Plugin attributes
                                                               # write access to this file for security.
 
 deployment:                    # Deployment configurations
-  role: traditional            # Set deployment mode: traditional, control_plane, data_plane.
+  role: traditional            # Set deployment mode: traditional, control_plane, or data_plane.
   role_traditional:
     config_provider: etcd      # Set the configuration center.
+
+    # role_data_plane:         # Set data plane details if role is data_plane.
+    #   config_provider: control_plane          # Set the configuration center: control_plane, or yaml.
+    #   control_plane:                          # Set control plane details if config_provider is control_plane.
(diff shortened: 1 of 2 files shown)


moonming:
Do we need to update https://github.com/apache/apisix/blob/master/conf/config-default.yaml too? I do not find these configures in this file.

kayx23:
It looks to me that `config-default.yaml` includes only the configurations of **traditional mode** by default.

It doesn't include the configurations of **decoupled mode**, so you won't find these configurations in there.

It probably doesn't make sense to include both (even with one commented out) in that file because these deployment modes and configurations are conceptually mutually exclusive.

moonming:
> It doesn't include the configurations of **decoupled mode**, so you won't find these configurations in there.

if so how do developers know decoupled mode configurations?

kayx23:
Added configurations to `config-default.yaml`

monkeyDluffy6017 (approved):

moonming (approved):

Issue #9802, opened 2023-07-07
bug: APISIX executes reload, but the old process does not exit for a long time


hansedong:
### Current Behavior

As APISIX executes reload, but the old process does not exit for a long time.

```
root      885278       1  0 Jul05 ?        00:00:00 nginx: master process /usr/bin/openresty -p /usr/local/apisix -c /usr/local/apisix/conf/nginx.conf
nobody    885280  885278  0 Jul05 ?        00:09:03 nginx: worker process is shutting down
nobody    885281  885278  0 Jul05 ?        00:08:39 nginx: worker process is shutting down
nobody    885282  885278  0 Jul05 ?        00:08:24 nginx: worker process is shutting down
nobody   1288295  885278  0 18:09 ?        00:00:03 nginx: worker process is shutting down
nobody   1288517  885278  5 18:10 ?        00:03:10 nginx: worker process
nobody   1288518  885278  4 18:10 ?        00:02:16 nginx: worker process
nobody   1288519  885278  4 18:10 ?        00:02:15 nginx: worker process
nobody   1288520  885278  4 18:10 ?        00:02:12 nginx: worker process
nobody   1288521  885278  0 18:10 ?        00:00:01 nginx: cache manager process
root     1288522  885278  3 18:10 ?        00:01:41 nginx: privileged agent process
root     1296427 1280594  0 19:04 pts/0    00:00:00 grep --color=auto nginx
```

From above, it can be seen that the creation time of the new process is 18:10, but it is already 19:05 now

```
[root@knode10-72-73-177 logs]# date
Fri Jul  7 19:05:05 CST 2023
```

The value of the configuration item "worker_shutdown_timeout" in the APISIX configuration file is set to 240 seconds.

```
nginx_config:                     # config for render the template to generate nginx.conf
  #user: root                     # specifies the execution user of the worker process.
                                  # the "user" directive makes sense only if the master process runs with super-user privileges.
                                  # if you're not root user,the default is current user.
  error_log: /DATA1/apisix/apisix/logs/error.log
  error_log_level:  warn          # warn,error
  worker_processes: 4             # if you want use multiple cores in container, you can inject the number of cpu as environment variable "APISIX_WORKER_PROCESSES"
  enable_cpu_affinity: false      # disable CPU affinity by default, if APISIX is deployed on a physical machine, it can be enabled and work well.
  worker_rlimit_nofile: 6553500     # the number of files a worker process can open, should be larger than worker_connections
[...]

PR #9909, opened 2023-07-26
fix: worker not exited when executing quit or reload command

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/9802

### Checklist


monkeyDluffy6017:
How long does the long connection prevent the quit operation?

kingluo:
@jiangfucheng Please show your steps to reproduce this issue.

jiangfucheng:
> How long does the long connection prevent the quit operation?

Accoding my test, the worker will never exit.

jiangfucheng:
> @jiangfucheng Please show your steps to reproduce this issue.

1. make run
2. make reload or make quit
3. ` ps -ef |grep nginx` check the worker if exited

apisix/core/config_etcd.lua
```diff
@@ -188,9 +190,43 @@ local function run_watch(premature)
...
+            end)
+
+            local check_worker_th = ngx_thread_spawn(function ()
+                while not exiting() do
+                    ngx_sleep(0.1)
+                end
+                sema:post()
+            end)
+
+            local _, sem_err = sema:wait(opts.timeout + 10)
```
kingluo, line 215: Could we use ngx.thread.wait instead of sema?

apisix/core/config_etcd.lua
```diff
@@ -188,9 +190,43 @@ local function run_watch(premature)
...
-            if log_level >= NGX_INFO then
-                log.info("res_func: ", inspect(res))
+            local res
+            local sema, err = semaphore.new()
+            if not sema then
+                log.error("create sema failed", err)
+                goto watch_event
+            end
+
+            local get_res_th = ngx_thread_spawn(function ()
```
kingluo, line 200: @leslie-tsang 
I'm concerned that each watch will spawn two threads, which will affect performance.
What if we check exiting() after the watch has timed out? It's currently 50 seconds, maybe we should lower it?

apisix/core/config_etcd.lua
jiangfucheng: Great suggestion, I didn't notice this method before, I will update it.

apisix/core/config_etcd.lua
monkeyDluffy6017: How long do you think is appropriate? 1s?

apisix/core/config_etcd.lua
```diff
@@ -188,9 +191,36 @@ local function run_watch(premature)
 
         ::watch_event::
         while true do
-            local res, err = res_func()
-            if log_level >= NGX_INFO then
-                log.info("res_func: ", inspect(res))
+            local res
+
+            local get_res_th = ngx_thread_spawn(function ()
```
kingluo, line 196: The code is fine, but there is a slight impact on performance, because each res_func() will spawn a new thread, and this thread will exit after 50 seconds.
So, is it feasible to run the whole run_watch() in one thread and watch for the exit signal in another thread?

apisix/core/config_etcd.lua
kingluo: Forget that, it doesn't seem like a good point.

apisix/core/config_etcd.lua
jiangfucheng: Good idea. I have a question, can we use `ngx_thread_spawn` instead of `ngx_timer_at`? we need kill the `run_watch` thread when worker exited.
https://github.com/apache/apisix/blob/c89ec15e85eded367bdc5e03fa61ac5f6e53792d/apisix/core/config_etcd.lua#L269

kingluo:
@jiangfucheng 
The code is ok.
Please check the failed ci: t/node/healthcheck-stop-checker.t.

kingluo (approved):

apisix/core/config_etcd.lua
```diff
@@ -257,6 +260,33 @@ local function run_watch(premature)
...
+    local ok, err = ngx_thread_wait(check_worker_th)
+
+    if not ok then
+        log.error("wait worker checker failed, retart checker, error: " .. err)
+        ngx_thread_kill(check_worker_th)
+        goto restart
+    end
+
+    -- waiting for do_run_watch thread to handle the final etcd event
+    ngx_sleep(1)
```
monkeyDluffy6017, line 282: How do you know no more etcd events will come in the last 1 second?
I think you should check ngx.exiting in `do_run_watch` too

apisix/core/config_etcd.lua
```diff
@@ -257,6 +260,33 @@ local function run_watch(premature)
...
+    local run_watch_th = ngx_thread_spawn(do_run_watch, premature)
+
+    ::restart::
+    local check_worker_th = ngx_thread_spawn(function ()
+        while not exiting() do
+            ngx_sleep(0.1)
+        end
+    end)
+
+    local ok, err = ngx_thread_wait(check_worker_th)
```
monkeyDluffy6017, line 273: I think you should monitor the `run_watch_th` too in case that the `run_watch_th` fails

apisix/core/config_etcd.lua
jiangfucheng: After we execute `make quit` command, the admin API will not be available, so before the admin API becomes unavailable, we can use`sleep(1)` to wait for the `do_run_watch` thread to handle the last possible event. Even if no events come, we also will kill the `do_run_watch` thread. 
And, check `ngx.exiting` in `do_run_watch` it means we need to wait for the `res_func()` function timeout,  that's why I create two threads to handle it.

apisix/core/config_etcd.lua
jiangfucheng: The `do_run_watch` function will not return, it has an internal retry mechanism.  The origin logic did not handle this so I didn't do this either.

apisix/core/config_etcd.lua
monkeyDluffy6017: The DP don't have admin API, it use etcd to sync the configuration from the CP which has admin API, when you execute `make quit` on DP, the CP is available.
So you should stop syncing after you execute `make quit` on DP

monkeyDluffy6017: By the way, do we really need to do this before quit? we will load full data after reload

jiangfucheng: Yes, it is not necessary. `t/node/healthcheck-stop-checker.t.` this test case failed seems related to this,  so I add the logic.

apisix/core/config_etcd.lua
monkeyDluffy6017: We should handle the situation that the `run_watch_th` thread crashes

apisix/core/config_etcd.lua
jiangfucheng: I will try it.

apisix/core/config_etcd.lua
jiangfucheng: Do we need restart `run_watch_th` when it fails?

monkeyDluffy6017: I think the code may like this
```
   ::restart::
    local run_watch_th = ngx_thread_spawn(do_run_watch, premature)
    local check_worker_th = ngx_thread_spawn(function ()
        while not exiting() do
            ngx_sleep(0.1)
        end
    end)

    local ok, err = ngx_thread_wait(run_watch_th, check_worker_th)
    if not ok then
        log.error("wait worker checker failed, retart checker, error: " .. err)
        ngx_thread_kill(run_watch_th)
        ngx_thread_kill(check_worker_th)
        goto restart
    end
```

t/node/healthcheck-stop-checker.t
```diff
@@ -256,7 +247,6 @@ ok
 --- grep_error_log eval
 qr/create new checker: table: 0x|try to release checker: table: 0x/
 --- grep_error_log_out
-try to release checker: table: 0x
```
jiangfucheng, line 259: This line can be remove, the reasons are as follows:

1.`healthchekcer` will be remove after worker exit since the status store in memroy, that's why `TEST 5` will not print `try to release checker: table: 0x` in first line.
2.Why the `try to release checker: table: 0x` will be print if we add `sleep(1)` before worker exited
    - Because the log is printed by old worker. Before old worker be killed, the new worker will be created, at this moment, both old worker and new worker can receive events, it's easy to be proved through print debug log before execute `fire_all_clean_handlers`
   
```
            if pre_index then
            local pre_val = self.values[pre_index]
            log.info("sync_data: check pre_val: ", inspect(pre_val.clean_handlers), worker_id_str, " pid: ", ngx.worker.pid())
            if pre_val then
                config_util.fire_all_clean_handlers(pre_val)
            end
```
 logs:
 
```
2023/08/08 18:04:31 [info] 77582#1004031: *312 [lua] config_etcd.lua:741: sync_data(): sync_data: check pre_val: { {
f = <function 1>,
id = 1
},
  _id = 2
} worker_id: 0 pid: 77582, context: ngx.timer

2023/08/08 18:04:31 [info] 77620#1004305: *447 [lua] config_etcd.lua:741: sync_data(): sync_data: check pre_val: {} worker_id: 0 pid: 77620, context: ngx.timer
```

We can see the etcd events be received with two `wokrer 0`, and there pid is different, we can easliy to confirm the worker is old worker which has `check_handlers` field

3.Why these test cases can passed in old version(before https://github.com/apache/apisix/issues/9456 be merged)
    - Because in the old version, the worker is not exit immediately too, it will exit after `make quit/reload` about 60s, so the reason is same as above.

monkeyDluffy6017 (dismissed):

apisix/core/config_etcd.lua
```diff
@@ -257,6 +260,30 @@ local function run_watch(premature)
 end
 
 
+local function run_watch(premature)
+    ::restart::
```
kingluo, line 264: **No! The watch routine must not be restarted!**
Because it's a stateful routine (e.g. last watch starts revision) and should start once and notify all child watchers that it started. Instead, you just wait for check_worker_th, if it exits, it means the worker process is exiting, then kill run_watch_th, and that's it.

kingluo (changes requested):

apisix/core/config_etcd.lua
kingluo: > What if the thread `do_run_watch` crashes?

It's not supposed to crash.
In fact, almost all timers did not have a crash guard.
And even if run_watch crashes, you cannot fix it by restarting it, 'cause it's a stateful routine.

kingluo: Frankly, restarting it in this scenario is an over-design.

monkeyDluffy6017: Agreed with @kingluo

jiangfucheng: Updated, thank you for your review.❤️

jiangfucheng:
@monkeyDluffy6017 @kingluo Please take a look again, thanks.❤️

moonming:
@kingluo please take a look again

kingluo (approved):

PR #9927, opened 2023-07-29
chore(ci): automate management of unresponded issues

shreemaan-abhishek:
### Description

Currently, we add the `wait for update` label to issues that require a response from reporters. Often, unresponded issues remain unclosed leading to clutter. Maintainers have to manually go through such issues and close them one by one which is very inefficient.

This PR will automate the "_closing_" of unresponded issues by adding a label `no response` two weeks after the day when the `wait for update` label was added. Also, the `wait for update` label will also be automatically removed when the reporter/issue author responds.

### Checklist


.github/workflows/remove-wait-for-update-label.yml
monkeyDluffy6017, line 1: I think you should rename this file?

shreemaan-abhishek: done

.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,18 @@
+name: Update label when user responds
+permissions:
+  issues: write
+
+on:
+  issue_comment:
+    types: [created]
```
monkeyDluffy6017, line 7: Do we need to trigger this every comment? how about just like the `close-unresponded.yml`?

.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  run-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove label when user responds
```
monkeyDluffy6017, line 13:
```suggestion
      - name: Rename the label when user responds
```

.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,18 @@
...
+        uses: andymckay/labeler@master
+        if: ${{ github.event.comment.user.login == github.event.issue.user.login && contains(github.event.issue.labels.*.name, 'wait for update') && !contains(github.event.issue.labels.*.name, 'user responded') }}
```
monkeyDluffy6017, line 15: What happends if the user respond an issue with the label "no response"?
Should we change the label "no response" to "user responded"?

.github/workflows/update-labels.yml
shreemaan-abhishek: Running this action on `issue_comment` event is important because the event details are needed on the conditional statement on line 15 (currently).

```yaml
if: ${{ github.event.comment.user.login == github.event.issue.user.login && contains(github.event.issue.labels.*.name, 'wait for update') && !contains(github.event.issue.labels.*.name, 'user responded') }}
```

.github/workflows/update-labels.yml
shreemaan-abhishek: I think `update labels` would be better
```suggestion
      - name: update labels when user responds
```

.github/workflows/update-labels.yml
shreemaan-abhishek: Hmm let me implement that.

.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,22 @@
...
+  run-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: update labels when user responds
+        uses: andymckay/labeler@master
+        if: ${{ github.event.comment.user.login == github.event.issue.user.login && contains(github.event.issue.labels.*.name, 'wait for update') && !contains(github.event.issue.labels.*.name, 'user responded') }}
+        with:
+          add-labels: 'user responded'
+          remove-labels: 'wait for update'
+        if: ${{ github.event.comment.user.login == github.event.issue.user.login && contains(github.event.issue.labels.*.name, 'no response') && contains(github.event.issue.labels.*.name, 'wait for update') && !contains(github.event.issue.labels.*.name, 'user responded') }}
```
monkeyDluffy6017, line 19: Why do you add this condition `contains(github.event.issue.labels.*.name, 'no response') && contains(github.event.issue.labels.*.name, 'wait for update')`

shreemaan-abhishek: 1st condition is for the following [question of yours](https://github.com/apache/apisix/pull/9927#discussion_r1282988073):
> What happends if the user respond an issue with the label "no response"?

2nd condition: We only care if the user responded when we are actually `waiting for response`.

monkeyDluffy6017: It's very werid, do you notice the logic relationship `&&`, does it mean the label should contain both `no response` and `wait for update`?

monkeyDluffy6017: OK, i get you

monkeyDluffy6017 (approved):

soulbird (approved):

Issue #9951, opened 2023-08-02
bug: route 404 after upgrade to 3.2.2


wklken:
### Current Behavior

register route and service into etcd, the changes not been watched and got 404

### Expected Behavior

apisix watch the changes in time, the url should not be 404

### Error Logs

I have create route/service 4 times, delete the instance and redo
only 1 time got an error log. another 3 times got no error log(still 404)

```
2023/08/02 08:28:44 [error] 1051#1051: *44434 upstream timed out (110: Connection timed out) while reading upstream, client: unix:, server: , request: "POST /v3/watch HTTP/1.1", upstream: "http://10.1.1.1:2379/v3/watch", host: "127.0.0.1"
```

---

`apisix reload`

```
2023/08/02 08:49:43 [warn] 38#38: *19 [lua] config_etcd.lua:168: main etcd watcher started, revision=437733, context: ngx.timer
2023/08/02 08:49:43 [warn] 41#41: *36 [lua] config_etcd.lua:168: main etcd watcher started, revision=437733, context: ngx.timer

....
# do reload

2023/08/02 09:05:05 [notice] 770#770: signal process started
2023/08/02 09:05:05 [emerg] 773#773: bind() to 0.0.0.0:6008 failed (98: Address already in use)
2023/08/02 09:05:05 [warn] 771#771: *24136 [lua] config_etcd.lua:168: main etcd watcher started, revision=437755, context: ngx.timer
```

### Steps to Reproduce

1. upgrade apisix from 3.2.1 to 3.2.2
2. register the service and route
3. apisix-dashboard can see the route and service
4. apisix admin api can fetch the route and service
5. call the url,  404
7. do `apisix reload`, call the url, 200;
8. downgrade to apisix 3.2.1, do 2-5, call the url, 200;

------

got an error log

maybe this feature cause the error: https://github.com/apache/apisix/pull/9456

### Environment

- APISIX version (run `apisix version`): 3.2.2
- Operating system (run `uname -a`): 
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.21.4.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 2.5.4
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

qihaiyan:
Same error occurs to me.

jiangfucheng:
[...]

PR #9967, opened 2023-08-03
fix: can't sync etcd data if key has special character

jiangfucheng:
### Description

Fixes https://github.com/apache/apisix/issues/9951

### Checklist

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -392,7 +392,7 @@ local function http_waitdir(self, etcd_cli, key, modified_index, timeout)
         if tonumber(res.result.header.revision) > self.prev_index then
             local res2
             for _, evt in ipairs(res.result.events) do
-                if evt.kv.key:find(key, 1, true) == 1 then
+                if core_str.find(evt.kv.key, key) == 1 then
                     if not res2 then
                         res2 = tablex.deepcopy(res)
                         table.clear(res2.result.events)
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -392,7 +392,7 @@ local function http_waitdir(self, etcd_cli, key, modified_index, timeout)
         if tonumber(res.result.header.revision) > self.prev_index then
             local res2
             for _, evt in ipairs(res.result.events) do
-                if evt.kv.key:find(key) == 1 then
+                if evt.kv.key:find(key, 1, true) == 1 then
                     if not res2 then
                         res2 = tablex.deepcopy(res)
                         table.clear(res2.result.events)
--- a/t/core/config_etcd.t
+++ b/t/core/config_etcd.t
@@ -446,7 +446,7 @@ qr/readdir key: fake res: \{.*"nodes":\[\{.*"value":\["bar"\].*\}\].*\}/
 
 
 
-=== TEST 13: add route with special character "-"
+=== TEST 13: test route with special character "-"
 --- yaml_config
 deployment:
   role: traditional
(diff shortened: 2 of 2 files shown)


apisix/core/config_etcd.lua
```diff
@@ -392,7 +392,7 @@ local function http_waitdir(self, etcd_cli, key, modified_index, timeout)
         if tonumber(res.result.header.revision) > self.prev_index then
             local res2
             for _, evt in ipairs(res.result.events) do
-                if evt.kv.key:find(key) == 1 then
+                if evt.kv.key:find(key, 1, true) == 1 then
```
monkeyDluffy6017, line 395: Use `core.string.find` instead

t/core/config_etcd.t
```diff
@@ -443,3 +443,81 @@ qr/readdir key: fake res: \{.*"nodes":\[\{.*"value":\["bar"\].*\}\].*\}/
 --- wait: 1
 --- no_error_log
 [error]
+
+
+
+=== TEST 13: test route with special character "-"
```
jiangfucheng, line 449: Add all tests in one "TEST" because we need to ensure nginx not be reloaded to cover this bug.

t/core/config_etcd.t
```diff
@@ -443,3 +443,81 @@ qr/readdir key: fake res: \{.*"nodes":\[\{.*"value":\["bar"\].*\}\].*\}/
...
+  role_traditional:
+    config_provider: etcd
+  admin:
+    admin_key: null
+  etcd:
+    prefix: "/apisix-test"
+--- config
+    location /t {
+        content_by_lua_block {
+            ngx.sleep(0.5)
```
jiangfucheng, line 462: This sleep is to wait for the watch thread to start, according to my testing, APISIX can't watch data if we send data immediately after APISIX starts.

t/core/config_etcd.t
monkeyDluffy6017: Just to make sure, the bug can be reproduced even nginx is reloaded, right?

t/core/config_etcd.t
monkeyDluffy6017: Is it better to set this before line 485
```
            ngx.sleep(0.5)
            -- hit
            local httpc = http.new()
```

t/core/config_etcd.t
jiangfucheng: Yes, even if nginx is reloaded it will also reproduce this bug as long as there are new events come.

t/core/config_etcd.t
jiangfucheng: It seems another issue, if we set data to etcd immediately after APISIX starts, the APISIX will can't update the data in memory. According to my debug, APISIX can watch the event, but the event will be ignored, the issue seems caused by blown code.  the real value will be `res.result.header.revision <= self.prev_index`, but I haven't found the specific reason yet. 

https://github.com/apache/apisix/blob/5917748495264977635e3a135195258b31944929/apisix/core/config_etcd.lua#L392C3-L392C3

monkeyDluffy6017 (approved):

t/core/config_etcd.t
jiangfucheng: > It seems another issue, if we set data to etcd immediately after APISIX starts, the APISIX will can't update the data in memory. According to my debug, APISIX can watch the event, but the event will be ignored, the issue seems caused by blown code. the real value will be `res.result.header.revision <= self.prev_index`, but I haven't found the specific reason yet.
> 
> https://github.com/apache/apisix/blob/5917748495264977635e3a135195258b31944929/apisix/core/config_etcd.lua#L392C3-L392C3

I tried to debug this issue, and will raise a new issue to describe how to reproduce it. So we can just keep the `sleep` logic in there to avoid this bug.

PR #9995, opened 2023-08-09
fix: safely check for upstream field in traffic split plugin

### LINKED ISSUE #9831: help request: traffic-split plugin occurs error
### Description

hello.
as this issue methioned:  [#5607](https://github.com/apache/apisix/issues/5607)

I also suffered such a situation. but I can not reproduce it. the error log is :
`

    /traffic-split.lua:209: attempt to index field 'upstream' (a string value)

`

according the source code. when the traffic-split plugin config "weighted_upstreams" . it's element has no property "upstream". but have property "upstream_id", it should not invode the follwing code (line 209, because **elseif upstream_obj.upstream then** return false )
image

so I confused why line 209 is invoked.

traffic_split config is below:
`

    {
	"traffic-split": {
		"disable": false,
		"rules": [{
			"weighted_upstreams": [{
				"weight": 0
			}, {
				"upstream_id": "466807194923303750",
				"weight": 100
			}, {
				"upstream_id": "466807194721978186",
				"weight": 0
			}]
		}]
	}

}

`

### Environment

- APISIX version (run `apisix version`): v2.12.0
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):3.5.0

Revolyssup:
### Description

Fixes #9831 

Although the schema check disallows anything other than the object inside `upstream` field but in certain edge cases there can be issues with etcd like in the above mentioned issue so having a concrete check whether the upstream field is a table is needed. Otherwise in edge case, if upstream is set as string then it will result in nil error. 

### Checklist

### DIFF
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -200,7 +200,7 @@ local function new_rr_obj(weighted_upstreams)
     for i, upstream_obj in ipairs(weighted_upstreams) do
         if upstream_obj.upstream_id then
             server_list[upstream_obj.upstream_id] = upstream_obj.weight
-        elseif type(upstream_obj.upstream) == "table" then
+        elseif upstream_obj.upstream and type(upstream_obj.upstream) == "table" then
             -- Add a virtual id field to uniquely identify the upstream key.
             upstream_obj.upstream.vid = i
             -- Get the table id of the nodes as part of the upstream_key,
--- a/apisix/plugins/traffic-split.lua
+++ b/apisix/plugins/traffic-split.lua
@@ -200,7 +200,7 @@ local function new_rr_obj(weighted_upstreams)
     for i, upstream_obj in ipairs(weighted_upstreams) do
         if upstream_obj.upstream_id then
             server_list[upstream_obj.upstream_id] = upstream_obj.weight
-        elseif upstream_obj.upstream then
+        elseif type(upstream_obj.upstream) == "table" then
             -- Add a virtual id field to uniquely identify the upstream key.
             upstream_obj.upstream.vid = i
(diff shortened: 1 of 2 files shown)


Sn0rt:
we need test case for no exsting resource. (such as http upstream)

apisix/plugins/traffic-split.lua
```diff
@@ -200,7 +200,7 @@ local function new_rr_obj(weighted_upstreams)
     for i, upstream_obj in ipairs(weighted_upstreams) do
         if upstream_obj.upstream_id then
             server_list[upstream_obj.upstream_id] = upstream_obj.weight
-        elseif upstream_obj.upstream then
+        elseif upstream_obj.upstream and type(upstream_obj.upstream) == "table" then
```
monkeyDluffy6017, line 203: It doesn't fix the user's problem, the `upstream` property shouldn't be a string, if we add the judgement here, it will cause other problem elsewhere

Revolyssup: @monkeyDluffy6017 Yes I agree with you but the user's bug wasn't reproducible and the data gets corrupted later on so adding this extra check just avoid that nil panic, and nothing else.

monkeyDluffy6017: We cannot just add the check here,  The `upstream_obj.upstream` is used in other parts of the code as well

Revolyssup: so should we not fix this on apisix's end since this data was not potentially corrupted by APISIX? Should I close this PR?

monkeyDluffy6017: Please close this pr and find the root cause of the problem

Issue #10047, opened 2023-08-17
docs: what is the syntax of {id} in apisix-admin api?


MirtoBusico:
### Current State

In the admin-api api chapter ( https://apisix.apache.org/docs/apisix/admin-api/ ) all the examples for defining SSL, Upstream and Route use an {id} 

In the Request methods it is indicated as {...}

In all examples this {id} is a numeric string

There is no explanation of the {id} syntax

### Desired State

Add the syntax rules for {id}:

1. is it numeric
2. is it a string

If it is a string:

1. what is the max length
2. what characters are allowed (letter, numbers, special as - _ . # @ )

shreemaan-abhishek:
Following is the schema for {id}:

https://github.com/shreemaan-abhishek/apisix/blob/6e4347840727c54dd6b79a60e53afc1204938242/apisix/schema_def.lua#L30-L38

```lua
local id_schema = {
    anyOf = {
        {
            type = "string", minLength = 1, maxLength = 64,
            pattern = [[^[a-zA-Z0-9-_.]+$]]
        },
        {type = "integer", minimum = 1}
    }
}
```

MirtoBusico:
Fantastic!
It is exactly what I needed!

PR #10056, opened 2023-08-18
Update admin-api.md

jaw163:
Added note on ID syntax in admin-api.md

### Description

Added note on ID syntax and links to this part in the areas where a clearer understanding was stated as needed (Upstream and SSL sections). Made this change as there was no explanation for the rules around this part of the API.

Fixes # (issue)

Refers to issue #10047

### DIFF
--- a/docs/en/latest/admin-api.md
+++ b/docs/en/latest/admin-api.md
@@ -277,6 +277,10 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
 
 Route resource request address: /apisix/admin/routes/{id}?ttl=0
 
+### Quick Note on ID Syntax
+
+ID's as a text string must be of a length between 1 and 64 characters and they should only contain uppercase, lowercase, numbers and no special characters apart from dashes ( - ) and underscores ( _ ). For integer values they simply must have a minimum character count of 1.
+
 ### Request Methods
 
 | Method | Request URI                      | Request Body | Description                                                                                                                   |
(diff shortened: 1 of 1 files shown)


docs/en/latest/admin-api.md
```diff
@@ -1078,7 +1084,7 @@ Example Configuration:
     ...
     ```
 
-    After the execution is successful, nodes will not retain the original data, and the entire update is:
+    After the executiis successful, nodes will not retain the original data, and the entire update is:
```
SylviaBABY, line 1087: is this right？ maybe `After executed successfully`？ i dont think `executiis` is the correct word

jaw163: Sorry about that, I'm not sure how that was changed, it shouldn't have been. I have change it back to how it was in the recent commit

Revolyssup (dismissed):

docs/en/latest/admin-api.md
```diff
@@ -277,6 +277,10 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
 
 Route resource request address: /apisix/admin/routes/{id}?ttl=0
 
+### Quick Note on ID Syntax
+
+ID's as a text string must be of a length between 1 and 64 characters and they should only contain uppercase, lowercase, numbers and no special characters apart from dashes ( - ) and underscores ( _ ). For integer values they simply must have a minimum character count of 1.
```
shreemaan-abhishek, line 282: It also accepts periods ".". Please include this as well.

shreemaan-abhishek (changes requested):

docs/en/latest/admin-api.md
jaw163: I've included that too. Should I squash the commits?

shreemaan-abhishek (approved):

Revolyssup (approved):

PR #10070, opened 2023-08-22
fix(inspect): jit should be recovered after all hooks are done

louxiu:
### Description

if all hooks are done and debug mode are disabled, it would be better turn on jit on immediately.

https://api7.ai/blog/apisix-lua-dynamic-debugging-plugin
`Once debugging ends and all breakpoints have been revoked, the system will return to normal JIT mode, and the cleared JIT cache will be regenerated once it re-enters the hot spot.`

Fixes # (issue)

### Checklist


kingluo (approved):
@louxiu Yes, it's a bug. Thanks for your bugfix.

monkeyDluffy6017 (approved):

PR #10072, opened 2023-08-22
feat: support pass json to wasm plugin

### LINKED ISSUE #10034: feat: As a user, I want to try wasm plugin, so that pass a json config to plugin instead of a string
### Description

[for example](https://api7.ai/blog/how-apisix-supports-wasm)

```json
curl  http://127.0.0.1:9180/apisix/admin/routes/1 \
-H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '{
    "uri":"/*",
    "upstream":{
        "type":"roundrobin",
        "timeout":{
            "connect":1,
            "read":1,
            "send":1
        },
        "nodes":{
            "httpbin.org:80":1
        }
    },
    "plugins":{
        "wasm_fault_injection":{
            "conf":"{\"http_status\":200, \"body\":\"Hello WebAssembly!\n\"}"
        }
    },
    "name":"wasm_fault_injection"
}'


```

the configure of the wasm_fault_injection plugin is a string. actually I want pass a json (table in lua).

This makes it easier to construct more complex configurations and transfer them to plugins

Sn0rt:
### Description

Fixes #10034

### Checklist


t/wasm/fault-injection2.t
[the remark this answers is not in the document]
Sn0rt: forget to delete

apisix/wasm.lua
```diff
@@ -26,8 +26,14 @@ local schema = {
             type = "string",
             minLength = 1,
         },
+        setting = {
+            type = "object",
+        },
+    },
+    oneOf = {
+        {required = {"conf"}},
+        {required = {"setting"}},
```
monkeyDluffy6017, line 35: Why not use the same option `conf`,  use `type` to check if conf is a object or string

Sn0rt: Makes sense, I'll modify it later.

apisix/wasm.lua
```diff
@@ -15,25 +15,31 @@
...
 local _M = {}
 
 
 local function check_schema(conf)
-    return core.schema.check(schema, conf)
+    if type(conf.conf) ~= "table" and type(conf.conf) ~= "string" then
+        return false, "invalid conf type"
+    end
+
+    if conf.conf == "" or conf.conf == {} then
```
leslie-tsang, line 38: Need test case

leslie-tsang (changes requested):

apisix/wasm.lua
AlinsRan: https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L289

Issue #9878, opened 2023-07-21
bug: a error "changing jit stack size is not allowed when some regexs have already been compiled and cached" is output


qwzhou89:
### Current Behavior

When I put the code to register a new variable inside my_hook.lua and execute apisix reload, an error is output inside the logs

### Expected Behavior

There should be no error.

### Error Logs

2023/07/21 06:53:10 [error] 1#1: init_by_lua error: /usr/local/apisix/apisix/init.lua:21: changing jit stack size is not allowed when some regexs have already been compiled and cached
stack traceback:
	[C]: in function 'error'
	/usr/local/openresty/lualib/ngx/re.lua:297: in function 'opt'
	/usr/local/apisix/apisix/init.lua:21: in main chunk
	[C]: in function 'require'
	init_by_lua:4: in main chunk

### Steps to Reproduce

1.Run APISIX 2.15.3 in docker.
2.Modify 3 lines in conf/config-default.yaml as follows：
```shell
  extra_lua_path: "/usr/local/apisix/conf/hook/?.lua"                # extend lua_package_path to load third party code
  extra_lua_cpath: ""               # extend lua_package_cpath to load third party code
  lua_module_hook: "my_hook"  # the hook module which will be used to inject third party code into APISIX
``` 
3.Create a new /usr/local/apisix/conf/hook/my_hook.lua file and insert the following contents
(this is copied from https://apisix.apache.org/docs/apisix/next/plugin-develop/#register-custom-variable)
```lua
local core = require "apisix.core"

core.ctx.register_var("a6_labels_zone", function(ctx)
    local route = ctx.matched_route and ctx.matched_route.value
    if route and route.labels then
        return route.labels.zone
    end
    return nil
end)
``` 
4.Execute the `apisix reload` command
5.In the log there is the following output
```
2023/07/21 07:11:38 [warn] 1#1: "ssl_stapling" ignored, issuer certificate not found for certificate "/usr/local/apisix/conf/cert/ssl_PLACE_HOLDER.crt"
2023/07/21 07:11:38 [error] 1#1: init_by_lua error: /usr/local/apisix/apisix/init.lua:21: changing jit stack size is not allowed when some regexs have already been compiled and cached
stack traceback:
	[C]: in function 'error'
	/usr/local/openresty/lualib/ngx/re.lua:297: in function 'opt'
	/usr/local/apisix/apisix/init.lua:21: in main chunk
	[C]: in function 'require'
	init_by_lua:4: in main chunk

``` 

[...]

PR #10086, opened 2023-08-25
test: register a custom variable in a custom Lua file and source it with `extra_lua_path`

kayx23:
### Description

* Updated doc for _Custom Variable._
* Added example and test to register a custom variable by sourcing a custom Lua file.

### Checklist

### DIFF
--- a/docs/en/latest/plugin-develop.md
+++ b/docs/en/latest/plugin-develop.md
@@ -1,5 +1,5 @@
 ---
-title: Plugin Develop
+title: Develop Plugin
 ---
 
 <!--
@@ -71,8 +71,7 @@ apisix:
     lua_module_hook: "my_hook"
 ```
 
-The `example/my_hook.lua` will be loaded when APISIX starts, and you can use this hook to replace a method in APISIX.
-The example of [my_hook.lua](https://github.com/apache/apisix/blob/master/example/my_hook.lua) can be found under the `example` directory of this project.
+The `example/my_hook.lua` will be loaded when APISIX starts, and you can use this hook to replace a method in APISIX. The example of [my_hook.lua](https://github.com/apache/apisix/blob/master/example/my_hook.lua) can be found under the `example` directory of this project.
 
 ## check dependencies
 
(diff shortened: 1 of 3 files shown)


example/register_custom_var.lua
```diff
@@ -0,0 +1,42 @@
...
+    ngx.log(ngx.EMERG, "my hook works in http")
+    old_http_init(...)
+end
+
+local old_stream_init = apisix.stream_init
+apisix.stream_init = function(...)
+    register_custom_var()
+    ngx.log(ngx.EMERG, "my hook works in stream")
+    old_stream_init(...)
+end
```
Revolyssup, line 42: To show that the variable was really registered, should we try to access and log this in another phase and then grep and assert in the tests?

kayx23: Do you mean something like this?

```lua
local old_http_access_phase = apisix.http_access_phase
apisix.http_access_phase = function(...)
    ngx.log(ngx.EMERG, ctx.a6_route_labels)
    old_http_access_phase(...)
end
```

My understanding is that the custom variable `a6_route_labels` would only has value if a route is configured with the prop `labels`. 

The `register_var("a6_route_labels")` example is kept consistent with this test case: https://github.com/apache/apisix/blob/89c1858662e34c76de01c84b665c79fcb2edaeb8/t/plugin/http-logger-log-format.t#L431-L441

Please advise.

Revolyssup: Yes part of the suggestion was to PUT that route (just like in the test you mentioned) and then send request to that route, after which to make assertion. But I don't have a strong opinion here. If others think that showing how it's registered is good enough then I think its fine.

shreemaan-abhishek (approved):
Since we already have a test:
https://github.com/apache/apisix/blob/89c1858662e34c76de01c84b665c79fcb2edaeb8/t/plugin/http-logger-log-format.t#L431-L441

That shows the given example function for registering vars is correct. LGTM.

Revolyssup:
> Since we already have a test:
> 
> https://github.com/apache/apisix/blob/89c1858662e34c76de01c84b665c79fcb2edaeb8/t/plugin/http-logger-log-format.t#L431-L441
> 
> That shows the given example function for registering vars is correct. This PR LGTM.

So we should reference this test in the doc

docs/en/latest/plugin-develop.md
```diff
@@ -488,25 +487,33 @@ curl -i -X GET "http://127.0.0.1:9090/v1/plugin/example-plugin/hello"
...
-```
+```lua
 local core = require "apisix.core"
 
-core.ctx.register_var("a6_labels_zone", function(ctx)
+core.ctx.register_var("a6_route_labels", function(ctx)
     local route = ctx.matched_route and ctx.matched_route.value
     if route and route.labels then
-        return route.labels.zone
+      return route.labels
```
leslie-tsang, line 500:
```suggestion
        return route.labels
```
Please standardize indentation style. `4` space will be please

example/register_custom_var.lua
```diff
@@ -0,0 +1,42 @@
...
+apisix.http_init = function(...)
+    register_custom_var()
+    ngx.log(ngx.EMERG, "my hook works in http")
+    old_http_init(...)
+end
+
+local old_stream_init = apisix.stream_init
+apisix.stream_init = function(...)
+    register_custom_var()
+    ngx.log(ngx.EMERG, "my hook works in stream")
```
leslie-tsang, line 40:
```suggestion
    ngx.log(ngx.EMERG, "my hook works in stream")
```
please use `core.log` instead.

example/register_custom_var.lua
leslie-tsang: We've already add test cases for this feature, and it doesn't seem necessary to do it again in the demo.
BTW, "my hook works in http" doesn't mean it actually works, it's just a log.

leslie-tsang (changes requested):

kayx23:
>  it doesn't seem necessary to do it again in the demo

@leslie-tsang this provides a working example for this issue (and doc): https://github.com/apache/apisix/issues/9878

example/register_custom_var.lua
kayx23: Yeah I understand this doesn't cover the entire use case. The initial motivation of this was users having difficulties understanding how to register a variable with a custom Lua file that gets sourced (i.e. `lua_module_hook`). Once it's sourced and APISIX starts successfully (aka here it doesn't error out), the rest is easy. I agree with you that the feature of sourcing external files is covered in test, however.

kayx23:
I agree this is more of an example than test coverage for the feature. Maybe we'd just update the docs alone to provide more instructions. https://github.com/apache/apisix/pull/10086#discussion_r1305408692

PR #10112, opened 2023-08-30
feat: support server func

lingsamuel:
### Description

Provides a way to access external service function in plugin development. Also contains an example in `example-plugin`.

### Checklist

### DIFF
--- a/apisix/core/utils.lua
+++ b/apisix/core/utils.lua
@@ -24,12 +24,14 @@ local core_str       = require("apisix.core.string")
 local rfind_char     = core_str.rfind_char
 local table          = require("apisix.core.table")
 local log            = require("apisix.core.log")
+local json            = require("apisix.core.json")
 local string         = require("apisix.core.string")
 local dns_client     = require("apisix.core.dns.client")
 local ngx_re         = require("ngx.re")
 local ipmatcher      = require("resty.ipmatcher")
 local ffi            = require("ffi")
 local base           = require("resty.core.base")
+local http           = require "resty.http"
 local open           = io.open
 local sub_str        = string.sub
 local str_byte       = string.byte
@@ -375,5 +377,27 @@ end
 -- Resolve {$1, $2, ...} in the given string
 _M.resolve_var_with_captures = resolve_var_with_captures
 
+local function server_func(func_name, body, headers)
+    local local_conf = config_local.local_conf()
+
+    if local_conf.apisix and local_conf.apisix.server_func_addr
+            and local_conf.apisix.server_func_addr ~= "" then
+        local addr = local_conf.apisix.server_func_addr
+        if not string.has_suffix(addr, "/") then
+            addr = addr .. "/"
+        end
+        log.warn("external server func to: " .. addr .. func_name .. ", body: " .. json.encode(body))
+        local httpc = http.new()
+        local res, err = httpc:request_uri(addr .. func_name, {
+            method = "POST",
+            body = body,
+            headers = headers,
+        })
+        return res, err
+    else
+        return nil, "no apisix.server_func_addr configured"
+    end
+end
+_M.server_func = server_func
 
 return _M
--- a/apisix/core/utils.lua
+++ b/apisix/core/utils.lua
@@ -386,7 +386,8 @@ local function server_func(func_name, body, headers)
         if not string.has_suffix(addr, "/") then
             addr = addr .. "/"
         end
-        log.info("external server func to: " .. addr .. func_name .. ", body: " .. json.encode(body))
+        log.info("external server func to: " .. addr .. func_name ..
+            ", body: " .. json.encode(body))
         local httpc = http.new()
         local res, err = httpc:request_uri(addr .. func_name, {
             method = "POST",
--- a/apisix/core/utils.lua
+++ b/apisix/core/utils.lua
@@ -24,14 +24,14 @@ local core_str       = require("apisix.core.string")
 local rfind_char     = core_str.rfind_char
 local table          = require("apisix.core.table")
 local log            = require("apisix.core.log")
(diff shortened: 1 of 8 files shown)


apisix/plugins/example-plugin.lua
```diff
@@ -27,6 +27,7 @@ local schema = {
         t = {type = "array", minItems = 1},
         ip = {type = "string"},
         port = {type = "integer"},
+        server_func = {type = "string"},
```
tao12345666333, line 30: rename it to `func_name` ?

apisix/plugins/example-plugin.lua
```diff
@@ -85,6 +86,27 @@ function _M.access(conf, ctx)
...
+        -- Example plugin server function call
+        local res, err = core.server_func(conf.server_func, core.json.encode({
+            ["data"] = "body-content"
+        }), {
+            ["Content-Type"] = "application/json",
+        })
+        if not err then
+            core.log.warn("plugin server func result: ", res.body)
+            if res.status == 200 then
+                core.response.set_header("Server-Func-Response", res.body)
```
tao12345666333, line 99: return the response body?

apisix/core/utils.lua
```diff
@@ -375,5 +377,27 @@ end
...
+local function server_func(func_name, body, headers)
+    local local_conf = config_local.local_conf()
+
+    if local_conf.apisix and local_conf.apisix.server_func_addr
+            and local_conf.apisix.server_func_addr ~= "" then
+        local addr = local_conf.apisix.server_func_addr
+        if not string.has_suffix(addr, "/") then
+            addr = addr .. "/"
+        end
+        log.warn("external server func to: " .. addr .. func_name .. ", body: " .. json.encode(body))
```
leslie-tsang, line 389: would be better to use `info` level ? seems no need to use `warn` as default.

Gallardot (commented):
Can this feature be used in the [serverless](https://apisix.apache.org/docs/apisix/plugins/serverless/) plugin?

apisix/plugins/example-plugin.lua
lingsamuel: User may need other information, such as res.status

jiangfucheng:
Hi, could you describe the specific usage scenario?  So we can know when to use it during development.

apisix/core/utils.lua
```diff
@@ -375,5 +377,28 @@ end
...
+        and local_conf.apisix.server_func_addr ~= "" then
+        local addr = local_conf.apisix.server_func_addr
+        if not string.has_suffix(addr, "/") then
+            addr = addr .. "/"
+        end
+        log.info("external server func to: " .. addr .. func_name ..
+            ", body: " .. json.encode(body))
+        local httpc = http.new()
+        local res, err = httpc:request_uri(addr .. func_name, {
+            method = "POST",
```
jiangfucheng, line 393: Would it be possible to replace this constant with a parameter? This would increase flexibility.

tao12345666333 (approved):

leslie-tsang (approved):

tao12345666333:
> Can this feature be used in the [serverless](https://apisix.apache.org/docs/apisix/plugins/serverless/) plugin?

Sure, it’s a function, you can call it if you want.

tao12345666333:
> Hi, could you describe the specific usage scenario? So we can know when to use it during development.

For example, when you use Lua Plugin as an extension entry, you can call external functions to implement specific logic at any stage.

This function just provides an example, or it stipulates a calling mode through which users can use this mode more conveniently.

Of course, you can also see that its implementation logic is very simple. If users do not want to use this mode, they can also implement the corresponding logic by themselves.

apisix/core/utils.lua
tao12345666333: Using parameters can indeed be more flexible.

https://github.com/apache/apisix/pull/10112#issuecomment-1716791638

As I mentioned before, a usage pattern is provided here. If users do not want to use this pattern, they can implement the logic by themselves.

Gallardot (approved):

bzp2010 (changes requested):
### Why do we have to add it?

In my opinion, it's a completely irrelevant feature, it can't even be called a feature, it's just a **poor wrapper** for `lua-resty-http`.

#### Problems I see

- It doesn't make sense, anyone can just use `lua-resty-http` to do this, and he'll get more flexibility in terms of using the library's parameters
- Users must configure a single service address using the APISIX configuration file, why? What if I want to use more than one service?
- This function does not support the use of special configurations such as connection pooling.
- The `lua-resty-http` itself is also an outdated thing, it does not support HTTP/2 HTTP/3 and does not support complex use cases.

#### So

It's not clear to me that there's any context beyond this PR such that we'd have to add something that seems irrelevant. I see it collecting enough Approve now, but I think it needs more info.

Currently, I hold a different opinion.



> Is it premature optimization?

monkeyDluffy6017:
This pr needs more detail

membphis:
I read the @bzp2010 's comment: https://github.com/apache/apisix/pull/10112#pullrequestreview-1623517064

I finally agree with him. I think we can use add a doc to `Q&A` about `How to make APISIX+Golang/Java/Nodejs work together`, use HTTP call is an easy way.

bzp2010:
@membphis Even if we implement it, we will still have the same problems we once had.

> Instead of being able to rewrite the response body in the normal datapath (using upstream), the Nginx upstream had to be bypassed by forwarding the request using the HTTP client. It's often used as response encryption and decryption, etc.

The problem still exists, and OpenResty prohibits performing this behavior. Developers don't understand these details and don't want to understand them, which is a serious mental burden. These things should be hidden under the hood, and this way doesn't help solve it either.

PR #10138, opened 2023-09-01
fix: provide error instead of nil panic when cache_zone is missing

### LINKED ISSUE #9957: bug: runtime error with `proxy-cache` plugin when using in memory caching without specifying the `memory_cache` zone
### Current Behavior

This issue summurizes https://github.com/apache/apisix/issues/8719 which was reported for the same behaviour.

If you configure `proxy-cache` plugin with ONLY `"cache_strategy": "memory"`, you get a `500 Internal Server Error` 

```bash
$ curl http://127.0.0.1:9180/apisix/admin/routes/1 \
{
    "uri": "/ip",
    "plugins": {
        "proxy-cache": {
            "cache_strategy": "memory",
            "cache_ttl": 10
        }
    },
    "upstream": {
        "nodes": {
            "httpbin.org": 1
        },
        "type": "roundrobin"
    }
}'
```

```bash
$ curl -i http://127.0.0.1:9080/ip
<html>
<head><title>500 Internal Server Error</title></head>
<body>
<center><h1>500 Internal Server Error</h1></center>
<hr><center>openresty</center>
<p><em>Powered by <a href="https://apisix.apache.org/">APISIX</a>.</em></p></body>
</html>
```

Logs:

```text
[error] 128#128: *7943331 lua entry thread aborted: runtime error: ./apisix/plugins/proxy-cache/memory.lua:47: attempt to index field 'dict' (a nil value)
stack traceback:
coroutine 0:
        ./apisix/plugins/proxy-cache/memory.lua: in function 'get'
        ./apisix/plugins/proxy-cache/memory_handler.lua:204: in function 'phase_func'
        ./apisix/plugin.lua:1099: in function 'run_plugin'
        ./apisix/init.lua:674: in function 'http_access_phase'
        access_by_lua(nginx.conf:340):2: in main chunk, client: 192.168.112.1, server: _, request: "POST /api HTTP/1.1", host: "127.0.0.1:9080"
```

The correct usage is to configure `"cache_strategy": "memory"`with `"cache_zone": "memory_cache"` (or any other custom memory cache zones defined in `config.yaml`), such as the below:

```json
"proxy-cache": {
      "cache_strategy": "memory",
      "cache_zone": "memory_cache"
}
```

### Expected Behavior

If a user incorrectly configure just the  `"cache_strategy": "memory"` without specifying a memory zone, the program should return an error message to help users identify the issue, instead of crashing like the current behaviour.

### Suggested Fix

- [ ] Add the error handler to return an error message to help users identify the issue, instead of crashing.
- [x] Update the [proxy-cache](https://apisix.apache.org/docs/apisix/plugins/proxy-cache/) doc. The current doc does not have an example for the in-memory `cache_strategy`, nor does it warn users of configuring the proper `cache_zone` for in-memory `cache_strategy`.

### Environment

APISIX version (run `apisix version`): 3.4.0

Revolyssup:
### Description

Fixes #9957 

### Checklist

### DIFF
--- a/apisix/plugins/proxy-cache/memory.lua
+++ b/apisix/plugins/proxy-cache/memory.lua
@@ -32,6 +32,10 @@ end
 
 
 function _M:set(key, obj, ttl)
+    if self.dict == nil then
+        return nil, "no cache_zone provided"
+    end
+
     local obj_json = core.json.encode(obj)
     if not obj_json then
         return nil, "could not encode object"
@@ -43,6 +47,9 @@ end
 
 
 function _M:get(key)
+    if self.dict == nil then
+        return nil, "no cache_zone provided"
+    end
     -- If the key does not exist or has expired, then res_json will be nil.
     local res_json, err = self.dict:get(key)
     if not res_json then
@@ -63,6 +70,9 @@ end
 
 
 function _M:purge(key)
+    if self.dict == nil then
+        return nil, "no cache_zone provided"
+    end
     self.dict:delete(key)
 end
 
--- a/t/plugin/proxy-cache/memory.t
+++ b/t/plugin/proxy-cache/memory.t
@@ -661,3 +661,58 @@ GET /hello
 --- more_headers
 Cache-Control: only-if-cached
 --- error_code: 504
+
+
+
+=== TEST 36: configure plugin without memory_cache zone
+--- config
+       location /t {
+           content_by_lua_block {
+               local t = require("lib.test_admin").test
+               local code, body = t('/apisix/admin/routes/1',
+                    ngx.HTTP_PUT,
+                    [[{
+                        "plugins": {
+                            "proxy-cache": {
+                               "cache_strategy": "memory",
+                               "cache_key":["$host","$uri"],
+                               "cache_bypass": ["$arg_bypass"],
+                               "cache_control": true,
+                               "cache_method": ["GET"],
+                               "cache_ttl": 10,
+                               "cache_http_status": [200]
+                            }
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:1986": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/hello"
+                   }]]
+                   )
+
+               if code >= 300 then
+                   ngx.status = code
+               end
+               ngx.say(body)
+           }
+       }
+--- request
+GET /t
+--- error_code: 200
+--- response_body
+passed
+
+
+
+=== TEST 37: hit route
+--- LAST
+--- request
+GET /hello
+--- grep_error_log eval
+qr/no cache_zone provided/
+--- grep_error_log_out
+no cache_zone provided
+no cache_zone provided


apisix/plugins/proxy-cache/memory.lua
```diff
@@ -43,6 +47,9 @@ end
 
 
 function _M:get(key)
+    if self.dict == nil then
+        return nil, "no cache_zone provided"
+    end
```
monkeyDluffy6017, line 52: Add a blank below

t/plugin/proxy-cache/memory.t
```diff
@@ -661,3 +661,57 @@ GET /hello
...
+
+
+=== TEST 37: hit route
+--- request
+GET /hello
+--- grep_error_log eval
+qr/no cache_zone provided/
+--- grep_error_log_out
+no cache_zone provided
+no cache_zone provided
```
monkeyDluffy6017, line 717: @kayx23 Does this meet your requirement?

monkeyDluffy6017, line 717: @Revolyssup Could you add a test case to show the response body and the error code ?

kayx23: There might be a misunderstanding? This should actually error out with an error in the 400 range and with the response body reminding users they should configure `cache_zone` to be an in-memory cache zone.

Current behaviour: If you configure `proxy-cache` plugin with `"cache_strategy": "memory"` and without `cache_zone`, currently you get a 500 Internal Server Error.

Revolyssup: @kayx23 That internal server error takes place when the requests are received and not when plugin is configured in admin API.
Now instead of 500 internal server error, the error logs will point out the fact that cache_zone is not configured so proxy-cache won't work as expected. The requests still pass through. 

I think your suggestion is more in line with, rejecting the configuration during schema validation if cache_zone is explicitly not provided while cache_strategy is `memory`

kayx23 (dismissed):

t/plugin/proxy-cache/memory.t
```diff
@@ -661,3 +661,46 @@ GET /hello
 --- more_headers
 Cache-Control: only-if-cached
 --- error_code: 504
+
+
+
+=== TEST 36: configure plugin without memory_cache zone
```
shreemaan-abhishek, line 667:
```suggestion
=== TEST 36: configure plugin without memory_cache zone for cache_strategy = memory
```

shreemaan-abhishek (changes requested):

shreemaan-abhishek (approved):

t/plugin/proxy-cache/memory.t
kayx23: Resolving as confirmed

kayx23 (dismissed):

Sn0rt (approved):

apisix/plugins/proxy-cache/init.lua
```diff
@@ -140,6 +141,12 @@ function _M.check_schema(conf)
         end
     end
 
+    -- For memory based cache, the default cache_zone cannot be used.
+    -- cache_zone will also be set as default value in case when passed empty.
+    if conf.cache_strategy == STRATEGY_MEMORY and conf.cache_zone == DEFAULT_CACHE_ZONE then
```
monkeyDluffy6017, line 146: It's not very reliable to judge by cache zone name, i think you could modified like this:
```
if local_conf.apisix.proxy_cache then
    for _, cache in ipairs(local_conf.apisix.proxy_cache.zones) do
        if cache.name == conf.cache_zone then
            if conf.cache_strategy == STRATEGY_MEMORY and not cache.disk_path then
                found = true
            elseif conf.cache_strategy == STRATEGY_DISK and cache.disk_path then
                found = true
            else
                found = false
            end
        end
    end
    if found == false then
        return false, "cache_zone " .. conf.cache_zone .. " not found"
    end
end
```

Revolyssup:
@monkeyDluffy6017 I added a slightly modified version of your suggestion with comments. Please review.

monkeyDluffy6017 (approved):

leslie-tsang (approved):

Issue #9415, opened 2023-05-05
bug: the exception of health check in special scene


zuiyangqingzhou:
### Current Behavior

When configuring similar clb nodes and IP nodes in upstream nodes at the same time, there will be node exception errors for clb nodes.

### Expected Behavior

All nodes should be normal.

### Error Logs

```
[error] 54#54: *338255 [lua] balancer.lua:82: fetch_health_nodes(): failed to get health check target status, addr: 39.156.66.14:80, host: nil, err: target not found, client: 172.18.0.1, server: _, request: "GET /hello HTTP/1.1", host: "127.0.0.1:9080"
```

### Steps to Reproduce

1. upstream config just like
```
{
    "id":"459301800810382014",
    "create_time":1683294806,
    "update_time":1683295767,
    "nodes":{
        "172.18.0.4:80":1,
        "172.18.0.5:80":1,
        "www.baidu.com:80":1
    },
    "timeout":{
        "connect":6,
        "send":6,
        "read":6
    },
    "type":"roundrobin",
    "checks":{
        "active":{
            "concurrency":10,
            "healthy":{
                "http_statuses":[
                    200,
                    302,
                    404
                ],
                "interval":1,
                "successes":2
            },
            "http_path":"/",
            "timeout":1,
            "type":"http",
            "unhealthy":{
                "http_failures":5,
                "http_statuses":[
                    500,
                    501,
                    502,
                    503,
                    504,
                    505
                ],
                "interval":1,
                "tcp_failures":2,
                "timeouts":3
            }
        }
    },
    "scheme":"http",
    "pass_host":"pass",
    "name":"test",
    "keepalive_pool":{
        "idle_timeout":60,
        "requests":1000,
        "size":320
    }
}
```
or the configuration in the background looks like this
![image](https://user-images.githubusercontent.com/17425270/236490177-e00c9237-0c75-43c2-960b-0c1602527277.png)

2. route config just like
```
{
    "id":"459301861040587454",
    "create_time":1683294842,
    "update_time":1683296659,
    "uri":"/*",
    "name":"test",
    "methods":[
        "GET",
        "POST",
        "PUT",
        "DELETE",
        "PATCH",
        "HEAD",
        "OPTIONS",
        "CONNECT",
        "TRACE",
        "PURGE"
    ],
    "upstream_id":"459301800810382014",
    "status":1
}
```
3. Initiate a request
```
 curl http://127.0.0.1:9080/hello -i
```
[...]

PR #10156, opened 2023-09-04
fix: use warn log for get healthcheck target status failure

shreemaan-abhishek:
### Description

https://github.com/apache/apisix/blob/35349ef508351364216c0c2f1c0fd8798905b9e2/apisix/balancer.lua#L82-L83
The above error log reports an error in some cases even if the request is valid and successful. In the initial requests, `get_target_status()` will always fail as the target status is not yet finalized. It does not make sense to report an error for positive outcomes.

Note that the `get_target_status()` failure just signifies that the system failed to get the health status only. We do not know if the target is actually unhealthy. And even if the node is unavailable/unhealthy `get_target_status()` does not report the error, but some warn logs like this:
```
2023/09/04 22:13:09 [warn] 14709#16724107: *2442 [lua] healthcheck.lua:650: [healthcheck] (upstream#/apisix/upstreams/1) unhealthy TCP increment (2/2) for '(127.0.0.1:8765)', context: ngx.timer, client: 127.0.0.1, server: 0.0.0.0:9080
```
#### Background information:

I encountered this problem while writing a test case for `https` based healthcheck.

### Checklist

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -79,7 +79,7 @@ local function fetch_health_nodes(upstream, checker)
         if ok then
             up_nodes = transform_node(up_nodes, node)
         elseif err then
-            core.log.error("failed to get health check target status, addr: ",
+            core.log.warn("failed to get health check target status, addr: ",
                 node.host, ":", port or node.port, ", host: ", host, ", err: ", err)
         end
     end


monkeyDluffy6017:
The error log occurs when the upstream is rebuilt as we add or delete an upstream node. The state is not correct, it will lead to unbalanced load. It's a problem, we should fix this, so the error log level is appropriate here.

shreemaan-abhishek:
> The state is not correct.

Do you mean `not correct` == `unhealthy`?

Also, in the beginning, there will be no state as the healthcheck won't be triggered unless the upstream has been accessed.
image


@monkeyDluffy6017 try running the following test case and check the servroot error logs and there you will find error logs even for positive outcomes.


>  t/stream-node/upstream-tls.t
```
use t::APISIX 'no_plan';

no_root_location();

add_block_preprocessor(sub {
    my ($block) = @_;

    if (!$block->http_config) {
        my $http_config = <<'_EOC_';
server {
    listen 8765 ssl;
    ssl_certificate ../../certs/mtls_server.crt;
    ssl_certificate_key ../../certs/mtls_server.key;
    ssl_client_certificate ../../certs/mtls_ca.crt;
    ssl_verify_client on;

    location /index {
        return 200 'gangnam style!';
    }
}
server {
    listen 8766 ssl;
    ssl_certificate ../../certs/mtls_server.crt;
    ssl_certificate_key ../../certs/mtls_server.key;
    ssl_client_certificate ../../certs/mtls_ca.crt;
    ssl_verify_client on;

    location /index {
        return 200 'gangnam style!';
    }
}
_EOC_
        $block->set_value("http_config", $http_config);
    }

    if (!$block->request) {
    }

});

run_tests;

__DATA__

=== TEST 7: https health check
--- log_level: info
--- config
    location /t {
        content_by_lua_block {
            local t = require("lib.test_admin")
            local cert = t.read_file("t/certs/mtls_client.crt")
            local key =  t.read_file("t/certs/mtls_client.key")
            local data = {
                retries = 2,
                checks = {
                    active = {
                    https_verify_certificate = false,
                    healthy = {
                        interval = 2,
                        successes = 1
                    },
                    unhealthy = {
                        interval = 1,
                        http_failures = 2
                    },
                    type = "https",
                    timeout = 5
                    },
                },
                type = "roundrobin",
                nodes = {
                    ["127.0.0.1:8766"] = 1,
                    ["127.0.0.1:8765"] = 1
                },
                scheme = "https",
                tls = {
                    client_cert = cert,
                    client_key = key
                }
            }
            local code, body = t.test('/apisix/admin/upstreams/1',
                 ngx.HTTP_PUT,
                 core.json.encode(data)
            )

            if code >= 300 then
                ngx.status = code
            end

            local code, body = t.test('/apisix/admin/routes/1',
                 ngx.HTTP_PUT,
                 [[{
                        "upstream_id": 1,
                        "uri": "/index"
                }]]
            )

            if code >= 300 then
                ngx.status = code
            end
        }
    }
--- error_code: 200
--- timeout: 15



=== TEST 8: ACCESS
--- pipelined_requests eval
["GET /index", "GET /index"]
--- response_body eval
["gangnam style!", "gangnam style!"]
--- error_code eval
[200, 200]
```

P.S: I am unable to assert the `error_log` using the test case so you will have to check the log manually, if you change the error log to warn log, the test will pass signifying the successful working of the feature.

leslie-tsang:
APISIX use [pre-defined preprocessor](https://github.com/apache/apisix/blob/bb228fadf660c664d462520ab326322968cd94e6/t/APISIX.pm#L912-L916) to handle the default `no_error_log`, maybe pass an `ignore_error_log` will feed you need.

monkeyDluffy6017:
@shreemaan-abhishek If @leslie-tsang's suggestion satisfy your requirement, the pr could be closed?

shreemaan-abhishek:
@monkeyDluffy6017, although the background for this PR was writing a test case, I am still convinced that using an error level log leads to a false negative.

monkeyDluffy6017:
@shreemaan-abhishek I have fix this by the following pr, so the error log won't happen again:
https://github.com/api7/lua-resty-healthcheck/pull/35
https://github.com/api7/lua-resty-healthcheck/pull/36

monkeyDluffy6017:
By the way, the log level `error` is as expected.  The `get_target_status` should not return an error

monkeyDluffy6017:

ZhangShangyu:
> The error log occurs when the upstream is rebuilt as we add or delete an upstream node. The state is not correct, it will lead to unbalanced load. It's a problem, we should fix this, so the error log level is appropriate here.

we also meet this error log, https://github.com/apache/apisix/issues/9415, finally we decide to ignore this target not find error, treat the node as health
```
if err == "target not found" then
    core.log.warn("health check target not found, ignore health check, addr: ",
            node.host, ":", port or node.port, ", host: ", host, ", err: ", err)
    up_nodes = transform_node(up_nodes, node)
else
    core.log.error("failed to get health check target status, addr: ",
            node.host, ":", port or node.port, ", host: ", host, ", err: ", err)
end
```

btw it may not lead to unbalanced load, because it add all node to the up_nodes as default.

monkeyDluffy6017:
I reopened this pull request because many people have encountered the same issue. This error should only occur when processing multiple requests concurrently. While the first request is in the process of creating a health check object, the second request encounters this error because it cannot access the object that hasn't been created yet. However, this error does not actually impact request forwarding. Lowering the error log level from 'error' to 'warn' is reasonable to prevent any misunderstanding.

monkeyDluffy6017 (approved):

soulbird (approved):

Issue #10197, opened 2023-09-14
bug: graphql post request route matching exception


AlinsRan:
### Current Behavior

Graphql post requests have a probability of causing routing matching exception.

The larger the body, the greater the probability of routing exceptions, which may be affected by other requests in the final result.

* expect
POST http://test.com/hello  -> route a
POST http://test2.com/hello -> rout b

* actual
POST http://test.com/hello  -> route a
POST http://test2.com/hello -> route a

### Steps to Reproduce

1. apisix configure `config.yaml`

```yaml
deployment:
  role: data_plane
  role_data_plane:
    config_provider: yaml

apisix:
  router:
    http: radixtree_uri

nginx_config:
  worker_processes: 1
  ```
 2. `apisix.yaml` configure
 
 ```yaml
 routes:
  - uri: "/hello"
    hosts:
      - test.com
    vars:
      - - "graphql_name"
        - "=="
        - "createAccount"
    priority: 30
    id: "graphql1"
    upstream_id: "invalid"

  - uri: "/hello"
    hosts:
      - test.com
    plugins:
      echo:
        body: "test server\n"
    priority: 20
    id: "graphql2"
    upstream_id: "invalid"
  
  - uri: "/hello"
    hosts:
      - test2.com
    plugins:
      echo:
        body: "test2\n"
    priority: 20
    id: "graphql3"
    upstream_id: "invalid"

upstreams:
  - nodes:
      127.0.0.1:1999: 1
    id: "invalid"
    ```
    
3. request params

```shell
➜  apisix git:(master) ✗ curl -Z http://test.com:9080/hello -H "Content-Type: application/json" -X POST -d @schema.json -Z http://test2.com:9080/hello
test2
test2
➜  apisix git:(master) ✗ curl -Z http://test.com:9080/hello -H "Content-Type: application/json" -X POST -d @schema.json -Z http://test2.com:9080/hello
test2
test server
➜  apisix git:(master) ✗ curl -Z http://test.com:9080/hello -H "Content-Type: application/json" -X POST -d @schema.json -Z http://test2.com:9080/hello
test2
test2
➜  apisix git:(master) ✗ curl -Z http://test.com:9080/hello -H "Content-Type: application/json" -X POST -d @schema.json -Z http://test2.com:9080/hello
test2
test2
```
 

### Environment

- APISIX version (run `apisix version`): master/3.5.0
- Operating system (run `uname -a`):
[...]

PR #10198, opened 2023-09-14
fix: graphql post request route matching exception

AlinsRan:
### Description

Reuse of `match_opts` causes the value of the next request to overwrite the value of the current request. GraphQL requests call the `ngx.req.read_body` API, which is a non-blocking call. If the request body has not been fully read (for example, if the body size exceeds the nginx buffer), the coroutine will yield. If there are other requests during this time period, `match_opts` will be contaminated by other requests, resulting in matching the wrong route.

Fixes: #10197 

### Checklist

### DIFF
--- a/apisix/http/router/radixtree_host_uri.lua
+++ b/apisix/http/router/radixtree_host_uri.lua
@@ -78,11 +78,10 @@ local function push_host_router(route, host_routes, only_uri_routes)
         vars = route.value.vars,
         filter_fun = filter_fun,
         handler = function (api_ctx, match_opts)
-            local matched = core.table.deepcopy(match_opts.matched)
             api_ctx.matched_params = nil
             api_ctx.matched_route = route
-            api_ctx.curr_req_matched = matched
-            api_ctx.real_curr_req_matched_path = matched._path
+            api_ctx.curr_req_matched = match_opts.matched
+            api_ctx.real_curr_req_matched_path = match_opts.matched._path
         end
     }
 
@@ -166,7 +165,7 @@ function _M.matching(api_ctx)
     match_opts.remote_addr = api_ctx.var.remote_addr
     match_opts.vars = api_ctx.var
     match_opts.host = api_ctx.var.host
-    match_opts.matched = {}
+    match_opts.matched = core.tablepool.fetch("matched_route_record", 0, 4)
 
     if host_router then
         local host_uri = api_ctx.var.host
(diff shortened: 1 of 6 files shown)


apisix/http/router/radixtree_host_uri.lua
```diff
@@ -162,12 +161,12 @@ end
 function _M.matching(api_ctx)
     core.log.info("route match mode: radixtree_host_uri")
 
-    core.table.clear(match_opts)
+    local match_opts = core.tablepool.fetch("route_match_opts", 0, 16)
     match_opts.method = api_ctx.var.request_method
     match_opts.remote_addr = api_ctx.var.remote_addr
     match_opts.vars = api_ctx.var
     match_opts.host = api_ctx.var.host
-    match_opts.matched = core.tablepool.fetch("matched_route_record", 0, 4)
```
monkeyDluffy6017, line 170: Why do you modified this?

apisix/http/router/radixtree_host_uri.lua
```diff
@@ -78,10 +78,11 @@ local function push_host_router(route, host_routes, only_uri_routes)
         vars = route.value.vars,
         filter_fun = filter_fun,
         handler = function (api_ctx, match_opts)
+            local matched = core.table.deepcopy(match_opts.matched)
```
monkeyDluffy6017, line 81: Why do you modified this?

apisix/http/router/radixtree_host_uri.lua
AlinsRan: Nesting memory requests from tablepool is a bad style and requires attention to the relationship between child and parent tables.
It may sacrifice some performance.

apisix/http/router/radixtree_host_uri.lua
AlinsRan: After routing matching, the table will be cleared.

apisix/http/router/radixtree_host_uri.lua
AlinsRan: I have decided to prioritize performance. 
udpate!

apisix/http/router/radixtree_host_uri.lua
AlinsRan: I have decided to prioritize performance. 
udpate!

monkeyDluffy6017 (approved):

leslie-tsang (approved):

Issue #10209, opened 2023-09-15
feat: support log with traceid when enable the zipkin plugin 


wizhuo:
### Description

feat: As a user, I would like to automatically include the traceId when logging, especially when I have enabled the Zipkin plugin. This will make it easier for me to quickly identify the logs associated with a particular traceId when troubleshooting issues. Currently, manually concatenating the traceId with each log entry is cumbersome and not user-friendly. I hope that this issue can be addressed at the architectural level.

wizhuo:
#10210

PR #8871, opened 2023-02-16
feat(opentelemetry): add opentelemetry variables

lework:
### Description

Span information recorded in the access log.

Fixes #  https://github.com/apache/apisix/issues/8765

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #10210, opened 2023-09-15
feat: support log with traceid when enable the zipkin plugin

wizhuo:
### Description

As a user, I would like to automatically include the traceId when logging, especially when I have enabled the Zipkin plugin. This will make it easier for me to quickly identify the logs associated with a particular traceId when troubleshooting issues. Currently, manually concatenating the traceId with each log entry is cumbersome and not user-friendly. I hope that this issue can be addressed at the architectural level.

feat #10209 

### Checklist

### DIFF
--- a/apisix/core/log.lua
+++ b/apisix/core/log.lua
@@ -61,6 +61,12 @@ local function update_log_level()
     end
 end
 
+local function get_trace_id()
+    if ngx_get_phase() ~= "init" and ngx.config.subsystem == "http"  then
+        return ngx.ctx.trace_id
+    end
+    return ''
+end
 
 function _M.new(prefix)
     local m = {version = _M.version}
@@ -91,6 +97,7 @@ function _M.new(prefix)
 end
 
 
+
 setmetatable(_M, {__index = function(self, cmd)
     local log_level = log_levels[cmd]
     local method
@@ -101,7 +108,7 @@ setmetatable(_M, {__index = function(self, cmd)
         method = do_nothing
     else
         method = function(...)
-            return ngx_log(log_level, ...)
+            return ngx_log(log_level, get_trace_id(), " ", ...)
         end
     end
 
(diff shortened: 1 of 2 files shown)


wizhuo:
#10210

monkeyDluffy6017:
Thank you for your contribution, but not everyone will be using the zipkin plugin. Perhaps you can modify it this way in your own project.

wizhuo:
> Thank you for your contribution, but not everyone will be using the zipkin plugin. Perhaps you can modify it this way in your own project.

Yes, indeed, not everyone will enable Zipkin, so the current changes should not affect users who don't use Zipkin. We've simply made user-friendly adjustments for those using Zipkin . after all, when Zipkin is in use, users expect traceId to be automatically recorded in the logs. This makes troubleshooting more convenient for everyone. I believe this capability would also be well-received by other users. What do you think?

monkeyDluffy6017:
There are some other similar plugins like `request-id`, and there will be more in the future, we should not assume that users will use a specific plugin.

wizhuo:
> There are some other similar plugins like `request-id`, and there will be more in the future, we should not assume that users will use a specific plugin.

Sounds reasonable, then I'll close it.

Gallardot:
Hi @wizhuo , I understand the purpose of your PR and I also use zipkin plugin. Perhaps you can refer to this [opentelemetry plugin PR](https://github.com/apache/apisix/pull/8871) for an implementation of the zipkin plugin. Thank you for your contribution.

wizhuo:
> Hi @wizhuo , I understand the purpose of your PR and I also use zipkin plugin. Perhaps you can refer to this [opentelemetry plugin PR](https://github.com/apache/apisix/pull/8871) for an implementation of the zipkin plugin. Would you like to help the community to explore this? Thank you for your contribution.

Yes，I would like to do

PR #10291, opened 2023-10-05
feat: support secret fetching in authz-keycloak

Revolyssup:
### Description

Currently $secret uri cannot be used in authz-keycloak plugin. This PR adds support for it.

### Checklist


apisix/plugins/authz-keycloak.lua
```diff
@@ -20,6 +20,7 @@ local sub_str   = string.sub
 local type      = type
 local ngx       = ngx
 local plugin_name = "authz-keycloak"
+local secret         = require("apisix.secret")
```
shreemaan-abhishek, line 23:
```suggestion
local secret        = require("apisix.secret")
```

t/plugin/authz-keycloak.t
```diff
@@ -645,3 +649,88 @@ GET /t
...
+                                "127.0.0.1:1982": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/api/token"
+                }]]
+            )
+
+            if code >= 300 then
+                ngx.status = code
```
shreemaan-abhishek, line 708: return when there is a failure.
```suggestion
                ngx.status = code
                return ngx.say(body)
```

t/plugin/authz-keycloak.t
shreemaan-abhishek, line 1: better add new tests in `t/plugin/authz-keycloak3.t`

shreemaan-abhishek (changes requested):

monkeyDluffy6017:
Please make the ci pass

Revolyssup:
> Please make the ci pass

done

apisix/plugins/authz-keycloak.lua
```diff
@@ -16,11 +16,13 @@
 --
 local core      = require("apisix.core")
 local http      = require "resty.http"
+local secret    = require("apisix.secret")
```
shreemaan-abhishek, line 19: I'm not sure but importing just `fetch_secrets()` (instead of the whole module) might be more efficient. cc: @monkeyDluffy6017

t/plugin/authz-keycloak4.t
```diff
@@ -0,0 +1,112 @@
...
+                                "127.0.0.1:1982": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/api/token"
+                }]]
+            )
+
+            if code >= 300 then
+                ngx.status = code
```
shreemaan-abhishek, line 84: return when there is a failure.
```suggestion
                ngx.status = code
                return ngx.say(body)
```

shreemaan-abhishek (changes requested):

Sn0rt:
pls add user document for guide

apisix/plugins/authz-keycloak.lua
monkeyDluffy6017: @Revolyssup please check this

Revolyssup: done

Revolyssup:
> pls add user document for guide

Updated the auth-keycloak plugin doc

monkeyDluffy6017 (approved):

docs/en/latest/plugins/authz-keycloak.md
```diff
@@ -48,7 +48,7 @@ Refer to [Authorization Services Guide](https://www.keycloak.org/docs/latest/aut
...
-| client_secret                                | string        | False    |                                               |                                                                    | The client secret, if required.                                                                                                                                                                                                                       |
+| client_secret                                | string        | False    |                                               |                                                                    | The client secret, if required. You can use APISIX secret to store and reference this value. APISIX currently supports storing secrets in two ways. [Environment Variables](#use-environment-variables-to-manage-secrets) and [HashiCorp Vault](#use-vault-to-manage-secrets)                                                                                                                                                                                                                       |
```
shreemaan-abhishek, line 51: The links are wrong. And please add an example as well.

shreemaan-abhishek (changes requested):

docs/en/latest/plugins/authz-keycloak.md
Revolyssup: I fixed the doc link. Though the lint points to the secret page which has the example in detail. So I think it will be redundant to add it again here.

shreemaan-abhishek: Does this PR introduce support to use env vars for accessing secrets? If yes, please add tests as well.

Regarding the docs, it would be nice to have an example like this:

(existing example that uses hard-coded value for secrets)

You can also set secrets using the secret resource to avoid having sensitive information in plain text <-- add this line

(add a basic configuration here)

See (add link to secrets.md documentation) for more information.

Revolyssup: @shreemaan-abhishek Though secret is the APISIX abstraction. Behind it can be vault or env variable or anything. So I thought just testing with any one of them should work. Okay I will add tests for env variable as well.

Revolyssup: okay i will modify the test case

Sn0rt:
Please take a look at this case, I think it might be better understood if you can directly check the value after reslove. https://github.com/apache/apisix/blob/1434335041c6e054f7b6afb0401aea0a069b0ad7/t/plugin/authz-keycloak3.t#L168

t/plugin/authz-keycloak4.t
```diff
@@ -0,0 +1,212 @@
...
+VAULT_TOKEN='root' VAULT_ADDR='http://0.0.0.0:8200' vault kv put kv/apisix/foo client_secret=d1ec69e9-55d2-4109-a3ea-befa071579d5
+--- response_body
+Success! Data written to: kv/apisix/foo
+
+
+
+=== TEST 2: set client_secret as a reference to secret
+--- yaml_config
+apisix:
+    data_encryption:
```
Sn0rt, line 55: why need this ?

Sn0rt (approved):

t/plugin/authz-keycloak4.t
Revolyssup: the test case whose configuration I picked also had this so I copy pasted. Should I remove it?

Revolyssup:
> Please take a look at this case, I think it might be better understood if you can directly check the value after reslove.
> 
> https://github.com/apache/apisix/blob/1434335041c6e054f7b6afb0401aea0a069b0ad7/t/plugin/authz-keycloak3.t#L168

@Sn0rt This variable extracts the exact string which is not a resolved secret. So the value of this will be `$secret://...something`. Here accessing this field doesn't give the resolved secret. So I am reverting back to previous test.

Revolyssup:
Closing this PR. Will recreate this.

PR #10339, opened 2023-10-16
fix: add missing properties to schema

pottekkat:
### Description

Adds missing properties to the schema so that it can be used for validating configuration files more accurately.

Also consolidates all the schemas to a single schema instead of having to validate with multiple schemas.

### Checklist


pottekkat:
I'm unsure why the tests are failing.

apisix/cli/schema.lua
```diff
@@ -246,18 +405,225 @@ local config_schema = {
...
+                            type = "string"
+                        },
+                        client_header_timeout = {
+                            type = "string"
+                        },
+                        client_body_timeout = {
+                            type = "string"
+                        },
+                        client_max_body_size = {
+                            type = "number"
```
Revolyssup, line 573:
```suggestion
                            type = "string"
```

Revolyssup (changes requested):

apisix/cli/schema.lua
pottekkat: Let me try. I originally had it as string but it threw an error.

[`f3ecf05` (#10339)](https://github.com/apache/apisix/pull/10339/commits/f3ecf055e025c1b92abc4a0804a9403e31ea8a68)

pottekkat:
@Revolyssup I got this error before so I tried changing it:

https://github.com/apache/apisix/actions/runs/6549717930/job/17787244082?pr=10339#step:5:287

pottekkat:
@Revolyssup Should we add a case where it can be either a string or a number?

monkeyDluffy6017:
LGTM, That's a pretty big change. So let's make the ci pass

apisix/cli/schema.lua
```diff
@@ -302,110 +683,87 @@ local config_schema = {
...
-            properties = {
-                items = {
-                    properties = {
-                        name = {type = "string"},
-                        key = {type = "string"},
-                        role = {type = "string"},
+            dependencies = {
+                role = {
+                    oneOf = {
+                        {
```
Revolyssup, line 709: @pottekkat Will you make sure that this dependency schema fits right with this[1] yaml snippet here?
1. https://github.com/apache/apisix/blob/88406dceab26ed7f25b663088b0b5e760bd45a38/t/cli/test_validate_config.sh#L210C1-L210C1

Revolyssup (changes requested):

pottekkat:
I reverted to the same changes as in the last commit where the tests were passing in this commit: https://github.com/apache/apisix/pull/10339/commits/bf8bdb87649dc18cabd85624ce301d83991e65a2

But now, the tests fail.

I'm now trying by removing all changes and checking if the tests pass now and build from there.

pottekkat:
All checks pass with these changes.

monkeyDluffy6017 (approved):

monkeyDluffy6017:
@Revolyssup please help to check

Revolyssup (approved):

PR #10342, opened 2023-10-16
fix: fix consul plugin worker will not exit while reload or quit

theweakgod:
### Description

**1.configure**

config.yaml

discovery:
  consul:
    servers:
       - "http://localhost:8500"
    keepalive: true

Version: 3.6.0

**2.situation**
Using the consul service, I found that there is a process leak. When reloading or quitting, there will be a nginx process shutting down and never exit.
When nginx reloads, the consul worker with keepalive turned on will not exit normally, will be in shutting down state, and will always access consul's health api.
You only need to make a judgment before creating a new timer to solve the process leak.
This problem does not exist when keepalive is false, but there will be an obvious memory leak.

htop -p $(pgrep -d',' -f "nginx")
image
process will not exit!

To process exit when testing, please change consul node state.

**3.testing**
start consul discovery, and try to `apisix reload && htop -p $(pgrep -d',' -f "nginx")`.  The process in shutting down, and nerver exit even if you set timeout.
after pull this commit, set shutting down timeout 300 seconds or change consul node info, the shutting down process will exit after node info changing or timeout.

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/discovery/consul/init.lua
+++ b/apisix/discovery/consul/init.lua
@@ -29,6 +29,7 @@ local tonumber           = tonumber
 local pairs              = pairs
 local ngx_timer_at       = ngx.timer.at
 local ngx_timer_every    = ngx.timer.every
+local exiting            = ngx.worker.exiting
 local log                = core.log
 local json_delay_encode  = core.json.delay_encode
 local ngx_worker_id      = ngx.worker.id
@@ -530,7 +531,9 @@ function _M.connect(premature, consul_server, retry_delay)
                 health_res.headers['X-Consul-Index'])
     end
 
-    check_keepalive(consul_server, retry_delay)
+    if not exiting() then
+        check_keepalive(consul_server, retry_delay)
+    end
 end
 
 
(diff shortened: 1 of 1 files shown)


theweakgod:
version: 3.6.0

theweakgod:
测试的时候在reload之后如果是还有nginx处于shutting down，可以更新一下consul的节点，这样worker就会退出了，因为worker此时处于http请求状态，请求完之后就会退出了😬

shreemaan-abhishek:
I checked this, the worker process still does not exit. Do you have a clue on how we can make the worker process exit automatically instead of worker reset?

theweakgod:
> I checked this, the worker process still does not exit. Do you have a clue on how we can make the worker process exit automatically instead of worker reset?

You can update consul node(eg. register a server). Because worker is blocked by http request. Or wait 300 seconds for request finished. Then worker will exit.
[reference](https://groups.google.com/g/openresty/c/HZGDE7XODeA)

theweakgod:
> During testing, if nginx is still shutting down after reload, you can update the consul node so that the worker will exit, because the worker is in the http request state at this time and will exit after the request is completed.😬

Check this. @shreemaan-abhishek

theweakgod:
Is there something wrong with this error check?

shreemaan-abhishek (approved):

monkeyDluffy6017 (approved):

PR #10363, opened 2023-10-19
feat: add method and status attr in prometheus

hugoray:
add method and status attr in prometheus

### Description

add method and status attribute in prometheus module

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/prometheus/exporter.lua
+++ b/apisix/plugins/prometheus/exporter.lua
@@ -54,8 +54,10 @@ end
 
 local plugin_name = "prometheus"
 local default_export_uri = "/apisix/prometheus/metrics"
--- Default set of latency buckets, 1ms to 60s:
-local DEFAULT_BUCKETS = {1, 2, 5, 10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000, 30000, 60000}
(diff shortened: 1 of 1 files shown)


monkeyDluffy6017:
Why we need this?

hugoray:
It fixes the prometheus labels inconsistency problem.

monkeyDluffy6017:
Could you explain the problem more clearly?

hugoray:
In promethus, we often need to specify the request type and the http_ret/status of one HTTP request.  By adding this two attributes,  the dashboard could draw new histogram based on them.

monkeyDluffy6017:
Maybe you could keep this in your own project? The metrics that each user cares about vary, and there's no way we can include all of them

hugoray:
I do think the change of metrics can help people get the accurate message of HTTP request, no matter the scene who use, need to get these two attributes when using prometheus.

nic-chen:
hi @hugoray 

There are enough metrics and metrics labels currently exposed, and some users have performance problems due to excessive data volume during use. 

I think the better solution is to customize the required metrics and labels by configuration of the prometheus plugin. Would you like to contribute this feature?

monkeyDluffy6017:

PR #10393, opened 2023-10-25
change(opentelemetry): make span name and attributes follow the standard spec

indrekj:
## Description

The previous implementation did not [follow the spec][1]. I'll list the reasoning below.

Span name:

> HTTP server span names SHOULD be {http.method} {http.route} if there
is a (low-cardinality) http.route available. HTTP server span names SHOULD be {http.method} if there is no (low-cardinality) http.route available. HTTP client spans have no http.route attribute since client-side instrumentation is not generally aware of the "route", and therefore HTTP client spans SHOULD use {http.method}. Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.

It's clearly mentioned that the span name MUST NOT use URI path, but the previous implementation did exactly that. Now it uses `http.method http.route` where the latter is added only when available.

`http.status_code`, `http.scheme`, `http.target`, `http.method`, and `net.host.name` are required according to the spec.

`http.route` is required if and only if it is available.

`http.user_agent` is recommended.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#http-server

## Motivation

The most important thing for us is the span name. The previous one had unlimited cardinality which would be a very bad idea in production (it can slow down the tracing platform and when using 3rd party services it could create huge bills).

### Checklist

  Not relevant.
  There obviously is a change, but I would consider it backward-compatible (or at least very low-risk one). I named this as `feat` but could also say it's a bug fix to make sure we are aligned with the OTel spec

### DIFF
--- a/apisix/plugins/opentelemetry.lua
+++ b/apisix/plugins/opentelemetry.lua
@@ -308,18 +308,35 @@ end
 
 
 function _M.rewrite(conf, api_ctx)
+    local vars = api_ctx.var
+
     local tracer, err = core.lrucache.plugin_ctx(lrucache, api_ctx, nil, create_tracer_obj, conf)
     if not tracer then
         core.log.error("failed to fetch tracer object: ", err)
         return
     end
 
-    -- extract trace context from the headers of downstream HTTP request
-    local upstream_context = trace_context_propagator:extract(context, ngx.req)
+    local span_name = vars.method
+
     local attributes = {
-        attr.string("service", api_ctx.service_name),
-        attr.string("route", api_ctx.route_name),
+        attr.string("net.host.name", vars.host),
+        attr.string("http.method", vars.method),
+        attr.string("http.scheme", vars.scheme),
+        attr.string("http.target", vars.request_uri),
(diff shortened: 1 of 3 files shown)


monkeyDluffy6017:
Thanks for your contribution! we will check this later

monkeyDluffy6017:
@yangxikun Could you help to check this pr?

indrekj:
@monkeyDluffy6017 @yangxikun friendly ping :bow: , one of the last things we would like to get in before we can deploy Apisix to non-test envs.

monkeyDluffy6017:
@indrekj I will check this in this week

yangxikun:
Approve, Thanks for the contribution!

monkeyDluffy6017:
@yangxikun click here if you have checked
image

monkeyDluffy6017 (approved):

yangxikun (approved):

AlinsRan (approved):

PR #10403, opened 2023-10-26
fix: make install failed on mac

AllenFang:
### Description

Since mac use M1/M2(arm64) chip, apisix will be always installed failure when execute `make install`, it's not about the permission, it's new system constraint and it doesn't work even mac user unlock it.

The change is adjust

* `ENV_INST_BINDIR` path from `/usr/bin` to `/usr/local/bin` for Mac arm64 only
* `ENV_INST_LUADIR` path from `/usr/share/lua/5.1` to `/opt/homebrew/Cellar/lua/{version}` for Mac arm64 only

Ideally for those non-distributed managed package, we better put in `/usr/local/bin` rather than `/usr/bin` regardless Linux/MacOS, `/usr/local/` is more for those self install packages. Anyway, this change only affect mac arm64 developer, please review. thank you

Fixes # (issue)
intend to fix https://github.com/apache/apisix/issues/6225 issue permanently

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,8 @@ endif
 ifeq ($(ENV_OS_NAME), darwin)
 	ifeq ($(ENV_OS_ARCH), arm64)
 		ENV_HOMEBREW_PREFIX := /opt/homebrew
+		ENV_INST_BINDIR := $(ENV_INST_PREFIX)/local/bin
+		ENV_INST_LUADIR := $(shell which lua | xargs realpath | sed 's/bin\/lua//g')
 	endif
 
 	# OSX archive `._` cache file


membphis:
I do not recommend installing APISIX via "make install" for mac os ENV

monkeyDluffy6017 (approved):

monkeyDluffy6017:
@membphis It make some sense when other services are dependent on APISIX

AllenFang:
> I do not recommend installing APISIX via "make install" for mac os ENV

May I know the reasons ? 🙏

jiangfucheng (approved):
LGTM, tested it on my mac.

PR #10404, opened 2023-10-26
refactor: add log after apisix.yml config reload in standalone mode

### LINKED ISSUE #10402: feat: As a user, I want to see a log message after apisix.yml config reload in standalone, so that I know if everything went fine
### Description

feat: As a user, I want to see a log message after config reload in standalone, so that I know if everything went fine.

Right now, only errors are logged in case apisix.yml reload fails. There's no indication of sucesful re-loading. Having a success log message would help to confirm when the new configuration is applied after reload.

markokocic:
Fixes #10402

Please review. Not sure why is the log message printed multiple times and how to avoid it.

### Checklist

### DIFF
--- a/apisix/core/config_yaml.lua
+++ b/apisix/core/config_yaml.lua
@@ -115,6 +115,7 @@ local function read_apisix_yaml(premature, pre_mtime)
 
     apisix_yaml = apisix_yaml_new
     apisix_yaml_ctime = last_change_time
+    log.warn("config file ", apisix_yaml_path, " reloaded.")
 end
 
 


monkeyDluffy6017:
The error log will be flooded by you newly added message.
```
 ngx.timer.every(1, read_apisix_yaml)
```

markokocic:
Hi @monkeyDluffy6017 
read_apisix_yaml is indeed called every second, but there's already a guard in place to skip execution if file is not changed:

``` lua
    local last_change_time = attributes.change
    if apisix_yaml_ctime == last_change_time then
        return
    end
```
Log statement is added at the end of the method, after successful execution.

monkeyDluffy6017:
I don't think it's very necessary, we don't have this log when use etcd too, the number of logs is also very lagre even the log is printed when you send a new configuration

markokocic:
This additional log entry won't much increase the log size, since it's applied only when config file changes, which is not very frequent, depending on the usage scenario.

If we add it, the benefit would be that we can confirm when changes are applied successfully.

As for the log verbosity, I created a new ticket that is not related to this topic, but for log entry duplication in general. See #10414

monkeyDluffy6017:
Some of our users change their configurations very frequently, like thousands of times in 1s

markokocic:
> Some of our users change their configurations very frequently, like thousands of times in 1s

In `apisix.yml` in standalone mode? I doubt how is that supposed to work, given that APISIX reads the config file only once per second.

monkeyDluffy6017:
What I mean is that even in standalone mode, configuration changes are frequent and should not be logged every time a configuration change is made without an error. This information is meaningless to most users. If you think this is necessary, you can keep this change yourself.

monkeyDluffy6017:

polaru:
I agree with @markokocic that when running apisix in standalone mode you don't get a lot of reloads; if ones needs that, they would use etcd or another config plane, not a static file. 

@monkeyDluffy6017: running apisix in standalone mode in a kubernetes pod, getting `apisix.yaml` from a configMap, I would like to know when the pod has finished loading the configuration so I can start sending traffic through it. how do you suggest to accomplish this?

> If you think this is necessary, you can keep this change yourself.
@monkeyDluffy6017 what do you mean by this? fork apisix?

monkeyDluffy6017:
> if ones needs that, they would use etcd or another config plane, not a static file.

Sounds reasonable.

AlinsRan:
> @monkeyDluffy6017: running apisix in standalone mode in a kubernetes pod, getting `apisix.yaml` from a configMap, I would like to know when the pod has finished loading the configuration so I can start sending traffic through it. how do you suggest to accomplish this?
> 

I agree with your point that static files usually do not change frequently.
We need to confirm whether the apisix.yaml configuration has been updated through logs.

monkeyDluffy6017 (approved):

AlinsRan (approved):

PR #10413, opened 2023-10-30
docs: make route matching/filtering information more visible

shreemaan-abhishek:
### Description

Information about grapql based route matching is almost invisble/unreachable. This has been fixed.

Fixes # (issue)

### Checklist


pottekkat (approved):

docs/en/latest/terminology/route.md
```diff
@@ -36,6 +36,12 @@ A Route mainly consists of three parts:
 2. Plugin configuration (current-limit, rate-limit)
 3. Upstream information
 
+:::note
+
+APISIX also supports GraphQL based route filtering, [see here](../../latest/router-radixtree.md#how-to-filter-route-by-graphql-attributes).
```
kayx23, line 41: While the [radix tree doc](https://apisix.apache.org/docs/apisix/router-radixtree) is indeed hard to find, I found it a bit strange to drop an admonition here just for "graphql based route filtering". In fact all these are not mentioned:

image

and what's missing from the above doc is the mention of APISIX variables (in addition to NGINX variables)

I would also see if we can mention these information in this doc without admonitions.

monkeyDluffy6017: @shreemaan-abhishek please check this

moonming:
@shreemaan-abhishek ping

docs/en/latest/terminology/route.md
shreemaan-abhishek: @kayx23 instead, I just moved the whole information to the `routes` page.

kayx23: @shreemaan-abhishek ok but I wouldn't remove `router-radixtree` doc entirely - it has important information about prefix matching and so on.

image

kayx23 (changes requested):
I wouldn't remove `router-radixtree` doc entirely - it has important information about prefix matching and so on.

image

shreemaan-abhishek:
> I wouldn't remove router-radixtree doc entirely - it has important information about prefix matching and so on.

I have moved all information to `route.md` no information is lost.

kayx23 (commented):
What's everybody's take on [lumping all these info into Routes](https://github.com/apache/apisix/blob/3a515dbd26ad67707e1a59e882b1f4f1d0a7ce7e/docs/en/latest/terminology/route.md)?

@pottekkat @monkeyDluffy6017 @moonming

monkeyDluffy6017:
@kayx23 it seems reasonable, users don't care what radix-tree is, they just want to how to configure the route

monkeyDluffy6017 (approved):

monkeyDluffy6017:
@kayx23 It's up to you to decide whether or not to merge.

kayx23:
A few more comments:

1. Apache APISIX doc has both **Route** and **Router** under terminology. I would, therefore, suggest moving the first half of the `router-radixtree.md` content into **Router** instead of **Route** as it is currently now.
2. The `docs/en/latest/router-radixtree.md` doc removed has a corresponding Chinese doc. More chores yes, but it should be updated for the change as well to avoid update anomalies.
3. Personally, I'm OK with the change once the above are addressed. As a community driven project, it probably should not be my sole decision, however. I'd tag a few contributors/stakeholders for food of thoughts.

monkeyDluffy6017:
@kayx23  I agree with you, I'm not a professional when it comes to documentation, whether or not to merge this pr still has to follow the community's opinion, please help track it down!

kayx23:
People I can think of: @pottekkat @nfrankel @yzeng25 @SylviaBABY @Neilblaze 

Any thought for this PR in addition to my [comment](https://github.com/apache/apisix/pull/10413#issuecomment-1821796385) here? If that sounds good we can move forward?

nfrankel:
@kayx23 It's a healthy discussion

I think "how to" configure a route should be more important than how it's implemented. The former needs to be a top-level item, while the later can be buried deeper.

However, it's perhaps better to address it in a different PR to avoid pulling an infinite thread, one thing leading to another.

shreemaan-abhishek:
@kayx23 could you please update the chinese docs?

yzeng25:
I'd suggest keep redundancies in 3.x doc and fully delete redundant contents in `radixtree.md` in 4.0 doc. If you really really want to delete `radixtree.md`, it is better to take this action in 4.0 and mention it in 4.0 changelog.

image

nfrankel:
Kudos for the diagram explanation @yzeng25

kayx23:
Thanks for everyone's input and the diagram.

It looks like what's suggested here is to have the same content across both docs until the `router-radixtree.md` can be fully removed in v4. with that I have two questions for @yzeng25 :

1. Do you have any references to back up this doc strategy and for us to learn more that it's a good practice? 
2. Assume we move forward with this suggestion, how do we plan to keep track for v4 and not confuse users for having the same content across two places?

Cheers.

nfrankel:
>Assume we move forward with this suggestion, how do we plan to keep track for v4 and not confuse users for having the same content across two places?

* The user is not confused if they have the same content across two places
* They will be confused if the content is different
* Two solutions:
** Either a link from one document to another
** Or synchronize the same content in different places

The link option is IMHO adverse to UX: it breaks the flow of reading. I think we should favor synchronization.

The answer to synchronization is includes. I don't know if it's an option in the existing system, but we should definitely go this way.

kayx23:
I have no objections but this should be tracked somehow, perhaps an issue? 

Also, I should mention that I'm participating as a contributor here and my primary focus is still API7 docs. So @shreemaan-abhishek while I can help facilitate the conversation, I can't finish the rest of the work for every doc PR. If the Chinese doc needs work, you can open new issues.

yzeng25:
> Do you have any references to back up this doc strategy and for us to learn more that it's a good practice?

Keeping UX consistency across all 3.x docs leads to the redundancies or synchronization strategy.

nfrankel:
After having to check the matching options in the Router Radix Tree page, I confirm that it's poor UX and having a dedicated Matching page instead (or in addition) can only improve the situation.

kayx23:
Ok can I understand and summarize the above suggested action above as to:

* keep the matching rules of `radixtree.md` page (aka first half of that page) intact, for at least 3.x docs, and optionally include the same content in the `terminology/router.md`; and
* move the three examples to `routes.md`, which is already done in this PR and the original motivation of this PR?

shreemaan-abhishek:
> Apache APISIX doc has both Route and Router under terminology. I would, therefore, suggest moving the first half of the router-radixtree.md content into Router instead of Route as it is currently now.

If someone wants to understand how routes are matched, they will refer the `route` doc instead of `router` or `radixtree`. So (not self praising 😂) but IMHO, the current state is ideal (the links with chinese docs need to be addressed tho)

@yzeng25's feedback is good but unimportant maintenance burden.

All in all this PR is too basic for a long discussion, and if the community thinks this PR makes the situation better can we just go ahead with approvals?

shreemaan-abhishek:
@kayx23, @pottekkat, @yzeng25 WDYT?

kayx23:
I am neutral because I struggle to see the pros outweigh the cons.

yzeng25:
I suggest leaving these changes to V4.0 milestone to deliver consistency among V3.x docs and consistency across EN and CN docs. Syncing multi-lang docs is also a huge workload.

shreemaan-abhishek:
I will open an issue instead to track this.

yzeng25:
> I will open an issue instead to track this.

@shreemaan-abhishek Could you refer the issue link?

shreemaan-abhishek:
@yzeng25 https://github.com/apache/apisix/issues/10735

PR #10377, opened 2023-10-23
fix: Solve the 403 problem caused by too large a request to upload the attachment body,error message"send() failed (32: Broken pipe)"

lplcc:
Solve the 403 problem caused by too large a request to upload the attachment body,error message"send() failed (32: Broken pipe)
#10375

Issue #10375, opened 2023-10-23
bug: When the request body is too large, forward-auth returns 403


lplcc:
### Current Behavior

When the request body is too large, forward-auth returns 403

### Expected Behavior

Hope to support the upload of more than 10M files

### Error Logs

send() failed (32: Broken pipe), client: 10.10.103.41, server: _, request: "POST /jljftest/v1/paas/service/meter HTTP/1.1", host: "172.70.2.82:9080"
[lua] forward-auth.lua:125: phase_func(): failed to process forward auth, err: broken pipe, client: 10.10.103.41, server: _, request: "POST /jljftest/v1/paas/service/meter HTTP/1.1", host: "172.70.2.82:9080"

### Steps to Reproduce

1,Configuring routes
2,add forward-auth plugin
3,Calls the gateway interface to upload a request containing a base64 string

### Environment

- APISIX version (run `[APISIX/3.6.0]`):
- Operating system (run `linux`):
- OpenResty / Nginx version (run `openresty/1.21.4.2` or `nginx -V`):
- etcd version, if relevant (run `3.5.0`):
- APISIX Dashboard version, run 3.0.1:
- Plugin runner version, for issues related to plugin runners: forward-auth 3.6.0
- LuaRocks version, for installation issues (run `3.8.0`):

shreemaan-abhishek:
> When the request body is too large, forward-auth returns 403

Can you describe the use case? Why would an auth request carry a too large body?

Revolyssup:
@lplcc Unlike Nginx, APISIX does not limit the body size by default.
what is the value of `client_max_body_size` in your apisix configuration. You can set it to 0(for infinite)

lplcc:
> > When the request body is too large, forward-auth returns 403
> 
> Can you describe the use case? Why would an auth request carry a too large body?

Our API is going to upload an attachment, which is done through a base64 string

monkeyDluffy6017:
Try to increase the [conf.keepalive](https://github.com/apache/apisix/blob/455804771124c15ff909e49ad8dfdd961f0e4a88/apisix/plugins/forward-auth.lua#L113) and the [conf.timeout](https://github.com/apache/apisix/blob/455804771124c15ff909e49ad8dfdd961f0e4a88/apisix/plugins/forward-auth.lua#L119)

lplcc:
[...]

PR #10425, opened 2023-10-31
fix:Solve the 403 problem caused by the post request body is too large

lplcc:
### Description

The gateway returns 403 forbidden in the forward-auth.lua file because the post request body is too large

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -117,7 +117,8 @@ function _M.access(conf, ctx)
 
     local httpc = http.new()
     httpc:set_timeout(conf.timeout)
-
+    local client_body_reader, err = httpc:get_client_body_reader()
+    params.body=client_body_reader
     local res, err = httpc:request_uri(conf.uri, params)
     if not res and conf.allow_degradation then
         return


Revolyssup:
THis is recreation of this PR https://github.com/apache/apisix/pull/10377/files trying to solve https://github.com/apache/apisix/issues/10375

lplcc:
but I can't solve this error  “ [error] 98340#98340: *40 [lua] init.lua:551: handle_upstream(): failed to set upstream: missing upstream configuration in Route or Service”,
 I just added two  lines of code in the forward-auth.lua file

apisix/plugins/forward-auth.lua
```diff
@@ -117,7 +117,8 @@ function _M.access(conf, ctx)
 
     local httpc = http.new()
     httpc:set_timeout(conf.timeout)
-
+    local client_body_reader, err = httpc:get_client_body_reader()
+    params.body=client_body_reader
```
monkeyDluffy6017, line 121: Why do you add this?

monkeyDluffy6017:
I will close this pr first, let's discuss this in the issue

monkeyDluffy6017:
Please make the ci pass

shreemaan-abhishek:
I am checking this, the current fix is causing some other failures that need to be fixed first.

apisix/plugins/forward-auth.lua
Gallardot: Agree with @monkeyDluffy6017 . Additionally, if this is a GET request, this approach may not be appropriate.

shreemaan-abhishek: @Gallardot `conf.timeout` or other conf parameters do not fix this issue and the root cause seems mysterious; I haven't been able to find it out.

> Additionally, if this is a GET request, this approach may not be appropriate.

Not a big deal, a condition check can be used. That said, there should be a better way to address this.

lplcc: Determine the request method or request body size, but this threshold is not easy to determine

apisix/plugins/forward-auth.lua
```diff
@@ -106,16 +106,20 @@ function _M.access(conf, ctx)
         method = conf.request_method
     }
 
+    local httpc = http.new()
     if params.method == "POST" then
         params.body = core.request.get_body()
+        local client_body_reader, err = httpc:get_client_body_reader()
+        if not err then
+            params.body = client_body_reader
+        end
```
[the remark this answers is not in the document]
shreemaan-abhishek: That won't work  if the `Content-Length` header is missing as `client_body_reader` requires it to be provided.

Apparently, reading `req_body` implicitly fixes this problem.

monkeyDluffy6017:
@shreemaan-abhishek is this pr ready?

shreemaan-abhishek:
@monkeyDluffy6017 yes.

t/plugin/forward-auth/fw-auth-large-file.bin
monkeyDluffy6017, line 1: It's too big for the project, could you create the file by code when run test cases?

apisix/plugins/forward-auth.lua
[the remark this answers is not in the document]
monkeyDluffy6017: I agree with @Gallardot, it's ok to use `core.request.get_body()`  if the `content-length` is missing, the client should ensure the input is correct. 
But we can't check the `err` value of `local client_body_reader, err = httpc:get_client_body_reader()`, because the `err` will be `nil` if the `content-length` is missing. https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L999

```
local httpc = http.new()
if params.method == "POST" then
    local client_body_reader, err = httpc:get_client_body_reader()
    if client_body_reader then
        params.body = client_body_reader
    else
        params.body = core.request.get_body()
    end
end
```

t/plugin/forward-auth.t
```diff
@@ -374,3 +403,34 @@ GET /hello
...
+                }
+            )
+
+            if not res then
+                ngx.log(ngx.ERR, err)
+                return
+            end
+        }
+    }
+--- error_code: 200
```
monkeyDluffy6017, line 436: Could you test the `response body` or the `response header` too?

apisix/plugins/forward-auth.lua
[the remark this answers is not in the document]
shreemaan-abhishek: if we go this way other test cases would need to be updated. This might result in a breaking change for users 🤔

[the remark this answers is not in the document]
shreemaan-abhishek: my bad, the  request already carries the content-length and other necessary headers... we just need to make sure that these headers are also sent to the auth server

monkeyDluffy6017:
@shreemaan-abhishek please fix the code lint error

t/plugin/forward-auth/fw-auth-large-file.bin
shreemaan-abhishek: done

Gallardot (approved):
LGTM

Sn0rt (approved):

monkeyDluffy6017:
It's fixed by https://github.com/apache/apisix/pull/10589

Issue #10474, opened 2023-11-10
help request: syslog plugin / attribute 'name' confusion


Wasta-Geek:
### Description

Hi

I'm kind of new to Apisix but I wanted to use the syslog plugin and I had trouble to understand the meaning of the plugin key 'name'.
My original need was to modify what would correspond to the 'tag' part (apisix[pid]) in the syslog message and I thought it would correspond to the plugin attribute 'name'.

But this attribute seems to be ignored as I couldn't find it in my remote logs. (other plugin attributes would get applied correctly so it seems not to be a config problem on my side)
I tried to see it via a tcpdump what's been received and I couldn't retrieves the value anywhere, I also tried to grep 'name' in the LUA syslog plugin files and I couldn't see a reference to this.

So I wanted to know what is the usage / impact of this attribute regarding the syslog plugin ?

Thanks for your help, best regards.

### Environment

- APISIX version (run `apisix version`): 3.6.0
- Operating system (run `uname -a`):Linux 5.10.0-25-amd64 #1 SMP Debian 5.10.191-1 (2023-08-16) x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx/1.18.0
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

kayx23:
APISIX loggers [send logs in batches](https://apisix.apache.org/docs/apisix/batch-processor). The name here is the ID of the batch processor for the plugin, which conveniently inherits the plugin's name.

However I noticed the value is hardcoded in, which means what you configure actually doesn't matter. I opened a PR to removed the `name` from attributes for `kafka-logger` here: https://github.com/apache/apisix/pull/10476 

If I'm not wrong on this, we should remove name attribute from the doc for all loggers that hard code the value in.

For sys log specifically, the value is hard coded here:

https://github.com/apache/apisix/blob/e1503dcb2b6220303c6023d7f0fe5e591efecd8f/apisix/plugins/syslog/init.lua#L26

You won't see this value in the actual logs.

Wasta-Geek:
Oh okk, I didn't see that the name in the documentation was similar as in this line and thus meant it would correspond to the batch processing part (even if ignored) :+1: 

Thanks for your help / time, have a nice day !

[...]

PR #10476, opened 2023-11-11
fix: remove redundant name attribute from logging plugins

kayx23:
### Description

Unique ID for batch processor (attribute `name`) are hardcoded in for quite many logging plugins. The docs, schema of certain plugins, and tests, previously treat `name` as a configurable attribute, which could create confusions, such as https://github.com/apache/apisix/issues/10474.

### Checklist


shreemaan-abhishek (approved):

kayx23:
Lmk if I should apply this change for other plugins in the same PR. Quite a few logging plugins have the same issues if I'm not mistaken.

moonming:
> Lmk if I should apply this change for other plugins in the same PR. Quite a few logging plugins have the same issues if I'm not mistaken.

yes, please do it

pottekkat (dismissed):
LGTM and +1 to go ahead and fix this in other plugins as well.

kayx23:
Just saw in `clickhouse-logger` the name is indeed configurable:

https://github.com/apache/apisix/blob/d8bd34414239ed36942c19f435ac10ad0d5e57c0/apisix/plugins/clickhouse-logger.lua#L28-L41

Are we going to:

1. make the other logger plugin names configurable as well; or 
2. Just update the docs for those that are not configurable?

monkeyDluffy6017 (dismissed):

moonming:
> Just saw in `clickhouse-logger` and `sys-logger` the namea are indeed configurable:
> 
> https://github.com/apache/apisix/blob/d8bd34414239ed36942c19f435ac10ad0d5e57c0/apisix/plugins/clickhouse-logger.lua#L28-L41
> 
> image
> 
> Are we going to:
> 
> 1. make the other logger plugin names configurable as well;
> 2. update `clickhouse logger` and `sys logger` to be consistent with others, which hard code the name; or
> 3. Just update the docs for those that are not configurable (done)?

kayx23:
Ok as it turns out (thanks to the failed CI tests) that this `name` attribute is actually configurable and its value gets exported to prometheus to identify batch processors.

The reason why the `name` attribute was documented to each of these logging plugin, despite being a batch processor attribute, was likely because these loggers initiate batch processor instances with specific names (the hard coded values), so they are the default values specific to these plugins.

In short, these docs should not be removed and the documented default values are consistent with what goes in the code.

PR #10482, opened 2023-11-12
feat: add multi-auth plugin

### LINKED ISSUE #10196: Allow multiple authentication methods per route
### Description

Right now, it is not possible to have multiple authentication methods defined for one route. If you want to expose one service to multiple consumers, where some consumers use basic auth, some key auth and some jwt, you need to define three different routes.

A better approach would be to allow different authentication methods for a route, maybe with some prioritization, e.g. if basic auth params present - use basic, if app-key present, use key auth, if jwt-key present, use jwt-auth, ....

madhawa-gunasekara:
### Description

Add multiple authentication for routes. This is the implementation of multi-auth plugin

```json
{
   "plugins":{
      "multi-auth":{
         "auth_plugins":[
            {
               "basic-auth":{ }
            },
            {
               "key-auth":{
                  "query":"apikey",
                  "hide_credentials":true,
                  "header":"apikey"
               }
            },
            {
               "jwt-auth":{
                  "cookie":"jwt",
                  "query":"jwt",
                  "hide_credentials":true,
                  "header":"authorization"
               }
            }
         ]
      }
   }
}
```

Fixes #10196

### Checklist

### DIFF
--- a/apisix/plugins/multi-auth.lua
+++ b/apisix/plugins/multi-auth.lua
@@ -15,16 +15,14 @@
 -- limitations under the License.
 --
 local core = require("apisix.core")
+local require = require
+local pairs = pairs
 
 local schema = {
     type = "object",
     title = "work with route or service object",
     properties = {
-        auth_plugins = { type = "array", minItems = 2 },
-        hide_credentials = {
-            type = "boolean",
-            default = false,
-        }
+        auth_plugins = { type = "array", minItems = 2 }
     },
     required = { "auth_plugins" },
 }
--- a/apisix/plugins/multi-auth.lua
+++ b/apisix/plugins/multi-auth.lua
@@ -69,7 +69,7 @@ function _M.rewrite(conf, ctx)
     for k, auth_plugin in pairs(auth_plugins) do
         for key, value in pairs(auth_plugin) do
             local auth = require("apisix.plugins." .. key)
-            local auth_code, b = auth.rewrite(value, ctx)
+            local auth_code = auth.rewrite(value, ctx)
             status_code = auth_code
             if auth_code == nil then
                 core.log.debug("Authentication is successful" .. key .. " plugin")
--- a/apisix/plugins/multi-auth.lua
+++ b/apisix/plugins/multi-auth.lua
@@ -67,14 +67,14 @@ function _M.rewrite(conf, ctx)
     for k, auth_plugin in pairs(auth_plugins) do
         for auth_plugin_name, auth_plugin_conf in pairs(auth_plugin) do
             local auth = require("apisix.plugins." .. auth_plugin_name)
-            -- returns 401 HTTP status code if authentication failed, otherwise nothing returns.
+            -- returns 401 HTTP status code if authentication failed, otherwise returns nothing.
             local auth_code = auth.rewrite(auth_plugin_conf, ctx)
             status_code = auth_code
             if auth_code == nil then
-                core.log.debug("Authentication is successful" .. auth_plugin_name .. " plugin")
+                core.log.debug(auth_plugin_name .. " succeed to authenticate the request")
                 goto authenticated
             else
-                core.log.warn("Authentication is failed" .. auth_plugin_name .. " plugin, code: "
+                core.log.debug(auth_plugin_name .. " failed to authenticate the request, code: "
                         .. auth_code)
             end
         end
--- a/apisix/plugins/multi-auth.lua
+++ b/apisix/plugins/multi-auth.lua
@@ -1,0 +1,89 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core = require("apisix.core")
+
+local schema = {
+    type = "object",
+    title = "work with route or service object",
+    properties = {
+        auth_plugins = { type = "array", minItems = 2 },
+        hide_credentials = {
+            type = "boolean",
+            default = false,
+        }
+    },
+    required = { "auth_plugins" },
+}
+
+
+local plugin_name = "multi-auth"
+
+local _M = {
+    version = 0.1,
+    priority = 2600,
+    type = 'auth',
+    name = plugin_name,
+    schema = schema
+}
+
+function _M.check_schema(conf)
+    local ok, err = core.schema.check(schema, conf)
+    if not ok then
+        return false, err
+    end
+
+    local auth_plugins = conf.auth_plugins
+    for k, auth_plugin in pairs(auth_plugins) do
+        for key, value in pairs(auth_plugin) do
+            local auth = require("apisix.plugins." .. key)
+            if auth == nil then
+                return false, key .. " plugin did not found"
+                else
+                if auth.type ~= 'auth' then
+                    return false, key .. " plugin is not supported"
+                end
+            end
+        end
+    end
+
+    return true
+end
+
+function _M.rewrite(conf, ctx)
+    local auth_plugins = conf.auth_plugins
+    local status_code
+    for k, auth_plugin in pairs(auth_plugins) do
+        for key, value in pairs(auth_plugin) do
+            local auth = require("apisix.plugins." .. key)
+            local auth_code, b = auth.rewrite(value, ctx)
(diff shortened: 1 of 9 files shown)


t/plugin/multi-auth.t
```diff
@@ -0,0 +1,244 @@
...
+                                },
+                                {
+                                    "key-auth": {
+                                        "query": "apikey",
+                                        "hide_credentials": true,
+                                        "header": "apikey"
+                                    }
+                                },
+                                {
+                                    "jwt-auth": {
```
moonming, line 86: There is no test case to cover jwt?

t/plugin/multi-auth.t
```diff
@@ -0,0 +1,244 @@
...
+GET /hello
+--- more_headers
+apikey: auth-two
+--- error_code: 401
+--- response_body
+{"message":"Authorization Failed"}
+
+
+
+=== TEST 8: enable multi auth plugin using admin api
```
moonming, line 172: Need a meaningful title

t/plugin/multi-auth.t
```diff
@@ -0,0 +1,244 @@
...
+    }
+--- request
+GET /t
+--- error_code: 400
+--- response_body_like eval
+qr/\{"error_msg":"failed to check the configuration of plugin multi-auth err: property \\"auth_plugins\\" is required"\}/
+
+
+
+=== TEST 9: enable multi auth plugin using admin api
```
moonming, line 207: Need a meaningful title

docs/en/latest/plugins/multi-auth.md
```diff
@@ -0,0 +1,140 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Description
+
+The `multi-auth` Plugin is used to add multiple authentication methods to a Route or a Service. Plugins with type 'auth' are supported.
```
moonming, line 33: For multiple different plug-ins, is there an "and" relationship or an "or" relationship between them? In other words, if there are two authentication plug-ins, do both plug-ins need to be verified?

apisix/plugins/multi-auth.lua
```diff
@@ -0,0 +1,87 @@
...
+
+function _M.rewrite(conf, ctx)
+    local auth_plugins = conf.auth_plugins
+    local status_code
+    for k, auth_plugin in pairs(auth_plugins) do
+        for key, value in pairs(auth_plugin) do
+            local auth = require("apisix.plugins." .. key)
+            local auth_code = auth.rewrite(value, ctx)
+            status_code = auth_code
+            if auth_code == nil then
```
moonming, line 72: What does it mean when the return value is `nil`? You can add some comments

apisix/plugins/multi-auth.lua
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    return true
+end
+
+function _M.rewrite(conf, ctx)
+    local auth_plugins = conf.auth_plugins
+    local status_code
+    for k, auth_plugin in pairs(auth_plugins) do
+        for key, value in pairs(auth_plugin) do
```
moonming, line 68: When will there be multiple `key` and `value` under an authentication plug-in? I see that the test case does not cover

t/plugin/multi-auth.t
madhawa-gunasekara: Noted .

apisix/plugins/multi-auth.lua
madhawa-gunasekara: I added comment.

t/plugin/multi-auth.t
madhawa-gunasekara: I added the test cases to cover jwt as well. 
This is kind of a wrapper plugin to enable different authentication methods using "or" relationship.
I just added these 3 plugins to show case the compatibility of any plugin with type 'auth'.

docs/en/latest/plugins/multi-auth.md
madhawa-gunasekara: I added details in here

apisix/plugins/multi-auth.lua
madhawa-gunasekara: I used proper variable names to avoid the confusion :)

madhawa-gunasekara:
Hi @moonming, 

I fixed the changes you mentioned, can you rereview again?

shreemaan-abhishek (changes requested):

apisix/plugins/multi-auth.lua
```diff
@@ -0,0 +1,89 @@
...
+        for auth_plugin_name, auth_plugin_conf in pairs(auth_plugin) do
+            local auth = require("apisix.plugins." .. auth_plugin_name)
+            -- returns 401 HTTP status code if authentication failed, otherwise nothing returns.
+            local auth_code = auth.rewrite(auth_plugin_conf, ctx)
+            status_code = auth_code
+            if auth_code == nil then
+                core.log.debug("Authentication is successful" .. auth_plugin_name .. " plugin")
+                goto authenticated
+            else
+                core.log.warn("Authentication is failed" .. auth_plugin_name .. " plugin, code: "
```
markokocic, line 77: `core.log.debug` instead of warn may be better here. Or just skip the whole `else` block.

docs/en/latest/plugins/multi-auth.md
markokocic: Actually, it will go through the list of configured auth plugins, and exit after the first one is successfully authenticated. That means route can be shared between different consumers using different auth methods. If a consumer is using basic, he would be able to authenticate using basic, if another user uses jwt, he would be able to authenticate using jwt.

Maybe the doc can be clearer about the intent and use case.

apisix/plugins/multi-auth.lua
```diff
@@ -0,0 +1,89 @@
...
+    return true
+end
+
+function _M.rewrite(conf, ctx)
+    local auth_plugins = conf.auth_plugins
+    local status_code
+    for k, auth_plugin in pairs(auth_plugins) do
+        for auth_plugin_name, auth_plugin_conf in pairs(auth_plugin) do
+            local auth = require("apisix.plugins." .. auth_plugin_name)
+            -- returns 401 HTTP status code if authentication failed, otherwise nothing returns.
```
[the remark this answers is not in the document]
madhawa-gunasekara: Thanks for the suggestion

apisix/plugins/multi-auth.lua
madhawa-gunasekara: It should be debug. Thanks for pointing it

docs/en/latest/plugins/multi-auth.md
shreemaan-abhishek: @madhawa-gunasekara did you see this?

shreemaan-abhishek (approved):

docs/en/latest/plugins/multi-auth.md
```diff
@@ -0,0 +1,144 @@
...
+
+For Route:
+
+| Name         | Type  | Required | Default | Description                                                           |
+|--------------|-------|----------|---------|-----------------------------------------------------------------------|
+| auth_plugins | array | True     | -       | Add supporting auth plugins configuration. expects at least 2 plugins |
+
+## Enable Plugin
+
+To enable the Plugin, you have to create a Consumer object with multiple authentication configurations:
```
markokocic, line 45: Better example and a more realistic use case is to have multiple consumers using different authentication plugins using the same route with multi-auth.

docs/en/latest/plugins/multi-auth.md
```diff
@@ -0,0 +1,144 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Description
+
+The `multi-auth` Plugin is used to add multiple authentication methods to a Route or a Service. It supports plugins of type 'auth'. You can combine different authentication methods using "or" relationship with `multi-auth` plugin. If you want to use multiple methods in an "and" relationship, apply specific authentication plugins directly to the route or service.
+
```
markokocic, line 34: As already mentioned, docs should state that authentication plugins will be tried in the order specified in plugin config, and as soon as one one match, it will let that consumer proceed with the service call.

markokocic (approved):
I had only doc changes suggested. Functionality is fine.

moonming (approved):

moonming:
@madhawa-gunasekara good job 👍
@markokocic thanks for your review

docs/en/latest/plugins/multi-auth.md
madhawa-gunasekara: I will add it as an example

docs/en/latest/plugins/multi-auth.md
madhawa-gunasekara: Yes, I will improve the docs

Neilblaze (approved):

SatriyoJati:
it said that it is only support for auth plugin with type 'auth', is authz-keycloak included to this type?

SatriyoJati:
since when i combine key-auth with authz-keycloak, it shown this error 
`{"error_msg":"failed to check the configuration of plugin multi-auth err: authz-keycloak plugin is not supported"}`

PR #10493, opened 2023-11-15
feat: add `required scopes` configuration property to `openid-connect` plugin

### LINKED ISSUE #10352: oidc-connect plugin, and restricting access to endpoints using `scope`
### Description

Hello, I am using the Ingress Controller and I am trying to support the following scenario using OIDC
- ClientID A should have access to Endpoints A,B
- ClientID B should have access to Endpoints B,C

All of them should be of type "**client_credentials**"

And on top of a test route, I have the following definition:

```
apiVersion: apisix.apache.org/v2
kind: ApisixRoute
metadata:
  name: echoserver-multiple-apisix
  namespace: ${kubernetes_namespace_v1.echonamespace2.metadata.0.name}
spec:
  http:
  - name: echohttp
    match:
      hosts:
      - echo.k8s.orb.local
      paths:
      - "/echo*"
    backends:
    - serviceName: echoserver2-service
      servicePort: 80
    plugins:

      - name: "openid-connect"
        enable: true
        config:
          client_id: "apisix2"
          client_secret: "..."
          discovery: "http://keycloak-proxy.keycloak.svc.cluster.local/realms/apisixrealm/.well-known/openid-configuration"
          introspection_endpoint: "http://keycloak-proxy.keycloak.svc.cluster.local/realms/apisixrealm/protocol/openid-connect/token/introspect"
          realm: "apisixrealm"
          scope: "route1scope"
```

I am using the OIDC plugin and Keycloak as my authentication provider. Although APISIX can provide access to the endpoints, with tokens accessed **it fails to deny access to clients that don't have the required scope**. No matter what scope I put in the `scope` parameter, it always allows anyone with a valid token (but without a scope) to access the route. Perhaps this is a relevant issue: https://github.com/apache/apisix/issues/1272

Therefore, I have two questions:
- For the scenario originally described, can I have multiple OIDC plugins with different client IDs and secrets on top of a route to achieve what I want? (since multiple clients must be able to access the same route)
- For the issue where the scopes are not taken into account, is there something I am missing or can do differently?
  - How can I restrict an API to specific scopes with the openid-connect plugin?

csotiriou:
Fixes #10352 

This adds the optional `required_scopes` configuration property in the `openid-connect` plugin configuration. In cases where the introspection endpoint of the OIDC server is called, the plugin will check if all required scopes are present in the scopes returned by the introspection endpoint. 

I have also updated the documentation to reflect this, in case anyone finds this PR valuable.


moonming:
@csotiriou thanks for this PR. It will be great if you can add test cases for it :)

csotiriou:
Hello @moonming I'm trying to wrap my head around the test implementation, my familiarity with lua is limited. However, I have reached the point where I'm making a sample test.

I noticed that for the openid-connect plugin there are some sample oidc providers with a dummy client-id and client-secret used. Must I use the same and have the same configuration in my keycloak? If not, how are tests going to pass during automation?

moonming:
@Revolyssup Please check how to write a more appropriate test case. I am not familiar with this plug-in.

Revolyssup:
@moonming @csotiriou I have added the test case.

Sn0rt (approved):

moonming (approved):

moonming:
@csotiriou thanks for your great work 👍

apisix/plugins/openid-connect.lua
```diff
@@ -315,6 +324,24 @@ local function add_access_token_header(ctx, conf, token)
...
+local function split_scopes_by_space(scope_string)
+    local scopes = {}
+    for scope in string.gmatch(scope_string, "%S+") do
+        scopes[scope] = true
+    end
+    return scopes
+end
+
+-- Function to check if all required scopes are present
+local function required_scopes_present(required_scopes, http_scopes)
```
soulbird, line 337: you can use `core.table.array_find` directly: https://github.com/apache/apisix/blob/master/apisix/core/table.lua#L94

soulbird:
I think you need to add a test case for scope validation

apisix/plugins/openid-connect.lua
Revolyssup: array_find(arr,ele) checks if `ele` is present in `arr`.
This function checks if all the elements in first table are present in second table.

Revolyssup:
> I think you need to add a test case for scope validation

@soulbird Do you mean the case where the correct scope is passed and the request succeeds?

moonming:
> Conflicting files
> docs/en/latest/plugins/openid-connect.md

@csotiriou please fix this when you have time.

soulbird:
> > I think you need to add a test case for scope validation
> 
> @soulbird Do you mean the case where the correct scope is passed and the request succeeds?

yes

soulbird (approved):

csotiriou:
> > Conflicting files
> > docs/en/latest/plugins/openid-connect.md
> 
> @csotiriou please fix this when you have time.

Thanks for the notification, @moonming , It seems that the merge conflicts have been resolved by @Revolyssup faster, thank you for the help!

apisix/plugins/openid-connect.lua
```diff
@@ -319,6 +328,24 @@ local function add_access_token_header(ctx, conf, token)
     end
 end
 
+-- Function to split the scope string into a table
+local function split_scopes_by_space(scope_string)
+    local scopes = {}
+    for scope in string.gmatch(scope_string, "%S+") do
```
monkeyDluffy6017, line 334: Why not use [ngx.re.split](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/re.md) ?

apisix/plugins/openid-connect.lua
```diff
@@ -319,6 +328,24 @@ local function add_access_token_header(ctx, conf, token)
...
+        scopes[scope] = true
+    end
+    return scopes
+end
+
+-- Function to check if all required scopes are present
```
monkeyDluffy6017, line 340: Two blanks are needed between functions

Revolyssup (approved):

monkeyDluffy6017 (approved):

Issue #9645, opened 2023-06-12
feat: will apisix support brotil compression algorithm


dream110fly:
### Description

 brotil has higher compression efficiency than gzip， nginx has support this feature by using module https://github.com/google/ngx_brotli

PR #10515, opened 2023-11-17
feat: add brotli plugin

yuweizzz:
### Description

add brotli plugin.  fixes https://github.com/apache/apisix/issues/9645

### Checklist


moonming:
@yuweizzz great work 👍 
Can you explain the usage scenarios of this plug-in? It would be better if there is documentation

yuweizzz:
sure, I will add a documentation about it.

monkeyDluffy6017:
Nice work! Could you make the ci pass?

yuweizzz:
I don't know why it couldn't work in redhat series, need some time to figure it out.

yuweizzz:
I think the `CI/build` and `CLI Test` should run again both, and `CLI Test` run failed cause server doesn't online but the request has been send already.

.github/workflows/fuzzing-ci.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
         sudo apt-get install -y git curl openresty-openssl111-dev unzip make gcc libldap2-dev
         ./utils/linux-install-luarocks.sh
 
-        make deps
+        sudo make deps
```
monkeyDluffy6017, line 67: Can you call `sudo` in `install_brotli ` ?

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    if type(types) == "table" then
+        local matched = false
+        local from = core.string.find(content_type, ";")
+        if from then
+            content_type = str_sub(content_type, 1, from - 1)
+        end
+
+        for _, ty in ipairs(types) do
```
rickyxucn, line 141: `"*"` cannot match any MIME type.

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,195 @@
...
+
+    ctx.brotli_matched = true
+    core.response.clear_header_as_body_modified()
+    core.response.add_header("Content-Encoding", "br")
+end
+
+
+function _M.body_filter(conf, ctx)
+    if ctx.brotli_matched then
+        local body = core.response.hold_body_chunk(ctx)
```
rickyxucn, line 179: Will calling this function have any performance side effects?

apisix/plugins/brotli.lua
yuweizzz: if we use `"*"` in `"types"`, it will pass this "ipairs", and will set `brotli_matched` as `true` in this func. same as 'gzip' plugin.

apisix/plugins/brotli.lua
yuweizzz: `hold_body_chunk` return the full payload, normally doesn't have performance side effects

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,195 @@
...
+        if ngx.arg[2] == false and not body then
+            return
+        end
+
+        local compressed = brotli_compress(conf, ctx, body)
```
monkeyDluffy6017, line 184: What happens if we don't install the `brotli` library?
Is it better to check if the `brotli` is installed before running the plugin?

conf/config-default.yaml
```diff
@@ -482,6 +482,7 @@ plugins:                           # plugin list (sorted by priority)
   - limit-count                    # priority: 1002
   - limit-req                      # priority: 1001
   #- node-status                   # priority: 1000
+  - brotli                         # priority: 996
```
monkeyDluffy6017, line 485: I suggest we don't enable the `brotli` plugin by default, since the user may not install the `brotli` library
```suggestion
  #- brotli                         # priority: 996
```

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,195 @@
...
+            -- 1: MODE_TEXT (for UTF-8 format text input)
+            -- 2: MODE_FONT (for WOFF 2.0)
+        },
+        comp_level = {
+            type = "integer",
+            minimum = 0,
+            maximum = 11,
+            default = 11,
+        },
+        lgwin = {
```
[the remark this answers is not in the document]
yuweizzz: `lg` mean base 2 logarithm, `lgwin` mean base 2 logarithm the of the sliding window size, in brotli, if `lgwin` is 24, then `sliding window size = 2**24 - 16 bytes`

apisix/plugins/brotli.lua
yuweizzz: if we don't install the brotli library, when `require("brotli")` will raise `no such file or directory` error,  I think check before running is better, but I'm not sure what should do in this part, any advice or example?

apisix/plugins/brotli.lua
yuweizzz: maybe we don't need the full payload here, I will adjust it.

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,195 @@
...
+            default = 0,
+            -- 0: MODE_GENERIC (default),
+            -- 1: MODE_TEXT (for UTF-8 format text input)
+            -- 2: MODE_FONT (for WOFF 2.0)
+        },
+        comp_level = {
+            type = "integer",
+            minimum = 0,
+            maximum = 11,
+            default = 11,
```
rickyxucn, line 65: Why is the default value `11` instead of `6`?
ref: https://github.com/google/ngx_brotli#brotli_comp_level

apisix/plugins/brotli.lua
monkeyDluffy6017: what happens if we only install the lua package `brotli-ffi = 0.3-1` and don't install the c library `brotli` ?
It will raise `no such file or directory` too or other error?

yuweizzz: > what happens if we only install the lua package `brotli-ffi = 0.3-1` and don't install the c library `brotli` ? It will raise `no such file or directory` too or other error?

will raise same error.

apisix/plugins/brotli.lua
yuweizzz: I follow the default value from brotli cmdline, I will adjust it.

apisix/plugins/brotli.lua
monkeyDluffy6017: Maybe we could do this at the top of the file, like this:
```
local brotli = pcall(require, "brotli")
function _M.header_filter(conf, ctx)
  if not brotli then
      core.log.error()
  end
end

function _M.body_filter(conf, ctx)
if not brotli then
    core.log.error()
end
```

apisix/plugins/brotli.lua
yuweizzz: > maybe we don't need the full payload here, I will adjust it.

Get it, now I'm trying to not use full payload here.

yuweizzz: could you help me to check it again? thank you very much.

rickyxucn: I have a few questions:
Why is the variable `compressor` not declared `local`?
What happens if compressor initialization fails?

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,200 @@
...
+    core.response.clear_header_as_body_modified()
+    core.response.add_header("Content-Encoding", "br")
+end
+
+
+function _M.body_filter(conf, ctx)
+    if not is_loaded then
+        core.log.error("please check the brotli library")
+        return
+    end
```
monkeyDluffy6017, line 182: I think this is unnecessary, because the `ctx.brotli_matched` is nil

docs/en/latest/plugins/brotli.md
```diff
@@ -0,0 +1,121 @@
...
+    }
+}'
+```
+
+## Example usage
+
+Once you have configured the Plugin as shown above, you can make a request as shown below:
+
+```shell
+curl http://127.0.0.1:9080/index.html -i -H "Accept-Encoding: br"
```
monkeyDluffy6017, line 89: Please use httpbin.org as example, users can not acccess 127.0.0.1:9080

docs/en/latest/plugins/brotli.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+```
+HTTP/1.1 404 Not Found
```
monkeyDluffy6017, line 93: Why is 404?

docs/en/latest/plugins/brotli.md
```diff
@@ -0,0 +1,121 @@
...
+{
+    "uri": "/index.html",
+    "plugins": {
+        "brotli": {
+        }
+    },
+    "upstream": {
+        "type": "roundrobin",
+        "nodes": {
+            "127.0.0.1:1980": 1
```
monkeyDluffy6017, line 78: Please don't use 127.0.0.1:1980, users cannot access the url, use httpbin.org instead

docs/en/latest/plugins/brotli.md
yuweizzz: "httpbin.org" doesn't has "/index.html" endpoint, I will change it.

monkeyDluffy6017 (dismissed):

apisix/plugins/brotli.lua
rickyxucn: The default value in the help documentation is still `11`?

rickyxucn (approved):

monkeyDluffy6017:
please make the ci pass

.github/workflows/fuzzing-ci.yaml
```diff
@@ -56,15 +56,6 @@ jobs:
...
         export_or_prefix
-        wget -qO - https://openresty.org/package/pubkey.gpg | sudo apt-key add -
-        sudo apt-get update
-        sudo apt-get -y install software-properties-common
-        sudo add-apt-repository -y "deb http://openresty.org/package/ubuntu $(lsb_release -sc) main"
-        sudo apt-get update
-        sudo apt-get install -y git curl openresty-openssl111-dev unzip make gcc libldap2-dev
-        ./utils/linux-install-luarocks.sh
-
-        make deps
```
monkeyDluffy6017, line 67: Why do you remove this?

yuweizzz: They are finished in `do_install` and `before_install`.

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,241 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core = require("apisix.core")
+local ngx = ngx
+local ngx_re_gmatch = ngx.re.gmatch
+local ngx_header = ngx.header
```
shreemaan-abhishek, line 20: consider using `core.req.header`

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,241 @@
...
+local req_http_version = ngx.req.http_version
```
shreemaan-abhishek, line 21: ditto

apisix/plugins/brotli.lua
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    type = "array",
+                    minItems = 1,
+                    items = {
+                        type = "string",
+                        minLength = 1,
+                    },
+                },
+                {
+                    enum = {"*"}
```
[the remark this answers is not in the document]
yuweizzz: `"*"` use to match any MIME type.

apisix/plugins/brotli.lua
yuweizzz: In this part, we need check the header from upstream response, not the request header.

apisix/plugins/brotli.lua
yuweizzz: request's http version not directly obtained from a request header.

shreemaan-abhishek (approved):

PR #10543, opened 2023-11-23
fix: use proxy_protocol_addr instead of remote_addr in proxy-protocol case

heresie:
Hello to APISIX maintainers ✋🏻 

### Description

We are currently using APISIX API Gateway and Ingress Controller on an Kubernetes infrastructure and we had some hard time to configure properly the way `X-Real-IP` is set in `proxy-protocol`.

By default, the APISIX Ingress Controller will use `$remote_addr` as value for `X-Real-IP` header.
But IMHO, the value should be `$proxy_protocol_addr` in proxy-protocol context.

Currently, in order to get the visitor IP address, we have to : 
 - Enable `proxy-rewrite` plugin on instance level with this configuration : 
   ```
   {
     "headers": {
       "set": {
         "X-Forwarded-For": "$proxy_protocol_addr",
         "X-Forwarded-Port": "$proxy_protocol_port"
       }
     },
   }
   ```
 - Enable `real-ip` plugin on each `APISIXRoute` we declare : 
   ```
        - name: real-ip
          enable: true
          config:
            source: http_x_forwarded_for
   ```

### Checklist

  => not covered by current tests
  => not covered by current documentation

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -715,7 +715,7 @@ http {
             proxy_set_header   Host              $upstream_host;
             proxy_set_header   Upgrade           $upstream_upgrade;
             proxy_set_header   Connection        $upstream_connection;
-            proxy_set_header   X-Real-IP         $proxy_protocol_addr;
+            proxy_set_header   X-Real-IP         {% if proxy_protocol then %} $proxy_protocol_addr {% else %} $remote_addr {% end %};
             proxy_pass_header  Date;
 
             ### the following x-forwarded-* headers is to send to upstream server
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -715,7 +715,7 @@ http {
             proxy_set_header   Host              $upstream_host;
             proxy_set_header   Upgrade           $upstream_upgrade;
             proxy_set_header   Connection        $upstream_connection;
-            proxy_set_header   X-Real-IP         $remote_addr;
+            proxy_set_header   X-Real-IP         $proxy_protocol_addr;
             proxy_pass_header  Date;
 
             ### the following x-forwarded-* headers is to send to upstream server


bzp2010:
Hi, @heresie. Thank you for your contribute!

bzp2010:
Hi, @heresie. Thank you for your contribution!

There may be some misunderstanding here.

As `$remote_addr` is a client-side source variable directly managed by the lower-level Nginx C module, we should use it.

And you mentioned the `real-ip` plugin; indeed, we should use it too.

Let us ignore some configurations used to filter out illegal request sources.

What the `real-ip` plugin does is that its core function is to get the value from an Nginx variable and call the Nginx C module with its internal logic to set it to the request source.

Yes, you see that I mentioned "get the value from an Nginx variable", which could also be `$proxy_protocol_addr`.

This gives us the "only correct" path to rewrite the request source, and other modules and plugins on the system can use `$remote_addr` to get the source address, which will work fine. There is no need to dispute the use of `$remote_addr` and `$proxy_protocol_addr` in every place it is used.

So, I think the behavior in the code is correct.

shreemaan-abhishek:
> So, I think the behavior in the code is correct. We should not create duality, not everyone is using the proxy protocol and it would be inappropriate to modify it directly.

I agree with zeping. Additionally @heresie, you can set the real-ip plugin as a global rule: https://docs.api7.ai/apisix/key-concepts/plugin-global-rules

heresie:
Hello @bzp2010 and @shreemaan-abhishek 

Thank you for your time and answers.

I understand that not everyone uses `proxy_protocol` with APISIX and my first contribution have not considered this. I updated it with `{% if %}` condition in order to address both cases, as the `ngx_tpl.lua` already uses similar conditions elsewhere.

I know that my problem can be adressed by enabling plugins on instance or route levels, but I think that it could be better to handle those both situations (with/without PP) directly by the lua configuration, without the need of enabling APISIX plugins.

Do you think it could be accepted?

Thank in advance for your help.

bzp2010:
Hi, @heresie.

No offense intended, but I think real-ip exists precisely to solve related problems, e.g. let the other plugins get correct IP of client and send x-real-ip to upstream.

If you think that turning on real-ip will cause some trouble for your use case, you can directly make the c module work normally at the nginx.conf level by modifying the realip-related items (they will be mapped to the directives provided by nginx realip module in nginx.conf) in the configuration file. This way is fine, but you will lose the dynamic benefit of using the lua plugin.

Back to your modification, in fact just adjusting the value of x-real-ip is not enough. You ignored some other issues. When we do not configure real-ip correctly (whether it is nginx directive or apisix plug-in), other parts of the system will not be able to get the correct source IP, such as the ip-restriction , it will not work as expected. This also That's why I recommend you use real-ip instead of modifying this. This is a once and for all solution.

bzp2010:
@heresie Excellent discussion, but perhaps there are some misunderstandings here.

We discuss this in two dimensions: Nginx and APISIX.

------

For Nginx, when you configure the proxy protocol `listen 80 proxy_protocol;`, this means that when Nginx receives the correct inbound request for the Proxy Protocol, it will extract the source IP in the protocol header and store it in `$proxy_protocol_addr`, right?

What's next? At this point, `$remote_addr` still stores the IP of the previous load balancer.

For the Nginx-only use case, you will need to rewrite the variables using the `ngx_http_realip_module`, referenced here in the Nginx documentation ([link](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/)):

> With the [RealIP](https://nginx.org/en/docs/http/ngx_http_realip_module.html) module which rewrites the values in the [$remote_addr](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_remote_addr) and [$remote_port](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_remote_port) variables, replacing the IP address and port of the load balancer with the original client IP address and port. The [$realip_remote_addr](https://nginx.org/en/docs/http/ngx_http_realip_module.html#var_realip_remote_addr) and [$realip_remote_port](https://nginx.org/en/docs/http/ngx_http_realip_module.html#var_realip_remote_port) variables retain the address and port of the load balancer, and the [$proxy_protocol_addr](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_proxy_protocol_addr) and [$proxy_protocol_port](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_proxy_protocol_port) variables retain the original client IP address and port anyway.

Now, you can use `$remote_addr` in any part of Nginx, as it is already the correct source IP. Either enforcing the source IP restriction (`deny 1.2.3.4;`) or setting it to the upstream header (`proxy_set_header X-Real-IP $remote_addr;`) works correctly.

This is the source IP that I mentioned that any other module in the system could get in a unique way without having to change every `$remote_addr` to `$proxy_protocol_addr`. Think about it: if you have 100 server blocks and you need to modify each one, that's obviously inappropriate, and using `ngx_http_realip_module` does this seamlessly and simply, right?

------

Let's look at APISIX, what's the difference.

According to the documentation ([link](https://nginx.org/en/docs/http/ngx_http_realip_module.html)), `ngx_http_realip_module`, only supports HTTP headers and proxy protocol as the client IP source.

However, the APISIX supports any Nginx variable and also provides dynamic on-and-off functionality.

You mention a complex process to manipulate IP, but may have missed the point that the `real-ip` plugin can be enabled on global rules just like you can with `proxy-rewrite`.

Any plugin that wishes to get the source IP gets it from `$remote_addr`, which is absolutely correct.

For example, take a look at the `ip-restriction` plugin ([code line](https://github.com/apache/apisix/blob/9d34d28fa7c1ab70e1acb61196fb91ad46c167cb/apisix/plugins/ip-restriction/init.lua#L92
)), which gets its value from `$remote_addr` in variables without even considering `$proxy_protocol_addr`, right? If your changes were implemented, we would have to decide which source to use in every place where IPs are used, and all I can say is that it would be a disaster.

Your modification is a very small part of the gateway, that directive is only responsible for passing the source IP upstream, in no other code does it read that value as the source address. It doesn't do anything to help `ip-restriction`, etc., get the correct address (which is the part handled by the complex rewrite operation you're using).

So, to summarize, turn on the `real-ip` plugin globally (global rules) and configure its address source as the variable `$proxy_protocol_addr`, nothing else needs to be done. No code changes or rewrites are required, and all functions will work correctly.

heresie:
Hello @bzp2010,

Thank for your detailed feedback and explanations.

I guess we will configure APISIX as you described.

Issue #10431, opened 2023-11-01
bug: the worker run into a dead end in events.lua


nanamikon:
### Current Behavior

![image](https://github.com/apache/apisix/assets/2010632/c874911e-1d96-4265-bfec-8610cade13b5)
One of worker run into a dead end

### Expected Behavior

Not run into a dead end

### Error Logs

Lua Thread stack as follow
```
(gdb) lbt
C:ngx_http_lua_ngx_sleep
builtin#21
@/scripts//deps/share/lua/5.1/resty/worker/events.lua:353
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:1156
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:621
builtin#21
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:505
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:527
@/scripts/apisix/balancer.lua:235
@/scripts/apisix/balancer.lua:363
@/scripts/apisix/init.lua:830
=balancer_by_lua:2
(gdb) c
Continuing.
^C
Program received signal SIGINT, Interrupt.
0x00007f08bd849000 in _Unwind_Find_FDE () from /lib/x86_64-linux-gnu/libgcc_s.so.1
(gdb) lbt
C:ngx_http_lua_ngx_sleep
builtin#21
@/scripts//deps/share/lua/5.1/resty/worker/events.lua:353
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:1156
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:621
builtin#21
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:505
@/scripts//deps/share/lua/5.1/resty/healthcheck.lua:527
@/scripts/apisix/balancer.lua:235
@/scripts/apisix/balancer.lua:363
@/scripts/apisix/init.lua:830

```

Thread stack as follow

```
(gdb) c
Continuing.

Breakpoint 1, lj_str_new (L=0x405b6380, 
    str=0x47b72010 "API disabled in the context of balancer_by_lua*-gz/upstreams/740001-cbca25b9-2f64-46b0-80f5-62f5cb408ea1:state:10.112.230.60:8080808080r4LYLNew!10.124.117.66direction=OUTdirection=IN475&clusters=&dire"..., lenx=47) at lj_str.c:351
351	in lj_str.c
(gdb) bt
[...]

PR #10550, opened 2023-11-27
feat: add new lua-resty-events as events implementation

bzp2010:
### Description

Provide lua-resty-events as a new implementation of the event system. It experimentally provides better performance and efficiency.
fix https://github.com/apache/apisix/issues/10431

### Checklist


ci/centos7-ci.sh
```diff
@@ -41,11 +41,15 @@ install_dependencies() {
...
     yum install -y yum-utils && yum-config-manager --add-repo https://openresty.org/package/centos/openresty.repo
     wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime-debug-centos7.sh"
-    wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime.sh"
+    #wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime.sh"
+    wget https://raw.githubusercontent.com/Sn0rt/apisix-build-tools/guohao/update-lua-resty-events/build-apisix-runtime.sh
     chmod +x build-apisix-runtime-debug-centos7.sh
     chmod +x build-apisix-runtime.sh
     ./build-apisix-runtime-debug-centos7.sh
 
+    # patch lua-resty-events
```
Sn0rt, line 50: why update this ?

ci/centos7-ci.sh
```diff
@@ -41,11 +41,15 @@ install_dependencies() {
     # install openresty to make apisix's rpm test work
     yum install -y yum-utils && yum-config-manager --add-repo https://openresty.org/package/centos/openresty.repo
     wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime-debug-centos7.sh"
-    wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime.sh"
+    #wget "https://raw.githubusercontent.com/api7/apisix-build-tools/apisix-runtime/${APISIX_RUNTIME}/build-apisix-runtime.sh"
+    wget https://raw.githubusercontent.com/Sn0rt/apisix-build-tools/guohao/update-lua-resty-events/build-apisix-runtime.sh
```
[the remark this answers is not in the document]
bzp2010: @Sn0rt Is the build for `apisix-runtime` already available and includes this library?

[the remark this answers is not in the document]
bzp2010: done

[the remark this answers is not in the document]
bzp2010: @Sn0rt I noticed that that project has not yet created a tag for the build that includes the events module, where should I use the build script.

ci/centos7-ci.sh
Sn0rt: ok.

ci/redhat-ci.sh
```diff
@@ -94,7 +100,7 @@ run_case() {
     make init
     set_coredns
     # run test cases
-    FLUSH_ETCD=1 prove --timer -Itest-nginx/lib -I./ -r ${TEST_FILE_SUB_DIR} | tee /tmp/test.result
+    FLUSH_ETCD=1 TEST_EVENTS_MODULE=lua-resty-worker-events prove --timer -Itest-nginx/lib -I./ -r ${TEST_FILE_SUB_DIR} | tee /tmp/test.result
```
Sn0rt, line 103: I don’t quite understand this part. I see that the code already uses lua-resty-event as the default event library. Why do I need to specify this variable?

t/node/healthcheck-discovery.t
```diff
@@ -153,8 +156,6 @@ qr/(create new checker|try to release checker): table/
 create new checker: table
 try to release checker: table
 create new checker: table
---- no_error_log
```
Sn0rt, line 156: Why should it be deleted?

ci/redhat-ci.sh
bzp2010: First, this is the same logic executed in the other test scripts, with the difference that under ubuntu we are executing the tests for both event modules at the same time, where an environment variable controlled by GitHub workflow is used as input.

You can see the header of some of the test cases, some of the tests are related to the events library and because of the implementation of the library they have slightly different behaviors, such as proactive health checks, which use different test cases under different events modules.

If an events module is not explicitly configured, it will default to the value in config-default, which is lua-resty-events, so simultaneous testing is not possible.

But it looks like some of the logic can be ported to APISIX.pm, and I'll be making that change.

t/node/healthcheck-discovery.t
bzp2010: This is the cost of the lua-resty-events library, whose events are asynchronous and delayed, so when the health checker is first created, the events are not communicated in a timely manner, so these error logs must appear.

bzp2010: They should have no impact because the main use of the test is to test that when the service discovers a change in the node provided, the new checker is created correctly rather than as a result of a health check.

If you are concerned about this, we will have to follow the example of the passive health check tests by listing them as separate test files and skipping them when they are not needed.

t/APISIX.pm
```diff
@@ -413,6 +413,11 @@ _EOC_
 
         require "resty.core"
 
+        local events_sock_path = "$apisix_home/t/servroot/logs/stream_worker_events.sock"
+        if os.getenv("TEST_NGINX_USE_HUP") ~= "1" then
```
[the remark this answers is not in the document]
bzp2010: Theoretically they will process properly, I can try. Maybe it's not necessary.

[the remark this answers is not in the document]
bzp2010: Removed, it can work.

monkeyDluffy6017 (approved):

Sn0rt (approved):

PR #10558, opened 2023-11-28
feat: create new event module for decouple logic and specific library

bzp2010:
### Description

Since we wanted to introduce a new implementation (#10550) of the event library, and the current code strongly relies on `lua-resty-worker-events`.

Therefore we need a new abstraction layer to accommodate the library API differences and to unify the usage logic within the APISIX gateway. So that the code no longer connects directly to a specific library.

This PR is supposed to not break any existing tests, as there are no substantive changes at this time.

### Checklist


soulbird:
> Should the `apisix.event` module use the plural form `apisix.events`? @monkeyDluffy6017 @soulbird

`apisix.events` will be better

monkeyDluffy6017:
@bzp2010 i prefer to `apisix.events`

apisix/event.lua
```diff
@@ -0,0 +1,70 @@
...
+    local ok, err = we.configure({shm = shm, interval = 0.1})
+    if not ok then
+        error("failed to init worker event: " .. err)
+    end
+
+    return we
+end
+
+
+function _M.init_worker()
```
monkeyDluffy6017, line 41: should we pass the option?

bzp2010: @monkeyDluffy6017 Do you mean some options for the events module, such as the configuration from yaml?

monkeyDluffy6017: Yeah,  it's better if we can do that in the `config.yaml`

bzp2010: It looks like we don't need any configuration for existing use cases. Perhaps we'll need to configure that later when we provide a switchable events module implementation.

bzp2010: @monkeyDluffy6017 To summarize, we may introduce some options in future PR, if you have concerns about the continuity of implementation, we can create a separate new branch for staging each PR and commit the integrated code to the mainline from the new branch.

monkeyDluffy6017 (approved):

AlinsRan (approved):

Issue #10494, opened 2023-11-15
chore: metadata log format is irrelevant


shreemaan-abhishek:
### Description

Taking an example of Kafka Logger, the plugin metadata log_format is sourced from log_util.lua:

https://github.com/apache/apisix/blob/4aeda49d853e274289792008b4c14928ebb46a5d/apisix/plugins/kafka-logger.lua#L133C1-L139C2

https://github.com/apache/apisix/blob/4aeda49d853e274289792008b4c14928ebb46a5d/apisix/utils/log-util.lua#L33C1-L43C1

But a totally different log_format goes unused and a different log format is in use by default:

https://github.com/apache/apisix/blob/4aeda49d853e274289792008b4c14928ebb46a5d/apisix/utils/log-util.lua#L132C1-L161C1

This is not inline with the documentation:

<img width="816" alt="image" src="https://github.com/apache/apisix/assets/61597896/30226f0f-8f3c-4ede-854f-23e7fef0fff8">

https://apisix.apache.org/docs/apisix/plugins/kafka-logger/#metadata

### How this can be fixed:

- if the metadata log format is not in use (can be done by removing it and running the CI) then get rid of it.
- update the documentation to show the correct log format.
- check if other logger plugins have the same issue.

kayx23:
You mean the default log_format of plugin metadata is not used right?

Because plugin metadata can still be used, providing that `meta_format` is set to `default`.

luoluoyuyu:
hi @shreemaan-abhishek , I have a question in the code seems to first determine whether to use a custom meta_format, no custom meta_format will use the default log_format, my understanding is correct? https://github.com/apache/apisix/blob/master/apisix/utils/log-util.lua#L240-L253

shreemaan-abhishek:
@kayx23 

> Because plugin metadata can still be used, providing that meta_format is set to default.

That is right according to the code but when I used `kafka-logger` without modifying the `log_format` the [default log format](https://github.com/shreemaan-abhishek/apisix/blob/105993f40f34768cd0a86035d01393346fff3a94/apisix/utils/log-util.lua#L37-L41) wasn't used instead the [full log format](https://github.com/shreemaan-abhishek/apisix/blob/105993f40f34768cd0a86035d01393346fff3a94/apisix/utils/log-util.lua#L133-L160) was used.

@Vacant2333 you might want to confirm this ⬆️. You can refer this documentation for kafka-logger: https://docs.api7.ai/hub/kafka-logger

[...]

PR #10574, opened 2023-11-30
feat: add http version to log

jbergstroem:
### Description

Expose http version to the logger which is useful for monitoring `HTTP 1.1` -> `HTTP 2.0` transition across the web.

Question: how should I update the documentation? Should I for instance go through all log plugin examples and add it in respective places?

### Checklist

### DIFF
--- a/apisix/utils/log-util.lua
+++ b/apisix/utils/log-util.lua
@@ -137,7 +137,9 @@ local function get_full_log(ngx, conf)
             method = ngx.req.get_method(),
             headers = ngx.req.get_headers(),
             querystring = ngx.req.get_uri_args(),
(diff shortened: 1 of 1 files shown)


FelipeLema:
hey @shreemaan-abhishek 

does the default value stated in docs (say [here](https://apisix.apache.org/docs/apisix/plugins/kafka-logger/) ) match the format in this modified function?

it's not easy to see whether we can get the same key-value entries if we set `log_format` , so that's why I'm asking

shreemaan-abhishek:
@FelipeLema, yes although right now there are some discrepancies in the default log format (tracked by #10494) using the admin api to modify the log format will work.

PR #10607, opened 2023-12-06
feat: allow to use environment variables for limit-count plugin settings

### LINKED ISSUE #10597: feat: As a user i want to configure sensitive limit-count plugin properties using environment variables
### Description

**Description**

[limit-count](https://github.com/apache/apisix/blob/master/apisix/plugins/limit-count/init.lua) plugin has several sensitive properties (contain passwords and username) or fields which should be the same for all instances of plugins in side apisix server, for example:

- redis_username
- redis_password
- redis_host
- redis_port
- redis_cluster_nodes
- redis_cluster_name

**Use case**

We have a single Redis cluster which should be used for all limit-count plugins we do not want to duplicate Redis-related settings and we do not want to include _redis_password_ directly in configuration scripts which create plugins inside routes.

**Proposal**

Make all string and string array properties support environment variables. 

If the property is an environment variable it should be prefixed with _env:_ or _$env:_. In other cases, properties will be treated as a plain value.

ikatlinsky:
### Description

As per https://github.com/apache/apisix/issues/10597.

Fixes #10597

### Checklist


moonming:
@ikatlinsky thank you for your contribution 👍

apisix/plugins/limit-count.lua
```diff
@@ -14,11 +14,12 @@
...
 --
-local limit_count = require("apisix.plugins.limit-count.init")
+local fetch_secrets = require("apisix.secret").fetch_secrets
+local limit_count   = require("apisix.plugins.limit-count.init")
 
-local plugin_name = "limit-count"
-local _M = {
-    version = 0.4,
+local plugin_name   = "limit-count"
+local _M            = {
```
monkeyDluffy6017, line 21: Why do you add the indent? it's better to keep the original style

apisix/plugins/limit-count/init.lua
```diff
@@ -68,6 +68,7 @@ local policy_to_additional_properties = {
                 type = "boolean", default = false,
             },
         },
+        encrypt_fields = {"redis_username", "redis_password"},
```
monkeyDluffy6017, line 71: Does the `redis_username` need  to be encrypted?

t/plugin/limit-count-redis-cluster3.t
```diff
@@ -77,3 +81,55 @@ __DATA__
...
+                        "limit-count": {
+                            "count": 2,
+                            "time_window": 60,
+                            "rejected_code": 503,
+                            "key": "remote_addr",
+                            "policy": "redis-cluster",
+                            "redis_timeout": 1001,
+                            "redis_cluster_nodes": [
+                                "$ENV://REDIS_NODE_0",
+                                "$ENV://REDIS_NODE_1"
```
monkeyDluffy6017, line 106: I think you are going to encrypt the `redis_password`, am i wrong?

t/plugin/limit-count5.t
```diff
@@ -89,3 +91,49 @@ remaining: 2
...
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "methods": ["GET"],
+                        "plugins": {
+                            "limit-count": {
+                                "count": 2,
+                                "time_window": 60,
+                                "rejected_code": 503,
+                                "key": "$ENV://LIMIT_COUNT_KEY"
```
monkeyDluffy6017, line 111: ditto

apisix/plugins/limit-count.lua
ikatlinsky: my bad, seems like my auto-formatter configuration

t/plugin/limit-count-redis-cluster3.t
ikatlinsky: it is not about encryption mostly, but about using environment variables or secrets to configure the plugin. Based on my changes each attribute of the plugin which is:
- string
- array of strings

can be configured via env variable or secret. 

So in this test i am checking if environment variables work on the example of array of strings field - redis_cluster_count.

t/plugin/limit-count5.t
ikatlinsky: same reasoning as above

t/plugin/limit-count-redis-cluster3.t
monkeyDluffy6017: I misunderstood you, so you did two different thins in the pr: like `redis_password`

apisix/plugins/limit-count/init.lua
```diff
@@ -68,6 +68,7 @@ local policy_to_additional_properties = {
                 type = "boolean", default = false,
             },
         },
+        encrypt_fields = {"redis_password"},
```
monkeyDluffy6017, line 71: I think this is not correct, please refer to https://github.com/apache/apisix/blob/master/apisix/plugins/kafka-proxy.lua#L36, and add a corresponding test case

ikatlinsky: your are correct, that i am doing 2 different things in a single PR, let me ever this change

ikatlinsky:
@monkeyDluffy6017 looking at failing tests (https://github.com/apache/apisix/actions/runs/7124716224/job/19435705333?pr=10607) and it seems like tests for multiple plugins are failed, like request id, openwisk and etc. It that something that is expected? Not sure if my changes can break other plugins.

Thanks

monkeyDluffy6017:
@ikatlinsky It has nothing to do with you changes

monkeyDluffy6017 (approved):

monkeyDluffy6017 (changes requested):
please resolve the conflicts

ikatlinsky:
> please resolve the conflicts

done

monkeyDluffy6017 (approved):

Revolyssup (approved):

PR #10612, opened 2023-12-07
feat: set minLength of redis_cluster_nodes to 1 for limit-count plugin

### LINKED ISSUE #10596: feat: as a user I want to configure limit-count plugin with single value in redis_cluster_nodes property
### Description

**Description**

Current implementation (https://github.com/apache/apisix/blob/master/apisix/plugins/limit-count/init.lua#L79) of schema in plugin _limit-count_ requires _redis_cluster_nodes_ to have a minimum length of 2, which seems too restrictive.

**Use case**

One can have the following example of setup:

1. Redis Cluster deployed in Kubernetes with 3 nodes (for example, 10.42.0.113:6379;10.42.0.114:6379;10.42.0.115:6379)
2. Redis Cluster nodes are accessible through [Kubernetes Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) (for example redis-headless-service

Kubernetes headless service is a Kubernetes service that does not assign an IP address to itself. Instead, it returns the IP addresses of the pods associated with it directly to the DNS system, allowing clients to connect to individual pods directly.

So, each time we request redis-headless service we get all 3 nodes ips back (not a single IP of a single node), and it is the task of Redis client to route to one of the nodes properly.

As a result, the following configuration should be valid:

```curl -i http://127.0.0.1:9180/apisix/admin/routes/1 \
{
    "uri": "/index.html",
    "plugins": {
        "limit-count": {
            "redis_cluster_nodes": [
              "redis-headless-service:6379"
            ],
        }
    },
    "upstream": {}
}'
```

and equivalent to

```
curl -i http://127.0.0.1:9180/apisix/admin/routes/1 \
{
    "uri": "/index.html",
    "plugins": {
        "limit-count": {
            "redis_cluster_nodes": [
              "10.42.0.113:6379",
              "10.42.0.114:6379",
              "10.42.0.115:6379"
            ],
        }
    },
    "upstream": {}
}'
```

**Proposal**

Update plugin schema validation to make _minLength=1_ for _limit-count_ plugin.

ikatlinsky:
### Description

As per https://github.com/apache/apisix/issues/10596

Fixes #10596

### Checklist


t/plugin/limit-count-redis-cluster.t
```diff
@@ -95,8 +95,7 @@ __DATA__
                             "policy": "redis-cluster",
                             "redis_timeout": 1001,
                             "redis_cluster_nodes": [
-                                "127.0.0.1:5000",
-                                "127.0.0.1:5001"
+                                "127.0.0.1:5000"
```
moonming, line 98: please add a new test case instead of change old one.
Better cover example in https://github.com/apache/apisix/issues/10596#issuecomment-1844685990, like
```
"limit-count": {
            "redis_cluster_nodes": [
              "redis-headless-service:6379"
            ],
        }
```

monkeyDluffy6017 (approved):

moonming (approved):

PR #10619, opened 2023-12-08
fix: do not invoke add_header if value resolved as nil

fengxsong:
### Description

If header_value is resolved to nil or empty table, `add_header` will throw an error https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/request.lua#L364

It's better to do nothing.

ref: https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/req.md#add_header

### Checklist

### DIFF
--- a/apisix/plugins/proxy-rewrite.lua
+++ b/apisix/plugins/proxy-rewrite.lua
@@ -365,8 +365,11 @@ function _M.rewrite(conf, ctx)
             local val = core.utils.resolve_var_with_captures(hdr_op.add[i + 1],
                                             ctx.proxy_rewrite_regex_uri_captures)
             val = core.utils.resolve_var(val, ctx.var)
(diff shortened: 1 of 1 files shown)


Revolyssup:
@fengxsong Can you add some test here?

fengxsong:
> @fengxsong Can you add some test here?

do we really need test case for this PR?

Revolyssup:
> > @fengxsong Can you add some test here?
> 
> do we really need test case for this PR?

A test case proving that logs have no error log in this condition might be helpful. @Sn0rt Do you have suggestions here?

monkeyDluffy6017 (approved):

monkeyDluffy6017:
@Revolyssup I think the code is very simple, no test cases is ok

AlinsRan (approved):

PR #10630, opened 2023-12-11
feat: validate `apisix.yaml` for config_provider yaml

### LINKED ISSUE #10583: feat: validate `apisix.yaml` before deployment for standalone mode
### Description

As a user I want to validate my apisix.yaml before deploying changes.

Currently calling `apisix test` only validates the `config.yaml` but not the `apisix.yaml` when the `deployment.role_data_plane.config_provider` is set to `yaml`.

Would it be possible to extend `apisix test` or create another command to validate the `apisix.yaml` using the cli?

Thanks in advance for having a look at this.

Related to https://the-asf.slack.com/archives/CUC5MN17A/p1695458645370659

boekkooi-lengoo:
### Description

This prototype tries to show a way to validate the `apisix.yaml` by starting openresty and only running the validation code.

As part of the validation we initialize the configuration and then check if no errors where logs. If any errors are logged then the `apisix test` command fails.

Special thanks to https://github.com/openresty/resty-cli/!

Fixes #10583

### Checklist

### DIFF
--- a/apisix/cli/ops.lua
+++ b/apisix/cli/ops.lua
@@ -153,7 +153,7 @@ local function get_lua_path(conf)
 end
 
 
-local function init(env)
+local function init_sysconf(env)
     if env.is_root_path then
         print('Warning! Running apisix under /root is only suitable for '
               .. 'development environments and it is dangerous to do so. '
@@ -742,6 +742,13 @@ Please modify "admin_key" in conf/config.yaml .
     sys_conf["extra_lua_path"] = get_lua_path(yaml_conf.apisix.extra_lua_path)
     sys_conf["extra_lua_cpath"] = get_lua_path(yaml_conf.apisix.extra_lua_cpath)
 
+    return sys_conf
+end
+
+
+local function init(env)
+    local sys_conf = init_sysconf(env)
+
     local conf_render = template.compile(ngx_tpl)
     local ngxconf = conf_render(sys_conf)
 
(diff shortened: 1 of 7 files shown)


monkeyDluffy6017:
Nice work! We really appreciate this pr!

apisix/cli/ops.lua
```diff
@@ -862,6 +869,106 @@ local function start(env, ...)
...
+        -- If an error is logged then mark the module as invalid
+        apisix_core.log.no_error_logs = true
+        apisix_core.log.error = function(...)
+             apisix_core.log.orig_error(...)
+             apisix_core.log.no_error_logs = false
+        end
+
+        local valid = true
+        for module_name, cfg_func in pairs(modules) do
+            apisix_core.log.no_error_logs = true
```
shreemaan-abhishek, line 927: i assume this was by mistake

apisix/cli/ops.lua
```diff
@@ -153,7 +153,7 @@ local function get_lua_path(conf)
 end
 
 
-local function init(env)
+local function init_sysconf(env)
```
shreemaan-abhishek, line 156: can you explain what have you done here?

boekkooi-lengoo: I extracted this method from `local function init(env)` so it can also be used by `local function test_standalone(env)`.

As the information from now in `init_sysconf` is used to [render `apisix.cli.ngx_cli_tpl`](https://github.com/apache/apisix/pull/10630/files#diff-228a51571b53709ad043d1c6c14fc8e508c248a84f8014fe548b5d259cedd5ddR949).

apisix/cli/ops.lua
boekkooi-lengoo: This was actually done on purpose. 
As `apisix_core.log.error` can set the variable to `false`.
By setting it to `true` here we can have a valid check per `module`.

monkeyDluffy6017:
hi @boekkooi-lengoo the code is a bit complicated, could you describe your thought in detail?
Which part of apisix.yaml can be validated?
Are you just going to check if the plugin is loaded correctly from your demos?

boekkooi-lengoo:
Hey @monkeyDluffy6017 

I agree the code is a bit complicated.
So my original goal was to validate the `apisix.yaml` schema (using "apisix.core.schema") at a minimum and if possible to validate the values of the config file as well.

Now it turns out that to use  "apisix.core.schema" I needed to require some of the `ngx` modules. 
To do this I created a setup similar to https://github.com/openresty/resty-cli/ which allows the [execution of lua in nginx as a command line command](https://github.com/apache/apisix/pull/10630/files#diff-228a51571b53709ad043d1c6c14fc8e508c248a84f8014fe548b5d259cedd5ddR957-R958).
This allow the ability to run code with all the requires modules.

After some reseacth it seems that it would be very hard to simply validate schemas as it seems that each module/part is doing validation differently. 
So for the sake of the prototype I split each module/part to it's own [validation call](https://github.com/apache/apisix/pull/10630/files#diff-228a51571b53709ad043d1c6c14fc8e508c248a84f8014fe548b5d259cedd5ddR898-R902).

Now for the validation we assume that any call to the [logger with level `error`](https://github.com/apache/apisix/pull/10630/files#diff-228a51571b53709ad043d1c6c14fc8e508c248a84f8014fe548b5d259cedd5ddR918-R923) means that the configuration is invalid.

Personally this entire thing feels a bit hacky but due to my lack of understanding of the internal design choices and lua knowledge I wanted to show this so we could analyses and discuss better options.

I hope this clarifies my thoughts a bit.

monkeyDluffy6017:
ok, we will figure out your design, please be patient

monkeyDluffy6017:
We have the command `apisix test` already, but it only can validate the generated nginx conf file, how can we unify the two functions?

boekkooi-lengoo:
@monkeyDluffy6017 I don't really understand your question as this code already adds the validation to apisix test in https://github.com/apache/apisix/pull/10630/files#diff-3936316fb5318f20b79ffe05648cf3d6b8c61e6987b51b52e2571c8a8e73a68bR795 can you maybe clarify what you mean?

monkeyDluffy6017:
@boekkooi-lengoo what i mean is that the `apisix test` comman only test `nginx.conf` before, you add the ability to test `apisix.yaml`, Can the 2 functions be confused?

boekkooi-lengoo:
@monkeyDluffy6017 they could be confusing as they confused me already because I assumed wrongly that `apisix test` would test the `apisix.yaml`. However, we could create a `apisix text-standalone` or something along those lines command.

monkeyDluffy6017:
OK, LGTM, please go ahead

boekkooi-lengoo:
Hey y'all,
I'm sorry but I sadly don't have time to spend on this feature at the moment.
For this reason I'm closing this PR.

Thanks again for the reviews, input and support I very much appreciate it!

PR #10645, opened 2023-12-13
feat: Provide `uri_arg_` when `radixtree_uri_with_parameter`

### LINKED ISSUE #10257: help request: about radixtree_uri_with_parameter
### Description

I used radixtree_uri_with_parameter to match parameters on the path, such as /path/:param. Currently, I want to use this param in the uri in proxy-rewrite. How should I handle it?

### Environment

- APISIX version (run `apisix version`): 3.2
- Operating system (run `uname -a`): debian

boekkooi-lengoo:
### Description

Add the prefix `uri_arg_` to the APISIX `ctx.var` allowing a APISIX configured to use the `radixtree_uri_with_parameter` router to access `opts.matched`. This allows the usage of the matches inside for example the `proxy_rewrite` plugin.

Also see https://github.com/api7/lua-resty-radixtree/#parameters-in-path

Fixes #10257

### Checklist


moonming:
@boekkooi-lengoo good catch 👍
Do we need to add more test cases for this PR?

boekkooi-lengoo:
@moonming I'm fine adding more just not sure which ones. Any suggestions what you think is missing?

monkeyDluffy6017:
@boekkooi-lengoo please make the ci pass

moonming:
> @moonming I'm fine adding more just not sure which ones. Any suggestions what you think is missing?

@boekkooi-lengoo you can add a test case where the uri parameter does not exist and see the failure situation. Others are LGTM

boekkooi-lengoo:
Good day @moonming and @monkeyDluffy6017,

Thanks for the review!
I added some extra test cases and a warning when a `uri_arg_<name>` is not found. As well as fixing the linting issue.

Please let me know if I missed anything :smile:

moonming:
> Good day @moonming and @monkeyDluffy6017,
> 
> 
> 
> Thanks for the review!
> 
> I added some extra test cases and a warning when a `uri_arg_<name>` is not found. As well as fixing the linting issue.
> 
> 
> 
> Please let me know if I missed anything :smile: 


great work. looks good to me👍

moonming (approved):

shreemaan-abhishek (dismissed):

apisix/core/ctx.lua
```diff
@@ -274,6 +274,15 @@ do
                     end
                 end
 
+            elseif core_str.has_prefix(key, "uri_arg_") then
+                if t._ctx.curr_req_matched then
+                    local arg_key = (sub_str(key, 9))
```
monkeyDluffy6017, line 279:
```suggestion
                    local arg_key = sub_str(key, 9)
```

apisix/core/ctx.lua
```diff
@@ -274,6 +274,15 @@ do
...
+                    val = t._ctx.curr_req_matched[arg_key]
+                    if val == nil then
+                        log.warn("failed to fetch uri argument: ", arg_key)
```
monkeyDluffy6017, line 282: I think we don't need the warn log like the other branches

apisix/core/ctx.lua
```diff
@@ -274,6 +274,15 @@ do
                     end
                 end
 
+            elseif core_str.has_prefix(key, "uri_arg_") then
```
monkeyDluffy6017, line 277: `uri_arg` always means URL query arguments, maybe you should use `uri_para_` instead?

monkeyDluffy6017, line 277: The built-in variable only works when the router engine is `radixtree_uri_with_parameter`, we should add comment about this

t/core/ctx_with_params.t
```diff
@@ -0,0 +1,174 @@
...
+
+            ngx.say("remote_addr: ", ctx.var["remote_addr"])
+            ngx.say("server_port: ", ctx.var["server_port"])
+        }
+    }
+--- request
+GET /t
+--- response_body
+remote_addr: 127.0.0.1
+server_port: 1984
```
monkeyDluffy6017, line 58: Why need this?

t/core/ctx_with_params.t
```diff
@@ -0,0 +1,174 @@
...
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "methods": ["GET"],
+                    "plugins": {
+                        "serverless-pre-function": {
+                            "phase": "access",
+                            "functions" : ["return function() ngx.log(ngx.INFO, \"uri_arg_path: \", ngx.ctx.api_ctx.var.uri_arg_path) end"]
```
monkeyDluffy6017, line 74: I think this test case is not appropriate, only the function `core.utils.resolve_var` can cover your newly added code

t/core/ctx_with_params.t
```diff
@@ -0,0 +1,174 @@
...
+
+
+
+=== TEST 3: `uri_arg_path` exist (hello)
+--- request
+GET /hello
+--- response_body
+hello world
+--- error_log
+uri_arg_path: hello
```
monkeyDluffy6017, line 106: Why do you test `uri_arg_path`? i think you should test the `param` in ` /path/:param` for example

apisix/core/ctx.lua
boekkooi-lengoo: This was inspired by https://github.com/apache/apisix/blob/master/apisix/core/ctx.lua#L245 
Not sure why this should not write a warning.

t/core/ctx_with_params.t
boekkooi-lengoo: I think you are confused because of the name see https://github.com/apache/apisix/pull/10645/files#diff-96d451e62e59f1ed3c91cabb0976f98151300573ee3a843e119b9b6d9001a2e2R83. 
I will change the name

t/core/ctx_with_params.t
boekkooi-lengoo: Not sure I agree with this as for example `cookie_` is tested in [ctx.t](https://github.com/apache/apisix/blob/master/t/core/ctx.t) and not in [utils.t](https://github.com/apache/apisix/blob/master/t/core/utils.t).
As far as I understand the reasoning for this it's because in `ctx*.t` tests are ran regarding the integration of nginx/openresty/third party context vars where as in `utils.t` is for testing the vars specific to apisix implementation.

However, please suggest the tests are you feel are missing and I will add them.

apisix/core/ctx.lua
boekkooi-lengoo: Partially correct, as `curr_req_matched` is set in https://github.com/apache/apisix/blob/538b964c3bb5a60d8b6016aa0b552439c6307f1b/apisix/http/router/radixtree_host_uri.lua#L83 and based on https://github.com/apache/apisix/blob/538b964c3bb5a60d8b6016aa0b552439c6307f1b/apisix/http/router/radixtree_host_uri.lua#L179 it seems `_host` is even valid without `radixtree_uri_with_parameter` also looking at [`lib/resty/radixtree.lua`](https://github.com/api7/lua-resty-radixtree/blob/master/lib/resty/radixtree.lua) it seems `_action`, `_path` and `_host` can be available.

boekkooi-lengoo: How about `uri_param`?

t/core/ctx_with_params.t
membphis: We can remove it if it is useless

apisix/core/ctx.lua
monkeyDluffy6017: LGTM

apisix/core/ctx.lua
monkeyDluffy6017: The reason why the place you pointed add a warn log is due to the the error returned needs to be handled, but you don't have an error to handle, right?
image

apisix/core/ctx.lua
monkeyDluffy6017: But your title is `Provide uri_arg_ when radixtree_uri_with_parameter`, and you need to tell the users what they can access

t/core/ctx_with_params.t
monkeyDluffy6017: ok

t/core/ctx_with_params.t
monkeyDluffy6017: I think you are right

apisix/core/ctx.lua
boekkooi-lengoo: I have remove the log message.

apisix/core/ctx.lua
boekkooi-lengoo: After some testing it's seems only `_path` is consistently available so I added a comment. 
Is this good enough?

t/core/ctx_with_params.t
boekkooi-lengoo: Renamed it to `id`

t/core/ctx_with_params.t
boekkooi-lengoo: Removed this test case.

monkeyDluffy6017:
please make the ci pass

monkeyDluffy6017:
please merge the master

boekkooi-lengoo:
@monkeyDluffy6017 Created merge commit https://github.com/apache/apisix/pull/10645/commits/45ee4a203a52d45e049294b295f670a4def1f279 as requested

monkeyDluffy6017 (approved):

AlinsRan (approved):

PR #10670, opened 2023-12-19
fix(openid-connect): use_jwks breaking authentication header

### LINKED ISSUE #10592: bug: open-id connect plugin sets bearer_only=true and returns 401 instead of checking the token.
### Current Behavior

I am facing a similar issue with APISIX. I deployed APISIX in GCP GKE and tried integrating it with Auth0. In my case my API GW is accessed from a FE app which already performs login with Auth0 and thus has a valid token.
I configured my APISIX route like this

{
  "uri": "/my-service/*",
  "name": "My Route",
  "plugins": {
    "openid-connect": {
      "bearer_only": true,
      "client_id": "AUTH0_CLIENTID",
      "client_secret": "AUTH0_SECRET",
      "discovery": AUTH0_DOMAIN.well-known/openid-configuration",
      "redirect_uri": "http://localhost:8082/my-service/redirect",  // same as Auth0 callback as mentioned in previous threads
      "scope": "openid profile"
    }
  },
  "upstream_id": "1",
  "status": 1
}
When calling the api with a -H "Authorization: Bearer AUTH0_TOKEN" i get a 401 Authorization Required error, why is this?
If I set the bearer_only to false and still call the API with -H "Authorization: Bearer AUTH0_TOKEN" it redirects me to login. It looks like API SIX ignores the fact that I have a valid token.
If I call the endpoint from browser then I get redirected to Auth0, I login, get a session and redirect to my api all good.

I see myself forced to use API SIX as a simple API GW without any security and handle the security in each upstream service individually. It is quite sad as this project has a lot of potential.

If anybody has encountered this issue before and has a solution please do share!

### Expected Behavior

_No response_

### Error Logs

_No response_

### Steps to Reproduce

{
  "uri": "/my-service/*",
  "name": "My Route",
  "plugins": {
    "openid-connect": {
      "bearer_only": true,
      "client_id": "AUTH0_CLIENTID",
      "client_secret": "AUTH0_SECRET",
      "discovery": AUTH0_DOMAIN.well-known/openid-configuration",
      "redirect_uri": "http://localhost:8082/my-service/redirect",  // same as Auth0 callback as mentioned in previous threads
      "scope": "openid profile"
    }
  },
  "upstream_id": "1",
  "status": 1
}

### Environment

### LINKED ISSUE #10669: bug: Openid-connect plugin `use_jwks` breaks Authorization headers when bearer_only is true
### Current Behavior

When using the openid-connect plugin with `bearer_only` as false, the plugin will normally authenticate both requests with the session cookie and requests with an explicit `Authorization: Bearer` header. This is very useful to allow both browser based and programmatic access to some resource.
E.g.
```yaml
plugins:
    openid-connect:
      client_id: ...
      client_secret: ...
      discovery: ...
      scope: "openid profile"
      bearer_only: false
      realm: ...
      public_key: ...
```

However, if you use jwks instead of explicitly setting the public key, requests with a proper `Authorization: Bearer` header still get the `302 Found` response.

### Expected Behavior

Authorization headers should get checked the same way whether you are manually setting a public key or using jwks.

### Error Logs

_No response_

### Steps to Reproduce

Create a route with an openid-connect configuration like so
```yaml
plugins:
    openid-connect:
      client_id: ...
      client_secret: ...
      discovery: ...
      scope: "openid profile"
      bearer_only: false
      realm: ...
      use_jwks: true
```
```bash
curl https://example.com -H "Authorization: Bearer $TOKEN"
```
will return a an HTML 302 error page.

I believe the issue is caused by not checking for `use_jwks` on [this line](https://github.com/apache/apisix/blob/4f0b85f5aa0667de177da7106447c11d7a051142/apisix/plugins/openid-connect.lua#L454C33-L454C55)

### Environment

- APISIX version (run `apisix version`): 3.7.0
- Operating system (run `uname -a`): Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): openresty/1.21.4.2
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): N/A (standalone mode)

jesse-r-s-hines:
### Description
When using the openid-connect plugin with `bearer_only` as false, the plugin will normally authenticate both requests with the session cookie and requests with an explicit `Authorization: Bearer` header. This is very useful to allow both browser based and programmatic access to some resource.

However, if you set `use_jwks` instead of explicitly setting the public key, requests with a proper `Authorization: Bearer` header still get the `302 Found` response.

Fix by adding `conf.use_jwks` to the check.

Fixes #10669

### DIFF
--- a/t/plugin/openid-connect.t
+++ b/t/plugin/openid-connect.t
@@ -1425,3 +1425,107 @@ passed
 --- response_body
 true
 --- error_code: 302
+
+
+
+=== TEST 38: Update plugin config to use_jwk and bear_only false
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "plugins": {
+                            "openid-connect": {
+                                "client_id": "course_management",
+                                "client_secret": "d1ec69e9-55d2-4109-a3ea-befa071579d5",
+                                "discovery": "http://127.0.0.1:8080/realms/University/.well-known/openid-configuration",
+                                "redirect_uri": "http://localhost:3000",
+                                "ssl_verify": false,
+                                "timeout": 10,
+                                "bearer_only": false,
+                                "use_jwks": true,
+                                "realm": "University",
+                                "introspection_endpoint_auth_method": "client_secret_post",
+                                "introspection_endpoint": "http://127.0.0.1:8080/realms/University/protocol/openid-connect/token/introspect"
+                            }
+                        },
+                        "upstream": {
+                            "nodes": {
+                                "127.0.0.1:1980": 1
+                            },
+                            "type": "roundrobin"
+                        },
+                        "uri": "/hello"
+                }]]
+                )
+
+            if code >= 300 then
+                ngx.status = code
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 39: Test that jwt with bearer_only false still allows a valid Authorization header
+--- config
+    location /t {
+        content_by_lua_block {
+            -- Obtain valid access token from Keycloak using known username and password.
+            local json_decode = require("toolkit.json").decode
+            local http = require "resty.http"
+            local httpc = http.new()
+            local uri = "http://127.0.0.1:8080/realms/University/protocol/openid-connect/token"
+            local res, err = httpc:request_uri(uri, {
+                    method = "POST",
+                    body = "grant_type=password&client_id=course_management&client_secret=d1ec69e9-55d2-4109-a3ea-befa071579d5&username=teacher@gmail.com&password=123456",
+                    headers = {
+                        ["Content-Type"] = "application/x-www-form-urlencoded"
+                    }
+                })
+
+            -- Check response from keycloak and fail quickly if there's no response.
+            if not res then
+                ngx.say(err)
+                return
+            end
+
+            -- Check if response code was ok.
+            if res.status == 200 then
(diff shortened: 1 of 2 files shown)


monkeyDluffy6017:
Could you add test cases like https://github.com/apache/apisix/blob/master/t/plugin/openid-connect.t#L900

jesse-r-s-hines:
I'm still a bit confused on how the APISIX tests work, but I've added a test that should check for the bug in #10670.

jesse-r-s-hines:
I can't replicate the linting failure by running `make lint` locally, how do I get it to tell me what's causing the failure?

kayx23:
Trying running reindex? https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#check-code-style-and-test-case-style

jesse-r-s-hines:
I've tried that, `reindex` makes no change to the file when I run it.

t/plugin/openid-connect.t
```diff
@@ -1241,7 +1345,7 @@ passed
 
 
 
-=== TEST 33: Access route to validate "x-userinfo" in request header
+=== TEST 35: Access route to validate "x-userinfo" in request header
```
monkeyDluffy6017, line 1348: I suggest put your newly added test cases to the end of the file, so that you won't need to modify the index of the rest test cases. 
The index of test cases below this one are not correct

jesse-r-s-hines: Oh of course, I'm not sure how I missed that. I've updated the test file.

jesse-r-s-hines:
I believe the linting should pass now.

monkeyDluffy6017 (approved):

moonming (approved):

Issue #9920, opened 2023-07-28
help request: content-based healthcheck


qwzhou89:
### Description

There is a need to check whether a field in the upstream's return body (in json format) is the specified value, and only the field with the specified value is considered to be a healthy node when doing a healthcheck on a certain route.
Is there a way to customize this to the current healthcheck?

theweakgod:
@qwzhou89 I think it is not necessary to check a certain field in the body. I think it is very simple to make an API specifically for health checks to match the entire content of the body. What you think.

PR #10697, opened 2023-12-22
feat: add body match in active health check

### LINKED ISSUE #10517: feat: As a user, I want to specify an "expected string" in http health checks
### Description

As a user, I want to specify an "expected string" parameter in the definition of an http health check, so that APISIX can tell when a webserver is not interpreting the programming language correctly and instead serves the files in plain-text.

For example, if the expected string is `pass` and I generate it with something like `echo 'p', 'a', 's', 's';`, the health check will disable the machine when the php code is not interpreted. Yes, it might take down all the nodes, but that's preferable to the other scenario, where the underlying code is displayed on the webpage.

theweakgod:
### Description
Add body match in active health check.

Fixes #10517  (https://github.com/apache/apisix/issues/10517)

Add the body_match_str parameter in the lua-resty-healthcheck module to check if the response body matches the expected string.

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -574,6 +574,9 @@ function _M.rewrite(plugin_conf, ctx)
             end
         end
     end
+    if session then
+        session:close()
+    end
 end
 
 
--- a/t/plugin/openid-connect5.t
+++ b/t/plugin/openid-connect5.t
@@ -1,0 +1,140 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+log_level('debug');
+repeat_each(1);
+no_long_string();
+no_root_location();
+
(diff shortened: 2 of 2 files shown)


monkeyDluffy6017:
The pr is not ready, right?

theweakgod:
> The pr is not ready, right?

https://github.com/api7/lua-resty-healthcheck/pull/43
This should be combined with this pr

monkeyDluffy6017:
OK, thanks for your contribution, this pr should merged after https://github.com/api7/lua-resty-healthcheck/pull/43, and we need to upgrade the lua-resty-healthcheck version in APISIX, the next version of lua-resty-healthcheck should be 3.2.1

monkeyDluffy6017:
I've taken a closer look at the requirement and I've realised that it's not very clear and there's probably no way to make it a very generic feature, e.g. whether the health check request sent needs to carry a specific body of the request, whether the health check result returned should be checked for a particular field or for the whole result.
These 2 issues describe 2 different requirements: https://github.com/apache/apisix/issues/10517 and https://github.com/apache/apisix/issues/9920

theweakgod:
> I've taken a closer look at the requirement and I've realised that it's not very clear and there's probably no way to make it a very generic feature, e.g. whether the health check request sent needs to carry a specific body of the request, whether the health check result returned should be checked for a particular field or for the whole result. These 2 issues describe 2 different requirements: #10517 and #9920



> I've taken a closer look at the requirement and I've realised that it's not very clear and there's probably no way to make it a very generic feature, e.g. whether the health check request sent needs to carry a specific body of the request, whether the health check result returned should be checked for a particular field or for the whole result. These 2 issues describe 2 different requirements: #10517 and #9920

I think it is better for the health check to do a simple whole-body verification, without considering the content-type. Some users may really need a simple body verification to determine whether they are healthy. If you target a certain field, different content-types will make the parsing more complicated. I think the easiest and best way at present is to verify the entire body.

monkeyDluffy6017:
let's hold this pr to see if others agree with this design, there are a number of designs that use a heartbeat mechanism, such as sending `aabb`, then checking to see if what is received is `bbaa`

theweakgod:
> let's hold this pr to see if others agree with this design, there are a number of designs that use a heartbeat mechanism, such as sending `aabb`, then checking to see if what is received is `bbaa`

Alright, I got it

PR #10754, opened 2024-01-05
feat: allow for conditional access logging

### LINKED ISSUE #10407: feat: As a user, I want to disable logs for certain endpoints for compliance reasons
### Description

Our system has a couple of endpoints where the query parameters can include sensitive information. One option is not to log query parameters or disable logging altogether. Neither of these solutions is very appealing. Instead, we would like to exclude these routes from access logging.

This is how other providers are doing it:
* ingress-nginx-controller uses [skip-access-log-urls](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#skip-access-log-urls)
* nginx itself does it through [access_log+if](https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#enabling-conditional-logging) directive

APISIX already dynamically sets some settings (e.g. [apisix_ngx_client.set_client_max_body_size](https://github.com/apache/apisix/blob/5911211d2292b4c96df0c666fcc74fabeb10a5d8/apisix/plugins/client-control.lua#L66)). Maybe something similar could be used where we can set `access_log` to `off` if request_uri matches?

kayx23:
### Description

Allow for conditional access logging with NGINX's `if` parameter.

Closes #10407

### Checklist

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -377,9 +377,9 @@ http {
     uninitialized_variable_warn off;
 
     {% if http.access_log_buffer then %}
-    access_log {* http.access_log *} main buffer={* http.access_log_buffer *} flush=3;
+    access_log {* http.access_log *} main {% if http.access_log_if_condition then %} if={* http.access_log_if_condition *}{% end %} buffer={* http.access_log_buffer *} flush=3;
     {% else %}
-    access_log {* http.access_log *} main buffer=16384 flush=3;
+    access_log {* http.access_log *} main {% if http.access_log_if_condition then %} if={* http.access_log_if_condition *}{% end %} buffer=16384 flush=3;
     {% end %}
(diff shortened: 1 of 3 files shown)


monkeyDluffy6017:
Impressive! but if is not suggested in nginx, https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

kayx23:
> Impressive! but if is not suggested in nginx, https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

The first sentence of the article mentions it is not recommended "when used in location context" and I think here we are not using if in location? Does the same concern still apply?

monkeyDluffy6017:
Yeah, `if` is forbidden in our project, we don't know what problem it will cause

kayx23:
Oops ok, didn't know that, but good to know now. This PR can be closed if implementation with `if` is a major concern.

PR #10771, opened 2024-01-07
change: unify the keyring and key_encrypt_salt fields, remove ssl:key_encrypt_salt

### LINKED ISSUE #10484: feat: As a user, I want to unify the keyring and key_encrypt_salt fields
### Description

As we can see, the [keyring](https://github.com/apache/apisix/blob/d8bd34414239ed36942c19f435ac10ad0d5e57c0/conf/config-default.yaml#L133) and the [key_encrypt_salt](https://github.com/apache/apisix/blob/d8bd34414239ed36942c19f435ac10ad0d5e57c0/conf/config-default.yaml#L111) are used in the same way and with the same meaning, so i suggest that if we could unify the keyring and key_encrypt_salt fields?

RitterHou:
### Description

This is a fix for issue 10484, the field ssl:key_encrypt_salt has been removed, all the encrypt and decrypt operations are using data_encryption:keyring field now.

key_encrypt_salt and keyring are all used for data encrypt and decrypt, but in fact the are doing the same thing, this change remove key_encrypt_salt only field keyring need to be reserved.

Fixes #10484 

### Checklist

- [ ] I have verified that this change is backward compatible

### DIFF
--- a/apisix/ssl.lua
+++ b/apisix/ssl.lua
@@ -90,17 +90,6 @@ local function init_iv_tbl(ivs)
 end
 
 
-local _aes_128_cbc_with_iv_tbl_ssl
-local function get_aes_128_cbc_with_iv_ssl(local_conf)
-    if _aes_128_cbc_with_iv_tbl_ssl == nil then
-        local ivs = core.table.try_read_attr(local_conf, "apisix", "ssl", "key_encrypt_salt")
-        _aes_128_cbc_with_iv_tbl_ssl = init_iv_tbl(ivs)
-    end
-
-    return _aes_128_cbc_with_iv_tbl_ssl
-end
-
-
 local _aes_128_cbc_with_iv_tbl_gde
 local function get_aes_128_cbc_with_iv_gde(local_conf)
     if _aes_128_cbc_with_iv_tbl_gde == nil then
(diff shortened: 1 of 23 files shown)


monkeyDluffy6017:
Thanks for contribution, we will check it later

monkeyDluffy6017:
I think you should update the related test cases too

conf/config-default.yaml
```diff
@@ -132,7 +122,7 @@ apisix:
     enable: false
```
monkeyDluffy6017, line 122: we should put this attribute `enable` under the `keyring`, because it's only for `encrypt_fields` in plugin schma, and change a name, modify the comments

RitterHou: how about changing `enable` to `enable_encrypt_fields` and leave it still at this place, because it's under `data_encryption` and the attribute name has shown its function

RitterHou:
> I think you should update the related test cases too

i will add them

RitterHou:
I have updated the related test cases and fixed some problems @monkeyDluffy6017

apisix/ssl.lua
```diff
@@ -125,21 +114,18 @@ end
...
-        local aes_128_cbc_with_iv_tbl_ssl = get_aes_128_cbc_with_iv_ssl(local_conf)
-        local aes_128_cbc_with_iv_ssl = aes_128_cbc_with_iv_tbl_ssl[1]
-        if aes_128_cbc_with_iv_ssl ~= nil and core.string.has_prefix(origin, "---") then
-            return encrypt(aes_128_cbc_with_iv_ssl, origin)
-        end
+    if aes_128_cbc_with_iv_gde == nil then
+        return origin
+    end
+
+    if not field and core.string.has_prefix(origin, "---") then
```
monkeyDluffy6017, line 124: Do we still need the `field` parameter?

RitterHou: I reserved this parameter because ssl encryption is need to check whether the `origin` has a `---` prefix and data  encryption is not. If I remove it, is the prefix check of `origin` should be also removed too? or just reserve the check and put it to ssl and data encryption?

monkeyDluffy6017: I prefer the later one: `reserve the check and put it to ssl and data encryption`, the `data encryption` don't have a prefix like `---`, so no need to consider the `data encryption`

monkeyDluffy6017:
please make the ci pass

RitterHou:
Hi @monkeyDluffy6017, I.am feeling confused on  how to fix `Error: nginx: [error] open() "/home/runner/work/apisix/apisix/logs/nginx.pid" failed (2: No such file or directory)` error, I have the same problem in local environment and I manual created the `logs` directory to fix this error. I am not sure how to fix this to pass the ci, is any prompt for fixing this error? thank you.

monkeyDluffy6017:
@RitterHou This error log can be ignored: `Error: nginx: [error] open() "/home/runner/work/apisix/apisix/logs/nginx.pid" failed (2: No such file or directory)`, please pay attention to the other errors

RitterHou:
@monkeyDluffy6017 I switched the branch to master and tested `t/router/radixtree-sni2.t`, the tests on the master branch still have errors. I am stuck fixing the test error now, can you give me some advice to fix the errors? I would be very grateful. Thank you.

monkeyDluffy6017:
@RitterHou I think the error is imported by your modifications
image

RitterHou:
I have found the reason for this error, the SSL key encryption must check whether the key is started with `---`, if the key is not started with `---`, the encryption will not work and the origin SSL key will be returned. However, the `data encryption` doesn't have this limit. If I add the check of the prefix, the SSL key check will be ok, but the data encryption will never work because it doesn't have a `---` prefix. But if I remove the check, the SSL key encryption will have the error I mentioned before.

Can you help me to solve this problem? @monkeyDluffy6017

> @RitterHou I think the error is imported by your modifications image

monkeyDluffy6017:
Maybe we should add the `field` back: https://github.com/apache/apisix/pull/10771/commits/b6501cd914df76b352e2d58a210fc2fe0b06abf4

monkeyDluffy6017 (approved):

shreemaan-abhishek (approved):

PR #10792, opened 2024-01-10
ci: remove unnecessary ci file

monkeyDluffy6017:
### Description

The ci file is unnecessary, we don't need to install the lua deps of master branch in current branch
The following ci file is enough which install the lua deps of current branch:
https://github.com/apache/apisix/blob/master/.github/workflows/cli.yml

Fixes # (issue)

### Checklist


AlinsRan:
Removing this file will cause `cli-master` to report an error, and we still need to delete the file https://github.com/apache/apisix/blob/master/.github/workflows/cli-master.yml#L23

AlinsRan (approved):

nic-6443 (approved):

PR #10854, opened 2024-01-19
fix: add healthcheck in response body

### LINKED ISSUE #10517: feat: As a user, I want to specify an "expected string" in http health checks

theweakgod:
### Description

Fixes #10517 (https://github.com/apache/apisix/issues/10517)
hold in another branch
[origin pr](https://github.com/apache/apisix/pull/10697)

### Checklist


Vacant2333:
hello, is the pr ready to review now? @theweakgod

PR #10862, opened 2024-01-23
feat: body-transformer plugin for other data formats without warnings

### LINKED ISSUE #10681: feat: As a user, I want to to use the body-transformer plugin for other data formats without warnings
### Description

The body-transformer plugin is a very flexible plugin to modify the body content of requests and responses, and we can already use the `body-transformer` plugin to manipulate arbitrary data formats with the lua templates. However, if you don't specify `input_format` and `Content-Type` isn't JSON or XML it will always log a warning about "no input format to parse request/response body".

Adding another format to the `input_format` enum, `none`, that explicitly tells `body-transformer` to not try to parse or validate the body content and silences this warning would be convenient.

smileby:
### Description

Fixes #10681

Adding another format to the input_format enum, none, that explicitly tells body-transformer to not try to parse or validate the body content and silences this warning would be convenient.

### Checklist

### DIFF
--- a/apisix/plugins/body-transformer.lua
+++ b/apisix/plugins/body-transformer.lua
@@ -34,7 +34,7 @@ local next              = next
 local transform_schema = {
     type = "object",
     properties = {
-        input_format = { type = "string", enum = {"xml", "json", "encoded", "args"} },
+        input_format = { type = "string", enum = {"xml", "json", "encoded", "args", "none"} },
         template = { type = "string" },
         template_is_base64 = { type = "boolean" },
     },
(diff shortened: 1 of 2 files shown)


apisix/plugins/body-transformer.lua
```diff
@@ -34,7 +34,7 @@ local next              = next
 local transform_schema = {
     type = "object",
     properties = {
-        input_format = { type = "string", enum = {"xml", "json", "encoded", "args"} },
+        input_format = { type = "string", enum = {"xml", "json", "encoded", "args", "none"} },
```
monkeyDluffy6017, line 37: Is `plain` better?

shreemaan-abhishek: I agree

smileby: I'll change it

monkeyDluffy6017 (approved):

shreemaan-abhishek (approved):

PR #10866, opened 2024-01-24
feat: add redis and redis-cluster in limit-conn

### LINKED ISSUE #10501: feat: centralized counter for limit conn and limit req
### Description

The limit-count plugin allows using `redis` or `redis-cluster` policy to centralise the counter for rate limiting.

> Rate-limiting policies to use for retrieving and increment the limit count. When set to local the counters will be locally stored in memory on the node. When set to redis counters are stored on a Redis server and will be shared across the nodes. It is done usually for global speed limiting, and setting to redis-cluster uses a Redis cluster instead of a single instance.

This is specially helpful when there are multiple instances of APISIX.

We should support similar feature for `limit-conn` and `limit-req` as well.

theweakgod:
### Description

Fixes #10501 (https://github.com/apache/apisix/issues/10501)

Add options for redis and redis-cluster in the limit-conn plugin, and use redis as the counter for the leaky bucket algorithm.

### Checklist

### DIFF
--- a/apisix/plugins/limit-conn/util.lua
+++ b/apisix/plugins/limit-conn/util.lua
@@ -60,7 +60,6 @@ function _M.incoming(self, red, key, commit)
 end
 
 
-
 function _M.leaving(self, red, key, req_latency)
     assert(key)
     key = "limit_conn" .. ":" .. key
--- a/apisix/utils/redis.lua
+++ b/apisix/utils/redis.lua
@@ -67,7 +67,6 @@ local function redis_cli(conf)
 end
 
 
-
 function _M.new(conf)
     return redis_cli(conf)
 end


shreemaan-abhishek:
Let's bring support for `limit-conn` first then you can raise another PR for `limit-req` it's easier to review.

theweakgod:
> Let's bring support for `limit-conn` first then you can raise another PR for `limit-req` it's easier to review.

I'll split it into two pull requests. I encountered an issue during development related to the management of Redis keys. I noticed that APISIX generates a significant number of Redis keys. I'm considering consolidating these keys into a hash for better management. What are your thoughts on this?

theweakgod:
@shreemaan-abhishek @monkeyDluffy6017 I have a problem now. I need to write a test case to test the plugin, but what address should I fill in for the redis cluster? How to fill in the cluster with password and the cluster without password? How to fill in the address of the redis singleton with password and without password?

shreemaan-abhishek:
@theweakgod you will find redis servers (singleton and cluster) defined as docker compose applications in `ci` directory. You can use them or create new ones. You can also take inspiration from existing test cases for redis policy in limit count plugin.

theweakgod:
> @theweakgod you will find redis servers (singleton and cluster) defined as docker compose applications in `ci` directory. You can use them or create new ones. You can also take inspiration from existing test cases for redis policy in limit count plugin.

Ok.I see, thank u.

theweakgod:
@shreemaan-abhishek @monkeyDluffy6017 I'm ready to review

apisix/plugins/limit-conn.lua
```diff
@@ -19,18 +19,95 @@ local limit_conn = require("apisix.plugins.limit-conn.init")
...
+                type = "string", minLength = 1,
+            },
+            redis_password = {
+                type = "string", minLength = 0,
+            },
+            redis_prefix = {
+                type = "string", minLength = 0, default = "limit_conn", pattern = "^[0-9a-zA-Z|_]+$"
+            },
+            redis_database = {
+                type = "integer", minimum = 0, default = 0,
```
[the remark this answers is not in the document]
theweakgod: Because the default database of redis is zero.

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,158 @@
...
+
+local _M = {version = 0.1}
+
+
+local mt = {
+    __index = _M
+}
+
+
+local function new_redis_cluster(conf)
```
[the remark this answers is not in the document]
theweakgod: In order to distinguish redis and redis-cluster, I will create two new files in apisix/utils/. One is redis.lua and the other is redis-cluster.lua. what you think?

apisix/plugins/limit-conn.lua
```diff
@@ -19,18 +19,95 @@ local limit_conn = require("apisix.plugins.limit-conn.init")
...
-        key_type = {type = "string",
+        key = {
+            type = "string"
+        },
+        key_type = {
+            type = "string",
             enum = {"var", "var_combination"},
             default = "var",
         },
+        counter_type = {
```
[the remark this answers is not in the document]
theweakgod: Ok

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,158 @@
...
+    return setmetatable(self, mt)
+end
+
+
+function _M.incoming(self, key, commit)
+    local max = self.max
+
+    -- init redis
+    local conf = self.conf
+    local red, err = new_redis_cluster(conf)
```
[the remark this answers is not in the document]
theweakgod: OK

theweakgod:
@shreemaan-abhishek all done

apisix/plugins/limit-conn.lua
```diff
@@ -19,18 +19,95 @@ local limit_conn = require("apisix.plugins.limit-conn.init")
...
+        },
+        key_type = {
+            type = "string",
             enum = {"var", "var_combination"},
             default = "var",
         },
+        policy = {
+            type = "string",
+            enum = {"redis", "redis-cluster", "shared-dict"},
+            default = "shared-dict",
```
shreemaan-abhishek, line 109: replace `shared-dict` with `local` and please make changes else where to ensure consistency in implementation with the `limit-count` redis and redis-cluster.

apisix/plugins/limit-conn.lua
```diff
@@ -19,18 +19,95 @@ local limit_conn = require("apisix.plugins.limit-conn.init")
 
 
 local plugin_name = "limit-conn"
+
+local policy_to_additional_properties = {
```
shreemaan-abhishek, line 23: I think this part of code is also duplicated, please look into how we can avoid code duplication.

shreemaan-abhishek: you also need to make `limit-count` use the code `utils/redis.lua` to avoid code duplication.

theweakgod: > I think this part of code is also duplicated, please look into how we can avoid code duplication.

No, I don't think so. There may be some parameters in "local" that need to be customized. So this does not need to be made publicly available. I think it would be nice to have a template for everyone to use.

apisix/plugins/limit-conn.lua
theweakgod: done

apisix/plugins/limit-conn.lua
theweakgod: done

theweakgod:
all done @shreemaan-abhishek

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,139 @@
...
+    local conf = self.conf
+
+    self.committed = false
+
+    local prefix = conf.redis_prefix
+    local hash_key = prefix .. ":connection_hash"
+
+    local conn, err
+    if commit then
+        conn, err = red:hincrby(hash_key, key, 1)
```
shreemaan-abhishek, line 63: I couldn't find the declaration of this function `hincrby` can you help me find it?

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,139 @@
...
+
+
+function _M.is_committed(self)
+    return self.committed
+end
+
+
+local function leaving_thread(premature, self, key, req_latency)
+
+    -- init redis
```
shreemaan-abhishek, line 102: remove this

apisix/plugins/limit-conn/limit-conn-redis-single.lua
shreemaan-abhishek, line 1: change the filename to `limit-conn-redis`

apisix/plugins/limit-conn/limit-conn-redis-single.lua
```diff
@@ -0,0 +1,143 @@
...
+    return setmetatable(self, mt)
+end
+
+
+function _M.incoming(self, key, commit)
+    local max = self.max
+
+    -- init redis
+    local conf = self.conf
+    local red, err = redis.new(conf)
```
shreemaan-abhishek, line 52: why create a new redis client for each request?

apisix/plugins/limit-conn/limit-conn-redis-single.lua
```diff
@@ -0,0 +1,143 @@
...
+function _M.is_committed(self)
+    return self.committed
+end
+
+
+local function leaving_thread(premature, self, key, req_latency)
+
+    -- init redis
+    local conf = self.conf
+    local red, err = redis.new(conf)
```
shreemaan-abhishek, line 105: why create a new redis here again??

apisix/utils/rediscluster.lua
shreemaan-abhishek, line 1: rediscluster.lua can be merged to redis.lua

apisix/plugins/limit-conn.lua
```diff
@@ -14,23 +14,41 @@
...
         burst = {type = "integer",  minimum = 0},
         default_conn_delay = {type = "number", exclusiveMinimum = 0},
         only_use_default_delay = {type = "boolean", default = false},
-        key = {type = "string"},
-        key_type = {type = "string",
+        key = {
+            type = "string"
+        },
+        key_type = {
+            type = "string",
```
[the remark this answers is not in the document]
shreemaan-abhishek: I agree this is clean. But as long as the linter is happy, let's not make any more changes.

Another reason is `git blame` will show this change is related to `feat: add redis .....` PR and this change doesn't relate well to this PR. I hope I didn't confuse you.

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,139 @@
...
+
+
+function _M.incoming(self, key, commit)
+    local max = self.max
+    local red = self.red_cli
+    local conf = self.conf
+
+    self.committed = false
+
+    local prefix = conf.redis_prefix
```
shreemaan-abhishek, line 58: why is it necessary to use prefix

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
shreemaan-abhishek: ok I remember this now. Thank you.

apisix/plugins/limit-conn.lua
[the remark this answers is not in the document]
theweakgod: Ok，I'll fix it

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
theweakgod: Because the keys called by income are similar to 192.168.1.1route40, the prefix is added to distinguish plugins and facilitate key management, because users may use the same redis, and all of them are suitable for current limiting plugins.

shreemaan-abhishek: I think we can prefix the key with the plugin name so the key can become `limit-conn192.168.1route40`.

shreemaan-abhishek: like this:

https://github.com/shreemaan-abhishek/apisix/blob/0bea3dbf6b1b868ba35acdb2fd0205f2a352479d/apisix/plugins/limit-count/limit-count-redis.lua#L109

theweakgod: OK, done.

apisix/plugins/limit-conn/limit-conn-redis.lua
```diff
@@ -0,0 +1,139 @@
...
+    }
+    return setmetatable(self, mt)
+end
+
+
+function _M.incoming(self, key, commit)
+    local max = self.max
+
+    local conf = self.conf
+    local red, err = redis.new(conf)
```
shreemaan-abhishek, line 51: move this to `_M.new()`

theweakgod: This cannot be placed in new because you cannot ensure whether it times out.

apisix/plugins/limit-conn/limit-conn-redis.lua
```diff
@@ -0,0 +1,139 @@
...
+        plugin_name = plugin_name,
+        burst = burst,
+        max = max + 0,    -- just to ensure the param is good
+        unit_delay = default_conn_delay,
+    }
+    return setmetatable(self, mt)
+end
+
+
+function _M.incoming(self, key, commit)
```
shreemaan-abhishek, line 47: the `incoming` function is same to same in `redis-cluster`. How do you plan to unify this?

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,134 @@
...
+
+
+function _M.is_committed(self)
+    return self.committed
+end
+
+
+local function leaving_thread(premature, self, key, req_latency)
+
+    local red = self.red_cli
```
shreemaan-abhishek, line 100: as per our previous communication we concluded that redis connection might be lost so we need to instantiate a new connection in `log_phase`. Why do you use existing connection here then?

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,134 @@
...
+    return 0, conn
+end
+
+
+function _M.is_committed(self)
+    return self.committed
+end
+
+
+local function leaving_thread(premature, self, key, req_latency)
```
shreemaan-abhishek, line 98: the `leaving_thread` function logic is also repeated (duplicate/same to same) please unify them.

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,134 @@
...
+    if req_latency then
+        local unit_delay = self.unit_delay
+        self.unit_delay = (req_latency + unit_delay) / 2
+    end
+
+    return conn
+end
+
+
+function _M.leaving(self, key, req_latency)
```
shreemaan-abhishek, line 118: ditto

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
theweakgod: Because this is a cluster, the cluster establishes a new connection every time do_cmd is used, so just maintain a red_cli here.
https://github.com/steve0511/resty-redis-cluster/blob/master/lib/resty/rediscluster.lua#L548-L559

apisix/utils/rediscluster.lua
```diff
@@ -0,0 +1,60 @@
...
+
+local _M = {version = 0.1}
+
+local function new_redis_cluster(conf)
+    local config = {
+        name = conf.redis_cluster_name,
+        serv_list = {},
+        read_timeout = conf.redis_timeout,
+        auth = conf.redis_password,
+        dict_name = "plugin-limit-conn-redis-cluster-slot-lock",
```
shreemaan-abhishek, line 29: this function is also being used by `limit-conn` so the dict name needs to be parameterized.

theweakgod: OK

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
shreemaan-abhishek: ok. btw, please don't resolve conversations by yourself.

apisix/plugins/limit-conn/limit-conn-redis.lua
shreemaan-abhishek: what do you mean?

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
theweakgod: > ok. btw, please don't resolve conversations by yourself.

Ok, my first time using😂

apisix/plugins/limit-conn/limit-conn-redis.lua
theweakgod: > what do you mean?

Because Redis uses a cosocket for its operations, it is susceptible to timeouts. However, RedisCluster is built on resty.redis. It establishes a connection every time do_cmd is executed. Therefore, we need to establish a Redis connection every time there is incoming or leaving activity to prevent cosocket timeouts.

apisix/utils/rediscluster.lua
theweakgod: done

apisix/plugins/limit-conn/limit-conn-redis.lua
theweakgod: Perhaps we can create a util.lua and pass red_cli as a parameter to avoid duplicating the code for the incoming and leaving functions?

apisix/plugins/limit-conn/limit-conn-redis-cluster.lua
```diff
@@ -0,0 +1,134 @@
...
+
+
+local mt = {
+    __index = _M
+}
+
+
+function _M.new(plugin_name, conf, max, burst, default_conn_delay)
+
+    local red_cli, err = redis_cluster.new(conf, "plugin-limit-conn-redis-cluster-slot-lock")
```
shreemaan-abhishek, line 35: why is the second parameter needed?

theweakgod: Because rediscluster requires a shared-dict.

shreemaan-abhishek:
@theweakgod I hope you noticed my comment that suggested prefixing the key with the plugin name, following this idea would eliminate the use of `red:hincr` thus your logic that depends on "_run_cmd()" instantiating a new redis connection to run the command will become incorrect. There will be more code changes. And I have left some comments regarding unification of the repeated logics. Please make this changes when you can and then I will review again.

theweakgod:
> @theweakgod I hope you noticed my comment that suggested prefixing the key with the name of the plugin, following this idea would eliminate the use of `red:hincr` thus your logic that depends on "`_run_cmd()` instantiating a new redis connection to run the command" will become incorrect. There will be more code changes. And I have left some comments regarding unification of the repeated logics. Please make these changes when you can and then I will review again.

1. The plugin name contains a hyphen; it is recommended not to use hyphens in Redis naming conventions. Directly using a plugin name with a hyphen is not recommended, as it contains a hyphen.
2. The source code of resty.rediscluster is written in a way that it calls the connect method of resty.redis every time do_cmd is invoked, establishing a connection. On the other hand, resty.redis simply establishes a connection. I'm not quite clear on the relationship you mentioned between "prefixing the key" and Redis connections.

theweakgod:
> @theweakgod I hope you noticed my comment that suggested prefixing the key with the name of the plugin, following this idea would eliminate the use of `red:hincr` thus your logic that depends on "`_run_cmd()` instantiating a new redis connection to run the command" will become incorrect. There will be more code changes. And I have left some comments regarding unification of the repeated logics. Please make these changes when you can and then I will review again.

My bad, I misunderstood earlier (it was a translation software issue).

Firstly, regardless of the command being invoked, rediscluster will establish a connection. Therefore, this is unrelated to the specific command being used. I hope you can notice [this part of the code](https://github.com/openresty/lua-resty-redis/blob/master/lib/resty/redis.lua#L468-L475) in resty.redis, which pre-processes the common_cmds and essentially calls do_cmd. So, when rediscluster invokes any Redis command, it is invoking resty.redis to establish a connection.

Are you suggesting changing it to incrby or zincrby? However, changing to these also does not address the issue of needing to establish a connection for both rediscluster and redis.

theweakgod:
@shreemaan-abhishek all done

shreemaan-abhishek:
> Are you suggesting changing it to incrby or zincrby? However, changing to these also does not address the issue of needing to establish a connection for both rediscluster and redis.

Okay I understand it now.

> Are you suggesting changing it to incrby or zincrby

Yes I am inclined towards using `incrby`. I don't think we need hashes here. Just prefix the key with plugin name to make it unique.

theweakgod:
> > Are you suggesting changing it to incrby or zincrby? However, changing to these also does not address the issue of needing to establish a connection for both rediscluster and redis.
> 
> Okay I understand it now.
> 
> > Are you suggesting changing it to incrby or zincrby
> 
> Yes I am inclined towards using `incrby`. I don't think we need hashes here. Just prefix the key with plugin name to make it unique.

In fact, I think redis key management is more difficult than hash key management. If a key needs to be created after a while, and the value is just an integer, then I think it will be a disaster for this db, because deleting these keys Sometimes, users need to unlink a lot of keys, and to delete the hash, they only need to unlink this key. This is my idea

theweakgod:
> I understand your point but using INCR would be more high performant isn't it? Hash will add one extra layer/lookup.

OK, I got it

theweakgod:
> I understand your point but using INCR would be more high performant isn't it? Hash will add one extra layer/lookup.

done

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,745 @@
...
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "plugins": {
+                            "limit-conn": {
+                                "conn": 100,
+                                "burst": 50,
+                                "default_conn_delay": 0.1,
+                                "rejected_code": 503,
+                                "key": "server_addr"
```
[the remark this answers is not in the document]
theweakgod: No, the policy here defaults to local, so redis_host is not necessary.

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,745 @@
...
+--- timeout: 10s
+--- response_body
+6
+4
+--- error_log
+limit key: 10.10.10.2route
+
+
+
+=== TEST 18: default rejected_code
```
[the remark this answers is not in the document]
theweakgod: no, it just set default rejected_code.

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,745 @@
...
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "plugins": {
+                            "limit-conn": {
+                                "burst": 1,
+                                "default_conn_delay": 0.1,
+                                "rejected_code": 503,
+                                "key": "remote_addr",
```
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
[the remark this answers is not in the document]
theweakgod: my bad, I misunderstood you before.

t/plugin/limit-conn-redis.t
```diff
@@ -675,8 +675,8 @@ passed
 GET /test_concurrency
 --- timeout: 10s
 --- response_body
-10
-0
+5
```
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,755 @@
...
+--- request
+GET /test_concurrency
+--- timeout: 10s
+--- response_body
+6
+4
+
+
+
+=== TEST 8: update plugin with username, password
```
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,755 @@
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
+
+
+
+=== TEST 7: exceeding the burst
```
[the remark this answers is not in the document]
theweakgod: done

t/plugin/limit-conn-redis.t
```diff
@@ -0,0 +1,745 @@
...
+            else
+                status_err_count = status_err_count + 1
+            end
+        end
+        ngx.say(status_ok_count)
+        ngx.say(status_err_count)
+    }
+}
+--- more_headers
+X-Real-IP: 10.0.0.1
```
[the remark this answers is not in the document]
theweakgod: sure

docs/en/latest/plugins/limit-conn.md
```diff
@@ -32,17 +32,30 @@ The `limit-conn` Plugin limits the number of concurrent requests to your service
...
+| allow_degradation        | boolean | False    | false       |                                   | When set to `true` enables Plugin degradation when the Plugin is temporarily unavailable and allows requests to continue.                                                                                                                                                                                                                                                                                 |
+| policy             | string  | False    | local | local, redis, redis-cluster | counter type to choose local, redis or redis-cluster                                                                                                                                                                                                                                                                                                                                                |
```
kayx23, line 46: the description could be a bit more descriptive for `policy` like [limit-count's](https://apisix.apache.org/docs/apisix/plugins/limit-count) doc?

docs/zh/latest/plugins/limit-conn.md
```diff
@@ -32,17 +32,30 @@ description: 本文介绍了 Apache APISIX limit-conn 插件的相关操作，
...
+| rejected_msg       | string | 否                                |        | 非空                       | 当请求数超过 `conn` + `burst` 阈值时，返回的信息。                                                                                                                                                                                    |
+| allow_degradation  | boolean | 否                                | false  |                            | 当设置为 `true` 时，启用插件降级并自动允许请求继续。                                                                                                                                                                                        |
+| policy             | string  | 否                                | local | local, redis, redis-cluster | 计数器类型                                                                                                                                                                                                                 |
```
kayx23, line 46: ditto

kayx23:
@shreemaan-abhishek I checked the doc and no more comments from me.

apisix/plugins/limit-conn/util.lua
```diff
@@ -0,0 +1,82 @@
...
+
+    -- we return a 0 delay by default
+    return 0, conn
+end
+
+
+
+function _M.leaving(self, red, key, req_latency)
+    assert(key)
+    local hash_key = "limit_conn"
```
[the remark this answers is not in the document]
theweakgod: done

apisix/utils/redis.lua
```diff
@@ -0,0 +1,75 @@
...
+        end
+    elseif err then
+        -- core.log.info(" err: ", err)
+        return nil, err
+    end
+    return red, nil
+end
+
+
+
```
monkeyDluffy6017, line 70: Only two blanks are needed between functions, please pay attention to other similar place

apisix/cli/ngx_tpl.lua
```diff
@@ -289,6 +289,7 @@ http {
 
     {% if enabled_plugins["limit-conn"] then %}
     lua_shared_dict plugin-limit-conn {* http.lua_shared_dict["plugin-limit-conn"] *};
+    lua_shared_dict plugin-limit-conn-redis-cluster-slot-lock {* http.lua_shared_dict["plugin-limit-conn-redis-cluster-slot-lock"] *};
```
monkeyDluffy6017, line 292: Why do we need to add a new share dict here?

apisix/utils/redis.lua
theweakgod: OK

apisix/utils/redis.lua
monkeyDluffy6017, line 1: Would it be better to keep these files in the `plugins` directory, since it will only be used by limit rate releated plugins?

apisix/cli/ngx_tpl.lua
theweakgod: Because resty.lock requires a shared dictionary, and rediscluster utilizes resty.lock.

apisix/utils/redis.lua
theweakgod: Yes, that's indeed a good idea. I will make the modification tomorrow.

apisix/plugins/utils/redis-schema.lua
monkeyDluffy6017, line 1: It's my mistake, i found that the [apisix.utils.log-util.lua](https://github.com/apache/apisix/blob/master/apisix/utils/log-util.lua) is referenced by many plugins too, could you revert this change and add a directory redis in the `apisix.utils` directory?

theweakgod: sure.

theweakgod: I have changed it, but found that the check failed.

monkeyDluffy6017: Maybe rename the directory will help?

theweakgod: I've decided to stick with the previous directory structure. I will revert the code because the check failed. What do u think?

monkeyDluffy6017: ok

monkeyDluffy6017: ok

theweakgod: done

theweakgod:
@shreemaan-abhishek @monkeyDluffy6017 Guys, do u think we can merge the code now?

monkeyDluffy6017 (approved):

PR #10889, opened 2024-01-30
chore: remove create-ssl.py

smileby:
### Description

use shell instead of python to configure ssls resources.

Remove the create-ssl.py script and modify the test_tls_over_tcp.sh test case

### Checklist


smileby:
@kayx23 @shreemaan-abhishek I don’t think the ci error is caused by this PR, please help check. Thanks

shreemaan-abhishek (approved):
Thank you.

starsz (approved):

PR #10905, opened 2024-02-02
feat(control): add plugins/reload to control api

sheharyaar:
### Description

`plugin/reload` is under admin API. The admin API can only be called at the control plane (if we set the role to control plane). But we need to call the plugin/reload API at the data plane, so if we set the role of the apisix to data plane we can't call this API.

This PR moves the `plugins/reload` to Control API so that we can call the endpoint from the data plane. For compatibility reasons, we also need to retain this API in admin API.

Refer: https://apisix.apache.org/docs/apisix/admin-api/#plugin

### Checklist


apisix/control/v1.lua
```diff
@@ -400,6 +400,13 @@ function _M.dump_plugin_metadata()
     return 200, metadata.value
 end
 
+function _M.plugins_reload()
```
starsz, line 403: Missing the logic of sending events to the other worker processes.
You can refer: https://github.com/apache/apisix/blob/master/apisix/admin/init.lua#L284

monkeyDluffy6017:
You mean `move`, do we need to remove the old API?

sheharyaar:
@monkeyDluffy6017  @starsz please review

t/control/plugins-reload.t
```diff
@@ -0,0 +1,52 @@
...
+            local code, body, res = t.test('/v1/plugins_reload',
+                ngx.HTTP_POST)
+            ngx.say(res)
+        }
+    }
+--- request
+GET /t
+--- error_code: 200
+--- response_body
+done
```
starsz, line 52: Can you add more test cases to test it ? 
You can refer https://github.com/apache/apisix/blob/master/t/admin/plugins-reload.t

monkeyDluffy6017:
Do we need to remove the old API?

monkeyDluffy6017:
Can you describe in detail why do you submit this pr?

apisix/control/router.lua
```diff
@@ -46,6 +47,80 @@ local function format_dismod_uri(mod_name, uri)
...
+
+    local new_set = {}
+    for _, p in ipairs(new) do
+        new_set[p.name] = p
+    end
+
+    return core.table.set_eq(old_set, new_set)
+end
+
+local function sync_local_conf_to_etcd(reset)
```
starsz, line 64: The code is the same as it is under admin API. Why not put it into the utils directory?

docs/en/latest/control-api.md
```diff
@@ -475,3 +475,9 @@ Dumps the metadata with the specified `plugin_name`:
     "id": "file-logger"
 }
 ```
+
+### POST /v1/plugins_reload
```
starsz, line 479: The method is `PUT` right?

docs/en/latest/control-api.md
```diff
@@ -475,3 +475,9 @@ Dumps the metadata with the specified `plugin_name`:
...
+
+Introduced in [v3.9.0](https://github.com/apache/apisix/releases/tag/3.9.0)
+
+Triggers a hot reload of the plugins.
```
starsz, line 483: Can you give an example of how to use this API?

apisix/control/v1.lua
```diff
@@ -474,5 +485,12 @@ return {
         uris = {"/plugin_metadata/*"},
         handler = _M.dump_plugin_metadata,
     },
+    -- /v1/plugins_reload
+    {
+        methods = {"PUT"},
+        uris = {"/plugins_reload"},
```
starsz, line 491:
```suggestion
        uris = {"/plugins/reload"},
```
Would be better.

starsz (approved):

starsz (approved):

apisix/control/router.lua
```diff
@@ -198,5 +201,26 @@ end
...
+
+    local local_conf = core.config.local_conf()
+    local deployment_role = core.table.try_read_attr(
+                       local_conf, "deployment", "role")
+    if deployment_role ~= "data_plane" then
+        -- data_plane should not write to etcd
+        if ngx_worker_id() == 0 then
+            sync_local_conf_to_etcd()
+        end
+    end
```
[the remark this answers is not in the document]
monkeyDluffy6017: @starsz please help to check this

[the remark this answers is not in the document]
starsz: The logic is the same as it in the admin API. 
https://github.com/apache/apisix/blob/master/apisix/admin/init.lua#L469-L475

apisix/utils/config_etcd.lua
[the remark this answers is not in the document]
monkeyDluffy6017: @sheharyaar

apisix/control/router.lua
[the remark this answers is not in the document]
starsz: After thinking about it, I think in the control API we don't need to update the `/plugins` resource in the etcd even if the role is `traditional`.

apisix/control/router.lua
```diff
@@ -198,5 +197,4 @@ end
 
 end -- do
 
-
```
starsz, line 201: Don't need to delete this line.

apisix/control/v1.lua
```diff
@@ -400,6 +403,12 @@ function _M.dump_plugin_metadata()
     return 200, metadata.value
 end
 
+function _M.reload_plugins()
+    core.log.info("start to hot reload plugins")
+    plugin.load()
```
starsz, line 408: You should keep the logic of event.

starsz (approved):

monkeyDluffy6017 (approved):

PR #10883, opened 2024-01-30
docs: jwe-decrypt secret length must be 32 chars

Vacant2333:
### Description

Fixes # [(issue 10825)](https://github.com/apache/apisix/issues/10825)

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #10928, opened 2024-02-08
fix: jwe-decrypt secret length restriction

Vacant2333:
### Description
add restriction for `jwe-decrypt` plugin
the secret length should be 32 chars only, it will be checked on the `check_scheme`

Fixes #
https://github.com/apache/apisix/pull/10883#issuecomment-1931380021

### Checklist

### DIFF
--- a/docs/en/latest/plugins/jwe-decrypt.md
+++ b/docs/en/latest/plugins/jwe-decrypt.md
@@ -44,6 +44,12 @@ For Consumer:
 | secret        | string  | True                                                 |         |                             | The decryption key. Must be 32 characters. The key could be saved in a secret manager using the [Secret](../terminology/secret.md) resource. |
 | is_base64_encoded | boolean | False                                                 | false   |                             | Set to true if the secret is base64 encoded.                                                                                                 |
 
+:::note
+
+After enabling `is_base64_encoded`, your `secret` length may exceed 32 chars. You only need to make sure that the length after Decode is still 32 chars.
+
+:::
+
 For Route:
 
 | Name   | Type   | Required | Default       | Description                                                         |
--- a/docs/en/latest/plugins/jwe-decrypt.md
+++ b/docs/en/latest/plugins/jwe-decrypt.md
@@ -46,7 +46,7 @@ For Consumer:
 
 :::note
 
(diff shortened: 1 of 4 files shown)


Vacant2333:
the previos ci was wrong, i will fix it on here https://github.com/apache/apisix/pull/10928/commits/7052e25ee8e64bec1f2a0a3dbe38ac3904604877
the test 2 and 3 test for the value type of `key` and `secert`,

and i will remove the schema min length, because the length of `secret` will be checked on `check_schema`

Vacant2333:
@kayx23 @shreemaan-abhishek can u help me take a look~~

docs/en/latest/plugins/jwe-decrypt.md
```diff
@@ -44,6 +44,12 @@ For Consumer:
 | secret        | string  | True                                                 |         |                             | The decryption key. Must be 32 characters. The key could be saved in a secret manager using the [Secret](../terminology/secret.md) resource. |
 | is_base64_encoded | boolean | False                                                 | false   |                             | Set to true if the secret is base64 encoded.                                                                                                 |
 
+:::note
+
+After enabling `is_base64_encoded`, your `secret` length may exceed 32 chars. You only need to make sure that the length after Decode is still 32 chars.
```
kayx23, line 49:
```suggestion
After enabling `is_base64_encoded`, your `secret` length may exceed 32 chars. You only need to make sure that the length after decoding is still 32 chars.
```

tbh i think this info should go into the description of `secret`, rather than a note at the bottom (if its not too difficult to read)

Vacant2333: if we add this description to `secret` it will be too long i think

t/plugin/jwe-decrypt.t
```diff
@@ -74,13 +74,13 @@ done
 
 
 
-=== TEST 3: wrong type of string
+=== TEST 3: wrong type of secret
```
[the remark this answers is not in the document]
Vacant2333: hi, i have explan this at here https://github.com/apache/apisix/pull/10928#issuecomment-1933498159, the test name was wrong, Test 3&4 was test for the wrong type of `secret` & `key`

kayx23 (approved):

Issue #9269, opened 2023-04-07
bug: datadog plugin failed to send metric when consumer is used with route for authentication


0prodigy:
### Current Behavior

datadog plugin failed to send metric

### Expected Behavior

plugin should send metric to datadog agent

### Error Logs

lua entry thread aborted: runtime error: /usr/local/apisix/apisix/plugins/datadog.lua:92: attempt to concatenate field 'consumer' (a table value)

### Steps to Reproduce

1. Enable datadog plugin in routes
2. Check log 

### Environment

- APISIX version (run apisix version): 3.0.0
- Operating system (run uname -a): Linux (x86_64 GNU/Linux)
- OpenResty / Nginx version (run openresty -V or nginx -V): openresty/1.21.4.1
- etcd version, if relevant (run curl http://127.0.0.1:9090/v1/server_info):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run luarocks --version):

Sn0rt:
can you share your config of the router ?

shreemaan-abhishek:
@0prodigy please share steps to reproduce, I have configured the dd plugin before and it worked.

monkeyDluffy6017:
I will close this due to long period of inactivity, please feel free to reopen it

zhuoyang:
I am facing this issue as well, it only happens when there's authentication enabled on the route

kayx23:
I was able to reproduce the issue with the following steps:

1. configure datadog on a route
2. configure a consumer with `key-auth`
3. patch the route with `key-auth`
4. send a request to the route and check logs

![image](https://github.com/apache/apisix/assets/39619599/4d75ef97-6b60-471b-aaed-aed319036bf8)

cc: @Vacant2333

shreemaan-abhishek:
looks like a good first issue to me

Vacant2333:
> I was able to reproduce the issue with the following steps:
> 
> 1. configure datadog on a route
> 2. configure a consumer with `key-auth`
> 3. patch the route with `key-auth`
> 4. send a request to the route and check logs
> 
[...]

PR #10934, opened 2024-02-14
fix(datadog): use the username property instead of the whole table when generating the tag

ecoupal-believe:
### Description

Use the username property instead of the whole table when generating the tag.

Fixes https://github.com/apache/apisix/issues/9269

### Checklist


shreemaan-abhishek:
@ecoupal-believe please add test cases as well.

Issue #6403, opened 2022-02-21
feat: apisix support QUIC+HTTP3


lizj3624:
### Issue description

[cloudflare/quiche](https://github.com/cloudflare/quiche) patch nginx-1.16 support quic+http3, I hope 
apisix support QUIC+HTTP3

moonming:
great idea! I'm not familiar with quiche, how much would it cost to use it instead of openresty? In addition, nginx is officially merging the functions of QUIC+HTTP3?
cc @membphis

lizj3624:
I patch [quiche](https://github.com/cloudflare/quiche) to OpenResty-1.15.8，I found error (epoll_ctl(1, 53) failed (17: File exists)) when access_by_lua of ngx_lua module enabled，I gruess ngx_lua not support quic.  [ngx_lua issue](https://github.com/openresty/lua-nginx-module/issues/1887)

1. build env
```shell
nginx version: openresty/1.15.8.1 (quiche-09c0e37)
built by gcc 4.8.5 20150623 (Red Hat 4.8.5-36) (GCC)
built with OpenSSL 1.1.1 (compatible; BoringSSL) (running with BoringSSL)
TLS SNI support enabled
[...]

PR #10989, opened 2024-03-04
feat: support to enable quic and add basic test case

zll600:
### Description

Fixes https://github.com/apache/apisix/issues/6403

support enable http3 for apisix.

I will add a how-to guide in another PR.

### Checklist


ci/linux_install_static_curl.sh
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+# use curl which supports http3 to test.
+install_static_curl
```
zll600, line 22: curl from apt does not support option --http3-only to test. So we need a newer version of curl

t/APISIX.pm
```diff
@@ -777,7 +779,7 @@ _EOC_
             set \$upstream_connection         '';
 
             set \$upstream_scheme             'http';
-            set \$upstream_host               \$http_host;
+            set \$upstream_host               \$host;
```
zll600, line 782: The value of variable $http_host header will be empty string when using http/3
ref: https://github.com/zulip/zulip/pull/26348

apisix/cli/ngx_tpl.lua
```diff
@@ -719,7 +722,7 @@ http {
             set $upstream_connection         '';
 
             set $upstream_scheme             'http';
-            set $upstream_host               $http_host;
+            set $upstream_host               $host;
```
zll600, line 725: The value of variable $http_host header will be empty string when using http/3
ref: zulip/zulip#26348

conf/config-default.yaml
```diff
@@ -97,9 +97,11 @@ apisix:
     listen:                                       # APISIX listening port for HTTPS traffic.
       - port: 9443
         enable_http2: true
+        enable_quic: true
```
membphis, line 100: I don't think we should use `true` as default
`false` is better

t/node/grpc-proxy.t
```diff
@@ -219,7 +219,7 @@ grpcurl -import-path ./t/grpc_server_example/proto -proto helloworld.proto -plai
 --- grep_error_log eval
 qr/grpc header: "(:authority|host): [^"]+"/
 --- grep_error_log_out eval
-qr/grpc header: "(:authority|host): 127.0.0.1:1984"/
+qr/grpc header: "(:authority|host): 127.0.0.1"/
```
membphis, line 222: I don't think we need to update this line

t/node/http_host.t
```diff
@@ -64,5 +64,5 @@ GET /uri
 Host: foo.com:1984
 --- response_body
 uri: /uri
-host: foo.com:1984
+host: foo.com
```
membphis, line 67: same issue, I do not know why we need to update this test case

t/quic/admin/basic.t
```diff
@@ -0,0 +1,108 @@
...
+        local code, body = t('/apisix/admin/routes/1',
+            ngx.HTTP_PUT,
+            [[{
+                "upstream": {
+                    "nodes": {
+                        "127.0.0.1:1980": 1
+                    },
+                    "type": "roundrobin"
+                },
+                "uris": ["/hello", "/world"]
```
membphis, line 86: only `/hello` should be enough

t/node/http_host.t
zll600: I change `$http_host` to `$host`, ref https://github.com/apache/apisix/pull/10989/files#r1510533390
`$host` has no `port`. So I will lost the port info here.

t/quic/admin/basic.t
zll600: OK, I will change it.

apisix/init.lua
```diff
@@ -576,6 +576,13 @@ end
 
 
 function _M.http_access_phase()
+    if ngx.req.http_version() == 3 then
+        local upstream_host = ngx.var.host
+        if ngx.var.server_port then
+            upstream_host = upstream_host .. ":" .. ngx.var.server_port
+        end
+        ngx.var.upstream_host = upstream_host
+    end
```
zll600, line 585: For http3, the `http_post` variable will be empty which is used to [set `$upstream_host`](https://github.com/apache/apisix/blob/d6a3db3ee2738436a634c3a6addd02c9cdfca035/apisix/cli/ngx_tpl.lua#L722). But because http/1.1 is used between apisix and upstream, `upstream_host` needs to be set manually here.

apisix/cli/ops.lua
```diff
@@ -449,7 +460,7 @@ Please modify "admin_key" in conf/config.yaml .
                 end
 
                 listen_table_insert(node_listen, "http", ip, port,
-                        enable_http2, enable_ipv6)
+                        enable_http2, false, enable_ipv6)
```
monkeyDluffy6017, line 463: The `enable_quic` option is fixed?

ci/common.sh
```diff
@@ -177,7 +186,7 @@ GRPC_SERVER_EXAMPLE_VER=20210819
 
 linux_get_dependencies () {
     apt update
-    apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev
+    apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev xz-utils
```
monkeyDluffy6017, line 189: Why do we need `xz-utils`?

zll600: I need it to install `static-url` from the `.tar.xz` file. ref: https://github.com/apache/apisix/pull/10989/files#diff-67362488684645b13d1314d9d45bdfd21fed3599970720bccd584a079971dff9R80

.github/workflows/quic.yml
```diff
@@ -0,0 +1,191 @@
...
+        run: |
+          echo "version=${GITHUB_REF##*/}" >>$GITHUB_OUTPUT
+          echo "fullname=apache-apisix-${GITHUB_REF##*/}-src.tgz" >>$GITHUB_OUTPUT
+
+      - name: Extract test type
+        shell: bash
+        id: test_env
+        run: |
+          test_dir="${{ matrix.test_dir }}"
+          if [[ $test_dir =~ 't/quic/plugin' ]]; then
```
shreemaan-abhishek, line 75: why do we need this if we are just testing basic.t?

apisix/cli/ops.lua
```diff
@@ -407,7 +413,12 @@ Please modify "admin_key" in conf/config.yaml .
 
             if ip_port_to_check[addr] == nil then
                 table_insert(listen_table,
-                        {ip = ip, port = port, enable_http2 = enable_http2})
+                        {
```
shreemaan-abhishek, line 416: can you explain the usage of this block briefly? And why this change is needed.

ci/common.sh
```diff
@@ -75,6 +75,15 @@ install_curl () {
     curl -V
 }
 
+install_static_curl () {
```
shreemaan-abhishek, line 78: if this is not going to be used anywhere else, better move it in quic.yaml.

ci/common.sh
shreemaan-abhishek: I hope you checked if there is any other alternative way to install static-url? Or install other tool? Or avoid installing it altogether.

apisix/cli/ops.lua
zll600: because I want to add a new field `enable_quic`. If I add `enable_quic` directly, the length of this line will too long to cause a lint error.

ci/common.sh
zll600: OK, will do

ci/common.sh
zll600: I will install xz-utils in quic.yml to avoid installing `xz-utils` altogether.

.github/workflows/quic.yml
zll600: I will add `t/quic/plugin` in another PR after merging this PR. So I didn't adjust code here which was copied from build.yml.

conf/config-default.yaml
```diff
@@ -97,9 +97,11 @@ apisix:
     listen:                                       # APISIX listening port for HTTPS traffic.
       - port: 9443
         enable_http2: true
+        enable_quic: false
```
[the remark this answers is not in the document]
zll600: `enable_quic` is already under `ssl`. I will add some comments for `enable_quic`.

apisix/init.lua
zll600: It will lost the `port` field by only using `ngx.var.http_host`.
Now Could we keep `ngx.var.host + ngx.var.server_port`?

t/quic/admin/basic.t
```diff
@@ -0,0 +1,108 @@
...
+--- request
+GET /t
+--- response_body
+passed
+
+
+
+=== TEST 4: Successfully access route with QUIC
+--- exec
+curl -k -v -H ":authority: test.com" -H "content-length: 0" --http3-only --resolve "test.com:1994:127.0.0.1" https://test.com:1994/hello 2>&1 | cat
```
membphis, line 106: this should be wrong: ":authority: test.com", pls fix it

.github/workflows/quic.yml
shreemaan-abhishek: ok

ci/common.sh
```diff
@@ -177,7 +177,7 @@ GRPC_SERVER_EXAMPLE_VER=20210819
 
 linux_get_dependencies () {
     apt update
-    apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev
+    apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev xz-utils
```
shreemaan-abhishek, line 180: you are already installing xz-utils in quic.yml. Is this needed again?

zll600: Sorry, I will revert this change

membphis (commented):
LGTM

need  a better title, eg: `feat: support to enable quic and add basic test case`

monkeyDluffy6017 (approved):

membphis (approved):
LGTM

Issue #11020, opened 2024-03-09
bug: forward-auth plugin doesn't work properly with POST method. 


suryaprabhakark:
### Current Behavior

forward-auth plugin doesn't work if the request is POST and authservice api is GET. Its happening due to forwading the POST method request headers like content-type and expect. Moved these headers only if authserver api is also POST.

### Expected Behavior

It should work without any problems, based on the authserver method, appropriate headers should be passed.

### Error Logs

`2024/03/09 15:35:28 [error] 51#51: *13215177 [lua] forward-auth.lua:134: phase_func(): failed to process forward auth, err: timeout, client: 10.117.4.40, server: _, request: "POST /post HTTP/1.1", host: "apiv2.ninjacart.in", referrer: "http://localhost:4000/"
2024/03/09 15:35:28 [warn] 51#51: *13215177 [lua] plugin.lua:1159: run_plugin(): forward-auth exits with http status code 403, client: 10.117.4.40, server: _, request: "POST /post HTTP/1.1", host: "apiv2.ninjacart.in", referrer: "http://localhost:4000/"
10.117.4.40 - - [09/Mar/2024:15:35:28 +0000] apiv2.ninjacart.in "POST /post HTTP/1.1" 403 225 3.043 "http://localhost:4000/" "Mozilla/5.0 (iPhone; CPU iPhone OS 13_2_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Mobile/15E148 Safari/604.1" - - - "http://apiv2.ninjacart.in"`

![Screenshot 2024-03-09 at 21 05 52](https://github.com/apache/apisix/assets/3763515/976419dd-7307-4ea0-bbf8-728691ae55d8)

### Steps to Reproduce

1. Run the apisix with docker image
2. Configure a route with post method, i used httporg post api like below
```
routes:
  - 
      uri: /post
      upstream:
        nodes:
          "httpbin.org:80": 1
        type: roundrobin
      plugins:
        forward-auth:
          _meta:
            disable: false
          request_headers:
            - Authorization
          ssl_verify: false
          request_method: GET
          uri: >-
            https://<AUTHSERVER_URL>
```
4.  Use the curl like below 
`shell
curl --location 'https://localhost:9080/post' \
--header 'Accept: application/json, text/plain, */*' \
--header 'Authorization: Bearer <TOKEN>' \
--header 'Content-Type: application/json' \
--data '{
    "payload" {
[...]

Issue #10927, opened 2024-02-07
bug: forward-auth with request_method GET causes timeouts


BrandonArp:
### Current Behavior

The request to the forward-auth request contains a Content-Length header and the GET request has no body.

### Expected Behavior

The forward-auth request has no Content-Length header.

### Error Logs

```
apisix-1  | 2024/02/07 20:27:39 [error] 242#242: *648462 [lua] forward-auth.lua:134: phase_func(): failed to process forward auth, err: timeout, client: 172.27.0.1, server: _, request: "POST /test HTTP/1.1", host: "localhost:9080"
apisix-1  | 2024/02/07 20:27:39 [warn] 242#242: *648462 [lua] plugin.lua:1159: run_plugin(): forward-auth exits with http status code 403, client: 172.27.0.1, server: _, request: "POST /test HTTP/1.1", host: "localhost:9080"
```

### Steps to Reproduce

1. Run APISIX
2. Deploy a service that can consume forward-auth requests
3. Create a route that has forward-auth enabled, with a GET request_method
    You can use the following plugin config:
   ```yaml
   forward-auth:
        keepalive: true
        keepalive_pool: 100
        keepalive_timeout: 2000
        request_headers:
          - Authorization
        request_method: GET
        uri: http://some_auth_service/auth```
4. Send 2 requests to the protected route
5. Observe that the second request will fail with a timeout

### Environment

- APISIX version 3.8.0

BrandonArp:
I believe that the issue was introduced in #10589 where some headers were added unconditionally [here](https://github.com/apache/apisix/pull/10589/files#diff-4b85e7666119852676bcd1c5fd15889e752f061a9ae891e10a0cbbefd30c2e0dR91-R93).

moonming:
welcome PR to fix it @BrandonArp

pshettyk:
I am also facing the same issue in 3.8.0. Works fine in 3.6.0.

moonming:
@shreemaan-abhishek please take a look

shreemaan-abhishek:
I have found the cause of this bug. I will fix this later.

### Case 1:
- request contains payload
- original request method: GET
- fw auth request method: GET

[...]

PR #11021, opened 2024-03-09
fix: add post request headers only if auth request method is post

suryaprabhakark:
### Description
forward-auth plugin doesn't work if the request is POST and authservice api is GET. Its happening due to forwading the POST method request headers like content-type and expect. Moved these headers only if authserver api is also POST.

Fixes # (issue)
Fixes the issue with forward-auth plugin with POST headers. 
https://github.com/apache/apisix/issues/10927
https://github.com/apache/apisix/issues/11020

### Checklist

### DIFF
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -89,10 +89,14 @@ function _M.access(conf, ctx)
         ["X-Forwarded-Host"] = core.request.get_host(ctx),
         ["X-Forwarded-Uri"] = ctx.var.request_uri,
         ["X-Forwarded-For"] = core.request.get_remote_client_ip(ctx),
-        ["Expect"] = core.request.header(ctx, "expect"),
-        ["Content-Length"] = core.request.header(ctx, "content-length"),
-        ["Transfer-Encoding"] = core.request.header(ctx, "transfer-encoding")
     }
+
+    if conf.request_method == "POST" then
+        auth_headers["Content-Length"] = core.request.header(ctx, "content-length")
+        auth_headers["Expect"] = core.request.header(ctx, "expect")
+        auth_headers["Transfer-Encoding"] = core.request.header(ctx, "transfer-encoding")
+        auth_headers["Content-Encoding"] = core.request.header(ctx, "content-encoding")
+    end
 
     -- append headers that need to be get from the client request header
     if #conf.request_headers > 0 then
--- a/t/plugin/forward-auth.t
+++ b/t/plugin/forward-auth.t
@@ -143,6 +143,24 @@ property "request_method" validation failed: matches none of the enum values
                                                end
                                            end
                                         end
+                                    end]],
+                                    [[return function(conf, ctx)
+                                        local core = require("apisix.core");
+                                        if core.request.header(ctx, "Authorization") == "token-headers-test" then
+                                            if core.request.get_method() == "POST" then
+                                                if core.request.header(ctx, "Content-Length") or core.request.header(ctx, "Transfer-Encoding") or core.request.header(ctx, "Content-Encoding") then
+                                                    core.response.exit(200)
+                                                else
+                                                    core.response.exit(403)
+                                                end
+                                            else
+                                                if core.request.header(ctx, "Content-Length") or core.request.header(ctx, "Transfer-Encoding") or core.request.header(ctx, "Content-Encoding") then
+                                                    core.response.exit(403)
+                                                else
+                                                    core.response.exit(200)
+                                                end
+                                            end
+                                        end
                                     end]]
                                 }
                             }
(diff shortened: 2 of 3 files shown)


BrandonArp:
Should the Transfer-Encoding header be removed? If APISIX is trying to just pass through the bytes of the request body, we'll need to send the Transfer-Encoding header to make sure the auth server knows how to interpret the bytes.

shreemaan-abhishek:
Hi could you also handle this case? 
image
https://github.com/apache/apisix/issues/10927#issuecomment-1975557939

and add test cases?

suryaprabhakark:
> Should the Transfer-Encoding header be removed? If APISIX is trying to just pass through the bytes of the request body, we'll need to send the Transfer-Encoding header to make sure the auth server knows how to interpret the bytes.

@BrandonArp My initial reading about the transfer encoding was that its set in the responses rather than request, further research suggests it can be used in both request and responses as well. Ill add back the header. Do you think we need to add content-encoding header as well, authserver should also know if the content is encoded.

motongxue:
> Hi could you also handle this case? image [#10927 (comment)](https://github.com/apache/apisix/issues/10927#issuecomment-1975557939)
> 
> and add test cases?

Hello abhishek. I have submitted a [PR](https://github.com/apache/apisix/pull/11023) regarding your issue, and I hope it can be of help to you. At the same time, I would greatly appreciate your assistance in reviewing and finalizing the test cases.

BrandonArp:
> Do you think we need to add content-encoding header as well, authserver should also know if the content is encoded.

Content-Encoding should be added at some point. I'm not very familiar with the APISIX architecture, so I don't know if it would be added somewhere else. If this is the only place we're adding headers, then yes, it should be added. Otherwise the auth server won't know how to interpret the body.

suryaprabhakark:
Thanks @BrandonArp. Let me know one of the contributor. 

@shreemaan-abhishek Can you help with the content-encoding header, I think its required by the authserver if its present in the request.

shreemaan-abhishek:
the content-encoding header should also be passed. Also, test cases should be added to support this PR.

suryaprabhakark:
Apologies for the delay, will add the testcases over the weekend(24th March).

suryaprabhakark:
@shreemaan-abhishek @BrandonArp Added the testcases, please review

t/plugin/forward-auth.t
```diff
@@ -143,6 +143,24 @@ property "request_method" validation failed: matches none of the enum values
                                                end
                                            end
                                         end
+                                    end]],
+                                    [[return function(conf, ctx)
```
shreemaan-abhishek, line 147: please write the test cases in a new file `t/plugin/forward-auth2.t` . This file already contains too many serverless functions and routes which can cause conflict.

suryaprabhakark: @shreemaan-abhishek Had split the testcases into new file.

suryaprabhakark:
@shreemaan-abhishek @BrandonArp Can you please give approval for workflows to run.

apisix/plugins/forward-auth.lua
```diff
@@ -89,11 +89,15 @@ function _M.access(conf, ctx)
         ["X-Forwarded-Host"] = core.request.get_host(ctx),
         ["X-Forwarded-Uri"] = ctx.var.request_uri,
         ["X-Forwarded-For"] = core.request.get_remote_client_ip(ctx),
-        ["Expect"] = core.request.header(ctx, "expect"),
-        ["Content-Length"] = core.request.header(ctx, "content-length"),
-        ["Transfer-Encoding"] = core.request.header(ctx, "transfer-encoding")
     }
 
+    if conf.request_method == "POST" then
+        auth_headers["Content-Length"] = core.request.header(ctx, "content-length")
```
membphis, line 95: For HTTP requests, if the request header contains the ["Content-Length"] field, the request body should be included.

So I think this request header should not be included.

suryaprabhakark:
@BrandonArp @shreemaan-abhishek Got these following errros in the tests, could you please help. 

`Error: nginx: [error] open() "/home/runner/work/apisix/apisix/logs/nginx.pid" failed (2: No such file or directory)
Error: 2024/03/24 06:39:06 [error] 94305#94305: *155 stream [lua] radixtree_sni.lua:178: match_and_set(): failed to find any SSL certificate by SNI: test.com, context: ssl_certificate_by_lua*, client: 127.0.0.1, server: 0.0.0.0:9100
`

apisix/plugins/forward-auth.lua
suryaprabhakark: @membphis If the auth server `request_method` is POST, the same upstream body is being forwarded to auth server as well, thats why we need this header incase its present. Below code reference is where body is being added to auth api request. 

https://github.com/apache/apisix/pull/11021/files/48112cc9057b16611ba86d0761a84dd635ac0e8b#diff-4b85e7666119852676bcd1c5fd15889e752f061a9ae891e10a0cbbefd30c2e0dL115-L124

shreemaan-abhishek:
@suryaprabhakark sorry for the delayed response, the CI failure was very likely a flaky error I have re-ran the CI.

shreemaan-abhishek (approved):

starsz (approved):

PR #11023, opened 2024-03-11
fix: Adjust POST headers and review Transfer-Encoding handling in forward-auth

motongxue:
…ward-auth

### Description
This is regarding the problem mentioned in Case 2 of [issue #10927](https://github.com/apache/apisix/issues/10927). I addressed the issue by retrieving the maximum body size using ngx.var.client_max_body_size and then comparing it with the Content-Length.

As a newcomer to APISIX, I have drafted the logic for the test case as I think it should be, but I would appreciate assistance in reviewing the correctness of the test case writing.

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/schema_def.lua
+++ b/apisix/schema_def.lua
@@ -309,6 +309,7 @@ local nodes_schema = {
                         description = "port of node",
                         type = "integer",
                         minimum = 1,
+                        maximum = 65535
                     },
                     weight = {
                         description = "weight of node",
@@ -342,6 +343,7 @@ _M.discovery_nodes = {
                 description = "port of node",
                 type = "integer",
                 minimum = 1,
+                maximum = 65535
             },
             weight = {
                 description = "weight of node",
@@ -901,6 +903,8 @@ _M.stream_route = {
         server_port = {
             description = "server port",
             type = "integer",
+            minimum = 1,
+            maximum = 65535
         },
         sni = {
             description = "server name indication",
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -451,6 +451,18 @@ local function check_upstream_conf(in_dp, conf)
             return false, "invalid configuration: " .. err
         end
 
+        if conf.nodes and not core.table.isarray(conf.nodes) then
+            local port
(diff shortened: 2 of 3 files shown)


shreemaan-abhishek:
@motongxue

1. the CI is failing.
2. I think it's okay to use `client_body_reader` by default as there are no negative implications. So just ensure to cache the req_body in a separate variable. Later, restore the request body from the variable.

t/plugin/forward-auth.t
```diff
@@ -488,3 +488,127 @@ Authorization: 333
         }
     }
 --- error_code: 200
+
```
monkeyDluffy6017, line 491: Three blanks are needed between test cases

motongxue: > Three blanks are needed between test cases

Ok, I will fix that.

motongxue:
> @motongxue
> 
> 1. the CI is failing.
> 2. I think it's okay to use `client_body_reader` by default as there are no negative implications. So just ensure to cache the req_body in a separate variable. Later, restore the request body from the variable.



> @motongxue
> 
> 1. the CI is failing.
> 2. I think it's okay to use `client_body_reader` by default as there are no negative implications. So just ensure to cache the req_body in a separate variable. Later, restore the request body from the variable.

@shreemaan-abhishek 
Hello abhishek, the CI has failed because the test cases have not been fixed yet.

Using client_body_reader by default is a more efficient method for handling large request bodies, as it prevents the entire large request body from being loaded into memory all at once. However, if the majority of the request bodies are small, frequently using streaming read can lead to unnecessary performance overheads, because each read operation requires time and resources, even if the data volume is small.

Therefore, a more balanced approach might be to select the appropriate reading strategy based on the anticipated size and usage of the request body. For example, a threshold could be set: use core.request.get_body() for request bodies smaller than this threshold, and use httpc:get_client_body_reader() for request bodies larger than the threshold.

shreemaan-abhishek:
@motongxue okay, it makes sense to me now.

shreemaan-abhishek:
@motongxue test cases are failing.

Issue #11046, opened 2024-03-15
bug: Disable All Prometheus-related Features When Prometheus Plugin is Turned Off


motongxue:
### Current Behavior

The Prometheus plugin is a crucial component for observability within Apache APISIX. However, I've encountered an issue where, even after the Prometheus plugin is disabled, all features related to Prometheus are not entirely shut down. This behavior could lead to unexpected performance impacts and data leakage.

Upon disabling the Prometheus plugin, certain Prometheus-related functionalities continue to operate, suggesting that not all components associated with the plugin are fully turned off.

### Expected Behavior

When the Prometheus plugin is disabled, all functionalities and metrics associated with Prometheus should be completely deactivated to ensure no unnecessary overhead or data exposure.

### Steps to Reproduce

1. Enable the Prometheus plugin in Apache APISIX.
2. Generate some traffic and observe Prometheus metrics being collected.
3. Disable the Prometheus plugin.
4. Observe that metrics or features associated with Prometheus are still active or being collected.

### Environment

- APISIX version (run `apisix version`): 3.8.0
- Operating system (run `uname -a`): Linux 4f398f8410bc 6.5.0-21-generic #21~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb  9 13:32:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.25.3.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

shreemaan-abhishek:
To be fixed by #11047. Thanks for your report.

PR #11047, opened 2024-03-15
fix(prometheus): disable features when plugin is turned off

motongxue:
### Description

Fixes # ([issue](https://github.com/apache/apisix/issues/11046))

### Checklist


apisix/utils/batch-processor.lua
```diff
@@ -170,21 +173,27 @@ function batch_processor:new(func, config)
...
         server_addr = config.server_addr,
+        prometheus_enabled = prometheus_enabled,
     }
 
     return setmetatable(processor, batch_processor_mt)
 end
 
 
 function batch_processor:push(entry)
+
```
monkeyDluffy6017, line 184: Remove this blank

apisix/cli/ngx_tpl.lua
```diff
@@ -114,6 +114,27 @@ http {
             stub_status;
         }
     }
+
+    {% else %}
+
+    init_by_lua_block {
```
monkeyDluffy6017, line 120: 1. this block is for `stream` only, please take care of the `http`
2. init_by_lua_block should be kept

apisix/utils/batch-processor.lua
motongxue: Okay, I have deleted this blank line.

apisix/cli/ngx_tpl.lua
motongxue: > 1. this block is for `stream` only, please take care of the `http`
> 2. init_by_lua_block should be kept

I have made a corresponding modification based on your two modification opinions

Revolyssup:
@motongxue Please fix the lint checks.

monkeyDluffy6017:
Could you add test cases to cover this?

t/xrpc/prometheus2.t
```diff
@@ -0,0 +1,127 @@
...
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 5: fetch the prometheus metric data while prometheus plugin is disabled
```
Revolyssup, line 121: Can you not add this case in the existing prometheus.t instead of creating a new test file?

motongxue: Thank you for your suggestion. I understand that consolidating related tests into a single file can enhance the readability and maintainability of the code. However, I have decided to create a new test file for the following reasons:
1. Currently, I have not passed all the tests in the existing Prometheus.t file. I am working on resolving these issues, but I am not very familiar with the tests yet, and I want to avoid affecting the existing related tests.
2. After consideration, I realized that the new test cases differ significantly from those in the existing Prometheus.t, as they test different functionality. I am concerned that merging them into the same file would make the file overly complex and could reduce the clarity and readability of the code.

Based on the above reasons, I believe creating a new test file is reasonable.

Revolyssup: @motongxue Okay makes sense. I see that you're saying you need to make changes to the test. Initially I saw only one extra test being added and rest of the tests same as prometheus.t. Please inform after you make final changes to the tests and ping if you need any help.

motongxue: @Revolyssup I think this is my final changes. Waiting for being merged.

t/xrpc/prometheus2.t
```diff
@@ -0,0 +1,127 @@
...
+run_tests;
+
+__DATA__
+
+=== TEST 1: sanity
+--- config
+    location /t {
+        content_by_lua_block {
+            local plugin = require("apisix.plugins.prometheus")
+            local ok, err = plugin.check_schema({})
```
monkeyDluffy6017, line 51: Why do you add this test case?

t/xrpc/prometheus2.t
```diff
@@ -0,0 +1,127 @@
...
+
+
+
+=== TEST 5: fetch the prometheus metric data while prometheus plugin is disabled
+--- yaml_config
+plugins:
+  - limit-count
+--- request
+GET /apisix/prometheus/metrics
+--- error_code: 404
```
monkeyDluffy6017, line 127: I think you only need to add this test case into the existing test cases?

motongxue: Okay, I will do it according to your suggestions, although existing test case (t/xprc/prometheus.t) I can't test it in my local development environment.

Revolyssup:
@motongxue Can you address the failing CI?

motongxue:
> @motongxue Can you address the failing CI?
@Revolyssup Regrettably, through log investigation, I found that the CI issue occurred not because of my actions but due to problems in other test files. Additionally, as a newcomer, I am still struggling with test cases and therefore do not know how to resolve these issues.

Issue #11018, opened 2024-03-08
Unable to validate audience claim using openid-connect


mc-round2:
### Description

Hi,

Im trying to use OIDC to authenticate my APIs, so far its working but its working a bit too well because every single token can pass as long as it is a valid token. 

Im trying to only allow tokens that were created for the specific API that im trying to connect to be allowed, and tokens that are valid BUT *not for this specific API* to be unauthorized.

My token provider provides the audience claim that ideally should be cross referenced with the client_id of the route. From my understanding of the documentation (and according to the APISIX DOCS AI) this should come out of the box... but it clearly doesnt.

Am I missing something? 

This is the current config that I have for the openid-connect plugin:
```yaml
bearer_only: true
client_id: wtv
client_secret: wtvwtvw
discovery: https://auth.wtv.wtv/auth/realms/wtv/.well-known/openid-configuration
introspection_endpoint: https://auth.wtv.wtv/auth/realms/wtv/protocol/openid-connect/token/introspect
introspection_endpoint_auth_method: client_secret_basic
use_jwks: true
```

shreemaan-abhishek:
It's a bit difficult to locate this problem for me, but did you find any relevant logs? It might be the case that this feature is not supported at all.

mc-round2:
There is no relevant logs no. 
I find it weird that the feature is not supported though... shouldn't this be a basic feature for an API Gateway? What is the point of having a token and not being able to verify that it is the correct one for the API that is trying to connect? I could find in other plugins the mention for the audience claim, so it is not unthought of.

I tried before quite a few combinations using scopes and required_scopes but Im not sure Im understanding how to use them well enough (or if it is related to begin with)

njanelt:
Did you try to leave the client_secret parameter unfilled? So just leave it as <"client_secret": "",>. As you use a bearer to authenticate there shoudn't be a need to provide the secret. Maybe the plugin just uses credentials grant instead if the bearer auth is unsuccessful.

mc-round2:
> Did you try to leave the client_secret parameter unfilled? So just leave it as <"client_secret": "",>. As you use a bearer to authenticate there shoudn't be a need to provide the secret. Maybe the plugin just uses credentials grant instead if the bearer auth is unsuccessful.

If I do this, I get an error message (when using the Dashboard):
```
Invalid plugin data
should have required property 'client_secret'
```

njanelt:
Yes you do need to provide the client secret, just leave it empty.

[...]

PR #11059, opened 2024-03-19
feat(openid-connect): Add extra validation if audience claim exists in the Bearer token 

dtsek:
### Description

>The "aud" (audience) claim identifies the recipients that the JWT is intended for.
>https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3

Issue: `openid-connect` currently does not validate if the audience claim matches the client_id which means that every valid JWT can authenticate using this plugin, even if it was generated for a different client.

Solution: This MR enhances the `introspect` function by adding a simple validation that the audience claim `aud` is the same as the provided `client_id`. 

Note: There is no documentation that can be updated, unless a note should be added to explain this functionality in the page of the plugin. 

Fixes # (issue)
https://github.com/apache/apisix/issues/11018

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -382,7 +382,16 @@ local function introspect(ctx, conf)
         -- Token successfully validated.
         local method = (conf.public_key and "public_key") or (conf.use_jwks and "jwks")
         core.log.debug("token validate successfully by ", method)
+
+        -- Validate audience claim if it exists
+        if res.aud then
+            if res.aud ~= conf.client_id then
+                return ngx.HTTP_UNAUTHORIZED, "Audience does not match the client_id"
+            end
+        end
+
         return res, err, token, res
+
     else
         -- Validate token against introspection endpoint.
         -- TODO: Same as above for public key validation.
@@ -397,6 +406,14 @@ local function introspect(ctx, conf)
         -- Token successfully validated and response from the introspection
         -- endpoint contains the userinfo.
         core.log.debug("token validate successfully by introspection")
+
+        -- Validate audience claim if it exists
+        if res.aud then
+            if res.aud ~= conf.client_id then
+                return ngx.HTTP_UNAUTHORIZED, "Audience does not match the client_id"
+            end
+        end
+
         return res, err, token, res
     end
 end


shreemaan-abhishek:
@dtsek the CI is faililng and could you please add test cases as well?

dtsek:
Hello @shreemaan-abhishek , could you give me some extra information regarding testing since I didn't see something relevant apart from the other tests?

In order to test this feature, I will need an IdP that can provide Bearer tokens for different clients so I can prove that the one with the valid `aud` field is the only one that passes.

Is there something available that I can use from your side? Otherwise I could take a look at something like https://www.oauth.com/playground/oidc.html but it will take some time.

shreemaan-abhishek:
> could you give me some extra information regarding testing since I didn't see something relevant apart from the other tests?

You can use the config that didn't work before your bugfix to show that it works now in the tests. You can take motivation from the existing test cases in `t/plugins/openid-connect.t` and other test cases.

> I will need an IdP that can provide Bearer tokens for different clients so I can prove that the one with the valid aud field is the only one that passes.

You can use existing services defined as docker compose applications [here](https://github.com/apache/apisix/blob/master/ci/pod/docker-compose.plugin.yml), define new services (not recommended because #11068) or use a mock server.

shreemaan-abhishek:
I forgot to share this link that shows how test cases can be written: https://apisix.apache.org/docs/apisix/internal/testing-framework/

dtsek:
Hello @shreemaan-abhishek , unfortunately I cannot provide for the time being a test case because in https://github.com/apache/apisix/blob/c0e3d9150f06c3140a52d145782085d26bc1ea67/ci/pod/keycloak/kcadm_configure_university.sh there is only one client being configured and I would like to avoid modifying many services outside of the plugin.

In order to create a test case to prove the bugfix I will need two valid clients. The correct one should get allowed by `openid-connect` while the other one should get denied. This is the scenario I was facing in my organization's environment and gave the inspiration for the security fix. 

I am also having some issues testing locally the test cases, but it is more related to the needed openresty modules. 

I can always provide some proof from my instance, but I imagine this is not enough for the PR.

markusmueller:
I came across this PR while working on another issue in the openid plugin. While I understand the need for an audience check in the use case mentioned I have concerns about the solution proposed with this change. 

As far as I know neither openid nor oauth2 specify a format, existence or content of an audience claim (e.g. https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3. `The interpretation of audience values is generally application specific.`). 

Some IDPs use JWTs using the client_id as value of the aud claim, but that's not a requirement and the case for every IDP.

Hard-coding the check to the client_id configured in the plugin would break the plugin for some IDPs which are not using the client_id in an audience claim but still conforming to the spec.

The apisix openid plugin is using lua-resty-openidc to verify access tokens, the method `jwt_verify` is accepting jwt-validators which can be dynamic. It might be a better way to add a configurable audience/general claim validation to solve this problem?

dtsek:
Hi @markusmueller, so you propose something like additional variables in the config of the plugin to 
1) enable verification of audience
2) specify the field that should be checked against aud 
?

markusmueller:
Yes, I'm suggesting additional config variables and reuse of existing methods. Let me illustrate and hope it gets more clear :-)

The plugin is using [lua-resty-openidc](https://github.com/zmartzone/lua-resty-openidc/tree/9f3a4fcade930f6f38ee0cb43cabf50cebffbcc9) for JWT validation.

Instead of implementing your own claim validation the idea is to reuse existing methods:
1. Find a JSON representation of the JWT validators supported by `lua-resty-openidc` and add it to the plugin config (validators are implemented in `lua-resty-jwt` see for example: https://github.com/SkyLothar/lua-resty-jwt/blob/master/t/validators.t)
3. Instantiate the validator(s) according to the configuration, chain them if its multiple instances
4. Pass the validator(s) to https://github.com/apache/apisix/blob/4df549c21278fbb99a1efba160b2ac9119ce4e1f/apisix/plugins/openid-connect.lua#L373https://docs.github.com/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

Rough draft of the additional config properties:

```{
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "type": {
        "type": "string",
        "title": "type",
        "enum": [
          "matches, equals"
        ]
      },
      "argument": {
        "type": "string",
        "title": "argument",
        "description": "Argument for the validator, for example validator of type matches is accepting a regex",
        "minLength": 1
      },
      "claim": {
        "type": "string",
        "title": "claim",
        "description": "Name of the claim the validator will be applied to",
        "minLength": 1
      }
    },
    "title": "validator",
    "required": [
      "argument",
      "claim",
      "type"
    ]
  },
  "title": "jwt-validators",
  "description": "Array of JWT validators applied to the JWT token"
}
```

Example config:
```
{
  "jwt-validators" : [
    {
      "claim" : "aud",
      "type": "equals",
      "argument" : "your_client_id"
    }
  ]
}
```

Baoyuantop:
Hi all, the issue described in this PR has been resolved in https://github.com/apache/apisix/pull/11987 and released in version 3.12.0.

Issue #8931, opened 2023-02-24
help request: file descriptor overflow when use request-id plugin nanoid algorithm


saiyan-qx:
### Description

When i use  request-id plugin nanoid algorithm，file descriptor increse to the limit of worker_rlimit_nofile : 20480，then cause the error : Too many open files。

I use lsof -p command, find a lot of file descriptors like this:
openresty 955 nfsnobody   45r      CHR                1,9      0t0 71005690 /dev/urandom
openresty 955 nfsnobody   46r      CHR                1,9      0t0 71005690 /dev/urandom
openresty 955 nfsnobody   47r      CHR                1,9      0t0 71005690 /dev/urandom
openresty 955 nfsnobody   48r      CHR                1,9      0t0 71005690 /dev/urandom
openresty 955 nfsnobody   49r      CHR                1,9      0t0 71005690 /dev/urandom
openresty 955 nfsnobody   50r      CHR                1,9      0t0 71005690 /dev/urandom

Thanks!

### Environment

- APISIX version (run `apisix version`): 3.0
- Operating system (run `uname -a`): Linux XXX 3.10.0-1160.80.1.el7.x86_64 #1 SMP Tue Nov 8 15:48:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):nginx version: openresty/1.21.4.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

tokers:
APISIX uses lua nanoid. It might be caused by the nanoid lib.

saiyan-qx:
Eureka!The lua-resty-nanoid lib  code does not execute the close method.

![source code](https://user-images.githubusercontent.com/29767218/221146049-4abf8014-2e0f-46e2-84aa-108d5973aa1f.png)

tokers:
You may try to submit a PR for the lua-resty-nanoid repo, then let's upgrade the dependency in APISIX.

liweitianux:
Interesting, but I found the upstream [lua-resty-nanoid](https://github.com/aikin-vip/lua-resty-nanoid) project doesn't provide the "Issues" section...

[...]

PR #11060, opened 2024-03-19
fix(request-id): file descriptor overflow when use request-id plugin nanoid algorithm

aikin-vip:
### Description

Fixes  (#8931)

### Checklist


t/plugin/request-id3.t
```diff
@@ -0,0 +1,186 @@
+#
```
zll600, line 1: what about `requiest-id-nanoid.t`?

aikin-vip: sure

docs/en/latest/plugins/request-id.md
```diff
@@ -47,6 +47,8 @@ The Plugin will not add a unique ID if the request already has a header with the
 | algorithm           | string  | False    | "uuid"         | ["uuid", "nanoid", "range_id"] | Algorithm to use for generating the unique request ID.                 |
 | range_id.char_set      | string | False | "abcdefghijklmnopqrstuvwxyzABCDEFGHIGKLMNOPQRSTUVWXYZ0123456789| The minimum string length is 6 | Character set for range_id |
 | range_id.length    | integer | False | 16             | Minimum 6 | Id length for range_id algorithm |
+| nanoid.char_set      | string | False | "abcdefghijklmnopqrstuvwxyzABCDEFGHIGKLMNOPQRSTUVWXYZ0123456789_-| The minimum string length is 6 | Character set for nanoid |
+| nanoid.length    | integer | False | 21             | Minimum 6 | Id length for range_id algorithm |
```
liweitianux, line 51: Typo. The description should be `Id length for nanoid algorithm`.

aikin-vip: You are right, done

apisix/plugins/request-id.lua
```diff
@@ -51,6 +51,22 @@ local schema = {
                 }
             },
             default = {}
+        },
+        nanoid = {
```
shreemaan-abhishek, line 55: why is this block needed when just replacing the dependency would fix the issue?

apisix/plugins/request-id.lua
```diff
@@ -79,8 +95,9 @@ local function get_request_id(conf)
     if conf.algorithm == "uuid" then
         return uuid()
     end
+
     if conf.algorithm == "nanoid" then
-        return nanoid.safe_simple()
+        return nanoid.generate(conf.nanoid.length,conf.nanoid.char_set)
```
shreemaan-abhishek, line 100: space needed after comma

t/plugin/request-id.t
```diff
@@ -437,71 +437,3 @@ GET /opentracing
...
-            local t = require("lib.test_admin").test
-            local http = require "resty.http"
-            local v = {}
-            local ids = {}
-            local code, body = t('/apisix/admin/routes/1',
-                 ngx.HTTP_PUT,
-                 [[{
-                        "plugins": {
-                            "request-id": {
-                                "algorithm": "nanoid"
```
shreemaan-abhishek, line 456: your PR suggests that usage method/configuration for nanoid has changed. Which means existing users will face incompatibility issues when upgrading to the newer version. We should avoid such a situation.

apisix/plugins/request-id.lua
aikin-vip: The new dependency provides more controllable parameters and more customized functions, so parameter options are added，just like range_id

apisix/plugins/request-id.lua
aikin-vip: > space needed after comma

OK

t/plugin/request-id.t
aikin-vip: I have noticed the question you raised before submitting. I have already set the initial default value in the parameters, the default parameter value will be the same as the default value generated in previous versions without modifying the parameter value. Therefore, I think the upgrade should be smooth. Is there anything missing from me?

t/plugin/request-id-nanoid.t
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 3: hit
+--- request
+GET /opentracing
```
shreemaan-abhishek, line 108: add error log and status code checks here.

Yokutto:
Uh... I think we should re-open this

PR #11076, opened 2024-03-21
feat: enable sensitive fields encryption by default

shreemaan-abhishek:
### Description

Fixes https://lists.apache.org/thread/30wf080qvrzokronrx283sy8x8kondqf

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -907,7 +907,7 @@ local function enable_gde()
     if enable_data_encryption == nil then
         enable_data_encryption =
             core.table.try_read_attr(local_conf, "apisix", "data_encryption",
-                    "enable_encrypt_fields") and (core.json.encode(core.config.type) == "etcd")
+                    "enable_encrypt_fields") and (core.config.type == "etcd")
         _M.enable_data_encryption = enable_data_encryption
     end
 
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -907,7 +907,7 @@ local function enable_gde()
     if enable_data_encryption == nil then
         enable_data_encryption =
             core.table.try_read_attr(local_conf, "apisix", "data_encryption",
-                    "enable_encrypt_fields")
+                    "enable_encrypt_fields") and (core.json.encode(core.config.type) == "etcd")
         _M.enable_data_encryption = enable_data_encryption
     end
 
--- a/t/admin/consumers.t
+++ b/t/admin/consumers.t
@@ -87,7 +87,7 @@ passed
                         "desc": "new consumer",
                         "plugins": {
                             "key-auth": {
-                                "key": "auth-one"
+                                "key": "4y+JvURBE6ZwRbbgaryrhg=="
                             }
                         }
                     },
@@ -126,7 +126,7 @@ passed
                         "desc": "new consumer",
                         "plugins": {
                             "key-auth": {
-                                "key": "auth-one"
+                                "key": "4y+JvURBE6ZwRbbgaryrhg=="
                             }
                         }
                     },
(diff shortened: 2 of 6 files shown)


moonming:
If a user upgrades APISIX from an old version to a new version, will this configuration be updated? Expected behavior is to not update

shreemaan-abhishek:
> Expected behavior is to not update

By default `encrypt_fields` will be updated to true, the users still have the freedom to set it to `false`.

moonming:
> > Expected behavior is to not update
> 
> 
> 
> By default `encrypt_fields` will be updated to true (only in the `config-default.yaml`, the users still have the freedom to set it to `false`.

Got it, so it's a break change

apisix/plugin.lua
```diff
@@ -907,7 +907,7 @@ local function enable_gde()
     if enable_data_encryption == nil then
         enable_data_encryption =
             core.table.try_read_attr(local_conf, "apisix", "data_encryption",
-                    "enable_encrypt_fields")
+                    "enable_encrypt_fields") and (core.json.encode(core.config.type) == "etcd")
```
shreemaan-abhishek, line 910: encrypt fields only when config type is etcd.

t/plugin/key-auth.t
```diff
@@ -23,6 +23,31 @@ use t::APISIX 'no_plan';
...
+    admin_key: null
+apisix:
+  node_listen: 1984
+  proxy_mode: http&stream
+  stream_proxy:
+    tcp:
+      - 9100
+  enable_resolv_search_opt: false
+  data_encryption:
+    enable_encrypt_fields: false
```
shreemaan-abhishek, line 45: disable encrypt fields for test because this line in the tests:
https://github.com/shreemaan-abhishek/apisix/blob/7c2994b844aef288030392d85bcc9bfbfb46ad0a/t/plugin/key-auth.t#L206-L207

expects response with sensitive fields in plain text... modifying the test to expect encrypted fields would make the tests too complex.

apisix/plugin.lua
```diff
@@ -907,7 +907,7 @@ local function enable_gde()
     if enable_data_encryption == nil then
         enable_data_encryption =
             core.table.try_read_attr(local_conf, "apisix", "data_encryption",
-                    "enable_encrypt_fields")
+                    "enable_encrypt_fields") and (core.config.type == "etcd")
```
shreemaan-abhishek, line 910: encrypt fields only when config type is etcd. Without this addition, enabling `encrypt_fields` on standalone mode would cause unexpected failures.

t/admin/consumers.t
```diff
@@ -87,7 +87,7 @@ passed
                         "desc": "new consumer",
                         "plugins": {
                             "key-auth": {
-                                "key": "auth-one"
+                                "key": "4y+JvURBE6ZwRbbgaryrhg=="
```
shreemaan-abhishek, line 90: since data encryption is `true` now, the body returned after a `PUT` operation would consist of encrypted information.

t/plugin/key-auth.t
```diff
@@ -23,6 +23,31 @@ use t::APISIX 'no_plan';
...
```
shreemaan-abhishek, line 45: disable encrypt fields for this test file because this line in the tests:
https://github.com/shreemaan-abhishek/apisix/blob/7c2994b844aef288030392d85bcc9bfbfb46ad0a/t/plugin/key-auth.t#L206-L207

expects response with sensitive fields in plain text... modifying the test to expect encrypted fields would make the tests too complex.

starsz: Could we only keep 

```
apisix:
    data_encryption:
        enable_encrypt_fields: false
```
this ?

shreemaan-abhishek: I think this should work. let me try

starsz (approved):

moonming (approved):

PR #10788, opened 2024-01-09
fix(openid-connect): openid not closing session and blocking until ttl…

markusmueller:
…l expired when using lockable session storage backend (#8017)

### Description

Fixes #8017

I'm not a Lua developer but I'd really like to see this fixed. It might be appropriate to wrap with a pcall and have the session closed in any case (try-catch-finally). Also, I couldn't quickly figure out if the testing would cover configuring and spinning up a locking session storage (memcache for example) to test this. If so, happy to add a test case.

Thank you for your feedback. 

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #11089, opened 2024-03-25
fix: close session in case of error to avoid blocked session

markusmueller:
…session on locking session storage (example unauth_action=deny)

### Description

This PR is fixing a bug where sessions are blocked until ttl is expired in case of an authentication error. For example unauthenticated session and `unauth_action='deny'` and locking session storage is configured.

Basically the same problem as in https://github.com/apache/apisix/pull/10788 but in case of error.

Fixes # (issue)

### Checklist

### DIFF
--- a/t/plugin/openid-connect5.t
+++ b/t/plugin/openid-connect5.t
@@ -144,18 +144,7 @@ hello world
 === TEST 2: Call to route with locking session storage, no authentication and unauth_action 'deny' should not block subsequent requests on same session
 --- config
     set $session_storage redis;
-    set $session_redis_prefix                   sessions;
-    set $session_redis_database                 0;
-    set $session_redis_connect_timeout          1000; # (in milliseconds)
-    set $session_redis_send_timeout             1000; # (in milliseconds)
-    set $session_redis_read_timeout             1000; # (in milliseconds)
-    set $session_redis_host                     127.0.0.1;
-    set $session_redis_port                     6379;
-    set $session_redis_ssl                      off;
-    set $session_redis_ssl_verify               off;
     set $session_redis_uselocking               on;
-    set $session_redis_spinlockwait             150;  # (in milliseconds)
-    set $session_redis_maxlockwait              30;   # (in seconds)
 
     location /t {
         content_by_lua_block {
--- a/t/plugin/openid-connect5.t
+++ b/t/plugin/openid-connect5.t
@@ -186,19 +186,38 @@ hello world
             -- Extract cookie which is not authenticated
             local cookie_str = concatenate_cookies(res.headers['Set-Cookie'])
 
-            -- Make the call to protected route cookie
+            -- Make the call to protected route with cookie
             local function firstRequest()
                local httpc = http.new()
-               httpc:request_uri(uri, {
+
+               local res, err = httpc:request_uri(uri, {
                         method = "GET",
                         headers = {
                             ["Cookie"] = cookie_str
                         }
                     })
-            end
-            ngx.thread.spawn(firstRequest)
 
-            -- Make second call to protected route and same cookie which should not timeout due ton a blocked session
+                if not res then
+                    ngx.log(ngx.ERR, "request failed with err: ", err)
+                    return
+                end
+                return res
+            end
+
+            local thread = ngx.thread.spawn(firstRequest)
+            ok, res = ngx.thread.wait(thread)
+
+            if not ok then
+                ngx.log(ngx.ERR, "First request did not complete: ", res)
+                return
+            end
+
+            if res.status ~= 401 then
+                ngx.log(ngx.ERR, "Expected status 401 received: ", res.status)
+                return
+            end
+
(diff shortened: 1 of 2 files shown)


shreemaan-abhishek:
the solution looks good in first observation, let's wait for the CI to run then I will review this again.

markusmueller:
@shreemaan-abhishek Is it possible that ci is flaky? The errors don't seem related to this PR to me.

t/plugin/openid-connect5.t
```diff
@@ -138,3 +138,97 @@ __DATA__
...
+                                "bearer_only" : false,
+                                "timeout": 10,
+                                "introspection_endpoint_auth_method": "client_secret_post",
+                                "introspection_endpoint": "http://127.0.0.1:8080/realms/University/protocol/openid-connect/token/introspect",
+                                "set_access_token_header": true,
+                                "access_token_in_authorization_header": false,
+                                "set_id_token_header": true,
+                                "set_userinfo_header": true,
+                                "set_refresh_token_header": true,
+                                "unauth_action": "deny"
```
shreemaan-abhishek, line 187: are all these configurations required? We should consider using a minimal configuration.

markusmueller: Yes, you are right, I have removed the properties which are not needed. Thank you!

shreemaan-abhishek: what about these? 
image

markusmueller: Also got minimal redis config now

shreemaan-abhishek:
@starsz does this feature look good to you?

t/plugin/openid-connect5.t
```diff
@@ -138,3 +138,77 @@ __DATA__
...
+                 [[{
+                        "plugins": {
+                            "openid-connect": {
+                                "discovery": "http://127.0.0.1:8080/realms/University/.well-known/openid-configuration",
+                                "realm": "University",
+                                "client_id": "course_management",
+                                "client_secret": "d1ec69e9-55d2-4109-a3ea-befa071579d5",
+                                "redirect_uri": "http://127.0.0.1:]] .. ngx.var.server_port .. [[/authenticated",
+                                "ssl_verify": false,
+                                "unauth_action": "deny"
```
shreemaan-abhishek, line 167: does unauth_action mean that all unauth reqs will be denied by default?

t/plugin/openid-connect5.t
```diff
@@ -138,3 +138,77 @@ __DATA__
     }
 --- response_body_like
 hello world
+
+
+
+=== TEST 2: Call to route with locking session storage, no authentication and unauth_action 'deny' should not block subsequent requests on same session
+--- config
+    set $session_storage redis;
+    set $session_redis_uselocking               on;
```
shreemaan-abhishek, line 147: why are these two required?

t/plugin/openid-connect5.t
```diff
@@ -138,3 +138,77 @@ __DATA__
...
+                )
+
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello"
+
+            -- Make the final call to protected route WITHOUT cookie
+            local httpc = http.new()
+            local res, err = httpc:request_uri(uri, {method = "GET"})
+
+            -- Extract cookie which is not authenticated
+            local cookie_str = concatenate_cookies(res.headers['Set-Cookie'])
```
shreemaan-abhishek, line 187: I don't think the response would contain `set-cookie` header 🤔

shreemaan-abhishek: okay my bad, I think the request will first be redirected to /authenticate

t/plugin/openid-connect5.t
markusmueller: yes

t/plugin/openid-connect5.t
markusmueller: To enable the locking session storage, thats what this bug is about

shreemaan-abhishek: oops, well it was too obvious 😂

t/plugin/openid-connect5.t
```diff
@@ -138,3 +138,77 @@ __DATA__
...
+            local function firstRequest()
+               local httpc = http.new()
+               httpc:request_uri(uri, {
+                        method = "GET",
+                        headers = {
+                            ["Cookie"] = cookie_str
+                        }
+                    })
+            end
+            ngx.thread.spawn(firstRequest)
```
starsz, line 199: How can you ensure that the first request is arrived before the second one?
I think maybe you can give a return value of the `firstRequest` function , and use `ngx.thread.wait` to get the return value.

markusmueller: @starsz Thank you for the suggestion, I have implemented a wait for the completion of the first request.

starsz (approved):

shreemaan-abhishek (approved):

PR #11090, opened 2024-03-26
feat(openid-connect): allow set headers in introspection request

### LINKED ISSUE #10601: help request: How to set header to introspection endpoint used within OIDC plugin?
### Description

In other API gateway I used similar way to set header to my introspection endpoint that I used within OIDC plugin. In pre-function plugin I did set header using their code `gateway.service.request.set_header('header', 'value')`. However, with APISIX, I'm unable to replicate this. Even though I added `serverless-pre-function` plugin with phase set to `rewrite` or `access`, with `ngx.req.set_header` method, header was not set and forwarded to introspection endpoint.

### Environment

- APISIX version: 3.7
- Operating system: Linux apisix-dev-7b89dc65c5-54f8w 5.10.176-157.645.amzn2.x86_64 #1 SMP Tue Mar 28 17:49:06 UTC 2023 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): openresty/1.21.4.2 / 
- Plugin runner version, for issues related to plugin runners: no sure what is this and how to get it
- LuaRocks version, for installation issues (run `luarocks --version`): no luarocks installed

yuweizzz:
### Description

Fixes #10601 

allow set headers in introspection request.

### Checklist


shreemaan-abhishek:
could you please explain about the use cases where this feature would be helpful in the linked issue thread?

yuweizzz:
this pr allows provide extra HTTP headers when talks to the OpenID Connect introspection endpoint, In other API gateway has this feature alrealy, like traefik. and some oauth libs has this feature too, like oauthlib. this may helpful when introspection endpoint is proxied.

shreemaan-abhishek:
@yuweizzz please fix the linter

yuweizzz:
done.

apisix/plugins/openid-connect.lua
```diff
@@ -386,7 +400,20 @@ local function introspect(ctx, conf)
     else
         -- Validate token against introspection endpoint.
         -- TODO: Same as above for public key validation.
+        if conf.introspection_addon_headers then
+            -- http_request_decorator option provides by lua-resty-openidc
```
shreemaan-abhishek, line 404:
```suggestion
            -- http_request_decorator option provided by lua-resty-openidc
```

t/plugin/openid-connect6.t
```diff
@@ -155,3 +155,213 @@ passed
...
+                                "discovery": "http://127.0.0.1:8080/realms/University/.well-known/openid-configuration",
+                                "redirect_uri": "http://localhost:3000",
+                                "ssl_verify": false,
+                                "timeout": 10,
+                                "bearer_only": true,
+                                "realm": "University",
+                                "introspection_endpoint_auth_method": "client_secret_post",
+                                "introspection_endpoint": "http://127.0.0.1:8080/realms/University/protocol/openid-connect/token/introspect",
+                                "introspection_addon_headers": {
+                                    "X-Addon-Header-A": "VALUE",
```
shreemaan-abhishek, line 182: hardcoding values in configuration doesn't make sense, I think the configuration should just specify header fields which would be extracted from the original request.

yuweizzz: do you mean change the headers from the original request?

shreemaan-abhishek: no, the plugin configuration should only specify which headers should be sent. APISIX should extract those headers from the request and then add them to the introspection request.

yuweizzz: I thinking it's hard to decide which headers should be sent, maybe we should just keep it open here and leave it to user customization?

t/plugin/openid-connect6.t
```diff
@@ -155,3 +155,213 @@ passed
...
+                local body = json_decode(res.body)
+                local accessToken = body["access_token"]
+
+                -- Access route using access token. Should work.
+                uri = "http://127.0.0.1:" .. ngx.var.server_port .. "/hello"
+                local res, err = httpc:request_uri(uri, {
+                    method = "GET",
+                    headers = {
+                        ["Authorization"] = "Bearer " .. body["access_token"],
+                        ["X-Addon-Header-A"] = "Value-A",
```
shreemaan-abhishek, line 240: what happens if we don't pass these headers?

t/plugin/openid-connect6.t
```diff
@@ -155,3 +155,213 @@ passed
...
+                ngx.say(false)
+            end
+        }
+    }
+--- response_body
+true
+--- grep_error_log eval
+qr/token validate successfully by \w+/
+--- grep_error_log_out
+token validate successfully by introspection
```
shreemaan-abhishek, line 263: I think we can check this just like so:

```perl
--- error_log
token validate successfully by introspection
```

t/plugin/openid-connect6.t
```diff
@@ -155,3 +155,213 @@ passed
...
+
+
+
+=== TEST 6: Access route with an invalid token, should work as expected too.
```
shreemaan-abhishek, line 267:
```suggestion
=== TEST 6: Access route with an invalid token, should fail
```

t/plugin/openid-connect6.t
```diff
@@ -155,3 +155,213 @@ passed
...
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 8: Check http headers from fake introspection endpoint.
```
shreemaan-abhishek, line 342: I don't understand what's happening in this test.

docs/en/latest/plugins/openid-connect.md
```diff
@@ -89,6 +89,7 @@ description: OpenID Connect allows the client to obtain user information from th
 | cache_segment | string | False |  |  | Optional name of a cache segment, used to separate and differentiate caches used by token introspection or JWT verification. |
 | introspection_interval | integer | False | 0 |  | TTL of the cached and introspected access token in seconds. |
 | introspection_expiry_claim | string | False |  |  | Name of the expiry claim, which controls the TTL of the cached and introspected access token. The default value is 0, which means this option is not used and the plugin defaults to use the TTL passed by expiry claim defined in `introspection_expiry_claim`. If `introspection_interval` is larger than 0 and less than the TTL passed by expiry claim defined in `introspection_expiry_claim`, use `introspection_interval`. |
+| introspection_addon_headers | string[] | False |  |  | Array of strings. Used in Append extras headers to the introspection http request, If configured header not exist in origin request, it will be ignore. |
```
kayx23, line 92:
```suggestion
| introspection_addon_headers | string[] | False |  |  | Array of strings. Used to append additional header values to the introspection HTTP request. If the specified header does not exist in origin request, value will not be appended. |
```

t/plugin/openid-connect6.t
yuweizzz: the test just for compare with test 6 and test 8. It should work normally if we don't pass these headers.

t/plugin/openid-connect6.t
yuweizzz: test 8 make a fake introspection endpoint which logs requests header, so we can check if those header same as the origin request.

yuweizzz:
any update?

shreemaan-abhishek (approved):

starsz (approved):

PR #11095, opened 2024-03-27
feat: support more sensitive fields for encryption

shreemaan-abhishek:
### Description

Some sensitive plugin fields aren't under the protection of `encrypt_fields` feature. It has been addressed in this PR.

### Checklist

### DIFF
--- a/apisix/plugins/jwe-decrypt.lua
+++ b/apisix/plugins/jwe-decrypt.lua
@@ -51,6 +51,7 @@ local consumer_schema = {
         is_base64_encoded = { type = "boolean" },
     },
     required = { "key", "secret" },
+    encrypt_fields = { "key", "secret" },
 }
 
 
--- a/apisix/plugins/jwe-decrypt.lua
+++ b/apisix/plugins/jwe-decrypt.lua
@@ -15,6 +15,7 @@
 -- limitations under the License.
 --
 local core            = require("apisix.core")
+local plugin          = require("apisix.plugin")
 local consumer_mod    = require("apisix.consumer")
 local base64          = require("ngx.base64")
 local aes             = require("resty.aes")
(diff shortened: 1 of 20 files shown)


t/plugin/authz-casdoor.t
```diff
@@ -506,6 +506,8 @@ apisix:
             -- get plugin conf from etcd, password is encrypted
             local etcd = require("apisix.core.etcd")
             local res = assert(etcd.get('/routes/1'))
+            local core = require("apisix.core")
+            core.log.warn("dibag", core.json.encode(res.body.node.value.plugins["authz-casdoor"].client_secret))
```
starsz, line 510: Debug line ?

shreemaan-abhishek: oops, let me remove it.

apisix/plugins/jwe-decrypt.lua
```diff
@@ -71,15 +72,26 @@ function _M.check_schema(conf, schema_type)
...
-        else
-            if #conf.secret ~= 32 then
-                return false, "the secret length should be 32 chars"
+        local local_conf, err = core.config.local_conf(true)
+        if not local_conf then
+            return false, "failed to load the configuration file: " .. err
+        end
+
+        local encrypted = core.table.try_read_attr(local_conf, "apisix", "data_encryption",
+        "enable_encrypt_fields") and (core.config.type == "etcd")
```
starsz, line 81: Why do we need to add this ? 
I think the `check_schema` is run before the fields encrypted.

shreemaan-abhishek: if data encryption is enabled then the secret length will be more than 32. So we should not check the length if data encryption is on.

> I think the check_schema is run before the fields encrypted.

yes. This is why we cannot use the code in `plugin.lua`, so I just copied the logic 😅

moonming (approved):

apisix/plugins/jwe-decrypt.lua
starsz: > if data encryption is enabled then the secret length will be more than 32.

Why the secret length will be more than 32. Is the secret is encrypted?

shreemaan-abhishek: yes, exactly.

starsz (approved):

PR #11103, opened 2024-03-28
fix(prometheus): disable features when plugin is turned off

motongxue:
### Description

Fixes # ([issue](https://github.com/apache/apisix/issues/11046))

### Checklist

### DIFF
--- a/apisix/utils/batch-processor.lua
+++ b/apisix/utils/batch-processor.lua
@@ -15,6 +15,7 @@
 -- limitations under the License.
 --
 local core = require("apisix.core")
+local plugin = require("apisix.plugin")
 local setmetatable = setmetatable
 local timer_at = ngx.timer.at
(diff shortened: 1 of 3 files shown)


apisix/utils/batch-processor.lua
```diff
@@ -184,8 +185,9 @@ function batch_processor:push(entry)
         return
     end
 
-    if prometheus and prometheus.get_prometheus() and not batch_metrics and self.name
-       and self.route_id and self.server_addr then
+    if not plugin.check_disable_with_metatable(plugin.get("prometheus"))
```
shreemaan-abhishek, line 188: your PR doesn't make sense, the existing logic is just as same as yours.

shreemaan-abhishek:
I will close this and create a PR myself.

PR #11117, opened 2024-04-03
fix(prometheus): disable features when prometheus plugin is turned off

### LINKED ISSUE #11046: bug: Disable All Prometheus-related Features When Prometheus Plugin is Turned Off

shreemaan-abhishek:
### Description

This PR allows disabling the export server when prometheus is removed from plugins list in config.yaml and then the plugin reload api is called.

Fixes #11046 

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -331,6 +331,8 @@ function _M.load(config)
         return local_plugins
     end
 
+    local exporter = require("apisix.plugins.prometheus.exporter")
+
     if ngx.config.subsystem == "http" then
         if not http_plugin_names then
             core.log.error("failed to read plugin list from local file")
@@ -344,6 +346,20 @@ function _M.load(config)
             if not ok then
                 core.log.error("failed to load plugins: ", err)
             end
+
+            if ngx.config.subsystem == "http" then
+                local enabled = core.table.array_find(http_plugin_names, "prometheus") ~= nil
+                local active  = exporter.get_prometheus() ~= nil
+                core.log.warn("inside subsystem: ", enabled, ". ", active, core.json.encode(http_plugin_names))
+                if not enabled and active then
+                    core.log.warn("disabled and active")
+                    exporter.destroy()
+                end
+                if enabled and not active then
+                    core.log.warn("enabled and inactive")
+                    exporter.http_init()
+                end
+            end
         end
     end
 
@@ -353,6 +369,14 @@ function _M.load(config)
         local ok, err = load_stream(stream_plugin_names)
         if not ok then
             core.log.error("failed to load stream plugins: ", err)
+        end
+
+        if ngx.config.subsystem == "stream" then
+            if not core.table.array_find(stream_plugin_names, "prometheus") then
+                exporter.destroy()
+            else
+                exporter.stream_init()
+            end
         end
     end
 
@@ -376,8 +400,6 @@ function _M.exit_worker()
     for name in pairs(stream_local_plugins_hash) do
         unload_plugin(name, PLUGIN_TYPE_STREAM)
     end
-
-    require("apisix.plugins.prometheus.exporter").destroy()
 end
 
 
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -347,15 +347,13 @@ function _M.load(config)
                 core.log.error("failed to load plugins: ", err)
             end
 
-            if ngx.config.subsystem == "http" then
-                local enabled = core.table.array_find(http_plugin_names, "prometheus") ~= nil
-                local active  = exporter.get_prometheus() ~= nil
-                if not enabled and active then
-                    exporter.destroy()
-                end
-                if enabled and not active then
-                    exporter.http_init()
-                end
+            local enabled = core.table.array_find(http_plugin_names, "prometheus") ~= nil
+            local active  = exporter.get_prometheus() ~= nil
+            if not enabled and active then
+                exporter.destroy()
+            end
+            if enabled and not active then
+                exporter.http_init()
             end
         end
     end
@@ -366,14 +364,6 @@ function _M.load(config)
         local ok, err = load_stream(stream_plugin_names)
         if not ok then
             core.log.error("failed to load stream plugins: ", err)
-        end
-
-        if ngx.config.subsystem == "stream" then
-            if not core.table.array_find(stream_plugin_names, "prometheus") then
-                exporter.destroy()
-            else
-                exporter.stream_init()
-            end
         end
     end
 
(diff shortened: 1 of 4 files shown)


t/cli/test_prometheus_reload.sh
```diff
@@ -0,0 +1,93 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+. ./t/cli/common.sh
+
+git checkout conf/config.yaml
+
+sleep 1
```
starsz, line 24: I think we don't need to sleep here.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -112,6 +113,9 @@ function _M.http_init(prometheus_enabled_in_stream)
     -- todo: support hot reload, we may need to update the lua-prometheus
     -- library
     if ngx.get_phase() ~= "init" and ngx.get_phase() ~= "init_worker"  then
+        if prometheus_bkp ~= nil then
```
starsz, line 116: prometheus_bkp would be better.

apisix/plugin.lua
```diff
@@ -344,6 +346,15 @@ function _M.load(config)
             if not ok then
                 core.log.error("failed to load plugins: ", err)
             end
+
+            local enabled = core.table.array_find(http_plugin_names, "prometheus") ~= nil
+            local active  = exporter.get_prometheus() ~= nil
```
starsz, line 351: I think we don't need to get the active status.
Since the `exporter.destroy` is harmless.

t/plugin/prometheus4.t
```diff
@@ -277,3 +277,140 @@ apisix_http_latency_bucket\{type="upstream",route="1",service="",consumer="",nod
...
+
+=== TEST 11: remove prometheus -> reload -> send batch request -> add prometheus for next tests
+--- config
+    location /t {
+        content_by_lua_block {
+            local http = require "resty.http"
+            local httpc = http.new()
+
+            local t = require("lib.test_admin").test
+                    ngx.sleep(0.1)
```
starsz, line 343: Code style.

t/plugin/prometheus4.t
```diff
@@ -277,3 +277,140 @@ apisix_http_latency_bucket\{type="upstream",route="1",service="",consumer="",nod
...
+            require("lib.test_admin").set_config_yaml(data)
+            local code, _, org_body = t('/v1/plugins/reload', ngx.HTTP_PUT)
+            local code, body = t('/batch-process-metrics',
+                 ngx.HTTP_GET
+                )
+
+            ngx.status = code
+            ngx.say(body)
+
+
```
starsz, line 368:
```suggestion
```

t/plugin/prometheus4.t
```diff
@@ -277,3 +277,140 @@ apisix_http_latency_bucket\{type="upstream",route="1",service="",consumer="",nod
...
+    example-plugin:
+        val: 1
+        ]]
+            require("lib.test_admin").set_config_yaml(data)
+            local code, _, org_body = t('/v1/plugins/reload', ngx.HTTP_PUT)
+            local code, body = t('/batch-process-metrics',
+                 ngx.HTTP_GET
+                )
+
+            ngx.status = code
```
starsz, line 365: Why the response code is 404 ?

shreemaan-abhishek: not sure I copied this test from here: https://github.com/apache/apisix/blob/e36db170f7e7d3ca0901bedd66ff175f55557250/t/plugin/prometheus2.t#L337C1-L389C20 😬 

I tried to find the reason but could not.

apisix/plugin.lua
shreemaan-abhishek: you are right

apisix/plugins/prometheus/exporter.lua
shreemaan-abhishek: done

t/cli/test_prometheus_reload.sh
shreemaan-abhishek: done

t/plugin/prometheus4.t
starsz: I know. The 404 is from upstream(1980) not from apisix.
Since we don't have any handler of `batch-process-metrics` in [lib/server.lua](https://github.com/apache/apisix/blob/master/t/lib/server.lua)

apisix/plugin.lua
shreemaan-abhishek: no I think the second if check needs the `active` status. Consider this case:

- apisix is running with prometheus active and enabled
- hit plugin reload

In this case `enabled == true` and `active == true`. But we need not call `http_init()` in prometheus because it is already running during this operation.

t/plugin/prometheus4.t
shreemaan-abhishek: ahhh yes! you are right.

apisix/plugin.lua
starsz: Got it.

starsz (approved):

moonming (approved):

apisix/plugins/prometheus/exporter.lua
```diff
@@ -545,4 +552,13 @@ function _M.get_prometheus()
     return prometheus
 end
 
+
+function _M.destroy()
+    if prometheus ~= nil then
+        prometheus_bkp = core.table.deepcopy(prometheus)
```
nic-6443, line 558: Why use `deepcopy`? Using the original object directly should be fine.

PR #11133, opened 2024-04-09
feat(kafka-logger): add max req/resp body size attributes

shreemaan-abhishek:
### Description

In cases where the req/response body is very large and when include_req_body/resp_body is enabled this leads to very high CPU usage and can even cause unavailability. To fix such event, two new attributes are introduced:

- max_resp_body_bytes
- max_req_body_bytes

### Checklist


apisix/core/response.lua
```diff
@@ -192,22 +192,35 @@ function _M.hold_body_chunk(ctx, hold_the_copy)
...
+            end
             return chunk
         end
 
-        body_buffer = concat_tab(body_buffer, "", 1, body_buffer.n)
+        local body_data = concat_tab(body_buffer, "", 1, body_buffer.n)
         ctx._body_buffer[ctx._plugin_name] = nil
-        return body_buffer
+        if max_resp_body_bytes and #body_data >= max_resp_body_bytes then
+            body_data = str_sub(body_data, 1, max_resp_body_bytes)
```
starsz, line 221: Why do we need to sub the body_data again here?

shreemaan-abhishek: in 213, we just sub the chunk this happens in the case when this is the first and last chunk of body_filter phase. (since eof == true and body_buffer is nil).

in 221 we need to sub because it is the last chunk but not the very first one.

starsz: I think if it's the first chunk, it will be dealed in line ~202

shreemaan-abhishek: hmm now it makes sense but I think we are calling sub on 213 on `chunk` because `body_buffer` is nil. And on 221 we are calling it on body data formed after `concat_tab` the `body_buffer` table.

starsz (approved):

moonming (approved):

docs/en/latest/plugins/kafka-logger.md
```diff
@@ -55,8 +55,10 @@ It might take some time to receive the log data. It will be automatically sent a
...
 | include_resp_body_expr | array   | False    |                |                       | Filter for when the `include_resp_body` attribute is set to `true`. Response body is only logged when the expression set here evaluates to `true`. See [lua-resty-expr](https://github.com/api7/lua-resty-expr) for more.                                                                                                                        |
+| max_resp_body_bytes    | integer | False    | 524288         | >=1                   | Request bodies within this size will be pushed to kafka, if the size exceeds the configured value it will be truncated before pushing to Kafka.                                                                                                                                                                                                  |
```
kayx23, line 61: @shreemaan-abhishek this should be response bodies eh?

PR #11135, opened 2024-04-09
fix(kafka-pubsub): restore `pb.state` appropriately

### LINKED ISSUE #10938: bug: PubSub/Kafka protobuf decode error when sending response
### Current Behavior

I use pubsub to connect apisix to kafka. I succeeded to create a route, to connect to it via websocket and to send a PubSubReq query. The messages are correctly fetched from kafka but then this error is thrown when apisix try to send the response: bad argument #&#8203;1 to '?' (type 'PubSubResp' does not exists) 

I believe this is a pb state problem  and adding this instruction
local pb_old_state = pb.state(pb_state) 
at the beginning of send_resp in  [pubsub.lua]( https://github.com/apache/apisix/blob/ec380945496324c00e7caaf77dd6ec220b12f1fc/apisix/core/pubsub.lua#L67) fixed the issue for me.

I am not very familiar with protobuf so I am not sure if this instruction is actually missing or if there is an error on my side.

### Expected Behavior

Response should be encoded successfully as the type PubSubResp is defined in [pubsub.proto](https://github.com/apache/apisix/blob/ec380945496324c00e7caaf77dd6ec220b12f1fc/apisix/include/apisix/model/pubsub.proto#L135)

### Error Logs

send_resp(): failed to encode response message, err: bad argument #&#8203;1 to '?' (type 'PubSubResp' does not exists), client: 172.22.0.1, server: _, request: "GET /kafka HTTP/1.1", host: "localhost:9080"

### Steps to Reproduce

1 - Create a route to connect to KAFKA Broker.
2 - Subscribe to this route via web socket.
3 - Send a PubSubReq  to a topic with partition and offset in order that Apisix recieve Kafka messages.
4 - The send_response function should fail to encode the response message.

### Environment

APISIX docker version, image  apache/apisix:3.6.0-debian

shreemaan-abhishek:
### Description

pb state should be restored before doing operations like `pb.encode()` as the state might have been changed/polluted by code in other places that use the pb module.

Fixes #10938

### Checklist


bzp2010 (approved):

apisix/core/pubsub.lua
```diff
@@ -184,11 +188,11 @@ function _M.wait(self)
             goto continue
         end
 
-        -- recovery of stored pb_store
-        local pb_old_state = pb.state(pb_state)
-
+        -- only recover state if it has changed
+        if pb.state() ~= pb_state then
```
bzp2010, line 192: It is a good idea to add a check to make sure that the same pb.state is indeed equal. i.e. verifying that == can indeed be used to check for equality.
I'm concerned that this behavior may be broken by updates to openresty and luajit.

starsz (approved):

apisix/core/pubsub.lua
shreemaan-abhishek: we can deal with that later.

PR #11174, opened 2024-04-20
fix: add post request headers only if auth request method is post

membphis:
sync https://github.com/apache/apisix/pull/11021


moonming (approved):

membphis:
A lot of CI fails for npm related reasons.

This PR only modifies the forward-auth plugin and has nothing to do with npm.

So I think test case blocking can be ignored, so please PMCs review. If you think there is no problem, please approve it.

image

image

image

image

image

nic-chen (approved):

shreemaan-abhishek (approved):

PR #11268, opened 2024-05-20
fix: validation fails causing etcd events not to be handled correctly

### LINKED ISSUE #11267: bug: error in checking schema with etcd change event handling
### Current Behavior

> **This bug was discovered by @juzhiyuan and confirmed by @nic-6443. Thanks for their keen observations!**

When APISIX is initially started, it loads the full configuration from etcd, and when there are any changes to the configuration in etcd, APISIX will get a change event to update its memory cache.

When loading full configuration and processing change events, APISIX will check the correctness of configuration entries to weed out configurations that do not conform to the schema specification.

When checking for change events, if an event in the event set does not pass the check, it will issue an error and interrupt subsequent event processing, so some configuration changes will not be loaded and applied to the data plane.

This is wrong, it should just issue the log and skip this event and move on to the others. When the user fixes the configuration, it will go through the check and apply it in a new event.

### Expected Behavior

Only incorrect events should be skipped and the correct configuration that follows should be applied correctly.

### Error Logs

No logging, logging is interrupted when an error is triggered.

### Steps to Reproduce

1. Modify multiple resources and introduce some error entries in them.
2. you will find that there are correct configurations that should be applied that are not loaded correctly.

### Environment

- APISIX version (run `apisix version`): at least 2.0 to latest
- Operating system (run `uname -a`): independent
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): independent
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): independent
- APISIX Dashboard version, if relevant: independent
- Plugin runner version, for issues related to plugin runners: independent
- LuaRocks version, for installation issues (run `luarocks --version`): independent

bzp2010:
### Description

Fixes #11267

Ensure that all etcd events are handled, rather than jumping out of the execution flow when a checking error is encountered.

Current wrong behavior will cause changes in etcd to no longer take effect on the started data plane.

The issue has existed since at least APISIX 2.0, which was long, 4-5 years ago.

### Checklist

### DIFF
--- a/t/cli/test_etcd_sync_event_handle.sh
+++ b/t/cli/test_etcd_sync_event_handle.sh
@@ -131,6 +131,3 @@ cat logs/error.log | grep "}, {" || (echo "failed: Log case 2 unexpected"; exit 
 
 ## Case3: Ensure that the check schema error is actually triggered.
 cat logs/error.log | grep "failed to check item data" || (echo "failed: Log case 3 unexpected"; exit 1)
-
-# Clean up
-make stop
--- a/t/cli/test_etcd_sync_event_handle.sh
+++ b/t/cli/test_etcd_sync_event_handle.sh
@@ -17,7 +17,7 @@
 # limitations under the License.
 #
 
-#. ./t/cli/common.sh
+. ./t/cli/common.sh
 
 # check etcd while enable auth
 git checkout conf/config.yaml
@@ -42,6 +42,7 @@ deployment:
       - http://127.0.0.1:2379
     prefix: /apisix
 nginx_config:
+  error_log_level: info
   worker_processes: 1
 ' > conf/config.yaml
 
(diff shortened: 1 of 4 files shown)


t/cli/test_etcd_sync_event_handle.sh
```diff
@@ -0,0 +1,133 @@
...
+##       }
+##     },
+##     ...
+##   } },
+##   header = {
+##     ...
+##   }
+## }
+## After check, it only appears when watch recovers and returns events in bulk.
+cat logs/error.log | grep "}, {" || (echo "failed: Log case 2 unexpected"; exit 1)
```
nic-6443, line 130: assert `failed to check item data` error log too.

shreemaan-abhishek: test needs to fail if expectations are not met `||` gives a exit code 0 which is not detected as a failure by the cli testing suite.

bzp2010: @shreemaan-abhishek  Actually, the error doesn't cause the test suite to exit either, I'm not sure what's happening. But I've checked the log manually and it has no errors.

t/cli/test_etcd_sync_event_handle.sh
```diff
@@ -0,0 +1,133 @@
...
+
+# Clean up
+make stop
```
shreemaan-abhishek, line 133: this last make stop is not needed, it is already run by default. Running twice give this error that pollutes the test output log.

image

t/cli/test_etcd_sync_event_handle.sh
bzp2010: @nic-6443 Good catch! Updated.

t/cli/test_etcd_sync_event_handle.sh
bzp2010: Fixed, that because I add the first line `. ./ci/common.sh` finally.

bzp2010: @shreemaan-abhishek Fixed, that because I add the first line `. ./ci/common.sh` finally.

nic-chen (approved):

nic-6443 (approved):

PR #11282, opened 2024-05-23
feat(jwt-auth): support configuring `key_claim_name`

### LINKED ISSUE #11276: feat: As a user, I'd like to use "iss" JWT claim to verify my Apisix Consumers
### Description

I'd like to extend a little [jwt-auth plugin](https://github.com/apache/apisix/blob/master/apisix/plugins/jwt-auth.lua) to add a new configuration parameter:
- `key_claim_name` (default value: `key`). that can be used to specify a different key claim name, to validate the JWT against an Apisix Consumer.

Lil snippet for reference:
```lua
    -- [...]

    local key_claim_name = conf.key_claim_name
    local user_key = jwt_obj.payload and jwt_obj.payload[key_claim_name]
    if not user_key then
        return 401, {message = "Missing " .. key_claim_name .. " claim in JWT token"}
    end

    -- [...]
end
```

Since it would be set to `key` by default, it shouldn't change/break any existing setting, but would be helpful for those who are using different JWT "formats".

For example, the company I'm working at is using `iss` claim as Consumer key. Since we're already using a custom version of the jwt-auth plugin (with this feature implemented), I'd like to integrate this contribution to the official plugin.

---

Do you think I can make a PR with the changes? 🙂

mikyll:
### Description

New **feature**:
- Added a **new configuration parameter** to the schema: `key_claim_name` (default = "key"), so for example one could use "iss" to check the validity of the JWT;

(similar behaviour to [Kong JWT plugin](https://docs.konghq.com/hub/kong-inc/jwt/configuration/#config-key_claim_name))

Fixes #11276 

### Checklist


shreemaan-abhishek (commented):
Please remove the style changes and add test cases to support your fix. Thanks.

mikyll:
@shreemaan-abhishek thank you for the feedback. I understand the importance of keeping PRs focused, but I made the style changes to align with the [code_style](https://github.com/apache/apisix/blob/master/CODE_STYLE.md) guide. Could you please clarify if:

- the style changes should be reverted entirely, despite the guidelines?
- generally speaking, you prefer these style changes to be submitted as a separate PR, or not be carried out at all?

Anyways, I'll be happy to add the necessary test cases, as soon as possible :-)

mikyll:
@shreemaan-abhishek
I did the following:
- pushed [`commit bd76c435a33c7dac91a938654c334eb3d10bd2ae`](https://github.com/apache/apisix/pull/11282/commits/bd76c435a33c7dac91a938654c334eb3d10bd2ae) to remove the style changes and add a part that was missing for the token generation (I actually forgot to mention this part in the commit description);
- pushed [`commit 2af7e156aa07ca29cf19f4934daa605eb0571902`](https://github.com/apache/apisix/pull/11282/commits/2af7e156aa07ca29cf19f4934daa605eb0571902) which includes a test case for the new feature;
- updated the PR title and details to remove anything style-related;

Could you double-check the tests for jwt-auth plugin, to ensure it behaves correctly and doesn't break anything?

mikyll:
@shreemaan-abhishek
I did the following:
- pushed [`commit bd76c435a33c7dac91a938654c334eb3d10bd2ae`](https://github.com/apache/apisix/pull/11282/commits/bd76c435a33c7dac91a938654c334eb3d10bd2ae) to remove the style changes and add a part that was missing for the token generation (I actually forgot to mention this part in the commit description);
- pushed [`commit 2af7e156aa07ca29cf19f4934daa605eb0571902`](https://github.com/apache/apisix/pull/11282/commits/2af7e156aa07ca29cf19f4934daa605eb0571902) which includes a test case for the new feature;
- pushed [`commit 92bd0257ace006260366d96dd2171199ab8ef260`](https://github.com/apache/apisix/pull/11282/commits/92bd0257ace006260366d96dd2171199ab8ef260) to clean the test code and use `/hello` local endpoint instead of external `httpbin`;
- updated the PR title and details to remove anything style-related;

Could you double-check the tests for jwt-auth plugin, to ensure it behaves correctly and doesn't break anything?

mikyll:
Hello, any news on this PR?

apisix/plugins/jwt-auth.lua
```diff
@@ -247,9 +251,9 @@ local function get_rsa_or_ecdsa_keypair(conf)
 end
 
 
-local function get_real_payload(key, auth_conf, payload)
+local function get_real_payload(key, auth_conf, payload, key_claim_name)
     local real_payload = {
-        key = key,
+        [key_claim_name] = key,
```
shreemaan-abhishek, line 256: although the key_claim_name has changed, the value of `key_claim_name` will always be `key`. That doesn't fit into the scene. For example, if the `key_claim_name` is `exp` then the value should be the expiry time, not the key. Right?

shreemaan-abhishek:
@mikyll, please rebase with master there have been some changes in the CI.

apisix/plugins/jwt-auth.lua
mikyll: I'm not sure if I understood your comment correctly, in case I didn't please correct me :)

---

The purpose of `key_claim_name` parameter is just to tell **in which claim the key must be stored**. 

The reason why I also made changes to `get_real_payload()` and the like, was to provide a way to **generate a signed token** with a custom key claim name, through the public API (otherwise I would have needed to make a test case with a hardcoded token, see [**line 182** of commit 2af7e156aa07ca29cf19f4934daa605eb0571902](https://github.com/apache/apisix/pull/11282/commits/2af7e156aa07ca29cf19f4934daa605eb0571902#diff-86d48d79e9b5631994cfc66c95b0dca12a2b03aa242d9e47092a746dc720734fR182) for reference).

Addressing your comment, if `key_claim_name` was "exp", then no, the value shouldn't be the expiry time, but still the key. However, that would cause some problem:
- even though according to JWT standard, [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4), the `exp` claim is optional, the **jwt-auth** plugin requires it to validate the token, and returns `401` if it's not present (at [this line](https://github.com/apache/apisix/blob/2fcfbd83e22301aea4f027738d628f19c262a458/apisix/plugins/jwt-auth.lua#L382)):
  ```bash
  [warn] 64#64: *16957 [lua] jwt-auth.lua:442: phase_func(): failed to verify jwt: Missing one of claims - [ nbf, exp ].,
  ```
- setting `key_claim_name` to "exp" is not feasible, since that would generate a token with duplicated `exp` claim.

At most, we could handle this case by preventing the usage of "exp" and "nbf" as `key_claim_name`, what do you think?

mikyll:
@shreemaan-abhishek I've done the rebase with master, and I also left a reply to your comment.

Possible enhancements for that `get_real_payload()` concern, that come to my mind are:

- (simplest one) **revert the changes to the public API** and make a test case for the `key_claim_name` scenario that just uses a hardcoded token, without generating it with the public API;
- prevent the usage of `exp` and `nbf` claims as values for the `key_claim_name` **query parameter in the public API** (e.g. calls to `apisix_host:9180/apisix/plugin/jwt/sign?key=custom-user-key&key_claim_name=exp` would return an error);

mikyll:
Closing this since it's outdated and is being implemented in https://github.com/apache/apisix/pull/11772 👀

Issue #11229, opened 2024-05-06
help request: openid-connect different clients unexpectedly shared the same auth session


RobWagMLP:
### Description

Hi,
first of all, thanks a lot for that great piece of software. So far the only open source gateway that has (almost) all you need.
I am just facing one issue, that i am not sure about, if it is intended or if i might be doing something wrong.

I am having 2 apps, with a frontend and a backend each, running under different hosts. 
These are handled in an apisisx-instance together with keycloak as idp and oidc-plugin.
my keycloak has 2 different clients, one for App A and one for App B.
I defined rules in apisix, to use the corresponding client-id and client-secret according to path and host.
So something like that:

```
- uri: /web/app1
          hosts:
            - app1.mydomain.de
          plugins:
            openid-connect:
              client_id: ${client_id_a}
              client_secret: ${client_secret_a}
              discovery: {kc_uri}/realms/{kc_realm}/.well-known/openid-configuration
              realm: {kc_realm}
              unauth_action: auth
              logout_path: /logout
              session: 
                secret: ${session_secret_a}             
          upstream:
            type: roundrobin
            discovery_type: dns
            nodes:
              "internal-web.domain.node_a": 1
1- uri: /api/*
          hosts:
            - app1.mydomain.de
          plugins:
            openid-connect:
              client_id: ${client_id_a}
              client_secret: ${client_secret_a}
              discovery: {kc_uri}/realms/{kc_realm}/.well-known/openid-configuration
              realm: {kc_realm}
              unauth_action: deny
              logout_path: /logout
              session: 
                secret: ${session_secret_a}             
          upstream:
            type: roundrobin
            discovery_type: dns
            nodes:
              "internal-api.domain.node_a": 1
              
 - uri: /web/app2
          hosts:
            - app2.mydomain.de
          plugins:
            openid-connect:
              client_id: ${client_id_b}
              client_secret: ${client_secret_b}
              discovery: {kc_uri}/realms/{kc_realm}/.well-known/openid-configuration
              realm: {kc_realm}
              unauth_action: auth
              session: 
                secret: ${session_secret_b}             
          upstream:
            type: roundrobin
            discovery_type: dns
            nodes:
              "internal-web.domain.node_b": 1
1- uri: /api/*
          hosts:
            hosts:
            - app2.mydomain.de
          plugins:
            openid-connect:
              client_id: ${client_id_b}
              client_secret: ${client_secret_b}
[...]

Issue #9834, opened 2023-07-14
400 Request Header Or Cookie Too Large


francoeder:
### Description

Estou nos meus primeiros passos com o Apache APISIX.
I just followed the step by step contained in this page [apisix-workshop](https://boburmirzo.github.io/apisix-workshop/#_route_configuration), promoted by Bobur Umurzokov.

What did I do:
- I uploaded the containers by the docker-compose command
- I created a route to an api that I already have published
- I created an upstream
- I edited the created route to consider the created upstream

However, when invoking a GET request for the mapped route, I get the error "400 Request Header Or Cookie Too Large".
`curl -i -X GET "http://127.0.0.1:9080/api/v1/conteudos"`

Can anyone help me?

### Environment

- APISIX version (run `apisix version`): 2.13.1
- Operating system (run `uname -a`): Windows
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): ???
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 3.4.15
- APISIX Dashboard version, if relevant: 2.10.1
- Plugin runner version, for issues related to plugin runners: -
- LuaRocks version, for installation issues (run `luarocks --version`): -

Revolyssup:
@francoeder Can you paste the exact route and upstream configuration that you have configured?

francoeder:
I managed to solve the problem.

This thread helped me 👉 https://github.com/apache/apisix/issues/6296

I needed to configure a reverse proxy plugin in my route, considering the domain of my API (Same domain used in the upstream) and Voilà! I was able to consume my API via APISIX Api Gateway.

I hope it helps someone in the future.

Revolyssup:
@francoeder You can close this issue. Thanks

sevensolutions:
Because APISIX uses nginx under the hood, i solved it by adding this to my config.yml

```yml
nginx_config:
  http_server_configuration_snippet: |
    large_client_header_buffers 4 32k;
```

wangjinxiang0522:
> Because APISIX uses nginx under the hood, i solved it by adding this to my config.yml
> 
> nginx_config:
>   http_server_configuration_snippet: |
>     large_client_header_buffers 4 32k;

am still getting 400 Request Header Or Cookie Too Large even after putting above changes

PR #11286, opened 2024-05-24
feat(openid-connect plugin): support configuring session name and path

damoshushu:
### Description

In some case, we need to provide different identity providers for different paths within the same host, however apisix uses the name "session" to store the cookie, which will cause the cookie be overwritten by the following authentication action.

for example,
- uri: /web/app1 use oidc with client_id_a
- uri: /web/app2 use another oidc with client_id_b

if we access /web/app1, after authentication, the auth cookie will be stored as "session", then we access /web/app2, after authentication, the auth cookie will be overwritten with `client_id_b`. Now we access /web/app1, the `client_id_b`'s token will be passed to app1.

Fixes # (issue)

https://github.com/apache/apisix/issues/11229
https://github.com/apache/apisix/issues/9834

### Checklist


shreemaan-abhishek:
instead of configuring name of the cookie for all plugins, can we append the path to the cookie name?

shreemaan-abhishek (commented):
instead of configuring name of the cookie for all plugins, can we append the path to the cookie name?

PR #11289, opened 2024-05-24
feat(trafiic-split): support rules based on json request body

wilsonGen:
### Description

After this change, the traffic-split will be able to support rules based on json body in POST request.                                                                                            For example: 
If we create a route like this:
```json
{
   "methods": [
      "POST"
   ],
   "name": "Route based on json body (POST)",
   "uri": "/*",
   "plugins": {
      "traffic-split": {
         "rules": [
            {
               "match": [
                  {
                     "vars": [
                        [
                           "json_body_arg_$.cat.sounds_like",
                           "==",
                           "meow"
                        ]
                     ]
                  }
               ],
               "weighted_upstreams": [
                  {
                     "upstream": {
                        "nodes": {
                           "httpbin.org:80": 1
                        },
                        "type": "roundrobin"
                     },
                     "weight": 3
                  }
               ]
            }
         ]
      }
   },
   "upstream": {
      "nodes": {
         "google.com": 1
      },
      "type": "roundrobin"
   }
}
```
In traffic-split we defined "json_body_arg_cat.sounds_like" == "mewo", which means it will match the json body like this:
```json
{
    "cat": {
        "sounds_like": "meow"
    },
    "other_fields": "hihi"
}
```

Motivation:
Currently the traffic-split plugin can create rules based on the default form of the POST request.
However it does not support rules based on application/json in the POST request body.
I think it will be great if it can support the rules based on json body as well.
As I have the need to have rules based on json body too.
Fixes # (issue)

### Checklist


shreemaan-abhishek:
please add test cases and docs

wilsonGen:
I have added test cases and docs, please let me know if there are anything missing, thank you. 😃

apisix/core/ctx.lua
```diff
@@ -274,6 +276,19 @@ do
...
                 end
 
+            elseif core_str.has_prefix(key, "json_body_arg_") then
+                -- only match json type request body
+                local content_type = request.header(nil, "Content-Type")
+                if content_type ~= nil and core_str.has_prefix(content_type,
+                        "application/json") then
+                    local arg_key = sub_str(key, 15)
+                    local raw_request_body = request.get_body()
+                    local body = cjson.decode(raw_request_body)
```
bzp2010, line 286:
```suggestion
                    local body, err = core.json.decode(raw_request_body)
```

apisix/core/ctx.lua
```diff
@@ -274,6 +276,19 @@ do
...
+                    if body then
+                        val = utils.get_nested_json_value(body, arg_key)
```
bzp2010, line 288: Can we use jsonpath?

wilsonGen: yes, just updated code using jsonpath, thank you. 😃

apisix/core/ctx.lua
```diff
@@ -274,6 +276,22 @@ do
...
                 end
 
+            elseif core_str.has_prefix(key, "json_body_arg_") then
+                -- only match json type request body
+                local content_type = request.header(nil, "Content-Type")
+                if content_type ~= nil and core_str.has_prefix(content_type,
+                        "application/json") then
+                    local arg_key = sub_str(key, 15)
+                    local raw_request_body = request.get_body()
+                    local body, err = json.decode(raw_request_body)
```
soulbird, line 286: Consider using a cache, as this code may be executed multiple times on hot paths.

PR #11305, opened 2024-05-30
fix(ssl): SSL key rotation caused request failure

### LINKED ISSUE #11148: bug: SSL error migrating to APISIX 3.9.0
### Current Behavior

I have an APISIX Gateway 3.6.0 installation with 1 upstream, 1 service, 1 route, and 1 ssl. After upgrading to 3.9.0 the most of the requests to my API end up with SSL error like this (curl verbose response):
```
...
* Send failure: Broken pipe
* LibreSSL SSL_write: LibreSSL/3.3.6: error:02FFF020:system library:func(4095):Broken pipe, errno 32
* Failed sending HTTP request
* Connection #0 to host example.com left intact
curl: (55) getpeername() failed with errno 22: Invalid argument
```

I've managed to reproduce it locally using docker-compose. I've added steps to reproduce and created a small repo with all needed files (docker-compose.yaml, apisix config etc., ssl keys): https://github.com/rastiehaiev/apisix-migration-to-3_9_0-bug.

I tested upgrades:
1. 3.6.0 -> 3.7.0, 
2. 3.7.0 -> 3.8.0, 
3. 3.8.0 -> 3.9.0, 
4. 3.6.0 -> 3.9.0 

And the problem is migrating to 3.9.0, so it seems like some defect was introduced in 3.9.0.

### Expected Behavior

API requests succeed after upgrading APISIX to 3.9.0.

### Error Logs

```
2024/04/10 16:53:53 [alert] 56#56: *2224 ignoring stale global SSL error (SSL: error:1C800064:Provider routines::bad decrypt) while SSL handshaking, client: MY_IP, server: 0.0.0.0:9443
2024-04-10T16:53:53.994117366Z 2024/04/10 16:53:53 [crit] 56#56: *2224 SSL_do_handshake() failed (SSL: error:1C800064:Provider routines::bad decrypt) while SSL handshaking, client: MY_IP, server: 0.0.0.0:9443
```

### Steps to Reproduce

1. Run APISIX Gateway 3.6.0 + etcd using docker-compose.
2. Create upstream, service, route and ssl like this (I used Admin API to create these entities):
```
---
upstreams:

  - id: api
    name: api
    desc: API upstream
    type: roundrobin
    scheme: https
    timeout:
      connect: 5
      read: 120
      send: 10
    nodes:
      - host: "httpbin.org"
        port: 443
        weight: 1

services:

  - id: api-http
    name: api-http
    plugins:
      gzip:
        min_length: 1100
        comp_level: 6
        types:
          - text/plain
          - application/json
        vary: true

routes:

  - id: api-http
    name: api-http
    uri: /*
    hosts:
      - "example.com"
    service_id: api-http
    upstream_id: api
    priority: 100

ssls:
  - id: ssl
    cert: "../certs/example.pem"
    key: "../certs/example.key"
    snis:
      - "example.com"
```
3. Make sure apisix installation works as expected:
```
for i in {1..10}; do curl -k --resolve 'example.com:9443:127.0.0.1' 'https://example.com:9443/get?query=1' -vvv -I || { echo "curl failed at attempt $i"; break; }; done
```
4. Run `docker-compose down` (and make sure etcd volume is not deleted).
5. Change the version of APISIX image from 3.6.0 to 3.9.0. Execute `docker-compose up -d` (there is some [breaking change in configuration](https://github.com/apache/apisix/blob/master/CHANGELOG.md#change) introduced in 3.9.0 - `enable_http2` should be on `apisix` level - make sure to change it).
6. Try curling your APISIX gateway once again:
```
for i in {1..10}; do curl -k --resolve 'example.com:9443:127.0.0.1' 'https://example.com:9443/get?query=1' -vvv -I || { echo "curl failed at attempt $i"; break; }; done
```

**Expected**: requests succeed.
**Actual**: some requests fail with SSL error.

To fix this, the entities created in step 2 should be re-created. Once they are re-created, the error disappears.

More verbose steps to reproduce are in the repository I created, please take a look:
https://github.com/rastiehaiev/apisix-migration-to-3_9_0-bug.

### Environment

- APISIX version (run `apisix version`): 3.6.0 - 3.6.9
- Operating system (run `uname -a`): Darwin NAME 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:11:05 PDT 2024; root:xnu-10063.101.17~1/RELEASE_X86_64 x86_64
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.25.3.1, built by gcc 10.2.1 20210110 (Debian 10.2.1-6), built with OpenSSL 3.2.0 23 Nov 2023, TLS SNI support enabled
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): bitnami/etcd:3.4.15
- APISIX Dashboard version, if relevant: irrelevant

AlinsRan:
### Description

Due to the failure to handle the OpenSSL error generated during the AES decryption process, this error was caught by Nginx during the TLS handshake request, resulting in the request failure.

In the scenario of key rotation, due to the attempt to traverse all keys and attempt decryption, the error messages generated earlier should be ignored when decryption is successful.

reference：
* https://github.com/openresty/lua-resty-string/pull/65#issuecomment-1185765216
* https://segmentfault.com/a/1190000015326570

Fixes #11148 

### Checklist

### DIFF
--- a/apisix/ssl.lua
+++ b/apisix/ssl.lua
@@ -163,7 +163,7 @@ local function aes_decrypt_pkey(origin, field)
         end
 
         if C.ERR_peek_error() then
-            -- clean up the error queue of OpenSSL to prevent 
+            -- clean up the error queue of OpenSSL to prevent
             -- normal requests from being interfered with.
             C.ERR_clear_error()
         end
--- a/apisix/ssl.lua
+++ b/apisix/ssl.lua
@@ -159,10 +159,12 @@ local function aes_decrypt_pkey(origin, field)
     for _, aes_128_cbc_with_iv in ipairs(aes_128_cbc_with_iv_tbl) do
         local decrypted = aes_128_cbc_with_iv:decrypt(decoded_key)
         if decrypted then
-            if C.ERR_peek_error() then
-                C.ERR_clear_error()
-            end
             return decrypted
+        end
+
+        if C.ERR_peek_error() then
+            -- clean up the error queue of OpenSSL to prevent normal requests from being interfered with.
+            C.ERR_clear_error()
         end
     end
 
--- a/apisix/ssl.lua
+++ b/apisix/ssl.lua
@@ -163,7 +163,8 @@ local function aes_decrypt_pkey(origin, field)
         end
 
         if C.ERR_peek_error() then
-            -- clean up the error queue of OpenSSL to prevent normal requests from being interfered with.
+            -- clean up the error queue of OpenSSL to prevent 
+            -- normal requests from being interfered with.
             C.ERR_clear_error()
         end
     end
--- a/apisix/ssl.lua
+++ b/apisix/ssl.lua
@@ -18,7 +18,9 @@ local core           = require("apisix.core")
 local secret         = require("apisix.secret")
 local ngx_ssl        = require("ngx.ssl")
 local ngx_ssl_client = require("ngx.ssl.clienthello")
+local ffi            = require("ffi")
 
+local C = ffi.C
 local ngx_encode_base64 = ngx.encode_base64
 local ngx_decode_base64 = ngx.decode_base64
 local aes = require("resty.aes")
@@ -28,6 +30,10 @@ local assert = assert
 local type = type
 local ipairs = ipairs
 
+ffi.cdef[[
+unsigned long ERR_peek_error(void);
+void ERR_clear_error(void);
+]]
 
 local cert_cache = core.lrucache.new {
     ttl = 3600, count = 1024,
@@ -153,6 +159,9 @@ local function aes_decrypt_pkey(origin, field)
     for _, aes_128_cbc_with_iv in ipairs(aes_128_cbc_with_iv_tbl) do
         local decrypted = aes_128_cbc_with_iv:decrypt(decoded_key)
         if decrypted then
+            if C.ERR_peek_error() then
+                C.ERR_clear_error()
+            end
             return decrypted
         end
     end
--- a/t/admin/ssl4.t
+++ b/t/admin/ssl4.t
@@ -404,3 +404,108 @@ location /t {
 }
 --- response_body
 passed
+
+
+
+=== TEST 12: set ssl(sni: www.test.com), encrypt with the first keyring
+--- yaml_config
+apisix:
+    node_listen: 1984
+    data_encryption:
+        keyring:
+            - edd1c9f0985e76a1
+--- config
+location /t {
+    content_by_lua_block {
+        local core = require("apisix.core")
+        local t = require("lib.test_admin")
+
+        local ssl_cert = t.read_file("t/certs/apisix.crt")
+        local ssl_key =  t.read_file("t/certs/apisix.key")
+        local data = {cert = ssl_cert, key = ssl_key, sni = "test.com"}
+
+        local code, body = t.test('/apisix/admin/ssls/1',
+            ngx.HTTP_PUT,
+            core.json.encode(data),
+            [[{
+                "value": {
+                    "sni": "test.com"
+                },
+                "key": "/apisix/ssls/1"
+            }]]
+            )
+
+        ngx.status = code
+        ngx.say(body)
+    }
+}
+--- response_body
+passed
+
+
+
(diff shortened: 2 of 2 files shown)


apisix/ssl.lua
```diff
@@ -153,6 +159,9 @@ local function aes_decrypt_pkey(origin, field)
     for _, aes_128_cbc_with_iv in ipairs(aes_128_cbc_with_iv_tbl) do
         local decrypted = aes_128_cbc_with_iv:decrypt(decoded_key)
         if decrypted then
+            if C.ERR_peek_error() then
+                C.ERR_clear_error()
+            end
```
nic-6443, line 164: I feel that it is more appropriate to pop up an error when decryption fails. Currently, there may be two unexpected behaviors:
1. When all keys fail to decrypt, the resulting error will be stored in OpenSSL's error queue, which may affect subsequent error checks.
2. When the first key decrypts successfully, no actual error occurs during decryption, but we still pop all errors from the queue.

AlinsRan: In a single call, it may generate multiple openssl errors, and we should not be responsible for handling errors thrown by the library. Instead, we should ensure that no matter how many errors it generates, it should not affect normal requests.


There are primarily two scenarios:

1. For successful decryption (normal request): We do not need to care about the contents of the error queue; the error queue must be cleared.
2. For total decryption failure (failed request): In fact, nginx will clear the error queue before the handshake, so we don't need to handle it additionally.

nic-6443 (approved):

moonming (dismissed):

apisix/ssl.lua
```diff
@@ -153,6 +159,9 @@ local function aes_decrypt_pkey(origin, field)
     for _, aes_128_cbc_with_iv in ipairs(aes_128_cbc_with_iv_tbl) do
         local decrypted = aes_128_cbc_with_iv:decrypt(decoded_key)
         if decrypted then
```
membphis, line 161: I prefer this style, clean the error if fail to call `decrypt`.

```lua

local decrypted = aes_128_cbc_with_iv:decrypt(decoded_key)
if not decrypted then
    -- clean the error in OpenSSL
    if C.ERR_peek_error() then
        C.ERR_clear_error()
    end

else
    return decrypted
end

... ...

Comment (bzp2010) [reply apisix/ssl.lua:161]: Basically, the current way is only clear of accumulated errors on the one occasion when it finally succeeds, while all failures of decryption cause errors to be thrown to Nginx.

I think this should be acceptable? Maybe it's not necessary to clean up the error every time?

**Comment by moonming** [comment]
@AlinsRan please make the code linter happy

**Comment by shreemaan-abhishek** [comment]
@AlinsRan why was this issue occuring only after apisix version upgrade?

**Comment by AlinsRan** [comment]
@shreemaan-abhishek Because a ssl key rotation occurred.  
`edd1c9f0985e76a1 -> qeddd145sfvddff3`
 Related to #10771

File: apisix/ssl.lua (hunk 2 above)
Comment (AlinsRan) [reply apisix/ssl.lua:161]: Agree.

> Can you confirm this? can it work properly?

Before performing a TLS handshake, nginx will clear the error queue of OpenSSL. It needs to ensure that the error is indeed generated during the TLS handshake, in order to reject the request and not be affected by other unexpected errors

File: t/admin/ssl4.t [hunk 3]
```diff
@@ -242,7 +242,6 @@ apisix:
             - qeddd145sfvddff3
 --- error_log
 decrypt ssl key failed
-[alert]
```
AlinsRan, line 245: Because the error generated by decryption has been cleared, ngx will not throw this error.

moonming (approved):

shreemaan-abhishek (approved):
LGTM

membphis (approved):
LGTM

bzp2010 (approved):

pfernandes21:
Hi, I'm experiencing this issue as well. Do you have any planned release for this? Thanks!

moonming:
A new version is expected to be released by the end of July. You can also patch this PR in the previous version first @pfernandes21

PR #11080, opened 2024-03-22
feat: Autogenerate admin api key if not passed

Revolyssup:
### Description

Fixes # (issue)
This PR is part of the proposal which removes hardcoding of sensitive API keys by autogenerating them at either application level(APISIX) or deployment level.

## Notes

1. The admin api key is autogenerated if it is passed empty.
2. All the CLI tests now retrieve the key from config.yaml first before sending request to the API server, which is why the change in tests related to CLI.
3. If key is null or a non empty string, then it will not be changed.
4. This is not recommended at application level and recommended at infra level. For instance, the helm chart can autogenerate key. Therefore users are advised to pass their own unique key. For simpler local, single host deployments, APISIX can autogenerate the key which the users can read from config.yaml.

Related PRs:
https://github.com/apache/apisix-helm-chart/pull/740
https://github.com/apache/apisix-docker/pull/548

### Checklist

- [ ] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

Issue #9600, opened 2023-06-05
discussion: refactor L4 and/or L7 Proxy Enablement


kayx23:
### Description

Enable L4 and/or L7 proxies in APISIX are currently somewhat confusing for users. 

Previous discussion in the mailing list can be found here: https://lists.apache.org/thread/w8681qbk3x9n9s4y6mrhcp18yzhddgyr

Opening this issue for tracking.

To summarize the discussion:

> ## Current logic: 
> 
>  1. To enable only the L7 proxy 
> This is the default configuration  
> 2. To enable both L7 & L4 proxy 
> enable apisix.stream_proxy  
> 3. To enable only the L4 proxy 
> enable apisix.stream_proxy and set enable_admin = false (not easy to understand)
> 
> NOTICE: The above mode is the operation of APISIX in the traditional mode (the default enable_admin is true).
> 
> If the user changes the mode of APISIX to control plane (at this time enable_admin is false by default), the above three become:
> 
> 1. To enable only the L7 proxy
> This is the default configuration
> 2. To enable both L7 \& L4 proxy
> enable apisix.stream_proxy and set apisix.stream_proxy.only = false
> 3. To enable only the L4 proxy
> enable apisix.stream_proxy and set apisix.stream_proxy.only = true
> 
> ## Proposed solution:
> Add a configuration and remove apisix.stream_proxy.only:
> * proxy_mode: http or stream or http&stream
> 
> This has nothing to do with enable_admin.
> 
> 1. To enable only the L7 proxy
> This is the default configuration
> 2. To enable both L7 & L4 proxy
> set apisix.proxy_mode = http&stream  
> 3. To enable only the L4 proxy 
> set apisix.proxy_mode = stream

### Environment

APISIX: current version is 3.2.1

Revolyssup:
@kayx23 Can I take this item?

kayx23:
Please chat with @leslie-tsang or @monkeyDluffy6017 about this, thanks

monkeyDluffy6017:
This is a break change, we need to discuss this in more detail.
What about your opinion? @kingluo @leslie-tsang

monkeyDluffy6017:
LGTM

PR #11312, opened 2024-06-02
feat: move tinyyaml to lyaml

bzp2010:
### Description

APISIX now uses a YAML parser called `tinyyaml`, which has a number of problems, such as:

- accept use cases that do not conform to the YAML specification
```yaml
routes:
    -       # Such indentation is obviously wrong, but certain developers think it is acceptable. Unbelievable!
    id: 1
    uri: "/test"
```
```yaml
such-key:
    - id: second
      service:
        schema: "http",    # This is also obviously wrong; YAML does not need to use "," as a end character.
        host: "127.0.0.1", # This is not "optional", the standard parser does not accept this error at all.
```
- No support for the new YAML standard (e.g. anchors syntax)

```yaml
upstream:
  - &u
    nodes: xxx
    type: roundrobin

routes:
  - id: 1
    upstream: *u
  - id: 2
    upstream: *u
```

There have been many issues logged for tinyyaml. https://github.com/apache/apisix/issues?q=tinyyaml

This PR therefore introduces a change to replace tinyyaml with a wrapper for the libyaml C library, which is the C reference implementation of the YAML project. It is more standardized and advanced.

------

Theoretically, if the user is using a fully YAML-compliant document, the change is insensitive. Otherwise it will report an error.

So I think it's largely backward compatible. But as stated in the background above, in the past our developers created so many buggy test cases that they had to be fixed one by one. The good thing is that they have been fixed completely and without substantially changing the testing process.

The outliers in config-default.yaml have also been fixed.

### Checklist


moonming (approved):

juzhiyuan (approved):
LGTM

AlinsRan (approved):

shreemaan-abhishek (approved):

bzp2010:
Please wait for some more REVIEWERS and don't merge for now.

> 请再等待一些reviewer，暂时不要合并。

ShenFeng312 (approved):
LGTM

bzp2010:
Regarding the changes in config-default.yaml:

- `1:2` will be parsed as number 62 instead of "1:2" in string
- `http&stream` contains the start character `&` of the YAML anchor syntax, which causes parsing errors

Revolyssup (changes requested):
I think this hack can be removed https://github.com/apache/apisix/pull/11080/files#diff-ab3c55c5825b9d6c1f2a2e296c658ea10092998b5a53feca14fd3988bcf8209aR87
where a string was replaced to get around null problems. Will you remove this in this PR?

moonming:
> * `http&stream` contains the start character `&` of the YAML anchor syntax, which causes parsing errors

Do we need to change it to `http, stream`?

bzp2010:
Reply to @moonming 

> Do we need to change it to `http, stream`?

This should be controlled by some boolean value, e.g.
```
proxy_mode:
  http: true
  stream: true
```

It should even be further integrated with port settings, e.g.
```
listen:
  - port: 443
    http: true
  - port: 5432
    stream: true
```

These bad designs will be corrected by me (maybe), but not in this PR. there are many things just waiting to be revolutionized.

bzp2010:
Reply to @Revolyssup 

It could probably be deleted, but I don't want it here. Because I also want to make some changes to where the YAML is generated.

The reason: overwriting the old config.yaml file with the regenerated YAML does not preserve user comments, which may break some user efforts.

So I think it's better to generate it at runtime and use logging output or output to file like `conf/.random_admin_key`

moonming:
> Reply to @moonming
> 
> > Do we need to change it to `http, stream`?
> 
> This should be controlled by some boolean value, e.g.
> 
> ```
> proxy_mode:
>   http: true
>   stream: true
> ```
> 
> It should even be further integrated with port settings, e.g.
> 
> ```
> listen:
>   - port: 443
>     http: true
>   - port: 5432
>     stream: true
> ```
> 
> These bad designs will be corrected by me (maybe), but not in this PR. there are many things just waiting to be revolutionized.

This design was imported by https://github.com/apache/apisix/issues/9600, looking forword a better one.

bzp2010:
> This design was imported by https://github.com/apache/apisix/issues/9600, looking forword a better one.

Yes, the current design is not good but it is still work, so it will not be modified in this PR.

Revolyssup (approved):

PR #11339, opened 2024-06-07
feat(secret): support store ssl.keys ssl.certs in secrets mamager

AlinsRan:
### Description

As part of #8319 

Support for storing certificates and private_keys on ssl through Secret Manager.

It is known that both `cert` and `key` support management through the Secret Manager, and `keys` and `certs` should also support it.

### Checklist


apisix/ssl.lua
```diff
@@ -278,8 +278,8 @@ function _M.check_ssl_conf(in_dp, conf)
     end
 
     for i = 1, numcerts do
-        if not secret.check_secret_uri(conf.cert[i]) and
-            not secret.check_secret_uri(conf.key[i]) then
```
moonming, line 282: Is this a previously hidden bug?

AlinsRan: yes

moonming: do we have test cases to cover this now?

AlinsRan:
```lua
local function check_secret_uri(secret_uri)
    -- ...
    if not string.has_prefix(secret_uri, PREFIX) and
        not string.has_prefix(upper(secret_uri), core.env.PREFIX) then
        return false, "error secret_uri prefix: " .. secret_uri
    end
    -- ...
end
```

It will check the `prefix`, and to trigger an error, the following conditions must be met simultaneously:
1. Use env or secret, which means having the conditions to match the prefix
2. Certs and keys configured

There is also an out of bounds exception, which requires the number of certificates to be greater than the number of key characters to trigger. This scenario is rare.

So this PR covers this point, but the master does not.

moonming (approved):

shreemaan-abhishek (approved):

PR #11341, opened 2024-06-07
feat(env): support environment variables in config-default.yaml

### LINKED ISSUE #11340: feat: support environment variables in config-default.yaml
### Description

Comment in `config.yaml` shows that APISIX support use environment variables in yaml
```
# To configure via environment variables, you can use `${{VAR}}` syntax. For instance:
#
# deployment:
#   role: traditional
#   role_traditional:
#     config_provider: etcd
#   etcd:
#     host:
#       - http://${{ETCD_HOST}}:2379
```

But in `config-defautl.yaml`, it does not work.

```
-- apisix/cli/file.lua

    local default_conf = yaml.load(default_conf_yaml)
    if not default_conf then
        return nil, "invalid config-default.yaml file"
    end

    local_conf_path = profile:customized_yaml_path()
    if not local_conf_path then
        local_conf_path = profile:yaml_path("config")
    end
    local user_conf_yaml, err = util.read_file(local_conf_path)
    if not user_conf_yaml then
        return nil, err
    end

    local is_empty_file = true
    for line in str_gmatch(user_conf_yaml .. '\n', '(.-)\r?\n') do
        if not is_empty_yaml_line(line) then
            is_empty_file = false
            break
        end
    end

    if not is_empty_file then
        local user_conf = yaml.load(user_conf_yaml)
        if not user_conf then
            return nil, "invalid config.yaml file"
        end

        local ok, err = resolve_conf_var(user_conf)
        if not ok then
            return nil, err
        end

        ok, err = merge_conf(default_conf, user_conf)
        if not ok then
            return nil, err
        end
    end
```

The ${{ VAR }} syntax is processed by `resolve_conf_var`, but it is only called for user_conf. 
To support it in config-default.yaml, run `resolve_conf_var` for default_conf as well.

flea1lt:
### Description

Fixes #11340 

### Checklist

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -239,6 +239,11 @@ function _M.read_yaml_conf(apisix_home)
         return nil, "invalid config-default.yaml file"
     end
 
+    local ok, err = resolve_conf_var(default_conf)
+    if not ok then
+        return nil, err
+    end
+
     local_conf_path = profile:customized_yaml_path()
     if not local_conf_path then
         local_conf_path = profile:yaml_path("config")
--- a/t/APISIX.pm
+++ b/t/APISIX.pm
@@ -239,6 +239,9 @@ add_block_preprocessor(sub {
     my ($block) = @_;
     my $wait_etcd_sync = $block->wait_etcd_sync // 0.1;
 
+    if ($block->extra_default_yaml_config) {
+        $default_yaml_config .= $block->extra_default_yaml_config;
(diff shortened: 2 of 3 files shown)


nfrankel:
Probably obsoleted by https://github.com/apache/apisix/pull/11343

flea1lt:
This PR is no longer needed due to the planned refactor of `config-default.yaml` in #11343.

PR #11284, opened 2024-05-23
feat: Normalize OpenTelemetry configuration

nfrankel:
### Description

Regular OpenTelemetry Tracing uses environment variable for configuration. This way, Ops don't need to understand the details of each stack to configure OTEL.

Here's a sample Docker Compose snippet where each stack is implemented on a different technology. Note that all abide by the [same environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/).

```yaml
  pricing:                      #python
    image: otel-pricing:1.1
    environment:
      OTEL_EXPORTER_OTLP_ENDPOINT: http://jaeger:4317
      OTEL_RESOURCE_ATTRIBUTES: service.name=pricing
      OTEL_METRICS_EXPORTER: none
      OTEL_LOGS_EXPORTER: none
  inventory:                 #rust
    image: otel-inventory:1.0
    environment:
      OTEL_EXPORTER_OTLP_ENDPOINT: http://jaeger:4317
      OTEL_RESOURCE_ATTRIBUTES: service.name=inventory
      OTEL_METRICS_EXPORTER: none
      OTEL_LOGS_EXPORTER: none
  warehouse-us:        #golang
    image: otel-warehouse-us:1.0
    environment:
      OTEL_EXPORTER_OTLP_ENDPOINT: http://jaeger:4317
      OTEL_RESOURCE_ATTRIBUTES: service.name=stocks USA
      OTEL_METRICS_EXPORTER: none
      OTEL_LOGS_EXPORTER: none
  warehouse-jp:      #ruby
    image: otel-warehouse-jp:1.0
    environment:
      OTEL_EXPORTER_OTLP_ENDPOINT: http://jaeger:4318
      OTEL_RESOURCE_ATTRIBUTES: service.name=stocks Japan
      OTEL_METRICS_EXPORTER: none
      OTEL_LOGS_EXPORTER: none
```

This PR is a starting point to allow Apache APISIX to follow the same conventions.

Two things of note, that should be discussed in this PR:

* The default OTEL values defined in the spec are not the same as those defined by APISIX
* The units defined in the spec are not the same as those defined by APISIX

### Checklist

- [X] I have explained the need for this PR and the problem it solves
- [X] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [X] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #11343, opened 2024-06-08
feat: move config-default.yaml to hardcoded lua file

bzp2010:
### Description

For the reasons mentioned in the email https://lists.apache.org/thread/nc7qr5bl0z2bmkfb5rqc01c32lkx0gv2, I'm submitting this PR. The front blocking has been resolved #11312 

It moves the config-default.yaml in APISIX to a hardcoded Lua. While it is still a script file in the distribution of APISIX, you can think of it as a "binary" and the user gets the clear and unambiguous message "it should never be modified".

And since it's a script file, it'll be more expressive than a pure Lua, e.g. we can read some of the values from environment variables (although I have a better way of doing that).

------

Email copy:

> Hi, community.
> 
> I am posting to this mailing list a proposal to remove the config-default.yaml default configuration file, and below I will describe the background and reasons for this and consider the benefits to us.
> 
> The proposal does not represent a community resolution, although I am a committer in the APISIX community. We still want to hear a variety of voices to make sure this is truly beneficial.
> 
> # Background
> 
> Now APISIX has two configuration files, the user configuration file config.yaml and the default configuration file config-default.yaml.
> When APISIX starts up, it reads these two configuration files and merges and overwrites the user configuration file with the default configuration to create the configuration that is actually used at runtime.
> 
> The recommendation from the developers is that users should only modify the user configuration file config.yaml, while ensuring that config-default.yaml remains intact.
> This is documented here: https://apisix.apache.org/docs/apisix/next/installation-guide/
> 
>   > APISIX's default configuration can be found in conf/config-default.yaml file and it should not be modified.
>   > It is bound to the source code and the configuration should only be changed by the methods mentioned above.
> 
> Users often ask how they should modify custom configurations, or encounter unintended problems after modifying the default configuration file.
> When explaining to them why they should not modify the default configuration file but copy those keys that need to be modified to the user configuration file and then modify them, and explaining the logic of merging the user configuration file with the default configuration file, I do find this issue to be a troubling one.
> 
> Also, some PRs encountered errors when improving config-default.yaml. For example: https://github.com/apache/apisix/pull/11284
> 
> Let's shift our attention to other major open source projects in the world, I have indeed seen very few implementations with such built-in annoyances. Often software can have its default configuration hard-coded into the program, bundled into a binary for distribution, and the user can use a new configuration file, which they will load at runtime and override the modified keys into the default configuration.
> This way, the user only needs to remember that he should modify config.yaml and nothing else; and APISIX will start normally without providing any configuration. This really reduces the cost of understanding.
> 
> Therefore, I suggest the following.
> 
> # Proposal
> 
> 1. Remove config-default.yaml and move its contents to a Lua file as a Lua table
> 2. Fix test errors
> 3. Use documentation to explicitly document all configuration items, which ideally should be generated from the schema.
> 
> This is APISIX enhancement work, which is not perceived by users unless they are using APISIX in a non-correct way.
> 
> # Benefit
> 
> 1. Reduce the cost of explaining about the section, and the cost of understanding for the user.
> 2. Reduce the APISIX code base
> 3. Improve the documentation quality through the entire process.
> 
> # The End
> 
> The discussion is asking for feedback, and you can reply to the email directly to the mailing list.
> We can discuss constructive feedback. If there is no constructive feedback, I will start working on it.

### Checklist


bzp2010:
This PR has completed most of the work in the proposal, but the update of the documentation is not yet complete.

The schema included in APISIX for checking configuration files is outdated and not at all representative of the actual situation. So it cannot be used to generate usable documentation.
https://github.com/apache/apisix/blob/master/apisix/cli/schema.lua

If the issue were to be addressed in this PR, this PR would become very large, which is not conducive to submission. So I would like to make changes in the next PR to include schema's and document generation.

moonming (approved):

conf/config-default.yaml
```diff
@@ -1,690 +0,0 @@
...
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-# CAUTION: DO NOT MODIFY DEFAULT CONFIGURATIONS IN THIS FILE.
-# Keep the custom configurations in conf/config.yaml.
-#
-
-apisix:
```
nic-6443, line 21: This file contains explanations for all configuration fields, some of which cannot even be found in the documentation. These comments and commented-out configurations were previously good examples for modifying `config.yaml`. Now, this part of the information is missing in `config.lua`. 
I am wondering if we can directly use the content from the original `config-default.yaml` as the new `config.yaml`, allowing users to directly modify or remove configurations within it. The `config.lua` will then fill in default values after user updated.

bzp2010: I think this is better, but would it be better to name it something like `config.yaml.example` as it is long and not easy to find what is needed.
This way, we will provide a few minimal default configuration files, `config.yaml`, and an example file containing all the configuration documentation, and if the user wants to use the full configuration file, he can replace `config.yaml` with `config.yamlexample`.

bzp2010: We don't have documentation for the configuration items in the config file right now, and the jsonschema used for checksums in the code is outdated (out of sync with the config file), which would be another improvement item.
That is, improve the jsonschema and use it to generate human-readable documentation.

conf/config-default.yaml
shreemaan-abhishek, line 1: since the default.yaml file is removed now, should we specify all configurable parameters in config.yaml as yaml comments?

t/core/etcd-mtls.t
```diff
@@ -82,12 +82,8 @@ deployment:
     }
 --- request
 GET /t
---- grep_error_log eval
-qr/init_by_lua:\d+: \S+/
---- grep_error_log_out
-init_by_lua:12: ab
```
shreemaan-abhishek, line 88: why did you remove these?

apisix/cli/config.lua
```diff
@@ -0,0 +1,365 @@
...
+    },
+    admin = {
+      admin_key_required = true,
+      admin_key = { {
+        name = "admin",
+        key = "",
+        role = "admin"
+      }, {
+        name = "viewer",
+        key = "4054f7cf07e344346cd3f287985e76a2",
```
juzhiyuan, line 341: Should we keep the default key in this change? Or generate it randomly as well? 

Other changes LGTM

nic-6443 (approved):

t/core/etcd-mtls.t
bzp2010: @shreemaan-abhishek 

As you can see on line 88, some smarty-pants developer even asserted in the log that the line numbers were fixed, and that it was impossible to init block unchanged, so the line numbers would change.

It is now modified to use multiple qr blocks at the same time with "and" conditions for log checking.

conf/config-default.yaml
bzp2010: As I mentioned at https://github.com/apache/apisix/pull/11343#discussion_r1632214492.

I will do this in `config.yaml.example`, which is just a placeholder that the APISIX script won't read.

apisix/cli/config.lua
bzp2010: That's a really good question, I need to think about it, it could probably be set to null by default.

conf/config-default.yaml
shreemaan-abhishek: got it. You will make that change in this PR itself?

SkyeYoung (approved):

apisix/cli/config.lua
shreemaan-abhishek: I agree, we can keep the default key as null and force the user to set a different admin key. Disallow apisix from starting if a different key is not set in config.yaml.

bzp2010: Now that the viewer user (which exists by default), has been removed and the admin user has an empty password, it will be automatically generated.

conf/config-default.yaml
bzp2010: Yes, it's done.

moonming (approved):

shreemaan-abhishek (approved):

SkyeYoung (approved):

juzhiyuan (approved):

PR #11374, opened 2024-06-26
feat(limit-count-redis): add sync_rate to reduce syncing frequency of counter data with redis

shreemaan-abhishek:
### Description

Add the sync_rate parameter to control the frequency of synchronizing counters with redis in redis based limit-count plugins.

Real-time synchronization will put a lot of pressure on redis, refer to https://docs.konghq.com/hub/kong-inc/rate-limiting/configuration/#config-sync_rate.

### Checklist


apisix/utils/redis-schema.lua
```diff
@@ -42,6 +42,9 @@ local policy_to_additional_properties = {
             redis_ssl_verify = {
                 type = "boolean", default = false,
             },
+            sync_interval = {
+                type = "integer", default = -1
```
zhoujiexiong, line 46: default to 0 and with restriction >= 0?

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+  to_be_synced[key] = true -- add to table for syncing
+  redis_confs[key] = conf
+
+  local incr, ierr = counter:incr(key .. consts.REDIS_COUNTER, cost, 1)
+  if not incr then
+    return nil, "failed to incr num req shdict: " .. ierr, 0
+  end
+  core.log.info("num reqs passed since sync to redis: ", incr)
+
+  local ttl = 0
```
zhoujiexiong, line 107: should update `ttl` before returning?

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+      counter:set(consts.REDIS_SYNCER, true)
+    else
+      core.log.error("failed to create main redis syncer timer: ", err, ". Will retry next time.")
+    end
+  end
+
+  to_be_synced[key] = true -- add to table for syncing
+  redis_confs[key] = conf
+
+  local incr, ierr = counter:incr(key .. consts.REDIS_COUNTER, cost, 1)
```
zhoujiexiong, line 101: why `init_val` is 1 rather than 0?

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+        -- next incoming request will pick this up and create a new timer
+        counter:set(consts.REDIS_SYNCER, false)
+      end
+    end
+  end
+end
+
+function _M.rate_limit_with_delayed_sync(conf, counter, to_be_synced, redis_confs, key, cost,
+                                         limit, window, script)
+  local syncer_started = counter:get(consts.REDIS_SYNCER)
```
zhoujiexiong, line 87: get and test not atomic, multiple timer would be started.

consider using DICT:add, if such. then start timer

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+  if not syncer_started then
+    local ok, err = ngx_timer.at(conf.sync_interval, sync_counter_data, counter, to_be_synced,
```
zhoujiexiong, line 89: `syncer` is globally for a NGX server instance but `conf.sync_interval` is plugin instance's configuration.

zhoujiexiong, line 89: `syncer` is globally for a NGX server instance but `conf.sync_interval` is plugin instance's configuration.

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+                                 redis_confs, script)
+    if ok then
+      counter:set(consts.REDIS_SYNCER, true)
+    else
+      core.log.error("failed to create main redis syncer timer: ", err, ". Will retry next time.")
+    end
+  end
+
+  to_be_synced[key] = true -- add to table for syncing
```
zhoujiexiong, line 98: `to_be_synced & redis_confs` are module scoped variables, worker processes that not running `syncer` would not able to tell `syncer` to sync the keys added by them through this way

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong, line 89: `syncer` is globally for a NGX server instance but `conf.sync_interval` is plugin instance's configuration.

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+      core.log.error("failed to sync shdict data to redis: ", err)
+      return
+    end
+
+    local remaining = res[1]
+    local ttl = res[2]
+    core.log.info("syncing shdict num_req counter to redis. remaining: ", remaining,
+                  " ttl: ", ttl, " reqs: ", num_reqs)
+    counter:set(key .. consts.SHDICT_REDIS_REMAINING, tonumber(remaining), tonumber(ttl))
+    counter:set(key .. consts.REDIS_COUNTER, 0)
```
zhoujiexiong, line 68:
```suggestion
    counter:incr(key .. consts.REDIS_COUNTER, -num_reqs)
```

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+local new_redis_cluster_cli = require("apisix.utils.rediscluster").new
+
+local ngx_timer = ngx.timer
+local redis_stop
+local redis_cluster_stop
+local _M = {}
+
+
+local function sync_counter_data(premature, counter, to_be_synced, redis_confs, script)
+  if premature then
```
zhoujiexiong, line 30: should delete/clear the `syncer` flag here?

and what if the `syncer` worker process crash?

apisix/plugins/limit-count/delayed_syncer.lua
```diff
@@ -0,0 +1,132 @@
...
+    return
+  end
+  for key, _ in pairs(to_be_synced) do
```
zhoujiexiong, line 33: using sub-function for keys?

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: syncer flag? wdym?

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: I didn't get you.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: counter denotes the number of requests past since last sync to redis. So after syncing to redis the number of requests should be zero.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: > get and test not atomic

the document doesn't say anything like that. Could you share the source of this info? 🤔

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: I don't understand you.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: because the first requests should also be counted.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: supporting ttl was complicating the code so I plan to support it in another PR.

apisix/utils/redis-schema.lua
```diff
@@ -42,6 +42,9 @@ local policy_to_additional_properties = {
             redis_ssl_verify = {
                 type = "boolean", default = false,
             },
+            sync_interval = {
```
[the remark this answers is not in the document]
shreemaan-abhishek: only redis based rate limiting plugins will support this feature.

apisix/utils/redis-schema.lua
shreemaan-abhishek: I don't get you. But the default value denotes normal mode of operation. i.e sync with redis immediately.

apisix/utils/redis-schema.lua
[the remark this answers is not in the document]
zhoujiexiong: For further reuse purpose, suggest separating redis-schema definition from concrete plugin policy.

May be we could have structure like this:

`apisix/utils/rate-limiting-policy.lua`
```
- redis[_cluster]_cli_properties -> ref. redis-schema
- sync_interval
```

apisix/utils/redis-schema.lua
zhoujiexiong: Because `delay(<=0)` arg. to the `ngx.timer.at` function means executing the timeout callback as soon as the current handler yield.

But negative number for `interval` have no meaning, so restrict `sync_interval >= 0`. 
And sync_interval = 0(default) means no delay(sync with redis immediately).

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: But for the 1st request after `counter:incr`, the counter value is equal to 2.

```lua
-- run: resty --shdict 'counter 10m' incr.lua
local counter = ngx.shared["counter"]
local incr, _ = counter:incr("key", 1, 1)
print("incr: ", incr)
-- out: incr: 2

```

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: TTL is read from redis and wrote to SHDICT by the `syncer`

```lua
    local res, err = red:eval(script, 1, key, conf.count, conf.time_window, num_reqs)
    if err then
      core.log.error("failed to sync shdict data to redis: ", err)
      return
    end

    local remaining = res[1]
    local ttl = res[2]
    core.log.info("syncing shdict num_req counter to redis. remaining: ", remaining,
                  " ttl: ", ttl, " reqs: ", num_reqs)
    counter:set(key .. consts.SHDICT_REDIS_REMAINING, tonumber(remaining), tonumber(ttl))
```

So just read it from SHDICT is OK?

```lua
  local ttl = 0
  local remaining, err = counter:incr(key .. consts.SHDICT_REDIS_REMAINING,
                                      0 - cost, limit, window)
  local ttl, err = counter:ttl(key .. consts.SHDICT_REDIS_REMAINING)

```

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: Encapsulate a function for synchronizing single key. :D

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: So your idea is `one syncer per worker process`?

If so, why using `SHDICT(consts.REDIS_SYNCER)` which is globally to all worker processes to check `syncer_started`?

And this made me mistakenly think that your idea is `one syncer per nginx server instance`.

 ```lua
 local syncer_started = counter:get(consts.REDIS_SYNCER)
  if not syncer_started then
    local ok, err = ngx_timer.at(conf.sync_interval, sync_counter_data, counter, to_be_synced,
                                 redis_confs, script)
    if ok then
      counter:set(consts.REDIS_SYNCER, true)
    else
      core.log.error("failed to create main redis syncer timer: ", err, ". Will retry next time.")
    end
  end

```

Comment (zhoujiexiong) [reply apisix/plugins/limit-count/delayed_syncer.lua:87]: > > get and test not atomic
> 
> the document doesn't say anything like that. Could you share the source of this info? 🤔

If you want to ensure `one syncer per nginx server instance`(but not), the following logic(get and test) should be considered to be atomic.

```lua
local syncer_started = counter:get(consts.REDIS_SYNCER)
if not syncer_started then
```

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: Even if `one syncer per worker process`, but which plugin instance's `conf.sync_interval` the `syncer` apply?

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: For the `one syncer per worker process` scenario, it's OK.

BTW:
1. `to_be_synced & redis_confs` are not necessary pass to ngx.timer.at as extra args?
2.  using redis client instance instead of redis config? 
     could be like this:  `to_be_synced[key] = { redis_cli = red_cli }`?

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: image

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: For the one syncer per worker process scenario,  ignore it. :D

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong, line 1: format indent whitespace from 2 to 4?

and pass make lint
```
ERROR: apisix/plugins/limit-count/delayed_syncer.lua: line 33: getting the Lua global "pairs"
ERROR: apisix/plugins/limit-count/delayed_syncer.lua: line 67: getting the Lua global "tonumber"
ERROR: apisix/plugins/limit-count/delayed_syncer.lua: line 67: getting the Lua global "tonumber"
```

zhoujiexiong:
@shreemaan-abhishek 

`one syncer per worker process` vs `one syncer per nginx server instance`

How do you think?

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: > Even if `one syncer per worker process`, but which plugin instance's `conf.sync_interval` the `syncer` apply?

but which plugin instance's `conf.sync_interval` the `syncer` apply? 
->
but which plugin `conf` instance's `sync_interval` the `syncer` apply?

apisix/utils/redis-schema.lua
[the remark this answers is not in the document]
shreemaan-abhishek: this would be over-optimisation.

apisix/utils/redis-schema.lua
shreemaan-abhishek: sync_interval is in seconds, 0 second would gives misleading idea of "sync immediately"

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: > using redis client instance instead of redis config?

cosockets cannot be shared to ngx.timer functions.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: nice catch thank you.

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: I get it now, you are right. Great catch!!

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: all plugins will have their own syncer as `key` is unique to each plugin instance.
https://github.com/apache/apisix/blob/4dbecfd560a0d47eb78391b5f8660dd51da9afec/apisix/plugins/limit-count/init.lua#L243

apisix/plugins/limit-count/delayed_syncer.lua
shreemaan-abhishek: > to_be_synced & redis_confs are module scoped variables

I get it now. We need to make `to_be_synced` and `redis_confs` visible to the worker process which has `syncer`.

Thanks for your review again.

zhoujiexiong: > > to_be_synced & redis_confs are module scoped variables
> 
> I get it now. We need to make `to_be_synced` and `redis_confs` visible to the worker process which has `syncer`.
> 
> Thanks for your review again.

:D

Yes.
But these module variables are originally visible to the `per worker process syncer`.

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: > all plugins will have their own syncer as `key` is unique to each plugin instance.
> 
> https://github.com/apache/apisix/blob/4dbecfd560a0d47eb78391b5f8660dd51da9afec/apisix/plugins/limit-count/init.lua#L243

But per current creation logic, syncer is  GLOBAL?
image

apisix/plugins/limit-count/delayed_syncer.lua
zhoujiexiong: > > using redis client instance instead of redis config?
> 
> cosockets cannot be shared to ngx.timer functions.

May be we could consider other implementation instead of ngx.timer. e.g. `request driven syncer`:
if reach` count or time threshold`, sync. immediately, in-place, in the req. handler context

apisix/plugins/limit-count/delayed_syncer.lua
ChuanFF: should get remaining and judge remaining<0  before  incr  key..consts.REDIS_COUNTER ?Otherwise, the request that is rejected will be counted

PR #11403, opened 2024-07-13
chore: warn log when sending requests to external services insecurely

shreemaan-abhishek:
### Description

Enforcing TLS is important when using third party services.

### Checklist


moonming (approved):

Revolyssup (approved):

apisix/plugins/cas-auth.lua
```diff
@@ -49,6 +49,8 @@ local _M = {
 }
 
 function _M.check_schema(conf)
+    local check = {"idp_uri"}
```
bzp2010, line 52: Should `cas_callback_uri` and `logout_uri` be listed as checks? Similar fields in `authz-casdoor` and `authz-keycloak` are listed as checks.

apisix/plugins/forward-auth.lua
```diff
@@ -78,6 +78,8 @@ local _M = {
 
 
 function _M.check_schema(conf)
+    local check = {"uri"}
```
bzp2010, line 81: The `ssl_verify` should also be checked, TLS without server certificate validation is meaningless.

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -124,6 +124,8 @@ function _M.check_schema(conf, schema_type)
     if schema_type == core.schema.TYPE_METADATA then
         return core.schema.check(metadata_schema, conf)
     end
+    local check = {"endpoint_addrs"}
```
bzp2010, line 127: Also check `ssl_verify`.

apisix/plugins/clickhouse-logger.lua
```diff
@@ -89,6 +89,8 @@ function _M.check_schema(conf, schema_type)
     if schema_type == core.schema.TYPE_METADATA then
         return core.schema.check(metadata_schema, conf)
     end
+    local check = {"endpoint_addrs"}
```
bzp2010, line 92: Also check `ssl_verify`.

apisix/plugins/authz-keycloak.lua
```diff
@@ -114,6 +114,9 @@ local _M = {
 
 
 function _M.check_schema(conf)
+    local check = {"discovery", "token_endpoint", "resource_registration_endpoint",
```
bzp2010, line 117: Also check `ssl_verify`.

apisix/plugins/http-logger.lua
```diff
@@ -82,6 +82,8 @@ function _M.check_schema(conf, schema_type)
         return core.schema.check(metadata_schema, conf)
     end
 
+    local check = {"uri"}
```
bzp2010, line 85: Also check `ssl_verify`.

apisix/plugins/cas-auth.lua
shreemaan-abhishek: cas_callback_uri and logout_uri are in the form of "/foo/bar" so we don't need to check them.

t/plugin/authz-casdoor.t
```diff
@@ -105,10 +105,43 @@ __DATA__
     }
 --- response_body
 done
+--- error_log
```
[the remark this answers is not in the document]
shreemaan-abhishek: it would be nice to have but there are 20+ files that would need update, very tedious job and not worth the effort IMHO.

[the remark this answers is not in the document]
bzp2010: I mean, if such is better (for reasons of complexity of modification now and on long term maintainability) and there is no such file now, we can create such a file and place the tests in one place.

Do you think this has advantages for long term maintenance? The problem I can imagine is that plugin contributors need to know that their tests need to be written in multiple test files to test different parts. This requires either contributor knowledge or a careful reviewer.

If so, let's move those tests to a new file.

and cc @moonming @membphis for advice

bzp2010 (commented):
LGTM, only the API style and testing files are still in question.

apisix/plugins/authz-keycloak.lua
```diff
@@ -114,6 +114,11 @@ local _M = {
...
+                    "access_denied_redirect_uri"}
+    core.utils.check_https(check, conf, plugin_name)
+    core.utils.check_tls_bool({conf.ssl_verify}, {"ssl_verify"}, plugin_name)
```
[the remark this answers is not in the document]
bzp2010: I mean `check` and `conf` seem to be reversed. And it seems that `check_tls_bool` can continue to use the string pattern in `check_https` to get data in conf instead of supplying it upfront in the code.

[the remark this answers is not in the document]
shreemaan-abhishek: great catch 🙏🏼 , thanks for your detailed attention.

t/plugin/authz-casdoor.t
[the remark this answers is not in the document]
shreemaan-abhishek: > on long term maintainability

I agree with you on this. 🙏🏼

bzp2010 (approved):

Revolyssup (approved):

moonming (approved):

PR #11415, opened 2024-07-18
fix(api-breaker):The actual number of failures of the circuit breaker is inconsistent with the set number of failures

FirstCatchYourHare:
### Description

Fixes # (issue)

### Checklist

###### api-breaker plug-in configuration
设置
###### Actual test result
QQ图片20240718100029
*You can see from the picture, in the test picture,The actual number of failures did not reach 3 times on the fuse.so that,In the access phase, it is not only necessary to judge the time problem, but also to determine whether the current total number of unhealthy times is equal to the number of unhealthy failures*

### DIFF
--- a/apisix/plugins/api-breaker.lua
+++ b/apisix/plugins/api-breaker.lua
@@ -179,7 +179,7 @@ function _M.access(conf, ctx)
     core.log.info("breaker_time: ", breaker_time)
 
     -- breaker
-    if lasttime + breaker_time >= ngx.time() then
+    if lasttime + breaker_time >= ngx.time() and unhealthy_count % conf.unhealthy.failures == 0 then
         if conf.break_response_body then
             if conf.break_response_headers then
                 for _, value in ipairs(conf.break_response_headers) do
--- a/t/plugin/api-breaker.t
+++ b/t/plugin/api-breaker.t
@@ -652,3 +652,53 @@ phase_func(): breaker_time: 10
 GET /t
 --- response_body eval
 qr/failed to check the configuration of plugin api-breaker err: property \"break_response_headers\" validation failed: failed to validate item 1: property \"(key|value)\" is required/
+
+
+
+=== TEST 21: continuous trigger break, The number of unhealthy instances is uncertain 
+--- config
+location /t {
+    content_by_lua_block {
+        local t = require("lib.test_admin").test
+        local json = require("toolkit.json")
+
+        -- trigger to unhealth
+        for i = 1, 4 do
+            local code = t('/api_breaker?code=500', ngx.HTTP_GET)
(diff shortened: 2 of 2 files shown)


zhoujiexiong:
@FirstCatchYourHare 

How is the behavor of your test client?
Could you please also add test case[s] to the PR to reproduce or prove to fix the "issue"?

FirstCatchYourHare:
continuous trigger break, The number of unhealthy instances is uncertain

zhoujiexiong:
> continuous trigger break, The number of unhealthy instances is uncertain

The new added test case fail the CI, try add `--- timeout: N` to adjust.

image

And your expected behavior like below?

`500 x 3`
[trigger break, 599...]
[sleep/after 2s+, breaker timeout]
`500 x 3`
[trigger break, 599...]
[sleep/after 4s+, breaker timeout]
`500 x 3`
[trigger break, 599...]
[sleep/after 8s+, breaker timeout]
...

shreemaan-abhishek:
this bug only occurs when both `healthy` and `max_breaker_sec` are configured together?

FirstCatchYourHare:
> this bug only occurs when both `healthy` and `max_breaker_sec` are configured together?

No,This bug is easy to reproduce.It can be reproduced with the default configuration.But it needs to be explained in conjunction with the api-breaker plug-in code.
As shown in the following picture:
api-breaker
In an unhealthy state,The request is then retried in 2, 4, 8, 16 ... seconds until the max_breaker_sec.In this process, there is often a situation:when the specified unhealthy state has not been reached, the breaker is triggered.
The reason is that when the specified number of unhealthy states is not reached, lasttime + breaker_time will occasionally appear to be greater than or equal to ngx.time().This causes a breaker to occur.
The solution is shown in the following figure:
api-breaker2

zhoujiexiong:
> > this bug only occurs when both `healthy` and `max_breaker_sec` are configured together?
> 
> No,This bug is easy to reproduce.It can be reproduced with the default configuration.But it needs to be explained in conjunction with the api-breaker plug-in code. As shown in the following picture: api-breaker In an unhealthy state,The request is then retried in 2, 4, 8, 16 ... seconds until the max_breaker_sec.In this process, there is often a situation:when the specified unhealthy state has not been reached, the breaker is triggered. The reason is that when the specified number of unhealthy states is not reached, lasttime + breaker_time will occasionally appear to be greater than or equal to ngx.time().This causes a breaker to occur. The solution is shown in the following figure: api-breaker2

Suppose there are 4 requests inbound, 3 failed, breaker triggered, then the 4th increase the `unhealthy_count` at `log_phase`, the modification of the PR would exit the breaking state?

PR #11417, opened 2024-07-18
feat: support aws secret manager

HuanXin-Chen:
### Description

Many enterprises are utilizing cloud services from AWS and GCP, relying on the secret manager provided by these platforms to handle sensitive information. Integrating Apache APISIX with these secret managers can streamline the process of using sensitive information within APISIX, enabling users to manage and utilize cloud-stored sensitive data more conveniently, thus enhancing the overall security and usability of the system.

This PR has completed the support for AWS. It added the aws.lua file to the original secret module, allowing users to store their secret information on AWS using the same reference method as before.

### Checklist


ci/common.sh
```diff
@@ -180,6 +180,8 @@ linux_get_dependencies () {
     apt update
     apt install -y cpanminus build-essential libncurses5-dev libreadline-dev libssl-dev perl libpcre3 libpcre3-dev libldap2-dev
     apt-get install -y libyaml-dev
+    wget https://github.com/libexpat/libexpat/releases/download/R_2_5_0/expat-2.5.0.tar.gz
```
moonming, line 183: What is the purpose of installing this software? I don't see where it is used.

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    schema = schema
+}
+
+local function initialize_aws()
+    AWS_ACCESS_KEY_ID = env.fetch_by_uri("$ENV://AWS_ACCESS_KEY_ID")
+    AWS_SECRET_ACCESS_KEY = env.fetch_by_uri("$ENV://AWS_SECRET_ACCESS_KEY")
+    AWS_SESSION_TOKEN = env.fetch_by_uri("$ENV://AWS_SESSION_TOKEN")
+    AWS_REGION = env.fetch_by_uri("$ENV://AWS_REGION")
+    AWS = aws()
+    initialize_aws= nil
```
moonming, line 65: please do not use global vars in Lua

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+end
+
+local function make_request_to_aws(conf,key)
+    if initialize_aws then
+        initialize_aws()
+    end
+
+    local region = conf.region or AWS_REGION
```
moonming, line 73: why we need `initialize_aws()` function? How about
```
local region = conf.region or env.fetch_by_uri("$ENV://AWS_REGION")
```

moonming, line 73: do we need to get region from `env.fetch_by_uri(conf.region)`?

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    if not region then
+        return nil, "aws secret manager requires region"
+    end
+
+    local access_key_id = env.fetch_by_uri(conf.access_key_id)
+    if not access_key_id then
+        access_key_id = conf.access_key_id or AWS_ACCESS_KEY_ID
```
moonming, line 80: what should we do if `access_key_id` is nil?

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    local secret_access_key = env.fetch_by_uri(conf.secret_access_key)
+    if not secret_access_key then
+        secret_access_key = conf.secret_access_key or AWS_SECRET_ACCESS_KEY
```
moonming, line 85: ditto

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    local session_token = env.fetch_by_uri(conf.session_token)
+    if not session_token then
+        session_token = conf.session_token or AWS_SESSION_TOKEN
```
moonming, line 90: ditto

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    if initialize_aws then
+        initialize_aws()
+    end
+
+    local region = conf.region or AWS_REGION
+    if not region then
+        return nil, "aws secret manager requires region"
+    end
+
+    local access_key_id = env.fetch_by_uri(conf.access_key_id)
```
moonming, line 78:
```suggestion
    local access_key_id = env.fetch_by_uri(conf.access_key_id) or conf.access_key_id or env.fetch_by_uri("$ENV://AWS_SECRET_ACCESS_KEY")
```

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    if access_key_id and secret_access_key then
+        my_creds = AWS:Credentials {
+            accessKeyId = access_key_id,
+            secretAccessKey = secret_access_key,
+            sessionToken = session_token,
+        }
+    end
+
+    if not my_creds then
+        return nil, "(invalid credentials)"
```
moonming, line 103: we can remove `(` and `)`

moonming, line 103: please add test cases for all invalid return

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    local res, err = sm:getSecretValue {
+        SecretId = key,
+        VersionStage = "AWSCURRENT",
+    }
+
+    if type(res) ~= "table" then
+        if err then
+            return nil, err
+        end
+        return nil, "(invalid response)"
```
moonming, line 127: we need to log error info

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+        local body = res.body
+        if type(body) == "table" then
+            err = core.json.decode(body)
+        end
+
+        if err then
+            return nil, err
+        end
+
+        return nil, "(invalid status)"
```
moonming, line 140: status ~= 200 does not mean invalid. And we need to log err and status.

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+        if err then
+            return nil, err
+        end
+        return nil, "(invalid response)"
+    end
+
+    if res.status ~= 200 then
+        local body = res.body
+        if type(body) == "table" then
+            err = core.json.decode(body)
```
moonming, line 133:
```suggestion
            local data, err = core.json.decode(body)
```

moonming, line 133: need to add test case

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+        end
+
+        if err then
+            return nil, err
+        end
```
moonming, line 138: test case

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    core.log.info("main: ", main_key, " sub: ", sub_key)
+
+    local res,err = make_request_to_aws(conf,main_key)
+    if not res then
+        return nil, "failed to retrtive data from aws secret manager: " .. err
+    end
+
+    local ret = core.json.decode(res)
```
moonming, line 182:
```suggestion
    local data, err = core.json.decode(res)
```

apisix/secret/aws.lua
```diff
@@ -0,0 +1,193 @@
...
+    if not ret then
+        return nil, "failed to decode result, res: " .. res
```
moonming, line 184: log err info

ci/common.sh
HuanXin-Chen: This is because luaexpact depends on Expat and requires a version greater than 2.4; otherwise, it will throw an error. The Expat package in the CLI environment is less than version 2.4. I used the SDK: https://luarocks.org/modules/kong/lua-resty-aws/1.5.1-1.

HuanXin-Chen: image

apisix/secret/aws.lua
HuanXin-Chen: The original SDK would read some system environment variables to make a global default configuration. User customization would override the global configuration. 
However, it calls `os.getenv` , which is inaccessible for APISIX. I want to maintain its original logic, so I made an initialization. 
This step can be removed and does not affect the logic.

ci/common.sh
moonming: So we also need to change the APISIX docker package, right?

HuanXin-Chen: Yes, if want to use the SDK.

t/secret/aws.t
```diff
@@ -0,0 +1,321 @@
...
+
+repeat_each(1);
+no_long_string();
+no_root_location();
+log_level("info");
+
+run_tests;
+
+__DATA__
+
```
bzp2010, line 35: A test for checking that the input configuration conforms to the rules is needed.

ref: https://github.com/apache/apisix/blob/1ded3a837c52fd268927d76f6708553934b77a03/t/plugin/loki-logger.t#L35-L67

apisix/secret/aws.lua
```diff
@@ -0,0 +1,157 @@
...
+
+    core.log.info("main: ", main_key, " sub: ", sub_key)
+
+    local res,err = make_request_to_aws(conf,main_key)
+    if not res then
+        return nil, "failed to retrtive data from aws secret manager: " .. err
+    end
+
+    local ret = core.json.decode(res)
+    if not ret then
```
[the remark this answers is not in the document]
HuanXin-Chen: Mentor, at present I have implemented this function, please review again.

apisix/secret/aws.lua
```diff
@@ -0,0 +1,160 @@
...
+    if main_key == "" then
+        return nil, "can't find main key, key: " .. key
+    end
+
+    local sub_key = idx and sub(key, idx + 1) or nil
+    if not sub_key then
+        core.log.info("main: ", main_key)
+    else
+        core.log.info("main: ", main_key, " sub: ", sub_key)
+    end
```
bzp2010, line 136: core.log.info("main: ", main_key, sub_key and ", sub: " .. sub_key or "")

t/secret/aws.t
```diff
@@ -0,0 +1,315 @@
...
+done
+err
+err
+done
+err
+done
+err
+done
+err
+done
```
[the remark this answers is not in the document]
bzp2010: https://github.com/apache/apisix/blob/1ded3a837c52fd268927d76f6708553934b77a03/t/plugin/loki-logger.t#L35-L67

ci/common.sh
bzp2010: We have removed this dependency from the fork and it is no longer needed.

apisix/secret/aws.lua
bzp2010: fixed

apisix/secret/aws.lua
bzp2010: removed

bzp2010: removed, it uses the schema's default value now

apisix/secret/aws.lua
bzp2010: It's already asserted through the schema, and the user can't create a manager instance with that value being null through the Admin API, so this may not be an issue.
Default values from environment variables have also been removed.

apisix/secret/aws.lua
bzp2010: fixed

apisix/secret/aws.lua
bzp2010: fixed

apisix/secret/aws.lua
bzp2010: I support configurations as a single source of credentials because users can create multiple instances of the same manager, which can cause misuse. Therefore the AWS_SECRET_ACCESS_KEY value should not be read, which is exactly what is done now.

Although AWS_SECRET_ACCESS_KEY is a common value used by the AWS CLI.

In short, fixed.

apisix/secret/aws.lua
bzp2010: fixed

bzp2010: @HuanXin-Chen

Does AWS:Credentials() return unintended null values? Please confirm it.

apisix/secret/aws.lua
```diff
@@ -0,0 +1,157 @@
...
+        local body = res.body
+        if type(body) == "table" then
+            local data = core.json.encode(body)
+            if data then
+                return nil, "invalid status code " .. res.status .. ", " .. data
+            end
+        end
+
+        return nil, "invalid status code " .. res.status
+    end
```
bzp2010, line 105: According `https://github.com/api7/lua-resty-aws/blob/main/src/resty/aws/init.lua#L392-L397`, this needs to be modified.

From the lib code, there are two possibilities for the return value of a call to the service API (like getSecretValue):

- res ~= nil, err == nil
- res == nil, err ~= nil

So the first thing you should do is to check if res is empty, and if it is, it means that err must have a value of type string.

```Lua
if not res then
  return nil, "invalid response: " .. err
end
```

Next, check the `res.status` value. `res` must not be null and must be of type table. record the status code and response as a log.

```Lua
if res.status ~= 200 then
  return nil, "invalid status code " .. res.status .. ", " .. data
end
```

Now `res` is not null and the response is as expected, so `res.body` is the data we need. Parsing it.

This simplifies your code.

apisix/secret/aws.lua
bzp2010: @HuanXin-Chen

apisix/secret/aws.lua
bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

apisix/secret/aws.lua
bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

apisix/secret/aws.lua
bzp2010: @HuanXin-Chen

apisix/secret/aws.lua
bzp2010: @HuanXin-Chen

apisix/secret/aws.lua
bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

apisix/secret/aws.lua
bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

apisix/secret/aws.lua
bzp2010: replace by https://github.com/apache/apisix/pull/11417#discussion_r1714764673

apisix/secret/aws.lua
bzp2010: @HuanXin-Chen

apisix/secret/aws.lua
HuanXin-Chen: resolved

t/secret/aws.t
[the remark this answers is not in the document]
HuanXin-Chen: resolved

apisix/secret/aws.lua
HuanXin-Chen: resolved

apisix/secret/aws.lua
bzp2010: won't

bzp2010 (approved):

bzp2010:
@moonming Please review this again, especially the reasonableness of the test.
@kayx23 When you have a moment, please help check the documentation. 🙏

docs/en/latest/terminology/secret.md
```diff
@@ -190,3 +191,105 @@ curl http://127.0.0.1:9180/apisix/admin/consumers \
 ```
 
 Through the above two steps, when the user request hits the `key-auth` plugin, the real value of the key in the Vault will be obtained through the APISIX Secret component.
+
+## Use AWS Secrets Manager to manage secrets
+
+Managing secrets with AWS Secrets Manager is a secure and convenient way to store and manage sensitive information. This method allows you to save secret information in AWS Secrets Manager and reference these secrets in a specific format when configuring APISIX plugins.
+
+APISIX currently supports two access methods: [long-term credential access](https://docs.aws.amazon.com/zh_cn/sdkref/latest/guide/access-iam-users.html) and [short-term credential access](https://docs.aws.amazon.com/zh_cn/sdkref/latest/guide/access-temp-idc.html).
```
kayx23, line 199:
```suggestion
APISIX currently supports two authentication methods: using [long-term credentials](https://docs.aws.amazon.com/sdkref/latest/guide/access-iam-users.html) and [short-term credentials](https://docs.aws.amazon.com/sdkref/latest/guide/access-temp-idc.html).
```

docs/en/latest/terminology/secret.md
```diff
@@ -190,3 +191,105 @@ curl http://127.0.0.1:9180/apisix/admin/consumers \
...
+| secret_access_key | Yes |  | AWS Secret Access Key |
+| session_token | No |  | Temporary access credential information |
+| region | No | us-east-1 | AWS Region |
+| endpoint_url | No | https://secretsmanager.{region}.amazonaws.com | AWS Secret Manager URL |
+
+### Example: use in key-auth plugin
+
+Here, we use the key-auth plugin as an example to demonstrate how to manage secrets through AWS Secrets Manager.
+
+Step 1: Create the corresponding key in the aws secrets manager.Here, [localstack](https://www.localstack.cloud/) is used for simulation, and you can use the following command:
```
kayx23, line 226:
```suggestion
Step 1: Create the corresponding key in the AWS secrets manager. Here, [localstack](https://www.localstack.cloud/) is used for as the example environment, and you can use the following command:
```

docs/en/latest/terminology/secret.md
```diff
@@ -190,3 +191,105 @@ curl http://127.0.0.1:9180/apisix/admin/consumers \
...
+
+### Required Parameters
+
+| Name | Required | Default Value | Description |
+| --- | --- | --- | --- |
+| access_key_id | Yes |  | AWS Access Key ID |
+| secret_access_key | Yes |  | AWS Secret Access Key |
+| session_token | No |  | Temporary access credential information |
+| region | No | us-east-1 | AWS Region |
+| endpoint_url | No | https://secretsmanager.{region}.amazonaws.com | AWS Secret Manager URL |
```
kayx23, line 220: Please update to True or False

image

docs/en/latest/terminology/secret.md
HuanXin-Chen: fixed

docs/en/latest/terminology/secret.md
HuanXin-Chen: fixed

docs/en/latest/terminology/secret.md
```diff
@@ -190,3 +191,105 @@ curl http://127.0.0.1:9180/apisix/admin/consumers \
...
+### Verification
+
+You can verify this with the following command:
+
+```shell
+#Replace the following your_route with the actual route path.
+curl -i http://127.0.0.1:9080/your_route -H 'apikey: value'
+```
+
+This will verify whether the key-auth plugin is correctly using the key from AWS Secrets Manager.
```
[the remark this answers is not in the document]
HuanXin-Chen: fixed

docs/en/latest/terminology/secret.md
```diff
@@ -190,3 +191,105 @@ curl http://127.0.0.1:9180/apisix/admin/consumers \
...
+
+Here, we use the key-auth plugin as an example to demonstrate how to manage secrets through AWS Secrets Manager.
+
+Step 1: Create the corresponding key in the aws secrets manager.Here, [localstack](https://www.localstack.cloud/) is used for simulation, and you can use the following command:
+
+```shell
+docker exec -i localstack sh -c "awslocal secretsmanager create-secret --name jack --description 'APISIX Secret' --secret-string '{\"auth-key\":\"value\"}'"
+```
+
+Step 2: Add APISIX Secrets resources through the Admin API, configure the connection information such as the address of AWS Secrets Manager:
```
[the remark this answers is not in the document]
HuanXin-Chen: fixed

docs/en/latest/terminology/secret.md
HuanXin-Chen: fixed

moonming (approved):

bzp2010 (dismissed):

apisix/secret/aws.lua
```diff
@@ -0,0 +1,139 @@
...
+    local idx = find(key, '/')
+
+    local main_key = idx and sub(key, 1, idx - 1) or key
+    if main_key == "" then
+        return nil, "can't find main key, key: " .. key
+    end
+
+    local sub_key = idx and sub(key, idx + 1) or nil
+
+    core.log.info("main: ", main_key, sub_key and ", sub: " .. sub_key or "")
```
membphis, line 115: style:

```suggestion
    core.log.info("main: ", main_key, sub_key and ", sub: ", sub_key or "")
```

apisix/secret/aws.lua
```diff
@@ -0,0 +1,139 @@
...
+
+    local data, err = core.json.decode(res)
+    if not data then
+        return nil, "failed to decode result, res: " .. res .. ", err: " .. err
+    end
+
+    return data[sub_key]
+end
+
+_M.get = get
```
membphis, line 134: I think this style is clearer and easier to understand

```lua
function _M.get(conf, key)
  ... ...
end

**Review by membphis — APPROVED** [review]
LGTM, except some minor issues

File: apisix/secret/aws.lua (hunk 25 above)
Comment (bzp2010) [reply apisix/secret/aws.lua:115]: It's wrong, `sub_key and ", sub: " .. sub_key or ""` snippet use specialized scope, we can't rely on the core.log.info concat the strings.

Comment (bzp2010) [reply apisix/secret/aws.lua:115]: @membphis It's wrong, `sub_key and ", sub: " .. sub_key or ""` snippet use specialized scope, we can't rely on the core.log.info concat the strings.

File: apisix/secret/aws.lua (hunk 26 above)
Comment (HuanXin-Chen) [reply apisix/secret/aws.lua:134]: I noticed that vault.lua is also using this style. I believe keeping the original style will help maintain consistency for future expansions.
image

File: apisix/secret/aws.lua [hunk 27]
```diff
@@ -0,0 +1,139 @@
...
+    local session_token = env.fetch_by_uri(conf.session_token) or conf.session_token
+
+    local credentials = aws_instance:Credentials({
+        accessKeyId = access_key_id,
+        secretAccessKey = secret_access_key,
+        sessionToken = session_token,
+    })
+
+    local default_endpoint = "https://secretsmanager." .. region .. ".amazonaws.com"
+    local pre, host, port, _, _ = unpack(http:parse_uri(conf.endpoint_url or default_endpoint))
```
nic-6443, line 71:
```suggestion
    local scheme, host, port, _, _ = unpack(http:parse_uri(conf.endpoint_url or default_endpoint))
```

apisix/secret/aws.lua
```diff
@@ -0,0 +1,139 @@
...
+    local sm = aws_instance:SecretsManager({
+        credentials = credentials,
+        endpoint = endpoint,
+        region = region,
+        port = port,
+    })
+
+    local res, err = sm:getSecretValue({
+        SecretId = key,
+        VersionStage = "AWSCURRENT",
```
nic-6443, line 83: It seems unnecessary to explicitly specify this parameter, because AWS's default behavior is definitely `AWSCURRENT`. There is no reason for it to return a historical version when querying a secret.

apisix/secret/aws.lua
nic-6443, line 115: duplicate with https://github.com/apache/apisix/blob/8ebd7f2ddf121967977aa052770d8da5c14f56ef/apisix/secret/vault.lua#L88-L104
we should reuse code that split the key

HuanXin-Chen: The splitting logic for AWS and Vault is different. AWS can have no sub key.

apisix/secret/aws.lua
HuanXin-Chen: I had thought about it this way before, but later I added it because having an extra field doesn't seem to have much impact. Additionally, if users have specific needs to retrieve previous versions, they can easily extend it themselves. Of course, if this is not elegant, I will modify it later.

apisix/secret/aws.lua
nic-6443: What? Isn't the variable name on line 113 of this file `sub_key`?

apisix/secret/aws.lua
nic-6443: http and https in the URL have standard naming, just like host and port, so there's no need to name it `pre` yourself.

apisix/secret/aws.lua
bzp2010: AWS serect does not force a sub key, it allows to get the whole value because AWS does not force the user to configure a secret in JSON format.

bzp2010: https://github.com/apache/apisix/pull/11417#discussion_r1699842025

HuanXin-Chen: image

nic-6443: got it

apisix/secret/aws.lua
HuanXin-Chen: resolved

nic-6443 (approved):

bzp2010 (dismissed):

moonming (dismissed):

t/secret/aws.t
```diff
@@ -0,0 +1,313 @@
...
+                }]]
+                )
+            if code >= 300 then
+                ngx.status = code
+                return ngx.say(body)
+            end
+
+            local secret = require("apisix.secret")
+            local value = secret.fetch_by_uri("$secret://aws/mysecret/jack/key")
+            ngx.say(value)
```
membphis, line 309: Just confirm, it will print a `nil` value in this test case? 
And it will be considered as expected output, it make me little confused


```lua
if value then
    ngx.say("secret value: ", value)
end

ngx.say("all done")
```

apisix/secret/aws.lua
membphis: if we need to reuse the function `get`, the current style is good.

In this code, we do not need to reuse function `get`, so `_M.get` is better to read and understand

membphis (approved):
LGTM, except minor code style issues

t/secret/aws.t
HuanXin-Chen: fixed

bzp2010 (approved):

moonming (approved):

apisix/secret/aws.lua
bzp2010: fixed

nic-6443 (approved):

membphis (approved):
LGTM

kayx23:
Hi, following the doc to configure this plugin (configured the `region`, `access_key_id`, and `secret_access_key`), I am running into `failed to retrtive data from aws secret manager` `unable to get local issuer certificate` error:

 [lua] secret.lua: fetch(): failed to fetch secret value: failed to retrtive data from aws secret manager: SecretsManager:getSecretValue() failed to connect to 'https://secretsmanager.ap-southeast-2.amazonaws.com:443': 20: unable to get local issuer certificate, client: 192.168.65.1, server: _, request: "GET /anything HTTP/1.1", host: "127.0.0.1:9080"

Any suggestion?

HuanXin-Chen:
> Hi, following the doc to configure this plugin (configured the `region`, `access_key_id`, and `secret_access_key`), I am running into `failed to retrtive data from aws secret manager` `unable to get local issuer certificate` error:
> 
>  [lua] secret.lua: fetch(): failed to fetch secret value: failed to retrtive data from aws secret manager: SecretsManager:getSecretValue() failed to connect to 'https://secretsmanager.ap-southeast-2.amazonaws.com:443': 20: unable to get local issuer certificate, client: 192.168.65.1, server: _, request: "GET /anything HTTP/1.1", host: "127.0.0.1:9080"
> 
> Any suggestion?

You need to install the CA certificate using the command: sudo apt-get install ca-certificates.

And then, make sure that APISIX has configured it with the following setting: ssl_trusted_certificate: /etc/ssl/certs/ca-certificates.crt.

kayx23:
When I used Vault previously I didn't remember having to go through this setup (but again the vault instance is running locally in Docker). What attributed to this differences? 🤔

HuanXin-Chen:
> When I used Vault previously I didn't remember having to go through this setup (but again the vault instance is running locally in Docker). What attributed to this differences? 🤔 

This is because HTTPS is different from HTTP.

PR #11430, opened 2024-07-24
fix: etcd sync data checker should work

ronething:
### Description

Fixes # (issue)

## Background

the `load_full_data` function in the file `config_etcd.lua`(load all exist data from etcd when apisix start), `checker` does not depend on `item_schema` to have a value

https://github.com/apache/apisix/blob/507df1262de88c4ed3ee8a637de46c4f4d2412f8/apisix/core/config_etcd.lua#L540-L546

But in the `sync_data` function (watch new events from etcd when apisix running), `checker` relies on `item_schema` should have value.

https://github.com/apache/apisix/blob/507df1262de88c4ed3ee8a637de46c4f4d2412f8/apisix/core/config_etcd.lua#L668

As a result, the plugin metadata resource cannot reach the checker logic when etcd watch.

https://github.com/apache/apisix/blob/507df1262de88c4ed3ee8a637de46c4f4d2412f8/apisix/plugin.lua#L816-L821

So we need to fix it.

###  about the test case:

The admin api is not used for the operation, because the admin api will directly fail to verify the schema and cannot go to the etcd checker logic.

error-log-logger-clickhouse.t: The request header sent to clickhouse should be decrypted to be correct.

### Checklist

### DIFF
--- a/apisix/plugins/sls-logger.lua
+++ b/apisix/plugins/sls-logger.lua
@@ -63,7 +63,11 @@ local schema = {
 
 local metadata_schema = {
     type = "object",
-    properties = {log_format = log_util.metadata_schema_log_format,},
+    properties = {
+        log_format = {
+            type = "object"
+        }
+    },
 }
 
 local _M = {
--- a/apisix/plugins/sls-logger.lua
+++ b/apisix/plugins/sls-logger.lua
@@ -61,15 +61,24 @@ local schema = {
     required = {"host", "port", "project", "logstore", "access_key_id", "access_key_secret"}
 }
 
+local metadata_schema = {
+    type = "object",
+    properties = {log_format = log_util.metadata_schema_log_format,},
+}
+
 local _M = {
     version = 0.1,
     priority = 406,
     name = plugin_name,
     schema = batch_processor_manager:wrap_schema(schema),
+    metadata_schema = metadata_schema,
 }
 
-function _M.check_schema(conf)
-   return core.schema.check(schema, conf)
+function _M.check_schema(conf,schema_type)
+    if schema_type == core.schema.TYPE_METADATA then
+      return core.schema.check(metadata_schema, conf)
+    end
+    return core.schema.check(schema, conf)
 end
 
 local function send_tcp_data(route_conf, log_message)
--- a/t/admin/plugin-metadata.t
+++ b/t/admin/plugin-metadata.t
@@ -274,7 +274,7 @@ qr/\{"error_msg":"invalid configuration: property \\"ikey\\" is required"\}/
         }
     }
 --- response_body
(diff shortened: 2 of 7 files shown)


moonming (approved):

nic-6443 (approved):

t/plugin/error-log-logger-clickhouse.t
```diff
@@ -98,7 +98,7 @@ done
 --- error_log
 this is a warning message for test2
 clickhouse body: INSERT INTO t FORMAT JSONEachRow
-clickhouse headers: x-clickhouse-key:dpwomMlEsHH2L7wSUi6YiQ==
+clickhouse headers: x-clickhouse-key:a
```
moonming, line 101: why change this?

t/admin/plugin-metadata.t
```diff
@@ -274,7 +274,7 @@ qr/\{"error_msg":"invalid configuration: property \\"ikey\\" is required"\}/
         }
     }
 --- response_body
-{"key":"/apisix/plugin_metadata/example-plugin","value":{"ikey":1,"skey":"val"}}
+{"key":"/apisix/plugin_metadata/example-plugin","value":{"id":"example-plugin","ikey":1,"skey":"val"}}
```
moonming, line 277: why change this?

moonming (changes requested):

apisix/plugins/sls-logger.lua
```diff
@@ -63,7 +63,11 @@ local schema = {
 
 local metadata_schema = {
     type = "object",
-    properties = {log_format = log_util.metadata_schema_log_format,},
+    properties = {
+        log_format = {
+            type = "object"
+        }
+    },
```
moonming, line 70: What does this change have to do with etcd?

t/plugin/error-log-logger-clickhouse.t
ronething: https://github.com/apache/apisix/pull/11076 
The modification of error-log-logger-clickhouse.t in this PR is wrong. The header sent to clickhouse should be decrypted.

t/admin/plugin-metadata.t
ronething: In the admin api, we have added an id field to the plugin metadata, so the response here needs to be added.

apisix/plugins/sls-logger.lua
ronething: Originally sls-logger didn't add the metadata_schema, and then because the etcd checker didn't take effect before, it couldn't detect the problem.
As for the metadata schema, I refer to file-logger.
https://github.com/apache/apisix/blob/2f6fd50121c545742a902cf838287406d103566a/apisix/plugins/file-logger.lua#L63-L70

moonming:
One PR is only for one task.

PR #11445, opened 2024-07-29
feat(cas-auth): support logged-in user header

### LINKED ISSUE #9524: help request: how can i get the user from cas-auth
### Description

 my route use cas-auth plugin， my service how get userinfo。

### Environment

- APISIX version (run `apisix version`): 3.3

nufissime:
### Description

The cas-auth plugin is great, but does not currently support retrieving the logged-in user from upstreams, making it really hard to do any kind of access control appart from "a user is logged-in".

This PR adds the `set_user_header` and `user_header` attributes that enables setting the logged-in user in a header, and allows for the header's name customization.

Some users have already asked for it.

As far as backward compatibility is concerned:  The header is set by default, so one might experience issues if setting the header before reaching apisix... which would make little sense considering apisix handles the authentication.

Fixes #9524.

### Checklist

### DIFF
--- a/apisix/plugins/cas-auth.lua
+++ b/apisix/plugins/cas-auth.lua
@@ -35,8 +35,7 @@ local schema = {
         idp_uri = {type = "string"},
         cas_callback_uri = {type = "string"},
         logout_uri = {type = "string"},
-        set_user_header = {type = "boolean", default = true},
-        user_header = {type = "string", default = "X-Proxy-User"},
+        user_header = {type = "string"},
     },
     required = {
         "idp_uri", "cas_callback_uri", "logout_uri"
@@ -89,7 +88,7 @@ local function with_session_id(conf, ctx, session_id)
     else
         -- refresh the TTL
         store:set(session_id, user, SESSION_LIFETIME)
-        if conf.set_user_header then
+        if conf.user_header then
             core.request.set_header(conf.user_header, user)
         end
     end
--- a/apisix/plugins/cas-auth.lua
+++ b/apisix/plugins/cas-auth.lua
@@ -35,6 +35,8 @@ local schema = {
         idp_uri = {type = "string"},
         cas_callback_uri = {type = "string"},
         logout_uri = {type = "string"},
+        set_user_header = {type = "boolean", default = true},
+        user_header = {type = "string", default = "X-Proxy-User"},
     },
     required = {
         "idp_uri", "cas_callback_uri", "logout_uri"
@@ -87,6 +89,9 @@ local function with_session_id(conf, ctx, session_id)
     else
         -- refresh the TTL
         store:set(session_id, user, SESSION_LIFETIME)
+        if conf.set_user_header then
+            core.request.set_header(conf.user_header, user)
+        end
     end
 end
 
--- a/docs/en/latest/plugins/cas-auth.md
+++ b/docs/en/latest/plugins/cas-auth.md
@@ -35,13 +35,12 @@ to do authentication, from the SP (service provider) perspective.
 
 ## Attributes
 
-| Name               | Type    | Required | Default        | Description                                                                                                           |
-| ------------------ | ------- | -------- | -------------- | --------------------------------------------------------------------------------------------------------------------- |
-| `idp_uri`          | string  | True     |                | URI of IdP.                                                                                                           |
-| `cas_callback_uri` | string  | True     |                | Redirect uri used to callback the SP from IdP after login or logout.                                                  |
-| `logout_uri`       | string  | True     |                | Logout uri to trigger logout.                                                                                         |
-| `set_user_header`  | boolean | False    | true           | When set to true, sets the logged-in user in a request header. By default, the `X-Proxy-User` request header is used. |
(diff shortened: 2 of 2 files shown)


zhoujiexiong:
@lucasarnulphy

Do we need to consider using a general method that can forward more info that returned from CAS to upstream, for extensibility purpose?

And test cases needed? :D

lucasarnulphy:
I think you might be right about returning more info, but I'm not sure it should be exclusive.

Many systems just read a single header with a username and then build all permissions on that. Returning a lot of information in a single header would make that harder to implement.

I do agree it would be great for the upstreams to have more information, maybe as a UserInfo header as it is done in the openid connect plugin, I'll probably implement that next.

As far as tests are concerned... it's only a few lines of codes to return a variable that was already stored and tested before, so I think we're in the clear... but I'll let you be the judge of that.

apisix/plugins/cas-auth.lua
```diff
@@ -87,6 +89,9 @@ local function with_session_id(conf, ctx, session_id)
     else
         -- refresh the TTL
         store:set(session_id, user, SESSION_LIFETIME)
+        if conf.set_user_header then
```
shreemaan-abhishek, line 92: the `set_user_header` field is not needed, this feature can be implemented using the `user_header` field alone. Also please choose a better name to allow any header to be passed.

eg: field name: `send_headers` and the type of this field should be array, to allow multiple headers to be sent.

lucasarnulphy: I'm fine with removing `set_user_header`, and having `user_header` empty by default.

As far as `send_headers` is concerned, I'm not sure I understand what you mean.
Do you mean allowing the plugin to put in a header any value from the cas response ? Because that would require a lot more work as currently only the user is stored.
Or do you mean having a general array which we can search in in the code when other headers are implemented ? because that would remove the ability to overwrite the headers names, which would make it painful to integrate with some apps.

zhoujiexiong: What about the `hash/mapping` way as following:

```
{
    "XPATH_such_as: /auth/user/node": "X-Forwarded-User", // could be the default setting
    "other_xpath_node_to_extract_value": "X-Forwarded-Foo" 
}
```

Some related Lua libs:
- https://github.com/basiliscos/lua-xpath
- https://github.com/bkersten/lua-xpath

zhoujiexiong:
Ping @lucasarnulphy :D

lucasarnulphy:
> Ping @lucasarnulphy :D

I'd be happy to implement that, but I'm waiting for @shreemaan-abhishek to confirm that it would be acceptable, as I have no intention to implement a change that would later be rejected again.

zhoujiexiong:
> > Ping @lucasarnulphy :D
> 
> I'd be happy to implement that, but I'm waiting for @shreemaan-abhishek to confirm that it would be acceptable, as I have no intention to implement a change that would later be rejected again.

@shreemaan-abhishek We need you. :-)

apisix/plugins/cas-auth.lua
shreemaan-abhishek: > Because that would require a lot more work as currently only the user is stored.

Why would it require a lot more work? I don't understand you. If any of the headers configured in `send_headers` is not present in the request then the plugin won't add it. Isn't this simple?

lucasarnulphy: @shreemaan-abhishek I still don't understand what you mean.

For this part:

> the set_user_header field is not needed, this feature can be implemented using the user_header field alone. Also please choose a better name to allow any header to be passed.

I agree.

But for this part:

> Why would it require a lot more work? I don't understand you. If any of the headers configured in send_headers is not present in the request then the plugin won't add it. Isn't this simple?

I have no understanding of what you mean. If send_headers is not present in which request ? The CAS response ? 

I feel like we're not talking about the same thing. The goal of this change is to allow an upstream service to retrieve a value already stored: the username of the authenticated user.

Yes it could be made more generic, by storing more values from the cas response body:

> ngx_re_match(res.body, "<cas:user>(.*?)</cas:user>", "jo");

But it's currently an issue for many users, and it can be improved afterwards to allow for any value for the response.

Other authentication plugins on Apisix have a user_header parameter to pass the authenticated user to upstream services, why does this have to be different ?

shreemaan-abhishek: > Other authentication plugins on Apisix have a user_header parameter to pass the authenticated user to upstream services, why does this have to be different ?

can you show an example?

shreemaan-abhishek: never mind, I understood what you are trying to do.

> I'm fine with removing set_user_header, and having user_header empty by default.

please push this change then we can resolve this conversation.

lucasarnulphy: @shreemaan-abhishek done

shreemaan-abhishek (commented):
could you please add test cases as well? Thanks for your cooperation.

nufissime:
> could you please add test cases as well? Thanks for your cooperation.

I'd dont really see what kind of tests you'd like me to add.

Issue #11216, opened 2024-04-30
bug: the plugin metadata obtained from etcd has not been validated by schema


nic-6443:
### Current Behavior

# Part 1
We created an etcd client to handle plugin metadata in:
https://github.com/apache/apisix/blob/64b81c48ed3bf4ed1bf5b2edbd4bcdea5137123c/apisix/plugin.lua#L816-L821
in `check_plugin_metadata` function, we need a `item.id`(expectation is the plugin name) to check metadata schema of specific plugin.
https://github.com/apache/apisix/blob/64b81c48ed3bf4ed1bf5b2edbd4bcdea5137123c/apisix/plugin.lua#L1040-L1042
but plugin metadata don't have a id field in our admin api, so when apisix startup with a exist plugin metadata, the checker will be triggered in `config_etcd:load_full_data`
https://github.com/apache/apisix/blob/64b81c48ed3bf4ed1bf5b2edbd4bcdea5137123c/apisix/core/config_etcd.lua#L489-L495
and then checker failed with this log:
<img width="1270" alt="image" src="https://github.com/apache/apisix/assets/22141303/b82f1352-4c73-4967-b17c-dddc6b5da61d">

# Part 2
When watch a new data from etcd, `config_etcd.lua` only check configuration that has configure with a `item_schema`
https://github.com/apache/apisix/blob/64b81c48ed3bf4ed1bf5b2edbd4bcdea5137123c/apisix/core/config_etcd.lua#L667-L685
because `plugin_metadata` only set `checker` field without `item_schema`, so the new plugin metadata watch from etcd don't validate by schema too.

### Steps to Reproduce

/

### Environment

/

PR #11452, opened 2024-07-31
fix: plugin metadata add id value for etcd checker

ronething:
### Description

Fixes https://github.com/apache/apisix/issues/11216

fix part 1 of this issue by adding an id field to the plugin metadata with the value of the plugin name

This PR is a partial fork of https://github.com/apache/apisix/pull/11430.

### Checklist

### DIFF
--- a/apisix/plugins/sls-logger.lua
+++ b/apisix/plugins/sls-logger.lua
@@ -61,15 +61,28 @@ local schema = {
     required = {"host", "port", "project", "logstore", "access_key_id", "access_key_secret"}
 }
 
+local metadata_schema = {
+    type = "object",
+    properties = {
+        log_format = {
+            type = "object"
+        }
+    },
+}
+
 local _M = {
     version = 0.1,
     priority = 406,
     name = plugin_name,
     schema = batch_processor_manager:wrap_schema(schema),
+    metadata_schema = metadata_schema,
 }
 
-function _M.check_schema(conf)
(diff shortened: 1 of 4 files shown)


apisix/plugins/sls-logger.lua
```diff
@@ -61,15 +61,28 @@ local schema = {
     required = {"host", "port", "project", "logstore", "access_key_id", "access_key_secret"}
 }
 
+local metadata_schema = {
```
AlinsRan, line 64: What is the relationship between this PR and sls-logger?
Will this fix cause the sls-logger ci-test to fail?

moonming (approved):

apisix/plugins/sls-logger.lua
ronething: Yes, sls-logger doesn't have a metadata schema, but the test case does.
https://github.com/apache/apisix/blob/07c4aa3a4287b4a829b76ae529705b51de56f8aa/t/plugin/sls-logger.t#L209-L222
because it doesn't go to the actual calibration logic so it doesn't report an error.

After adding the id, the check here will fail https://github.com/apache/apisix/blob/6edd86313b800e79cb53a8bb10c3908a7f5348fe/apisix/plugin.lua#L1041-L1042

AlinsRan (approved):

nic-6443 (approved):

PR #11450, opened 2024-07-30
fix: httpc connection close

madhawa-gunasekara:
### Description

This will close httpc connection. We observed too many connections in activemq.

Fixes #11449 

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #11499, opened 2024-08-15
feat: ai-proxy plugin

shreemaan-abhishek:
### Description

Implement the AI-Proxy Plugin.

### Checklist


apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,103 @@
...
+    schema = schema,
+}
+
+
+function _M.check_schema(conf)
+    return core.schema.check(schema.plugin_schema, conf)
+end
+
+
+local CONTENT_TYPE_JSON = "application/json"
```
bzp2010, line 43: This may not be possible to assert simply as a static string, as the content-type may have the following format

```http
Content-Type: application/json;charset=utf8
```

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,103 @@
...
+
+
+local function get_request_table()
+    local req_body, err = core.request.get_body()
+    if not req_body then
+        return nil, "failed to get request body: " .. (err or "request body is empty")
+    end
+    req_body, err = req_body:gsub("\\\"", "\"") -- remove escaping in JSON
```
membphis, line 47: it seems unsafe. I do not think we need to call this step

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,103 @@
...
+    if conf.model.options and conf.model.options.response_streaming then
+        request_table.stream = true
+        ngx.ctx.disable_proxy_buffering = true
+    end
+
+    if conf.model.name then
+        request_table.model = conf.model.name
+    end
+
+    local ai_driver = require("apisix.plugins.ai-proxy.drivers." .. conf.model.provider)
```
membphis, line 86: it may fail, we need to check the return value

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,67 @@
...
+        args[conf.auth.name] = conf.auth.value
+        core.request.set_uri_args(ctx, args)
+    end
+
+    if conf.model.options then
+        for opt, val in pairs(conf.model.options) do
+            request_table[opt] = val
+        end
+    end
+    return true, nil
```
membphis, line 64: remove `, nil`

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,103 @@
...
+        req_schema = schema.chat_completion_request_schema
+    end
+    local ok, err = core.schema.check(req_schema, request_table)
+    if not ok then
+        return 400, "request format doesn't match schema: " .. err
+    end
+
+    if conf.model.options and conf.model.options.response_streaming then
+        request_table.stream = true
+        ngx.ctx.disable_proxy_buffering = true
```
membphis, line 79: why do we need this line? `ngx.ctx.disable_proxy_buffering = true`

apisix/init.lua
```diff
@@ -893,7 +906,16 @@ function _M.http_balancer_phase()
...
 
-    load_balancer.run(api_ctx.matched_route, api_ctx, common_phase)
+    if api_ctx.custom_upstream_ip then
+        local ok, err = balancer.set_current_peer(api_ctx.custom_upstream_ip,
+                                                  api_ctx.custom_upstream_port)
+        if not ok then
+            core.log.error("failed to overwrite upstream for ai_proxy: ", err)
+            return core.response.exit(500)
+        end
+    else
```
bzp2010, line 916: Why not use the `upstream.set` API to modify the upstream endpoints instead of doing this? Is this mandatory?

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,67 @@
...
+local test_scheme = os.getenv("AI_PROXY_TEST_SCHEME")
+local ngx = ngx
+local pairs = pairs
+
+-- globals
+local DEFAULT_HOST = "api.openai.com"
+local DEFAULT_PORT = 443
+
+local path_mapper = {
+    ["llm/completions"] = "/v1/completions",
```
bzp2010, line 29: Why we need to support the v1 API marked as `Legacy`?

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: json decode will fail if we don't do this.

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: when `response_streaming` is configured. Upstream will send response via SSE, proxy_buffering should be disabled for smooth functioning of SSE.

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,67 @@
...
+    ctx.custom_upstream_port = conf.model.options.upstream_port or DEFAULT_PORT
+
+    local ups_path = (conf.model.options and conf.model.options.upstream_path)
+                        or path_mapper[conf.route_type]
+    ngx.var.upstream_uri = ups_path
+    ngx.var.upstream_scheme = test_scheme or "https"
+    ngx.req.set_method(ngx.HTTP_POST)
+    ngx.var.upstream_host = conf.model.options.upstream_host or DEFAULT_HOST
+    ctx.custom_balancer_host = conf.model.options.upstream_host or DEFAULT_HOST
+    ctx.custom_balancer_port = conf.model.options.port or DEFAULT_PORT
```
bzp2010, line 50: Should we use the same `upstream.set` API as `traffic-split` to implement this?

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+}
+
+local model_schema = {
+    type = "object",
+    properties = {
+        provider = {
+            type = "string",
+            description = "AI provider request format - kapisix translates "
```
bzp2010, line 104: What is the `kapisix`?

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,67 @@
...
+    if conf.auth.source == "header" then
```
bzp2010, line 51: What do you mean `source`? This doesn't seem to mean that we will get a value from the header of the client's request, but only use the configuration, so I don't think the naming of the source makes sense.

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: done, I added the check in `check_schema`

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,167 @@
...
+                    },
+                    content = {
+                        type = "string",
+                        minLength = "1",
+                    },
+                },
+                additionalProperties = false,
+                required = {"role", "content"},
+            },
+        }
```
bzp2010, line 151: What design specifications do they follow? for OpenAI?
Does this mean that every new provider in the future will use this format?

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,167 @@
...
+    },
+    required = {"provider", "name"}
+}
+
+_M.plugin_schema = {
+    type = "object",
+    properties = {
+        route_type = {
+            type = "string",
+            enum = { "llm/chat", "llm/completions", "passthrough" }
```
bzp2010, line 123: I don't see the need to support the old and inflexible `llm/completions` API.

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,167 @@
...
+            description = "Full auth-header/param value.",
+            minLength = 1,
+             -- TODO encrypted = true,
+        },
+    },
+    required = { "source", "name", "value" },
+    additionalProperties = false,
+}
+
+local model_options_schema = {
```
bzp2010, line 42: Is it compatible with other potential AI Providers?

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,167 @@
...
+        },
+        upstream_path = {
+            type = "string",
+            description = "To be specified to override the URL to the AI provider endpoints",
+        },
+        response_streaming = {
+            description = "Stream response by SSE",
+            type = "boolean",
+            default = false,
+        }
```
bzp2010, line 95: We'd better move the configuration about the HTTP request itself like upstream or response to a higher level.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,107 @@
...
+
+local CONTENT_TYPE_JSON = "application/json"
+
+
+local function get_request_table()
+    local req_body, err = core.request.get_body()
+    if not req_body then
+        return nil, "failed to get request body: " .. (err or "request body is empty")
+    end
+    req_body, err = req_body:gsub("\\\"", "\"") -- remove escaping in JSON
```
bzp2010, line 51: Please provide an example to explain the scenario specifically, otherwise why would we want to do such a hack on a string that is indeed JSON.

bzp2010, line 51: Please provide an example to explain the scenario specifically, otherwise why would we want to do such a hack on a string that is indeed JSON.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,107 @@
...
+        local final_body, err = core.json.encode(request_table)
+        if not final_body then
+            core.log.error("failed to encode request body to JSON: ", err)
+            return 500
+        end
+        ngx_req.set_body_data(final_body)
+    end
+end
+
+return _M
```
bzp2010, line 107: I didn't notice that we used any sub-requests, in the future, how will we post-process the AI API response?

apisix/plugins/ai-proxy/drivers/openai.lua
bzp2010, line 50: This should be implemented using the `upstream.set` API, just like the traffic-split plugin.

apisix/init.lua
shreemaan-abhishek: done

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: The current spec is for OpenAI, new providers can have different format which can be specified when they are being implemented.

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: Yea I will remove it since the chat API can do what the completion api does.

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: yes, later other AI providers' options can be specified here too.

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: ok

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: https://github.com/apache/apisix/pull/11499#discussion_r1726422581

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: I realised that we don't need to modify OpenAI's response body.

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: it was a mistake sorry.

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: renamed it to `type`

apisix/plugins/ai-proxy.lua
bzp2010: Will this make us have to change a lot of code? You can envision that, because it may be on the roadmap and implemented in the future, and if we think about it only by the time we start doing it, chances are we'll need to do more.

shreemaan-abhishek: no, the subrequest part is a separate piece of independent code.

apisix/init.lua
```diff
@@ -722,7 +722,19 @@ function _M.http_access_phase()
         plugin.run_plugin("access", plugins, api_ctx)
     end
 
-    _M.handle_upstream(api_ctx, route, enable_websocket)
+    if not api_ctx.custom_upstream_ip then
+        _M.handle_upstream(api_ctx, route, enable_websocket)
+    end
```
bzp2010, line 727: There is no longer any place to use `ctx.custom_upstream_ip`, this should be modified.

apisix/plugins/traffic-split.lua
```diff
@@ -18,12 +18,9 @@ local core       = require("apisix.core")
 local upstream   = require("apisix.upstream")
 local schema_def = require("apisix.schema_def")
 local roundrobin = require("resty.roundrobin")
-local ipmatcher  = require("resty.ipmatcher")
```
bzp2010, line 21: We'd prefer not to modify it in this PR, you can indeed create reusable code snippets first, but please modify the call in traffic-split in another PR.

apisix/plugins/ai-proxy/schema.lua
bzp2010: So each model in each provider may have different options? 🤔

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,92 @@
...
+
+local CONTENT_TYPE_JSON = "application/json"
+
+
+function _M.access(conf, ctx)
+    local route_type = conf.route_type
+    ctx.ai_proxy = {}
+
+    local ct = core.request.header(ctx, "Content-Type") or CONTENT_TYPE_JSON
+    if ct ~= CONTENT_TYPE_JSON and ct ~= CONTENT_TYPE_JSON .. ";charset=utf8" then
```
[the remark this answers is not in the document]
shreemaan-abhishek: oops, dumb me 😂
I'll fix it.

apisix/init.lua
shreemaan-abhishek: done

shreemaan-abhishek:
> So each model in each provider may have different options? 🤔

yes

apisix/cli/ngx_tpl.lua
```diff
@@ -809,6 +809,64 @@ http {
             }
         }
 
+        location @disable_proxy_buffering {
```
bzp2010, line 812: If the HTTP client way works, is it no longer needed?

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,83 @@
...
+    local httpc, err = http.new()
+    if not httpc then
+        return nil, "failed to create http client to send request to LLM server: " .. err
+    end
+    httpc:set_timeout(conf.timeout)
+
+    local custom_host = core.table.try_read_attr(conf, "override", "host")
+    local custom_port = core.table.try_read_attr(conf, "override", "port")
+    local custom_path = core.table.try_read_attr(conf, "override", "path")
+    local custom_scheme = core.table.try_read_attr(conf, "override", "scheme")
```
bzp2010, line 39: Would it be better to just use a combined ENDPOINT configuration? 
I.e. `override.endpoint = "https://api.openai.com:443/api-balabala/"`

In this way schema, host, and port can be combined into a single configuration. If we do need such a separate section later, APISIX can parse it internally instead of asking the user to configure it one by one.

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,83 @@
...
+
+    local ok, err = httpc:connect({
+        scheme = custom_scheme or "https",
+        host = custom_host or DEFAULT_HOST,
+        port = custom_port or DEFAULT_PORT,
+        ssl_verify = conf.ssl_verify,
+        ssl_server_name = custom_host or DEFAULT_HOST,
```
bzp2010, line 46: Is it required, and should lua-resty-http automatically infer it?

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,83 @@
...
+            ["Content-Type"] = "application/json",
+        },
+        keepalive = conf.keepalive,
+        ssl_verify = conf.ssl_verify,
+        path = custom_path or "/v1/chat/completions",
+    }
+
+    if conf.auth.type == "header" then
+        params.headers[conf.auth.name] = conf.auth.value
+    end
```
bzp2010, line 66: I'm wondering if we need to provide specific authentication configuration items, which are not easily scalable, e.g. if OpenAI requires both headers, apikey and organization id, then the current mechanism would need major modifications.

Could we expose a raw headers configuration item that would allow the user to directly specify the header that should be attached to the request as a JSON object. This is more flexible and allows us to not specify too much schema.

I.e.

```json
{
  "auth": {
    "args": {},
    "headers": {
      "apikey": "apikey",
      "org-id": "org-id"
    }
  }
}
```

Is this more favorable to users and what do you think?

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,127 @@
...
+                break
+            end
+            if not chunk then
+                break
+            end
+            content_length = content_length + #chunk
+            ngx.print(chunk)
+            ngx.flush(true)
+        end
+        httpc:set_keepalive(10000, 100)
```
bzp2010, line 79: Is this because we have to manually put the connection back into the connection pool at the end of the entire SSE response?
What could be the problem if we don't do that?

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,127 @@
...
+                core.log.error("failed to read response chunk: ", err)
+                break
+            end
+            if not chunk then
+                break
+            end
+            content_length = content_length + #chunk
+            ngx.print(chunk)
+            ngx.flush(true)
+        end
```
bzp2010, line 78: https://github.com/ledgetech/lua-resty-http/blob/master/lib/resty/http.lua#L564-L589

The logic is similar to the above link, except we add print and flush. Can we copy the above verified code and just add our own print/flush logic.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,127 @@
...
+    end
+
+    return send_request(conf, ctx)
+end
+
+
+function _M.delayed_access(conf, ctx)
+    return send_request(conf, ctx)
+end
+
```
bzp2010, line 126: What is this? 🤔

apisix/constants.lua
```diff
@@ -42,5 +42,5 @@ return {
         ["/ssls"] = true,
         ["/stream_routes"] = true,
         ["/plugin_metadata"] = true,
-    },
+    }
```
bzp2010, line 45: no changes required

apisix/init.lua
```diff
@@ -722,10 +722,25 @@ function _M.http_access_phase()
...
 end
 
 
+function _M.disable_proxy_buffering_access_phase()
+    ngx.ctx = fetch_ctx()
+    local api_ctx = ngx.ctx.api_ctx
+    local plugins = plugin.filter(api_ctx, api_ctx.matched_route)
+
+    -- plugins to be run after proxy_buffering is disabled
+    plugin.run_plugin("delayed_access", plugins, api_ctx)
```
bzp2010, line 740: What is its necessity?

apisix/upstream.lua
```diff
@@ -20,11 +20,13 @@ local discovery = require("apisix.discovery.init").discovery
 local upstream_util = require("apisix.utils.upstream")
 local apisix_ssl = require("apisix.ssl")
 local events = require("apisix.events")
+
 local error = error
 local tostring = tostring
 local ipairs = ipairs
 local pairs = pairs
 local pcall = pcall
+local table_insert = table.insert
```
bzp2010, line 29: no changes required

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,170 @@
...
+            maximum = 1,
+
+        },
+        stream = {
+            description = "Stream response by SSE",
+            type = "boolean",
+            default = false,
+        }
+    }
+}
```
bzp2010, line 84: Should the configurations related to the LLM provider be moved to `openai.lua` or are they the same on every LLM provider?

apisix/core/request.lua
```diff
@@ -334,6 +335,26 @@ function _M.get_body(max_size, ctx)
 end
 
 
+function _M.get_request_body_table()
+    local body, err = _M.get_body()
+    if not body then
+        return nil, { message = "could not get body: " .. (err or "request body is empty") }
+    end
+
+    body, err = body:gsub("\\\"", "\"") -- remove escaping in JSON
```
bzp2010, line 344: I still don't understand the need for it. Can you provide a specific example? For example, curl or postman?

bzp2010 (commented):
The code has been checked and the problem has been significantly improved and we are on the right path. 🫡

apisix/cli/ngx_tpl.lua
shreemaan-abhishek: if proxy buffering is enabled, openresty would still wait for buffers to be full before sending response to client, isn't it?

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: it is a general practice to use keepalive isn't it?

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: if streaming is enabled then we set `ctx.disable_proxy_buffering` in access_phase but proxy_buffering will be disabled only after the end of the `access_phase` hence a new dummy phase `delayed_access` phase is created that will run after `proxy_buffering` is disabled.

apisix/init.lua
shreemaan-abhishek: https://github.com/apache/apisix/pull/11499#discussion_r1743429703

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: most LLMs have similar options that's why I kept them here.

apisix/core/request.lua
shreemaan-abhishek: hmm now I can no longer reproduce this issue. Maybe I was doing something wrong before.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,127 @@
...
+local function send_request(conf, ctx)
+    local request_table = ctx.request_table
+    local ai_driver = require("apisix.plugins.ai-proxy.drivers." .. conf.model.provider)
+    local res, err, httpc = ai_driver.request(conf, request_table, ctx)
+    if not res then
+        core.log.error("failed to send request to AI service: ", err)
+        return 500
+    end
+
+    if conf.passthrough then
```
zhoujiexiong, line 54:
```suggestion
    if conf.passthrough or not core.table.try_read_attr(conf, "model", "options", "stream") then
```

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,83 @@
...
+
+    if conf.model.options then
+        for opt, val in pairs(conf.model.options) do
+            request_table[opt] = val
+        end
+    end
+    params.body = core.json.encode(request_table)
+
+    local res, err = httpc:request(params)
```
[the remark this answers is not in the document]
shreemaan-abhishek: `request_uri` doesn't allow streaming responses.

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: I agree, I will make these changes.

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: we should provide is compulsorily.

apisix/plugins/ai-proxy/drivers/openai.lua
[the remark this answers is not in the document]
zhoujiexiong: > Replace with httpc:request_uri to avoid invoking close or keepalive elsewhere?
> 
> #11450

You need to support streaming chunk outside, so `httpc:request` here.
But remember httpc:close when conf.keepalive is false.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,127 @@
...
+    if not ok then
+        return 400, "request format doesn't match schema: " .. err
+    end
+
+    if conf.model.name then
+        request_table.model = conf.model.name
+    end
+    ctx.request_table = request_table
+
+    if conf.model.options and conf.model.options.stream then
```
zhoujiexiong, line 113:
```suggestion
    if core.table.try_read_attr(conf, "model", "options", "stream") then
```

apisix/init.lua
zhoujiexiong, line 740:
```suggestion
    plugin.run_plugin("disable_proxy_buffering_access_phase", plugins, api_ctx)
```

apisix/cli/ngx_tpl.lua
zhoujiexiong: confirm that `ngx.flush(true)` is already satisfy for streaming output

```
$ restydoc -s ngx.flush

When "ngx.flush(true)" is called immediately after ngx.print or
       ngx.say, it causes the latter functions to run in synchronous mode.
       This can be particularly useful for streaming output.
```

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: you are right

apisix/cli/ngx_tpl.lua
shreemaan-abhishek: true but this is what I meant: https://docs.api7.ai/hub/proxy-buffering#optional-see-buffering-in-effect

apisix/init.lua
shreemaan-abhishek: done

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: now we should not pass the response from LLM to upstream if stream == true.

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: done

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: what do you mean? Right now, the logic is already same as you said.

apisix/cli/ngx_tpl.lua
bzp2010: No, it's probably a misunderstanding, when I reviewed you hadn't deleted the code for the subrequest, but on commit it had, so GitHub recorded the wrong comment location.
If the subrequests have been cleaned up correctly, I don't see a problem.

apisix/plugins/ai-proxy/drivers/openai.lua
bzp2010: I'm not aware of any other APISIX calls passing that parameter, I mean `ssl_server_name`, so I suspect it should be automatically inferred from the host, but no matter, it doesn't mess up the functionality and provides greater certainty, I'll take it.

apisix/plugins/ai-proxy/drivers/openai.lua
bzp2010: So should it still be kept?

```
if conf.auth.type == "header" then
  params.headers[conf.auth.name] = conf.auth.value
end
```

apisix/plugins/ai-proxy.lua
bzp2010: Since I use `request_uri` usually and it handles this automatically, I'd like to ask, for `request` it's manual right?

apisix/plugins/ai-proxy/schema.lua
bzp2010: cool

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: yes.

apisix/cli/ngx_tpl.lua
shreemaan-abhishek: conclusion: the subrequest block should not be removed now?

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: whoops, must be removed. Thanks for the good observation, it has been removed now.

apisix/cli/ngx_tpl.lua
bzp2010: It should.

apisix/upstream.lua
bzp2010: `upstream.lua` also contains invalid changes, if we remove it we will simplify the PR.

https://github.com/apache/apisix/pull/11499/files#diff-9af33577e041ece8fef32d8f8193f46eac83dab9e947a2cc940f36b83d5150e8R23

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,88 @@
...
+        ssl_verify = conf.ssl_verify,
+        ssl_server_name = parsed_url.host or DEFAULT_HOST,
+        pool_size = conf.keepalive and conf.keepalive_pool,
+    })
+
+    if not ok then
+        return nil, "failed to connect to LLM server: " .. err
+    end
+
+    local query_params = core.utils.table_to_query_params(conf.auth.query)
```
bzp2010, line 56: I think `ngx.encode_args` should be exactly what you want?
https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#ngxencode_args

bzp2010: If so, we can remove the custom function.

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -0,0 +1,153 @@
...
+        ["^[a-zA-Z0-9._-]+$"] = {
+            type = "string"
+        }
+    }
+}
+
+local auth_schema = {
+    type = "object",
+    patternProperties = {
+        ["^(query|header)$"] = auth_item_schema
```
bzp2010, line 31: I guess this could be written in a regular flattened way? Given the different state of implementation of json schema in various programming languages, I would suggest that we write it in the most common way rather than any special syntax.

apisix/cli/ngx_tpl.lua
zhoujiexiong: According to research, `proxy_http_version 1.1;` is SSE friendly, which is the default setting of APISIX.

So `proxy_buffering no;` seems unnecessary.

https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: done

apisix/plugins/ai-proxy/drivers/openai.lua
shreemaan-abhishek: done

apisix/upstream.lua
shreemaan-abhishek: whoops, fixed it.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,128 @@
...
+        end
+        httpc:set_keepalive(10000, 100)
+        return
+    else
+        local res_body, err = res:read_body()
+        if not res_body then
+            core.log.error("failed to read response body: ", err)
+            return 500
+        end
+        return res.status, res_body
```
zhoujiexiong, line 88:
```suggestion
        httpc:set_keepalive(10000, 100)
        return res.status, res_body
```

also `keepalive` before returning?

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,128 @@
...
+        return 500
+    end
+
+    if conf.passthrough then
+        local res_body, err = res:read_body()
+        if not res_body then
+            core.log.error("failed to read response body: ", err)
+            return 500
+        end
+        ngx_req.set_body_data(res_body)
```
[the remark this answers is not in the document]
shreemaan-abhishek: wow I didn't know these APIs existed. Thanks a lot 🧠

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: yes, thank you.

apisix/plugins/ai-proxy.lua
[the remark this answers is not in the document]
shreemaan-abhishek: but we need to send response back to the client, and I'm not sure how to use the ngx.req APIs to send response.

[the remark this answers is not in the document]
zhoujiexiong: `send response back to the client` is the responsibility of Ngx.

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,136 @@
...
+            if not chunk then
+                break
+            end
+            ngx_print(chunk)
+            ngx_flush(true)
+        end
+        keepalive_or_close(conf, httpc)
+        return
+    else
+        local res_body, err = res:read_body()
```
zhoujiexiong, line 126: also using body_reader, ngx_print * N + ngx_flush * 1?

shreemaan-abhishek: I don't understand what you mean.

zhoujiexiong: The purpose is to save the intermediate buffering process and directly
write the response body to the response buffer by calling ngx.print
multiple times. Finally, call ngx.flush once more.

But,

$ restydoc -s ngx.print

> Please note that both "ngx.print" and ngx.say will always invoke the
> whole Nginx output body filter chain, which is an expensive operation.
> So be careful when calling either of these two in a tight loop; buffer
> the data yourself in Lua and save the calls.

shreemaan-abhishek: got it, I think we don't need ` ngx_print * N + ngx_flush * 1` as reading the response body all at once is favourable for `"stream": false`

apisix/init.lua
```diff
@@ -726,6 +726,7 @@ function _M.http_access_phase()
 end
 
 
+
```
bzp2010, line 729: If there are no substantive changes, please do not change the code style.

t/plugin/ai-proxy.t
```diff
@@ -0,0 +1,694 @@
...
+                    if header_auth ~= "Bearer token" and query_auth ~= "apikey" then
+                        ngx.status = 401
+                        ngx.say("Unauthorized")
+                        return
+                    end
+
+                    if header_auth == "Bearer token" or query_auth == "apikey" then
+                        ngx.req.read_body()
+                        local body, err = ngx.req.get_body_data()
+                        local esc = body:gsub('"\\\""', '\"')
```
bzp2010, line 105: Is this still necessary? Just a confirmation that if it is then no changes are needed.

docs/en/latest/plugins/ai-proxy.md
```diff
@@ -0,0 +1,144 @@
...
+      "nodes": {
+        "somerandom.com:443": 1
+      },
+      "scheme": "https",
+      "pass_host": "node"
+    }
+  }'
+```
+
+Since `passthrough` is not enabled upstream node can be any arbitrary value because it won't be contacted.
```
kayx23, line 110: 1. design question: why do we need the parameter `passthrough`? is it meant to be a toggle?
2. is this sentence needed here?

bzp2010 (commented):
no more comment now ✨

docs/en/latest/plugins/ai-proxy.md
```diff
@@ -0,0 +1,144 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Description
+
+The `ai-proxy` plugin simplifies access to AI providers and models by defining a standard request format
+that allows configuring key fields in plugin configuration to embed into the request.
```
kayx23, line 33: do you mean

```suggestion
that allows key fields in plugin configuration to be embedded into the request.
```

docs/en/latest/plugins/ai-proxy.md
```diff
@@ -0,0 +1,144 @@
...
+
+Proxying requests to OpenAI is supported for now, other AI models will be supported soon.
```
kayx23, line 35:
```suggestion
Proxying requests to OpenAI is supported now. Other LLM services will be supported soon.
```

svc providers != models
LLM models != AI models  (just a subset)

docs/en/latest/plugins/ai-proxy.md
```diff
@@ -0,0 +1,144 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+## Description
+
+The `ai-proxy` plugin simplifies access to AI providers and models by defining a standard request format
```
kayx23, line 32:
```suggestion
The `ai-proxy` plugin simplifies access to LLM providers and models by defining a standard request format
```

docs/en/latest/config.json
```diff
@@ -96,7 +96,8 @@
             "plugins/fault-injection",
             "plugins/mocking",
             "plugins/degraphql",
-            "plugins/body-transformer"
+            "plugins/body-transformer",
+            "plugins/ai-proxy"
```
kayx23, line 100: The recently added `ai-*` plugins should probably all go into a new category. Can be amended in a different PR.

docs/en/latest/plugins/ai-proxy.md
shreemaan-abhishek: > design question: why do we need the parameter passthrough? is it meant to be a toggle?

yes, kinda. when it's enabled, the response from LLM will be set as the request body to the upstream.

> is this sentence needed here?

I think it's important to let users know, instead to let them be puzzled about it and then figure out later.

apisix/core/request.lua
```diff
@@ -334,6 +335,21 @@ function _M.get_body(max_size, ctx)
 end
 
 
+function _M.get_request_body_table()
```
zhoujiexiong, line 338: Decode body according content-type or rename the method, e.g. get_json_request_body_table?

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,136 @@
...
+end
+
+
+function _M.access(conf, ctx)
+    local ct = core.request.header(ctx, "Content-Type") or CONTENT_TYPE_JSON
+    if not core.string.has_prefix(ct, CONTENT_TYPE_JSON) then
+        return 400, "unsupported content-type: " .. ct
+    end
+
+    local request_table, err = core.request.get_request_body_table()
```
zhoujiexiong, line 61: new naming style?

request_table, count_num, flag_boolean?

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,136 @@
...
+                break
+            end
+            ngx_req.append_body(chunk)
+        end
+        ngx_req.finish_body()
+        keepalive_or_close(conf, httpc)
+        return
+    end
+
+    if core.table.try_read_attr(conf, "model", "options", "stream") then
```
zhoujiexiong, line 110:
```suggestion
    if request_table.stream then
```

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,136 @@
...
+        request_table.stream = true
+    end
+
+    local ai_driver = require("apisix.plugins.ai-proxy.drivers." .. conf.model.provider)
+    local res, err, httpc = ai_driver.request(conf, request_table, ctx)
+    if not res then
+        core.log.error("failed to send request to AI service: ", err)
+        return 500
```
zhoujiexiong, line 83: use the predefined Ngx constants?

> ngx.HTTP_NOT_ACCEPTABLE = 406
> ngx.HTTP_REQUEST_TIMEOUT = 408
> ngx.HTTP_CONFLICT = 409
> ngx.HTTP_GONE = 410
> ngx.HTTP_UPGRADE_REQUIRED = 426
> ngx.HTTP_TOO_MANY_REQUESTS = 429
> ngx.HTTP_CLOSE = 444
> ngx.HTTP_ILLEGAL = 451
> ngx.HTTP_INTERNAL_SERVER_ERROR = 500
> ngx.HTTP_METHOD_NOT_IMPLEMENTED = 501
> ngx.HTTP_BAD_GATEWAY = 502
> ngx.HTTP_SERVICE_UNAVAILABLE = 503
> ngx.HTTP_GATEWAY_TIMEOUT = 504
> ngx.HTTP_VERSION_NOT_SUPPORTED = 505
> ngx.HTTP_INSUFFICIENT_STORAGE = 507

apisix/plugins/ai-proxy.lua
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    local body_reader = res.body_reader
+    if not body_reader then
+        core.log.error("LLM sent no response body")
+        return 500
```
zhoujiexiong, line 89: ditto

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,92 @@
...
+
+local core = require("apisix.core")
+local http = require("resty.http")
+local url  = require("socket.url")
+
+local pairs = pairs
+local type  = type
+
+-- globals
+local DEFAULT_HOST = "api.openai.com"
```
zhoujiexiong, line 27: hard code?

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,92 @@
...
+        ssl_server_name = parsed_url.host or DEFAULT_HOST,
+        pool_size = conf.keepalive and conf.keepalive_pool,
+    })
+
+    if not ok then
+        return nil, "failed to connect to LLM server: " .. err
+    end
+
+    local query_params = ""
+    if conf.auth.query and type(conf.auth.query) == "table" then
```
zhoujiexiong, line 58:
```suggestion
    if type(conf.auth.query) == "table" then
```

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,92 @@
...
+        query_params = core.string.encode_args(conf.auth.query)
+        if query_params and query_params ~= "" then
+            query_params = "?" .. query_params
+        end
+    end
+
+    local path = (parsed_url.path or "/v1/chat/completions") ..  query_params
```
zhoujiexiong, line 65:
```suggestion
    local path = (parsed_url.path or DEFAULT_URI) ..  query_params
```

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,92 @@
...
+        pool_size = conf.keepalive and conf.keepalive_pool,
+    })
+
+    if not ok then
+        return nil, "failed to connect to LLM server: " .. err
+    end
+
+    local query_params = ""
+    if conf.auth.query and type(conf.auth.query) == "table" then
+        query_params = core.string.encode_args(conf.auth.query)
```
zhoujiexiong, line 59: FYI, from lua-resty-http, `params.query` already support table type



```
local function _format_request(self, params)
    local version = params.version
    local headers = params.headers or {}

    local query = params.query or ""
    if type(query) == "table" then
        query = "?" .. ngx_encode_args(query)
    elseif query ~= "" and str_sub(query, 1, 1) ~= "?" then
        query = "?" .. query
    end

    -- Initialize request
    local req = {
        str_upper(params.method),
        " ",
        self.path_prefix or "",
        params.path,
        query,
        HTTP[version],
        -- Pre-allocate slots for minimum headers and carriage return.
        "",
        "",
        "",
    }
    local c = 7 -- req table index it's faster to do this inline vs table.insert

    -- Append headers
    for key, values in pairs(headers) do
        key = tostring(key)

        if type(values) == "table" then
            for _, value in pairs(values) do
                req[c] = key .. ": " .. tostring(value) .. "\r\n"
                c = c + 1
            end

        else
            req[c] = key .. ": " .. tostring(values) .. "\r\n"
            c = c + 1
        end
    end

    -- Close headers
    req[c] = "\r\n"

    return tbl_concat(req)
end

```

t/sse_server_example/main.go
```diff
@@ -0,0 +1,59 @@
...
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+
+	// A simple loop that sends a message every 2 seconds
+	for i := 0; i < 5; i++ {
+		// Create a message to send to the client
+		fmt.Fprintf(w, "data: %s\n\n", time.Now().Format(time.RFC3339))
+
+		// Flush the data immediately to the client
+		if f, ok := w.(http.Flusher); ok {
```
zhoujiexiong, line 40: do type assertion outside the loop?   :D

apisix/plugins/ai-proxy.lua
shreemaan-abhishek: I don't understand what you mean.

apisix/plugins/ai-proxy/drivers/openai.lua
```diff
@@ -0,0 +1,86 @@
...
+    if conf.model.options then
+        for opt, val in pairs(conf.model.options) do
+            request_table[opt] = val
+        end
+    end
+    params.body = core.json.encode(request_table)
+
+    local res, err = httpc:request(params)
+    if not res then
+        return 500, "failed to send request to LLM server: " .. err
```
zhoujiexiong, line 80:
```suggestion
        return internal_server_error, "failed to send request to LLM server: " .. err
```

t/sse_server_example/main.go
```diff
@@ -0,0 +1,58 @@
...
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+
+	f, ok := w.(http.Flusher);
+	if !ok {
+		fmt.Fprintf(w, "[ERROR]")
+		return
+	}
+	// A simple loop that sends a message every 2 seconds
```
zhoujiexiong, line 39:
```suggestion
	// A simple loop that sends a message every 500ms
```

Parameterize it?

apisix/plugins/ai-proxy.lua
zhoujiexiong: Although adding a suffix makes it more readable, it is not the style used in the project.

shreemaan-abhishek: oh I got it now 😂  . `request_table` makes it more readable and it can be understood that it's not a string.

apisix/plugins/ai-proxy.lua
bzp2010: Hi @shreemaan-abhishek Is it resolved?

shreemaan-abhishek: yes

bzp2010 (approved):

moonming (approved):

nic-6443 (approved):

PR #11502, opened 2024-08-16
feat: Add the cwt(chain web token) plugin.

yezhenghao:
…ilar to the jwt-auth plugin, which belongs to the permission authentication plugin. The difference is that for cwt plugin, apisix uses the chain wallet address in the consumer whitelist configuration, does not touch the key, the key only belongs to the client user, which is a decentralized mode, while jwt-auth plugin needs to configure the key in the consumer of the apisix to generate the token. It's a centralized model. The cwt plugin can be used in scenarios where mutual trust is required or where it is not convenient to expose the key.

### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/plugins/cwt.lua
+++ b/apisix/plugins/cwt.lua
@@ -1,0 +1,561 @@
+local core              = require("apisix.core")
+local consumer_mod      = require("apisix.consumer")
+local cjson             = require("cjson.safe")
+local resty_string      = require("resty.string")
+local openssl_digest    = require("resty.openssl.digest")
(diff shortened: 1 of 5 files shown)


shreemaan-abhishek:
please create an issue with the proposal describing "why" first, then "how" and then "what".

shreemaan-abhishek:
also, please fix the PR title. It looks horrific right now.

yezhenghao:
> please create an issue with the proposal describing "why" first, then "how" and then "what".

https://github.com/apache/apisix/issues/11542

PR #11570, opened 2024-09-12
fix(ci): foo.com is no longer a bad gateway

shreemaan-abhishek:
### Description

foo.com used in traffic split test is a valid domain now and sending plain http request will yield `302` status code instead of `502`.
https://github.com/apache/apisix/actions/runs/10823712557/job/30029684940?pr=11499#step:21:380

### Checklist

### DIFF
--- a/t/plugin/traffic-split3.t
+++ b/t/plugin/traffic-split3.t
@@ -285,7 +285,6 @@ passed
 === TEST 8: domain name resolved successfully
 --- request
 GET /server_port
---- error_code: 502
 --- error_log eval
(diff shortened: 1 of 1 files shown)


bzp2010:
Can we change that domain to something that can't possibly exist? 
For example using a non-existent TLD like `foo.local`, which would not exist on any well known DNS, or even, we could leave the suffix alone.

bzp2010 (approved):

Revolyssup (approved):

moonming (approved):

nic-6443 (approved):

PR #11581, opened 2024-09-17
feat: rewrite hmac-auth plugin for usability

Revolyssup:
Currently the hmac-auth plugin has a lot of headers that need to be configured, which makes it confusing to use the plugin. This PR also makes it so that it follows the [RFC](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-05).
This PR refactors the plugin as following:
- Use single `Authorization` header instead of multiple headers for ease of use. Now you can provide 4 parameters(key_id,signature, headers,algorithm) in Authorization header separated by comma in any order, this makes it easy to use
- change the name of access_key to key_id
- simplifies the schema of consumer resource and moves most parameters at route/service level.(except key_id and secret_key)
- Use @Request-target as mentioned in RFC in signing string.

@Request-target definition comes from: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-03#name-request-target
Signing string generation logic: https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-00#section-2.1.2


membphis:
bad title, we can not use `backport` in this PR

pls add more description about why we need this PR?

Revolyssup:
> bad title, we can not use `backport` in this PR
> 
> pls add more description about why we need this PR?

done

moonming (dismissed):

nic-6443 (dismissed):

apisix/plugins/hmac-auth.lua
```diff
@@ -148,204 +118,163 @@ function _M.check_schema(conf, schema_type)
...
+            if not canonical_header then
+              if h == "@request-target" then
+                local request_target = request_method .. " " .. uri
+                core.table.insert(signing_string_items, request_target)
+                core.log.info("canonical_header name:", core.json.delay_encode(h))
+                core.log.info("canonical_header value: ",
+                              core.json.delay_encode(request_target))
+              end
+            else
+              core.table.insert(signing_string_items,
```
membphis, line 166: bad alignment

image

apisix/plugins/hmac-auth.lua
```diff
@@ -354,86 +283,62 @@ local function validate(ctx, params)
...
-                      core.json.delay_encode(auth_data))
-
-        if #auth_data == 6 and auth_data[1] == "hmac-auth-v1" then
-            app_key = auth_data[2]
-            signature = auth_data[3]
-            algorithm = auth_data[4]
-            date = auth_data[5]
-            signed_headers = auth_data[6]
-        end
+    if not auth_string:match("^Signature") then
```
membphis, line 293: a high performance way: `core.string.has_prefix`

apisix/plugins/hmac-auth.lua
```diff
@@ -354,86 +283,62 @@ local function validate(ctx, params)
...
+    end
+    local validated_consumer, err = validate(ctx, conf, params)
     if not validated_consumer then
         core.log.warn("client request can't be validated: ", err or "Invalid signature")
         return 401, {message = "client request can't be validated"}
     end
 
     local consumer_conf = consumer.plugin(plugin_name)
     consumer.attach_consumer(ctx, validated_consumer, consumer_conf)
-    core.log.info("hit hmac-auth rewrite")
```
membphis, line 436: we do not need to remove it

membphis (approved):
LGTM, except some minor issues

apisix/plugins/hmac-auth.lua
membphis: https://github.com/apache/apisix/blob/master/apisix/core/string.lua#L62

moonming (approved):

nic-6443 (approved):

membphis (approved):
LGTM

ssquan:
What was the reason for changing the property name from 'access_key' to 'key_id'?
Applications that automatically create consumers will need to update their code when upgrading to the new version of APISIX.

coder2z:
Why not update the documentation?

coder2z:
Why is this completely incompatible update not opening a new plugin? Or compatible with the previous usage method? Have you not considered us users? Do you want us to remain on one version? Not to update?

caibirdme:
Maybe we can rename this newly hmac-auth as hmac-auth-v2? Since hmac-auth is widely used in many different users' prod environment, this breaking change will rise many problems. There're many different consumers that use hmac-auth, it's impossible to have them adopt the new hmac sign algorithm, which means we have to keep the old hmac-auth plugin, thus we only have two choices: donnot upgrade apisix or remove this change in my fork version. Neither is good way.
Since this change isn't really that much important, not related to architecture, just for ease of use, I really suggest that we don't need to bring this new breaking change. Renaming is better

PR #11597, opened 2024-09-23
feat: Jwt-auth plugin no longer requires a private_key to be uploaded.

dspo:
### Description

For security reasons, the API gateway should not accept uploads of users' private keys for issuing jwt tokens, and should no longer provide an API for issuing jwt tokens.
This PR removes the `/apisix/plugin/jwt/sign` API.

### Checklist


apisix/plugins/jwt-auth.lua
```diff
@@ -138,14 +137,10 @@ function _M.check_schema(conf, schema_type)
     end
 
     if conf.algorithm == "RS256" or conf.algorithm == "ES256" then
-        -- Possible options are a) public key is missing
-        -- b) private key is missing
+        -- Possible options is: public key is missing
```
nic-6443, line 140: This comment doesn't seem to make much sense, it feels like it can be removed.

apisix/plugins/jwt-auth.lua
```diff
@@ -235,15 +230,10 @@ local function get_rsa_or_ecdsa_keypair(conf)
...
-        return public_key, private_key
-    elseif public_key and not private_key then
-        return nil, nil, "missing private key"
-    elseif not public_key and private_key then
+    if not public_key then
         return nil, nil, "missing public key"
-    else
-        return nil, nil, "public and private keys are missing"
     end
+    return public_key, private_key
```
nic-6443, line 236: Why is there still a `private_key`?

dspo: To facilitate generating jwt tokens in tests, I added a `_M.gen_token` method. In the test, I can use `local gen_token = require("apisix.plugins.jwt-auth").gen_token`. `gen_token` depends on this function, so there is still a private_key.

nic-6443: I think you need to completely distinguish between verifying the legality of the token in client requests and `gen_token` in tests, otherwise this function makes it very misleading that a `private_key` field is still needed in the `jwt-auth` configuration.

apisix/plugins/jwt-auth.lua
```diff
@@ -387,52 +287,87 @@ function _M.rewrite(conf, ctx)
...
+    end
+    if not auth_conf.algorithm then
+        auth_conf.algorithm = "HS256"
+    end
+    local sign_handler = get_sign_handler(auth_conf.algorithm)
+    local jwt_token, err = sign_handler(auth_conf.key, auth_conf, payload)
+    return jwt_token, err
 end
 
+_M.gen_token = gen_token
```
nic-6443, line 370:
```suggestion
-- only for test
_M.gen_token = gen_token
```

nic-6443 (dismissed):

apisix/plugins/jwt-auth.lua
```diff
@@ -138,14 +137,9 @@ function _M.check_schema(conf, schema_type)
     end
 
     if conf.algorithm == "RS256" or conf.algorithm == "ES256" then
```
nic-6443, line 139: It looks like this validation is redundant with the json schema, see if it can be removed.

membphis:
the `wolf-rbac` is not a popular auth plugin, using it for best practices is not good.

I suggest another two ways:
1. use JWT as before, but we generate the jwt-token by other popular tools
2. change to `basic-auth` which is popular

docs/en/latest/plugin-develop.md
```diff
@@ -439,20 +439,21 @@ end
 
 ## register public API
 
-A plugin can register API which exposes to the public. Take jwt-auth plugin as an example, this plugin registers `GET /apisix/plugin/jwt/sign` to allow client to sign its key:
+A plugin can register API which exposes to the public. Take wolf-rbac plugin as an example, this plugin registers `POST /apisix/plugin/wolf-rbac/login` to allow a client to login and get the wolf rbac_token:
```
membphis, line 442: we can use `batch-request` as new example for developer

apisix/plugins/jwt-auth.lua
```diff
@@ -387,52 +287,88 @@ function _M.rewrite(conf, ctx)
...
+    if not auth_conf.algorithm then
+        auth_conf.algorithm = "HS256"
+    end
+    local sign_handler = get_sign_handler(auth_conf.algorithm)
+    local jwt_token, err = sign_handler(auth_conf.key, auth_conf, payload)
+    return jwt_token, err
 end
 
+-- only for test
+_M.gen_token = gen_token
```
moonming, line 371: we don't need this.

apisix/plugins/jwt-auth.lua
```diff
@@ -387,52 +287,88 @@ function _M.rewrite(conf, ctx)
...
+    return jwt_token
+end
 
-    core.log.info("consumer: ", core.json.delay_encode(consumer))
-
-    local sign_handler = algorithm_handler(consumer, true)
-    local jwt_token = sign_handler(key, consumer, payload)
-    if jwt_token then
-        return core.response.exit(200, jwt_token)
+local function sign_jwt_with_RS256_ES256(key, auth_conf, payload)
```
moonming, line 326: we don't need this.

apisix/plugins/jwt-auth.lua
```diff
@@ -387,52 +287,88 @@ function _M.rewrite(conf, ctx)
...
+        core.table.merge(extra_payload, real_payload)
+        return extra_payload
     end
+    return real_payload
+end
 
-    local consumer_conf = consumer_mod.plugin(plugin_name)
-    if not consumer_conf then
-        return core.response.exit(404)
+local function sign_jwt_with_HS(key, auth_conf, payload)
```
moonming, line 303: ditto

moonming (changes requested):

apisix/plugins/jwt-auth.lua
dspo: `_M.gen_token` is only for test. It is often necessary to generate jwt tokens for requests in tests. `_M.gen_token` can be introduced in tests (`local gen_token = require(‘apisix.plugins.jwt-auth’).gen_token`). This function is completely separate from the gateway authentication jwt.

dspo: I removed the `gen_token` related code from apisix/plugins/jwt-auth.lua.
I put `gen_token` in `t/lib/apisix/plugins/jwt-auth.lua`, following the practice of other tests of putting functions used in tests in the `t/lib` directory.

t/lib/apisix/plugins/jwt-auth.lua
```diff
@@ -0,0 +1,122 @@
+--
```
moonming, line 1: please use https://jwt.io/#debugger-io to make your test data

dspo: I modified the PR to use https://jwt.io/#debugger-io to make test data in most cases.

t/fips/jwt-auth.t
```diff
@@ -180,7 +145,7 @@ passed
 
 
 
-=== TEST 6: JWT sign and verify use RS256 algorithm(private_key numbits = 1024)
+=== TEST 5: JWT sign and verify use RS256 algorithm
```
moonming, line 148: ?

t/fips/jwt-auth.t
moonming, line 1: I don't understand this test file

t/perf/test_http.py
moonming, line 1: why change this file? I think you can remove the perf CI

t/plugin/jwt-auth.t
moonming, line 1: there are many `private_key` in this test file, I don't know why we need it.

moonming (changes requested):

dspo:
@moonming @membphis please review again

moonming (approved):

membphis (approved):
Nice, LGTM

nic-6443 (approved):

PR #11648, opened 2024-10-15
feat: Changed the logic to detect token generation uri with parameterized content type.

foogao:
### Description

In HTTP specification, the form of Content-Type as following:
Content-Type: <media-type>[; key=value]

For examples:

Content-Type: text/html; charset=utf-8
Content-Type: multipart/form-data; boundary=boundary-string

But in the authz-keycloak plugin, the condition to decide if current request is for granting token from Keycloak must be full match the "application/x-www-form-urlencoded". It will cause some client libraries to be unable to obtain the token.

For example, the RestTemplate in SpringFramework, it will add parameter after the content-type directive automatically.

This pull request resolved this issue.

### Checklist


foogao:
Hello @tzssangglass, @jenskeiner and @sshniro  can you help to review my request? thanks a lot.

Issue #11647, opened 2024-10-15
bug: failed to retrtive data from aws secret manager when the secret name contains a slash `/`


kayx23:
### Current Behavior

When configuring secrets in AWS, the suggestive text for secret name uses slashes, so that it could be quite common that users have slashes in their secret names.

![img_v3_02fm_c169f4d3-98d5-482f-9285-0d608daa0c9g](https://github.com/user-attachments/assets/5f6d94ee-0001-4009-a73a-59d8e56aa8be)

Suppose the secret name is `john/secret` in AWS and the key of the credential is `john-key-auth`. In apisix, the secret is referenced by `key-auth` as such:

```shell
curl "http://127.0.0.1:9180/apisix/admin/consumers/john/credentials" -X PUT \
  -d '{
    "id": "cred-key-auth",
    "plugins": {
      "key-auth": {
        "key": "$secret://aws/1/john/secret/john-key-auth"
      }
    }
  }'
```

Suppose the `key-auth` is enabled on a route. When requesting this route, you would receive 401 from APISIX and see the below error in the error log:

```
failed to fetch secret value: failed to retrtive data from aws secret manager: invalid status code 400, {"Message":"Secrets Manager can't find the specified 
...
run_plugin(): key-auth exits with http status code 401, client: [192.168.65.1](http://192.168.65.1/), server: _, request: "GET /anything HTTP/1.1", host: "[127.0.0.1:9080](http://127.0.0.1:9080/)"
```

This is likely due to the slash in the secret name. There might be a parsing case to catch. From testing, it seems that a slash in the secret name leads to errors, but a slash in the key of the credential key-value pair does not.

### Expected Behavior

APISIX to accommodate the situation where secret names contain slashes.

### Environment

- APISIX version (run `apisix version`): `3.11.0`

kayx23:
just fyi @HuanXin-Chen

HuanXin-Chen:
> just fyi @HuanXin-Chen

Ok, I know what the problem is, it may take a little time to do a fix.

HuanXin-Chen:
> just fyi @HuanXin-Chen

Could you please help me test if GCP has a similar problem, I referred to a bit less test case at the time, thank you❤️

HuanXin-Chen:
@kayx23 Could you please help me test if GCP has a similar problem, I referred to a bit less test case at the time, thank you❤️

PR #11655, opened 2024-10-15
fix(secret): support aws secret name contains a slash /

HuanXin-Chen:
### Description

Currently, AWS Secrets Manager supports secret names that contain "/".

Fixes https://github.com/apache/apisix/issues/11647

### Checklist

### DIFF
--- a/apisix/secret/aws.lua
+++ b/apisix/secret/aws.lua
@@ -18,12 +18,12 @@
 --- AWS Tools.
 local core = require("apisix.core")
 local http = require("resty.http")
-local aws = require("resty.aws")
+local aws  = require("resty.aws")
 
-local sub = core.string.sub
-local find = core.string.find
-local env = core.env
-local unpack = unpack
+local sub        = core.string.sub
+local rfind_char = core.string.rfind_char
+local env        = core.env
+local unpack     = unpack
 
 local schema = {
     type = "object",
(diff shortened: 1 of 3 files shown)


HuanXin-Chen:
@kayx23 It has been fixed; check if it meets your expectations.

HuanXin-Chen:
### My Thoughts:
AWS differs from others like Vault in that it can not only store JSON text but also String types.

Therefore, for instance, when referencing: APISIX/JSON/JACK
- If it's a JSON type, then the main_key would be APISIX/JSON, and the sub_key would be JACK.
- If it's a String type, both the main_key and the key should be APISIX/JSON/JACK.

You cannot differentiate solely based on APISIX/JSON/JACK, so you need to try both scenarios.

- When main_key == key, it is definitely a String type.
- When main_key != key, it could be either JSON or String.

If the JSON does not meet the requirements, then it's a String type.

### Other Issues:
I noticed some errors in CI, but upon initial inspection, they do not seem to be related to my code.

shreemaan-abhishek:
@HuanXin-Chen, if the key contains `/` main_key and key can never be same.

I think it's better to query with the whole key first, if that fails then fall back to querying aws with main key and then parse the JSON and extract the value from subkey. WDYT?

Baoyuantop:
Hi @HuanXin-Chen, following up on the previous review comments. Please let us know if you have any updates. Thank you.

HuanXin-Chen:
@moonming @Baoyuantop I apologize for the delayed response; I have only just now seen your email. My team is currently handling a critical project, and I am unlikely to be available in the coming period. Would it be possible to assign this task to another community member? I appreciate your understanding.

Baoyuantop:
Hi @HuanXin-Chen, thank you for your contribution. I will close this PR now, and other community members are welcome to contribute.

PR #11675, opened 2024-10-23
feat(jwt-auth): store JWT in the request context

### LINKED ISSUE #11281: feat: As a user, I want to obtain the previously validated JWT from the context, so that I can use it in my custom plugins with lower priority
### Description

I'd like to extend [jwt-auth plugin](https://github.com/apache/apisix/blob/master/apisix/plugins/jwt-auth.lua) to **store the authenticated JWT in the request context**, so that it can easily be used in a plugin which executes after (lower priority), such as a custom ACL based on JWT payload.

### Benefits

These are the benefits that come to my mind:

- hiding the token from headers - this allows to enable `hide_credentials` and hide the JWT from the server receiving the request, but still allows subsequent plugins to access it;
- less redundant code - the following plugins do not have to import extra modules to parse _**again**_ the token, but can simply access its claims as a table.

### Example

Suppose the `jwt-auth` plugin stores the decoded jwt object in request context (`jwt-auth.lua`):

```lua
    -- [...]

    ctx.jwt_obj = jwt_obj
end
```

Custom plugin example (`my-acl.lua`):

```lua
    -- [...]

    if not ctx.jwt_obj then
        -- Technically this can only happen when the JWT plugin didn't run for this request, or the priority was lower
        core.log.warn("Token object not found in ctx.")
        return 401, {message = "Missing JWT in request"}
    end

    local authorized = false
    local roles = ctx.jwt_obj.payload.roles

    -- Do something with roles
    for i, value in ipairs(tab) do
        if value == "admin" then
            authorized = true
        end
    end

    if not authorized then
        core.log.warn("Request not authorized")
        return 403, {message = "You are not allowed to access this resource"}
    end

    -- [...]
end
```

### Additional Information

Kong JWT Plugin stores the encoded token in context, which I find interesting but a little bit unhandy: https://github.com/Kong/kong/blob/d4ab528fa2414d996861a43e58406c02b4978157/kong/plugins/jwt/handler.lua#L146

mikyll:
### Description

The changes introduce a new parameter to [jwt-auth](https://apisix.apache.org/docs/apisix/plugins/jwt-auth/) plugin's route schema: `store_in_ctx`. If set to `true` (default is `false`) the plugin will store the validated JWT object in the request context.

This feature is useful especially for custom plugins. For example, my company is currently using a custom ACL plugin, based on Casbin ([lua-casbin](https://github.com/casbin/lua-casbin)), that parses a JWT to extract permissions.

Therefore, I would say this feature is useful for at least for two reasons:

- since the JWT can be removed from the request attributes (via `hide_credential = true` configuration), this provides another way to pass the token, without exposing it;
- this prevents custom plugins from duplicated code (retrieve and parse JWT object), which `jwt-auth` plugin already does.

Fixes #11281 

### Checklist


juzhiyuan:
@mikyll Thank you! I will request reviewers.

apisix/plugins/jwt-auth.lua
```diff
@@ -295,6 +299,10 @@ local function find_consumer(conf, ctx)
         return nil, nil, "failed to verify jwt"
     end
 
+    if conf.store_in_ctx then
+        ctx.jwt_obj = jwt_obj
```
bzp2010, line 303: Awesome, maybe it would be better called `jwt_payload`, or `jwt_auth_payload`? 

Use `payload`, (and specifically prefix it with the jwt-auth plugin name, as some other plugins such as openid-connect also use JWT, which may cause conflicts

bzp2010 (dismissed):

bzp2010 (approved):
LGTM

> One suggestion for the test, it would be nice if it could be done in a clearer order like “Setup 1 - Test 1 - Setup 2 - Test 2”.

mikyll:
> LGTM
> 
> > One suggestion for the test, it would be nice if it could be done in a clearer order like “Setup 1 - Test 1 - Setup 2 - Test 2”.

@bzp2010 I moved the tests in a more proper order and fixed the docs a little 🙂

bzp2010 (approved):

shreemaan-abhishek (approved):

membphis (approved):

nic-chen (approved):

PR #11788, opened 2024-11-28
fix: allow special characters in uri params

shreemaan-abhishek:
### Description

The underlying `lua-resty-radixtree` library used a regex that would cause route matching to fail when there are special characters in the uri parameters. This has been fixed in the radixtree router library, so we upgrade the version and add a test case that ensures special characters are allowed in uri parameters when using `radixtree_uri_with_parameter`.

### Checklist


Revolyssup:
@shreemaan-abhishek lint issue

mikyll:
I tested this change and it doesn't seem to work for `%2F` (encoded `/`)...

shreemaan-abhishek:
@mikyll, this is expected behaviour as `/` is a delimiter for urls. If `/` is allowed as a special character, there will be no regularity in the boundary of the url parameter. Consider the following example:

route uri: `/v1/:id/products/:type/list`
request uri: `/v1/123/products/footwear/list`

If `/` is also accepted as a special character then the `id` url parameter would be incorrectly evaluated to `123/products/footwear/list`.

nic-6443 (approved):

moonming (approved):

Revolyssup (approved):

PR #11822, opened 2024-12-11
Updated route name length το 3 times the default, to cope with Ingress Controller problems

### LINKED ISSUE #11821: Ingress Controller ApisixRoutes: Reaching limit with less than 100 Characters
### Current Behavior

Currently, the gateway has a limitation of 100 characters for a route name. This is applied to all routes, either they are created by the ingress controller or not.

### Expected Behavior

When creating an ingress route with a route name of less than 100 characters, the route should be accepted and working

### Error Logs

Kubernetes Resources with route names less than 100 characters, will NOT always be able to be processed. This happens because in Kubernetes, the **route names are automatically concatenated with the namespace, and the name of the resource.**

That creates impossible situations, confusion, and vague situations, since developers and devops cannot foresee in advance how much the route name will take under the hood.

### Steps to Reproduce

- Create a setup with Apisix Ingress Controller
- Create a route with 99 characters as one of the route names (to be sure)
- Apply.
- Watch the Ingress throw errors due to validation issues.

### Environment

- All versions of Apisix
- Kubernetes (any version)

### Question
Is there any reason why there is a limit of 100 characters in the route name in the first place? I have a PR ready, I don't know if it's the correct way of solving, this - but when this happens, it's affecting everyone using the cluster.

csotiriou:
## Description

Fixes #11821

## What it does.
It makes the route name accept a name length that is 3 times the default one. That is because in the Ingress Controller, the route name is actually automatically given by concatenating the namespace name + other factors, therefore.

Question:
- Is this better to be a static number (like 600, without depending on the default rule)?

### DIFF
--- a/apisix/schema_def.lua
+++ b/apisix/schema_def.lua
@@ -549,7 +549,11 @@ _M.route = {
             minItems = 1,
             uniqueItems = true,
         },
-        name = rule_name_def,
+        name = {
+            type = "string",
+            maxLength = rule_name_def.maxLength * 3,
+            minLength = rule_name_def.minLength,
+        },
         desc = desc_def,
         priority = {type = "integer", default = 0},
 


bzp2010:
I don't think it's a good idea to just do dirty patches like this on the route spec. If it is ok to extend the limit values, then the same thing should be done for all resources.

It's arguable that such a restriction might only be done due to some “inertia” on the part of the API creators to follow what they've done before, which might not really be necessary or justified.

The length of the value is meaningless in terms of functionality, and may only affect the speed of etcd synchronization and the amount of memory used by APISIX when large strings are used.

Perhaps you could start a discussion to expand the value on all resources.

Or perhaps the limit could be moved to a configuration file and allow the user to define it at runtime without modifying the code.

What do others think?

csotiriou:
I agree, on changing the route spec, but I found that this is a central space where all of this is gathered. This value is then picked up by multiple places inside the code. 

I believe that placing a character limit of 300 by default for all resources like this, and then allowing Helm Charts to change those values during Apisix installation time to be ideal. What would be a good way to achieve this?

Baoyuantop:
> Or perhaps the limit could be moved to a configuration file and allow the user to define it at runtime without modifying the code.

LGTM

Baoyuantop:
Hi @csotiriou, are you still working on this PR?

Baoyuantop:
cc @moonming  @membphis

membphis:
> Perhaps you could start a discussion to expand the value on all resources.
> 
> Or perhaps the limit could be moved to a configuration file and allow the user to define it at runtime without modifying the code.

Personally, I prefer this approach: allowing users to redefine in the configuration file.

After all, for most users, a 100-byte name is more than enough.

csotiriou:
> After all, for most users, a 100-byte name is more than enough.

Actually, that's the problem it **isn't** 100 bytes at all. It's less than half sometimes. 

100 bytes is the **technical** limit, but internally the Ingress controller concatenates the namespace name, and service name to that, thereby exceeding 100 bytes with ease even if the user has given 30 characters as a name. And that is opaque to the user. The result is in production we will get errors without actually knowing the root cause.

membphis:
ok, I got it.

we can change the name's max length to `256`

image

Baoyuantop:
Hi @csotiriou, any updates?

csotiriou:
> Hi @csotiriou, any updates?

I made the change and pushed it just now. 

If it was up to me I would push it to 512 just to be on the safe side - since the concatenation made by ingress controller means that you are getting 30% the space you think you do. 

But at this point I will take what I can get :)

membphis:
@csotiriou pls rebase the master branch, CI failed now

csotiriou:
@membphis I believe it should be fine now, can you check?

Baoyuantop:
Hi @csotiriou, I see errors related to this PR in the failing tests, please check.

SkyeYoung:
@csotiriou any update?

csotiriou:
@SkyeYoung I made the changes a few days ago, and I see that all tests have already passed.

apisix/schema_def.lua
```diff
@@ -104,7 +104,7 @@ local labels_def = {
 
 local rule_name_def = {
     type = "string",
-    maxLength = 100,
+    maxLength = 256,
```
SkyeYoung, line 107: Unrelated to PR:

I read the documentation and found that the API previously defined using this schema didn't specify any maximum field lengths.

I think this should be improved in the future, for example, by using OpenAPI documentation.

bzp2010:
@AlinsRan

Please evaluate this and see if it's really necessary.

For example, does the concatenation of kubernetes resource names to form APISIX resource names still exist in ingress controller v2?
If so, should we continue with this approach? Is using a short hash string as an alternative?

If we still use the old way, are the changes in this PR necessary?

This requires input from the APISIX ingress controller maintainers.

AlinsRan:
> @AlinsRan
> 
> Please evaluate this and see if it's really necessary.
> 
> For example, does the concatenation of kubernetes resource names to form APISIX resource names still exist in ingress controller v2? If so, should we continue with this approach? Is using a short hash string as an alternative?
> 
> If we still use the old way, are the changes in this PR necessary?
> 
> This requires input from the APISIX ingress controller maintainers.

The uniqueness of APISIX resources is guaranteed by the id, while the name is mainly for readability. Currently, Ingress Controller v2 still generates APISIX resource names by concatenating Kubernetes resource names. Therefore, increasing the length limit from 100 to 256 is reasonable, as it aligns with Kubernetes’ 253-character limit.

Baoyuantop:
cc @bzp2010

bzp2010:
Although I don't understand why filling in the name is mandatory, as it serves no purpose for APISIX, I note that the value length in labels is limited to 256. Therefore, I partially agree with modifying this restriction. However, there is no fundamental justification that this change actually makes sense.

Once we expand this limitation, we won't be able to reimpose restrictions later.

Storing key-value pairs using labels might be more beneficial for APISIX to utilize them.

PR #7965, opened 2022-09-22
fix: use modifiedIndex as lru key when merge plugins from route and consumer

tzssangglass:
…consumer

### Description

Fixes # (issue)

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #11840, opened 2024-12-18
perf: accelerate the creation of the consumer cache

xuruidong:
### Description

In my case, there are approximately 17,000+ consumers. Any update from a consumer will trigger a long-tail request. I found the methods `consumer.plugin` and `consumer.consumers_kv` cost too much time.

#### Test steps 
Update one consumer, send a request. Repeat 5 times.

#### Test result:
Before optimization:
unit: second
|consumer.plugin|consumer.consumers_kv|
|------------------|--------------------------|
|0.192824| 3.209864 |
| 0.178377 | 2.837157 |
| 0.156748 | 3.573214 |
|0.152296 |  2.720143 |
|0.161006| 2.530867 |

After optimization:
unit: second
|consumer.plugin|consumer.consumers_kv|
|------------------|--------------------------|
| 0.050382 |  0.036338 |
| 0.021982 | 0.016851 |
| 0.023093 | 0.017966 |
| 0.022090 | 0.016934 |

Fixes # (issue)

### Checklist


praswicaksono:
To purge local cache need to restart apisix?

membphis:
this is the core reason whey the new way is better performance.

use the cached data, reduce to recall `secret.fetch_secrets`

image

I think we can use `lrucache`, which is easier way, here is the code example: 

```lua
local lrucache = core.lrucache.new({
    ttl = 300, count = 5000  -- we can change the `count` as we need, in your case, we may change it to more than 17,000+
})

local new_consumer, err = lrucache(consumer, nil,
                    secret.fetch_secrets, consumer.auth_conf, false)
```

I can not 100% confirm if it works, welcome to make a test in this way, many thx

xuruidong:
> this is the core reason whey the new way is better performance.
> 
> use the cached data, reduce to recall `secret.fetch_secrets`
> 
> image
> I think we can use `lrucache`, which is easier way, here is the code example:
> 
> ```lua
> local lrucache = core.lrucache.new({
>     ttl = 300, count = 5000  -- we can change the `count` as we need, in your case, we may change it to more than 17,000+
> })
> 
> local new_consumer, err = lrucache(consumer, nil,
>                     secret.fetch_secrets, consumer.auth_conf, false)
> ```
> 
> I can not 100% confirm if it works, welcome to make a test in this way, many thx

Let me make the ci happy first, and then I'll try your suggestion.

xuruidong:
ping @membphis

apisix/consumer.lua
```diff
@@ -101,10 +104,19 @@ local function plugin_consumer()
...
                     }
                 end
 
+                local cached_consumer = consumers_id_cache[val.value.id]
+                if cached_consumer and
+                    cached_consumer.modifiedIndex == val.modifiedIndex then
+                    plugins[name].len = plugins[name].len + 1
+                    core.table.insert(plugins[name].nodes, plugins[name].len,
+                                        cached_consumer)
+                    goto CONTINUE_INTERNAL
```
membphis, line 118: shorter is better

```suggestion
                    goto CONTINUE
```

apisix/consumer.lua
```diff
@@ -101,10 +104,19 @@ local function plugin_consumer()
                 if not plugins[name] then
                     plugins[name] = {
                         nodes = {},
+                        len = 0,
                         conf_version = consumers.conf_version
                     }
                 end
 
+                local cached_consumer = consumers_id_cache[val.value.id]
```
membphis, line 112: I do not know why we have to add `consumers_id_cache `

it seems useless

apisix/consumer.lua
```diff
@@ -186,24 +204,43 @@ function _M.consumers()
...
+        consumers_plugin_key_lrucache_tab[plugin_name] = lru_cache
+    end
 
     for _, consumer in ipairs(consumers_conf.nodes) do
         core.log.info("consumer node: ", core.json.delay_encode(consumer))
-        local new_consumer = core.table.clone(consumer)
-        new_consumer.auth_conf = secret.fetch_secrets(new_consumer.auth_conf, true,
-                                                        new_consumer.auth_conf, "")
-        consumer_names[new_consumer.auth_conf[key_attr]] = new_consumer
+        local new_consumer = lru_cache(consumer.auth_conf[key_attr],
```
membphis, line 230: I think: this should be enough, and it seems easier

```suggestion
        local new_consumer = lru_cache(consumer, nil, create_new_consumer, consumer)
```

apisix/consumer.lua
```diff
@@ -186,24 +204,43 @@ function _M.consumers()
 end
 
 
-local function create_consume_cache(consumers_conf, key_attr)
+local create_consume_cache
+do
+    local consumers_plugin_key_lrucache_tab = {}
+
+local function create_new_consumer(consumer)
```
membphis, line 211: we need a better name, it is not creating a new consumer

it should be `transform` or `fill secret`, what do you think?

membphis (changes requested):

apisix/consumer.lua
xuruidong: The purpose of using consumer_id_cache is to avoid unnecessary table copies. I will refactor this part of the code to replace the table with an LRU cache.

apisix/consumer.lua
xuruidong: Yes，I‘ll refactor this code.

xuruidong:
> To purge local cache need to restart apisix?

yes

apisix/consumer.lua
```diff
@@ -101,46 +152,21 @@ local function plugin_consumer()
                 if not plugins[name] then
                     plugins[name] = {
                         nodes = {},
+                        len = 0,
```
xuruidong, line 155: The purpose of introducing the len variable here is to reduce the redundant computation of the table length. Refer to the `table.insert` section in this link:  https://api7.ai/learning-center/openresty/lua-table-and-metatable

xuruidong:
The code introduces a variable `consumers_count_for_lrucache`, which is used to set the size of the LRU cache. In actual use, this value needs to be configured to be greater than the actual number of consumers. Additionally, if a consumer has multiple authentication plugins, each of them will be considered as a separate consumer and cached individually in the LRU cache.

When the number of consumers exceeds twice the size of the LRU cache, the process of creating cache during the execution of the `consumer.plugin` function may result in cache misses, leading to performance degradation. This happens because the LRU cache can only store consumers within a single window, and each time a consumer is updated, the order of the consumer IDs in the `consumers.values` array is likely to either remain the same or change very little. As a result, new consumer objects are continuously created, causing the old consumer cache entries to be evicted.

---

代码中引入了一个变量 consumers_count_for_lrucache , 用来设置 lrucache 的大小。实际使用的时候，这个值需要配置为大于 consumer 的实际数量，并且如果一个consumer 中存在多个认证插件，则会被认为是不同的consumer 缓存到 lrucache 中。

当 consumer 数量大于2倍的lrucache 大小的时候，在执行 `consumer.plugin` 函数创建缓存的过程中可能会一直不能命中缓存而导致性能下降。因为 lrucache 只能缓存一个窗口内的 consumer，而每次 consumer 有更新时，consumers.values 数组中的 consumer id 的顺序大概率可能不变或者变化很小，此时会一直存在生成新的 consumer object ，淘汰旧的 consumer 缓存。

xuruidong:
ping @membphis

apisix/consumer.lua
```diff
@@ -186,20 +215,34 @@ function _M.consumers()
...
 
-local function create_consume_cache(consumers_conf, key_attr)
+local create_consume_cache
+do
+    local consumer_lrucache = core.lrucache.new({
+            count = consumers_count_for_lrucache
+        })
+
+local function fill_consumer_secret(consumer)
+    local new_consumer = core.table.clone(consumer)
```
membphis, line 225: do you think we should use `deep_clone`?
is it safer?

xuruidong: I think deep-copy is not necessary. The consumer in the function `fill_consumer_secret(consumer)` is created in the `construct_consumer_data` function and is a copy of an element from the `consumers.values` list.

membphis (commented):
I will check it carefully next week, I am busy this week

apisix/consumer.lua
membphis: ok, got it

thx for your explanation

apisix/consumer.lua
```diff
@@ -186,20 +215,33 @@ function _M.consumers()
...
 
-local function create_consume_cache(consumers_conf, key_attr)
+local create_consume_cache
+do
+    local consumer_lrucache = core.lrucache.new({
+            count = consumers_count_for_lrucache
+        })
+
+local function fill_consumer_secret(consumer)
+    consumer.auth_conf = secret.fetch_secrets(consumer.auth_conf, false)
```
membphis, line 225: it is not safe to update the original `consumer` object.

more safe way:

```lua
local new_consumer = table.clone(consumer)
new_consumer.auth_conf = secret.fetch_secrets(consumer.auth_conf, false)
return new_consumer
```

apisix/consumer.lua
```diff
@@ -36,6 +36,11 @@ local lrucache = core.lrucache.new({
     ttl = 300, count = 512
 })
 
+-- Please calculate and set the value of the "consumers_count_for_lrucache"
+-- variable based on the number of consumers in the current environment,
+-- taking into account the appropriate adjustment coefficient.
+local consumers_count_for_lrucache = 4096
```
xuruidong, line 42: the usage of the variable is here.

apisix/consumer.lua
```diff
@@ -80,7 +85,53 @@ local function filter_consumers_list(data_list)
...
+            -- Normally wouldn't get here:
+            -- it should belong to a consumer for any credential.
+            core.log.error("failed to get the consumer for the credential,",
+                " a wild credential has appeared!",
+                " credential key: ", val.key, ", consumer name: ", consumer_name)
+            return nil, "failed to get the consumer for the credential"
+        end
+    else
+        consumer = core.table.clone(val.value)
+        consumer.modifiedIndex = val.modifiedIndex
```
membphis, line 116: we do not need this line, I think we can remove it

membphis (approved):
LGTM ^_^

shreemaan-abhishek (approved):

moonming (approved):

apisix/consumer.lua
xuruidong: https://github.com/apache/apisix/pull/7965    
This PR may explain the purpose of the "modifiedIndex" variable.

membphis: I think `consumer.modifiedIndex` equal `val.modifiedIndex` after we called `core.table.clone(val.value)1`

We do not need to set this value again

Only comments should be enough.

membphis: What do you think?

xuruidong: `modifiedIndex` is a member of `val`, not a member of `val.value`,

membphis: sorry, it is my fault, you're right

membphis:
many thx @xuruidong 's contribution, merged your PR right now

apisix/consumer.lua
```diff
@@ -101,46 +152,21 @@ local function plugin_consumer()
...
-                    consumer = core.table.clone(val.value)
-                    consumer.modifiedIndex = val.modifiedIndex
-                end
-
-                -- if the consumer has labels, set the field custom_id to it.
-                -- the custom_id is used to set in the request headers to the upstream.
-                if consumer.labels then
-                    consumer.custom_id = consumer.labels["custom_id"]
+                local consumer = consumers_id_lrucache(val.value.id .. name,
+                        val.modifiedIndex, construct_consumer_data, val, config)
```
nic-6443, line 161: We cannot cache the result of `construct_consumer_data` here, because even when there is no change in credential (`val.modifiedIndex` has not changed), the consumer might have been modified. We need to design different versions for `consumer` and `credential`.

PR #11844, opened 2024-12-19
fix: corrupt data in routes() response due to healthchecker data

Revolyssup:
### Description

Fixes # (issue)
Currently there is an intermittent bug. Below are the steps to reproduce:
Reproduction steps
1. Create a route and enable health check

```bash
curl http://127.0.0.1:9180/apisix/admin/routes/1 -H 'X-API-KEY: wqGBivhkYbFXiMdcMEAOMZanCGnLSbWE' -X PUT -d '
{
    "uri": "/*",
    "upstream": {
         "nodes": {
            "httpbin.org:80": 1,
            "mockbin.org:80": 1
        },
        "type": "roundrobin",
        "checks": {
            "active": {
                "timeout": 5,
                "type": "tcp",
                "healthy": {
                    "interval": 2,
                    "successes": 1
                },                
                "unhealthy": {
                    "interval": 2,
                    "successes": 1
                }
            }
        }
    }
}'
```

2. **access routes to trigger health checks**

```bash
curl http://127.0.0.1:9080/get
```

3. **When accessing the control api, 500 will be returned and the error in the problem description will be generated.**
### Checklist

This error is intermittent and multiple tries might be needed to reproduce. 
On debugging it was found that the cause of issue was http_router.user_routes table being polluted by adding healthchecks leaving unexpected data which while trying to parse, the error was generated.

As you can see in the generated logs, user_route returned healthchecker data.
Screenshot from 2024-12-19 15-11-00
Screenshot from 2024-12-19 15-10-51

## Fix
Remove healthcheck related objects before sending to user.

### DIFF
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -269,6 +269,14 @@ local function iter_add_get_routes_info(values, route_id)
         if new_route.value.upstream and new_route.value.upstream.parent then
             new_route.value.upstream.parent = nil
         end
+        -- remove healthcheck info
+        new_route.checker = nil
+        new_route.checker_idx = nil
+        new_route.checker_upstream = nil
+
+        if new_route.clean_handlers then
+            new_route.clean_handlers = {}
+        end
         core.table.insert(infos, new_route)
         -- check the route id
         if route_id and route.value.id == route_id then
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -273,7 +273,6 @@ local function iter_add_get_routes_info(values, route_id)
         new_route.checker = nil
         new_route.checker_idx = nil
         new_route.checker_upstream = nil
-
         if new_route.clean_handlers then
             new_route.clean_handlers = {}
         end
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -273,9 +273,7 @@ local function iter_add_get_routes_info(values, route_id)
         new_route.checker = nil
         new_route.checker_idx = nil
         new_route.checker_upstream = nil
-        if new_route.clean_handlers then
-            new_route.clean_handlers = {}
-        end
+        new_route.clean_handlers = nil
         core.table.insert(infos, new_route)
         -- check the route id
         if route_id and route.value.id == route_id then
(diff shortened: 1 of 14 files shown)


membphis:
we can not update the watched data from etcd, it is dangerous

Revolyssup:
> we can not update the watched data from etcd, it is dangerous

The original data is not being modified. It is already deepcopied. But to make it easier to review I have refactored to move the modification logic just below the deepcopy function so that it's clear that the original data is not being modified.

apisix/control/v1.lua
```diff
@@ -269,6 +269,13 @@ local function iter_add_get_routes_info(values, route_id)
         if new_route.value.upstream and new_route.value.upstream.parent then
             new_route.value.upstream.parent = nil
         end
+        -- remove healthcheck info
+        new_route.checker = nil
```
membphis, line 273: @nic-chen pls take a look when you have time

this changing seem fine to me, but I can not sure 100%

nic-chen: No problem, I discussed it with @Revolyssup  yesterday

apisix/control/v1.lua
```diff
@@ -269,6 +269,13 @@ local function iter_add_get_routes_info(values, route_id)
...
+        new_route.checker_idx = nil
+        new_route.checker_upstream = nil
+        if new_route.clean_handlers then
+            new_route.clean_handlers = {}
+        end
```
nic-chen, line 278: why not just also set to nil

Revolyssup: some existing test cases grep logs and expect this to {}

nic-chen: Then you can fix the test case as well.
Checking the logs has no meaning for the control api.

apisix/control/v1.lua
```diff
@@ -269,6 +269,13 @@ local function iter_add_get_routes_info(values, route_id)
         if new_route.value.upstream and new_route.value.upstream.parent then
             new_route.value.upstream.parent = nil
         end
+        -- remove healthcheck info
```
nic-6443, line 272: we need remove healthcheck info from `iter_add_get_services_info` too

Revolyssup: done

apisix/control/v1.lua
Revolyssup: done

nic-chen (commented):
please add test cases to ensure error is gone

Revolyssup:
> please add test cases to ensure error is gone

added

nic-chen (approved):

shreemaan-abhishek (approved):

nic-6443 (approved):

Issue #11711, opened 2024-11-05
bug: Two sessions shown when using openid-connect


arnab-sharma:
### Current Behavior

We see two sessions 'session' and 'session_2' being set as cookies when using openid-connect. Wanted to understand why session_2 is being set as well.

### Expected Behavior

Expected to see only one cookie 'session' being set.

### Error Logs

![Two_session_cookies](https://github.com/user-attachments/assets/6a9e5ea7-d3e3-4c9f-8b17-763731c81af2)

### Steps to Reproduce

Sample Code:
![image](https://github.com/user-attachments/assets/7c073840-3024-40ea-bdea-7c56431718dc)

### Environment

- APISIX version (run `apisix version`): 3.9.1
- Operating system (run `uname -a`): Kubernetes-compatible Linux x86
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 3.5.10
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

arnab-sharma:
@juzhiyuan - Could you please take a look at above issue?

yardenshoham:
I am also seeing this behavior and want to understand it.

yardenshoham:
Looks like it's happening because the session is too large. You could reduce it's size by adding this config to the plugin:

```yaml
openid-connect:
  ...
  bearer_only: false
  # this comes from https://github.com/zmartzone/lua-resty-openidc, if this is missing, all information will be stored in the session. This
  # causes the cookies to be too large and the request to fail. We have control over id_token, user, enc_id_token, and access_token.
  # We currently include only access_token in the session. If this gets too big we can remove it as well.
  session_contents: 
    access_token: true
  session:
    secret: "NTgzZWY1MzAtOTcyNi00MzA5LWI1MjItNjFhMzI2NzRkZWQxCg==" # used `uuidgen|base64` to generate this secret
```

Baoyuantop:
If there is still a problem, please open it again.

PR #11889, opened 2025-01-06
docs: improve `openid-connect` plugin doc and add keycloak OIDC tutorial

kayx23:
### Description

Improve `openid-connect` plugin doc and add keycloak OIDC tutorial.

Also include the update from this PR https://github.com/apache/apisix/pull/11892

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #11892, opened 2025-01-06
docs: add troubleshooting option for openid-connect

yardenshoham:
This solved my issue when nothing else did.

References https://github.com/apache/apisix/issues/11711

### DIFF
--- a/docs/en/latest/plugins/openid-connect.md
+++ b/docs/en/latest/plugins/openid-connect.md
@@ -275,6 +275,21 @@ upstream sent too big header while reading response header from upstream
 
 If so, try adjusting `proxy_buffers`, `proxy_buffer_size`, and `proxy_busy_buffers_size` to larger values.
 
+Another option is to set the `session_content` attribute:
+
+```yaml
+openid-connect:
+  ...
+  bearer_only: false
+  # this comes from https://github.com/zmartzone/lua-resty-openidc, if this is missing, all information will be stored in the session. This
+  # causes the cookies to be too large and the request to fail. We have control over id_token, user, enc_id_token, and access_token.
+  # We currently include only access_token in the session. If this gets too big we can remove it as well.
+  session_contents: 
+    access_token: true
+  session:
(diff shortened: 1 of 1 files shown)


docs/en/latest/plugins/openid-connect.md
```diff
@@ -275,6 +275,21 @@ upstream sent too big header while reading response header from upstream
...
 
+Another option is to set the `session_content` attribute:
+
+```yaml
+openid-connect:
+  ...
+  bearer_only: false
+  # this comes from https://github.com/zmartzone/lua-resty-openidc, if this is missing, all information will be stored in the session. This
+  # causes the cookies to be too large and the request to fail. We have control over id_token, user, enc_id_token, and access_token.
+  # We currently include only access_token in the session. If this gets too big we can remove it as well.
```
kayx23, line 286: Could you simplify the comment to

```text
# use session_contents option to configure which information among id_token, user, enc_id_token, and access_token, get stored.
```

and add the same doc update to the Chinese doc?

docs/en/latest/plugins/openid-connect.md
```diff
@@ -275,6 +275,21 @@ upstream sent too big header while reading response header from upstream
...
+  session_contents: 
+    access_token: true
```
kayx23, line 288: Technically this is an attribute that was previously not explicitly exposed, so should be also be added to attribute section of the doc; as well as to the plugin schema in the source code (relevant tests need to be updated as well) 💭 cc: @juzhiyuan

yardenshoham:
Maybe you could take over and do everything in https://github.com/apache/apisix/pull/11889

kayx23:
Ok I will evaluate and update accordingly

kayx23:
I think it's better to separate this change from https://github.com/apache/apisix/pull/11889 as the update requires a modification to the plugin schema and test case (albeit simple).

Will wait till #11889 is merged and proceed with this change. Thanks for your contribution.

kayx23:
I'm including this info in the other doc PR now, thanks.

PR #11943, opened 2025-01-31
feat(datadog): Improve Datadog plugin tag support

deiwin:
### Description

Currently the metrics reported by Apisix's Datadog plugin are not sufficient for our observability needs (which I believe are common) and we need to rely on service-reported metrics instead for most use-cases. This is problematic for us, because service-reported metrics are fundamentally unable to include some important information. For example: rate-limited response counts, gateway errors when the service is unreachable, and delays within the network between gateway and the service.

The changes included in this PR make the metrics reported by the plugin compatible with most common observability use-cases in a backwards-compatible and cost-conscious way. Each commit includes detailed reasoning within the commit message for the specific change but an overview of the changes is provided here below as well.

* feat(datadog): add include_path option
  * If a service starts reporting error response codes or slow responses, then it's important to know if the issue affects the whole service or only some endpoints, and if only some, then which particular endpoints. Including the `path` pattern in a tag allows metrics to be grouped by endpoint in Datadog dashboards and monitors.
* feat(datadog): add include_method option
  * Similarly, `method` is required to distinguish different endpoints with the same path that use a different HTTP method. A GET and a PUT for the same path can have very different performance characteristics and failure reasons, so need to be distinguished.
* feat(datadog): add route-level constant_tags support
  * Another common observability requirement is to filter metrics by other metadata such as team ownership for example. In cases where multiple teams own different endpoints within a single service, there needs to be a way to provide custom metric tags for a group of endpoints. This allows creating monitors that alert on high error rates in any endpoint owned by a particular team, for example.
* feat(datadog): add response_status_class tag to metrics
  * Datadog tags are simple string values and so numerical operations are not supported for grouping metrics. Given this, the `response_status_class` tag is required to be able to create monitors that alert when the rate of user errors (4xx responses) is high, for example. Without such a tag, the monitor would have to explicitly group by every possible 4xx value (400, 401, 402, etc) and then add the values reported by such groups together, which is not practical in most cases.
* feat(datadog): validate constant_tags values
  * Do basic validation to the provided custom tags to avoid simple configuration mistakes.

### Checklist


Baoyuantop:
Hi @deiwin, I see PR is still in draft status, is there still time to advance?

deiwin:
Hi! I haven't been able to work on this recently, but would love to see this completed.

Based on how I read the code, the code changes should work as-is, but the docs & tests still need work.

Unfortunately I wasn't able to get the tests to run on mac (following [this guide](https://apisix.apache.org/docs/apisix/build-apisix-dev-environment-on-mac/)), so I put this on hold for now.

Baoyuantop:
> Unfortunately I wasn't able to get the tests to run on mac

Can you share the specific problem?

Baoyuantop:
Hi @deiwin, the running Docker container may have limited resources, causing the process to execute slowly and fail to complete within the time expected by the test framework. You can try to modify the timeout parameters of the test framework to give the process more time to complete the operation. You can check the configuration of the Test::Nginx::Socket module to see if the timeout parameter can be adjusted.

deiwin:
Any specific conf changes (file & parameter) you could recommend? My Docker should have 8 (M1) cores & 15.5GB of memory available, so it shouldn't be too constrained.

Baoyuantop:
> Any specific conf changes (file & parameter) you could recommend? My Docker should have 8 (M1) cores & 15.5GB of memory available, so it shouldn't be too constrained.

For example timeout in `t/APISIX.pm` file

deiwin:
Looks like that worked and I am now able to reliably run at least the relevant Datadog plugin tests! 🎉 Thank you, @Baoyuantop! 🙇

I've finished the changes now and updated the PR with them. I'll now open it for review.

deiwin:
I fixed the lint issue (I hope) but the other CI errors seemed unrelated to my changes.

deiwin:
I realized there's one more change required for what I'm trying to achieve, so I added another commit to also add a `response_status_class` tag. Additional details in the commit message again.

apisix/plugins/datadog.lua
```diff
@@ -80,10 +89,24 @@ local function generate_tag(entry, const_tags)
         tags = {}
     end
 
+    if entry.constant_tags and #entry.constant_tags > 0 then
+        for _, tag in pairs(entry.constant_tags) do
+            core.table.insert(tags, tag)
```
Baoyuantop, line 111: It is recommended that label format validation be added to ensure that labels conform to Datadog's label specifications (e.g., character limits, length limits, etc.)

apisix/plugins/datadog.lua
```diff
@@ -80,10 +89,24 @@ local function generate_tag(entry, const_tags)
...
+        end
+    end
+
     if entry.route_id and entry.route_id ~= "" then
         core.table.insert(tags, "route_name:" .. entry.route_id)
     end
 
+    if entry.path and entry.path ~= "" then
```
Baoyuantop: Can you use functions to reduce duplicate code?

apisix/plugins/datadog.lua
deiwin: Added here: https://github.com/apache/apisix/pull/11943/commits/b761540f61f26c5072cb6513adbbffbe5c99514b

Although to be clear, these new tags previously followed the current approach with `constant_tags` provided via plugin configuration, which were also included as-is, without validation. Both are validated now.

apisix/plugins/datadog.lua
deiwin: Did you have something like this in mind: https://github.com/apache/apisix/pull/11943/commits/004b574382cc8dd8f75b4e1693face763da8f730?

apisix/plugins/datadog.lua
```diff
@@ -80,25 +107,32 @@ local function generate_tag(entry, const_tags)
         tags = {}
     end
 
-    if entry.route_id and entry.route_id ~= "" then
-        core.table.insert(tags, "route_name:" .. entry.route_id)
+    if entry.constant_tags and #entry.constant_tags > 0 then
+        for _, tag in pairs(entry.constant_tags) do
```
Baoyuantop: should use ipairs to iterate over the array.

deiwin: Right, not sure why I used `pairs` here. Fixed: https://github.com/apache/apisix/compare/004b574382cc8dd8f75b4e1693face763da8f730..3d61e5984e58e121478bfcad35c033cdd90f2c12

deiwin:
I also rebased this onto latest `master`, just in case

Baoyuantop:
Hi @deiwin, could you please add some descriptive information so that other maintainers can understand your intention and review this PR more effectively, such as the reason for adding this feature improvement and the application of these features in an actual business context?

deiwin:
@Baoyuantop each commit already has detailed reasoning for the specific change within that commit. But I can try improving the overall PR description as well.

deiwin:
Updated the PR description as well

membphis:
> Updated the PR description as well

many thx for your nice explain, I totally understand why we need this PR

membphis (approved):

Baoyuantop (approved):

PR #11987, opened 2025-02-22
feat(openid-connect): add jwt audience validator

### LINKED ISSUE #11968: add extra validation if audience claim exists in the Bearer token
refer to https://github.com/apache/apisix/pull/11059

bzp2010:
### Description

Add JWT audience authentication to the OpenID Connect plugin, which allows:

- Asserts that the claim must exist, otherwise the request is rejected.
- Asserts that it should be equal to or contain the client_id to comply with the OIDC specification requirements, otherwise the request is rejected.
- The claim can be customized.

Fixes #11968 #11059

One of the developers in #11059 mentioned that it is possible to use some of the APIs in `jwt-validators` to implement JWT validation in `lua-resty-openidc`, but it doesn't work, that library only works with local verification that uses a public key, and not with the Introspection API. We have to implement the functionality directly in the plugin code to support it in both scenarios.

To keep compatibility, these features are not turned on by default and it is up to you to decide if you want to turn them on. Although the OIDC spec requires this to be the default behavior.

### Checklist


moonming (dismissed):

apisix/plugins/openid-connect.lua
```diff
@@ -89,6 +90,33 @@ local schema = {
...
+                            default = "aud",
+                        },
+                        required = {
+                            type = "boolean",
+                            description = "audience claim is required",
+                            default = false,
+                        },
+                        match_with_client_id = {
+                            type = "boolean",
+                            description = "audience must euqal to or includes client_id",
```
nic-chen, line 112: From the code, it can only be equals?

apisix/plugins/openid-connect.lua
```diff
@@ -547,6 +575,40 @@ function _M.rewrite(plugin_conf, ctx)
...
+                            matched = true
+                        end
+                    end
+                    if not matched then
+                        core.log.error("OIDC introspection failed: ",
+                                        "audience list does not contain the client id")
+                        return 403, core.json.encode(error_response)
+                    end
+                elseif conf.client_id ~= audience_value then
+                    core.log.error("OIDC introspection failed: ",
```
nic-chen, line 606: Compared to `error`, `warn` level seems more appropriate for these logs.

nic-chen (approved):
LGTM

apisix/plugins/openid-connect.lua
bzp2010: Yes, if the match is true but not required, the correlation check with `client_id` is only performed if the `aud` value is not empty.

apisix/plugins/openid-connect.lua
bzp2010: Makes sense, but the rest of the code uses error, so I'm choosing to follow the precedent, should we break that "convention"?

bzp2010:
We should consider the design of the configuration items more carefully, as it may not align with ergonomic best practices. I'd appreciate your feedback.

bzp2010:
No feedback so far, should go ahead, review and merge it.

apisix/plugins/openid-connect.lua
```diff
@@ -547,6 +575,40 @@ function _M.rewrite(plugin_conf, ctx)
...
+            if core.table.try_read_attr(conf, "claim_validator", "audience", "required")
+                and not audience_value then
+                core.log.error("OIDC introspection failed: required audience (",
+                                audience_claim, ") not present")
+                local error_response = { error = "required audience claim not present" }
+                return 403, core.json.encode(error_response)
+            end
+            if core.table.try_read_attr(conf, "claim_validator", "audience", "match_with_client_id")
+                and audience_value ~= nil then
+                local error_response = { error = "mismatched audience" }
```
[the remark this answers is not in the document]
bzp2010: If so, should I need to set the `Content-Type: application/json` header manually? 🤔

nic-chen (dismissed):

apisix/plugins/openid-connect.lua
```diff
@@ -89,6 +90,33 @@ local schema = {
...
+                        },
+                        match_with_client_id = {
+                            type = "boolean",
+                            description = "audience must euqal to or includes client_id",
+                            default = false,
+                        }
+                    },
+                },
+            },
+            default = {},
```
membphis, line 118: can we remove this default value?
empty table `{}`, it seems a little weird

membphis (dismissed):
LGTM, only one minor thing, I have commented

nic-6443 (approved):

apisix/plugins/openid-connect.lua
bzp2010: This can be remove, but `claim_validator` will not appear in configurations that have populated the defaults, nor will configuration items in `audience`, and the populating of the defaults is not recursive.
However, after checking, the code logic correctly handles the case where these config items don't exist.

The line will be removed.

nic-chen (approved):

membphis (approved):

nic-6443 (approved):

SkyeYoung (approved):

PR #11988, opened 2025-02-23
fix(grpc-web): response contains two trailer chunks

bzp2010:
### Description

This PR is intended to address the following issues:

- Sometimes grpc-web requests incorrectly contain two trailer blocks, which according to grpc's protocol requirements must exist and exist only once. (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2 Message framing)

The reason for this is easy to explain, the old code used `status` to assert whether the response body had finished.
But this is unreliable, and it can go wrong in the following cases:

Since the response is chunked, the trailer chunk from upstream may not be fully included in the first response block, so the upstream_trailer_grpc_message variable is not available in the first body_filter, which causes APISIX to trigger the body_filter call a second time and incorrectly add a duplicate trailer block.

```
RESP HEADER: balabala

--- first chunk, eof = false 
RESP BODY:
balabala
grpc-status: 0
--- second chunk, eof = true
RESP BODY:
grpc-message: "demo"
--- done
```

Based on the pseudo-code above, you'll see that when this happens, the `upstream_trailer_grpc_status` value is already available to us at the first `body_filter` call, which causes the plugin code to send a trailer chunk.
When the next upstream chunk is received, `body_filter` is called a second time, and `upstream_trailer_grpc_status` still has the value it had before, so the plugin code sends the trailer block again.

This results in multiple duplicate trailers and accidentally breaks the grpc-web js client in the browser.

This PR changes the end of the response asserted using the `status` value to use the `ngx.arg[2]` value, which will be the most reliable and trigger the code to send the trailer only on the last chunk. It does not add or modify any existing behavior, so it does not result in test case modifications either.

### Checklist

### DIFF
--- a/apisix/plugins/grpc-web.lua
+++ b/apisix/plugins/grpc-web.lua
@@ -172,9 +172,15 @@ function _M.body_filter(conf, ctx)
         n bytes: trailer
 
     --]]
-    local status = ctx.var.upstream_trailer_grpc_status
-    local message = ctx.var.upstream_trailer_grpc_message
-    if status ~= "" and status ~= nil then
+    if ngx_arg[2] then -- if eof
+        local status = ctx.var.upstream_trailer_grpc_status
+        local message = ctx.var.upstream_trailer_grpc_message
+
+        -- When the response body completes and still does not receive the grpc status
+        if status == nil or status == "" then
+            core.log.error("upstream grpc status not received")
+        end
+
         local status_str = "grpc-status:" .. status
         local status_msg = "grpc-message:" .. ( message or "")
         local grpc_web_trailer = status_str .. "\r\n" .. status_msg .. "\r\n"


moonming (approved):

membphis (approved):

nic-6443 (approved):

membphis (approved):

nic-6443 (approved):

juzhiyuan (approved):
I have reviewed the test cases, LGTM

apisix/plugins/grpc-web.lua
```diff
@@ -110,16 +150,17 @@ function _M.access(conf, ctx)
...
-    if err then
+    if err or not body then
         core.log.error("failed to read request body, err: ", err)
-        return 400
+        return exit(ctx, 400)
     end
 
     if encoding == CONTENT_ENCODING_BASE64 then
         body = decode_base64(body)
+        ngx.log(ngx.WARN, "DECODE BODY: ", body)
```
shreemaan-abhishek, line 160: 1. why do we need warn log here? 
2. we can use core.log.warn instead

apisix/plugins/grpc-web.lua
```diff
@@ -68,6 +68,46 @@ function _M.check_schema(conf)
...
+    trailer_buf = trailer_buf .. string.char(
+        bit.band(bit.rshift(len, 24), 0xff),
+        bit.band(bit.rshift(len, 16), 0xff),
+        bit.band(bit.rshift(len, 8), 0xff),
+        bit.band(len, 0xff)
+    )
+    -- n bytes: trailer
+    trailer_buf = trailer_buf .. grpc_web_trailer
+
+    return  trailer_buf
```
shreemaan-abhishek, line 108:
```suggestion
    return trailer_buf
```

apisix/plugins/grpc-web.lua
bzp2010: Obviously here leaves out some unneeded code, which I will remove in another PR. 🫡

apisix/plugins/grpc-web.lua
bzp2010: Good catch

apisix/plugins/grpc-web.lua
bzp2010: This is probably code for testing and debugging.

apisix/plugins/grpc-web.lua
shreemaan-abhishek: this is a good first issue, hopefully someone from the community can fix this.

PR #11993, opened 2025-02-25
feat: set default value of `ssl_trusted_certificate` to `system`

### LINKED ISSUE #11720: bug: Vault Secret Engine Issue LUA Trust Certificates
### Current Behavior

I am running Api Six in stand-alone mode and want to use Vault for secret management.

I am using the Docker images, and I keep getting this error. I don't know how to add certificates to the trust.

global_rules:
-
id: 1
plugins:
key-auth:
header: "Authorization"

routes:
- id: "test_route"
uri: "/test"
plugins:
key-auth: {}
upstream:
type: roundrobin
scheme: "https"
nodes:
"postb.in:443": 1

consumers:
- username: nemus_dupper
plugins:
key-auth:
key: $secret://vault/1/nemus_dupper/auth-key

secrets:
- id: vault/1
ssl_verify: false
prefix: apisix
token: hvs.asdfasdfasdfasdf
uri: https://vault.mydomain.com:8200/

api-gateway-1 | 2024/11/07 06:41:12 [error] 37#37: *1755 [lua] secret.lua:180: fetch(): failed to fetch secret value: failed to retrtive data from vault kv engine: 20: unable to get local issuer certificate, client: 172.18.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:8080"
api-gateway-1 | 2024/11/07 06:41:12 [warn] 37#37: *1755 [lua] plugin.lua:1174: run_plugin(): key-auth exits with http status code 401, client: 172.18.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:8080"

### Expected Behavior

I would like it to call the vault server I've specified in the config.

### Error Logs

api-gateway-1  | 2024/11/07 06:41:12 [error] 37#37: *1755 [lua] secret.lua:180: fetch(): failed to fetch secret value: failed to retrtive data from vault kv engine: 20: unable to get local issuer certificate, client: 172.18.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:8080"
api-gateway-1  | 2024/11/07 06:41:12 [warn] 37#37: *1755 [lua] plugin.lua:1174: run_plugin(): key-auth exits with http status code 401, client: 172.18.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:8080"
api-gateway-1  | 172.18.0.1 - - [07/Nov/2024:06:41:12 +0000] 127.0.0.1:8080 "GET / HTTP/1.1" 401 52 0.006 "-" "curl/8.2.1" - - - "http://127.0.0.1:8080"

### Steps to Reproduce

services:
    api-gateway:
        image: apache/apisix:latest
        environment:
            - APISIX_STAND_ALONE=true
            - LUA_SSL_TRUSTED_CERTIFICATE=/usr/local/share/ca-certificates/vault-ca.crt
        volumes:
            - ${CONFIGS:-./configs}/apisix/apisix.yaml:/usr/local/apisix/conf/apisix.yaml:ro
            - ./vault_ca.crt:/usr/local/share/ca-certificates/vault-ca.crt  # Mount the CA cert into the container
        extra_hosts:
            - "vault.mydomain.com:192.168.10.60"
        ports:
            - '${LISTEN_ADDRESS:-127.0.0.1}:8080:9080'
            - '${LISTEN_ADDRESS:-127.0.0.1}:8443:9443'
        networks:
            - public
networks:
   public:
    external: true
    
    
   curl -H 'Authorization:asdfasdfasdfas' -H "Content-Type: application/json" -i http://127.0.0.1:8080   

### Environment

 docker exec -it  apisix-api-gateway-1 bash                                                                                                                                                                                        ❌1 00:01
apisix@b7e90f1785f2:/usr/local/apisix$ apisix version
/usr/local/openresty//luajit/bin/luajit ./apisix/cli/apisix.lua version
3.11.0
apisix@b7e90f1785f2:/usr/local/apisix$ uname -a
Linux b7e90f1785f2 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 GNU/Linux
apisix@b7e90f1785f2:/usr/local/apisix$ openresty -V` or `nginx -V`
> ^C
apisix@b7e90f1785f2:/usr/local/apisix$ 'penresty -V` or `nginx -V`
> ^C
apisix@b7e90f1785f2:/usr/local/apisix$ 'Openresty -V` or `nginx -V`
> ^C
apisix@b7e90f1785f2:/usr/local/apisix$ `Openresty -V` or `nginx -V`
bash: Openresty: command not found
nginx version: openresty/1.25.3.2
built by gcc 10.2.1 20210110 (Debian 10.2.1-6)
built with OpenSSL 3.2.0 23 Nov 2023
TLS SNI support enabled
configure arguments: --prefix=/usr/local/openresty/nginx --with-cc-opt='-O2 -DAPISIX_RUNTIME_VER=1.2.1 -DNGX_LUA_ABORT_AT_PANIC -I/usr/local/openresty/zlib/include -I/usr/local/openresty/pcre/include -I/usr/local/openresty/openssl3/include' --add-module=../ngx_devel_kit-0.3.3 --add-module=../echo-nginx-module-0.63 --add-module=../xss-nginx-module-0.06 --add-module=../ngx_coolkit-0.2 --add-module=../set-misc-nginx-module-0.33 --add-module=../form-input-nginx-module-0.12 --add-module=../encrypted-session-nginx-module-0.09 --add-module=../srcache-nginx-module-0.33 --add-module=../ngx_lua-0.10.26 --add-module=../ngx_lua_upstream-0.07 --add-module=../headers-more-nginx-module-0.37 --add-module=../array-var-nginx-module-0.06 --add-module=../memc-nginx-module-0.20 --add-module=../redis2-nginx-module-0.15 --add-module=../redis-nginx-module-0.3.9 --add-module=../ngx_stream_lua-0.0.14 --with-ld-opt='-Wl,-rpath,/usr/local/openresty/luajit/lib -Wl,-rpath,/usr/local/openresty/wasmtime-c-api/lib -L/usr/local/openresty/zlib/lib -L/usr/local/openresty/pcre/lib -L/usr/local/openresty/openssl3/lib -Wl,-rpath,/usr/local/openresty/zlib/lib:/usr/local/openresty/pcre/lib:/usr/local/openresty/openssl3/lib' --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../mod_dubbo-1.0.2 --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../ngx_multi_upstream_module-1.2.0 --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../apisix-nginx-module-1.16.1 --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../apisix-nginx-module-1.16.1/src/stream --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../apisix-nginx-module-1.16.1/src/meta --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../wasm-nginx-module-0.7.0 --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../lua-var-nginx-module-v0.5.3 --add-module=/tmp/tmp.0vt0zLPiwq/openresty-1.25.3.2/../lua-resty-events-0.2.0 --with-poll_module --with-pcre-jit --with-stream --with-stream_ssl_module --with-stream_ssl_preread_module --with-http_v2_module --with-http_v3_module --without-mail_pop3_module --without-mail_imap_module --without-mail_smtp_module --with-http_stub_status_module --with-http_realip_module --with-http_addition_module --with-http_auth_request_module --with-http_secure_link_module --with-http_random_index_module --with-http_gzip_static_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-threads --with-compat --with-stream --without-pcre2 --with-http_ssl_module
bash: or: command not found

apisix@b7e90f1785f2:/usr/local/apisix$ luarocks --version
bash: luarocks: command not found

Revolyssup:
### Description
When testing AI plugins, we found that `ssl_trusted_certificate` should be set to `system`, otherwise APISIX will continuously report errors when accessing external AI services.
```
unable to get local issuer certificate
```

Further refactoring:
Currently two ways are used for reading configuration.
1. While generating nginx.conf file.read_yaml_conf is called 
2. During lifecycle, config_local.local_conf is called which calls read_yaml_conf internally.

The problem - Schema validation for configuration happens only in local_conf so at startup, the local conf schema is not validated. This is an inconsistency.

This is especially problematic when my change uses schema to set the default to "system" because this default is only set during lifecycle and not during generation of nginx.conf.

This PR moves the schema validation inside read_yaml_conf for consistency. Now local_conf is just calling read_yaml_conf internally and adding cache.

## Effect on PR of refactoring
This makes sure that consistently at all places when yaml file is read, first schema validation happens which sets the default value of trusted certificate to "system" and then later the overrides take place including the one which replaces "system" with cert paths.

Based on the discussion below, this PR also removes support for combining multiple certs
Fixes # (issue)

### Checklist

Fixes #11720


nic-6443 (dismissed):

apisix/cli/schema.lua
```diff
@@ -208,6 +208,7 @@ local config_schema = {
                     properties = {
                         ssl_trusted_certificate = {
                             type = "string",
+                            default = "system",
```
shreemaan-abhishek, line 211: please update the example yaml file too

Revolyssup: done

shreemaan-abhishek (dismissed):

nic-6443 (dismissed):

nic-6443 (dismissed):

conf/config.yaml.example
```diff
@@ -101,7 +101,7 @@ apisix:
       #   enable_http3: true
     ssl_trusted_combined_path: /usr/local/apisix/conf/ssl_trusted_combined.pem # All the trusted certificates will be combined into a single file
     #ssl_trusted_certificate: system              # Specifies comma separated list of trusted CA. Value can be either "system"(for using system available ca certs) or
-                                                # a file path with trusted CA certificates in the PEM format
+                                                # a file path with trusted CA certificates in the PEM format. The default value is "system".
```
membphis, line 104: add one more space

image

membphis (dismissed):

apisix/cli/ops.lua
```diff
@@ -529,7 +531,7 @@ Please modify "admin_key" in conf/config.yaml .
         end
 
         local combined_cert_filepath = yaml_conf.apisix.ssl.ssl_trusted_combined_path
-                                       or "/usr/local/apisix/conf/ssl_trusted_combined.pem"
+                                       or profile.apisix_home .. "/conf/ssl_trusted_combined.pem"
```
bzp2010, line 534: Why don't we use the system's CA file directly when it is set to `system`, but use the one processed by APISIX itself?

Revolyssup: we do use the system CA file. This path is to store the combined cert only.

Revolyssup: okay i got what you were saying. let me make that change

bzp2010: Specifically, we should hardcode some list of paths based on the default system certificate paths of some of the major Linux distributions (Debian, Redhat) and find and use them at boot time.
As currently designed, this file will be copied and generated in /usr/local/apisix, a place where permissions management is typically more complex than in /etc/ssl/xxx.

This introduces security risks to APISIX that are not otherwise within its administrative purview. This is bad if a malicious user is not authorized to modify /etc and it can modify /usr/local/apisix.
A TLS CA file silently copied by APISIX to authenticate an authoritative certificate chain could be corrupted by a malicious user, unbeknownst to APISIX itself or the user, jeopardizing the user's system.

bzp2010: As long as the system's own ca bundle is used, we just comply with whatever changes it makes, even though it may be broken as well, and that's not APISIX's problem.

bzp2010: To add a trusted certificate, the user simply follows the general guide of placing the new certificate somewhere, and later uses `sudo update-ca-certificates` to update the system certificate chain.

bzp2010: Alternatively, instead of using the system certificate chain at all, users can copy that system certificate themselves or start with a blank file, add their own self-signed CA to the end of the file, and specify a custom file directory inside the APISIX configuration file.

The user will know what they are doing and what the risks are, this cannot be a black box.

bzp2010: I have no problem with the main thrust of that PR, i.e., the chain of trust system certificates, but we would be better off not copying it.

nic-6443: >  start with a blank file, add their own self-signed CA to the end of the file, and specify a custom file directory inside the APISIX configuration file

In this case, the usage cost will increase, especially for containerized deployment. Users need to build their own image based on the APISIX container image or write corresponding shell scripts to complete this task before starting the APISIX image. Both methods are somewhat troublesome.
Imagine if the etcd deployed by users has mTLS enabled, in order to use custom CA certificates and system certificates simultaneously, they would need to handle the above steps themselves when deploying APISIX, which becomes inconvenient.

> he user will know what they are doing and what the risks are, this cannot be a black box.

We can explain the actual functions of `ssl_trusted_certificate` and `ssl_trusted_combined_path` in the documentation for `config.yaml`. Just as APISIX generates `nginx.conf` at startup, we can explain the function of those temporary files.

nic-6443: > As currently designed, this file will be copied and generated in /usr/local/apisix, a place where permissions management is typically more complex than in /etc/ssl/xxx.

yes, we should put it in `$profile.apisix_home/conf` like `nginx.conf`, using `/usr/local/apisix` is a mistake. cc @Revolyssup

bzp2010: @nic-6443 

> In this case, the usage cost will increase, especially for containerized deployment. Users need to build their own image based on the APISIX container image or write corresponding shell scripts to complete this task before starting the APISIX image. Both methods are somewhat troublesome.
Imagine if the etcd deployed by users has mTLS enabled, in order to use custom CA certificates and system certificates simultaneously, they would need to handle the above steps themselves when deploying APISIX, which becomes inconvenient.

I don't agree with the idea that costs will go up, if a user decides to turn on mTLS and the server doesn't use trusted certificates, it should handle any resulting extra work itself, which should follow any debian or redhat guidelines that exist as well, more people will learn about standard packages (ca-certificates) than will learn and understand us weird black box customization mechanisms.
For debian, first make sure the ca-certificates package is installed on your system, then place your own CA certificates in /usr/local/share/ca-certificates (the container will need to be mounted), and then just run sudo update-ca-certificates and those customized certificates will be automatically added to the root certificate package by the script.

For containers, this can be done by custom ENTRYPOINT or CMD, such as 
`docker run -d apache/apisix:3.11.0-debian sh -c "update-ca-certificates && apisix start"` (just an example), this doesn't require a script. I'm doing this for a couple containers and it's working fine.

The difficulties you mention should be improved by some better docker entrypoint scripts, for example, by providing some directory to place and execute user's custom scripts before startup, as many are doing https://docs.linuxserver.io/general/container-customization/#custom-scripts. this is outside the scope of APISIX, and it's enough for APISIX to ensure that its behavior is safe.

nic-6443: I maintain my viewpoint that, from my perspective, the design of allowing `ssl_trusted_certificate` to support multiple files is very intuitive.
Based on this, in order for nginx/openresty's `ssl_trusted_certificate` to correctly read multiple files, we have no choice but to merge the files. 
This logic is very simple and can be clearly explained in the description of this configuration, so I don't consider it a black-box design.
It is merely an encapsulation and enhancement of nginx/openresty's `ssl_trusted_certificate`.
@membphis @moonming @juzhiyuan Need more discussion here.

bzp2010: This is not a rebuttal to the “simple wrapping” statement, it is simple wrapping. The problem is that we shouldn't do it at all.

Any action other than _**reading**_ the system certificate file (when this value is set to `system`) is outside the scope of the APISIX project. Modifying the system certificate chain is a matter for the user's system administrators to deal with, not ours.

### **In short, we do not endorse or take responsibility for any potential security risks from this behavior.**

If the file being copied is controlled by APISIX in terms of content and lifecycle, the responsibility for ensuring its security is APISIX's. With great power comes great responsibility.
Unless the user reads the code, no one will know that we have copied the file and placed it in the unintended unknown, and I think it's fair for me to call it a black box. This is a serious discussion for any security and trust issues.

System administrators are familiar with the way certificates are managed in the Linux distributions their companies use. The `ca-certificates` is a well known software package, and there are many documents on how to work with it. And what about APISIX's documentation of this? I don't think there's more than that on the config file.

---

There is also more than one way to modify the certificate chain, as I've already mentioned.

1. Follow the Linux guide and use the `update-ca-certificates` command
2. Point `ssl_trusted_certificate` to a CA certificate bundle file that you know exactly what it is and want to trust.

It's intuitive, and whether users are more familiar with ca-certificates or nginx they will know what should be done.

apisix/cli/ops.lua
```diff
@@ -503,38 +501,27 @@ Please modify "admin_key" in conf/config.yaml .
...
+            util.die(err)
         end
-
-        local combined_cert_filepath = yaml_conf.apisix.ssl.ssl_trusted_combined_path
-                                       or "/usr/local/apisix/conf/ssl_trusted_combined.pem"
-        util.gen_trusted_certs_combined_file(combined_cert_filepath, cert_paths)
-
-        yaml_conf.apisix.ssl.ssl_trusted_certificate = combined_cert_filepath
+        yaml_conf.apisix.ssl.ssl_trusted_certificate = trusted_certs_path
+    elseif yaml_conf.apisix.ssl.ssl_trusted_certificate ~= nil then
```
nic-6443, line 513: This condition is redundant, the default value for `ssl_trusted_certificate` has already been set above.

Revolyssup: fixed

nic-6443 (dismissed):

membphis (dismissed):
LGTM

apisix/cli/ops.lua
```diff
@@ -503,38 +501,27 @@ Please modify "admin_key" in conf/config.yaml .
...
-                -- Therefore we need to check the absolute version instead
-                cert_path = pl_path.abspath(cert_path)
-                if not pl_path.exists(cert_path) then
-                    util.die("certificate path", cert_path, "doesn't exist\n")
-                end
-
-                table.insert(cert_paths, cert_path)
-            end
+    if not yaml_conf.apisix.ssl.ssl_trusted_certificate then
+        yaml_conf.apisix.ssl.ssl_trusted_certificate = "system"
```
bzp2010, line 505: As it relates to default value changes, this should be added to the default values table in `apisix/cli/config.lua` rather than only setting the variable directly here. if that table is also used here, we can reduce codes.

apisix/cli/schema.lua
```diff
@@ -209,9 +209,6 @@ local config_schema = {
                         ssl_trusted_certificate = {
                             type = "string",
```
bzp2010, line 210: Do we also need to set system to the json schema default?

apisix/cli/ops.lua
bzp2010: - Add default to `apisix/cli/config.lua`
- Check if `ops.lua` is using `conf/config.yaml` and `apisix/cli/config.lua` aggregation
  -  if so, remove the assignment here

apisix/cli/schema.lua
Revolyssup: I think if we add it to apisix/cli/config.lua. Its fine to not add this here.

apisix/cli/ops.lua
Revolyssup: done

apisix/cli/schema.lua
bzp2010: I still recommend adding it so that we can explore the feasibility of generating documents for it in the future.
Although we don't need it to fill us with default values, as a specification it should fit the status quo.

apisix/core/config_local.lua
```diff
@@ -65,12 +63,6 @@ function _M.local_conf(force)
         return nil, err
     end
 
-    -- fill the default value by the schema
```
nic-6443, line 68: why remove this?

bzp2010: @nic-6443 

https://github.com/apache/apisix/pull/11993#discussion_r1982986021
https://github.com/apache/apisix/pull/11993#discussion_r1982994653

We have a two guarantees, so this is not necessary.

Revolyssup: This has been moved to read_yaml_conf. Please check description for reason

nic-6443 (approved):

membphis (approved):

bzp2010 (approved):

PR #12004, opened 2025-02-27
feat(ai-proxy): support proxying openai compatible LLMs

shreemaan-abhishek:
### Description

This PR introduces a new provider `openai-compatible` that can proxy requests to openai compatible LLMs.

### Checklist


nic-6443 (dismissed):

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    if extra_opts.model_options then
+        for opt, val in pairs(extra_opts.model_options) do
+            request_table[opt] = val
+        end
+    end
+    params.body = core.json.encode(request_table)
+
+    httpc:set_timeout(conf.keepalive_timeout)
```
AlinsRan, line 98: >     local httpc, err = http.new()
>    if not httpc then
>        return nil, "failed to create http client to send request to LLM server: " .. err
>    end
>    httpc:set_timeout(conf.timeout)

Why set two timeouts?

the `keepalive_timeout` should be the maximum idle time, which one should be called here?
https://github.com/ledgetech/lua-resty-http?tab=readme-ov-file#set_keepalive

shreemaan-abhishek: `set_timeout` is correct. And it's needed in cases where the LLMs takes too long to respond.

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -0,0 +1,107 @@
...
+        path = path,
+        query = query_params
+    }
+
+    if extra_opts.model_options then
+        for opt, val in pairs(extra_opts.model_options) do
+            request_table[opt] = val
+        end
+    end
+    params.body = core.json.encode(request_table)
```
membphis, line 96: it may fail, need to capture the error case

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -84,8 +84,7 @@ local model_schema = {
         provider = {
             type = "string",
             description = "Name of the AI service provider.",
-            oneOf = { "openai" }, -- add more providers later
-
+            enum = { "openai", "openai-compatible" }, -- add more providers later
```
membphis, line 87: missing `deepseek`?

apisix/plugins/ai-drivers/openai-base.lua
AlinsRan: If it is a read timeout, why not use `set_timeouts` directly?
The name `keepalive_timeout` made me misunderstand that it is related to nginx connection pool management.

apisix/plugins/ai-drivers/openai-base.lua
shreemaan-abhishek: done

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: fixed

apisix/plugins/ai-drivers/openai-base.lua
AlinsRan: The same name, but the behavior of this PR is completely different from other plugins.
https://github.com/apache/apisix/blob/master/apisix/plugins/forward-auth.lua#L135

shreemaan-abhishek: > If it is a read timeout, why not use set_timeouts?

I think `set_timeout` is just fine. And it is in-line with other plugins/implementations.

nic-6443: I can't understand why set timout twice too

shreemaan-abhishek: okay, I finally understand what you meant. I have fixed this now.

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):

PR #12040, opened 2025-03-11
fix: consumer key duplication check

### LINKED ISSUE #11197: API KEY Unique
### Description

API KEY can be use for purpose Authentication, to do that API KEY need require to unique  for all Consumer

Baoyuantop:
### Description
This PR adds a duplicate consumer key check when creating consumers and credentials using the admin api to prevent different consumers from having the same key.

Fixes #11197

### Checklist


t/admin/credentials.t
```diff
@@ -110,7 +110,7 @@ passed
                      "desc": "basic-auth for jack",
                      "plugins": {
                          "basic-auth": {
-                             "username": "the-user",
+                             "username": "the-new-user",
```
Baoyuantop, line 113: After adding the duplicate judgment, it cannot be added here again, so I made a modification.

t/secret/aws.t
```diff
@@ -314,3 +314,5 @@ GET /t
     }
 --- response_body
 all done
+--- error_log
```
Baoyuantop, line 317: The testing of these changes is an expected change. These tests have an admin api call to the consumer and use a non-existent secret reference, which in the new logic triggers a secret lookup failure and an error log. It didn't break the original test path.

apisix/admin/consumers.lua
```diff
@@ -32,7 +34,50 @@ local function check_conf(username, conf, need_username, schema)
     if conf.plugins then
         ok, err = plugins.check_schema(conf.plugins, core.schema.TYPE_CONSUMER)
         if not ok then
-            return nil, {error_msg = "invalid plugins configuration: " .. err}
+            return nil, {
+                error_msg = "invalid plugins configuration: " .. err
```
shreemaan-abhishek, line 38: we can keep this line unchanged.

apisix/admin/consumers.lua
```diff
@@ -32,7 +34,50 @@ local function check_conf(username, conf, need_username, schema)
...
+                local key_field = plugin_key_map[plugin_name]
+
+                if key_field then
+                    local key_value = decrypted_conf[key_field]
+
+                    if key_value then
+                        local consumer, _, err = consumer
+                            .find_consumer(plugin_name, key_field, key_value)
+                        if err then
+                            core.log.warn("failed to find consumer: ", err)
```
shreemaan-abhishek, line 69: I don't understand the logic here. In the `find_consumer()` method call above. We are checking if the given consumer key already exists, no? If it doesn't exist then the duplication check should be successful (no duplicate found), right?

If yes, then there is no point in printing a warn log.

apisix/admin/consumers.lua
```diff
@@ -32,7 +34,50 @@ local function check_conf(username, conf, need_username, schema)
...
+                        end
+
+                        if consumer and consumer.username ~= conf.username then
```
shreemaan-abhishek, line 72: 1. Is it guaranteed that the returned consumer will have the field `username` in it? Different consumers have different key fields right? Take your own code snippet for example:
image

2. This block will execute only if a consumer with given `key_field` already exists. So, if it does, we should directly raise an error, no?

Baoyuantop: The purpose of this is to allow the user to edit a pre-existing consumer. If the consuner being edited matches the username of the consumer being looked up, then the same consumer is being edited, and this behavior should be allowed.

shreemaan-abhishek: got it. What about my first question?

> Is it guaranteed that the returned consumer will have the field username in it? Different consumers have different key fields right? Take your own code snippet for example:

apisix/admin/consumers.lua
shreemaan-abhishek: I think you can add this info as a comment in the code.

apisix/admin/consumers.lua
Baoyuantop: username is a unique field for the consumer and must exist. https://apisix.apache.org/docs/apisix/terminology/consumer/#configuration-options

apisix/admin/consumers.lua
```diff
@@ -34,6 +38,40 @@ local function check_conf(username, conf, need_username, schema)
...
+                local key_field = utils.plugin_key_map[plugin_name]
+
+                if key_field then
+                    local key_value = plugin_conf_copy[key_field]
+
+                    if key_value then
+                        local consumer, _, err = consumer
+                            .find_consumer(plugin_name, key_field, key_value)
+                        if err then
+                            core.log.warn("failed to find consumer: ", err)
```
AlinsRan, line 62:
```suggestion
                        local consumer = consumer.find_consumer(plugin_name, key_field, key_value)
                        if consumer then
                           ...
```

AlinsRan: The expected outcome is that it cannot be found, so I believe there is no need to record any logs.

apisix/admin/credentials.lua
```diff
@@ -32,14 +34,41 @@ local function check_conf(_id, conf, _need_id, schema)
...
             if not plugin_obj then
                 return nil, {error_msg = "unknown plugin " .. name}
             end
+
             if plugin_obj.type ~= "auth" then
                 return nil, {error_msg = "only supports auth type plugins in consumer credential"}
             end
+
+            -- check duplicate key
+            local decrypted_conf = core.table.deepcopy(plugin_conf)
```
AlinsRan, line 48: These two pieces of code are basically the same. I suggest using a universal function.

AlinsRan (commented):
Do we not need to check for duplicate keys when watching etcd data?

Baoyuantop:
> Do we not need to check for duplicate keys when watching etcd data?

Do you mean that etcd can have duplicate keys written to it by other components ? @AlinsRan

apisix/admin/credentials.lua
```diff
@@ -32,14 +34,37 @@ local function check_conf(_id, conf, _need_id, schema)
...
```
AlinsRan, line 48: There are too many nested judgments inside, but I think we can extract this logic into a universal function that can be used by both consumers and credentials.

AlinsRan:
> > Do we not need to check for duplicate keys when watching etcd data?
> 
> Do you mean that etcd can have duplicate keys written to it by other components ? @AlinsRan



> > Do we not need to check for duplicate keys when watching etcd data?
> 
> Do you mean that etcd can have duplicate keys written to it by other components ? @AlinsRan

Components that do not use admin API, such as dashboard.

apisix/admin/consumers.lua
```diff
@@ -17,8 +17,51 @@
...
+        if not plugin_obj then
+            return nil, "unknown plugin " .. plugin_name
+        end
+
+        if plugin_obj.type ~= "auth" then
+            goto continue
+        end
+
+        local plugin_conf_copy = core.table.deepcopy(plugin_conf)
+        plugin.decrypt_conf(plugin_name, plugin_conf_copy, core.schema.TYPE_CONSUMER)
```
nic-6443, line 42: Is this code useful? How can the input of the admin API be encrypted?

t/admin/credentials.t
```diff
@@ -492,3 +492,138 @@ GET /t
...
+=== TEST 23: delete consumer jack
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/consumers/jack', ngx.HTTP_DELETE)
+        }
+    }
+--- request
+GET /t
```
nic-6443, line 629: we don't need clean up thoes resources manually, the test framework will delete all keys in etcd after each run

apisix/admin/consumers.lua
```diff
@@ -73,14 +70,14 @@ local function check_conf(username, conf, need_username, schema)
     end
 
     if conf.plugins then
-        ok, err = plugins.check_schema(conf.plugins, core.schema.TYPE_CONSUMER)
+      local ok, err = check_duplicate_key(conf.username, conf.plugins)
```
nic-6443, line 73: why change the order of those checks? I think we should keep checking static schema first indeed

Baoyuantop: The plugin's configuration will be encrypted during check_schema, and will need to be decrypted later.

nic-6443 (dismissed):

nic-6443 (dismissed):

Baoyuantop:
> > > Do we not need to check for duplicate keys when watching etcd data?
> > 
> > 
> > Do you mean that etcd can have duplicate keys written to it by other components ? @AlinsRan
> 
> Components that do not use admin API, such as dashboard.

I think it can be discussed in a new issue, and this PR only deals with the admin api section

apisix/admin/utils.lua
```diff
@@ -110,4 +112,51 @@ function _M.decrypt_params(decrypt_func, body, schema_type)
...
+
+local plugin_key_map = {
+    ["key-auth"] = "key",
+    ["basic-auth"] = "username",
+    ["jwt-auth"] = "key",
+    ["hmac-auth"] = "key_id"
+}
+
+
+function _M.check_duplicate_key(plugins_conf, username, credential_id)
```
AlinsRan, line 124: I suggest putting the `check_deuplicate_key` function in `apisix/consumer.lua`

AlinsRan (approved):

nic-6443 (approved):

moonming (changes requested):
I’m having trouble understanding all the test cases — they’re quite hard to read. Could you please use more meaningful names and descriptions to make them easier to follow?

dvdieu:
that good change thanks

Issue #10945, opened 2024-02-19
bug: The route plugin and consumer plugin are run repeatedly. Is this the correct result?


dbit-xia:
### Current State

apisix version: 3.4.1

https://apisix.apache.org/docs/apisix/3.4/terminology/plugin/
```
While configuring the same plugin, only one copy of the configuration is valid. The order of precedence is always Consumer > Consumer Group > Route > Plugin Config > Service.
```
see "only one copy of the configuration is valid"

During implementation, it was found that the same plug-in first executed the routing configuration plug-in, and then executed the consumer configuration plug-in. Is this correct?

### Desired State

Documentation is consistent with actual behavior

kayx23:
> During implementation, it was found that the same plug-in first executed the routing configuration plug-in, and then executed the consumer configuration plug-in. Is this correct?

What plugin is this and steps to reproduce?

dbit-xia:
I'm using a custom plugin, outputting logs during the rewrite phase for verification. 

```
local core = require("apisix.core")
local _M = {
    version = 0.1,
    priority = 22000,
    name = "wm-test",
    schema = {
      type = "object",
      properties = {},
    }
}

function _M.check_schema(conf)
    return core.schema.check(schema, conf)
end

function _M.rewrite(conf, ctx)
    core.log.error(core.json.encode(conf))
end

return _M
```

hanqingwu:
can you provide some logs about it  ?

dbit-xia:
The following two request logs: 
The first line `"max_body_size":1000` is the plugin configuration set in the route.
The second line `"max_body_size":2097152` is the plugin configuration set in the consumer.

```
2024/02/26 17:38:58 [error] 90#90: *144 [lua] wm-test.lua:48: phase_func(): trace_id: ac11e468b8dd269df4f9bec53394e6be {"_meta":{"disable":false},"max_body_size":1000}, client: 127.0.0.1, server: _, request: "POST /api3/page/save HTTP/1.1", host: "127.0.0.1:9080"
2024/02/26 17:38:58 [error] 90#90: *144 [lua] wm-test.lua:48: phase_func(): trace_id: ac11e468b8dd269df4f9bec53394e6be {"_meta":{"disable":false},"types":{"multipart/form-data":20000000},"max_body_size":2097152}, client: 127.0.0.1, server: _, request: "POST /api3/page/save HTTP/1.1", host: "127.0.0.1:9080"
```

dbit-xia:
[...]

Issue #10873, opened 2024-01-29
help request: when using `hmac-auth` and `proxy-rewrite` plugin simultaneously, why does the rewrite phase of `proxy-rewrite` gets executed twice? Is this a Bug?


FinerKeysen:
### Current Behavior

In apisix release/2.13 and release/3.2, when using `hmac-auth` and `proxy-rewrite` plugins simultaneously on a route, the rewrite phase of proxy-rewrite is observed to be executed twice.

### Error Logs

2024/01/29 13:04:36 [warn] 26073#26073: *2 [lua] plugin.lua:252: load_stream(): new plugins: {"mqtt-proxy":true,"ip-restriction":true,"limit-conn":true,"syslog":true}, context: init_worker_by_lua*
2024/01/29 13:04:36 [warn] 26071#26071: *60 [lua] config_etcd.lua:168: main etcd watcher started, revision=1374, context: ngx.timer
2024/01/29 13:04:36 [warn] 26074#26074: *80 [lua] config_etcd.lua:168: main etcd watcher started, revision=1374, context: ngx.timer
2024/01/29 13:04:36 [warn] 26073#26073: *93 [lua] config_etcd.lua:168: main etcd watcher started, revision=1374, context: ngx.timer
2024/01/29 13:05:29 [warn] 26073#26073: *3478 [lua] proxy-rewrite.lua:261: phase_func(): 开始 》》》 proxy rewrite, client: 127.0.0.1, server: _, request: "GET /api/v1/books?id=2 HTTP/1.1", host: "127.0.0.1:27151"
2024/01/29 13:05:29 [warn] 26073#26073: *3478 [lua] proxy-rewrite.lua:351: phase_func(): 结束 << << << proxy rewrite, client: 127.0.0.1, server: _, request: "GET /api/v1/books?id=2 HTTP/1.1", host: "127.0.0.1:27151"
2024/01/29 13:05:29 [warn] 26073#26073: *3478 [lua] proxy-rewrite.lua:261: phase_func(): 开始 》》》 proxy rewrite, client: 127.0.0.1, server: _, request: "GET /api/v1/books?id=2 HTTP/1.1", host: "127.0.0.1:27151"
[...]

PR #12048, opened 2025-03-13
fix: plugins repeats the rewrite phase when consumer configs with a plugin

hachi029:
### Description
plugins repeats the rewrite phase when consumer configs with a plugin.  

Fixes # (issue)

 #10945 #10873

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -494,31 +494,34 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
             goto continue
         end
 
-        if not check_disable(plugin_conf) then
-            if plugin_obj.run_policy == "prefer_route" and route_plugin_conf ~= nil then
-                local plugin_conf_in_route = route_plugin_conf[name]
-                local disable_in_route = check_disable(plugin_conf_in_route)
-                if plugin_conf_in_route and not disable_in_route then
-                    goto continue
-                end
-            end
-
-            if plugin_conf._meta and plugin_conf._meta.priority then
-                custom_sort = true
-            end
-
-            -- in the rewrite phase, the plugin executes in the following order:
-            -- 1. execute the rewrite phase of the plugins on route(including the auth plugins)
-            -- 2. merge plugins from consumer and route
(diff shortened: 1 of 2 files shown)


membphis (approved):

apisix/plugin.lua
```diff
@@ -506,6 +506,16 @@ function _M.filter(ctx, conf, plugins, route_conf, phase)
             if plugin_conf._meta and plugin_conf._meta.priority then
                 custom_sort = true
             end
+
```
nic-6443, line 509: how about move this check to 
https://github.com/apache/apisix/blob/52d8fea3dbc93c4b59ac6682902955c6146af4ad/apisix/plugin.lua#L1184-L1186
as 
```
if plugins[i].type == "auth" or (not plugins[i + 1]._from_consumer) then
    plugins[i + 1]._skip_rewrite_in_consumer = true
end
```
so that we can keep skip logic in one place.

hachi029: Ok. Is it better move  
```
        if plugins[i].type == "auth" then
            plugins[i + 1]._skip_rewrite_in_consumer = true
        end
```
into _M.filter(...) ? because it is the function _M.filter determines or marks whether a plugin will be executed or not .

nic-6443: I found that there is a place where the `run_plugin` function is executed without executing `filter` first: https://github.com/apache/apisix/blob/52d8fea3dbc93c4b59ac6682902955c6146af4ad/apisix/init.lua#L437-L451
so I am not sure if moving it to `filter` will cause any issues.

hachi029: these modifications only affect the 'rewrite_in_consumer' phase.

nic-6443 (approved):

t/node/plugin1.t
```diff
@@ -0,0 +1,104 @@
+#
```
membphis, line 1: You can make a try, give this file a meaningful name

membphis (approved):

Issue #12028, opened 2025-03-08
feat: As a user, I want to custom oidc session, so that update openid-connect.lua


mengxzh:
### Description

when i want to custom oidc session, i found this code looks like already supported

<img width="774" alt="Image" src="https://github.com/user-attachments/assets/6c7d93a9-fe68-409b-a400-02f5c2cb1dec" />

but when i use below code to update custom session use  admin api

`
curl --request PATCH \
  --url http://127.0.0.1:9180/apisix/admin/routes/oauth2-resource-server \
  --header 'X-API-KEY: eZTssHqsNuUlJXzkCPqbhhWRxwVsZCRm' \
  --header 'content-type: application/json' \
  --data '{
	"plugins": {
		"openid-connect": {
			"bearer_only": false,
			"client_id": "oauth2-test",
			"client_secret": "123456",
			"discovery": "http://192.168.0.29:9303/.well-known/openid-configuration",
			"introspection_endpoint_auth_method": "client_secret_basic",
			"post_logout_redirect_uri": "http://127.0.0.1:9080",
			"realm": "apisix",
			"redirect_uri": "http://127.0.0.1:9080/call-back",
			"session": {
				"secret": "12e16f870756980a46d6c888086fe84afb96458616e175161fe3230e4fc4ae78",
				"redis": {
					"host": "192.168.0.29",
					"port": "6379",
					"prefix": "apisix-sessions:",
					"database": 2
				}
			}
		}
	}
}'`

**And got a response like this** 

![Image](https://github.com/user-attachments/assets/7692a5f8-3db5-4ecc-a4f9-f574f776caed)

**due to below check**

<img width="758" alt="Image" src="https://github.com/user-attachments/assets/1fbeb788-6473-43bf-b91d-c15c2fa0c9a9" />

how could i custom oidc session?

Baoyuantop:
What version are you using?

mengxzh:
@Baoyuantop version: 3.11

Baoyuantop:
https://github.com/apache/apisix/blob/release/3.11/apisix/plugins/openid-connect.lua#L67-L87
There is an additionalProperties restriction in this configuration that does not allow other fields, which is why you are returning errors with the admin api.

mengxzh:
@Baoyuantop Yes, cuz this restriction as you say, so I could not custom session where i store. But lua-resty-session has variety for session storage. I want use it, but in apisix could not set. If lift this limitaion, it will be flexible

[...]

PR #12051, opened 2025-03-13
feat: openid-connect plugin: Support extra session config options, remove unused option

### LINKED ISSUE #12050: bug: openid-connect plugin: the session.cookie.lifetime has no effect
### Current Behavior

The session cookie set by the openid-connect plugin is always set as a "session cookie" - the lifetime option has no effect

### Expected Behavior

the resty session lib doesn't support setting the lifetime of the cookie, so i guess remove that option?

### Error Logs

_No response_

### Steps to Reproduce

use the openid-connect plugin, log in, look at the Set-Cookie returned by the request to .apisix/redirect at the end of the flow

### Environment

arthurdarcet:
### Description

In the openid-connect plugin:
- Remove the session.cookie.lifetime option: that option doesn't exist in the lua-resty-session lib, and has no impact
- allow extra properties to support all the possible config options supported by the underlying lib

Needed because we need to change the cookie domain to set it to the base tld, and not the subdomain of the initial request, so that different product can share auth status.

Without that cookie_domain option, each subdomain needs to trigger an auth flow which are extra un-needed redirections for the user, and more importantly, if one subdomain is set to "unauth_action = pass", the user will be logged out on that domain until they explicitly log in even if they are actually already logged in on other subdomains.

Change is backward compatible because the allowed extra props allow the removed option to be still present.

Fixes
#12050 
#12028

### Checklist

- [*] I have explained the need for this PR and the problem it solves
- [*] I have explained the changes or the new features added to this PR
- [*] I have updated the documentation to reflect this change
- [*] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -67,24 +67,16 @@ local schema = {
         },
         session = {
             type = "object",
+            description = [[configuration of the session cookie. Anything supported by
+            the lua-resty-session init constructor can be set in this object]],
             properties = {
                 secret = {
                     type = "string",
                     description = "the key used for the encrypt and HMAC calculation",
                     minLength = 16,
-                },
-                cookie = {
-                    type = "object",
-                    properties = {
-                        lifetime = {
-                            type = "integer",
-                            description = "it holds the cookie lifetime in seconds in the future",
-                        }
-                    }
                 }
             },
             required = {"secret"},
-            additionalProperties = false,
         },
         realm = {
             type = "string",


arthurdarcet:
I wasn't reading the right version of the code for the resty.session lib - sorry for the noise.

cookie.lifetime is supported, and the cookie domain can be configured by setting the `session.cookie.domain` conf option of the openid-connect plugin (which is allowed even if not documented).

I think the session options could be improved in that plugin, but that's not a blocker for us anymore (for instance, to set a custom cookie name, i would need to set the `session.name` option on the plugin, but the schema doesn't allow that (even though the underlying lib supports it). I would change this to accept anything in the `session` key, and document it only with a link to the exact version of the resty.session lib that is bundled to be both exhaustive and exact.

mengxzh:
Sorry, How custom config session in this plugin at version 3.10

arthurdarcet:
> @arthurdarcet Sorry, How custom config session using redis in this plugin at version 3.10

afaics, you can't. You will need to write a custom plugin that can delegate to the openid-connect plugin after setting a few global vars that the underlying [resty.session](https://github.com/bungle/lua-resty-session/blob/v3.10/README.md) plugin will pick up.

mengxzh:
@arthurdarcet You mean i should custom plugin which is priority higher than openid-connect? And then i should set global vars that oidc-connect could read it?

arthurdarcet:
@mengxzh this isn't really a support forum. I meant that you need to write a custom plugin to replace the existing openid-connect plugin, but can do that with very little custom code by importing the existing plugin into your new plugin

mengxzh:
@arthurdarcet Why lift this limitaion as your code? As a custom feature, it will be flexible.

arthurdarcet:
> @arthurdarcet Why lift this limitaion as your code? As a custom config feature, it will be flexible.

yes, feel free to open another PR to remove the additionalProps: false or to document all the possible options

fernando-arellano:
hello @arthurdarcet can you confirm how you were able to configure the domain.

```
  - config:
      access_token_in_authorization_header: true
      bearer_only: false
      client_id: ...
      client_secret: ...
      discovery: ...
      redirect_after_logout_uri: http://localhost:4200
      redirect_uri: http://localhost:4200/login/redirect
      scope: openid 48bd3841-c7fc-4dd2-b973-080d1d7e1b47/.default
      session:
        cookie:
          domain: example.com
        secret: "1234567890123456"
      use_jwks: true
    enable: true
```

but the cookie domain seems to be ignored after a successful login.

PR #12055, opened 2025-03-17
chore: rectify business logic/code in ai-proxy

shreemaan-abhishek:
### Description

This PR does the following fixes:

checks if the response content-type header is not nil before doing string.find
return 504 if timeout error occurs
weight is required when configuring upstream nodes in ai-proxy-multi
bypass nginx upstream in ai-proxy-multi
add debug logs for ai usage
remove max timeout limit in schema
return http code when read chunk causes error

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -453,6 +453,12 @@ end
 
 
 function _M.handle_upstream(api_ctx, route, enable_websocket)
+    -- some plugins(ai-proxy...) request upstream by http client directly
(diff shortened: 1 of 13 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

nic-6443 (approved):

AlinsRan (approved):

PR #12070, opened 2025-03-19
fix: timeout risk in usages of `lua-resty-aws`

shreemaan-abhishek:
### Description

When initializing the lua-resty-aws module, it may send requests to IDMS endpoint which may timeout in non-aws environment or just introduce delays in some cases. Thus it is good to remove the initialisation from the module level.

Also, it is unnecessary to initialise the module for each request. This problem has also been fixed which ensures that if any there are any delays in initialising the module, such delay would occur only once.

### Checklist

### DIFF
--- a/apisix/plugins/ai-aws-content-moderation.lua
+++ b/apisix/plugins/ai-aws-content-moderation.lua
@@ -14,8 +14,12 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+require("resty.aws.config") -- to read env vars before initing aws module
+
 local core = require("apisix.core")
-local aws_instance = require("resty.aws")()
+local aws = require("resty.aws")
+local aws_instance
+
 local http = require("resty.http")
 local fetch_secrets = require("apisix.secret").fetch_secrets
 
@@ -96,6 +100,9 @@ function _M.rewrite(conf, ctx)
 
     local comprehend = conf.comprehend
 
+    if not aws_instance then
+        aws_instance = aws()
+    end
     local credentials = aws_instance:Credentials({
         accessKeyId = comprehend.access_key_id,
         secretAccessKey = comprehend.secret_access_key,
--- a/apisix/secret/aws.lua
+++ b/apisix/secret/aws.lua
@@ -16,9 +16,12 @@
 --
 
 --- AWS Tools.
+require("resty.aws.config") -- to read env vars before initing aws module
+
 local core = require("apisix.core")
 local http = require("resty.http")
 local aws = require("resty.aws")
+local aws_instance
 
 local sub = core.string.sub
 local find = core.string.find
@@ -51,7 +54,9 @@ local _M = {
 }
 
 local function make_request_to_aws(conf, key)
-    local aws_instance = aws()
+    if not aws_instance then
+        aws_instance = aws()
+    end
 
     local region = conf.region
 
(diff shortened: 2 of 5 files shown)


apisix/plugins/ai-aws-content-moderation.lua
```diff
@@ -14,8 +14,12 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+require("resty.aws.config") -- to read env vars before initing aws module
+
 local core = require("apisix.core")
-local aws_instance = require("resty.aws")()
+local aws = require("resty.aws")
+local aws_instance
```
bzp2010, line 21: Just to confirm:
This basically just represents an http client wrapper right?
I mean, if a user configures this plugin on multiple routes and uses different credentials, they won't interfere with each other?

bzp2010: If you can confirm that the following call to `aws_instance:Credentials` does not save in or affect `aws_instance`, we can do so.

apisix/plugins/ai-aws-content-moderation.lua
```diff
@@ -14,8 +14,12 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+require("resty.aws.config") -- to read env vars before initing aws module
```
bzp2010, line 17: Do you want the default value of `false` for `AWS_EC2_METADATA_DISABLED` to cause access to IDMS to be turned off by default unless the user explicitly turns it on?

Can you give a more detailed argument for its use? 🤔

apisix/plugins/ai-aws-content-moderation.lua
shreemaan-abhishek: > I mean, if a user configures this plugin on multiple routes and uses different credentials, they won't interfere with each other?

correct

apisix/plugins/ai-aws-content-moderation.lua
shreemaan-abhishek: I added this line mainly as a standard practice. (To read AWS specific ENV vars when APISIX is deployed in AWS environment)

> Do you want the default value of false for AWS_EC2_METADATA_DISABLED to cause access to IDMS to be turned off by default unless the user explicitly turns it on?

No, I added this to ensure the IDMS query can be disabled if need be.

apisix/plugins/ai-aws-content-moderation.lua
bzp2010: Great!

apisix/plugins/ai-aws-content-moderation.lua
bzp2010: Therefore, it obeys those predefined environment variables anyway, and the initialization process for aws instance is moved to when the AWS API request actually occurs.
IDMS requests don't happen if the plugin configuration contains user-set AK/SK?

shreemaan-abhishek: > IDMS requests don't happen if the plugin configuration contains user-set AK/SK?

as per my last test, IDMS request happened anyway.

bzp2010: So this will block requests when API calls are incoming, or block the secret manager.
Unless the user explicitly specifies `AWS_EC2_METADATA_DISABLED` as `true` to disable IDMS request?

shreemaan-abhishek: > So this will block requests when API calls are incoming, or block the secret manager.

yes, but only for the first time.

bzp2010 (approved):

nic-6443 (approved):

Revolyssup (approved):

PR #12030, opened 2025-03-10
refactor(ai): ai-proxy and ai-proxy-multi

shreemaan-abhishek:
### Description

This PR accomplishes the following: 
1. Adjust priorities of ai plugins in order to incorporate upcoming ai-rate-limiting plugin.
2. Refactors ai-proxy and ai-proxy-multi, integrating their proxy logic with APISIX's phase concept to work in conjunction with the `ai-rate-limiting` plugin under development
3. Handles AI requests and responses accordingly, also obtaining usage data for AI provider

### Checklist

- [ ] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #12083, opened 2025-03-25
fix: ai-request-rewrite test

LiteSun:
### Description

In the ai-request-rewrite2.ts file, the TEST 3 case was incorrectly configured with the wrong plugin. It has now been corrected.

Fixes # (issue)

### Checklist

### DIFF
--- a/t/plugin/ai-request-rewrite2.t
+++ b/t/plugin/ai-request-rewrite2.t
@@ -188,19 +188,16 @@ override.endpoint is required for openai-compatible provider
                  [[{
                     "uri": "/anything",
                     "plugins": {
-                        "ai-proxy": {
+                        "ai-request-rewrite": {
+                            "prompt": "some prompt to test",
                             "auth": {
                                 "query": {
                                     "api_key": "apikey"
                                 }
                             },
-                            "model": {
-                                "provider": "openai",
-                                "name": "gpt-35-turbo-instruct",
-                                "options": {
-                                    "max_tokens": 512,
-                                    "temperature": 1.0
-                                }
+                            "provider": "openai",
+                            "options": {
+                                "model": "gpt-35-turbo-instruct"
                             },
                             "override": {
                                 "endpoint": "http://localhost:6724/test/params/in/overridden/endpoint?some_query=yes"


apisix/plugins/ai-request-rewrite.lua
```diff
@@ -124,13 +124,12 @@ local function request_to_llm(conf, request_table, ctx)
         model_options = conf.options
     }
 
-    local res, err, httpc = ai_driver:request(conf, request_table, extra_opts)
```
LiteSun, line 127: No need to manully close according to the docs 
https://github.com/openresty/lua-nginx-module#tcpsockclose
image

t/plugin/ai-request-rewrite2.t
```diff
@@ -188,19 +188,16 @@ override.endpoint is required for openai-compatible provider
                  [[{
                     "uri": "/anything",
                     "plugins": {
-                        "ai-proxy": {
+                        "ai-request-rewrite": {
```
nic-6443, line 191: It seems that this test is useless, as it even got the plugin wrong and still passed. We need to add request validation for this route to ensure `override` works, making this test case a bit more meaningful.

apisix/plugins/ai-request-rewrite.lua
AlinsRan: https://github.com/apache/apisix/blob/master/apisix/plugins/ai-drivers/openai-base.lua#L131

It seems that only two parameters were returned, why didn't painc occur before?

AlinsRan: I think at least one pass case is needed, and there should be no error level logs.

LiteSun: > https://github.com/apache/apisix/blob/master/apisix/plugins/ai-drivers/openai-base.lua#L131
> 
> It seems that only two parameters were returned, why didn't painc occur before?

This issue occurred because when I started development and created my branch, the function was returning three parameters. After this PR was merged: https://github.com/apache/apisix/pull/12030/files#diff-b9bb0990d3ea842c42b094837737acfb597e2c8bf3a976dda8be56c84ef5b3c0L109, the returned parameters were adjusted. The code in my branch hasn't been synchronized with upstream, which caused this inconsistency.

> I think at least one pass case is needed, and there should be no error level logs.

The test cases are already covered, and the tests cannot pass without making this change.

nic-6443 (approved):

AlinsRan (approved):

Baoyuantop (commented):
Please revise to a meaningful PR title

PR #12086, opened 2025-03-26
feat: add proxy_nginx_upstream param

LiteSun:
### Description

Fixes # (issue)

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -59,6 +59,7 @@ local tonumber        = tonumber
 local type            = type
 local pairs           = pairs
 local ngx_re_match    = ngx.re.match
+local HTTP_BAD_REQUEST = ngx.HTTP_BAD_REQUEST
 local control_api_router
 
 local is_http = false
@@ -451,6 +452,10 @@ local function common_phase(phase_name)
 end
 
 
+local function proxy_upstream(api_ctx)
+    -- core.log.error("proxy nginx upstream", api_ctx)
+    return HTTP_BAD_REQUEST
+end
 
 function _M.handle_upstream(api_ctx, route, enable_websocket)
     -- some plugins(ai-proxy...) request upstream by http client directly
@@ -532,6 +537,13 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
     -- handle on its own
     if api_ctx.matched_upstream and api_ctx.matched_upstream.scheme == "kafka" then
         return pubsub_kafka.access(api_ctx)
+    end
+
+    -- if proxy_nginx_upstream is true, then proxy the request to upstream
+    if api_ctx.proxy_nginx_upstream then
+        core.log.error("proxy nginx upstream")
+        proxy_upstream(api_ctx)
+        return
     end
 
     local code, err = set_upstream(route, api_ctx)
--- a/apisix/plugins/ai-request-rewrite.lua
+++ b/apisix/plugins/ai-request-rewrite.lua
@@ -174,6 +174,7 @@ end
 
 
 function _M.access(conf, ctx)
+    ctx.proxy_nginx_upstream = true
     local client_request_body, err = core.request.get_body()
     if err then
         core.log.warn("failed to get request body: ", err)


apisix/plugins/example-plugin.lua
```diff
@@ -85,6 +90,10 @@ function _M.access(conf, ctx)
     core.log.warn("plugin access phase, conf: ", core.json.encode(conf))
     -- return 200, {message = "hit example plugin"}
 
+    if conf.lua_proxy_upstream then
```
AlinsRan, line 93: This is a feature PR that has added plugin configuration and requires updating schema and documentation.

apisix/init.lua
```diff
@@ -550,6 +550,12 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
...
     set_upstream_headers(api_ctx, server)
 
+    -- lua proxy the request to upstream
+    if api_ctx.lua_proxy_upstream then
+        common_phase("before_proxy")
+        return
+    end
+
     -- run the before_proxy method in access phase first to avoid always reinit request
     common_phase("before_proxy")
```
AlinsRan, line 560:
```lua
common_phase("before_proxy")
if api_ctx.lua_proxy_upstream then
     return
end
```

apisix/init.lua
```diff
@@ -550,6 +550,12 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
 
     set_upstream_headers(api_ctx, server)
 
+    -- lua proxy the request to upstream
+    if api_ctx.lua_proxy_upstream then
```
nic-6443, line 554: I thought about the naming of the existing variables `bypass_nginx_upstream` and `lua_proxy_upstream`, and felt that their meanings are unclear. I think we can design them as two: 
* `bypass_nginx_upstream` to indicate bypassing nginx's upstream module
* `bypass_balancer` to indicate bypassing the existing balancer logic
cc @membphis

apisix/lua_proxy.lua
```diff
@@ -0,0 +1,165 @@
...
+
+local function build_request_opts(conf, ctx)
+    -- Get upstream server
+    local server = ctx.picked_server
+    if not server then
+        return nil, "no picked server"
+    end
+
+    local headers = core.request.headers(ctx)
+    -- When content-length is cleared, the HTTP server will automatically calculate and
```
nic-6443, line 41:
```suggestion
    -- When content-length is cleared, the HTTP client will automatically calculate and
```

apisix/lua_proxy.lua
```diff
@@ -0,0 +1,165 @@
...
+    end
+
+    local content_type = res.headers["Content-Type"]
+    core.response.set_header("Content-Type", content_type)
+
+    -- TODO: support event stream
+    if content_type and core.string.find(content_type, "text/event-stream") then
+        core.log.error("event stream is not supported")
+       return HTTP_INTERNAL_SERVER_ERROR
+    end
```
nic-6443, line 93: This already has example code in ai-proxy.

t/plugin/example.t
```diff
@@ -339,3 +339,56 @@ plugin body_filter phase, eof: false
...
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "uri": "/anything",
+                    "plugins": {
+                        "example-plugin": {
+                            "i": 0,
+                            "lua_proxy_upstream": true,
+                            "request_uri": "http://httpbin.org/get",
```
nic-6443, line 358: rename this `request_uri`  to indicate it's used to change the rseponse body, it make me confused with `upstream.nodes`

apisix/plugins/example-plugin.lua
```diff
@@ -107,6 +116,49 @@ function _M.access(conf, ctx)
...
+        end
+        core.log.warn("lua proxy upstream response: ", core.json.encode(body))
+        plugin_lua_body_filter(conf, ctx, body)
+    end
+end
+
+
+function _M.lua_body_filter(conf, ctx, body)
+    core.log.warn("plugin lua_body_filter phase, conf: ", core.json.encode(conf))
+    core.log.warn("plugin lua_body_filter phase, body: ", core.json.encode(body))
```
nic-6443, line 136: you can assert those log in test cases to confirm the body come from upstream is passed to `lua_body_filter` as expect.

t/plugin/example.t
```diff
@@ -339,3 +339,56 @@ plugin body_filter phase, eof: false
 plugin delayed_body_filter phase, eof: false
 plugin body_filter phase, eof: true
 plugin delayed_body_filter phase, eof: true
+
+
+
+=== TEST 14: lua body filter
```
nic-6443, line 345: We need to add more test cases that combine example-plugin with other plugins like proxy-rewrite, which modify downstream requests. This will help verify that the new Lua HTTP client way works with existing plugins in the rewrite/access phase.

apisix/lua_proxy.lua
```diff
@@ -0,0 +1,165 @@
...
+        scheme = server.scheme or ctx.upstream_scheme or "http",
+        host = server.domain or server.host,
+        port = server.port,
+        path = ctx.var.uri,
+        query = ctx.var.args,
+        method = core.request.get_method(),
+        headers = headers,
+        ssl_verify = conf.ssl_verify,
+        keepalive = conf.keepalive,
+        keepalive_timeout = conf.timeout,
```
AlinsRan, line 57: This `timeout` is unclear.
The meanings of connection timeout and idle connection timeout are different.
I suggest using `keepalive_timeout`

apisix/lua_proxy.lua
```diff
@@ -0,0 +1,165 @@
...
+
+    -- Set upstream URI
+    if ctx.var.upstream_uri ~= "" then
+        opts.path = ctx.var.upstream_uri
+    end
+
+    -- Get request body
+    local body, err = core.request.get_body()
+    if err then
+        core.log.error("failed to get request body: ", err)
```
AlinsRan, line 69: Why isn't there a return for this error?

apisix/plugins/example-plugin.lua
```diff
@@ -107,6 +116,49 @@ function _M.access(conf, ctx)
     return
 end
 
+
+function _M.before_proxy(conf, ctx)
+    core.log.warn("plugin before_proxy phase, conf: ", core.json.encode(conf))
```
AlinsRan, line 121: This is not a warm level log, please use core.json.delay_encode

apisix/plugins/example-plugin.lua
AlinsRan: ditto

bzp2010 (changes requested):
What is the mean of another complex mechanism introduced by this PR? It's introducing a new plugin phase again, and continuing to introduce new learning costs, and I can't accept it without **significant justification**.
> Now, even a professional APISIX developer would be hard pressed to say what APISIX phases are and how they correspond to NGINX phases.

How does it differ from performing HTTP requests at rewrite/access using `lua-resty-http`?
Has the solution been properly evaluated for performance differences with access phase + subrequests? This would also allow you to implement running external requests in the `header_filter` and `body_filter` phases, and there would be no need to introduce new request phases, just a new core API is enough.

apisix/init.lua
shreemaan-abhishek: I agree with this

Baoyuantop:
Hi @LiteSun, is this PR still being processed? Can it be marked as draft first?

PR #12126, opened 2025-04-08
fix: release healthchecker on 0 nodes

Revolyssup:
### Description
Follow up to this fix: https://github.com/apache/apisix/pull/12118
The healthchecker should also be released when new node count is 0.

### Checklist

### DIFF
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -327,6 +327,7 @@ function _M.set_by_route(route, api_ctx)
 
     local nodes_count = up_conf.nodes and #up_conf.nodes or 0
     if nodes_count == 0 then
+        release_checker(up_conf.parent)
         return HTTP_CODE_UPSTREAM_UNAVAILABLE, "no valid upstream node"
     end
 
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -327,7 +327,9 @@ function _M.set_by_route(route, api_ctx)
 
     local nodes_count = up_conf.nodes and #up_conf.nodes or 0
     if nodes_count == 0 then
-        release_checker(up_conf.parent)
+        if up_conf.parent and up_conf.parent.checker then
+            release_checker(up_conf.parent)
+        end
         return HTTP_CODE_UPSTREAM_UNAVAILABLE, "no valid upstream node"
     end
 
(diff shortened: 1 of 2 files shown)


apisix/upstream.lua
```diff
@@ -327,6 +327,9 @@ function _M.set_by_route(route, api_ctx)
 
     local nodes_count = up_conf.nodes and #up_conf.nodes or 0
     if nodes_count == 0 then
+        if up_conf.parent and up_conf.parent.checker then
```
nic-chen, line 330: should check the parent and checker in function `release_checker` ?

Revolyssup: ok sounds right. done

t/discovery/reset-healthchecker.t
```diff
@@ -112,3 +112,58 @@ ok
...
+            local httpc = http.new()
+            local res, err = httpc:request_uri(uri, {method = "GET", keepalive = false})
+            ngx.status = res.status
+            ngx.sleep(5)
+        }
+    }
+--- request
+GET /t
+--- timeout: 22
+--- no_error_log
```
nic-6443, line 167: assert `try to release checker:` log too.

apisix/upstream.lua
```diff
@@ -327,6 +330,7 @@ function _M.set_by_route(route, api_ctx)
 
     local nodes_count = up_conf.nodes and #up_conf.nodes or 0
     if nodes_count == 0 then
+        release_checker(up_conf.parent)
```
nic-6443, line 333: Why not skip creating a new checker for upstream with no nodes?  ref: 
https://github.com/apache/apisix/blob/1d3581c95287dbab0359fb9553e1b2db5affc7bb/apisix/upstream.lua#L142

Revolyssup: @nic-6443 Healthchecker is not created when node == 0. This is for when previously there were non zero nodes and healthchecker was created. Later when it goes to 0, the previous healthchecker is deleted. But healthchecker is not created in case of no nodes.

nic-chen (approved):

membphis (approved):

nic-6443 (approved):

PR #12151, opened 2025-04-16
feat: add mcp-bridge plugin

bzp2010:
### Description

This PR introduces a new plugin called mcp-bridge that allows users to convert any stdio-based mcp server to HTTP SSE-based.

It consists of two parts, a plugin for handling stdio and managing subprocess lifecycles, and a submodule for supporting mcp session management.

#### `mcp-bridge` plugin

The `mcp-bridge` plugin will start an HTTP SSE response when a request arrives and create a queue to cache MCP RPC calls from the client.
The plugin will also start a child process and prepare its stdio according to the configuration.
After starting the SSE response and the child process is ready, it enters a loop and in each loop takes the cached RPC call out of the queue and puts it into the stdin of the child process, then reads the response from the stdout and the stderr, and if the response can't be read starts the next round of the loop, and the process of reading the response fails fast.

#### MCP session management

This PR also includes a lightweight MCP session management module that will assign session ids and manage queues and ping timers on sessions. It uses shared dict to ensure that it works properly when SSE connections and RPC calls are not handled by the same nginx worker.

#### Additionally

The plugin has implemented prototype functionality for managing MCP server protocol conversions and proxies, but still has some shortcomings.

1. Right now, the MCP session will not be shared between multiple APISIX instances, if your APISIX cluster consists of multiple nodes, you have to configure the session stickiness correctly on the front LB to ensure that requests from the same client will always be forwarded to the same APISIX instance, only then it will work properly

2. Current MCP SSE connections are loop-driven, and although loops don't take up too many resources (where reading and writing stdio would be synchronous non-blocking calls), this is not efficient, and we need to connect to some message queues to make them event-driven and scalable in a clustered fashion.

3. The MCP session management module is just a prototype, we should also abstract another MCP proxy server module to support starting an MCP server inside APISIX to support advanced scenarios. And that proxy server module, it should be event-driven rather than loop-driven.

### Checklist


moonming (approved):

membphis (approved):

juzhiyuan (approved):

Flcwl:
@bzp2010 how ise it?

PR #12173, opened 2025-04-22
docs: update rocketmq-logger.md, add max req/resp body size attributes description(3.10-)

yuanrui01:
This PR updates the documentation to include descriptions for two new configuration attributes introduced in version 3.10: max_req_body_bytes and max_resp_body_bytes. These attributes allow users to limit the recorded size of request and response bodies when include_req_body or include_resp_body is enabled, respectively, with a default value of 524288 bytes. Bodies exceeding these limits are truncated, enhancing memory efficiency and system stability.


Baoyuantop:
I didn't find these two attributes in the code. 🤔

yuanrui01:
> I didn't find these two attributes in the code. 🤔

I found it in log-util.lua
conf.max_req_body_bytes
conf.max_resp_body_bytes
Maybe this is implemented as a public property of all logging plugins?

Baoyuantop:
The schema of the plugin also needs to be changed so that users can configure these two parameters in the plugin.

yuanrui01:
> The schema of the plugin also needs to be changed so that users can configure these two parameters in the plugin.

I can add these two properties directly under the plugin to take effect as well.
ex:
`
"plugins": {
       "rocketmq-logger": {
           ...
           "max_resp_body_bytes" : 1048576,
           "max_req_body_bytes ": 1048576
       }
    }
`
So you say modifying the schema, is it for integrity purposes? I read lua for the first time and i'm not familiar with the rules.

Baoyuantop:
> I can add these two properties directly under the plugin to take effect as well.

Did you verify that the function is normal?

In order for the plugin to support this capability, we need to provide some code and tests for the plugin, you can refer to https://github.com/apache/apisix/pull/11133

yuanrui01:
> > I can add these two properties directly under the plugin to take effect as well.
> 
> Did you verify that the function is normal?
> 
> In order for the plugin to support this capability, we need to provide some code and tests for the plugin, you can refer to #11133

Yes, it works. In fact, that was my original need, at first I was trying to control the size logging via include_req_body_expr and include_resp_body_expr, then realized I couldn't do it, so I went through the code for a while and found these two lovely attributes. I upgraded the version from 3.3 to 3.12 then turned on show specified size and the size could be truncated.

Basically, 
config: "plugins": { "rocketmq-logger": { ... "max_resp_body_bytes" : 30, "max_req_body_bytes ": 30} }
and i got the message's resposne body like this
truncate

Baoyuantop:
Hi @yuanrui01, it's not enough to just change the documentation, please add code and tests.

PR #12180, opened 2025-04-28
feat: support health checker for stream subsystem

nic-6443:
### Description

support health checker for stream subsystem

Fixes #

### Checklist


Revolyssup (dismissed):

Revolyssup (approved):

AlinsRan (dismissed):

apisix/cli/config.lua
```diff
@@ -109,7 +109,8 @@ local _M = {
         ["lrucache-lock-stream"] = "10m",
         ["plugin-limit-conn-stream"] = "10m",
         ["worker-events-stream"] = "10m",
-        ["tars-stream"] = "1m"
+        ["tars-stream"] = "1m",
+        ["upstream-healthcheck-stream"] = "10m"
```
bzp2010, line 113:
```suggestion
        ["upstream-healthcheck-stream"] = "10m",
```

t/stream-node/healthcheck.t
```diff
@@ -0,0 +1,290 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+BEGIN {
+    if ($ENV{TEST_EVENTS_MODULE} ne "lua-resty-events") {
+        $SkipReason = "Only for lua-resty-events events module";
+    }
+}
```
bzp2010, line 21: Is this necessary? If lua-resty-worker-event is used, events can't cross subsystems?

Revolyssup: For lua-resty-events, a second request need to be initiated to trigger the passive health check. Ref: https://github.com/apache/apisix/blob/1bac43a0517a675e4fe77145c2ad3d26f61033a0/t/node/healthcheck-passive-resty-events.t#L346.

For lua-resty-worker-events, when I add second request the test case is timing out, So the test case will be a litlte different based on which library is used. So I pinned the test to only lua-resty events.

Revolyssup (approved):

t/stream-node/healthcheck.t
bzp2010: Do we need additional test cases to make sure it works? As I understand it, the current use cases with special handling can also override the lua-resty-worker-event library? 🤔

Revolyssup: lua-resty-worker-event case requires a little different handling in test. Okay I will separate these into two tests.

Revolyssup (approved):

nic-chen (approved):

PR #12185, opened 2025-04-28
update rocketmq-logger schema and markdown , add max req/resp body size attributes description

yuanrui01:
This PR updates the documentation to include descriptions for two new configuration attributes introduced in version 3.10: max_req_body_bytes and max_resp_body_bytes. These attributes allow users to limit the recorded size of request and response bodies when include_req_body or include_resp_body is enabled, respectively, with a default value of 524288 bytes. Bodies exceeding these limits are truncated, enhancing memory efficiency and system stability.


yuanrui01:
later

PR #12200, opened 2025-05-08
feat: add healthcheck for sync configuration

Revolyssup:
image

### Description

- Standalone mode: This PR aims to add a healthcheck endpoint which can be used by to indicate where APISIX is ready and has loaded the configuration from clients such as ingress controller.
- Config provider etcd: The above endpoint can be used to get the healthcheck status which will be false until all the workers have first time loaded the configuration from etcd.

Refer: https://github.com/apache/apisix/pull/12179#pullrequestreview-2820918241
### Checklist


apisix/cli/config.lua
```diff
@@ -176,6 +176,7 @@ local _M = {
         ["cas-auth"] = "10m",
         ["ocsp-stapling"] = "10m",
         ["mcp-session"] = "10m",
+        status_report_standalone = "1m"
```
nic-6443, line 179: keep a consistent style with the code above.

apisix/cli/ngx_tpl.lua
```diff
@@ -550,6 +554,23 @@ http {
...
+        listen {* status_server_addr *} enable_process=privileged_agent;
+        access_log off;
+        location /status {
+            content_by_lua_block {
+                apisix.status_standalone()
+            }
+        }
+        location /status/ready {
+            content_by_lua_block {
+                apisix.status_standalone_ready()
```
nic-6443, line 568: We should provide a unified `/status/ready` endpoint that not only supports standalone mode but also does not support the mode based on etcd as the configuration center.

apisix/init.lua
```diff
@@ -868,6 +869,68 @@ local function healthcheck_passive(api_ctx)
...
 
+function _M.status()
+    core.response.exit(200, "ok")
+end
+
+function _M.status_ready()
+    local_conf = core.config.local_conf()
+    local provider = core.table.try_read_attr(local_conf, "deployment", "role_traditional",
+                                              "config_provider")
+    if provider == "yaml" then
```
membphis, line 880: code style: much clearer

```lua
if provider == "yaml" then
  ....
  core.response.exit(200, "ok")
  return
end

if provider == "etcd" then
  ....
  core.response.exit(200, "ok")
  return
end

core.response.exit(503, "unknown config provider: " .. provider)
return

**Review by membphis — COMMENTED** [review]
the current `title` is not suitable and confuse me

**Comment by Revolyssup** [comment]
> the current `title` is not suitable and confuse me

Modified the title and description

File: apisix/admin/standalone.lua [hunk 4]
```diff
@@ -193,8 +206,28 @@ function _M.run()
...
+        local pid = tostring(ngx.worker.pid())
+        status_shdict:delete(pid)
+        core.log.info("worker ", pid, " removed itself on exit")
+        return
+    end
+
+    -- Reschedule the timer to check again
+    local ok, err = ngx.timer.at(0.1, cleanup_on_exit)
+    if not ok then
+        core.log.error("Failed to reschedule cleanup timer: ", err)
```
nic-6443, line 221:
```suggestion
        core.log.error("failed to reschedule cleanup timer: ", err)
```

apisix/cli/config.lua
```diff
@@ -176,6 +176,8 @@ local _M = {
         ["cas-auth"] = "10m",
         ["ocsp-stapling"] = "10m",
         ["mcp-session"] = "10m",
+        status_report_standalone = "10m",
+        status_report = "10m"
```
nic-6443, line 180: 10m is too large, I only recorded the pid and a bool value, 1m should be enough?

apisix/admin/standalone.lua
Revolyssup: done

apisix/cli/config.lua
Revolyssup: done

apisix/core/config_etcd.lua
```diff
@@ -644,6 +662,7 @@ local function sync_data(self)
 
     if not dir_res then
         if err == "compacted" or err == "restarted" then
+            sync_status_to_shdict(self.need_reload)
```
nic-6443, line 665: In this scenario, the status should be false, right?

apisix/core/config_etcd.lua
```diff
@@ -482,6 +485,20 @@ local function short_key(self, str)
 end
 
 
+local function sync_status_to_shdict(status)
```
nic-6443, line 488:
```suggestion
local function sync_status_to_shdict(need_reload)
```

apisix/core/config_etcd.lua
Revolyssup: yes it will be false. you can look at the logic on status_ready() handler

apisix/core/config_etcd.lua
Revolyssup: done

apisix/init.lua
```diff
@@ -868,6 +869,70 @@ local function healthcheck_passive(api_ctx)
...
+        local res, err
+        for _, host in ipairs(yaml_conf.etcd.host) do
+            res, err = etcd_util.request(host .. "/version", yaml_conf)
+            if res then
+                core.response.exit(200)
+                return
+            end
+        end
+
+        core.response.exit(503, "none of the configured dp_manager available: " .. err)
```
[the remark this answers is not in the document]
Revolyssup: done

apisix/admin/standalone.lua
```diff
@@ -193,8 +206,28 @@ function _M.run()
...
+    if ngx.worker.exiting() then
+        local status_shdict = ngx.shared.status_report_standalone
+        local pid = tostring(ngx.worker.pid())
+        status_shdict:delete(pid)
+        core.log.info("worker ", pid, " removed itself on exit")
+        return
+    end
+
+    -- Reschedule the timer to check again
+    local ok, err = ngx.timer.at(0.1, cleanup_on_exit)
```
nic-6443, line 219: Using a timer to periodically check whether the shared dict needs cleaning feels wasteful. 
Can we change the records in this shared dict to **a list of worker PIDs waiting for synchronization configuration**? When `init_worker` is completed, add the current worker's PID to this list. When `update_config` is completed, directly remove this PID from the list, thus eliminating the need for subsequent checks.

Revolyssup: In the case when worker exits(like reload), it wont be removed from the list causing health check to fail indefinitely.

apisix/core/config_etcd.lua
Revolyssup: done

apisix/admin/standalone.lua
nic-6443: got it.

nic-6443 (approved):

AlinsRan (approved):

apisix/init.lua
```diff
@@ -868,6 +869,70 @@ local function healthcheck_passive(api_ctx)
...
+                break
+            end
+        end
+
+        if resp and resp ~= "" then
+            core.response.exit(503, resp)
+            return
+        end
+
+        core.response.exit(200, "ok")
```
bzp2010, line 900: We're better to return structured data (in JSON), and using text like this can't be interactive with downstream component. HTTP 503 isn't enough to convey enough detail.

apisix/init.lua
```diff
@@ -868,6 +869,70 @@ local function healthcheck_passive(api_ctx)
...
+
+        local yaml_conf = local_conf
+        local res, err
+        for _, host in ipairs(yaml_conf.etcd.host) do
+            res, err = etcd_util.request(host .. "/version", yaml_conf)
+            if res then
+                core.response.exit(200)
+                return
+            end
+        end
```
bzp2010, line 927: What does etcd reachability have to do with whether an instance is READY?

This is a complete departure from the APISIX design intent of using old configurations to provide services when etcd is unavailable to ensure that services are not interrupted. 

If we use it in Kubernetes Probes, as soon as etcd becomes unavailable, that APISIX instance is reported as unhealthy and thus stops receiving incoming traffic. If there are no replaceable instances, the service stops completely.

bzp2010: This status ready API should be limited to whether the instance has loaded any configuration to make it work, either etcd or standalone. etcd availability or whether the configuration is up-to-date is not part of this scope.

apisix/init.lua
```diff
@@ -868,6 +869,70 @@ local function healthcheck_passive(api_ctx)
...
+function _M.status_ready()
+    local_conf = core.config.local_conf()
+    local provider = core.table.try_read_attr(local_conf, "deployment", "role_traditional",
+                                              "config_provider")
+    if provider == "yaml" then
+        local status_shdict = ngx.shared.status_report_standalone
+        local pids = status_shdict:get_keys()
+        local resp = ""
+
+        for _, pid in pairs(pids) do
```
bzp2010, line 885: The `get_keys` function returns an array right? Should we use `ipairs`?

apisix/cli/config.lua
```diff
@@ -176,6 +176,8 @@ local _M = {
         ["cas-auth"] = "10m",
         ["ocsp-stapling"] = "10m",
         ["mcp-session"] = "10m",
+        status_report_standalone = "1m",
+        status_report = "1m"
```
bzp2010, line 180: Is this necessary to use two different shared dicts? Since they won't be enabled at the same time and are (rightfully) functionally identical, this doesn't seem necessary.

bzp2010: And why alone it is using `_` instead of `-`.

conf/config.yaml.example
```diff
@@ -132,7 +132,10 @@ apisix:
   events:                             # Event distribution module configuration
     module: lua-resty-events          # Sets the name of the events module used.
                                       # Supported module: lua-resty-worker-events and lua-resty-events
-
+# status:                       # When enabled, APISIX will provide `/status` and `/status/ready` endpoints
+  #   ip: 127.0.0.1               # /status endpoint will return 200 status code if APISIX has successfully started and running correctly
+  #   port: 7085                  # /status/ready endpoint will return 503 status code if none of the configured etcd (dp_manager) are available
```
bzp2010, line 137: For Apache APISIX, what is dp manager, there is no such thing. Delete it.

bzp2010 (commented):
Again, we must clarify the scope of work of this PR, what is it?

1. is it responsible for reporting that at least one worker has been started? `/status`
2. is it responsible for reporting if all workers have at least one version configuration loaded? `/status/ready`
3. is it responsible for reporting whether all workers have the same version of the configuration loaded?
4. is it responsible for reporting if all workers have loaded the latest configuration version from the configuration provider?
5. is it responsible for reporting whether the configuration provider is available?

I think it should cover at least 1 and 2, with optional support for 3 (but that doesn't seem too difficult). 4 and 5 cannot be implemented by the APISIX data plane alone, or should not be implemented by it.

I have reservations about introducing a new nginx server block to make the status API work on a separate port, we'll need to see how other projects handle this. If this is common practice, then we can do the same.

bzp2010 (changes requested):
I have checked the code part and it seems that there are still some implementation issues and there are some key questions left unanswered that we need to discuss to come to an agreement.

apisix/init.lua
nic-6443: I agree with @bzp2010's opinion that the `/status/ready` endpoint in etcd mode should indeed not be related to the success or failure of connecting to etcd. In that case, it seems no conditions are needed in etcd mode, and a direct return of 200 is sufficient.

apisix/admin/standalone.lua
```diff
@@ -30,12 +30,25 @@ local events       = require("apisix.events")
...
 
 local EVENT_UPDATE = "standalone-api-configuration-update"
 
 local _M = {}
 
 
+local function sync_status_to_shdict(status)
+    if process.type() ~= "worker" then
+        return
+    end
```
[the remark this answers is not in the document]
Revolyssup: init_worker also runs for priviliged agent. This logic skips that.

Revolyssup:
> Again, we must clarify the scope of work of this PR, what is it?
> 
> 1. is it responsible for reporting that at least one worker has been started? `/status`
> 2. is it responsible for reporting if all workers have at least one version configuration loaded? `/status/ready`
> 3. is it responsible for reporting whether all workers have the same version of the configuration loaded?
> 4. is it responsible for reporting if all workers have loaded the latest configuration version from the configuration provider?
> 5. is it responsible for reporting whether the configuration provider is available?
> 
> I think it should cover at least 1 and 2, with optional support for 3 (but that doesn't seem too difficult). 4 and 5 cannot be implemented by the APISIX data plane alone, or should not be implemented by it.
> 
> I have reservations about introducing a new nginx server block to make the status API work on a separate port, we'll need to see how other projects handle this. If this is common practice, then we can do the same.

Keeping this in mind, I have refactored the current PR. Check description.

apisix/init.lua
```diff
@@ -868,6 +869,54 @@ local function healthcheck_passive(api_ctx)
...
+        status = "ok" }),
+        { ["Content-Type"] = "application/json"
+    })
+end
+
+function _M.status_ready()
+    local local_conf = core.config.local_conf()
+    local provider = core.table.try_read_attr(local_conf, "deployment",
+                                              "role_traditional", "config_provider")
+    if provider == "yaml" or provider == "etcd" then
```
bzp2010, line 883: The fact that you only use `role_traditional.config_provider` for assertions makes file-based standalone modes not work properly, so they always fall back to the default value of `etcd` for `role_traditional.config_provider`, which uses the `role_data_plane.config_provider`, which you should adapt.
Otherwise, this one status API can't simply be enabled in environments like Kubernetes probes, and we'd have to make special adaptations in the helm chart, which is neither reasonable nor sustainable.

apisix/core/config_etcd.lua
```diff
@@ -961,7 +979,9 @@ function _M.new(key, opts)
     if not health_check_timeout or health_check_timeout < 0 then
         health_check_timeout = 10
     end
-
+    if key then
+        sync_status_to_shdict(true)
+    end
```
bzp2010, line 984: What does the fact that the key is not null mean? You may need to explain this by adding some additional comments, I don't understand what this is.

apisix/admin/standalone.lua
```diff
@@ -193,8 +206,28 @@ function _M.run()
...
+        core.log.info("worker ", pid, " removed itself on exit")
+        return
+    end
+
+    -- Reschedule the timer to check again
+    local ok, err = ngx.timer.at(0.1, cleanup_on_exit)
+    if not ok then
+        core.log.error("failed to reschedule cleanup timer: ", err)
+    end
+end
```
bzp2010, line 223: What scenario is this function used to handle?

I.e. why the worker is ready to exit, and we should also remove its shdict entry properly.
If this is due to a reload, the old worker is no longer listening and therefore no longer receiving incoming traffic, which does not affect overall availability. After the new worker comes online, it uses the new pid anyway and sets its initial state to false.

So I have some questions:

1. is the use of pid as expected? pid is necessarily changed when the worker restarts, while the worker id does not change.
2. since the initial state of the worker is always false, is it necessary to actively perform cleanup?
3. can you write this logic in standalone and handle the behavior of automatically restoring the latest configuration from shdict on restart in standalone + API mode well? (This does not use broadcasts and therefore does not execute any logic in standalone.lua)

bzp2010 (changes requested):

apisix/admin/standalone.lua
```diff
@@ -30,12 +30,25 @@ local events       = require("apisix.events")
...
 
 
+local function sync_status_to_shdict(status)
+    if process.type() ~= "worker" then
+        return
+    end
+    local status_shdict = ngx.shared.status_report
+    local pid = worker_pid()
+    core.log.info("sync status to shared dict, pid: ", pid, " status: ", status)
+    status_shdict:set(pid, status, 5*60)
```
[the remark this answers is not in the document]
Revolyssup: you're right it should not have a ttl. I will change

apisix/cli/ngx_tpl.lua
```diff
@@ -550,6 +558,23 @@ http {
...
+            content_by_lua_block {
+                apisix.status()
+            }
+        }
+        location /status/ready {
+            content_by_lua_block {
+                apisix.status_ready()
+            }
+        }
+    }
```
bzp2010, line 575: Depending on your implementation (and the as is situation with NGINX), this server block will always be defined under the http subsystem, so when the user tries to access `/status/ready`, the request is handled by `apisix.status_ready()`.

You use https://github.com/apache/apisix/pull/12200/files#diff-7439ffe94cc4bcad12066d5d83f6c1004b8e9b3ff5097b645fa7c700eb3b3743R886 with 

```
local status_shdict = ngx.shared["status-report"] or ngx.shared["status-report-stream"]
```

to get the desired shdict, and since the request is handled in the http subsystem, it will always get the `status-report` instead of the `status-report-stream`, so the stream health check you implemented is meaningless.

Now, adding http lua_shared_dict to nginx.conf is conditional on

```
http.lua_shared_dict["status-report"]
```

being present, which means that, by default, it he always present, so there's no chance of you getting `status-report-stream`.

---

Considering it from another side, both config_etcd and config_yaml will register some timers in `init_worker` to load the full configuration from the configuration provider (both route and stream route included).

This always happens in http and stream (i.e. the loaded configuration for http also contains the stream route, and vice versa), and since the user always has at least one subsystem enabled (either http or stream), the status of the worker's health can be reported by either subsystem, and reporting at the same time is probably fine.

Also, although http and stream are different subsystems, they both run in the same worker process, so when the http route configuration is loaded successfully, it also means that the stream route is loaded successfully, and a successful configuration load means that the worker is healthy.

Therefore, using a shared shdict would be feasible and necessary.

apisix/core/config_yaml.lua
```diff
@@ -45,11 +45,16 @@ local pcall        = pcall
...
 local apisix_yaml_path = profile:yaml_path("apisix")
 local created_obj  = {}
 local shared_dict
-
-
+local status_report_shared_dict_name = "status-report"
+local is_http = ngx.config.subsystem == "http"
+if not is_http then
+    status_report_shared_dict_name = status_report_shared_dict_name .. "-stream"
+end
```
bzp2010, line 57: We don't need this separate branch anymore do we?

Revolyssup: we need separate shared dict for stream

apisix/cli/ngx_tpl.lua
```diff
@@ -70,6 +70,7 @@ lua {
     {% if standalone_with_admin_api then %}
     lua_shared_dict standalone-config {* meta.lua_shared_dict["standalone-config"] *};
     {% end %}
+    lua_shared_dict status-report {* meta.lua_shared_dict["status-report"] *};
```
bzp2010, line 73:
```suggestion
    {% if status then %}
    lua_shared_dict status-report {* meta.lua_shared_dict["status-report"] *};
    {% end %}
```

apisix/core/config_yaml.lua
```diff
@@ -73,6 +75,16 @@ local apisix_yaml
 local apisix_yaml_raw -- save a deepcopy of the latest configuration for API
 local apisix_yaml_mtime
 
+local function sync_status_to_shdict(status)
+    if process.type() ~= "worker" then
+        return
+    end
+    local status_shdict = ngx.shared[status_report_shared_dict_name]
+    local pid = worker_id()
```
bzp2010, line 83: rename the `pid`

Revolyssup:
@bzp2010 I applied all suggestions

apisix/core/config_etcd.lua
```diff
@@ -65,9 +68,7 @@ local err_etcd_grpc_engine_timeout = "context deadline exceeded"
 local err_etcd_grpc_ngx_timeout = "timeout"
 local err_etcd_unhealthy_all = "has no healthy etcd endpoint available"
 local health_check_shm_name = "etcd-cluster-health-check"
-if not is_http then
-    health_check_shm_name = health_check_shm_name .. "-stream"
-end
```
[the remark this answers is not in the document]
Revolyssup: mistake

apisix/init.lua
```diff
@@ -868,6 +869,56 @@ local function healthcheck_passive(api_ctx)
...
+
+        for _, pid in ipairs(pids) do
+            local ready = status_shdict:get(pid)
+            if not ready then
+                core.log.warn("worker pid: ", pid, " has not received configuration")
+                core.table.insert(errors, "worker pid: " .. pid ..
+                                  " has not received configuration")
+                break
+            end
+        end
```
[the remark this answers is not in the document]
Revolyssup: I think ngx.worker.count is not needed. All workers initialise and set the id-> to false in init_worker() so we will iterate over all started workers.

apisix/core/config_yaml.lua
```diff
@@ -73,6 +75,19 @@ local apisix_yaml
...
+local function sync_status_to_shdict(status)
+    if process.type() ~= "worker" then
+        return
+    end
+    local status_shdict = ngx.shared[status_report_shared_dict_name]
+    if not status_shdict then
+        return
+    end
+    local id = worker_id()
+    log.info("sync status to shared dict, pid: ", id, " status: ", status)
```
bzp2010, line 87: pid => worker id

apisix/core/config_etcd.lua
```diff
@@ -29,17 +29,20 @@ local core_str     = require("apisix.core.string")
 local new_tab      = require("table.new")
 local inspect      = require("inspect")
 local errlog       = require("ngx.errlog")
+local process      = require("ngx.process")
 local log_level    = errlog.get_sys_filter_level()
 local NGX_INFO     = ngx.INFO
 local check_schema = require("apisix.core.schema").check
 local exiting      = ngx.worker.exiting
+local worker_id   = ngx.worker.id
```
bzp2010, line 37:
```suggestion
local worker_id    = ngx.worker.id
```

apisix/init.lua
```diff
@@ -868,6 +869,54 @@ local function healthcheck_passive(api_ctx)
...
+
+function _M.status_ready()
+    local local_conf = core.config.local_conf()
+    local role = core.table.try_read_attr(local_conf, "deployment", "role")
+    local provider = core.table.try_read_attr(local_conf, "deployment", "role_" .. role, "config_provider")
+    if provider == "yaml" or provider == "etcd" then
+        local status_shdict = ngx.shared["status-report"]
+        local ids = status_shdict:get_keys()
+        local errors = {}
+
```
bzp2010, line 887: If the `ids` length is not equal to `ngx.worker.count`, it should also be reported as unhealthy.
Note that worker ID start at 0.

apisix/init.lua
```diff
@@ -868,6 +869,67 @@ local function healthcheck_passive(api_ctx)
...
+                                              role, "config_provider")
+    if provider == "yaml" or provider == "etcd" then
+        local status_shdict = ngx.shared["status-report"]
+        local ids = status_shdict:get_keys()
+        local error
+        local worker_count = ngx.worker.count()
+       if #ids ~= worker_count then
+            core.log.warn("worker count: ", worker_count, " but status report count: ", #ids)
+            error = "worker count: " .. ngx.worker.count() ..
+            " but status report count: " .. #ids
```
bzp2010, line 892: It's a good idea to also print a list of ids to show which worker is not working properly. Just `core.json.encode()` it.

bzp2010 (approved):
There's a small problem left.

The rest of it LGTM, you'll be adding documentation, right?

apisix/init.lua
Revolyssup: Okay

Revolyssup:
> There's a small problem left.
> 
> The rest of it LGTM, you'll be adding documentation, right?

Yes

bzp2010 (commented):
LGTM

nic-6443 (dismissed):

docs/en/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+This API can be used to perform readiness probes on APISIX before APISIX starts receiving user requests.
+
+### GET /status
+
+Returns a JSON reporting the status of APISIX workers
+
+```json
+{
+  "status": "ok"
```
kayx23, line 46: What if the status is not ok? The section below shows output when its "not ok" - what about this section?

docs/en/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+
+### GET /status/ready
+
+Returns `ok` when all workers have loaded the configuration, otherwise returns the specific error with `503` error code. Below are specific examples.
+
+When all workers have loaded the configuration
```
kayx23, line 54:
```suggestion
When all workers have loaded the configuration:
```

Please do the same for other sentences missing punctuation in this doc.

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+在 Apache APISIX 中，状态 API 用于：
```
kayx23, line 24:
```suggestion
在 Apache APISIX 中，Status API 用于：
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+* 检查 APISIX 是否已成功启动并正确运行
+* 检查所有工作人员是否已收到并加载配置。
```
kayx23, line 27:
```suggestion
* 检查所有 workers 是否已收到配置并加载。
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+  status:
+    ip: "127.0.0.1"
+    port: 7085
+```
+
+此 API 可用于在 APISIX 开始接收用户请求之前对 APISIX 执行就绪探测。
+
+### GET /status
+
+返回报告 APISIX 工作者状态的 JSON
```
kayx23, line 42:
```suggestion
返回报告 APISIX worker 状态的 JSON：
```

similarly add missing punctuation in this doc

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+当 1 个 workers 尚未初始化时
+
+```json
+{
+  "status": "error",
+  "error": "worker count: 16 but status report count: 15"
+}
+```
+
+当特定工作器尚未加载配置时
```
kayx23, line 71:
```suggestion
当特定 worker 尚未加载配置时：
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+{
+  "status": "ok"
+}
+```
+
+### GET /status/ready
+
+当所有 Worker 都已加载配置时，返回 `ok`；否则，返回特定错误，错误代码为 `503`。以下是具体示例。
+
+当所有 Worker 都已加载配置时
```
kayx23, line 54: image

lower case the `Worker`, thanks

docs/en/latest/status-api.md
Revolyssup: If APISIX is running, this endpoint will always return 200 and no other response. If APISIX is not running, then it wouldn't return any HTTP response, it will error out while establishing tcp connection.

kayx23: Make sense. Maybe add a sentence at the end of this section?

> If APISIX is not running, the request will error out while establishing TCP connection.

wdyt

docs/en/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+When all workers have loaded the configuration:
+
+```json
+{
+  "status": "ok"
+}
+```
+
+When 1 workers has't been initialised
```
kayx23, line 62:
```suggestion
When 1 workers has't been initialised:
```

docs/en/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+```json
+{
+  "status": "error",
+  "error": "worker count: 16 but status report count: 15"
+}
+```
+
+When a particular worker hasn't loaded the configuration
```
kayx23, line 71:
```suggestion
When a particular worker hasn't loaded the configuration:
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+在 Apache APISIX 中，Status API 用于：
+
+* 检查 APISIX 是否已成功启动并正确运行
+* 检查所有 workers 是否已收到配置并加载。
```
kayx23, line 27:
```suggestion
* 检查所有 workers 是否已收到配置并加载
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+
+当所有 worker 都已加载配置时
+
+```json
+{
+  "status": "ok"
+}
+```
+
+当 1 个 workers 尚未初始化时
```
kayx23, line 62:
```suggestion
当 1 个 workers 尚未初始化时：
```

docs/zh/latest/status-api.md
```diff
@@ -0,0 +1,78 @@
...
+{
+  "status": "ok"
+}
+```
+
+### GET /status/ready
+
+当所有 worker 都已加载配置时，返回 `ok`；否则，返回特定错误，错误代码为 `503`。以下是具体示例。
+
+当所有 worker 都已加载配置时
```
kayx23, line 54:
```suggestion
当所有 worker 都已加载配置时：
```

kayx23 (commented):
Others LGTM. Please fix for doc lint errors as well, likely due to punctuations.

kayx23 (approved):

nic-6443 (approved):

bzp2010 (approved):

PR #12207, opened 2025-05-13
feat: display a warning message when the data plane has write access to etcd

LiteSun:
### Description

In short, as long as APISIX is deployed in decoupled mode with the role set to data_plane, and it can successfully write to etcd, a warning message will be displayed.

#### A warning message will be displayed under the following conditions:

1. Etcd allows anonymous access, enabling the data plane to write to etcd.
2. The configured etcd user has write permissions.

#### No warning will be shown in the following case:

1. The configured etcd user has read-only permissions.

### Checklist


LiteSun:
After discussions with @moonming  and @nic-6443, we found that checking etcd write permissions during the APISIX startup phase is not very accurate, especially under different etcd cluster permission configurations. A new PR will be submitted to display a warning message when a write operation to etcd is attempted.

PR #12214, opened 2025-05-14
feat(standalone): support revision in API-driven standalone mode like etcd

AlinsRan:
### Description

In standalone mode, APISIX’s Admin API currently requires clients to retrieve the entire configuration on every sync. As configurations grow in size or change more often, this full-sync approach can lead to unnecessary network traffic and increased latency in applying updates. Moreover, frequent resource changes force a full rebuild of the internal radixtree, which significantly degrades lookup performance under heavy churn.

In the scenario of service discovery, the upstream will always update frequently, and we hope to only update the upstream without affecting the usage of other resources.

**Control the changes of each resource by adding `<resource-type>_conf_version`**:

```json
{
    "consumer_groups_conf_version": 1000,
    "consumers_conf_version": 1000,
    "global_rules_conf_version": 1000,
    "plugin_configs_conf_version": 1000,
    "plugin_metadata_conf_version": 1000,
    "protos_conf_version": 1000,
    "routes_conf_version": 1000,
    "secrets_conf_version": 1000,
    "services_conf_version": 1000,
    "ssls_conf_version": 1000,
    "upstreams_conf_version": 1000
    "routes": [
        {
            "id": "r1",
            "uri": "/hello",
            "upstream_id": "u1"
        }
    ],
    "upstreams": [
        {
            "id": "u1",
            "nodes": {
                "127.0.0.1:1980": 1,
                "127.0.0.1:1980": 1
            },
            "type": "roundrobin"
        }
    ]
    "services":[]
}
```

### Checklist

### DIFF
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -30,25 +30,61 @@ local events       = require("apisix.events")
 local core         = require("apisix.core")
 local config_yaml  = require("apisix.core.config_yaml")
 local check_schema = require("apisix.core.schema").check
+local tbl_deepcopy = require("apisix.core.table").deepcopy
+
 
 local EVENT_UPDATE = "standalone-api-configuration-update"
 
 local _M = {}
 
 
-local function update_and_broadcast_config(apisix_yaml, conf_version)
-    local config = core.json.encode({
+
+local function get_config()
+    local config = shared_dict:get("config")
+    if not config then
+        return nil, "not found"
+    end
+
+    local err
+    config, err = core.json.decode(config)
+    if not config then
+        return nil, "failed to decode json: " .. err
+    end
+    return config
+end
+
+
+local function update_and_broadcast_config(old_config, apisix_yaml, conf_version_kv)
+    local config = {
         conf = apisix_yaml,
-        conf_version = conf_version,
-    })
+        conf_version = conf_version_kv
+    }
+
+    if old_config then
+        local ori_config = tbl_deepcopy(old_config)
+        ori_config.conf_version = ori_config.conf_version or {}
+        ori_config.conf = ori_config.conf or {}
+        for key, v in pairs(conf_version_kv) do
+            if v ~= ori_config.conf_version[key] then
+                ori_config.conf[key] = apisix_yaml[key]
+                ori_config.conf_version[key] = v
+            end
+        end
+        config = ori_config
+    end
+
+    local encoded_config, encode_err = core.json.encode(config)
+    if not encoded_config then
+        core.log.error("failed to encode json: ", encode_err)
+        return nil, "failed to encode json: " .. encode_err
+    end
 
     if shared_dict then
-        -- the worker that handles Admin API calls is responsible for writing the shared dict
-        local ok, err = shared_dict:set("config", config)
+        local ok, save_err = shared_dict:set("config", encoded_config)
         if not ok then
-            return nil, "failed to save config to shared dict: " .. err
+            return nil, "failed to save config to shared dict: " .. save_err
         end
-        core.log.info("standalone config updated: ", config)
+        core.log.info("standalone config updated: ", encoded_config)
     else
         core.log.crit(config_yaml.ERR_NO_SHARED_DICT)
     end
@@ -59,24 +95,38 @@ end
 local function update(ctx)
     local content_type = core.request.header(nil, "content-type") or "application/json"
 
-    local conf_version
-    if ctx.var.arg_conf_version then
-        conf_version = tonumber(ctx.var.arg_conf_version)
-        if not conf_version then
-            return core.response.exit(400, {error_msg = "invalid conf_version: "
-                                            .. ctx.var.arg_conf_version
-                                            .. ", should be a integer" })
+
+    local update_keys
+
+    local config, err = get_config()
+    if not config then
+        if err ~= "not found" then
+            core.log.error("failed to get config from shared dict: ", err)
+            return core.response.exit(500, {
+                error_msg = "failed to get config from shared dict: " .. err
+            })
         end
     else
-        conf_version = ngx.time()
-    end
-    -- check if conf_version greater than the current version
-    local _, ver = config_yaml._get_config()
-    if conf_version <= ver then
-        return core.response.exit(400, {error_msg = "invalid conf_version: conf_version ("
-                                        .. conf_version
-                                        .. ") should be greater than the current version ("
-                                        .. ver .. ")"})
+        for key, version in pairs(config.conf_version) do
+            local header = "x-apisix-conf-version" .. "-" .. key
+            local req_conf_version = core.request.header(ctx, header)
+            if req_conf_version then
+                if not tonumber(req_conf_version) then
+                    return core.response.exit(400, {error_msg = "invalid header: [" .. header ..
+                        ": " .. req_conf_version .. "]" ..
+                        " should be a integer"})
+                end
+                req_conf_version = tonumber(req_conf_version)
+                if req_conf_version <= version then
+                    return core.response.exit(400, {error_msg = "invalid header: [" .. header ..
+                        ": " .. req_conf_version .. "]" ..
+                        " should be greater than the current version (" .. version .. ")"})
+                else
+                    update_keys = update_keys or {}
+                    update_keys[key] = req_conf_version
+                end
+            end
+        end
     end
 
     -- read the request body
@@ -107,9 +157,19 @@ local function update(ctx)
 
     -- check input by jsonschema
     local apisix_yaml = {}
+    local confi_version_kv = tbl_deepcopy(config and config.conf_version or {})
     local created_objs = config_yaml.fetch_all_created_obj()
+
     for key, obj in pairs(created_objs) do
-        if req_body[key] and #req_body[key] > 0 then
+        local updated = false
+        if not update_keys then
+            confi_version_kv[key] = confi_version_kv[key] and confi_version_kv[key] + 1 or 1
+            updated = true
+        elseif update_keys[key] then
+            confi_version_kv[key] = update_keys[key]
+            updated = true
+        end
+        if updated and req_body[key] and #req_body[key] > 0 then
             apisix_yaml[key] = table_new(1, 0)
             local item_schema = obj.item_schema
             local item_checker = obj.checker
@@ -137,12 +197,15 @@ local function update(ctx)
         end
     end
 
-    local ok, err = update_and_broadcast_config(apisix_yaml, conf_version)
+    local ok, err = update_and_broadcast_config(config, apisix_yaml, confi_version_kv)
     if not ok then
         core.response.exit(500, err)
     end
 
-    core.response.set_header("X-APISIX-Conf-Version", tostring(conf_version))
+    for key, version in pairs(confi_version_kv) do
+        core.response.set_header("X-APISIX-Conf-Version-" .. key, tostring(version))
+    end
+
     return core.response.exit(202)
 end
 
@@ -151,9 +214,26 @@ local function get(ctx)
     local accept = core.request.header(nil, "accept") or "application/json"
     local want_yaml_resp = core.string.has_prefix(accept, "application/yaml")
 
-    local _, ver, config = config_yaml._get_config()
+    local config, err = get_config()
+    if not config then
+        if err ~= "not found" then
+            core.log.error("failed to get config from shared dict: ", err)
+            return core.response.exit(500, {
+                error_msg = "failed to get config from shared dict: " .. err
+            })
+        end
+        config = {}
+        local created_objs = config_yaml.fetch_all_created_obj()
+        for _, obj in pairs(created_objs) do
+            core.response.set_header("X-APISIX-Conf-Version-" .. obj.key,
+                                        tostring(obj.conf_version))
+        end
+    else
+        for key, version in pairs(config.conf_version) do
+            core.response.set_header("X-APISIX-Conf-Version-" .. key, tostring(version))
+        end
+    end
 
-    core.response.set_header("X-APISIX-Conf-Version", tostring(ver))
     local resp, err
     if want_yaml_resp then
         core.response.set_header("Content-Type", "application/yaml")
@@ -195,7 +275,7 @@ end
 
 
 function _M.init_worker()
-    local function update_config()
+    local function update_config(data, event, resource, pid)
         local config, err = shared_dict:get("config")
         if not config then
             core.log.error("failed to get config from shared dict: ", err)
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -17,8 +17,6 @@
 local type         = type
 local pairs        = pairs
 local ipairs       = ipairs
-local tonumber     = tonumber
-local tostring     = tostring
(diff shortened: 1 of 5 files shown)


apisix/admin/standalone.lua
```diff
@@ -151,9 +179,21 @@ local function get(ctx)
...
+            core.log.error("failed to get config from shared dict: ", err)
+            return core.response.exit(500, {
+                error_msg = "failed to get config from shared dict: " .. err
+            })
+        end
+        config = {}
+        local created_objs = config_yaml.fetch_all_created_obj()
+        for _, obj in pairs(created_objs) do
+            config[obj.conf_version_key] = obj.conf_version
+        end
```
nic-6443, line 194: This branch indicates that the shared dict is not configured, so just return an empty object? There is no need to populate these version fields.

apisix/core/config_yaml.lua
```diff
@@ -158,29 +148,55 @@ local function sync_data(self)
...
+                end
+            end
+            local new_values = new_tab(8, 0)
+            self.values_hash = new_tab(0, 8)
+            for _, item in ipairs(self.values) do
+                local id = item.value.id
+                if not exist_items[id]  then
+                    config_util.fire_all_clean_handlers(item)
+                else
+                    insert_tab(new_values, item)
```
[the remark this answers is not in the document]
nic-6443: Another issue is that I think we need to determine whether the `modifiedIndex` has changed, and if it has, we need to trigger the old item's `fire_all_clean_handlers` function. This needs to be triggered not only when the item is deleted but also when it is updated.

docs/en/latest/deployment-modes.md
```diff
@@ -115,73 +115,160 @@ This method is more suitable for two types of users:
...
+        }
+    ],
+    "upstreams": [
+        {
+            "id": "u1",
+            "nodes": {
+                "127.0.0.1:1980": 1,
+                "127.0.0.1:1980": 1
+            },
+            "type": "roundrobin"
```
nic-6443, line 259: add a `modifiedIndex` field to those resources.

apisix/admin/standalone.lua
AlinsRan: It is used to view the currently available resources, as well as their corresponding confw_version keys and initial confw_version.
The purpose is for ease of use, and if you find it meaningless, I can delete it.

nic-6443:
The title of this PR is not quite appropriate, it can be changed to:
support revision in API-driven standalone mode like etcd

apisix/admin/standalone.lua
nic-6443: I thought about it, and I think that when the shared dict is not found, it should return 500 with an error message indicating that this shared dict needs to be configured.

apisix/core/config_yaml.lua
```diff
@@ -395,6 +430,14 @@ function _M.new(key, opts)
...
     end
 
+    if is_use_admin_api() then
+        if item_schema and item_schema.properties then
+            -- allow clients to specify modifiedIndex to control resource changes.
+            item_schema.properties.modifiedIndex = {
+                type = "integer",
+            }
+        end
+    end
```
bzp2010, line 440: It seems that this should also be implemented in standalone. Should we be concerned that the schema is permanently modified, i.e. should we modify a deepcopy?

apisix/admin/standalone.lua
AlinsRan: Should the initial state be 500?
 I think the behavior of this interface itself should have no internal errors.
 Its purpose is to return the current configuration information.
 If not, it is empty.
@bzp2010  What's your opinion?

apisix/admin/standalone.lua
```diff
@@ -137,12 +166,11 @@ local function update(ctx)
         end
     end
 
-    local ok, err = update_and_broadcast_config(apisix_yaml, conf_version)
+    local ok, err = update_and_broadcast_config(apisix_yaml)
```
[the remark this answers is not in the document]
AlinsRan: You're right.

apisix/admin/standalone.lua
bzp2010: Do we need to indicate any special meaning through the 500 response code of this interface? If it's just an empty object, I guess the client can understand that the server doesn't have any configurations to do the diff correctly, which doesn't seem to be a sufficient basis to support the default return of 500 and an error message?

@nic-6443 @AlinsRan 🤔

bzp2010: So I think HTTP 200 + `{}` response would be sufficient.

docs/en/latest/deployment-modes.md
AlinsRan: fixed.

apisix/admin/standalone.lua
```diff
@@ -151,9 +179,21 @@ local function get(ctx)
     local accept = core.request.header(nil, "accept") or "application/json"
     local want_yaml_resp = core.string.has_prefix(accept, "application/yaml")
 
-    local _, ver, config = config_yaml._get_config()
+    local config, err = get_config()
```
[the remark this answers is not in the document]
AlinsRan: fixed.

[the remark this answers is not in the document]
AlinsRan: https://github.com/apache/apisix/pull/12214/files#diff-0b1c3ea45d7bed65c686b7fc5179dee2a89c2e4c7462bc3f1f59abaf087cb9e9R127-R130

apisix/admin/standalone.lua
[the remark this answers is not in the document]
AlinsRan: fixed.
https://github.com/apache/apisix/pull/12214/files#diff-0b1c3ea45d7bed65c686b7fc5179dee2a89c2e4c7462bc3f1f59abaf087cb9e9R127-R130

apisix/core/config_yaml.lua
AlinsRan: fixed.

apisix/admin/standalone.lua
```diff
@@ -195,7 +235,7 @@ end
 
 
 function _M.init_worker()
-    local function update_config()
+    local function update_config(data, event, resource, pid)
```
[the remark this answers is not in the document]
AlinsRan: fixed.

apisix/admin/standalone.lua
nic-6443: Sorry, I misunderstood. I thought the returned `not found` error was due to missing shared dict, but it is actually because no configuration has been submitted yet. It should not return 500; I agree with using 200.

nic-6443 (dismissed):

membphis (dismissed):

apisix/core/config_yaml.lua
```diff
@@ -158,29 +149,59 @@ local function sync_data(self)
...
+                end
+            end
+
+            local new_values = new_tab(8, 0)
+            self.values_hash = new_tab(0, 8)
+            for _, item in ipairs(self.values) do
+                local id = item.value.id
+                if not exist_modifiedIndex_items[id]  then
+                    config_util.fire_all_clean_handlers(item)
+                else
```
bzp2010, line 193: Why does resource cleanup use `modifiedIndex` to determine if a resource needs to be deleted?
They don't seem to be related, it should be based on the resource id, such as the route id or consumer username.

bzp2010: A correct process is:
1. Check which resources (of ids) existed in the old configuration set but not in the new configuration set, which means that these resources need to be removed and cleaned up.
2. check which resources exist in both the old and new configuration sets, but have inconsistent modifiedIndex, meaning they need to be updated.
3. check for resources that exist in the new configuration set but not in the old, meaning they need to be created.

Where 1, should be realized by the resource id, not the modifiedIndex, they have nothing to do with each other. The actual situation is that if for a given resource id, if the id is changed to some other value, it means deleting a resource represented by the old id and creating a resource represented by the new id, which would be the simplest and most reliable pattern.

ADC's "differ" is the exact implementation of this idea. Except that it allows a comparison range beyond modifiedIndex, and you can check its code and implement a simple equivalent.
https://github.com/api7/adc/blob/main/apps/cli/src/differ/differv3.ts#L308-L639

apisix/core/config_yaml.lua
```diff
@@ -158,29 +149,59 @@ local function sync_data(self)
...
+                    insert_tab(new_values, item)
+                    self.values_hash[id] = #new_values
+                end
+            end
+            self.values = new_values
```
bzp2010, line 198: Why is the new value set here? Shouldn't it be set after the schema check below?

apisix/core/config_yaml.lua
```diff
@@ -256,12 +281,26 @@ local function sync_data(self)
...
+                    local pre_val = self.values[pre_index]
+                    if pre_val and
+                        (not pre_val.modifiedIndex or pre_val.modifiedIndex ~= modifiedIndex) then
+                        log.error("fire all clean handlers for ", self.key, " id: ", id,
+                                    " modifiedIndex: ", modifiedIndex)
+                        config_util.fire_all_clean_handlers(pre_val)
+                        self.values[pre_index] = conf_item
+                        conf_item.value.id = item_id
+                        conf_item.clean_handlers = {}
+                    end
```
bzp2010, line 297: Since self.values has been overwritten with a new table above (different pointer address), what is this dealing with?

apisix/core/config_yaml.lua
AlinsRan: The logic here is to remove old values that do not exist in the new list.

apisix/core/config_yaml.lua
AlinsRan: Isn't that how we handle it now?
The prerequisite for exist_rodifiedIndex_items is to use id as the index, followed by modifiedIndex.

apisix/core/config_yaml.lua
AlinsRan: In the above processing, if the ID is the same and there is a modifiedIndex, it will not be overwritten. Here, the old table that satisfies the change will be overwritten and updated.

apisix/core/config_yaml.lua
AlinsRan: In order to make it easier to understand, I have optimized the logic here: only ID is detected, and the detection of modifiedIndex will be handled in the subsequent logic changes

t/admin/standalone.spec.ts
```diff
@@ -228,5 +295,48 @@ describe("Admin - Standalone", () => {
...
+      expect(resp3.status).toEqual(202);
+
+      // Check route /r1 not exists
+      const resp3_1 = await client.get("/r1").catch((err) => err.response);
+      expect(resp3_1.status).toEqual(404);
+
+      // Check route /r2 exists
+      const resp3_2 = await client.get("/r2");
+      expect(resp3_2.status).toEqual(200);
+    });
```
bzp2010, line 340: ditto

AlinsRan: done.

t/admin/standalone.spec.ts
```diff
@@ -210,6 +245,38 @@ describe("Admin - Standalone", () => {
...
+      expect(resp2.status).toEqual(202);
+
+      const resp2_1 = await client.get(ENDPOINT);
+      expect(resp2_1.status).toEqual(200);
+      expect(resp2_1.data.routes_conf_version).toEqual(17);
+      expect(resp2_1.data.ssls_conf_version).toEqual(5);
+      expect(resp2_1.data.services_conf_version).toEqual(5);
+      expect(resp2_1.data.upstreams_conf_version).toEqual(5);
+      expect(resp2_1.data.consumers_conf_version).toEqual(5);
+    });
```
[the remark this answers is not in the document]
AlinsRan: Done.

docs/en/latest/deployment-modes.md
```diff
@@ -115,73 +115,162 @@ This method is more suitable for two types of users:
...
+  role: data_plane
+  role_data_plane:
+    config_provider: yaml
+```
 
-2. The API-driven is an emerging paradigm for standalone.
+This makes it possible to disable the Admin API and discover configuration changes and reloads based on the local file system.
 
-    The routing rules will be entirely in memory and not in a file, and it will need to be updated using the dedicated Standalone Admin API.
+#### API-driven (Experimental)
```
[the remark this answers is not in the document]
AlinsRan: done.

juzhiyuan:
Hi, please also check the checklist items to make the pr clear

t/admin/standalone.spec.ts
```diff
@@ -159,7 +192,82 @@ describe("Admin - Standalone", () => {
...
+
+      // Check route /r1 exists
+      const resp_1 = await client.get("/r1");
+      expect(resp_1.status).toEqual(200);
+
+      // Update with c2
+      const resp2 = await client.put(ENDPOINT, c2);
+      expect(resp2.status).toEqual(202);
+
+      // Check route /r1 exists
```
bzp2010, line 251:
```suggestion
      // Check route /r1 exists
      // But it is not applied because the modifiedIndex is the same as the old value
```

bzp2010 (dismissed):
Agree in principle with the PR and we can merge when you resolve all the issues on the documentation.

docs/en/latest/deployment-modes.md
```diff
@@ -115,73 +115,162 @@ This method is more suitable for two types of users:
...
+3. update based on resource type
+
+In APISIX memory, the current configuration is:
+
+```json
+{
+    "routes_conf_version": 1000,
+    "upstreams_conf_version": 1000,
+}
+
```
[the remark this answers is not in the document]
AlinsRan: fixed.

docs/en/latest/deployment-modes.md
```diff
@@ -115,73 +115,162 @@ This method is more suitable for two types of users:
...
-      role: traditional
-      role_traditional:
-        config_provider: yaml
-    #END
-    ```
+```yaml
+deployment:
+  role: traditional
+  role_traditional:
+    config_provider: yamls
```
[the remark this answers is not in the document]
AlinsRan: fixed.

docs/en/latest/deployment-modes.md
```diff
@@ -115,73 +115,162 @@ This method is more suitable for two types of users:
...
+      "routes": [],
+      "upstreams": []
+    }
     ```
 
-    The update API validates the input and returns an error if it is invalid. If the configuration is accepted, it responds with a `202 Accepted` status and includes the latest configuration version in the `X-APISIX-Conf-Version` header.
+    APISIX compares each provided `<resource>_conf_version` against its in-memory `<resource>_conf_version` for that resource type:
+
+  - **Greater than** the current `conf_version`
+    If your `<resource>_conf_version` is **higher**, APISIX will **rebuild/reset** that resource type’s data to match your payload.
```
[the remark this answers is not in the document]
AlinsRan: done.

docs/en/latest/deployment-modes.md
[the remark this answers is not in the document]
AlinsRan: fixed.

kayx23 (approved):
Doc LGTM

nic-6443 (approved):

bzp2010 (approved):

PR #12216, opened 2025-05-15
fix: only parse and validate apisix.yaml in cli when startup

nic-6443:
### Description

The purpose of the `read_yaml_conf` function is to read `config.yaml`, but it also parses and validates `apisix.yaml`. This validation only needs to be performed when the CLI module starts, and there is no need to repeatedly parse and validate apisix.yaml during runtime when try to get `config.yaml`:
https://github.com/apache/apisix/blob/958312fb016f5a63eac22f0955087e1fdc390c54/apisix/core/config_local.lua#L61

Fixes #

### Checklist

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -15,6 +15,7 @@
 -- limitations under the License.
 --
 
+local ngx = ngx
 local yaml = require("lyaml")
 local profile = require("apisix.core.profile")
 local util = require("apisix.cli.util")
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -293,7 +293,9 @@ function _M.read_yaml_conf(apisix_home)
         end
     end
 
-    if default_conf.deployment.config_provider == "yaml" then
+    --- using `not ngx` to check whether the current execution environment is apisix cli module,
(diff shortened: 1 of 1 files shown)


membphis (approved):

moonming (approved):

apisix/cli/file.lua
```diff
@@ -293,7 +294,9 @@ function _M.read_yaml_conf(apisix_home)
         end
     end
 
-    if default_conf.deployment.config_provider == "yaml" then
+    --- using `not ngx` to check whether the current execution environment is apisix cli module,
+    --- because it is only necessary to parse and validate `apisix.yaml` in apisix cli.
+    if default_conf.deployment.config_provider == "yaml" and not ngx then
         local apisix_conf_path = profile:yaml_path("apisix")
```
bzp2010, line 300: Actually I don't see what the following code does, it just reads the yaml and parses the variables whose values it contains. Other than that, there is no utilization of the values it reads.
So whether we add this extra judgment or not doesn't affect the functionality of using it. I read your PR description and this is just an efficiency difference? I didn't get the need for it.

nic-chen (approved):

bzp2010 (approved):

PR #12218, opened 2025-05-15
change: delete the `server-info` plugin 

LiteSun:
### Description

The server-info plugin in Apache APISIX periodically reports basic server information to etcd. When a large number of APISIX instances start simultaneously, these periodic writes can lead to increased load on etcd. If etcd is not adequately provisioned or configured to handle this load, it may become overwhelmed, potentially causing APISIX startup failures. ￼

### Checklist


nic-6443 (approved):

AlinsRan (approved):

bzp2010:
> Apologies for using so many header lines to increase the font size, but it's important and necessary to do so.

# The author and all who approved this PR, I have questions:


## Where is the discussion on this **BREAK CHANGE**?
### Why is such a **break change** being merged without public discussion?
## The plugin does not contain a security vulnerability and we are not going to release a patch for it, so why perform such an urgent change operation?
## When you merge this break change, what is the equivalent alternative to this plugin what is it?
### If there are no alternatives and there is no advance notice of any deprecation plan, why did you just merge this plugin?

---

# This change isn't urgent at all, and we have plenty of time before the next release, which should have been handled in a better way! Clearly documenting the behavior and risks of the plugin, and telling users what's going on here from a technical perspective would all help to better address the issue. If it needs to be removed, start by initiating a change to the documentation to mark it as “Deprecated and about to be removed”, after which you can modify it to not start by default and print a log to remind users not to use it anymore.
# Is it acceptable to remove a plugin so easily? Basically you're telling developers that there are no rules to follow and you can make any disruptive change you want, even if that rule is clearly documented in the PR checklist. There is no reason whatsoever to support this non-backwards compatible PR being merged so hastily and quickly.

bzp2010:
> Apologies for using so many header lines to increase the font size, but it's important and necessary to do so.

@LiteSun @nic-6443 @AlinsRan @ronething @membphis 

# The author and all who approved this PR, I have questions:


## Where is the discussion on this **BREAK CHANGE**?
### Why is such a **break change** being merged without public discussion?
## The plugin does not contain a security vulnerability and we are not going to release a patch for it, so why perform such an urgent change operation?
## When you merge this break change, what is the equivalent alternative to this plugin what is it?
### If there are no alternatives and there is no advance notice of any deprecation plan, why did you just merge this plugin?

---

# This change isn't urgent at all, and we have plenty of time before the next release, which should have been handled in a better way! Clearly documenting the behavior and risks of the plugin, and telling users what's going on here from a technical perspective would all help to better address the issue. If it needs to be removed, start by initiating a change to the documentation to mark it as “Deprecated and about to be removed”, after which you can modify it to not enable by default and print a log to remind users not to use it anymore.

# Is it acceptable to remove a plugin so easily? Basically you're telling developers that there are no rules to follow and you can make any disruptive change you want, even if that rule is clearly documented in the PR checklist. There is no reason whatsoever to support this non-backwards compatible PR being merged so hastily and quickly.

# I believe that we need to revert this commit and then complete the process in a better way.

PR #12223, opened 2025-05-16
test: merge separately http3 cases

bzp2010:
### Description

APISIX workflow now tests HTTP3 and other cases separately, which results in more than double the number of workflows, more time-consuming, and more error possibilities.
Given that ngx http3 has been added to the (openresty's) default build since openresty 1.27, and apisix-runtime has long had it as a default build module, I think they can be merged in one place.

This PR aims to remove the separate http3 tests and reduce our mental overhead.

In addition to this, this PR also reduces the time required by a small amount by simply reordering the workflow execution.
Technically, the original workflow started all the required external components at the beginning, such as kafka, etc., which would have consumed some resources and undoubtedly slowed down the subsequent builds of the apisix-runtime, thus slowing down the overall time-consumption. This order has been adjusted to build first, then start the external components.

### Checklist


juzhiyuan (approved):

moonming (approved):

Baoyuantop (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #12226, opened 2025-05-18
test: improve typescript-based test framework

bzp2010:
### Description

Split from #12196, modifications in functionality and testing frameworks were introduced there, making those PR not pure.

First we can merge this PR, after that rebase that PR, we can make that PR to focus on the function, also easy to backport.

### Checklist

### DIFF
--- a/t/ts/admin_api.ts
+++ b/t/ts/admin_api.ts
@@ -14,18 +14,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { Config } from 'jest';
+import axios, { type Method, type RawAxiosRequestHeaders } from 'axios';
 
-const config: Config = {
-  coverageProvider: 'v8',
-  testEnvironment: 'node',
-  testRegex: '(/__tests__/.*|(\\.|/)(spec|test))\\.(ts|mts)$',
-  transform: {
-    '^.+\\.ts$': ['ts-jest', { useESM: false }],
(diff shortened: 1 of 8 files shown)


juzhiyuan (dismissed):
LGTM

juzhiyuan (approved):

t/ts/admin_api.ts
```diff
@@ -0,0 +1,35 @@
...
+  body?: object,
+  headers?: RawAxiosRequestHeaders,
+) => {
+  return axios.request({
+    method,
+    // TODO: use 9180 for admin api
+    url: `http://127.0.0.1:1984/${uri}`,
+    data: body,
+    headers: {
+      'X-API-KEY': 'edd1c9f034335f136f87ad84b625c8f1',
```
SkyeYoung, line 31: can we pick it from apisix config?

SkyeYoung (approved):
Others LGTM

t/ts/admin_api.ts
bzp2010: It could be. But the hardcoded ones seem to be sufficient for now.

This is because this server is spawned by test nginx and not by the new test framework itself (it doesn't even use the standard APISIX startup process), which sucks and will be improved later so that ts test can autonomously start or stop APISIX instances (right now it has to rely on a xxx.t shell file)

guoqqqi (approved):

PR #12256, opened 2025-05-26
fix(standalone): API-driven mode does not properly handle consumer schema

### LINKED ISSUE #12253: bug: API-driven standalone mode does not properly handle consumer schema checks
### Current Behavior

The `consumers` does not support inputs that contain a `modifiedIndex` field.

The `core.config.new(“/consumers”, cfg)` does not include an `item_schema` input similar to the other resources because consumers needs to handle both consumer and consumer_credential resources under the same key.
It just passes in a function that checks the schema and determines the resource type by key and performs specific checks if needed.
The additional field `modifiedIndex` that we have now introduced in config_yaml cannot be applied to the internal check function of consumers, so the check will output an error.

### Expected Behavior

Consistent with other resources.

### Error Logs

```text
invalid consumers at index 0, err: additional properties forbidden, found modifiedIndex
```

### Steps to Reproduce

1. Use the API to submit a consumer containing the `modifiedIndex`.

### Environment

- Operating system (run `uname -a`): N/A
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): N/A
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): N/A
- Plugin runner version, for issues related to plugin runners: N/A
- LuaRocks version, for installation issues (run `luarocks --version`): N/A

AlinsRan:

Fixes #12253 

### Checklist

### DIFF
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -156,7 +156,12 @@ local function update(ctx)
                     end
                 end
                 if item_checker then
-                    valid, err = item_checker(item_temp)
+                    local item_checker_key
+                    if item.id then
+                        -- credential need to check key
+                        item_checker_key = "/" .. key .. "/" .. item_temp.id
+                    end
+                    valid, err = item_checker(item_temp, item_checker_key)
                     if not valid then
                         core.log.error(err_prefix, err)
                         core.response.exit(400, {error_msg = err_prefix .. err})
@@ -235,6 +240,37 @@ function _M.run()
 end
 
 
+local schema_patch
+do
+    local resource_schema = {
+        "proto",
+        "global_rule",
+        "route",
+        "service",
+        "upstream",
+        "consumer",
+        "consumer_group",
+        "credential",
+        "ssl",
+        "plugin_config",
+    }
+    function schema_patch()
+        for _, name in ipairs(resource_schema) do
+            local schema = core.schema[name]
+            if not schema then
+                core.log.error("schema for ", name, " not found")
+                return nil, "schema for " .. name .. " not found"
+            end
+            if schema.properties then
+                schema.properties.modifiedIndex = {
+                    type = "integer",
+                }
+            end
+        end
+    end
+end
+
+
 function _M.init_worker()
     local function update_config()
         local config, err = shared_dict:get("config")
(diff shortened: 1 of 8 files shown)


apisix/core/config_yaml.lua
```diff
@@ -274,7 +275,7 @@ local function sync_data(self)
                           ", it should be an object")
             end
 
-            local id = item.id or ("arr_" .. idx)
+            local id = item.id or item.username or ("arr_" .. idx)
```
nic-6443, line 278: Is this change necessary? We have already set `id` to `username` for consumer in `apisix/consumer.lua`.
```
consumer.value.id = consumer.value.username
```

AlinsRan: If not, the default is values_hash ["arr_1"] = consumer{}, 

The filter is called after the cache is set. In the following calls, it will not get the consumer


image


image

nic-6443 (dismissed):

apisix/schema_def.lua
```diff
@@ -729,7 +729,18 @@ _M.credential = {
...
+            oneOf = {
+                id_schema,
+                -- id: "${consumer_name}/credentials/${credential_id}" from standalone mode
+                {
+                    type = "string",
+                    minLength = 15,
+                    maxLength = 128,
+                    pattern = [[^[a-zA-Z0-9]+/credentials/[a-zA-Z0-9]+$]],
+                }
+            }
```
bzp2010, line 742: If this is standalone-specific, it should be overridden when standalone is enabled, rather than using oneOf.
oneOf is ambiguous at some point and is not conducive to various downstream components using its type.

apisix/admin/standalone.lua
```diff
@@ -235,6 +240,36 @@ function _M.run()
 end
 
 
+local schema_patch
+do
+    local resource_schema = {
```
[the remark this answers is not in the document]
AlinsRan: I don't think it's necessary, `plugin_metadata` is not defined in schema_def, and its structure depends on plugin customization.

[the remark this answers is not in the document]
bzp2010: So the `modifiedIndex` on `plugin_metadata` actually benefits from the fact that the plugin itself doesn't declare `additionalProperties`, which doesn't prevent the plugin from writing without recognizing the field. Right?

apisix/schema_def.lua
bzp2010: Now, instead of modifying the schema after making a deep copy, this module just modifies it, which is as expected?

apisix/admin/standalone.lua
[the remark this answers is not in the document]
AlinsRan: Yes.

bzp2010 (approved):

PR #12276, opened 2025-06-02
feat: add embedded apisix dashboard ui

### LINKED ISSUE #12160: feat: support to expose APISIX Dashboard UI within APISIX
### Description

The APISIX team is developing a new `apache/apisix-dashboard`. According to https://github.com/apache/apisix-dashboard/issues/2981 and https://github.com/apache/apisix-dashboard/issues/2988, we plan to bundle the APISIX Dashboard UI within the APISIX Gateway. This will allow the APISIX Gateway to expose the Dashboard UI as Admin APIs.

## TODO

- [ ] Confirm Technical Design on `apache/apisix`

bzp2010:
### Description

Add a new embedded UI to APISIX as part of the https://github.com/apache/apisix-dashboard/issues/2981 plan.

This is the infrastructure for enabling the embedded dashboard. 
Subsequent work needs to be completed in the build process to include the built frontend files in the APISIX container build. This is expected to slightly increase the container image size.

Fixes #12160

### Checklist

### DIFF
--- a/t/cli/test_admin_ui.sh
+++ b/t/cli/test_admin_ui.sh
@@ -88,6 +88,28 @@ fi
 
 make stop
 
+# test ip restriction
+
+git checkout conf/config.yaml
+
+echo "
+deployment:
+    admin:
+        enable_admin_ui: true
+        allow_admin:
+            - 1.1.1.1/32
+" > conf/config.yaml
+
+make run
+
+code=$(curl -v -k -i -m 20 -o /dev/null -s -w %{http_code} http://127.0.0.1:9180/ui/)
+if [ ! $code -eq 403 ]; then
+    echo "failed: ip restriction not working, expected 403, got $code"
+    exit 1
+fi
+
+make stop
+
 # test admin ui disabled
 
 git checkout conf/config.yaml
@@ -100,8 +122,27 @@ deployment:
 
 make init
 
-grep "location ^~ /ui/" conf/nginx.conf > /dev/null
-if [ $? -eq 0 ]; then
+#### When grep cannot find the value, it uses 1 as the exit code.
+#### Due to the use of set -e, any non-zero exit will terminate the
+#### script, so grep is written inside the if statement here.
+if grep "location ^~ /ui/" conf/nginx.conf > /dev/null; then
     echo "failed: failed to disable embedded admin ui"
+    exit 1
+fi
+
+# test admin UI explicitly enabled
+
+git checkout conf/config.yaml
+
+echo "
+deployment:
+    admin:
+        enable_admin_ui: true
+" > conf/config.yaml
+
+make init
+
+if ! grep "location ^~ /ui/" conf/nginx.conf > /dev/null; then
+    echo "failed: failed to explicitly enable embedded admin ui"
     exit 1
 fi


apisix/cli/ngx_tpl.lua
```diff
@@ -646,6 +646,21 @@ http {
                 apisix.http_admin()
             }
         }
+
+        {% if enable_admin_ui then %}
+        location = /ui {
+            return 301 /ui/;
+        }
+        location ^~ /ui/ {
```
moonming, line 654: There are no IP or permission restrictions, which may cause security issues

conf/config.yaml.example
```diff
@@ -674,6 +674,7 @@ deployment:                    # Deployment configurations
       #   role: viewer
 
     enable_admin_cors: true       # Enable Admin API CORS response header `Access-Control-Allow-Origin`.
+    enable_admin_ui: true         # Enable embedded APISIX Dashboard UI.
```
moonming, line 677: if the role of APISIX is data plane, will `enable_admin_ui` be false automatically? Or we can set `enable_admin_ui` to false by default.

bzp2010: On the data plane, the Admin API won't actually start at all.
Since the ui enforces the use of the Admin API port (to minimize problems with CORS), it is implicit that if the Admin API is not enabled, the ui will not be enabled.

So the answer to that question is that if the Admin API is not enabled (e.g. by turning it off on traditional, or by using data plane mode), the config item will have no effect. That is, whatever it is set to, the ui will not be exposed.

This is accomplished by a template in `ngx_tpl`, where location blocks about `/ui/` are not rendered if the server block used by the Admin API does not exist.

apisix/cli/ngx_tpl.lua
bzp2010: There are no security concerns about unauthorized access as you understand it.

Even if that ui endpoint is exposed on the Internet, the user still needs to be able to access the Admin API (/apisix/admin/xx) and have the proper credentials. And this part will undoubtedly be IP-restricted depending on the configuration.
Separate access to the UI makes no sense; that thing will be built in CI and will not contain any data about the user's environment. Without credentials and permissions, that UI will always just report errors and not display any data.

However, I recognize that this does carry some risk, mainly from the fact that if the endpoint is unintentionally exposed, it may cause bandwidth consumption on the server, but not face any potential data leakage or illegal operations.

For that reason, I'll be introducing the IP restrictions that are currently in place on the Admin API to the new endpoint at the same time to alleviate that concern.

bzp2010: Moved allow/deny on `/apisix/admin` to the server block level, which will resolve this issue.

conf/config.yaml.example
bzp2010: So I don't think anything needs to be done about it.

apisix/cli/ngx_tpl.lua
bzp2010: Added test cases for this.

bzp2010:
### About the documentation
This integration process is not yet complete. We need to modify the APISIX build process to complete the front-end compilation in advance.
Therefore, the documentation may be submitted later via another PR.

SkyeYoung:
> We need to modify the APISIX build process to complete the front-end compilation in advance.

Perhaps you could try reusing the Dockerfile used to build the image in `apisix-dashboard`, and then copy out the final build artifacts.

bzp2010:
@SkyeYoung 

> Perhaps you could try reusing the Dockerfile used to build the image in apisix-dashboard, and then copy out the final build artifacts.

I would try to compile the frontend in the workflow and copy dist directly when building the container. Probably not modifying the Dockerfile significantly. For example, change to a multi-stage build.

membphis (approved):
LGTM

SkyeYoung (approved):

juzhiyuan:
> How should the Dashboard be integrated with APISIX? Should we embed the Dashboard's commit ID as part of the integration?

submodule (I prefer this way) or follow https://github.com/apache/apisix/blob/master/.requirements.

juzhiyuan (approved):
LGTM

bzp2010:
@moonming @juzhiyuan 

> How should the Dashboard be integrated with APISIX? Should we embed the Dashboard's commit ID as part of the integration?

> submodule (I prefer this way) or follow https://github.com/apache/apisix/blob/master/.requirements.

We need to adjust the APISIX build and release process to shift the functionality of the apisix-docker repository to the APISIX repository itself.
At that point we decided how to integrate in the build, which is now inconclusive.

No doubt these will be done before the next release.

First, let me merge this PR. After that, I will try to contact the ASF Infra team to configure Docker Hub credentials for the apisix repository. https://issues.apache.org/jira/browse/INFRA-26888

PR #12288, opened 2025-06-04
fix: correct pre/post hook typos in Kubernetes discovery and improve cleanup safety

ChuanFF:
there is a spelling error in k8s discovery module. `post_List` should be `post_list`

### Description

There is a spelling error in the k8s service discovery module code.   As shown below, the judgment condition in the code of `apisix/discovery/kubernetes/informer_factory.lua` should be `informer.post_list` instead of `informer.post_List`. The current bug causes the `post_list` function to not be executed.
```lua
    informer.fetch_state = "list finished"
    if informer.post_List then
        informer:post_list()
    end
```

### Checklist


Baoyuantop:
Hi @ChuanFF, need add test for this change.

ChuanFF:
@Baoyuantop just a typo, is the test  necessary?

Baoyuantop:
> The current bug causes the post_list function to not be executed.

We need to verify this.

ChuanFF:
There are two typos in the following code:  
`informer.pre_List` should be `informer.pre_list`,  
`informer.post_List` should be `informer.post_list`.

https://github.com/apache/apisix/blob/58066abc88df37a490f6c04011ed9588a0bda0d1/apisix/discovery/kubernetes/informer_factory.lua#L288-L303

#### Function Behavior:
- **`pre_list`** calls `shared_dict:flush_all()` ([docs](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#ngxshareddictflush_all)), which expires all keys in the shared dictionary but doesn't physically delete them.
- **`post_list`** calls `shared_dict:flush_expired()` ([docs](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#ngxshareddictflush_expired)), which physically deletes expired keys from the shared dictionary.

#### Issue After Fixing Typos:
When the typos are corrected, a new problem emerges:  
`flush_all` function seems will kill 3 key-value pairs data in shared_dict ([code reference](https://github.com/openresty/lua-nginx-module/blob/1f4d846f02cd3980d1f64fbc3e5359fd0e7716f6/src/ngx_http_lua_shdict.c#L1930)). This creates a critical window between the `pre_list` call and when new endpoints are populated. During this interval:
1. some endpoint data in the shared dictionary will be missing
2. requests may fail due to unavailable endpoint information

Therefore, this typos bug seems not to be fixed directly. I'm not sure whether the behavior of `shared_dict:flush_all()` deleting some data is a bug

Baoyuantop:
Hi @ChuanFF, I think we can optimize this by modifying the pre_list and pre_post methods: 
-  pre_list function only records existing data keys and does not clean up any data.
-  pre_post compares old and new data and removes old data that no longer exists.

ChuanFF:
@Baoyuantop if we want to  removes old data that no longer exists, there is a problem that how to records existing data keys,`ngx.shared.DICT.get_keys` does not seem to be suitable（official document have a caution: Avoid calling this method on dictionaries with a very large number of keys as it may lock the dictionary for significant amount of time and block Nginx worker processes trying to access the dictionary.）

I think we can  prepare two tables, `data_history` and `data_existing`, recording data to `data_existing` during `list` and `watch`, assigning `data_existing` to `data_history` in `pre_list` and setting `data_existing` to {}, then comparing `data_history` with `data_existing` in `post_list` and removes old data that no longer exists.

Test cases seem  difficult to write, dirty data is generated only after the watch ends and before the next list call, i do not know how to verify old data that no longer exists, do you have any suggest？

Or can we ignore dirty data? After all, the conditions for dirty data to be generated are harsh, and it seems to have no effect.

Baoyuantop:
Lua table cannot share data between multiple workers.

ChuanFF:
Lua table just used for record data to clear dirty data in shared_dict.

Baoyuantop:
Hi @ChuanFF, before continuing to complete the code, could you please describe your current modification plan in the description? This will help other maintainers review this PR.

ChuanFF:
@Baoyuantop Sorry, I've been trying to fix this issue in my own repository for the past couple of days and didn't realize the PR would sync here as well. I have now completed the code changes. The approach is:

In the `pre_list` function, I record the existing keys from the shared_dict into a temporary table `handle.existing_keys`. Then, when APISIX lists Kubernetes endpoints/endpointSlices, I record the keys written to the shared_dict into another temporary table `handle.current_keys_hash`. Essentially, `handle.existing_keys` represents the old list of keys, while `handle.current_keys_hash` represents the new list of keys. Finally, in the `post_list` function, I compare the old and new data to identify and remove any dirty data.

ChuanFF:
Additionally, the test case requires waiting for the list query to finish, so a 2-second sleep was added in the location `/t`. I plan to submit an additional PR that improve  the `function _M.status_ready()` in apisix/init.lua  to check if APISIX's Kubernetes service discovery is ready. Once that is implemented, the 2-second sleep will be replaced with a wait for the readiness check.

ChuanFF:
> Hi @ChuanFF, before continuing to complete the code, could you please describe your current modification plan in the description? This will help other maintainers review this PR.

@Baoyuantop The modification has been completed. Can we start the review？

Baoyuantop:
Hi @ChuanFF, could you please fix the code lint error?

ChuanFF:
> Hi @ChuanFF, could you please fix the code lint error?

@Baoyuantop hello, lint error has been fix

t/kubernetes/discovery/kubernetes3.t
```diff
@@ -242,6 +242,13 @@ _EOC_
             }
         }
 
+        location /t {
+            content_by_lua_block {
+                ngx.sleep(2)
+                ngx.exit(200)
+            }
+        }
+
```
Baoyuantop, line 251: Why modify this part?

Baoyuantop (commented):
It is recommended to update the title and description of the PR, as it is currently not just a typo fix.

t/kubernetes/discovery/kubernetes3.t
Baoyuantop: Hi @ChuanFF, please check

ChuanFF: the test case requires waiting for the k8s endpoints list query to finish, so i add a location /t to sleep 2 second. I plan to submit an additional PR that improve the function _M.status_ready() in apisix/init.lua to check if APISIX's Kubernetes service discovery is ready. Once that is implemented, the 2-second sleep will be replaced with a wait for the readiness check.

ChuanFF:
i see there are 2 failing check in ci, but they seem unrelated to my modifications?

Baoyuantop:
> i see there are 2 failing check in ci, but they seem unrelated to my modifications?

The main branch contains fixes for failed tests; you can merge them.

Baoyuantop:
Hi @ChuanFF, we've made some fixes to CI in https://github.com/apache/apisix/pull/12805. You can merge the latest main branch to get CI pass.

ChuanFF:
lastest main branch now merged

Baoyuantop (approved):

PR #12291, opened 2025-06-06
feat: enhance admin api filter

bzp2010:
### Description

Added filter enhancements to the Admin API, allowing routes and stream routes to be filtered by `service_id` and `upstream_id`.

### Checklist

### DIFF
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -120,11 +120,44 @@ local function pagination(body, args)
 end
 
 
-local function filter(body, args)
-    if not args.name and not args.label and not args.uri then
-        return
+local function _filter(item, args, resource)
+    if not args.filter then
+        return true
     end
 
+    local filters, err = ngx.decode_args(args.filter or "", 100)
+    if not filters then
+        log.error("failed to decode filter args: ", err)
+        return false
+    end
+
+    for key, value in pairs(filters) do
+        if not resource.list_filter_fields[key] then
+            log.warn("filter field '", key, "' is not supported by resource: ", resource.name)
+            goto CONTINUE
+        end
+
+        if not item[key] then
+            return false
+        end
+
+        if type(value) == "table" then
+            value = value[#value] -- get the last value in the table
+        end
+
+        local matched = re_find(item[key], value, "jo")
+        if not matched then
+            return false
+        end
+
+        ::CONTINUE::
+    end
+
+    return true
+end
+
+
+local function filter(body, args, resource)
     for i = #body.list, 1, -1 do
         local name_matched = true
         local label_matched = true
@@ -168,14 +201,15 @@ local function filter(body, args)
             end
         end
 
-        if not name_matched or not label_matched or not uri_matched then
+        if not name_matched or not label_matched or not uri_matched
+                            or not _filter(body.list[i].value, args, resource) then
             table.remove(body.list, i)
         end
     end
 end
 
 
-function _M.filter(body)
+function _M.filter(body, resource)
     if not enable_v3() then
         return body
     end
@@ -194,7 +228,7 @@ function _M.filter(body)
 
     -- filter and paging logic for list query only
     if processed_body.list then
-        filter(processed_body, args)
+        filter(processed_body, args, resource)
 
         -- calculate the total amount of filtered data
         processed_body.total = processed_body.list and #processed_body.list or 0
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -250,10 +250,8 @@ local function run()
         else
             core.response.set_header("X-API-VERSION", "v2")
         end
-        if resource.need_v3_filter then
-            data = v3_adapter.filter(data)
-        end
 
+        data = v3_adapter.filter(data, resource)
         data = strip_etcd_resp(data)
 
         core.response.exit(code, data)
--- a/apisix/admin/resource.lua
+++ b/apisix/admin/resource.lua
@@ -25,10 +25,8 @@ local type = type
 
 
 local _M = {
-    need_v3_filter = true,
+    list_filter_fields = {},
 }
-
-
 local mt = {
     __index = _M
 }
--- a/apisix/admin/routes.lua
+++ b/apisix/admin/routes.lua
@@ -143,5 +143,6 @@ return resource.new({
     name = "routes",
     kind = "route",
     schema = core.schema.route,
-    checker = check_conf
+    checker = check_conf,
+    list_filter_fields = { "service_id", "upstream_id" },
 })
--- a/apisix/admin/stream_routes.lua
+++ b/apisix/admin/stream_routes.lua
@@ -73,5 +73,6 @@ return resource.new({
     kind = "stream route",
     schema = core.schema.stream_route,
     checker = check_conf,
-    unsupported_methods = {"patch"}
+    unsupported_methods = { "patch" },
+    list_filter_fields = { "service_id", "upstream_id" },
 })


docs/en/latest/admin-api.md
```diff
@@ -278,6 +278,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
 }
 ```
 
+### Support reference filtering query
+
+:::note
+
+This feature was introduced in APISIX 3.13.0.
```
kayx23, line 285: Don't think this sentence is necessary? This doc change will go into "next" and only become visible in 3.13.0 version once released.

I think we can get rid of this note and just put the sentence below in a regular paragraph.

> APISIX supports querying routes and stream routes by `service_id` and `upstream_id`. Other resources or fields are not currently supported.

WDYT?

docs/en/latest/admin-api.md
```diff
@@ -278,6 +278,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
...
+
+When getting a list of resources, it supports a `filter` for filtering resources by filters.
+
+It is encoded in the following manner.
+
+```text
+filter=escape_uri(key1=value2&key2=value2)
+```
+
+The following example filters routes using `service_id`. When multiple filters are input at the same time, the result will be the intersection of them.
```
kayx23:
```suggestion
The following example filters routes using `service_id`. Applying multiple filters simultaneously will return results that match all filter conditions.
```

docs/en/latest/admin-api.md
bzp2010: I did encounter this. Users do not take care to manually select the older version of the documentation they are using, and try a particular new version of a supported feature on an older version directly from the documentation.

This is confusing. Therefore, I added it and tried to reduce some of the explanation costs. If you think it is unnecessary, I will delete it.

SkyeYoung (approved):

AlinsRan (approved):

apisix/admin/v3_adapter.lua
```diff
@@ -120,11 +122,44 @@ local function pagination(body, args)
...
+
+        if not item[key] then
+            return false
+        end
+
+        if type(value) == "table" then
+            value = value[#value] -- get the last value in the table
+        end
+
+        local matched = re_find(item[key], value, "jo")
```
nic-6443: The filter operator we provide is equality, but here a substring containment implementation is used, which feels mismatched.

nic-6443: For example, for `service_id=1`, the user is expected to filter explicitly based on service_id being equal to 1 rather than containing 1.

bzp2010: In fact, it's hard to tell which fields need to be exact matches and which need to be fuzzy matches, and the current implementation is a duplication of previous filter parameters. I recognize that `name`, `uri` are not the same concept with exact ID in nature, but I don't have any good ideas at the moment for the following reasons.

When the feature was written at the beginning, I expected it to be able to filter on any field inside any resource without more development work, so this fuzzy match was used. I later realized that how it could filter on any field, I would have a hard time dealing with fields that have complex nesting, such as upstream. so I finally decided to limit the scope to service_id and upstream_id on routes and stream routes.

I didn't want to exclusively develop based on specific fields, but wanted to keep it generic. Right now the schema can't provide this additional information for this goal, so I have to resort to fuzzy matching.

bzp2010: So I'm going to change it to an exact match?

SkyeYoung: @bzp2010  Maybe a table can be maintained to distinguish patterns? Or perform an exact match ending with `_id`?

nic-6443: I think if we haven't figured out how to support fields like `name` and `uri`, I suggest implementing `=` as an exact match first for `xx_id` filter. This way, if we need to add support for other fields in the future, we can extend it at the operator position (like `~=` in lua-resty-expr lib).

bzp2010: Moved to exact match

docs/en/latest/admin-api.md
```diff
@@ -278,6 +278,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
 }
 ```
 
+### Support reference filtering query
+
+:::note
+
+This feature was introduced in APISIX 3.13.0.
+
+It now supports querying by `service_id` and `upstream_id` on routes and stream routes. Other resources or other fields are not supported.
```
kayx23:
```suggestion
APISIX supports querying routes and stream routes by `service_id` and `upstream_id`. Other resources or fields are not currently supported.
```

others LGTM

apisix/admin/v3_adapter.lua
```diff
@@ -120,11 +122,43 @@ local function pagination(body, args)
...
+
+    local filters, err = ngx.decode_args(args.filter or "", 100)
+    if not filters then
+        log.error("failed to decode filter args: ", err)
+        return false
+    end
+
+    for key, value in pairs(filters) do
+        if not resource.list_filter_fields[key] then
+            log.warn("filter field '", key, "' is not supported by resource: ", resource.name)
```
membphis, line 138: need a test case to confirm it

docs/en/latest/admin-api.md
```diff
@@ -278,6 +278,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
...
+It is encoded in the following manner.
+
+```text
+filter=escape_uri(key1=value2&key2=value2)
+```
+
+The following example filters routes using `service_id`. Applying multiple filters simultaneously will return results that match all filter conditions.
+
+```shell
+curl 'http://127.0.0.1:9180/apisix/admin/routes?filter=service_id%3D1' \
```
SkyeYoung, line 302: one question: can we adjust to use the format `a[b]=c&a[d]=e`? It seems more common. https://www.npmjs.com/package/qs

bzp2010: There is no existing API for parsing query strings for such objects in OpenResty, so it may have to use regular expressions to do so.
I don't want to do that right now. The existing format can be implemented in any language client, it's nothing more than a URI-encoded string.


```lua
ngx.log(ngx.ERR, require("cjson").encode(ngx.decode_args("filter%5Bservice_id%5D=a&filter%5Bservice_id%5D=b&b=c&c=c")))
```

```text
[lua] test.lua:1: {"filter[service_id]":["a","b"],"b":"c","c":"c"}
```

apisix/admin/v3_adapter.lua
bzp2010: Added

docs/en/latest/admin-api.md
bzp2010: Fixed

docs/en/latest/admin-api.md
bzp2010: image

It's too complicated to support something like this.

kayx23 (dismissed):
docs LGTM

SkyeYoung (approved):

nic-6443 (dismissed):

docs/en/latest/admin-api.md
```diff
@@ -278,6 +278,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
...
+APISIX supports querying routes and stream routes by `service_id` and `upstream_id`. Other resources or fields are not currently supported.
+
+:::
+
+When getting a list of resources, it supports a `filter` for filtering resources by filters.
+
+It is encoded in the following manner.
+
+```text
+filter=escape_uri(key1=value2&key2=value2)
```
juzhiyuan:
```suggestion
filter=escape_uri(key1=value1&key2=value2)
```

docs/zh/latest/admin-api.md
```diff
@@ -284,6 +284,42 @@ curl 'http://127.0.0.1:9180/apisix/admin/routes?name=test&uri=foo&label=' \
...
+APISIX 支持通过 `service_id` 和 `upstream_id` 查询路由和 Stream 路由。现在不支持其他资源或字段。
+
+:::
+
+在获取资源列表时，你可以使用 `filter` 参数过滤资源。
+
+它以以下方式编码：
+
+```text
+filter=escape_uri(key1=value2&key2=value2)
```
juzhiyuan:
```suggestion
filter=escape_uri(key1=value1&key2=value2)
```

docs/en/latest/admin-api.md
bzp2010: Fixed

docs/zh/latest/admin-api.md
bzp2010: Fixed

membphis (approved):

kayx23 (approved):

SkyeYoung (approved):

AlinsRan (approved):

PR #12304, opened 2025-06-10
fix: handle consul nil port cases by defaulting to port 80

oil-oil:
### Description

When using consul as a service registry, if a user registers a service without filling in the port number, the default value is not set in apisix, resulting in `svc_port` possibly being `nil`, which in turn results in `local service_id = svc_address ... “:” ... svc_port` cannot be spliced.

Fixes https://github.com/apache/apisix/issues/11134

### Checklist

### DIFF
--- a/apisix/discovery/consul/init.lua
+++ b/apisix/discovery/consul/init.lua
@@ -516,8 +516,8 @@ function _M.connect(premature, consul_server, retry_delay)
                     end
 
                     local svc_address, svc_port = node.Service.Address, node.Service.Port
-                    -- Handle nil port case - default to 80 for HTTP services
-                    if not svc_port then
+                    -- Handle nil or 0 port case - default to 80 for HTTP services
+                    if not svc_port or svc_port == 0 then
                         svc_port = 80
                     end
                     -- if nodes is nil, new nodes table and set to up_services
--- a/apisix/discovery/consul/init.lua
+++ b/apisix/discovery/consul/init.lua
@@ -225,8 +225,8 @@ local function watch_catalog(consul_server)
     ::RETRY::
     local watch_result, watch_err = client:get(consul_server.consul_watch_catalog_url)
     local watch_error_info = (watch_err ~= nil and watch_err)
-                             or ((watch_result ~= nil and watch_result.status ~= 200)
-                             and watch_result.status)
+                            or ((watch_result ~= nil and watch_result.status ~= 200)
+                            and watch_result.status)
     if watch_error_info then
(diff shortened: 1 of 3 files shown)


apisix/discovery/consul/init.lua
```diff
@@ -225,8 +225,8 @@ local function watch_catalog(consul_server)
     ::RETRY::
     local watch_result, watch_err = client:get(consul_server.consul_watch_catalog_url)
     local watch_error_info = (watch_err ~= nil and watch_err)
```
membphis: shorter is clearer, you can take a look. if you would like my way, you can make a try

```lua
    local res, err = client:get(consul_server.consul_watch_catalog_url)
    err = err or (res and res.status ~= 200 and res.status)
```

membphis (dismissed):
LGTM

nic-6443 (approved):

apisix/discovery/consul/init.lua
```diff
@@ -225,8 +225,8 @@ local function watch_catalog(consul_server)
...
-                             or ((watch_result ~= nil and watch_result.status ~= 200)
-                             and watch_result.status)
+                            or ((watch_result ~= nil and watch_result.status ~= 200)
+                            and watch_result.status)
```
bzp2010: This appears to be a code style modification only. If it does not contain substantive changes, please leave it as is.

apisix/discovery/consul/init.lua
```diff
@@ -516,6 +516,10 @@ function _M.connect(premature, consul_server, retry_delay)
                     end
 
                     local svc_address, svc_port = node.Service.Address, node.Service.Port
+                    -- Handle nil or 0 port case - default to 80 for HTTP services
+                    if not svc_port or svc_port == 0 then
```
bzp2010, line 520: TCP and UDP ports can be 0. 

Although it stands for "special use", it should be clearly distinguished from port is null. If this value is explicitly configured as 0, then we should accept it and not change it.

bzp2010: We have verified that nginx/openresty does not allow `0` as a port, even though TCP/UDP is designed to allow it.
As verified by @oil-oil, Consul's default behavior on Port has changed; it used to be `null` when registered without entering it; now it auto-populates with `0` as the default.
Since APISIX is primarily oriented towards HTTP scenarios, we fallback to `80` (the HTTP default) if the Port is `null` (the old behavior) or `0` (the new behavior, but not allowed by nginx).

This behavior needs to be documented.

bzp2010 (approved):

PR #12322, opened 2025-06-13
fix(workflow): use runners with different architectures instead of QEMU

SkyeYoung:

Fixes #

### Checklist

### DIFF
--- a/.github/workflows/push-dev-image-on-commit.yml
+++ b/.github/workflows/push-dev-image-on-commit.yml
@@ -12,8 +12,16 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
+  build-test-push:
+    strategy:
+      matrix:
+        include:
+          - runner: ubuntu-latest
+            arch: amd64
+          - runner: ubuntu-latest-arm64
+            arch: arm64
+
+    runs-on: ${{ matrix.runner }}
 
     env:
       APISIX_DOCKER_TAG: master-debian-dev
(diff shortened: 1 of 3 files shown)


.github/workflows/push-dev-image-on-commit.yml
```diff
@@ -47,22 +57,25 @@ jobs:
             exit 125
           fi
 
-      - name: Login to Docker Hub
```
bzp2010, line 50: Don't we need to log in `build-test-push`?

Makefile
```diff
@@ -497,23 +497,27 @@ ci-env-stop:
...
+	$(ENV_DOCKER) push $(IMAGE_NAME):dev-$(ENV_OS_ARCH)
+	@$(call func_echo_success_status, "$@ -> [ Done ]")
+
+### merge-dev-tags : Merge architecture-specific dev tags into a single dev tag
+.PHONY: merge-dev-tags
+merge-dev-tags:
+	@$(call func_echo_status, "$@ -> [ Start ]")
+	$(ENV_DOCKER) manifest create $(IMAGE_NAME):dev \
+		$(IMAGE_NAME):dev-amd64 \
+		$(IMAGE_NAME):dev-arm64
```
bzp2010, line 521: Do we need to remove these individual images from Docker Hub?

bzp2010:
BTW, they seem too long.

image

.github/workflows/push-dev-image-on-commit.yml
SkyeYoung: They were removed by mistake, let me add them back.

Makefile
SkyeYoung: Doesn't it seem like a necessary operation? Just a matter of choice

SkyeYoung: If not deleted, it's convenient to pull images of non-device architectures on some devices (like a MacBook) for testing purposes, but this use case is rare.

SkyeYoung: I tried searching, but it seems there isn't a related command to delete this remote tag. Only this https://docs.docker.com/reference/cli/docker/manifest/rm/

SkyeYoung: I haven't found a faster way to provide dev image.

In order to only upload the dev tag image, I seem to have to cache and download the built image to the current job, then upload it.

SkyeYoung:
> BTW, they seem too long.
> 
> image

We can do this in another PR

.github/workflows/push-dev-image-on-commit.yml
SkyeYoung: done

bzp2010: I didn't see it. 🤔

bzp2010 (approved):

SkyeYoung:
This is much faster than the previous method, so please review it as soon as possible.

image
https://github.com/apache/apisix/actions/runs/15670730008/job/44141280587

CleanShot 2025-06-16 at 17 50 10@2x
https://github.com/apache/apisix/actions/runs/15677420274/job/44160830531?pr=12322

nic-6443 (dismissed):

SkyeYoung:
@membphis ping

bzp2010 (approved):

juzhiyuan (approved):

PR #12333, opened 2025-06-17
feat(standalone): support JSON format

SkyeYoung:
### Description

**What**

Support JSON as configuration format in standalone file mode

**How**

Extend `apisix/core/config_yaml.lua`. 

Therefore, without modifying `config_provider: yaml`, it is compatible with the JSON format.

While this is a bit weird at the moment, I think we can change `yaml` to `file` or other more appropriate words in the future, rather than providing a new option. 

Because as you can see from the code, the only difference is how to read different configuration files. The current method helps us support other different formats in the future, such as `toml` or other file formats that users prefer.

**Why**

Yaml parsing is much slower than JSON.

**Test Cases**

All cases are copy and changed from `t/config-center-yaml/*.t`.

Fixes #

### Checklist

### DIFF
--- a/docs/en/latest/deployment-modes.md
+++ b/docs/en/latest/deployment-modes.md
@@ -138,7 +138,7 @@ deployment:
     config_provider: yaml
 ```
 
-You can also provide the configuration in JSON format by placing it in `conf/apisix.json`. At this point, you need to change the `deployment.role_data_plane.config_provider` to `json`.
+You can also provide the configuration in JSON format by placing it in `conf/apisix.json`. Before proceeding, you should change the `deployment.role_data_plane.config_provider` to `json`.
 
 Refer to the example below:
 


membphis:
some test cases failed, pls fix them first

docs/en/latest/deployment-modes.md
```diff
@@ -136,6 +136,8 @@ deployment:
 
 This makes it possible to disable the Admin API and discover configuration changes and reloads based on the local file system.
 
+*Note*: When using `config_provider: yaml`, we can also use JSON, just provide the configuration in `conf/apisix.json`.
```
kayx23:
```suggestion
*Note*: When using `config_provider: yaml`, you can also provide the configuration in JSON format by placing it in `conf/apisix.json`.
```

docs/en/latest/deployment-modes.md
```diff
@@ -293,7 +295,7 @@ routes:
 #END
 ```
 
-*WARNING*: APISIX will not load the rules into memory from file `conf/apisix.yaml` if there is no `#END` at the end.
+*WARNING*: APISIX will not load the rules into memory from file `conf/apisix.yaml` if there is no `#END` at the end. But using `conf/apisix.json` eliminates the need for `#END`, as APISIX will directly parse to validate the legitimacy of the JSON.
```
kayx23:
```suggestion
*WARNING*: APISIX will not load the rules into memory from file `conf/apisix.yaml` if there is no `#END` at the end. However, when using `conf/apisix.json`, the `#END` marker is not required, as APISIX can directly parse and validate the JSON structure.
```

docs/en/latest/deployment-modes.md
```diff
@@ -327,6 +329,23 @@ routes:
 #END
 ```
 
+```json5
+// JSON configuration can be directly translated from YAML, this is just an example.
```
kayx23: Presentation wise this is not the best as there are a number of examples provided for standalone configuration in parallel: https://apisix.apache.org/docs/apisix/next/deployment-modes/#standalone

Better presentation is to use tabs (YAML, JSON) and add JSON configs for all examples. Can use GPT to do so if the details are all 1:1.

SkyeYoung: OK, I will wait for others to finish the feedback on the code part and repair the document part.

If I remember correctly, this might involve converting the document to MDX.

docs/en/latest/deployment-modes.md
```diff
@@ -293,7 +295,7 @@ routes:
 #END
 ```
 
-*WARNING*: APISIX will not load the rules into memory from file `conf/apisix.yaml` if there is no `#END` at the end.
+*WARNING*: APISIX will not load the rules into memory from file `conf/apisix.yaml` if there is no `#END` at the end. However, when using `conf/apisix.json`, the `#END` marker is not required, as APISIX can directly parse and validate the JSON structure.
```
AlinsRan: You need to clarify the priority of the `apisix.yaml` and `apisix.json` configuration files.
When both files exist, which one is valid.

apisix/core/config_yaml.lua
```diff
@@ -69,9 +68,36 @@ local mt = {
...
+
+local apisix_yaml_path = profile:yaml_path("apisix")
+local apisix_json_path = ngx.re.sub(apisix_yaml_path, [[\.yaml$]], ".json", "jo")
+local file_configs = {
+    {path = apisix_yaml_path, type = "yaml"},
+    {path = apisix_json_path, type = "json"}
+}
+
+local function get_config_file_info()
+    local paths_str = ""
```
AlinsRan: The variable name is not clear, it is only used to record errors.

This function has many return values, you can consider using a table.

refer to:
```
local function get_config_file_info()
    local not_found_paths = {}

    for _, config in ipairs(file_configs) do
        local attr, err = lfs.attributes(config.path)
        if attr then
            return config.path, config.type, attr.modification
        else
            table.insert(not_found_paths, config.path)
            log.warn("failed to fetch attributes for ", config.path, ": ", err)
        end
    end

    local paths_str = table.concat(not_found_paths, ", ")
    return nil, nil, nil, "Failed to find any configuration file with path " .. paths_str
end
```

SkyeYoung: After communicating with membphis and others, I found that it needed to be reconstructed, and the current one is indeed not good enough.

apisix/core/config_yaml.lua
```diff
@@ -127,36 +186,15 @@ local function read_apisix_yaml(premature, pre_mtime)
...
-    f:seek('set')
-    local yaml_config = f:read("*a")
-    f:close()
-
-    local apisix_yaml_new = yaml.load(yaml_config)
-    if not apisix_yaml_new then
-        log.error("failed to parse the content of file " .. apisix_yaml_path)
+    local config_new, err = config_file.parse()
+    if err then
+        log.error("failed to parse the content of file " .. config_file.path .. ": " .. err)
```
membphis: performance

membphis (dismissed):
I mainly checked the Lua code, it seems good to me

pls fix one minor issue

apisix/core/config_yaml.lua
SkyeYoung: done

SkyeYoung:
I tried to use `yarn start:website` of `apisix-website` locally for preview, and it also works without changing to `mdx`:

image

Well, the style is very ugly, but it works. 😅

docs/en/latest/deployment-modes.md
```diff
@@ -134,6 +138,17 @@ deployment:
     config_provider: yaml
 ```
 
+You can also provide the configuration in JSON format by placing it in `conf/apisix.json`. At this point, you need to change the `deployment.role_data_plane.config_provider` to `json`.
```
kayx23:
```suggestion
You can also provide the configuration in JSON format by placing it in `conf/apisix.json`. Before proceeding, you should change the `deployment.role_data_plane.config_provider` to `json`.
```

kayx23:
> I tried to use `yarn start:website` of `apisix-website` locally for preview, and it also works without changing to `mdx`:
> 
> image
> Well, the style is very ugly, but it works. 😅

The changes look ok to me. However if you have any constructive suggestion, please bring it forward.

SkyeYoung:
> The changes look ok to me. However if you have any constructive suggestion, please bring it forward.

I'm just saying these CSS styles are making my eyes hurt. 

If I had time, I'd rather rebuild the APISIX website docs part from scratch to solve a lot of the existing problems.

membphis (approved):
LGTM now

bzp2010 (approved):
code looks good

nic-6443 (approved):

PR #12338, opened 2025-06-17
feat: add max pending entries option to batch-processor

### LINKED ISSUE #12628: discuss: set maximum limit to avoid memory issue when burst requests
### Current Behavior

If APISIX receives a burst requests over a short period of time, the APISIX Batch Processor may lead to memory pressure because it doesn't support to set a maximum queues capacity limit.

- https://github.com/apache/apisix/blob/master/docs/en/latest/batch-processor.md

### Expected Behavior

Allow to set the queue capacity limit to avoid burst requests causing memory issues.

### Error Logs

_No response_

### Steps to Reproduce

N/A

### Environment

- APISIX version (run `apisix version`): 3.13

Revolyssup:
Fixes #
When Kafka server is slow or unavailable, the logger objects keep accumulating faster than they are released causing memory spike. 

The solution is to implement a limit in batch processor manager that will allow it to drop new objects pushed to it, if a given number of callbacks have not been processed.

## Reproduction steps
1.  Start APISIX (1 worker process), Kafka Server 
2. Create a route and enable kafka-logger. Instance configuration:
kafka-logger:
  batch_max_size: 80
  brokers:
    - host: 127.0.0.1
      port: 9092
  buffer_duration: 180
  cluster_name: 1
  inactive_timeout: 5
  kafka_topic: test2
  max_retry_count: 1
  meta_format: default
  meta_refresh_interval: 30
  name: service kafka logger
  producer_batch_num: 200
  producer_batch_size: 104857600
  producer_max_buffering: 50000
  producer_time_linger: 1
  producer_type: async
  required_acks: 1
  retry_delay: 1
  timeout: 5
3. Test the request with the curl command and make sure the request information is sent to the kafka server
4.  Limit the CPU usage of the kafka server (simulating the slow receiving speed of the kafka server), for example, limit the CPU to 0.1 CUP 
sudo docker update --cpus 0.1 apisix_kafka
5.  Use wrk/wrk2 to send pressure to the specified route, and you will soon see the memory usage of the APISIX worker increase rapidly. When it grows to a certain size, it will be killed by the system, causing the worker process to  crash and exit , and then APISIX will start a new worker process. 
wrk2 -c 200 -d 60 -t 4 -R 50000 http://xxxxx/test

### Checklist

Fixes #12628


apisix/plugins/kafka-logger.lua
```diff
@@ -128,6 +128,7 @@ local schema = {
         producer_max_buffering = {type = "integer", minimum = 1, default = 50000},
         producer_time_linger = {type = "integer", minimum = 1, default = 1},
         meta_refresh_interval = {type = "integer", minimum = 1, default = 30},
+        max_pending_entries = {type = "integer", minimum = 0}
```
Revolyssup: this limit is per worker

apisix/utils/batch-processor-manager.lua
```diff
@@ -29,6 +29,7 @@ function _M.new(name)
     return setmetatable({
         stale_timer_running = false,
         buffers = {},
+        total_pushed_entries = 0,
```
nic-6443, line 32: If you record this number through a member of this module, it means that all places where the module is loaded via require 
```
local batch_processor = require("apisix.utils.batch-processor")
```
share the same counter. Shouldn't there be an independent counter for each plugin configuration since `max_pending_entries` can be configured at the specific plugin level? I think we should add this counter to every `batch_processor ` instance. https://github.com/apache/apisix/blob/cba5ba93d49886972228c0db44fb49746d4ec156/apisix/utils/batch-processor.lua#L148-L176

apisix/plugins/kafka-logger.lua
nic-6443: we should add this configuration in all other logger plugins that support batch-processor.

apisix/utils/batch-processor.lua
```diff
@@ -107,7 +107,7 @@ function execute_func(premature, self, batch)
         end
         return
     end
-
+    self.processed_entries = self.processed_entries + #batch.entries
```
nic-6443, line 110: please read above code, there have `first_fail` case that meaning sent part of `batch_entries` successfully.

nic-6443, line 112: please read the above code, for whole `not ok` case, the failed batch will be discard after exceed `max_retry_count`, in this case, batch will be free too, so it should be counting in `processed_entries` too.

nic-6443: If the number of these discarded entries is not recorded, the `processed_entries` count will be inaccurate when there is an fault in the Kafka server. This way, even if Kafka returns to normal, the plugin will not work due to the inaccurate `pending_entries`.

nic-6443 (commented):
can we add test cases for this feature? start a mocking tcp server that don't read data from connected socket, this will accumulating pending log entry in apisix, and we can set `max_pending_entries` to small value to confirm discard code is working.

apisix/utils/batch-processor.lua
Revolyssup: fixed it

nic-6443: please read `first_fail` case again that will update `batch.entries` after sent paritial batch successfully, in this case, `self.processed_entries` still has problem.

apisix/utils/batch-processor-manager.lua
```diff
@@ -94,11 +108,17 @@ function _M:add_entry(conf, entry)
...
+    self.total_pushed_entries = self.total_pushed_entries + 1
     return true
 end
 
 
-function _M:add_entry_to_new_processor(conf, entry, ctx, func)
+function _M:add_entry_to_new_processor(conf, entry, ctx, func, max_pending_entries)
+    if max_pending_entries and
+       self.total_pushed_entries - total_processed_entries(self) > max_pending_entries then
+        core.log.error("max pending entries limit exceeded. discarding entry")
```
membphis: need to print more useful information, eg: max_pending_entries, `self.total_pushed_entries - total_processed_entries(self)`

apisix/utils/batch-processor-manager.lua
```diff
@@ -85,7 +86,20 @@ do
 end
 
 
-function _M:add_entry(conf, entry)
+local function total_processed_entries(self)
+    local processed_entries = 0
+    for _, log_buffer in pairs(self.buffers) do
```
membphis, line 91: I am little worried it will cause bad performance

need a benchmark to confirm it

membphis (commented):
In this pr, I think we only need to update one plugin `kafka-logger`, this should be enough

We can change other plugins later via separate PRs

apisix/utils/batch-processor-manager.lua
```diff
@@ -85,7 +86,25 @@ do
...
+        processed_entries = processed_entries + log_buffer.processed_entries
+    end
+    return processed_entries
+end
+
+function _M:add_entry(conf, entry, max_pending_entries)
+    if max_pending_entries then
+        local total_processed_entries = total_processed_entries(self)
+        if self.total_pushed_entries - total_processed_entries > max_pending_entries then
+        core.log.error("max pending entries limit exceeded. discarding entry.",
```
membphis: bad indentation

image

apisix/utils/batch-processor-manager.lua
```diff
@@ -94,11 +113,22 @@ function _M:add_entry(conf, entry)
...
     return true
 end
 
 
-function _M:add_entry_to_new_processor(conf, entry, ctx, func)
+function _M:add_entry_to_new_processor(conf, entry, ctx, func, max_pending_entries)
+    if max_pending_entries then
+        local total_processed_entries = total_processed_entries(self)
+        if self.total_pushed_entries - total_processed_entries > max_pending_entries then
+        core.log.error("max pending entries limit exceeded. discarding entry.",
```
membphis: ditto

apisix/plugins/kafka-logger.lua
```diff
@@ -128,6 +129,7 @@ local schema = {
         producer_max_buffering = {type = "integer", minimum = 1, default = 50000},
         producer_time_linger = {type = "integer", minimum = 1, default = 1},
         meta_refresh_interval = {type = "integer", minimum = 1, default = 30},
+
```
nic-6443: useless change.

apisix/utils/batch-processor-manager.lua
nic-6443: According to my understanding, the number of `self.buffers` is proportional to the number of plugin configurations. For example, if kafka-logger is configured on 10 routes, there will be 10 `buffers`, so the quantity is controllable. 
@Revolyssup Am I right?

Revolyssup: Yes. On 10K created unique plugin configurations, there will be 10K self.buffers. On performance testing the QPS dropped from 6K to 3K. But having 10K unique plugin configuration is not a normal use case.

apisix/plugins/kafka-logger.lua
```diff
@@ -140,7 +142,12 @@ local metadata_schema = {
     properties = {
         log_format = {
             type = "object"
-        }
+        },
+        max_pending_entries = {
+            type = "integer",
+            description = "maximum number of pending entries in the batch processor",
+            minimum = 0,
```
membphis: the minimum should be `1` here

0 is meaningless for APISIX user.

membphis (dismissed):
LGTM, except one minor issue

nic-6443 (approved):

membphis (approved):

nic-chen (approved):

apisix/plugins/kafka-logger.lua
```diff
@@ -255,7 +264,7 @@ function _M.log(conf, ctx)
         entry = log_util.get_log_entry(plugin_name, conf, ctx)
     end
 
-    if batch_processor_manager:add_entry(conf, entry) then
+    if batch_processor_manager:add_entry(conf, entry, max_pending_entries) then
         return
```
bzp2010, line 268: 1. These need to be handled by the batch processor and not some plugin. This is a generic capability.

bzp2010: 2. If all the other batch processor configurations are in `conf`, why add a special configuration and place it in the plugin metadata?

apisix/utils/batch-processor-manager.lua
```diff
@@ -94,11 +113,22 @@ function _M:add_entry(conf, entry)
...
 
-function _M:add_entry_to_new_processor(conf, entry, ctx, func)
+function _M:add_entry_to_new_processor(conf, entry, ctx, func, max_pending_entries)
+    if max_pending_entries then
+        local total_processed_entries_count = total_processed_entries(self)
+        if self.total_pushed_entries - total_processed_entries_count > max_pending_entries then
+            core.log.error("max pending entries limit exceeded. discarding entry.",
+                           " total_pushed_entries: ", self.total_pushed_entries,
+                           " total_processed_entries: ", total_processed_entries_count,
+                           " max_pending_entries: ", max_pending_entries)
```
bzp2010, line 128: Also both achieve their purpose by discarding log entries, how does it differ from the existing control of the maximum buffer achieved by reducing the number of retries?

> Specifically, by reducing `max_retry_count` to `0` or `1` and configuring the `retry_delay` to a shorter time, the maximum amount of time that each log stays in the buffer can be precisely controlled.
> All of this is an existing mechanism, what added value is provided by the additional, new mechanism that will bring more confusion?

This mechanism discards logs before data is added to the buffer, dealing a serious blow to the reliability and meaning of logging. What's the point of a logging mechanism that isn't even remotely explainable if logs can be discarded on the fly and without warning?
A log may be discarded at runtime under a variety of different conditions, and the user will find that a batch of logs critical to his diagnosis of a failure has strangely disappeared from the dataset, an obvious blow to his confidence.

Indeed, this is an optional feature. But in what ways is it advanced compared to existing configurations.

----

If some guys are causing massive memory consumption or even OOM because of all this caching. 

Obviously the first thing that should come under scrutiny is whether its log-receiving system has enough capacity to accept the entirety of the logs, and if it doesn't, that's not APISIX's concern, and it's not our problem. Reliable delivery of diagnostic material at all times is what we should be looking at.

The second step that should be scrutinized is whether these users have sufficient memory resources (both the memory limit of the pod and the entire node) reserved for APISIX operation. What does it matter to us if none of their memory consumption has been properly evaluated and thus pre-configured with unreasonable memory limits, resulting in processes being killed by the kernel?

bzp2010 (changes requested):

apisix/plugins/kafka-logger.lua
bzp2010: This will be done in the future.

bzp2010 (approved):

PR #12347, opened 2025-06-18
ci: add GitHub Action to check CHANGELOG consistency with PRs

LiteSun:
### Description
This PR provides a script to check the changelog for completeness. If a PR is not of type docs, chore, test, or ci, and is not listed in the IGNORE_PRS variable, but is missing from the changelog, the script will raise an error.

Note: This script currently only checks changelog entries for versions after 3.8.0.

#### Test record:
1. Manually removed a changelog entry for version 3.12.0, and the script correctly reported an error.
image
image

2. Tested an upcoming version (e.g., 3.13.0) that hasn’t been tagged yet to verify if the script works correctly. In this case, it compares the current latest code from the main branch with the previous version (3.12.0).
image

### Checklist

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -24,4 +24,5 @@ jobs:
       - name: Run check_changelog_prs script
         working-directory: ci
         run: |
+          export PATH="$HOME/.bun/bin:$PATH"
           bun run check_changelog_prs.ts
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install bun
         run: |
           curl -fsSL https://bun.sh/install | bash
-          export PATH="$HOME/.bun/bin:$PATH"
+          source /home/runner/.bashrc
 
       - name: Run check_changelog_prs script
         working-directory: ci
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -5,6 +5,10 @@ on:
     paths:
       - 'CHANGELOG.md'
       - 'ci/check_changelog_prs.ts'
+  pull_request:
+    paths:
+      - 'CHANGELOG.md'
+      - 'ci/check_changelog_prs.ts'
 
 jobs:
   check-changelog:
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -18,11 +18,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Bun
-        uses: oven-sh/setup-bun@v2
-        with:
-          bun-version: latest
 
       - name: Run check_changelog_prs script
         working-directory: ci
-        run: bun run check_changelog_prs.ts
+        run: |
+          curl -fsSL https://bun.sh/install | bash
+          export PATH="$HOME/.bun/bin:$PATH"
+          bun run check_changelog_prs.ts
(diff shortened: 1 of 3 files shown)


ci/check_changelog_prs.ts
```diff
@@ -0,0 +1,235 @@
...
+
+// Configuration
+const IGNORE_TYPES = [
+    'docs',
+    'chore',
+    'test',
+    'ci'
+];
+
+const IGNORE_PRS = [
```
LiteSun, line 42: refer: https://github.com/apache/apisix/pull/12308#issue-3132286015

ci/check_changelog_prs.ts
SkyeYoung, line 1: In practice, you can try some AST parsing tools to ensure the content you obtain is accurate.

ci/check_changelog_prs.ts
```diff
@@ -0,0 +1,235 @@
...
+        if (hasErrors) {
+            process.exit(1);
+        }
+    } catch (error) {
+        console.error('Error:', error);
+        process.exit(1);
+    }
+}
+
+main();
```
SkyeYoung: `await main()` ?

ci/check_changelog_prs.ts
LiteSun: > In practice, you can try some AST parsing tools to ensure the content you obtain is accurate.

Actually, our changelog currently has a simple and consistent format. The existing code is also straightforward and effective, so there’s no need to use an AST parsing tool for now — it would be overkill. I also prefer not to introduce additional dependencies, as I’d like to keep the script as simple as possible.

ci/check_changelog_prs.ts
LiteSun: done.

SkyeYoung (dismissed):

.github/workflows/check-changelog.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run check_changelog_prs script
+        working-directory: ci
+        run: |
+          curl -fsSL https://bun.sh/install | bash
```
Baoyuantop, line 25: use https://github.com/oven-sh/setup-bun

.github/workflows/check-changelog.yml
```diff
@@ -0,0 +1,24 @@
+name: Check Changelog
+
+on:
+  push:
```
juzhiyuan, line 4: I'm wondering if we should just listen for changes in the CHANGEsLOG file?

juzhiyuan (dismissed):

.github/workflows/check-changelog.yml
SkyeYoung: We can't use these directly.

SkyeYoung: check here https://github.com/apache/infrastructure-actions/blob/main/actions.yml

Baoyuantop: Looks like it ran successfully?

.github/workflows/check-changelog.yml
LiteSun: > I'm wondering if we should just listen for changes in the CHANGEsLOG file?

we actually need to detect any missing changelog updates as early as possible when a PR is created. I just tested it and found that the push event does not trigger the CI when the PR is first submitted.

SkyeYoung (approved):

Baoyuantop (approved):

juzhiyuan (approved):

PR #12353, opened 2025-06-19
feat: replace events library with shdict

Revolyssup:
Breakout from: https://github.com/apache/apisix/pull/12263
Fixes #
Currently nacos discovery uses event library and during testing it was found out that when using lua-resty-events, sometimes not all workers get the events. And inconsistencies emerge. Moreover the idiomatic way to share data between workers is through a shared dict. Therefore this PR migrates nacos from older events mechanism to newer shared dict way. Now only priviliged agent will fetch data from nacos and write to shdict while all workers will read from it.
### Checklist

### DIFF
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -332,14 +332,13 @@ local function fetch_full_registry(premature)
         end
         ::CONTINUE::
     end
-    local old_curr_service_in_use = curr_service_in_use
-    curr_service_in_use = service_names
     -- remove services that are not in use anymore
-    for key, _ in pairs(old_curr_service_in_use) do
+    for key, _ in pairs(curr_service_in_use) do
         if not service_names[key] then
             nacos_dict:delete(key)
         end
     end
+    curr_service_in_use = service_names
 end
 
 
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -20,6 +20,7 @@ local local_conf         = require('apisix.core.config_local').local_conf()
 local http               = require('resty.http')
 local core               = require('apisix.core')
 local ipairs             = ipairs
+local pairs              = pairs
 local type               = type
 local math               = math
 local math_random        = math.random
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -35,6 +35,10 @@ local log                = core.log
 
 local default_weight
 local nacos_dict = ngx.shared.nacos --key: namespace_id.group_name.service_name
+if not nacos_dict then
+    error("lua_shared_dict \"nacos\" not configured")
+end
+
 local auth_path = 'auth/login'
 local instance_list_path = 'ns/instance/list?healthyOnly=true&serviceName='
 local default_namespace_id = "public"
--- a/apisix/discovery/nacos/init.lua
+++ b/apisix/discovery/nacos/init.lua
@@ -34,7 +34,7 @@ local str_find           = core.string.find
 local log                = core.log
 
 local default_weight
-local applications
+local nacos_dict = ngx.shared.nacos --key: namespace_id.group_name.service_name
 local auth_path = 'auth/login'
 local instance_list_path = 'ns/instance/list?healthyOnly=true&serviceName='
 local default_namespace_id = "public"
@@ -42,19 +42,12 @@ local default_group_name = "DEFAULT_GROUP"
 local access_key
 local secret_key
 
-local events
-local events_list
-
 
 local _M = {}
 
-local function discovery_nacos_callback(data, event, source, pid)
-    applications = data
-    log.notice("update local variable application, event is: ", event,
-               "source: ", source, "server pid:", pid,
-               ", application: ", core.json.encode(applications, true))
+local function get_key(namespace_id, group_name, service_name)
+    return namespace_id .. '.' .. group_name .. '.' .. service_name
 end
-
 
 local function request(request_uri, path, body, method, basic_auth)
     local url = request_uri .. path
@@ -289,15 +282,11 @@ local function fetch_full_registry(premature)
     local token_param, err = get_token_param(base_uri, username, password)
     if err then
         log.error('get_token_param error:', err)
-        if not applications then
-            applications = up_apps
-        end
         return
     end
 
     local infos = get_nacos_services()
     if #infos == 0 then
-        applications = up_apps
         return
     end
 
@@ -318,29 +307,14 @@ local function fetch_full_registry(premature)
             goto CONTINUE
         end
 
-        if not up_apps[namespace_id] then
-            up_apps[namespace_id] = {}
-        end
-
-        if not up_apps[namespace_id][group_name] then
-            up_apps[namespace_id][group_name] = {}
-        end
-
+        local nodes = {}
+        local key = get_key(namespace_id, group_name, service_info.service_name)
         for _, host in ipairs(data.hosts) do
-            local nodes = up_apps[namespace_id]
-                [group_name][service_info.service_name]
-            if not nodes then
-                nodes = {}
-                up_apps[namespace_id]
-                    [group_name][service_info.service_name] = nodes
-            end
-
             local node = {
                 host = host.ip,
                 port = host.port,
                 weight = host.weight or default_weight,
             }
-
             -- docs: https://github.com/yidongnan/grpc-spring-boot-starter/pull/496
             if is_grpc(scheme) and host.metadata and host.metadata.gRPC_port then
                 node.port = host.metadata.gRPC_port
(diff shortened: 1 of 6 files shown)


apisix/discovery/nacos/init.lua
```diff
@@ -417,7 +359,20 @@ end
 
 
 function _M.dump_data()
-    return {config = local_conf.discovery.nacos, services = applications or {}}
```
Revolyssup, line 420: config is removed from here as it's a security risk.

t/discovery/nacos2.t
```diff
@@ -148,18 +148,43 @@ routes:
...
-]
---- no_error_log
-[error, error]
+--- config
+    location /t {
+        content_by_lua_block {
+            local http = require("resty.http")
+            local uri = "http://127.0.0.1:" .. ngx.var.server_port
+
+            -- Wait for 2 seconds for APISIX initialization
```
Revolyssup, line 157: Test is modified to add this sleep because: Since retry mechanism is removed, priviliged agent fills data in shdict might take some time so we want to send the request after that.

apisix/discovery/nacos/init.lua
```diff
@@ -34,28 +34,21 @@ local str_find           = core.string.find
 local log                = core.log
 
 local default_weight
-local applications
+local nacos_dict = ngx.shared.nacos --key: namespace_id.group_name.service_name
```
membphis, line 38: check if `nacos_dict` is nil

Revolyssup: done

apisix/discovery/nacos/init.lua
nic-6443: make sense.

apisix/discovery/nacos/init.lua
```diff
@@ -417,7 +363,20 @@ end
...
 
 function _M.dump_data()
-    return {config = local_conf.discovery.nacos, services = applications or {}}
+    local keys = nacos_dict:get_keys(0)
+    local applications = {}
+    for _, key in ipairs(keys) do
+        local value = nacos_dict:get(key)
+        if value then
+            local nodes = core.json.decode(value)
+            if nodes and #nodes > 0 then
```
nic-6443: `dump_data` is for debug purpose, if value in shared dict for this key is empty , we should return the real empty value for debugging instead hide it.

nic-6443 (commented):
one thing you missed: remove unused key from shared dict in every `fetch_full_registry` run

apisix/discovery/nacos/init.lua
```diff
@@ -318,50 +310,35 @@ local function fetch_full_registry(premature)
...
-        log.error("post_event failure with ", events_list._source,
-                  ", update application error: ", err)
+    local old_curr_service_in_use = curr_service_in_use
+    curr_service_in_use = service_names
+    -- remove services that are not in use anymore
+    for key, _ in pairs(old_curr_service_in_use) do
+        if not service_names[key] then
+            nacos_dict:delete(key)
+        end
     end
```
nic-6443:
```suggestion
    -- remove services that are not in use anymore
    for key, _ in pairs(curr_service_in_use) do
        if not service_names[key] then
            nacos_dict:delete(key)
        end
    end
    curr_service_in_use = service_names 
```

nic-6443 (approved):

apisix/discovery/nacos/init.lua
```diff
@@ -371,40 +348,18 @@ function _M.nodes(service_name, discovery_args)
...
-        end
-        ngx.sleep(step)
-        waiting_time = waiting_time - step
-    end
-
-    if not applications or not applications[namespace_id]
-        or not applications[namespace_id][group_name]
-    then
+    local key = get_key(namespace_id, group_name, service_name)
+    local value = nacos_dict:get(key)
```
bzp2010, line 352: You're using `shdict:get` here to read a shared memory, and I'd like to point out that it locks the shdict even when it reads the shared memory.

> ref: https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_shdict.c#L1593-L1609

Each request will use this nodes function, which is obviously a very "hot" function, which will limit the concurrency that can be handled by waiting for locks, is this appropriate?

Revolyssup:
@bzp2010 I ran a benchmark to test whether the `shdict:get` brings too much performance degradation. But based on below results I don't think this should be an issue.

## Configuration used
config.yaml
```yaml
apisix:
  node_listen: 1984
deployment:
  role: data_plane
  role_data_plane:
    config_provider: yaml
nginx_config:
  worker_processes: 8
discovery:
  nacos:
      host:
        - "http://nacos:nacos@127.0.0.1:8848"
      prefix: "/nacos/v1/"
      fetch_interval: 1
      weight: 1
      timeout:
        connect: 2000
        send: 2000
        read: 5000
```

apisix.yaml
```yaml
routes:
  -
    uri: /hello
    upstream:
      service_name: APISIX-NACOS
      discovery_type: nacos
      type: roundrobin
#END
```
Here are the results.

### On master
```bash
❯ wrk -c 10 -t 5 -d 30s -R 50000 http://localhost:1984/hello
Running 30s test @ http://localhost:1984/hello
  5 threads and 10 connections
  Thread calibration: mean lat.: 2999.418ms, rate sampling interval: 10395ms
  Thread calibration: mean lat.: 2885.546ms, rate sampling interval: 10125ms
  Thread calibration: mean lat.: 2857.330ms, rate sampling interval: 10059ms
  Thread calibration: mean lat.: 2967.000ms, rate sampling interval: 10231ms
  Thread calibration: mean lat.: 2953.790ms, rate sampling interval: 10256ms
    Latency    11.12s     3.06s   16.47s    57.49%
    Req/Sec     4.52k    82.05     4.63k    60.00%
  689405 requests in 30.00s, 110.45MB read
Requests/sec:  22980.47
Transfer/sec:      3.68MB

```

### After changes
```bash
❯ wrk -c 10 -t 5 -d 30s -R 50000 http://localhost:1984/hello
Running 30s test @ http://localhost:1984/hello
  5 threads and 10 connections
  Thread calibration: mean lat.: 3060.053ms, rate sampling interval: 10584ms
  Thread calibration: mean lat.: 3084.401ms, rate sampling interval: 10772ms
  Thread calibration: mean lat.: 3254.041ms, rate sampling interval: 11198ms
  Thread calibration: mean lat.: 3093.539ms, rate sampling interval: 10985ms
  Thread calibration: mean lat.: 3062.424ms, rate sampling interval: 10772ms
    Latency    11.89s     3.36s   17.74s    56.68%
    Req/Sec     4.13k    39.11     4.18k    60.00%
  623009 requests in 30.00s, 99.81MB read
Requests/sec:  20767.18
Transfer/sec:      3.33MB
```

## Result
The QPS dropped from 22980.47 to 20767.18 with ~2k difference. I think this should be acceptable given the issue with events library this PR solves. According to me, it's an okay tradeoff

apisix/discovery/nacos/init.lua
bzp2010: When we encounter performance issues, consider adding caching with TTL at the worker level to further improve locality.

bzp2010 (approved):

membphis (approved):
LGTM

imsuperli:
> Breakout from: #12263 Fixes # Currently nacos discovery uses event library and during testing it was found out that when using lua-resty-events, sometimes not all workers get the events. And inconsistencies emerge. Moreover the idiomatic way to share data between workers is through a shared dict. Therefore this PR migrates nacos from older events mechanism to newer shared dict way. Now only priviliged agent will fetch data from nacos and write to shdict while all workers will read from it.
> 
> ### Checklist
> * [ ]  I have explained the need for this PR and the problem it solves
> * [ ]  I have explained the changes or the new features added to this PR
> * [ ]  I have added tests corresponding to this change
> * [ ]  I have updated the documentation to reflect this change
> * [ ]  I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)



How to understand:   Now only priviliged agent will fetch data from nacos and write to shdict while all workers will read from it ？

In the current code, does each worker periodically fetch registration information from Nacos?

9268:
> Currently nacos discovery uses event library and during testing it was found out that when using lua-resty-events, sometimes not all workers get the events. And inconsistencies emerge. Moreover the idiomatic way to share data between workers is through a shared dict. Therefore this PR migrates nacos from older events mechanism to newer shared dict way. Now only priviliged agent will fetch data from nacos and write to shdict while all workers will read from it.

@Revolyssup I'm upgrade apisix from 3.7 to 3.14.1，my nacos server get 40x requests than original apisxi（worker=40），it seems every worker is fetching data from nacos，i checked `apisix/discovery/nacos/init.lua` and not found restriction applied on timer running on special worker， the concurrent of nacos reqs is unacceptable，is this a feature or a bug？

Baoyuantop:
Hi @9268, thank you for your report. After reviewing the code, I found that the Nacos service discovery is currently handling data requests from every worker, which needs to be changed so that only the privileged agent can retrieve the data. Please feel free to submit an issue to document this problem or submit PR modifications.

9268:
> Hi @9268, thank you for your report. After reviewing the code, I found that the Nacos service discovery is currently handling data requests from every worker, which needs to be changed so that only the privileged agent can retrieve the data. Please feel free to submit an issue to document this problem or submit PR modifications.

@Baoyuantop hi，i have commit a pr to fix this issue，pls have a review https://github.com/apache/apisix/pull/12867

PR #12358, opened 2025-06-20
fix: assign new_nodes of service discovery to upstream every time

nic-6443:
### Description

I introduced a performance optimization in #12258, and after testing, we found that the kubernetes performance optimization would invalid after some time. 
The reason is that after watch progress is done in kubernetes service discovery for some time, it will re-list all data. At this point, addresses of all tables that store service discovery are changed, but the content of nodes does not change. This causes `up_conf.nodes` to never reference the `new_nodes` table at https://github.com/apache/apisix/blob/9e5163c60b74589672b10201355b855098166f4a/apisix/upstream.lua#L305-L317 because their addresses are different and it goes to the branch where equality is determined by value in every downstream requests.
This branch has significant performance overhead: https://github.com/apache/apisix/blob/9e5163c60b74589672b10201355b855098166f4a/apisix/utils/upstream.lua#L50-L61

### Checklist

### DIFF
--- a/apisix/utils/upstream.lua
+++ b/apisix/utils/upstream.lua
@@ -47,8 +47,8 @@ local function compare_upstream_node(up_conf, new_t)
 
     -- slow path
     core.log.debug("compare upstream nodes by value, ",
-                    "old: ", tostring(old_t) , " ", core.json.delay_encode(old_t, true),
-                    ", new: ", tostring(new_t) , " ", core.json.delay_encode(new_t, true))
+                    "old: ", tostring(old_t) , " ", core.json.delay_encode(old_t, true))
+    core.log.debug("new: ", tostring(new_t) , " ", core.json.delay_encode(new_t, true))
 
     if up_conf.original_nodes then
         -- if original_nodes is set, it means that the upstream nodes


nic-chen (dismissed):

membphis (approved):

AlinsRan (approved):

shreemaan-abhishek (dismissed):

AlinsRan (approved):

apisix/utils/upstream.lua
```diff
@@ -43,6 +45,18 @@ local function compare_upstream_node(up_conf, new_t)
         return false
     end
 
+    -- slow path
+    core.log.debug("compare upstream nodes by value, ",
+                    "old: ", tostring(old_t) , " ", core.json.delay_encode(old_t, true),
+                    ", new: ", tostring(new_t) , " ", core.json.delay_encode(new_t, true))
```
nic-chen: Call `delay_encode` once for each log record, otherwise the content of the log records will be garbled.

```suggestion
    core.log.debug("compare upstream nodes by value, ",
                    "old: ", tostring(old_t) , " ", core.json.delay_encode(old_t, true))
   core.log.debug(" new: ", tostring(new_t) , " ", core.json.delay_encode(new_t, true))
```

nic-6443: fixed

nic-chen (approved):

PR #12361, opened 2025-06-23
refactor: change log level to debug to avoid unnecessary logs

Revolyssup:
Fixes #
For each request, the following log is printed multiple times and the logs get flooded with it. The primary purpose for this log was for testing so it should be set to debug.
### Checklist


membphis:
you need to update your test case

change the log level in test case `t/core/ctx3.t`

bzp2010 (approved):

PR #12364, opened 2025-06-23
fix: support multiple versions of Elasticsearch logger

Revolyssup:
Fixes 
Error comes when using elasticsearch version 9+
Screenshot From 2025-06-23 12-29-02

We want to remove the compat header and have a general support for all major versions.

For elasticsearch logger plugin comptability check. Following are the observations and conclusion based on reading docs and self testing.

The `type` field issue:
Before version 7, the `type` field was required in json schema https://github.com/elastic/elasticsearch/blob/8453f7701de25c467ce08088ddddac185cb8028c/rest-api-spec/src/main/resources/rest-api-spec/api/create.json#L21.
The current tests without `type` fail on version 6(here it's 6.7) with the following error

Batch Processor[elasticsearch-logger] failed to process entries: elasticsearch server returned status: 400, body: {"error":{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: type is missing;"}]

Adding a _type in the code fixes the above test.
Conclusion
As suggested in the migration guide, we can hardcode the `type` to "_doc" in versions prior to 7 and on 7+ we can omit `type` completely
Indices created in 6.x only allow a single-type per index. Any name can be used for the type, but there can be only one. The preferred type name is _doc, so that index APIs have the same path as they will have in 7.0: PUT {index}/_doc/{id} and POST {index}/_doc

The `id` field issue:
Another problem was detected while running tests on older version(6), when _id was not passed along with _index,. the following error was reported:

Batch Processor[elasticsearch-logger] failed to process entries: elasticsearch server returned status: 400, body: {"error":{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: an id must be provided if version type or value are set;"}],"

In older version, _id is also required https://github.com/elastic/elasticsearch/blob/8453f7701de25c467ce08088ddddac185cb8028c/rest-api-spec/src/main/resources/rest-api-spec/api/create.json#L9
But I when I used index API instead of create API, request passed even without _id for all versions 6,7,8

-            create = {
+            index = {
                 _index = conf.field.index,
-                _type = conf.field.type
+                _type = "_doc",
             }
         }) .. "\n" ..

The doc says
A create action fails if a document with the same ID already exists in the target An index action adds or replaces a document as necessary.
An answer in the discuss forum says
Neither is used to create an index (although they may do so as a side-effect of inserting the first document into a new index). "index" indexes a document. "create" only indexes a document if it doesn't already exist. It used to be that "create" was faster if you knew that your docs didn't already exist. But now, since versioning was added, "index" performs as well as "create"
Another answer suggests
you use create if a doc with that id already exists, if you use index it will create a document or update it if xisting.
Conclusion: To me it looks okay to replace create with index

Conclusion:
We can remove `type` from plugin configuration and for detected version below 7, hardcode it as `_doc`
We can replace Create with Index in the request body.
We run all tests on version 8 and my PR adds a comptability test with version 9. We can add two more compatibilty tests with 6 and 7 to confirm our changes.

Result: We will be able to support ES versions 6,7,8,9

References and other relevant information:
https://www.elastic.co/guide/en/elasticsearch/reference/6.7/release-notes-6.7.0.html. Search "Deprecate types" in this doc.
From reading this issue: https://github.com/elastic/elasticsearch/issues/35190
Though a temporary parameter was added `include_type_name` which defaults to true between 6.7 upto 7 meaning by default `type` was being sent but it could be turned off.
Then between 7 and 8 this field was default to false meaning by default the `type` will no longer be sent and a deprecated warning will come.
Then after 8+ this include_type_name was removed and `type` was no longer supported.
References:
https://www.elastic.co/docs/reference/elasticsearch/rest-apis/compatibility
https://www.elastic.co/guide/en/elasticsearch/reference/8.18/rest-api-compatibility.html
https://github.com/elastic/elasticsearch-ruby/issues/2665
### Checklist

### DIFF
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -128,19 +128,7 @@ function _M.check_schema(conf, schema_type)
     local check = {"endpoint_addrs"}
     core.utils.check_https(check, conf, plugin_name)
     core.utils.check_tls_bool({"ssl_verify"}, conf, plugin_name)
-
     return core.schema.check(schema, conf)
-end
-
-
-local function get_logger_entry(conf, ctx)
-    local entry = log_util.get_log_entry(plugin_name, conf, ctx)
-    return core.json.encode({
-            create = {
-                _index = conf.field.index
-            }
-        }) .. "\n" ..
-        core.json.encode(entry) .. "\n"
 end
 
 
@@ -186,24 +174,50 @@ local function get_es_major_version(uri, conf)
 end
 
 
-local function send_to_elasticsearch(conf, entries)
-    local httpc, err = http.new()
-    if not httpc then
-        return false, str_format("create http error: %s", err)
+local function get_logger_entry(conf, ctx)
+    local entry = log_util.get_log_entry(plugin_name, conf, ctx)
+    local body = {
+        index = {
+            _index = conf.field.index
+        }
+    }
+    -- for older version type is required
+    core.log.info("elasticsearch version:", conf._version,"]", " and type is")
+    if conf._version == "6" or conf._version == "5" then
+        body.index._type = "_doc"
     end
+    return core.json.encode(body) .. "\n" ..
+        core.json.encode(entry) .. "\n"
+end
 
-    local selected_endpoint_addr
-    if conf.endpoint_addr then
-        selected_endpoint_addr = conf.endpoint_addr
-    else
-        selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
-    end
+local function set_version(conf)
     if not conf._version then
+        local selected_endpoint_addr
+        if conf.endpoint_addr then
+            selected_endpoint_addr = conf.endpoint_addr
+        else
+            selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
+        end
         local major_version, err = get_es_major_version(selected_endpoint_addr, conf)
         if err then
             return false, str_format("failed to get Elasticsearch version: %s", err)
         end
         conf._version = major_version
+    end
+end
+
+
+local function send_to_elasticsearch(conf, entries)
+    local httpc, err = http.new()
+    if not httpc then
+        return false, str_format("create http error: %s", err)
+    end
+    set_version(conf)
+    local selected_endpoint_addr
+    if conf.endpoint_addr then
+        selected_endpoint_addr = conf.endpoint_addr
+    else
+        selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
     end
     local uri = selected_endpoint_addr .. "/_bulk"
     local body = core.table.concat(entries, "")
@@ -251,6 +265,11 @@ function _M.body_filter(conf, ctx)
     log_util.collect_body(conf, ctx)
 end
 
+function _M.access(conf)
+    -- set_version will call ES server only the first time
+    -- so this should not amount to considerable overhead
+    set_version(conf)
+end
 
 function _M.log(conf, ctx)
     local entry = get_logger_entry(conf, ctx)
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -187,20 +187,22 @@ local function get_logger_entry(conf, ctx)
         core.json.encode(entry) .. "\n"
 end
 
-local function set_version(conf)
-    if not conf._version then
-        local selected_endpoint_addr
-        if conf.endpoint_addr then
-            selected_endpoint_addr = conf.endpoint_addr
-        else
-            selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
-        end
-        local major_version, err = get_es_major_version(selected_endpoint_addr, conf)
-        if err then
-            return false, str_format("failed to get Elasticsearch version: %s", err)
-        end
-        conf._version = major_version
+local function fetch_and_update_es_version(conf)
+    if conf._version then
+        return
     end
+    local selected_endpoint_addr
+    if conf.endpoint_addr then
+        selected_endpoint_addr = conf.endpoint_addr
+    else
+        selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
+    end
+    local major_version, err = get_es_major_version(selected_endpoint_addr, conf)
+    if err then
+        core.log.error("failed to get Elasticsearch version: ", err)
+        return
+    end
+    conf._version = major_version
 end
 
 
@@ -209,7 +211,7 @@ local function send_to_elasticsearch(conf, entries)
     if not httpc then
         return false, str_format("create http error: %s", err)
     end
-    set_version(conf)
+    fetch_and_update_es_version(conf)
     local selected_endpoint_addr
     if conf.endpoint_addr then
         selected_endpoint_addr = conf.endpoint_addr
@@ -256,9 +258,9 @@ function _M.body_filter(conf, ctx)
 end
 
 function _M.access(conf)
-    -- set_version will call ES server only the first time
+    -- fetch_and_update_es_version will call ES server only the first time
     -- so this should not amount to considerable overhead
-    set_version(conf)
+    fetch_and_update_es_version(conf)
 end
 
 function _M.log(conf, ctx)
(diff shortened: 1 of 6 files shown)


apisix/plugins/elasticsearch-logger.lua
```diff
@@ -156,12 +204,30 @@ local function send_to_elasticsearch(conf, entries)
...
+        ["Accept"] = "application/vnd.elasticsearch+json"
     }
+    if conf._version == "8" then
+        headers["Content-Type"] = headers["Content-Type"] .. compat_header_7
+        headers["Accept"] = headers["Accept"] .. compat_header_7
+    elseif conf._version == "9" then
+        headers["Content-Type"] = headers["Content-Type"] .. compat_header_8
+        headers["Accept"] = headers["Accept"] .. compat_header_8
+        if conf.field.type then
+            core.log.warn("type is not supported in Elasticsearch 9, removing `type`")
```
membphis: I think we should use error log level

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -195,6 +264,11 @@ function _M.body_filter(conf, ctx)
     log_util.collect_body(conf, ctx)
 end
 
+function _M.access(conf)
```
Revolyssup, line 260: set_version calls ES server the first time and this http connection is not allowed in log phase so I had no other way but to add this here because version information is needed at the start of log phase when entries are being created.

ci/pod/docker-compose.plugin.yml
```diff
@@ -225,6 +225,66 @@ services:
       http.port: 9201
       xpack.security.enabled: 'true'
 
+  elasticsearch-noauth-2:
```
nic-6443: Running so many Elasticsearch containers simultaneously in CI may strain computing resources, and causing tests to fail.

nic-6443: Has the auth method changed between these versions? If not, is it unnecessary to test the auth case separately for each version?

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -124,32 +124,94 @@ function _M.check_schema(conf, schema_type)
...
+end
+
 
 local function send_to_elasticsearch(conf, entries)
     local httpc, err = http.new()
     if not httpc then
         return false, str_format("create http error: %s", err)
     end
-
+    set_version(conf)
```
nic-6443: we already call `set_version` here, why still need it in `access` phase?

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -159,9 +221,16 @@ local function send_to_elasticsearch(conf, entries)
     local uri = selected_endpoint_addr .. "/_bulk"
     local body = core.table.concat(entries, "")
     local headers = {
-        ["Content-Type"] = "application/x-ndjson;compatible-with=7",
-        ["Accept"] = "application/vnd.elasticsearch+json;compatible-with=7"
+        ["Content-Type"] = "application/x-ndjson",
+        ["Accept"] = "application/vnd.elasticsearch+json"
     }
+    if conf._version == "8" then
+        headers["Content-Type"] = headers["Content-Type"] .. compat_header_7
```
membphis: can we remove `compat_header_7` and `compat_header_8`?

If yes, then we can move `if conf._version == "8" then` and `elseif conf._version == "9" then`.

just use the default header, pls confirm if it is ok:

```
local headers = {
        ["Content-Type"] = "application/x-ndjson",
        ["Accept"] = "application/vnd.elasticsearch+json"
}
```

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -195,6 +264,11 @@ function _M.body_filter(conf, ctx)
     log_util.collect_body(conf, ctx)
 end
 
+function _M.access(conf)
+    -- set_version will call ES server only the first time
+    -- so this should not amount to considerable overhead
+    set_version(conf)
```
membphis: how about `fetch_and_update_es_version`?

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -124,32 +124,94 @@ function _M.check_schema(conf, schema_type)
...
+    }
+    -- for older version type is required
+    if conf._version == "6" or conf._version == "5" then
+        body.index._type = "_doc"
+    end
+    return core.json.encode(body) .. "\n" ..
         core.json.encode(entry) .. "\n"
 end
 
+local function set_version(conf)
```
membphis:
```lua
local function set_version(conf)
    if conf._version then
        return
    end

        local selected_endpoint_addr
        if conf.endpoint_addr then
            selected_endpoint_addr = conf.endpoint_addr
        else
            selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
        end
        local major_version, err = get_es_major_version(selected_endpoint_addr, conf)
        if err then
            return false, str_format("failed to get Elasticsearch version: %s", err)
        end
        conf._version = major_version
end

File: apisix/plugins/elasticsearch-logger.lua [hunk 8]
```diff
@@ -124,32 +124,94 @@ function _M.check_schema(conf, schema_type)
...
+    if not conf._version then
+        local selected_endpoint_addr
+        if conf.endpoint_addr then
+            selected_endpoint_addr = conf.endpoint_addr
+        else
+            selected_endpoint_addr = conf.endpoint_addrs[math_random(#conf.endpoint_addrs)]
+        end
+        local major_version, err = get_es_major_version(selected_endpoint_addr, conf)
+        if err then
+            return false, str_format("failed to get Elasticsearch version: %s", err)
```
membphis: we should write error log here

we do not capture the error msg in `access` phase

apisix/plugins/elasticsearch-logger.lua
membphis: I do not like `if conf._version == "8" then ... elseif conf._version == "9" then ... end` style

ci/pod/docker-compose.plugin.yml
```diff
@@ -225,6 +225,66 @@ services:
...
+    environment:
+      ES_JAVA_OPTS: -Xms512m -Xmx512m
+      discovery.type: single-node
+      ELASTIC_USERNAME: elastic
+      ELASTIC_PASSWORD: 123456
+      http.port: 9201
+      xpack.security.enabled: 'true'
+
+  elasticsearch-auth-4:
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.7.0
```
membphis, line 255: we started more instances of es

the github action, is it power enough to support this? I am little worried

apisix/plugins/elasticsearch-logger.lua
Revolyssup: this function is run later in batch processor. But we need the version in the log phase when entry is created. That is why we need it in access phase.

ci/pod/docker-compose.plugin.yml
Revolyssup: i removed two of these containers(not needed)

apisix/plugins/elasticsearch-logger.lua
Revolyssup: removed it. not needed anymore

ci/pod/docker-compose.plugin.yml
Revolyssup: i removed two of these containers(not needed)

apisix/plugins/elasticsearch-logger.lua
Revolyssup: done

apisix/plugins/elasticsearch-logger.lua
Revolyssup: done

Revolyssup:
> Do we need to remove `type` field in this PR?

Yes. If we don't remove `type` field then we will need the compat headers that I removed

membphis:
> > Do we need to remove `type` field in this PR?
> 
> Yes. If we don't remove `type` field then we will need the compat headers that I removed

pls update the doc too

Revolyssup:
> > > Do we need to remove `type` field in this PR?
> > 
> > 
> > Yes. If we don't remove `type` field then we will need the compat headers that I removed
> 
> pls update the doc too

Removed `type` from docs

membphis (approved):
LGTM

nic-6443 (dismissed):

nic-6443 (approved):

PR #12366, opened 2025-06-23
fix: added restriction for TLSv1.3 cross-SNI session resumption

nic-6443:

Fixes #

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -300,6 +300,24 @@ local function set_upstream_headers(api_ctx, picked_server)
 end
 
 
+-- verify the TLS session resumption by checking if the SNI in the client hello
+-- matches the hostname of the SSL session, this is to prevent the mTLS bypass security issue.
+local function verify_tls_session_resumption()
+    local session_hostname, err = apisix_ssl.session_hostname()
+    if err then
+        core.log.error("failed to get session hostname: ", err)
+        return false
+    end
+    if session_hostname and session_hostname ~= ngx.ctx.client_hello_sni then
+        core.log.error("sni in client hello mismatch hostname of ssl session, ",
+                         "sni: ", ngx.ctx.client_hello_sni, ", hostname: ", session_hostname)
+        return false
+    end
+
+    return true
+end
+
+
 local function verify_tls_client(ctx)
     local matched = router.router_ssl.match_and_set(ctx, true)
     if not matched then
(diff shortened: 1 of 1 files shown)


nic-chen (dismissed):

membphis (dismissed):

AlinsRan (dismissed):

apisix/init.lua
```diff
@@ -391,6 +400,17 @@ local function verify_https_client(ctx)
...
+        local session_hostname, err = apisix_ssl.session_hostname()
+        if err then
+            core.log.error("failed to get session hostname: ", err)
+            return false
+        end
+        if session_hostname and session_hostname ~= ngx.ctx.client_hello_sni then
+            core.log.error("sni in client hello mismatch hostname of ssl session, ",
+                             "sni: ", ngx.ctx.client_hello_sni, ", hostname: ", session_hostname)
+            return false
+        end
```
bzp2010: What is the difference between this and the other modification that looks the same above?

bzp2010: Maybe they could be moved to a new function with some comments explaining why this verification is done "to fix a security issue".

nic-6443: The above is the http subsystem, and this is the stream subsystem.

nic-6443: The HTTP subsystem also has matching verification for the HTTP host and SNI, so the verification logic of these two subsystems is not exactly the same and cannot be fully reused.

nic-6443: > Maybe they could be moved to a new function with some comments explaining why this verification is done "to fix a security issue".

done

membphis (approved):

apisix/init.lua
```diff
@@ -1022,25 +1046,6 @@ function _M.http_control()
...
-    core.tablepool.release("api_ctx", api_ctx)
-    ngx_ctx.api_ctx = nil
-
-    if not ok then
-        if err then
-            core.log.error("failed to fetch ssl config: ", err)
-        end
-        ngx_exit(-1)
-    end
-end
```
bzp2010, line 1041: The stream ssl phase is merged into a code snippet that is consistent with http. Is this as expected?

nic-6443: Yes, previously the stream subsystem did not use the `client_hello_by_lua` phase. But in this fix, it is necessary to use this phase. 
After comparing the code of the http and stream subsystems, I confirmed that the code for both `client_hello_by_lua` and `certificate_by_lua` phases can be reused in these two subsystems, making it unnecessary to maintain a separate `stream_ssl_phase`.

PR #12375, opened 2025-06-24
fix:  deployment data plane with read-only etcd

SkyeYoung:
### Description

**What**

This PR implements skipping write operations to etcd when the `deployment.role` is `data_plane`.

**How**

Make a judgment in `etcd.init()`.

**Why**

Since `init_etcd` is called directly in https://github.com/apache/apisix-docker/blob/release/apisix-3.2.2/debian-dev/docker-entrypoint.sh#L39-L40, it does not make judgments like in https://github.com/apache/apisix/blob/3.2.2/apisix/cli/ops.lua#L821-L823. Causes apisix running as a container, there will be a problem when `deployment.role` is `data_plane`.

From a maintenance perspective, I also want to avoid adding things in places like `docker-entrypoint.sh`, as this would obviously duplicate more code. As for testing, there's currently no direct testing of the APISIX container, and it's difficult for me to do this in a short amount of time.

After this modification, it can currently avoid similar issues when calling `init_etcd` or `etcd.init()` anywhere.

In the future, we can actually remove the redundant checks in https://github.com/apache/apisix/blob/3.2.2/apisix/cli/ops.lua#L821-L823.

Fixes #

### Checklist

### DIFF
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -379,10 +379,14 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    -- The data plane does not have write permissions,
+    -- and it should not continue to prepare dirs.
     if yaml_conf.deployment.role == "data_plane" then
+        print("data plane does not have write permissions, skip preparing dirs")
         return true
     end
 
+    print("prepare dirs")
     local etcd_ok = false
     for index, host in ipairs(etcd_healthy_hosts) do
         if prepare_dirs(yaml_conf, args, index, host, host_count) then
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -381,7 +381,8 @@ function _M.init(env, args)
 
     -- access from the data plane to etcd must be read-only
     if yaml_conf.deployment.role == "data_plane" then
-        print("access from the data plane to etcd must be read-only, skip initializing the data of etcd")
+        print("access from the data plane to etcd must be read-only, "
+              .."skip initializing the data of etcd")
         return true
     end
 
--- a/apisix/cli/etcd.lua
+++ b/apisix/cli/etcd.lua
@@ -379,6 +379,10 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    if yaml_conf.deployment.role == "data_plane" then
+        return true
+    end
+
     local etcd_ok = false
     for index, host in ipairs(etcd_healthy_hosts) do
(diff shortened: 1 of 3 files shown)


apisix/cli/etcd.lua
```diff
@@ -379,6 +379,10 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    if yaml_conf.deployment.role == "data_plane" then
```
SkyeYoung, line 384: Because yaml_conf has already been parsed above and there are similar checks already in place, I've chosen to put it here instead of adding it to the `init_etcd` function.

membphis: @SkyeYoung add some comments, to tell other delevelopers about why we `return true` directly

membphis: and we can write some error log(info or debug level)
then it can be verified in your test case

SkyeYoung: done

SkyeYoung: I found that there were no examples of writing logs in the cli, they were all printed directly, so I also used `print`.

membphis (commented):
the test case is very good

The comment is wrong. For the `data_plane` role, this means that `etcd` should avoid to write. But APISIX allows the user to set a writable etcd account.

apisix/cli/etcd.lua
membphis: you are write, we should use `print`

apisix/cli/etcd.lua
```diff
@@ -379,6 +379,14 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    -- The data plane does not have write permissions,
```
moonming: Need clearer comments on what component permissions to write and why they cannot be written

apisix/cli/etcd.lua
```diff
@@ -379,6 +379,16 @@ function _M.init(env, args)
         util.die("the etcd cluster needs at least 50% and above healthy nodes\n")
     end
 
+    -- access from the data plane to etcd should be read-only.
+    -- data plane writes to etcd may cause security issues,
+    -- or multiple instance writes may cause performance degradation.
```
membphis: remove this line, it not a key point

membphis (dismissed):

apisix/cli/etcd.lua
SkyeYoung:
```suggestion
    -- data plane writes to etcd may cause security issues.
```

SkyeYoung: done. pls approve again.

moonming (approved):

PR #12380, opened 2025-06-26
fix: `get_keys` only return first 1024 items in shared dict by default

nic-6443:
### Description

From OpenResty Doc:
> By default, only the first 1024 keys (if any) are returned. When the <max_count> argument is given the value 0, then all the keys will be returned even there is more than 1024 keys in the dictionary.

We should return complete data in the `/v1/discovery/{service}/dump` control API, otherwise, it will confuse users.

Fixes #

### Checklist

### DIFF
--- a/apisix/discovery/kubernetes/init.lua
+++ b/apisix/discovery/kubernetes/init.lua
@@ -634,7 +634,7 @@ end
 
 
 local function dump_endpoints_from_dict(endpoint_dict)
-    local keys, err = endpoint_dict:get_keys()
+    local keys, err = endpoint_dict:get_keys(0)
     if err then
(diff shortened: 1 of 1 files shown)


Baoyuantop:
I found a similar call https://github.com/apache/apisix/blob/master/apisix/init.lua#L907, do we need to handle it here?

moonming (approved):

nic-6443:
> I found a similar call https://github.com/apache/apisix/blob/master/apisix/init.lua#L907, do we need to handle it here?

@Baoyuantop The data stored in this place is the state of each worker. I think there will not be more than 1024 workers for any apisix instance, so it does not need to be changed.

PR #12383, opened 2025-06-26
feat: decoupled prometheus exporter's calculation and output

SkyeYoung:
### Description

This PR decouples the calculation and output processes of the Prometheus exporter. The "calculation" is performed in the privileged agent process at intervals defined by the `refresh_interval`(default is 15s) and written to a shared dict, while the "output" (i.e., the `/apisix/prometheus/metrics` API) is moved to the worker process, which only reads and returns the cached data in the shared dict.

The above are just the core changes. In fact, I encountered many other problems, which have been commented or annotated in the corresponding positions, and will not be repeated here.

For the testing part, since the Prometheus exporter currently refreshes data every 15 seconds, I used a smaller interval in the relevant tests to pass the original tests.

Fixes #

#### How to

0. install wrk2, git clone apisix
1. deploy etcd(`./test.sh init-etcd`), nginx(as upstream, `./test.sh start-nginx`)
2. `make run`
3. create 10k routes(`./test.sh create`)
4. enable promethues(`./test.sh enable-prometheus`)
5. run benchmark(`./test.sh benchmark`)

#### Key Performance Indicators

| Scenario | CPU Usage | Memory Usage | P50 Latency | P90 Latency | P99 Latency |
|---|---|---|---|---|---|
| Prometheus Disabled | 33.0% | 144.3MB | 3.58ms | 5.99ms | 8.99ms |
| Prometheus Enabled | 38.3% (+5.2%) | 100.2MB (-44.1MB) | 4.22ms (+0.64ms) | 7.07ms (+1.08ms) | 10.66ms (+1.67ms) |
| 1 Metrics Reqs | 38.6% (+5.6%) | 101.3MB (-42.9MB) | 4.25ms (+0.68ms) | 7.87ms (+1.87ms) | 12.17ms (+3.18ms) |
| 3 Metrics Reqs | 39.1% (+6.1%) | 100.1MB (-44.2MB) | 3.88ms (+0.30ms) | 7.17ms (+1.18ms) | 11.92ms (+2.93ms) |

#### Performance Impact Summary

| Scenario | CPU Impact | Memory Impact | P50 Latency Impact | P90 Latency Impact | P99 Latency Impact |
|---|---|---|---|---|---|
| Prometheus Disabled (Baseline) | 0.0% | 0.0MB | 0.00ms | 0.00ms | 0.00ms |
| Prometheus Enabled | +5.2% | -44.1MB | +0.64ms | +1.08ms | +1.67ms |
| 1 Metrics Reqs  | +5.6% | -42.9MB | +0.68ms | +1.87ms | +3.18ms |
| 3 Metrics Reqs | +6.1% | -44.2MB | +0.30ms | +1.18ms | +2.93ms |

worker_performance_individual

latency_metrics_individual

### Checklist


apisix/plugins/prometheus/exporter.lua
```diff
@@ -310,39 +309,50 @@ function _M.stream_log(conf, ctx)
...
+    {name = "accepted", var = "ngx_stat_accepted"},
+    {name = "handled", var = "ngx_stat_handled"},
+    {name = "total", var = "ngx_stat_requests"},
+    {name = "reading", var = "ngx_stat_reading"},
+    {name = "writing", var = "ngx_stat_writing"},
+    {name = "waiting", var = "ngx_stat_waiting"},
+}
+
+-- Use FFI to get nginx status directly from global variables
 local function nginx_status()
```
SkyeYoung, line 367: Because `API disabled in the context of ngx.timer, context: ngx.timer`.

Here use `FFI` to rewrite the `nginx_status` logic.

|OLD(master)|NEW(current)|
|-|-|
|image|image|

apisix/plugins/prometheus/exporter.lua
```diff
@@ -454,10 +458,11 @@ local function collect(ctx, stream_only)
     local config = core.config.new()
 
     -- config server status
-    local vars = ngx.var or {}
-    local hostname = vars.hostname or ""
+    local hostname = core.utils.gethostname() or ""
```
SkyeYoung, line 489: Because `API disabled in the context of ngx.timer, context: ngx.timer`.

apisix/plugins/prometheus/exporter.lua
bzp2010: Underhood: The `ngx.var` is related to the request context, which is inaccessible; nginx subrequests are also not available. Hence the rewrite to FFI.

This not only eliminates the total connections offset caused by the APISIX plugin requesting the `/apisix/status` API itself but also improves efficiency.

apisix/core/config_etcd.lua
```diff
@@ -1001,7 +1001,7 @@ function _M.new(key, opts)
         sync_times = 0,
         running = true,
         conf_version = 0,
-        values = nil,
+        values = {},
```
SkyeYoung, line 1004: Without this modification, the following problems will be encountered:

```
Failed test 't/core/config_etcd.t TEST 12: test route with special character "-" - pattern "[error]" should not match any line in error.log but matches line "2025/07/09 14:24:01 [error] 1169428#1169428: *26 [lua] exporter.lua:534: Failed to collect metrics: 

/home/xxx/apisix/apisix/consumer.lua:77: attempt to get length of local 'data_list' (a nil value), context: ngx.timer" (req 0)
```

In practice, even in [init.lua](https://github.com/apache/apisix/pull/12383/files#diff-7439ffe94cc4bcad12066d5d83f6c1004b8e9b3ff5097b645fa7c700eb3b3743), placing the `plugin.init_prometheus()` after all `.init_worker`, the relevant values may still be `nil` as shown in the error above. Therefore, it is necessary to handle the `values` before initialization.

----

After the discussion with yuansheng and bzp a few days ago, there are two ways to handle it, one is the current one, and the other is as follows:

```lua
-- add between https://github.com/SkyeYoung/apisix/blob/7a06f5c2f7b74cf621a0c2ff6878e8aa0e4e99a7/apisix/core/config_etcd.lua#L1030-L1031
else
  load_full_data(obj, { nodes: {}}, nil)
  self.need_reload = true
-- ^ This is why not use this method, 
-- there is a hidden variable in `load_full_data` that needs to be reset
```

Without resetting this variable, the [etcd-sync.t#TEST 5](https://github.com/apache/apisix/blob/4377d05491ae2af228d40e3900362371363a8f46/t/core/etcd-sync.t#L141-L160) will be failed. (The reason is that the subsequent execution of `load_full_data` is missing)

Thanks for the help from @bzp2010 , I was lost in the code.

apisix/cli/ngx_tpl.lua
```diff
@@ -100,18 +100,14 @@ http {
     }
 
     server {
-        {% if use_apisix_base then %}
```
SkyeYoung, line 103: Now, this API can be run in a normal worker

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +506,70 @@ local function collect(ctx, stream_only)
...
+
+    exporter_timer_running = false
+end
+
+local function init_exporter_timer()
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    ngx.timer.at(0, exporter_timer)
```
SkyeYoung: After the modification, this place can still only be asynchronous at the moment.

If synchronization is needed, we need to modify the following parts:

https://github.com/apache/apisix/blob/6fb9bf94281525c1fca397f681b4890b69440369/apisix/plugins/prometheus/exporter.lua#L467-L494

The reason is that this part requests the API, which will lead to the following error when directly using `exporter_timer()`:

```txt
2025/07/21 14:41:12 [error] 464992#464992: init_worker_by_lua error: /home/xxx/apisix//deps/share/lua/5.1/resty/http.lua:74: API disabled in the context of init_worker_by_lua*
stack traceback:
  [C]: in function 'co_create'
  /home/xxx/apisix//deps/share/lua/5.1/resty/http.lua:74: in function '_body_reader'
  /home/xxx/apisix//deps/share/lua/5.1/resty/http.lua:821: in function 'request'
  /home/xxx/apisix//deps/share/lua/5.1/resty/etcd/v3.lua:120: in function 'request_uri_via_unix_socket'
  /home/xxx/apisix//deps/share/lua/5.1/resty/etcd/v3.lua:160: in function 'http_request_uri'
  /home/xxx/apisix//deps/share/lua/5.1/resty/etcd/v3.lua:251: in function 'server_version'
  /home/xxx/apisix/apisix/core/config_etcd.lua:1066: in function 'server_version'
  /home/xxx/apisix/apisix/plugins/prometheus/exporter.lua:474: in function 'collect'
  /home/xxx/apisix/apisix/plugins/prometheus/exporter.lua:537: in function 'exporter_timer'
  /home/xxx/apisix/apisix/plugins/prometheus/exporter.lua:553: in function 'init_exporter_timer'
  /home/xxx/apisix/apisix/plugin.lua:808: in function 'init_prometheus'
  /home/xxx/apisix/apisix/init.lua:161: in function 'http_init_worker'
  init_worker_by_lua:2: in main chunk
```

So, if we need to ensure synchronized initialization, we need to continue the discussion: **whether to remove or move the collection of this part of the indicators.**

apisix/plugins/prometheus/exporter.lua
```diff
@@ -211,6 +221,14 @@ end
 
 
 function _M.stream_init()
+    -- FIXME:
+    -- Now the HTTP subsystem loads the stream plugin unintentionally, which shouldn't happen.
+    -- It breaks the initialization logic of the plugin,
+    -- here it is temporarily fixed using a workaround.
+    if ngx.config.subsystem ~= "stream" then
+        return
+    end
```
SkyeYoung: As mentioned in the comments, the http subsystem also loads the stream plugins. **This is an issue that needs to be resolved.**

Image

Image

apisix/plugin.lua
```diff
@@ -356,15 +354,6 @@ function _M.load(config)
...
             end
-
-            local enabled = core.table.array_find(http_plugin_names, "prometheus") ~= nil
-            local active  = exporter.get_prometheus() ~= nil
-            if not enabled then
-                exporter.destroy()
-            end
-            if enabled and not active then
-                exporter.http_init()
-            end
```
bzp2010, line 367: Add some description under this comment explaining why we removed it and moved to init and destroy hooks.

apisix/plugin.lua
```diff
@@ -808,18 +797,18 @@ do
...
+    local enabled_in_stream = core.table.array_find(stream_plugin_names, "prometheus")
 
-    -- some plugins need to be initialized in init* phases
-    if is_http and core.table.array_find(http_plugin_names, "prometheus") then
-        local prometheus_enabled_in_stream =
-            core.table.array_find(stream_plugin_names, "prometheus")
-        require("apisix.plugins.prometheus.exporter").http_init(prometheus_enabled_in_stream)
-    elseif not is_http and core.table.array_find(stream_plugin_names, "prometheus") then
-        require("apisix.plugins.prometheus.exporter").stream_init()
+    if is_http and (enabled_in_http or enabled_in_stream) then
```
bzp2010, line 808: ### NOTE

We will always only handle metrics generation in the http subsystem.

1. This will ensure that there is no duplication of execution on http and stream to waste compute resources.
2. This simplifies the design.
3. Whether or not the user has http enabled (i.e., whether or not it is in stream only mode), an http block for the Prometheus export API and its server block (`:9091`) will always be present, otherwise Prometheus would be pointless. This means that we can always have an http subsystem context for periodic generation of timers and metrics anyway, even if we are currently in stream only mode.

bzp2010: Please add some comments to the code to document the design intent. @SkyeYoung

apisix/plugins/prometheus.lua
```diff
@@ -35,6 +34,7 @@ local _M = {
     priority = 500,
     name = plugin_name,
     log  = exporter.http_log,
+    destroy = exporter.destroy,
```
bzp2010, line 37: ### NOTE

This will always destroy the plugin (the prometheus instance in it) when reloading it using the Admin API and loading it again based on the latest configuration.
If a reload is performed after a plugin is removed from the profile list, this plugin will not be restored again. Until the next reload.

Technically, `exporter.destroy` just backs up that instance of the prometheus module and copies it to another variable.
This will cause the export API to stop working, at which point it will always return a `{}`, which is consistent with the current behavior.
Under the hood, the timer will also stop working, no longer generating metrics based on interval timing, and the metrics computation overhead introduced by APISIX is completely eliminated.
When the next plugin reload occurs, if prometheus is re-enabled, the timer will resume running.

Regarding the background timer introduced by the prometheus third-party library, unfortunately, it never stops running.
It is registered with `ngx.timer.every` to perform the task of synchronizing the shdict at regular intervals, and this overhead cannot be paused or resumed by external intervention unless we fork and modify the library itself.

So this destruction does not mean that the prometheus instance is actually destroyed, the synchronization timer is stopped, and the shdict is cleared. none of this happens.

apisix/plugins/prometheus.lua
```diff
@@ -55,4 +55,11 @@ function _M.api()
 end
 
 
+function _M.init()
```
bzp2010, line 58: ### NOTE

We turned to using the built-in hooks of the plugin system, namely `init` to initialize the prometheus instance and prometheus metrics registration.

Note, however, that data padding only occurs the first time the plugin is started (it is usually when the worker is started, i.e. the `init_prometheus` call in `init.lua http_init_worker`) and every timer.
This initialization just registers the metrics, but doesn't really populate the data.

apisix/plugins/prometheus.lua
```diff
@@ -55,4 +55,11 @@ function _M.api()
...
+    local local_conf = core.config.local_conf()
+    local enabled_in_stream = core.table.array_find(local_conf.stream_plugins, "prometheus")
+    exporter.http_init(enabled_in_stream)
```
bzp2010, line 61: ### NOTE

The prometheus plugin, loaded by the http subsystem, will register http metrics there, and will decide whether to register stream metrics (xrpc) depending on whether the stream subsystem has been started.
This is mainly for metrics generation needs in privileged processes, stream data is not really reported at any phase in the http subsystem.

apisix/stream/plugins/prometheus.lua
```diff
@@ -35,6 +35,8 @@ local _M = {
     priority = 500,
     name = plugin_name,
     log  = exporter.stream_log,
+    destroy = exporter.destroy,
+    init = exporter.stream_init,
```
bzp2010, line 39: ### NOTE

Initialization in the stream subsystem is very pure, it just requires unconditional registration of stream metrics and xrpc metrics. It doesn't need to care if the http subsystem is enabled or not.

t/cli/test_prometheus_run_in_privileged.sh
bzp2010, line 1: ### NOTE

The export API will no longer be exposed to privileged processes, which provides isolation of HTTP traffic from root privileges for enhanced security.
Therefore this is no longer needed.

apisix/cli/ngx_tpl.lua
bzp2010: The export API will no longer be exposed to privileged processes, which provides isolation of HTTP traffic from root privileges for enhanced security.
Therefore this is no longer needed.

t/plugin/prometheus.t
```diff
@@ -37,6 +37,14 @@ add_block_preprocessor(sub {
     if (!defined $block->request) {
         $block->set_value("request", "GET /t");
     }
+
+    if (!defined $block->yaml_config) {
+        $block->set_value("yaml_config", <<'EOF');
+plugin_attr:
+    prometheus:
+        refresh_interval: 0.1
```
bzp2010, line 45: ### NOTE

We use smaller values when we need an extremely fast refresh.
You don't need to worry about timers being executed too often or overlapping, if a metrics generation process doesn't complete, the next one will never start.

apisix/plugins/prometheus/exporter.lua
bzp2010: Please create an issue for this. thx @SkyeYoung

apisix/plugins/prometheus/exporter.lua
```diff
@@ -454,34 +477,37 @@ local function collect(ctx, stream_only)
...
+    local stream_only = local_conf.apisix.proxy_mode == "stream"
     -- we can't get etcd index in metric server if only stream subsystem is enabled
     if config.type == "etcd" and not stream_only then
         -- etcd modify index
         etcd_modify_index()
 
-        local version, err = config:server_version()
-        if version then
-            metrics.etcd_reachable:set(1)
+        if yieldable then
```
bzp2010, line 499: ### NOTE

Metrics contains the etcd reachability report and the etcd latest modified index report, which relies on communication with etcd.
According to openresty's restrictions, yield is prohibited in the `init_worker` phase, i.e. cosocket-based communication with etcd is not allowed.
So we skip a capture here until the timer does them.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +524,71 @@ local function collect(ctx, stream_only)
...
+    return core.table.concat(prometheus:metric_data())
+end
+
+
+local function exporter_timer(premature, yieldable)
+    if premature then
+        return
+    end
+
+    if not prometheus then
```
bzp2010, line 545: ### NOTE

This is used for dynamic disable (by plugin reload API) of the plugin, i.e. done in `exporter.destroy`.
This is where we stop if the prometheus instance is "destroyed", and as you can see, this will happen before scheduling the next timer task, which means the timer will stop.

Technically, this is the advantage of `ngx.timer.at` over `ngx.timer.every`, every is not terminable, the developer can't get an "instance" of a timer to pause or stop it.
But by using `ngx.timer.at`, we can precisely control whether or not to continue scheduling the next timed task, which allows us to stop the timer. If you need to resume it, just re-execute `ngx.timer.at(0)`.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +524,71 @@ local function collect(ctx, stream_only)
...
+    exporter_timer_running = false
+end
+
+
+local function init_exporter_timer()
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    exporter_timer(false, false)
```
bzp2010: ### NOTE

The initialization of the timer will perform an acquisition task synchronously, i.e. the first acquisition will always occur in the init_worker phase, which provides initial access to the metrics data.

If at any time , the metrics data (the string in the `prometheus-cache` shdict) is not available, the API will report an error and log it. This is by design not very likely to happen.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +524,71 @@ local function collect(ctx, stream_only)
...
+                     " 'prometheus_metrics' shared dict is present in nginx template")
+        return 500, {message = "An unexpected error occurred"}
+    end
+
     core.response.set_header("content_type", "text/plain")
-    return 200, core.table.concat(prometheus:metric_data())
+    local cached_metrics_text = shdict_prometheus_cache:get(CACHED_METRICS_KEY)
+    if not cached_metrics_text then
+        core.log.error("Failed to retrieve cached metrics: data is nil")
+        return 500, "Failed to retrieve metrics: no data available"
```
bzp2010: Is it more standard to return a JSON with reference to 
https://github.com/apache/apisix/pull/12383/files#diff-390eaff60bfa1071dd1850bce9c7689452eaa13f07bf45a927921ecf05886d1bR580 ?

apisix/plugins/prometheus/exporter.lua
```diff
@@ -530,16 +617,16 @@ end
 _M.get_api = get_api
 
 
-function _M.export_metrics(stream_only)
+function _M.export_metrics()
     if not prometheus then
-        core.response.exit(200, "{}")
+       return core.response.exit(200, "{}")
```
bzp2010: JFI, this behavior seems to be inconsistent with what is in get_cached_metrics and we need to confirm which mode should be used. cc @membphis 

BTW, prometheus to nil will happen when the plugin is dynamically disabled.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -110,6 +116,10 @@ end
 
 
 function _M.http_init(prometheus_enabled_in_stream)
+    if ngx.config.subsystem ~= "http" then
+        return
```
membphis, line 128: we can print some warning error log, it is unexpected

apisix/plugins/prometheus/exporter.lua
```diff
@@ -211,6 +221,14 @@ end
 
 
 function _M.stream_init()
+    -- FIXME:
+    -- Now the HTTP subsystem loads the stream plugin unintentionally, which shouldn't happen.
+    -- It breaks the initialization logic of the plugin,
+    -- here it is temporarily fixed using a workaround.
+    if ngx.config.subsystem ~= "stream" then
+        return
```
membphis, line 238: ditto

apisix/plugins/prometheus/exporter.lua
membphis: another way:

we can use `resty.http` to send the http request to itself for fetching `nginx status`

which is easier to read

anyway, I can accept current way but I can not sure it is easier enough to maintain for other developers

I highly recommend to add some comments which is useful

apisix/plugins/prometheus/exporter.lua
```diff
@@ -438,13 +467,7 @@ local function shared_dict_status()
...
 
 
-local function collect(ctx, stream_only)
-    if not prometheus or not metrics then
-        core.log.error("prometheus: plugin is not initialized, please make sure ",
-                     " 'prometheus_metrics' shared dict is present in nginx template")
-        return 500, {message = "An unexpected error occurred"}
-    end
-
+local function collect(yieldable)
```
membphis, line 479: can we return directly if the `yieldable` is `false`?

apisix/plugins/prometheus/exporter.lua
membphis: the current way is good to me

conf/config.yaml.example
```diff
@@ -170,6 +170,7 @@ nginx_config:                     # Config for render the template to generate n
   meta:
     lua_shared_dict:              # Nginx Lua shared memory zone. Size units are m or k.
       prometheus-metrics: 15m
+      prometheus-cache: 10m
```
membphis: pls add some comments, tell users when they need to modify it

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +524,71 @@ local function collect(ctx, stream_only)
...
+    end
+
+    exporter_timer_running = true
+
+    local ok, res = pcall(collect, yieldable)
+    if not ok then
+        core.log.error("Failed to collect metrics: ", res)
+        return
+    end
+    shdict_prometheus_cache:set(CACHED_METRICS_KEY, res)
```
membphis: need to capture the return value, it maybe fail

if `err`, tell user what is the reason, and tell the user to change the default size if the `shdict` is small

apisix/plugins/prometheus/exporter.lua
bzp2010: I suggest to perform the first capture and metrics generation at `init_worker`, it is synchronously blocking, which makes the metrics almost always available, otherwise we can't be sure when the metrics will be available, which can lead to some unexpected exception responses (HTTP 500).

apisix/plugins/prometheus/exporter.lua
bzp2010: This is a problem that needs to be solved, and after it is solved, it seems that we will no longer need these assertions.

apisix/plugins/prometheus/exporter.lua
bzp2010: @membphis 

Actually, I recommend using these FFI API for data acquisition.
I confirmed from the nginx code that this data is synchronized between workers by shared memory, so FFI API can access them.

And this mechanism is entirely LuaJIT, not nginx fake request or openresty cosocket. 
The former doesn't introduce any noise. 
And when using any of the latter, no matter which one, the fetching behavior itself causes accepted/active/handled/reading/waiting/writing metrics to increase. waiting/writing. This is due to the fact that these mechanisms are always requesting APIs as network sockets, and they themselves cause the metrics to go up. This has always been a problem, and while I can understand it and it doesn't cause serious problems, it's always been confusing.

I came up with this idea and @SkyeYoung  did it independently after some simple research, which I'm sure is not a difficult task for developers with almost any AI LLM assistance.
These c-variables haven't changed in years, and I don't think it's going to change nearly as much in the future (there's no need), so it's not really an area that needs constant attention. If any future openresty/nginx breaks this convention, our test cases can find them.

SkyeYoung: I remember I tried using `resty.http` locally and ran into some problems. This was also one of the first methods suggested by @bzp2010. Because there were a lot of problems with the code, I didn't even submit it, and now I can't find it.

As for cosocket, it's even harder for me, a beginner, to understand.

Later, I switched to FFI under the suggestion of @bzp2010 , and found this approach to be actually very simple and straightforward.

apisix/plugin.lua
SkyeYoung: [The original code](https://github.com/apache/apisix/pull/11117) skipped [plugin.init()](https://github.com/SkyeYoung/apisix/blob/72c6f229e681b36efa226b51713ebdb23075a575/apisix/plugin.lua#L191-L192) and [old_plugin.destroy()](https://github.com/SkyeYoung/apisix/blob/72c6f229e681b36efa226b51713ebdb23075a575/apisix/plugin.lua#L121-L122) used in https://github.com/apache/apisix/blob/6fb9bf94281525c1fca397f681b4890b69440369/apisix/plugin.lua, and implemented the overload of the prometheus plugin for some reason that I have not yet understood (perhaps because `prometheus.lua` originally did not contain two functions init and destroy).

---

The initial reason was that even after separating the `init_prometheus` part and placing it at the end of `init_worker`, directly calling `exporter_timer()` would still cause an error. After debugging, another initialization logic was found here. This is obviously redundant.

Currently, we provide `init` and `destroy` functions in `prometheus.lua`, allowing the initialization and reloading of the prometheus plugin to be handled within the plugin's own files, reducing coupling.

This also allows the prometheus plugin to revert to the mechanism provided by `plugin.lua`, reducing special cases, lowering the cost of understanding, and making the code easier to maintain.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: This parameter is currently used to solve the problem of not being able to obtain the following data during initialization.

I elaborated in detail in https://github.com/apache/apisix/pull/12383#discussion_r2220706155.

Image

apisix/plugin.lua
SkyeYoung: done.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: done.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: done.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: done

conf/config.yaml.example
SkyeYoung: done.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: done

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +530,83 @@ local function collect(ctx, stream_only)
...
+    if not ok then
+        core.log.error("Failed to collect metrics: ", res)
+        exporter_timer_running = false
+        return
+    end
+
+    local _, err, forcible = shdict_prometheus_cache:set(CACHED_METRICS_KEY, res)
+
+    if err then
+        core.log.error("Failed to save metrics to shdict: ", err)
```
membphis: need more information, eg: the name of `shdict` and the size of value

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +530,83 @@ local function collect(ctx, stream_only)
...
+local function get_cached_metrics()
+    if not prometheus or not metrics then
+        core.log.error("prometheus: plugin is not initialized, please make sure ",
+                     " 'prometheus_metrics' shared dict is present in nginx template")
+        return 500, {message = "An unexpected error occurred"}
+    end
+
     core.response.set_header("content_type", "text/plain")
-    return 200, core.table.concat(prometheus:metric_data())
+    local cached_metrics_text = shdict_prometheus_cache:get(CACHED_METRICS_KEY)
```
membphis: just confirm, no "err" is returned?

apisix/plugins/prometheus/exporter.lua
SkyeYoung: I'll create it a little later

apisix/plugins/prometheus/exporter.lua
membphis: if the value is to large, will fail when we call `set` method

apisix/plugins/prometheus/exporter.lua
SkyeYoung: I checked [the part code](https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/shdict.lua#L589C16) and found that the second return value, marked as "flags" in [its documentation](https://github.com/openresty/lua-nginx-module#ngxshareddictget), actually returns an error when an error occurs.

Let me fix this part of the code.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +530,88 @@ local function collect(ctx, stream_only)
...
+        return
+    end
+
+    local refresh_interval = DEFAULT_REFRESH_INTERVAL
+    local attr = plugin.plugin_attr("prometheus")
+    if attr and attr.refresh_interval then
+        refresh_interval = attr.refresh_interval
+    end
+
+    ngx.timer.at(refresh_interval, exporter_timer, true)
```
membphis: it may fail: https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#ngxtimerat

I think we have to use a new way

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +530,88 @@ local function collect(ctx, stream_only)
...
+    exporter_timer_running = true
+
+    local ok, res = pcall(collect, yieldable)
+    if not ok then
+        core.log.error("Failed to collect metrics: ", res)
+        exporter_timer_running = false
+        return
+    end
+
+    local _, err, forcible = shdict_prometheus_cache:set(CACHED_METRICS_KEY, res)
```
membphis: I think we need to set an expire time

image


if the refresh interval is  15sec, I think the ttl should be about 30sec

membphis (changes requested):

apisix/plugins/prometheus/exporter.lua
```diff
@@ -498,10 +533,105 @@ local function collect(ctx, stream_only)
...
+    local ok, res = pcall(collect, yieldable)
+    if not ok then
+        core.log.error("Failed to collect metrics: ", res)
+        exporter_timer_running = false
+        return
+    end
+
+    -- Clear the cached data after cache_exptime to prevent stale data in case of an error.
+    local _, err, forcible = shdict_prometheus_cache:set(CACHED_METRICS_KEY, res, cache_exptime)
+
```
membphis: remove this blank line

membphis (approved):
LGTM

apisix/plugin.lua
bzp2010: It might be better to add a link to this PR comment here.

https://github.com/apache/apisix/pull/12383/files#r2221993953

apisix/plugins/prometheus/exporter.lua
SkyeYoung:
```suggestion
```

bzp2010 (approved):

apisix/plugin.lua
SkyeYoung: @bzp2010 I think this part of the code can be found by modifying the history, just like the old code.

apisix/plugins/prometheus/exporter.lua
SkyeYoung: done

apisix/plugin.lua
```diff
@@ -808,18 +797,21 @@ do
...
+    -- And for other modes, whether in stream or http plugins,
+    -- the prometheus exporter needs to be initialized.
+    if is_http and (enabled_in_http or enabled_in_stream) then
+        require("apisix.plugins.prometheus.exporter").init_exporter_timer()
     end
+end
 
+
+function _M.init_worker()
     -- someone's plugin needs to be initialized after prometheus
```
yurkovoznyak, line 815: Hi @SkyeYoung, I'm using Prometheus metrics in my custom plugin, and with this change, there is no guarantee that the prometheus plugin is loaded before mine. What is the correct approach for defining custom metrics now?

Sorry for commenting on a closed PR

Baoyuantop: Hi @yurkovoznyak, could you please file an issue describing your problem in detail?

Issue #11200, opened 2024-04-25
bug: After the forward-auth verification is passed, the upstream cannot obtain the request body (payload) with 504 Timeout


OnGoingLzy:
### Current Behavior

After the forward-auth verification is passed, the upstream cannot obtain the request body (payload)
![image](https://github.com/apache/apisix/assets/65748123/998b7906-a740-4732-a90c-2620c447061d)
getData2 can be called normally, but getData() doesn't, and after many tests I think there are no parameters in the payload

![image](https://github.com/apache/apisix/assets/65748123/e17948a3-fc1b-4341-9900-2466b8d46eea)

This screenshot shows the error on the C# backend
![image](https://github.com/apache/apisix/assets/65748123/6b0b47e0-2169-4720-97d3-3d47d00e060a)

(GetData call failure occurs only when forward-auth is used, and can be called normally if it is not used)

Here's my Foward-auth configuration:
"forward-auth": {
      "_meta": {
        "disable": false
      },
      "client_headers": [
        "apk",
        "timestamp",
        "nonce",
        "sign",
        "Content-Length",
        "Expect",
        "Transfer-Encoding",
        "Content-Encoding"
      ],
      "request_headers": [
        "apk",
        "timestamp",
        "nonce",
        "sign",
        "Content-Length",
        "Expect",
        "Transfer-Encoding",
        "Content-Encoding"
      ],
      "request_method": "POST",
      "status_on_error": 500,
      "upstream_headers": [
        "apk",
        "timestamp",
        "nonce",
        "sign",
        "Content-Length",
        "Expect",
        "Transfer-Encoding",
        "Content-Encoding"
      ],
      "uri": "http://localhost:9080/authApi/api/authorization/authorization1"
    },

other QS：forward-auth's configuration request_method:"GET" will result in an error: “time out”. Like mentioned in #11021,How can I avoid these issues in apisix-docker?

### Expected Behavior

I hope that when using the forward-auth plugin, my request will be successfully requested to my upstream

### Error Logs

apisix log :<2024-04-25 13:19:46 127.0.0.1 - - [25/Apr/2024:05:19:43 +0000] localhost:9080 "POST /authApi/api/authorization/authorization1 HTTP/1.1" 200 56 0.022 "-" "lua-resty-http/0.16.1 (Lua) ngx_lua/10026" 172.30.14.255:15976 200 0.021 "http://localhost:9080/api/authorization/authorization1"
[...]

PR #10589, opened 2023-12-04
fix: forward-auth request body is too large

Sn0rt:
## Description

The gateway returns 403 forbidden in the forward-auth.lua file because the post request body is too large

Fixes https://github.com/apache/apisix/issues/10375

PR #12404, opened 2025-07-07
revert: fix: forward-auth request body too large

Revolyssup:
### Description
This PR https://github.com/apache/apisix/pull/10589/files, introduced a bug due to which the request body is emptied after being read by the auth service and the upstream doesn't receive the request body. Therefore, this PR needs to be reverted.

The new solution after this revert is accepted will be:
Do not pass the whole request body to the auth service and extract data from body into headers and then forward to the auth service. The headers to be extracted will be specified with a new field: `extra_headers`

Fixes https://github.com/apache/apisix/issues/11200

### Checklist

### DIFF
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -118,23 +118,17 @@ function _M.access(conf, ctx)
         method = conf.request_method
     }
 
-    local httpc = http.new()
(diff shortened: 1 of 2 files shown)


bzp2010 (approved):

membphis (approved):
LGTM

PR #12424, opened 2025-07-14
fix(standalone): lack of configuration validation in api

AlinsRan:

In standalone mode, the lack of code validation related to admin API results in incorrect configurations being able to pass validation, as shown in the following example:
* Invalid plugin
* route/service with invalid upstream

##### When configuring an invalid upstream:

**Expect:**
```bash
curl "http://127.0.0.1:9180/apisix/admin/configs" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -v -d '{
    "routes": [
        {
            "id": "r1",
            "uri": "/r2",
            "upstream": {
                "type": "chash",
                "nodes": {
                    "127.0.0.1:1980": 1
                },
                "hash_on": "vars",
                "key": "args_abc"
            }
        }
    ]
}'

< HTTP/1.1 400 Bad Request
< Server: APISIX/3.13.0
< 
{"error_msg":"invalid configuration: failed to match pattern \"^((uri|server_name|server_addr|request_uri|remote_port|remote_addr|query_string|host|hostname|mqtt_client_id)|arg_[0-9a-zA-z_-]+)$\" with \"args_abc\""}
```

**Actual:**
```
curl "http://127.0.0.1:9180/apisix/admin/configs" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -v -d '{
    "routes": [
        {
            "id": "r1",
            "uri": "/r2",
            "upstream": {
                "type": "chash",
                "nodes": {
                    "127.0.0.1:1980": 1
                },
                "hash_on": "vars",
                "key": "args_abc"
            }
        }
    ]
}'
> 
< HTTP/1.1 202 Accepted
< Server: APISIX/3.13.0
< 
```

##### When configuring an invalid plugin

**Actual:**

```bash
 curl "http://127.0.0.1:9180/apisix/admin/configs" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -v -X PUT -d '{
    "routes": [
        {
            "id": "r1",
            "uri": "/r2",
            "upstream": {
                "type": "chash",
                "nodes": {
                    "127.0.0.1:1980": 1
                },
                "hash_on": "vars",
                "key": "arg_abc"
            },
            "plugins": {
                "proxy-rewrite234": {
                    "uri": "/hello"
                }
            }
        }
    ]
}'
>
< HTTP/1.1 202 Accepted
< Server: APISIX/3.13.0
<
```

**Expect:**

```bash
curl "http://127.0.0.1:9180/apisix/admin/configs" -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '{
    "routes": [
        {
            "id": "r1",
            "uri": "/r2",
            "upstream": {
                "type": "chash",
                "nodes": {
                    "127.0.0.1:1980": 1
                },
                "hash_on": "vars",
                "key": "args_abc"
            },
            "plugins": {
                "proxy-rewrite234": {
                    "uri": "/hello"
                }
            }
        }
    ]
}'
>
< HTTP/1.1 400 Bad Request
< Server: APISIX/3.13.0
< 
{"error_msg":"unknown plugin [proxy-rewrite234]"}
```

### Checklist

### DIFF
--- a/apisix/admin/consumers.lua
+++ b/apisix/admin/consumers.lua
@@ -19,7 +19,7 @@ local plugins = require("apisix.admin.plugins")
 local resource = require("apisix.admin.resource")
 
 
-local function check_conf(username, conf, need_username, schema)
+local function check_conf(username, conf, need_username, schema, skip_etcd_check)
     local ok, err = core.schema.check(schema, conf)
     if not ok then
         return nil, {error_msg = "invalid configuration: " .. err}
@@ -36,7 +36,7 @@ local function check_conf(username, conf, need_username, schema)
         end
     end
 
-    if conf.group_id then
+    if conf.group_id and not skip_etcd_check then
         local key = "/consumer_groups/" .. conf.group_id
         local res, err = core.etcd.get(key)
         if not res then
@@ -60,6 +60,11 @@ return resource.new({
     name = "consumers",
     kind = "consumer",
     schema = core.schema.consumer,
-    checker = check_conf,
+    checker = function (username, conf, need_username, schema)
+        return check_conf(username, conf, need_username, schema)
+    end,
+    standalone_checker = function (username, conf, need_username, schema)
+        return check_conf(username, conf, need_username, schema, true)
+    end,
     unsupported_methods = {"post", "patch"}
 })
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -34,6 +34,24 @@ local EVENT_UPDATE = "standalone-api-configuration-update"
 
 local _M = {}
 
+local resources = {
+    routes          = require("apisix.admin.routes"),
+    services        = require("apisix.admin.services"),
+    upstreams       = require("apisix.admin.upstreams"),
+    consumers       = require("apisix.admin.consumers"),
+    credentials     = require("apisix.admin.credentials"),
+    schema          = require("apisix.admin.schema"),
+    ssls            = require("apisix.admin.ssl"),
+    plugins         = require("apisix.admin.plugins"),
+    protos          = require("apisix.admin.proto"),
+    global_rules    = require("apisix.admin.global_rules"),
+    stream_routes   = require("apisix.admin.stream_routes"),
+    plugin_metadata = require("apisix.admin.plugin_metadata"),
+    plugin_configs  = require("apisix.admin.plugin_config"),
+    consumer_groups = require("apisix.admin.consumer_group"),
+    secrets         = require("apisix.admin.secrets"),
+}
+
 local function check_duplicate(item, key, id_set)
     local identifier, identifier_type
     if key == "consumers" then
@@ -135,6 +153,7 @@ local function update(ctx)
         local conf_version = config and config[conf_version_key] or obj.conf_version
         local items = req_body[key]
         local new_conf_version = req_body[conf_version_key]
+        local resource = resources[key]
         if not new_conf_version then
             new_conf_version = conf_version + 1
         else
@@ -151,37 +170,33 @@ local function update(ctx)
             end
         end
 
+
         apisix_yaml[conf_version_key] = new_conf_version
         if new_conf_version == conf_version then
             apisix_yaml[key] = config and config[key]
-        elseif items and #items > 0 then
+        elseif resource and items and #items > 0 then
             apisix_yaml[key] = table_new(#items, 0)
-            local item_schema = obj.item_schema
-            local item_checker = obj.checker
+            local item_schema = resource.schema
+            local item_checker = resource.standalone_checker or resource.checker
             local id_set = {}
 
-            for index, item in ipairs(items) do
+            for _, item in ipairs(items) do
                 local item_temp = tbl_deepcopy(item)
                 local valid, err
-                -- need to recover to 0-based subscript
-                local err_prefix = "invalid " .. key .. " at index " .. (index - 1) .. ", err: "
-                if item_schema then
-                    valid, err = check_schema(obj.item_schema, item_temp)
-                    if not valid then
-                        core.log.error(err_prefix, err)
-                        core.response.exit(400, {error_msg = err_prefix .. err})
-                    end
-                end
                 if item_checker then
-                    local item_checker_key
-                    if item.id then
-                        -- credential need to check key
-                        item_checker_key = "/" .. key .. "/" .. item_temp.id
-                    end
-                    valid, err = item_checker(item_temp, item_checker_key)
-                    if not valid then
-                        core.log.error(err_prefix, err)
-                        core.response.exit(400, {error_msg = err_prefix .. err})
+                    if core.string.find(tostring(item.id), "/credentials/") then
+                        local credential_item_checker = resources.credentials.standalone_checker or resources.credentials.checker
+                        local item_schema2 = resources.credentials.schema
+                        valid, err = credential_item_checker(item.id, item_temp, false, item_schema2)
+                        if not valid then
+                            --core.log.error(err_prefix, err)
+                            core.response.exit(400, err)
+                        end
+                    else
+                        valid, err = item_checker(item.id, item_temp, false, item_schema)
+                        if not valid then
+                            core.response.exit(400, err)
+                        end
                     end
                 end
                 -- prevent updating resource with the same ID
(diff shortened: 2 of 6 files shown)


apisix/admin/consumers.lua
```diff
@@ -19,7 +19,7 @@ local plugins = require("apisix.admin.plugins")
 local resource = require("apisix.admin.resource")
 
 
-local function check_conf(username, conf, need_username, schema)
+local function check_conf(username, conf, need_username, schema, skip_etcd_check)
```
nic-6443: I personally recommend using a more general naming convention, such as: `skip_references_check`.

apisix/admin/standalone.lua
```diff
@@ -27,13 +28,30 @@ local yaml         = require("lyaml")
...
+    ssls            = require("apisix.admin.ssl"),
+    plugins         = require("apisix.admin.plugins"),
+    protos          = require("apisix.admin.proto"),
+    global_rules    = require("apisix.admin.global_rules"),
+    stream_routes   = require("apisix.admin.stream_routes"),
+    plugin_metadata = require("apisix.admin.plugin_metadata"),
+    plugin_configs  = require("apisix.admin.plugin_config"),
+    consumer_groups = require("apisix.admin.consumer_group"),
+    secrets         = require("apisix.admin.secrets"),
+}
```
nic-6443, line 55: I think we can rename `ssl.lua`, `proto.lua`, `plugin_config.lua`, and `consumer_group.lua` to keep the naming of code files consistent with resource identifiers, so we don't have to maintain this mapping table anymore.

apisix/admin/consumers.lua
```diff
@@ -60,6 +60,11 @@ return resource.new({
     name = "consumers",
     kind = "consumer",
     schema = core.schema.consumer,
-    checker = check_conf,
+    checker = function (username, conf, need_username, schema)
+        return check_conf(username, conf, need_username, schema, false)
+    end,
+    standalone_checker = function (username, conf, need_username, schema)
+        return check_conf(username, conf, need_username, schema, true)
+    end,
```
nic-6443: I feel that providing two separate `checkers` makes it easy for developers to overlook the `skip_etcd_check` parameter when adding a new resource or adding check code to an existing resource. 
It would be better to add the `skip_etcd_check` parameter to the `check_conf` function of all resources, even if this parameter is not used in that resource like `upstreams`、`ssl` etc.

apisix/admin/standalone.lua
AlinsRan: The processing here can cache the results of require calls without the need to call require multiple times in the function

apisix/admin/consumers.lua
AlinsRan: https://github.com/apache/apisix/blob/eed93f315c917e7bedcddf18d7df56af556cb918/apisix/admin/resource.lua#L126

I had intended to make this change before, but I found that some resources may have more than just four parameters
 So my idea is to create two independent checkers as much as possible to reflect the differences

AlinsRan: fixed.

apisix/admin/consumers.lua
AlinsRan: done.

apisix/admin/standalone.lua
```diff
@@ -151,37 +172,45 @@ local function update(ctx)
...
-                    local item_checker_key
-                    if item.id then
-                        -- credential need to check key
-                        item_checker_key = "/" .. key .. "/" .. item_temp.id
-                    end
-                    valid, err = item_checker(item_temp, item_checker_key)
-                    if not valid then
-                        core.log.error(err_prefix, err)
-                        core.response.exit(400, {error_msg = err_prefix .. err})
+                    if core.string.find(id, "/credentials/") then
```
membphis: https://github.com/apache/apisix/blob/c4ead141b311d43dbfdd0099f975f109b4b89807/apisix/admin/standalone.lua#L189-L213

The code is a bit deep, which makes it difficult to read.
We can move this code to a separate function.

membphis (approved):
LGTM

only one minor issue(about code style)

apisix/admin/standalone.lua
AlinsRan: Good.

nic-6443 (approved):

apisix/admin/consumer_group.lua
```diff
@@ -23,7 +23,7 @@ local tostring = tostring
 local ipairs = ipairs
 
 
-local function check_conf(id, conf, need_id, schema)
+local function check_conf(id, conf, need_id, schema, secret_type, skip_references_check)
```
bzp2010: What is the relevance of secret type to consumer group? In fact, according to these code changes, it has nothing to do with any code other than the secret.

Normally, we wouldn't write something like `fu(nil, true)` in codes. A required field (infer from your use case) should not come after an optional field. 

Most functions don't care about the value of `secret_type`, and more importantly, we can't accept an infinite rise in the number of input arguments to this function. I would suggest using a table directly as an input parameter, so that you can be completely unconcerned about which fields are not needed.

```lua
local function fn1(i1, i2, i3, opts)
  if opts.skip_references_check then
    -- do something
  end
end

local function fn2(i1, i2, i3, opts)
  if opts.secret_type == "xx" then
    -- do something
  end
end
```

apisix/admin/stream_routes.lua
```diff
@@ -72,7 +72,12 @@ return resource.new({
     name = "stream_routes",
     kind = "stream route",
     schema = core.schema.stream_route,
-    checker = check_conf,
+    checker = function (id, conf, need_id, schema)
+        return check_conf(id, conf, need_id, schema, false)
+    end,
+    standalone_checker = function (id, conf, need_id, schema)
+        return check_conf(id, conf, need_id, schema, true)
+    end,
```
bzp2010: Why is `standalone_checker` added only to this resource type and not to the others?

AlinsRan: fixed.

apisix/admin/consumer_group.lua
AlinsRan: fixed.

apisix/admin/consumers.lua
```diff
@@ -36,7 +37,7 @@ local function check_conf(username, conf, need_username, schema)
         end
     end
 
-    if conf.group_id then
+    if conf.group_id and not opts.skip_references_check then
```
bzp2010, line 40: The `skip_references_check` may be nil, when it is `nil` this means that validation is skipped (`not nil` = `not false`), is this as expected?

AlinsRan: Nil or false are default values and will not skip reference checks
When set to true, only standalone will be used and it will skip the reference check

```lua
local opts_tests ={
    {},
    {
        skip_references_check = nil,
    },
    {
        skip_references_check = false,
    },
    {
        skip_references_check = true,
    },
}

local inspect = require("inspect")

for i, opt in ipairs(opts_tests) do
    if not opt.skip_references_check then
        print("fase ----- skip_references_check = false, Test " .. i, ": " .. inspect(opt))
    else
        print("true ------ skip_references_check = true, Test " .. i, ": " .. inspect(opt))
    end
end
```

result:
```bash
fase ----- skip_references_check = false, Test 1: {}
fase ----- skip_references_check = false, Test 2: {}
fase ----- skip_references_check = false, Test 3: {
  skip_references_check = false
}
true ------ skip_references_check = true, Test 4: {
  skip_references_check = true
}
```

t/admin/standalone.spec.ts
```diff
@@ -427,7 +480,7 @@ describe("Admin - Standalone", () => {
       expect(resp.status).toEqual(400);
       expect(resp.data).toMatchObject({
         error_msg:
-          'invalid routes at index 0, err: property "uri" validation failed: wrong type: expected string, got number',
+          'invalid configuration: property \"uri\" validation failed: wrong type: expected string, got number',
```
bzp2010: It seems to be that these error indications are backwards, and there's no easy way to know which resource is wrong.

AlinsRan: fixed.

nic-6443 (approved):

bzp2010 (approved):

PR #12426, opened 2025-07-14
feat: add healthcheck manager to decouple upstream

Revolyssup:

Fixes #

### Checklist

### DIFF
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -1,0 +1,156 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core = require("apisix.core")
+local healthcheck
+local events = require("apisix.events")
+local tab_clone = core.table.clone
+
+local _M = {
+    working_pool = {},     -- resource_path -> {version = ver, checker = checker}
+    waiting_pool = {}      -- resource_path -> resource_ver
+}
+
+local function fetch_latest_conf(resource_path)
+    return core.config.fetch_conf(resource_path)
+end
+
+local function create_checker(up_conf)
+    if not healthcheck then
+        healthcheck = require("resty.healthcheck")
+    end
+
+    core.log.info("creating new healthchecker for ", up_conf.key)
+
+    local checker, err = healthcheck.new({
+        name = "upstream#" .. up_conf.key,
+        shm_name = "upstream-healthcheck",
+        checks = up_conf.checks,
+        events_module = events:get_healthcheck_events_modele(),
+    })
+
+    if not checker then
+        core.log.error("failed to create healthcheck: ", err)
+        return nil
+    end
+
+    -- Add target nodes
+    local host = up_conf.checks and up_conf.checks.active and up_conf.checks.active.host
+    local port = up_conf.checks and up_conf.checks.active and up_conf.checks.active.port
+    local up_hdr = up_conf.pass_host == "rewrite" and up_conf.upstream_host
+    local use_node_hdr = up_conf.pass_host == "node" or nil
+
+    for _, node in ipairs(up_conf.nodes) do
+        local host_hdr = up_hdr or (use_node_hdr and node.domain)
+        local ok, err = checker:add_target(node.host, port or node.port, host,
+                                        true, host_hdr)
+        if not ok then
+            core.log.error("failed to add healthcheck target: ", node.host, ":", 
+                          port or node.port, " err: ", err)
+        end
+    end
+
+    return checker
+end
+
+function _M.fetch_checker(resource_path, resource_ver)
+    -- Check working pool first
+    local working_item = _M.working_pool[resource_path]
+    if working_item and working_item.version == resource_ver then
+        return working_item.checker
+    end
+
+    if _M.waiting_pool[resource_path] == resource_ver then
+        return nil
+    end
+
+    -- Add to waiting pool with version
+    _M.waiting_pool[resource_path] = resource_ver
+    return nil
+end
+
+function _M.timer_create_checker()
+    if core.table.nkeys(_M.waiting_pool) == 0 then
+        return
+    end
+
+    local waiting_snapshot = tab_clone(_M.waiting_pool)
+    for resource_path, resource_ver in pairs(waiting_snapshot) do
+        local res_conf = fetch_latest_conf(resource_path)
+        if not res_conf then
+            _M.waiting_pool[resource_path] = nil
+            goto continue
+        end
+
+        local current_ver = res_conf.modifiedIndex .. "#" .. tostring(res_conf.value) .. "#" ..
+                            tostring(res_conf.value._nodes_ver or '')
+        if resource_ver ~= current_ver then
+            _M.waiting_pool[resource_path] = nil
+            goto continue
+        end
+
+        local checker = create_checker(res_conf.value)
+        if not checker then
+            _M.waiting_pool[resource_path] = nil
+            goto continue
+        end
+
+        _M.working_pool[resource_path] = {
+            version = resource_ver,
+            checker = checker
+        }
+
+        _M.waiting_pool[resource_path] = nil
+        core.log.info("created healthchecker for ", resource_path, " version: ", resource_ver)
+
+        ::continue::
+    end
+end
+
+function _M.timer_working_pool_check()
+    if core.table.nkeys(_M.working_pool) == 0 then
+        return
+    end
+
+    local working_snapshot = tab_clone(_M.working_pool)
+    for resource_path, item in pairs(working_snapshot) do
+        -- Fetch latest configuration
+        local res_conf = fetch_latest_conf(resource_path)
+        if not res_conf then
+            item.checker:delayed_clear(10)
+            item.checker:stop()
+            _M.working_pool[resource_path] = nil
+            goto continue
+        end
+
+        local current_ver = res_conf.modifiedIndex .. "#" .. tostring(res_conf.value) .. "#" ..
+                            tostring(res_conf.value._nodes_ver or '')
+        if item.version ~= current_ver then
+            item.checker:delayed_clear(10)
+            item.checker:stop()
+            _M.working_pool[resource_path] = nil
+        end
+
+        ::continue::
+    end
+end
+
+function _M.init_worker()
+    core.timer.every(1, _M.timer_create_checker)
+    core.timer.every(60, _M.timer_working_pool_check)
+end
+
+return _M
--- a/apisix/upstream.lua
+++ b/apisix/upstream.lua
@@ -14,6 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 --
+local inspect = require("inspect")
 local require = require
 local core = require("apisix.core")
 local config_local   = require("apisix.core.config_local")
@@ -30,6 +31,7 @@ local ngx_var = ngx.var
 local is_http = ngx.config.subsystem == "http"
 local upstreams
 local healthcheck
+local healthcheck_manager
 
 local healthcheck_shdict_name = "upstream-healthcheck"
 if not is_http then
@@ -103,100 +105,17 @@ end
 _M.get_healthchecker_name = get_healthchecker_name
 
 
-local function create_checker(upstream)
-    local local_conf = config_local.local_conf()
-    if local_conf and local_conf.apisix and local_conf.apisix.disable_upstream_healthcheck then
-        core.log.info("healthchecker won't be created: disabled upstream healthcheck")
-        return nil
-    end
-    if healthcheck == nil then
-        healthcheck = require("resty.healthcheck")
-    end
-
-    local healthcheck_parent = upstream.parent
-    if healthcheck_parent.checker and healthcheck_parent.checker_upstream == upstream
-        and healthcheck_parent.checker_nodes_ver == upstream._nodes_ver then
-        return healthcheck_parent.checker
-    end
-
-    if upstream.is_creating_checker then
-        core.log.info("another request is creating new checker")
-        return nil
-    end
-    upstream.is_creating_checker = true
-
-    core.log.debug("events module used by the healthcheck: ", events.events_module,
-                    ", module name: ",events:get_healthcheck_events_modele())
-
-    local checker, err = healthcheck.new({
-        name = get_healthchecker_name(healthcheck_parent),
-        shm_name = healthcheck_shdict_name,
-        checks = upstream.checks,
-        -- the events.init_worker will be executed in the init_worker phase,
-        -- events.healthcheck_events_module is set
-        -- while the healthcheck object is executed in the http access phase,
-        -- so it can be used here
-        events_module = events:get_healthcheck_events_modele(),
-    })
-
-    if not checker then
-        core.log.error("fail to create healthcheck instance: ", err)
-        upstream.is_creating_checker = nil
-        return nil
-    end
-
-    if healthcheck_parent.checker then
-        local ok, err = pcall(core.config_util.cancel_clean_handler, healthcheck_parent,
-                                              healthcheck_parent.checker_idx, true)
-        if not ok then
-            core.log.error("cancel clean handler error: ", err)
-        end
-    end
-
-    core.log.info("create new checker: ", tostring(checker))
-
-    local host = upstream.checks and upstream.checks.active and upstream.checks.active.host
-    local port = upstream.checks and upstream.checks.active and upstream.checks.active.port
-    local up_hdr = upstream.pass_host == "rewrite" and upstream.upstream_host
-    local use_node_hdr = upstream.pass_host == "node" or nil
-    for _, node in ipairs(upstream.nodes) do
-        local host_hdr = up_hdr or (use_node_hdr and node.domain)
-        local ok, err = checker:add_target(node.host, port or node.port, host,
-                                           true, host_hdr)
-        if not ok then
-            core.log.error("failed to add new health check target: ", node.host, ":",
-                    port or node.port, " err: ", err)
-        end
-    end
-
-    local check_idx, err = core.config_util.add_clean_handler(healthcheck_parent, release_checker)
-    if not check_idx then
-        upstream.is_creating_checker = nil
-        checker:clear()
-        checker:stop()
-        core.log.error("failed to add clean handler, err:",
-            err, " healthcheck parent:", core.json.delay_encode(healthcheck_parent, true))
-
-        return nil
-    end
-
-    healthcheck_parent.checker = checker
-    healthcheck_parent.checker_upstream = upstream
-    healthcheck_parent.checker_nodes_ver = upstream._nodes_ver
-    healthcheck_parent.checker_idx = check_idx
-
-    upstream.is_creating_checker = nil
-
-    return checker
-end
-
-
 local function fetch_healthchecker(upstream)
     if not upstream.checks then
         return nil
     end
 
-    return create_checker(upstream)
+    local parent = upstream.parent
+    local resource_path = parent.key
+    local resource_ver = parent.modifiedIndex .. "#" .. tostring(upstream) .. "#" ..
+                         tostring(upstream._nodes_ver or '')
+
+    return healthcheck_manager.fetch_checker(resource_path, resource_ver, upstream)
 end
 
 
@@ -633,6 +552,8 @@ function _M.init_worker()
         error("failed to create etcd instance for fetching upstream: " .. err)
         return
     end
+    healthcheck_manager = require("apisix.healthcheck_manager")
+    healthcheck_manager.init_worker()
 end
 
 


apisix/balancer.lua
```diff
@@ -213,7 +214,9 @@ local function pick_server(route, ctx)
 
     local version = ctx.upstream_version
     local key = ctx.upstream_key
-    local checker = ctx.up_checker
```
nic-6443: why change this? I think for single request, we only need fetch health checker once, and reuse same checker during this request by ctx variable.

Revolyssup: checker is no longer attached to request context and fetched by resource_key and version

Revolyssup: fixed

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+end
+
+
+local function create_checker(up_conf)
+    local local_conf = config_local.local_conf()
+    if local_conf and local_conf.apisix and local_conf.apisix.disable_upstream_healthcheck then
+        core.log.info("healthchecker won't be created: disabled upstream healthcheck")
+        return nil
+    end
+    core.log.warn("creating healthchecker for upstream: ", up_conf.resource_key)
```
Revolyssup: warn here to be replaced with info

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+                               resource_ver)
+                goto continue
+            end
+            local res_conf = fetch_latest_conf(resource_path)
+            if not res_conf then
+                goto continue
+            end
+            local upstream = res_conf.value.upstream or res_conf.value
+            local new_version = res_conf.modifiedIndex .. tostring(upstream._nodes_ver or '')
+            core.log.warn("checking waiting pool for resource: ", resource_path,
```
Revolyssup: ditto. to be changed to info

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+local function fetch_latest_conf(resource_path)
+    local resource_type, id
+    -- Handle both formats:
+    -- 1. /apisix/<resource_type>/<id>
+    -- 2. /<resource_type>/<id>
+    if resource_path:find("^/apisix/") then
+        resource_type, id = resource_path:match("^/apisix/([^/]+)/([^/]+)$")
+    else
+        resource_type, id = resource_path:match("^/([^/]+)/([^/]+)$")
+    end
```
nic-6443: why we still need to handle those two format?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+end
+
+local function find_in_working_pool(resource_path, resource_ver)
+    local checker = _M.working_pool[resource_path]
+    if not checker then
+        return nil  -- not found
+    end
+
+    if checker.version ~= resource_ver then
+        core.log.warn("version mismatch for resource: ", resource_path,
```
nic-6443: This is a log that would appear in a normal case, and there is no need to use the warn level.

apisix/healthcheck_manager.lua
nic-6443: sure, why still keep it as warn?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+                    " current version: ", checker.version, " requested version: ", resource_ver)
+        return nil  -- version not match
+    end
+    return checker
+end
+
+local timer_create_checker_running = false
```
nic-6443: this variable only be used in `init_worker`, maybe we should move it to before `init_worker` function to easy find it's usage.

apisix/upstream.lua
```diff
@@ -569,9 +448,9 @@ local function filter_upstream(value, parent)
     if not value then
         return
     end
-
-    value.parent = parent
-
+    value.resource_key = ((parent and parent.key) or value.key)
```
nic-6443: Are you sure `or value.key` is necessary?
https://github.com/apache/apisix/blob/81caea1e4b02809e1202bcea20529948d8aeac4f/apisix/upstream.lua#L627
no key field in this `upstream` parameter?

nic-6443: Try to keep the code consistent and reduce writing separate code for different cases.

apisix/upstream.lua
```diff
@@ -377,10 +256,10 @@ function _M.set_by_route(route, api_ctx)
     if not ok then
         return 503, err
     end
-
-    local checker = fetch_healthchecker(up_conf)
+    local resource_version = up_conf.resource_version and (up_conf.resource_version ..
```
nic-6443: `up_conf.resource_version` always exists, I think this check is redundant. and you don't check it in above code too.

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,288 @@
...
+        --- remove from working pool if resource doesn't exist
+        local res_conf = fetch_latest_conf(resource_path)
+        if not res_conf then
+            item.checker:delayed_clear(10)
+            item.checker:stop()
+            core.log.info("try to release checker: ", tostring(item.checker))
+            _M.working_pool[resource_path] = nil
+            goto continue
+        end
+        local current_ver = res_conf.modifiedIndex ..  tostring(res_conf.value._nodes_ver or '')
```
nic-6443: `res_conf.modifiedIndex .. tostring(res_conf.value._nodes_ver or '')` this line of code duplicate in at least four places, can we provide a function like `upstream_version()` to wrap it?

apisix/healthcheck_manager.lua
Revolyssup: We will have to handle these two cases anyways either here or in filter_upstream to construct a resource key by stripping the "/apisix". And even then inside this function we will have to break down /routes/1 into "routes" and 1 to first data = fetch_created_obj(routes) and then data:get(id). 
So I thought it will be better to have all the key construction logic and fetching logic at one place for easier to read.
Do you think there is advantage of moving this to filter_upstream?

Revolyssup: Either based on matching like this, or explicitly using resource_type in filter_upstream, we will have to handle these two cases. I picked here to have all key generation logic + fetching data from generated key logic at one place.

apisix/upstream.lua
Revolyssup: It actually was nil in one test case which caused a panic. Let me find that test case.

apisix/healthcheck_manager.lua
Revolyssup: done

apisix/healthcheck_manager.lua
Revolyssup: done

apisix/healthcheck_manager.lua
Revolyssup: done

apisix/healthcheck_manager.lua
Revolyssup: done

apisix/upstream.lua
Revolyssup: removed

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,297 @@
...
+    if not data then
+        core.log.error("failed to fetch configuration for type: ", key)
+        return nil
+    end
+    local resource = data:get(id)
+    if not resource then
+        -- this can happen if the resource was deleted
+        -- after the this function was called so we don't throw error
+        core.log.warn("resource not found: ", id, " in ", key)
+        return nil
```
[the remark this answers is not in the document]
Revolyssup: no need error is already logged in fetch_latest_conf

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,297 @@
...
+    for resource_path, item in pairs(working_snapshot) do
+        --- remove from working pool if resource doesn't exist
+        local res_conf = fetch_latest_conf(resource_path)
+        if not res_conf then
+            item.checker:delayed_clear(10)
+            item.checker:stop()
+            core.log.info("try to release checker: ", tostring(item.checker))
+            _M.working_pool[resource_path] = nil
+            goto continue
+        end
```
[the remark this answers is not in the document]
Revolyssup: made changes

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,298 @@
...
+end
+_M.get_healthchecker_name = get_healthchecker_name
+
+
+local function fetch_latest_conf(resource_path)
+    local resource_type, id
+    -- Handle both formats:
+    -- 1. /apisix/<resource_type>/<id>
+    -- 2. /<resource_type>/<id>
+    if resource_path:find("^/apisix/") then
```
nic-chen: key prefix may not be `/apisix`

Revolyssup: It will be either like /apisix/upstreams/1 or /upstreams/1
Can you give an example of something else?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,298 @@
...
+                core.log.info("resource: ", resource_path,
+                             " already in working pool with version: ",
+                               resource_ver)
+                goto continue
+            end
+            local res_conf = fetch_latest_conf(resource_path)
+            if not res_conf then
+                goto continue
+            end
+            local upstream = res_conf.value.upstream or res_conf.value
```
nic-chen, line 227: if not config `checks` for the upstream, we should also go to continue

apisix/balancer.lua
```diff
@@ -27,7 +27,7 @@ local get_last_failure = balancer.get_last_failure
 local set_timeouts     = balancer.set_timeouts
 local ngx_now          = ngx.now
 local str_byte         = string.byte
-
+local healthcheck_manager = require("apisix.healthcheck_manager")
```
membphis: image

apisix/balancer.lua
```diff
@@ -75,7 +75,8 @@ local function fetch_health_nodes(upstream, checker)
     local port = upstream.checks and upstream.checks.active and upstream.checks.active.port
     local up_nodes = core.table.new(0, #nodes)
     for _, node in ipairs(nodes) do
-        local ok, err = checker:get_target_status(node.host, port or node.port, host)
+        local ok, err = healthcheck_manager.fetch_node_status(checker,
```
membphis, line 78: we can cache `healthcheck_manager.fetch_node_status`, a short local function

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,298 @@
...
+local exiting      = ngx.worker.exiting
+local pairs    = pairs
+local tostring = tostring
+local core = require("apisix.core")
+local config_local   = require("apisix.core.config_local")
+local healthcheck
+local events = require("apisix.events")
+local tab_clone = core.table.clone
+local timer_every = ngx.timer.every
+local _M = {
```
membphis: it is not safe to export `working_pool` and `waiting_pool`.

We will never allow the "working_pool" or "waiting_pool" to be destroyed.

they only can be use in this lua module

apisix/healthcheck_manager.lua
membphis: this is the prefix:

image

apisix/healthcheck_manager.lua
Revolyssup: If an upstream config is updated to have `checks` removed, then we want the further logic to process as this function deletes the previous checker. If we continue, then after `checks` is removed, the checker will not be removed.

Revolyssup: I updated the create_checker function to return early if up_conf.checks is nil. You can check again. I think that's a better point to detect and continue as the previous checker would be deleted by then.

apisix/balancer.lua
Revolyssup: do you mean using the lrucache with `checker` as key?

membphis:
it looks good to me now, but we have to wait for more time, for more comments, it is a big change

BTW, the PR title should be `refactor` ^_^

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,297 @@
...
+            core.log.info("checking waiting pool for resource: ", resource_path,
+                    " current version: ", new_version, " requested version: ", resource_ver)
+            if resource_ver ~= new_version then
+                goto continue
+            end
+
+            -- if a checker exists then delete it before creating a new one
+            local existing_checker = _M.working_pool[resource_path]
+            if existing_checker then
+                existing_checker.checker:delayed_clear(10)
```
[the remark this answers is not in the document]
nic-chen: @Revolyssup

nic-chen:
Most of our tests are for scenarios which the config_provider is etcd. Could you add health check tests for scenarios which the config_provider is yaml or json?

Revolyssup:
> Most of our tests are for scenarios which the config_provider is etcd. Could you add health check tests for scenarios which the config_provider is yaml or json?

A lot of existing tests here already use yaml config_provider. See https://github.com/apache/apisix/pull/12426/files#diff-d89c41bb4b1cc7c97936090edd39c05e1fcbc35c94039a16dbb392515d4f38d3R34

nic-6443 (dismissed):

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,313 @@
...
+function _M.timer_working_pool_check()
+    if core.table.nkeys(working_pool) == 0 then
+        return
+    end
+
+    local working_snapshot = tab_clone(working_pool)
+    for resource_path, item in pairs(working_snapshot) do
+        --- remove from working pool if resource doesn't exist
+        local res_conf = fetch_latest_conf(resource_path)
+        if not res_conf or not res_conf.value then
```
membphis:
```
working_pool[resource_path] = nil
item.checker:delayed_clear(DELAYED_CLEAR_TIMEOUT)
item.checker:stop()
core.log.info("try to release checker: ", ...(a useful name))

Comment (membphis) [reply apisix/healthcheck_manager.lua]: we can print the `resource_path` and `version` in the error log

File: apisix/healthcheck_manager.lua [hunk 19]
```diff
@@ -0,0 +1,313 @@
...
+
+
+function _M.timer_working_pool_check()
+    if core.table.nkeys(working_pool) == 0 then
+        return
+    end
+
+    local working_snapshot = tab_clone(working_pool)
+    for resource_path, item in pairs(working_snapshot) do
+        --- remove from working pool if resource doesn't exist
```
membphis, line 266:
```lua
--- remove from working pool if resource doesn't exist
    local need_free
    local res_conf = fetch_latest_conf(resource_path)
    if not res_conf or not res_conf.value then
        need_free = true
        goto continue
    end
    local current_ver = _M.upstream_version(res_conf.modifiedIndex,
                                            res_conf.value._nodes_ver)
    core.log.info("checking working pool for resource: ", resource_path,
                " current version: ", current_ver, " item version: ", item.version)
    if item.version ~= current_ver then
        need_free = true
        goto continue
    end
    
    ::continue::
    if need_free then
        working_pool[resource_path] = nil           -- delete the ref ASAP
        item.checker.dead = true                          -- mark it ASAP
        item.checker:delayed_clear(DELAYED_CLEAR_TIMEOUT)
        item.checker:stop()
        core.log.info("try to release checker: ", ... ...)   -- need a useful information, eg: resource_path and name
    end

Comment (membphis) [reply apisix/healthcheck_manager.lua:266]: the free process is same, we can merge them

File: apisix/healthcheck_manager.lua [hunk 20]
```diff
@@ -0,0 +1,313 @@
...
+    local timer_create_checker_running = false
+    local timer_working_pool_check_running = false
+    timer_every(1, function ()
+        if not exiting() then
+            if timer_create_checker_running then
+                core.log.warn("timer_create_checker is already running, skipping this iteration")
+                return
+            end
+            timer_create_checker_running = true
+            pcall(_M.timer_create_checker)
```
membphis: we always should capture the error message, and print them to error log

they are useful if the code run fail

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,313 @@
...
+        end
+    end)
+    timer_every(1, function ()
+        if not exiting() then
+            if timer_working_pool_check_running then
+                core.log.warn("timer_working_pool_check is already running skipping iteration")
+                return
+            end
+            timer_working_pool_check_running = true
+            pcall(_M.timer_working_pool_check)
```
membphis: ditto

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,313 @@
...
+            if resource_ver ~= new_version then
+                goto continue
+            end
+
+            -- if a checker exists then delete it before creating a new one
+            local existing_checker = working_pool[resource_path]
+            if existing_checker then
+                existing_checker.checker:delayed_clear(10)
+                existing_checker.checker:stop()
+                core.log.info("releasing existing checker: ", tostring(existing_checker.checker))
```
membphis: need more useful information, pls fix those similar issues together

apisix/control/v1.lua
```diff
@@ -20,6 +20,7 @@ local plugin = require("apisix.plugin")
 local get_routes = require("apisix.router").http_routes
 local get_services = require("apisix.http.service").services
 local upstream_mod = require("apisix.upstream")
+local healthcheck_manager = require("apisix.healthcheck_manager")
```
bzp2010, line 23: It might make sense to put it in the same directory as the batch processor manager. The root directory isn't the right place for it.

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,322 @@
...
+    local data = core.config.fetch_created_obj(key)
+    if not data then
+        core.log.error("failed to fetch configuration for type: ", key)
+        return nil
+    end
+    local resource = data:get(id)
+    if not resource then
+        -- this can happen if the resource was deleted
+        -- after the this function was called so we don't throw error
+        core.log.warn("resource not found: ", id, " in ", key)
```
membphis: print more information, help the OPS to confirm the reason

`this can happen if the resource was deleted`

apisix/healthcheck_manager.lua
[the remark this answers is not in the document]
membphis: pls remove this magic number

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,322 @@
...
+function _M.timer_working_pool_check()
+    if core.table.nkeys(working_pool) == 0 then
+        return
+    end
+
+    local working_snapshot = tab_clone(working_pool)
+    for resource_path, item in pairs(working_snapshot) do
+        --- remove from working pool if resource doesn't exist
+        local res_conf = fetch_latest_conf(resource_path)
+        local need_free = true
```
membphis: another name: `need_destroy`, seems better?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,322 @@
...
+    local timer_create_checker_running = false
+    local timer_working_pool_check_running = false
+    timer_every(1, function ()
+        if not exiting() then
+            if timer_create_checker_running then
+                core.log.warn("timer_create_checker is already running, skipping this iteration")
+                return
+            end
+            timer_create_checker_running = true
+            local ok, err = pcall(_M.timer_create_checker)
```
membphis: `_M.timer_create_checker` vs `local function timer_create_checker`

I think we do not need to export `timer_create_checker`, local function should be enough

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,322 @@
...
+        end
+    end)
+    timer_every(1, function ()
+        if not exiting() then
+            if timer_working_pool_check_running then
+                core.log.warn("timer_working_pool_check is already running skipping iteration")
+                return
+            end
+            timer_working_pool_check_running = true
+            local ok, err = pcall(_M.timer_working_pool_check)
```
membphis: same issue as before

membphis (dismissed):
LGTM, only some minor issues

apisix/control/v1.lua
```diff
@@ -66,14 +67,13 @@ function _M.schema()
     return 200, schema
 end
 
-
```
bzp2010, line 69: Just a reminder, please try to avoid pointless formatting changes, this noise will affect review efficiency.

membphis (approved):
LGTM ^_^

nic-6443 (approved):

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+                        resource_path, " and version : ", item.version)
+        end
+    end
+end
+
+function _M.init_worker()
+    local timer_create_checker_running = false
+    local timer_working_pool_check_running = false
+    timer_every(1, function ()
+        if not exiting() then
```
bzp2010, line 294: Why not use the `premature` parameter on the timer callback instead of `exiting()`?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+            timer_create_checker_running = true
+            local ok, err = pcall(timer_create_checker)
+            if not ok then
+                core.log.error("failed to run timer_create_checker: ", err)
+            end
+            timer_create_checker_running = false
+        end
+    end)
+    timer_every(1, function ()
+        if not exiting() then
```
bzp2010, line 308: ditto

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+
+local _M = {}
+local working_pool = {}     -- resource_path -> {version = ver, checker = checker}
+local waiting_pool = {}      -- resource_path -> resource_ver
+
+local DELAYED_CLEAR_TIMEOUT = 10
+local healthcheck_shdict_name = "upstream-healthcheck"
+local is_http = ngx.config.subsystem == "http"
+if not is_http then
+    healthcheck_shdict_name = healthcheck_shdict_name .. "-" .. ngx.config.subsystem
```
bzp2010, line 39: Once again, is a separate shdict for stream a must? Can't it share data with http?

apisix/healthcheck_manager.lua
```diff
@@ -0,0 +1,323 @@
...
+local exiting      = ngx.worker.exiting
+local pairs    = pairs
+local tostring = tostring
+local core = require("apisix.core")
+local config_local   = require("apisix.core.config_local")
+local healthcheck
+local events = require("apisix.events")
+local tab_clone = core.table.clone
+local timer_every = ngx.timer.every
+local string_sub     = string.sub
```
bzp2010, line 29: Please align their equals signs.

```
local xxx   = x
local x     = x
local xxxxx = x
```

apisix/upstream.lua
```diff
@@ -569,9 +448,9 @@ local function filter_upstream(value, parent)
     if not value then
         return
     end
-
-    value.parent = parent
-
+    value.resource_key = parent and parent.key
+    value.resource_version = ((parent and parent.modifiedIndex) or value.modifiedIndex)
+    value.resource_id = ((parent and parent.value.id) or value.id)
```
bzp2010, line 453: I'm concerned if this is creating some more special case where: I can get upstream metadata via `res.resource_id/res.resource_version` but can't do the same with the route.
This may reduce readability and maintainability, and developers need to understand why `upstream.resource_id` is available but `route.resource_id` and `consumer.resource_id` are not.

Revolyssup: We already have this problem of special case with maintaining `.parent` everywhere. According to me this `.parent` is much less maintainable and spread at different places in code.

apisix/healthcheck_manager.lua
Revolyssup: removed

nic-chen (approved):

PR #12435, opened 2025-07-16
fix(forward-auth): extra_headers not resolving variable on $post_arg.

Revolyssup:
The existing regex in util.resolve_var doesn't handle all the variable cases properly. For eg: when passed with $post_arg.xyz, it fails to resolve the variable. So regex is being modified to support the aforementioned key.

This was not caught previously because the test cases didn't assert the values properly.

### Checklist

### DIFF
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -21,6 +21,7 @@ local http     = require("resty.http")
 local pairs    = pairs
 local type     = type
 local tostring = tostring
+local sub_str  = string.sub
 
 local schema = {
     type = "object",
@@ -124,13 +125,19 @@ function _M.access(conf, ctx)
             if type(value) == "number" then
                 value = tostring(value)
             end
-            local resolve_value, err = core.utils.resolve_var(value, ctx.var)
-            if not err then
-                auth_headers[header] = resolve_value
-            end
-            if err then
-                core.log.error("failed to resolve variable in extra header '",
-                                header, "': ",value,": ",err)
+            if core.string.has_prefix(value, "$") then
+                -- If the value starts with a dollar sign, treat it as a variable
+                -- and resolve it from the context.
+                local sub_str_len = #value - 1  -- Exclude the dollar sign
+                if sub_str_len > 0 then
+                    local trimmed_value = sub_str(value, 2)  -- Remove the dollar sign
+                    local resolve_value = ctx.var[trimmed_value]
+                    if resolve_value then
+                        auth_headers[header] = resolve_value
+                    end
+                end
+            else
+                auth_headers[header] = value
             end
         end
     end
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -21,7 +21,6 @@ local http     = require("resty.http")
 local pairs    = pairs
 local type     = type
 local tostring = tostring
-local sub_str  = string.sub
 
 local schema = {
     type = "object",
@@ -125,19 +124,13 @@ function _M.access(conf, ctx)
             if type(value) == "number" then
                 value = tostring(value)
             end
-            if core.string.has_prefix(value, "$") then
-                -- If the value starts with a dollar sign, treat it as a variable
-                -- and resolve it from the context.
-                local sub_str_len = #value - 1  -- Exclude the dollar sign
-                if sub_str_len > 0 then
-                    local trimmed_value = sub_str(value, 2)  -- Remove the dollar sign
-                    local resolve_value = ctx.var[trimmed_value]
-                    if resolve_value then
-                        auth_headers[header] = resolve_value
-                    end
-                end
-            else
-                auth_headers[header] = value
+            local resolve_value, err = core.utils.resolve_var(value, ctx.var)
+            if not err then
+                auth_headers[header] = resolve_value
+            end
+            if err then
+                core.log.error("failed to resolve variable in extra header '",
+                                header, "': ",value,": ",err)
             end
         end
     end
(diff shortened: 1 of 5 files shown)


docs/en/latest/plugins/forward-auth.md
```diff
@@ -227,8 +228,8 @@ curl -X PUT 'http://127.0.0.1:9180/apisix/admin/routes/1' \
 Send a POST request with the `tenant_id` header:
 
 ```shell
-curl -i http://127.0.0.1:9080/post -X POST -d '{
-   "tenant_id": 123
+curl -i http://127.0.0.1:9080/post -H "Content-Type: application/json" -X POST -d '{
+   "tenant_id": "123"
```
kayx23, line 232: If user accidentally use the number instead of the string, they will run into a 500 error.

image

It is not entirely clear based on the current example why (as a user) I have to use string. I think this error should be caught, but I'll leave the judgement to you.

docs/en/latest/plugins/forward-auth.md
```diff
@@ -187,11 +187,12 @@ curl -X PUT 'http://127.0.0.1:9180/apisix/admin/routes/auth' \
             "functions": [
                 "return function(conf, ctx)
                  local core = require(\"apisix.core\")
-                 if core.request.header(ctx, \"tenant_id\") then
+                 local tenant_id = core.request.header(ctx, \"tenant_id\")
+                 if tenant_id == \"123\" then
                      core.response.set_header(\"X-User-ID\", \"i-am-an-user\");
```
kayx23, line 192: Maybe remove this line to avoid any confusion? Since users will not actually observe this response header in this example.

docs/en/latest/plugins/forward-auth.md
```diff
@@ -173,7 +173,7 @@ Location: http://example.com/auth
 When the decision is to be made on the basis of POST body, then it is recommended to use `$post_arg.*` with `extra_headers` field and make the decision on Authorization service on basis of headers rather than using POST `request_method` to pass the entire request body to Authorization service.
 :::
 
-Create a serverless function on the `/auth` route that checks for the presence of the `tenant_id` header. If present, the route responds with HTTP 200 and sets the `X-User-ID` header to a fixed value `i-am-an-user`. If `tenant_id` is missing, it returns HTTP 400 with an error message.
+Create a serverless function on the `/auth` route that checks for the presence of the `tenant_id` header and confirms its value. If present, the route responds with HTTP 200 and sets the `X-User-ID` header to a fixed value `i-am-an-user`. If `tenant_id` is missing, it returns HTTP 400 with an error message.
```
kayx23: If the above is change, this paragraph should be updated as:

```suggestion
Create a serverless function on the `/auth` route that checks for the presence of the `tenant_id` header and confirms its value. If present, the route responds with HTTP 200.. If `tenant_id` is missing, it returns HTTP 400 with an error message.
```

kayx23 (commented):
Other doc changes LGTM

docs/en/latest/plugins/forward-auth.md
Revolyssup: This is required to pass/deny the request. Users don't need to care about it. This is forward-auth server simulated.

nic-6443 (dismissed):

apisix/plugins/forward-auth.lua
```diff
@@ -124,13 +125,19 @@ function _M.access(conf, ctx)
...
+                local sub_str_len = #value - 1  -- Exclude the dollar sign
+                if sub_str_len > 0 then
+                    local trimmed_value = sub_str(value, 2)  -- Remove the dollar sign
+                    local resolve_value = ctx.var[trimmed_value]
+                    if resolve_value then
+                        auth_headers[header] = resolve_value
+                    end
+                end
+            else
+                auth_headers[header] = value
```
bzp2010: I disagree with this.

The `core.utils.resolve_var` not only resolves `$var_1` but also templates of strings with mixed formatting like `fixed${var_1}fixed${var_2}`, where `$` does not need to be an initial letter.
Your snippet asserts this, which subverts the basis of any previous implementation of variable resolution we've had.

bzp2010: If `$post_arg.` is not supported, change the format of `$post_arg.`. This change has not been included in any of the new releases, but variable parsing support is widely used.
Not to mention that there is no precedent for using symbols other than underscores `_` in APISIX and nginx variables.

https://nginx.org/en/docs/varindex.html

docs/en/latest/plugins/forward-auth.md
bzp2010, line 1: I don't see the need for these modifications. It doesn't really provide an example of `$post_arg.xx`.
We don't need to mention that specifically. These are details of the APISIX variable system, and they can be used anywhere if it is supported there.
So these documentation changes seem to be able to be reverted.

Revolyssup: There are some issues in the existing doc here llke:
1. Addition of ` -H "Content-Type: application/json" ` in example without which the post_arg wont be passed and users wont the result mentioned in doc.
2. Since we claim to pass the header, it is good to modify the serverless function in doc that actually checks the forwarded header. So that users can be assured that headers are passed. Previous to the current change, only it's existence is checked and value not verified.

docs/zh/latest/plugins/forward-auth.md
```diff
@@ -202,7 +203,7 @@ curl -X PUT 'http://127.0.0.1:9180/apisix/admin/routes/auth' \
 }'
 ```
 
-创建一个接受 POST 请求的路由，并使用 `forward-auth` 插件通过请求中的 `tenant_id` 调用身份验证端点。只有当身份验证检查返回 200 时，请求才会转发到上游服务。
+创建一个接受 POST 请求的路由，并使用“forward-auth”插件通过请求中的“tenant_id”调用身份验证端点。仅当身份验证检查返回 200 时，请求才会转发到上游服务。
```
bzp2010: Some of the ` have been accidentally modified, and they should be kept.

Revolyssup: fixed

nic-6443 (approved):

PR #12445, opened 2025-07-19
feat(nacos): add metadata filtering support to nacos discovery

beardnick:
### Description

Add metadata filtering support to nacos discovery.

#### Which issue(s) this PR fixes:

Fixes https://github.com/apache/apisix/issues/12392

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #12448, opened 2025-07-20
feat(consul): filter nodes in upstream with metadata

### LINKED ISSUE #12464: feat: I hope we could filter nodes via metadata when config upstream
### Description

As a user, I hope that for the same service name, different instances can be filtered in different routes, upstreams, and services.

## Q. Why do we need metadata filtering?

A. For the same service, we will have different metadata for management. For example, when the recommendation system model changes, the model version will be modified and traffic switching and ab experiments will be performed through client coloring.

## Q. Why not use different service names?

A. We will create multiple environments based on the same service template. They only have some different metadata, such as env or lane, and model version. As the metadata dimension increases, the service names we need to locate specific coordinates will increase nonlinearly and client traffic is not colored so we have no way to use multi-routes.

## Q. What upstream types need to be supported?

A. Including Consul, Eureka, Nacos and other registration center types, as well as the registration center selection that may be added in the future or within the enterprise can directly inherit this capability

## Q. Any other context?

A1. I'm sorry I broke the process. I have implemented filtering in the upstream dimension internally and submitted PR https://github.com/apache/apisix/pull/12448. During the communication process, I conservatively changed to only support Consul's implementation, but following the DRY principle, I think it may be suitable to put it in the upstream instead of implementing it once for each discovery.

A2. Currently, someone in Nacos is also implementing the same function https://github.com/apache/apisix/issues/12392 https://github.com/apache/apisix/pull/12445, but only supports single value matching. However, we hope to support multi-value matching. This is because user requests often do not have any tags, so if we want to allow multiple instances with different metadata to provide services together, we must register a special metadata to group them, but this grouping cannot be changed dynamically.

Real scenarios I have experienced:

In order to control the explosion radius, we deploy services in multiple data centers, multiple availability zones, and even multiple regions. The gateway, as a global load balancer, will forward requests evenly (possibly with a geographic location bias). Ideally, each availability zone is the same and deployed in a unit, but in reality there will always be various accidents that cause a single data center to be unavailable. Due to various human errors, the probability is much greater than "Snakes in a Facebook Datacenter" or "excavators digging up cables" (sounds absurd, but these two also happen frequently). At this time, there needs to be a way to directly remove this data center from the global load balancer to ensure SLA, and it is obvious that removing this data center from the list is the simplest and most efficient way.

jizhuozhi:
## Description

This PR introduces **metadata-based node filtering** for consul discovery, supporting Consul service discovery based upstreams

### Motivation

Currently, APISIX selects upstream nodes based on service name from discovery without additional filtering logic. In real-world scenarios like canary release or swimlane routing, users often tag backend instances with custom metadata (e.g., `version`, `env`, `lane`, `dc`, and etc) and expect the gateway to route only to specific subsets.

This change allows users to define a `metadata_match` field in `discovery_args` configuration, which filters nodes before load balancing based on their metadata values.

### Changes

* **Consul discovery**: 
- Include `Service.Meta` in the node definition and respect its `weight` if available (aligned with Eureka).
- Filter with `discovery_args` when fetch upstream nodes.

* **Tests**: Add test cases to cover both:

  * discovery-based upstream (Consul) with `metadata_match`

### Example Usage

```yaml
upstream:
  type: roundrobin
  scheme: http
  discovery_type: eureka
  discovery_args:
    metadata_match:
      lane:
      - prod
      - canary
      dc:
      - us-east-1
      - us-east-2
```

Only nodes with `metadata.lane in [prod, canary] and metadata.dc in [us-east-1, us-east-2]` will be used for load balancing.

---

## Fixes

Fixes #12464 

---

## Checklist

* [x] I have explained the need for this PR and the problem it solves
* [x] I have explained the changes or the new features added to this PR
* [x] I have added tests corresponding to this change
* [x] I have updated the documentation to reflect this change
* [x] I have verified that this change is backward compatible (no breaking changes)


Baoyuantop:
Hi @jizhuozhi, thanks for your contribution.

I think it is useful for node filtering of Consul discovery. But I don't understand static upstream filtering. It seems that I need to mark metadata for each node in the upstream object, and then use metadata_match to configure filtering? Because each node is manually defined, if it is not needed, can I just add or delete the node?

jizhuozhi:
> Hi @jizhuozhi, thanks for your contribution.
> 
> I think it is useful for node filtering of Consul discovery. But I don't understand static upstream filtering. It seems that I need to mark metadata for each node in the upstream object, and then use metadata_match to configure filtering? Because each node is manually defined, if it is not needed, can I just add or delete the node?

Here is a unified approach: I only determine whether there is a filtering rule, without distinguishing whether it is a service discovery or static list.

However, according to my previous experience as a gateway administrator, there will be corresponding business developers who temporarily add rules for some debugging considerations but do not want to change the original instance list (for quick adjustment)

Baoyuantop:
> there will be corresponding business developers who temporarily add rules for some debugging considerations but do not want to change the original instance list 

Thanks for your reply. Could you please describe the scenario in detail? Why can't the existing methods solve this problem?

jizhuozhi:
> Thanks for your reply. Could you please describe the scenario in detail? Why can't the existing methods solve this problem?

It is not a production environment, but it is common in the testing and verification phase. We need to specify specific instances frequently (for example, to capture flame graphs for performance analysis), but we need to add other instances back after deleting them, so we need to specify instances by filtering.

In fact, we also matched according to the dynamic colored metadata when loading balancing but not predefine the routes, similar to https://github.com/kitex-contrib/loadbalance-tagging (I am also using lua to implement the same capabilities, but this is not within the scope of this discussion).

Baoyuantop:
> > Thanks for your reply. Could you please describe the scenario in detail? Why can't the existing methods solve this problem?
> 
> It is not a production environment, but it is common in the testing and verification phase. We need to specify specific instances frequently (for example, to capture flame graphs for performance analysis), but we need to add other instances back after deleting them, so we need to specify instances by filtering.
> 
> In fact, we also matched according to the dynamic colored metadata when loading balancing but not predefine the routes, similar to https://github.com/kitex-contrib/loadbalance-tagging (I am also using lua to implement the same capabilities, but this is not within the scope of this discussion).

I still have doubts about what is in Example Usage. Do you mean that if I need to adjust the nodes used, I don't need to change the content of the nodes list, but adjust metadata_match?
By the way, are you using static nodes or service discovery?

jizhuozhi:
> I still have doubts about what is in Example Usage. Do you mean that if I need to adjust the nodes used, I don't need to change the content of the nodes list, but adjust metadata_match?

Yes, just adjust metadata_match (but the discussion of this use case has been separated from this PR). For the runtime, it is a unified filtering rule for the service list that does not need to distinguish the source.

> By the way, are you using static nodes or service discovery?

We are currently using Consul on kubernetes. When I was working in another company a few years ago, we were using cloud virtual machines (or EC2). The cloud platform did not provide an API interface, but we used scripts to synchronize static instance lists at regular intervals. At this time, the static list was also a kind of dynamic discovery. (why not filter in the script? Because we were lazy:)

Baoyuantop:
Hi @jizhuozhi, There is currently no modification to the upstream schema, which means that the current modifications in the upstream only serve consul. Is it more appropriate to put all these logics into the consul module?

jizhuozhi:
Hello, @Baoyuantop, thanks for your reply. 

> Hi @jizhuozhi, There is currently no modification to the upstream schema, which means that the current modifications in the upstream only serve consul. Is it more appropriate to put all these logics into the consul module?

Not only consul, but also Eureka (which has already supported metadata in apisix) will inherit this function. In my forked dashboard has already supported configuring metadata_match for Consul and Eureka
https://github.com/jizhuozhi/apisix-dashboard/blob/9bd72c82e4fcbfa0d2bf34420280028c2ca853c8/web/src/components/Upstream/components/ServiceDiscovery.tsx#L30-L37

(The examples in the PR description are just examples, because this allows testing without the registry, and we don't need to care about service discovery or static nodes.)

And the current discovery package is responsible for pulling all instances, and the filtering in discovery is effective for all service names and upstreams, it means that I can only configure general filtering rules, but cannot configure differentiated matching for different routes and upstreams. This is our current online effect

image

jizhuozhi:
Hello @Baoyuantop , I see. Currently, upstream has passed the discovery args to nodes, so the loop can be closed in discovery. I will modify it.

```lua
        local new_nodes, err = dis.nodes(up_conf.service_name, up_conf.discovery_args)
        if not new_nodes then
            return HTTP_CODE_UPSTREAM_UNAVAILABLE, "no valid upstream node: " .. (err or "nil")
        end
```

jizhuozhi:
Hello @Baoyuantop , PTAL, thanks :)

We also have Spring Cloud applications with Eureka, but I have no time to write test case now, so I will create a new PR for Eureka later.

Baoyuantop:
Hi @jizhuozhi, we are still discussing whether to accept the feature of this PR, and we need to reach a consensus before we can start the review. Since there is no separate issue to discuss this issue, you need to clearly tell the maintainer what this feature does and why it is needed in the PR description (the current description already exists).

> The examples in the PR description are just examples, because this allows testing without the registry, and we don't need to care about service discovery or static nodes.

This is inappropriate and you need to replace it with an example from a real scenario. The current example will confuse other maintainers. In the latest changes, I see that you have cancelled the upstream related code. The current PR seems to focus on the filtering of consul services. Please update the PR description to reflect this. Thanks again for your contribution.

jizhuozhi:
> This is inappropriate and you need to replace it with an example from a real scenario. The current example will confuse other maintainers.

Thank you for your reminder, the PR content has been updated

Baoyuantop:
Please fix the failed CI.

docs/zh/latest/discovery/consul.md
SkyeYoung, line 1: Weird. 

Why are there so much content here that seem to be completely unrelated to this PR feature.

Welcome to submit these extra good content individually through a new PR, thx.

But please make sure that one PR focuses on doing only one thing as much as possible.

apisix/discovery/consul/init.lua
```diff
@@ -511,11 +514,17 @@ function _M.connect(premature, consul_server, retry_delay)
...
                         goto CONTINUE
                     end
 
-                    local svc_address, svc_port = node.Service.Address, node.Service.Port
+                    local svc_address = service.Address
+                    local svc_port = service.Port
+                    local metadata = service.Meta
+                    if type(metadata) ~= "table" then
+                       metadata = nil
+                    end
```
SkyeYoung: Why do we need this logic? Pls add some comments.

apisix/discovery/consul/init.lua
```diff
@@ -532,7 +541,8 @@ function _M.connect(premature, consul_server, retry_delay)
                         core.table.insert(nodes, {
                             host = svc_address,
                             port = tonumber(svc_port),
-                            weight = default_weight,
+                            weight = metadata and tonumber(metadata.weight) or default_weight,
```
SkyeYoung: Should we get the weight from the metadata? Is this a common practice? Should some comments be added?

apisix/utils/discovery.lua
```diff
@@ -0,0 +1,55 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+local core   = require("apisix.core")
+local ipairs = ipairs
+local pairs  = pairs
+
+local _M = {}
+
+local function do_metadata_match(inst, filters)
```
SkyeYoung: What does "inst" mean? 

Looking at the logic of how you use this function, I think "node" or "upstream_node" would be better than the current meaning.

Regardless, please adjust the variable name or add some comments.

apisix/utils/discovery.lua
```diff
@@ -0,0 +1,55 @@
...
+    local metadata = inst.metadata or {}
```
SkyeYoung: Should we return early if `inst.metadata` does not exist?

The current logic seems to still iterate once when `metadata={}`:

```lua
for _, value in ipairs(values) do
    ...
end

File: t/discovery/consul2.t [hunk 5]
```diff
@@ -145,7 +145,7 @@ location /consul3 {
     "PUT /consul2/deregister/service_a2",
     "PUT /consul3/deregister/service_a1",
     "PUT /consul3/deregister/service_a2",
-    "PUT /consul1/register\n" . "{\"ID\":\"service_a1\",\"Name\":\"service_a\",\"Tags\":[\"primary\",\"v1\"],\"Address\":\"127.0.0.1\",\"Port\":30511,\"Meta\":{\"service_a_version\":\"4.0\"},\"EnableTagOverride\":false,\"Weights\":{\"Passing\":10,\"Warning\":1}}",
+    "PUT /consul1/register\n" . "{\"ID\":\"service_a1\",\"Name\":\"service_a\",\"Tags\":[\"primary\",\"v1\"],\"Address\":\"127.0.0.1\",\"Port\":30511,\"Meta\":{\"service_a_version\":\"4.0\", \"weight\":\"100\"},\"EnableTagOverride\":false,\"Weights\":{\"Passing\":10,\"Warning\":1}}",
```
SkyeYoung: Based on [my questions regarding the changes related to `weight`](https://github.com/apache/apisix/pull/12448#discussion_r2244553007), I think the testing for this part needs to be reconsidered.

t/discovery/consul_dump.t
```diff
@@ -95,7 +95,7 @@ discovery:
 --- request
 GET /t
 --- response_body
-{"service_a":[{"host":"127.0.0.1","port":30511,"weight":1}],"service_b":[{"host":"127.0.0.1","port":8002,"weight":1}]}
+{"service_a":[{"host":"127.0.0.1","metadata":{"service_a_version":"4.0"},"port":30511,"weight":1}],"service_b":[{"host":"127.0.0.1","metadata":{"service_b_version":"4.1"},"port":8002,"weight":1}]}
```
SkyeYoung: Why does it affect this place? I don't think this test should be changed

apisix/discovery/consul/init.lua
SkyeYoung: IMO, even though this is a common practice, I think we should initiate another discussion and submit a separate PR to adjust it. Because this doesn't seem to be related to the PR title (if the title represents the topic of this PR)

SkyeYoung:
@jizhuozhi ping

apisix/utils/discovery.lua
jizhuozhi: fixed

apisix/discovery/consul/init.lua
jizhuozhi: weight has been deleted from this PR and will submit in another, fixed

t/discovery/consul_dump.t
jizhuozhi: Since we are now relying on metadata, we need to persist it when persisting.

SkyeYoung:
@jizhuozhi 

Please try not to use force-push. It forces me to review the entire PR from the beginning, rather than just the parts you've changed since my last review.

Thanks 😸

t/discovery/consul_dump.t
SkyeYoung: @jizhuozhi Ok. Another question is, if this is added, do we now lack a test case for the situation where there is "no metadata"?

jizhuozhi:
> @jizhuozhi
> 
> Please try not to use force-push. It forces me to review the entire PR from the beginning, rather than just the parts you've changed since my last review.
> 
> Thanks 😸

Thanks for your reminder, I will pay attention to it. The previous problem here was because of cross-environment submission (I testing in EC2 and coding in office computer), resulting in multiple different and invalid submitters. This situation will not occur in the future.

t/discovery/consul_dump.t
jizhuozhi: I will add it :)

apisix/utils/discovery.lua
```diff
@@ -0,0 +1,66 @@
...
+local core   = require("apisix.core")
+local ipairs = ipairs
+local pairs  = pairs
+
+local _M = {}
+
+local function do_metadata_match(node, metadata_match)
+    local metadata = node.metadata
+    -- because metadata_match has already been checked in nodes_metadata_match,
+    -- there is at least one role, if there is no metadata in node, it's mast not matched
```
SkyeYoung:
```suggestion
    -- there is at least one role, if there is no metadata in node, it's must not matched
```

SkyeYoung (commented):
Others LGTM

Pls complete these two things, and then we can ask other maintainers to review it:

1. add no metadata test case
2. fix ci

SkyeYoung (dismissed):
LGTM

docs/en/latest/discovery/consul.md
```diff
@@ -232,6 +232,61 @@ $ curl http://127.0.0.1:9180/apisix/admin/stream_routes/1 -H "X-API-KEY: $admin_
...
+```shell
+$ curl http://127.0.0.1:9180/apisix/admin/routes/5 -H "X-API-KEY: $admin_key" -X PUT -i -d '
+{
+    "uri": "/nacosWithMetadata/*",
+    "upstream": {
+        "service_name": "APISIX-NACOS",
+        "type": "roundrobin",
+        "discovery_type": "nacos",
+        "discovery_args": {
+          "metadata": {
```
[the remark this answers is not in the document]
SkyeYoung: @jizhuozhi Please fix the document.

jizhuozhi:
Hello @SkyeYoung. Please help try again, it seems to be unrelated to this change

SkyeYoung (dismissed):

apisix/discovery/consul/init.lua
```diff
@@ -511,11 +515,19 @@ function _M.connect(premature, consul_server, retry_delay)
...
                         goto CONTINUE
                     end
 
-                    local svc_address, svc_port = node.Service.Address, node.Service.Port
+                    local svc_address = service.Address
+                    local svc_port = service.Port
+                    local metadata = service.Meta
+                    -- ensure that metadata is an accessible table,
+                    -- avoid userdata likes `null` returned by cjson
+                    if type(metadata) ~= "table" then
```
membphis: pls take a look, skip current service if the `metadat` is invalid

```
if type(metadata) == "cdata" then
   metadata = nil
elseif type(metadata) ~= "table" then
   core.log.error("wrong meta data, ...", ...)
   goto CONTINUE
end

Comment (jizhuozhi) [reply apisix/discovery/consul/init.lua]: Hello @membphis. I will add cdata check and error log, but I am not sure whether it is appropriate to skip this node. 

If it is a container template release, it means that all the containers released in this batch will be skipped, and users only need to pay attention to whether it is valid when they need to use metadata. If it is skipped directly, the service quality will be affected. So I tend to keep these nodes, print logs and leave them empty.

**Comment by SkyeYoung** [comment]
@jizhuozhi Please fix the errors reported in the CI.

**Review by SkyeYoung — DISMISSED** [review]
LGTM

File: t/discovery/consul.t [hunk 10]
```diff
@@ -781,3 +781,71 @@ location /sleep {
     qr//
 ]
 --- ignore_error_log
+
+
+
+=== TEST 16: test metadata_match with consul discovery
```
SkyeYoung: @jizhuozhi I've found a problem. If I understand correctly, the current test only checks routing to `v2` and `v3`, but it's missing the cases after adding or removing `metadata_match`. For example, adding `- v4` and removing `- v2` means the request should only be routed to those marked as `v3` or `v4`.

Maybe I missed something, please trouble you to answer or add the cases.

SkyeYoung (commented):
At the same time, it seems that some explanations are missing, such as which tests are conducted without metadata. Including these comments would better assist others in reviewing your code. Thx.

t/discovery/consul.t
jizhuozhi: Yes, after adding `- v4` and removing `- v2`, the request should only be routed to those marked as `v3` or `v4`. 

It seems to be equivalent to the current test case, as `discovery_args` is atomic updated, however I will add new case with adding `- v4` and removing `- v2` means the request should only be routed to those marked as `v3` or `v4` :)

SkyeYoung: @jizhuozhi The intention is to prove that after the configuration is updated, the data can correctly point to their marks.
Perhaps there's a better way to describe this situation, what are your thoughts?

SkyeYoung: @jizhuozhi Hello, may I ask if there's any progress?

apisix/schema_def.lua
```diff
@@ -488,6 +488,18 @@ local upstream_schema = {
                     description = "group name",
                     type = "string",
                 },
+                metadata_match = {
```
Baoyuantop: It is recommended to unify it as metadata.

docs/en/latest/discovery/consul.md
```diff
@@ -232,6 +232,61 @@ $ curl http://127.0.0.1:9180/apisix/admin/stream_routes/1 -H "X-API-KEY: $admin_
...
+$ curl http://127.0.0.1:9180/apisix/admin/routes/5 -H "X-API-KEY: $admin_key" -X PUT -i -d '
+{
+    "uri": "/consulWithMetadata/*",
+    "upstream": {
+        "service_name": "APISIX-CONSUL",
+        "type": "roundrobin",
+        "discovery_type": "consul",
+        "discovery_args": {
+          "metadata_match": {
+            "version": ["v1", "v2"]
```
Baoyuantop, line 259: I have a question here, is it necessary to set the value to an array? I checked the service meta format of consul https://developer.hashicorp.com/consul/docs/reference/service#meta, and I think setting the value to a string should be sufficient to reduce complexity.

jizhuozhi:
I'm sorry, due to continuous business trips and illness, there has been no progress. I will continue to work on it next week.

t/discovery/consul.t
jizhuozhi: Hello, I'm sorry that I haven't updated MR in the past two months due to project progress. Test cases for dynamically updating routes through etcd have been added.

docs/en/latest/discovery/consul.md
jizhuozhi: Hello @Baoyuantop , this feature has been running in our system since June, and we've benefited repeatedly from the ability to quickly recover from issues using an array rather than a string.

To limit the blast radius of new feature releases, we divide service releases into three phases: canary, prod-blue, and prod-green. Canary handles only a minimal amount of traffic to verify system stability, while prod-blue receives 50% of the online traffic to verify business issues. We also implement modular deployments for our online services, ensuring that all services are fully integrated across canary, prod-blue, and prod-green environments. If an issue is discovered in prod-blue, we quickly remove that phase from the gateway entry point and forward all traffic to prod-green.

With the current array-based configuration, if canary has some problem, we simply need to modify the apisix routing configuration:

From
```json
{
    "upstream": {
        "type": "roundrobin",
        "discovery_type": "consul",
        "discovery_args": {
            "metadata_match": {
                "env": ["canary", "prod-blue", "prod-green"]
            }
        }
    }
}
```

To
```json
{
    "upstream": {
        "type": "roundrobin",
        "discovery_type": "consul",
        "discovery_args": {
            "metadata_match": {
                "env": ["prod-blue", "prod-green"]
            }
        }
    }
}
```

But if we are using single string value, we must predefine a key likes `env: prod` to present this is a prod instance (even canary), and for staging deployment, we need define a secondary key likes `staging: canary/prod-blue/prod-green`. Normally, it would be easy to use `env: prod` (traffic is distributed to canary/prod-blue/prod-green based on instance weights), but when canary fails, we need to use a secondary key `staging` to avoid using canary. However, since only one value can be configured, I can only choose `prod-blue` or `prod-green`, which means I have to choose half of the instances to serve all the traffic, even if the other half are available.

From
```json
{
    "upstream": {
        "type": "roundrobin",
        "discovery_type": "consul",
        "discovery_args": {
            "metadata_match": {
                "env": "prod"
            }
        }
    }
}
```

To
```json
{
    "upstream": {
        "type": "roundrobin",
        "discovery_type": "consul",
        "discovery_args": {
            "metadata_match": {
                "env": "prod",
                "staging": "prod-blue" // or "prod-green", cannot both "prod-blue" and "prod-green"
            }
        }
    }
}
```

jizhuozhi: And as the dimensions increase, I have to define multiple routes by cross-product to achieve the functions originally implemented using arrays.

jizhuozhi:
Hello @Baoyuantop , the test case `t/plugin/mcp-bridge.t` seems not effected by this PR, maybe need retry the CI, thanks :)

apisix/schema_def.lua
jizhuozhi: Hello @Baoyuantop , because metadata is a noun. In our user communication, many people said that they mistakenly thought it was a label without side effects. Therefore, we used metadata_match with verb semantics.

Baoyuantop:
The main branch should have corresponding optimizations, you can merge it.

apisix/schema_def.lua
Baoyuantop: We need to maintain consistent naming conventions with other service discovery services.

Baoyuantop:
> Hello @Baoyuantop , the test case `t/plugin/mcp-bridge.t` seems not effected by this PR, maybe need retry the CI, thanks :)

The main branch has fixed this bug.

jizhuozhi:
I will see the reason and fix the testcases

silence8013:
This feature is fantastic, I can't wait!

apisix/utils/discovery.lua
```diff
@@ -0,0 +1,66 @@
...
+-- limitations under the License.
+--
+local core   = require("apisix.core")
+local ipairs = ipairs
+local pairs  = pairs
+
+local _M = {}
+
+local function do_metadata(node, metadata)
+    local metadata = node.metadata
```
Baoyuantop: What is the purpose of the second parameter, metadata?

Baoyuantop:
Hi @jizhuozhi, following up on the previous review comments. Let us know if you have an update. Thanks!

apisix/utils/discovery.lua
silence8013: It appears that the "rename metadata_match to metadata" commit partially overlaps with this one.

jizhuozhi: Yes, I will fix that caused by batch renaming variables

jizhuozhi:
I apologize that due to business demands, KPI pressure, and year-end performance reviews, my response to this request was delayed in Q4.  I can now dedicate my full attention to this PR.  I apologize again for the delay.

jizhuozhi:
The past failed job seems not caused by this PR change, I have merged latest master branch, please retry it, thanks.

moonming (changes requested):
Hi @jizhuozhi, thank you for the Consul metadata filtering support!

This is a highly requested feature (31 reviews!) for canary/lane routing with Consul-based service discovery. Combined with #12445 (Nacos metadata filtering), this covers the two most popular discovery backends.

**Current state**: The PR is marked as CONFLICTING. To move forward:
1. Please rebase on the latest master to resolve conflicts
2. Confirm all 31 review comments have been addressed
3. Consider coordinating with #12445 (Nacos) to ensure a consistent metadata filtering API across discovery backends

Once rebased, I'd like to prioritize this for merge. Thank you for the substantial community engagement! 🙏

PR #12465, opened 2025-07-28
feat: support limit-conn in workflow plugin

Revolyssup:
This PR adds support for using limit-conn plugin inside the workflow plugin extending the functionality of workflow plugin.
### Checklist


apisix/plugins/workflow.lua
```diff
@@ -101,10 +101,11 @@ local support_action = {
 }
 
 
-function _M.register(plugin_name, handler, check_schema)
+function _M.register(plugin_name, handler, check_schema, log_handler)
     support_action[plugin_name] = {
         handler        = handler,
-        check_schema   = check_schema
+        check_schema   = check_schema,
+        log_handler    = log_handler
```
nic-6443: I think it would be better to adjust the order and naming of this parameter to `check_schema`, `access_handler`, `log_handler`.

Revolyssup: Do you mean the order of arguments in the register function?

Revolyssup: I purposefully added log_handler to last to avoid making changes to other files keeping surface area of this PR low.

nic-6443: However, there are few places where the `register` function is used, only limit-count. I think it would be better to redesign the order of parameters.

apisix/plugins/workflow.lua
```diff
@@ -157,5 +158,22 @@ function _M.access(conf, ctx)
     end
 end
 
+function _M.log(conf, ctx)
+    for _, rule in ipairs(conf.rules) do
+        local match_result = true
+        if rule.case then
+            local expr, _ = expr.new(rule.case)
```
membphis: performance: we'd better to cache the `expr` and reuse it

membphis: same issue in `access` phase:

https://github.com/apache/apisix/pull/12465/files#diff-fdd97cfadd9b485076ddd89c4611747faf1f6c4cb6e833c07729a9e1aa70844eL149

membphis: we can store `expr` in `conf.rules`

Revolyssup: instead of caching expr in access phase. How about doing it in check_schema?

membphis: yes, it is acceptable

apisix/plugins/workflow.lua
```diff
@@ -116,10 +119,17 @@ function _M.check_schema(conf)
 
     for idx, rule in ipairs(conf.rules) do
          if rule.case then
-            local ok, err = expr.new(rule.case)
+            local expr, err = expr.new(rule.case)
             if not ok then
                 return false, "failed to validate the 'case' expression: " .. err
             end
+            setmetatable(rule, {
```
membphis: we should `getmetatable` first

nic-6443 (approved):

nic-6443 (dismissed):

apisix/plugins/workflow.lua
```diff
@@ -123,13 +123,13 @@ function _M.check_schema(conf)
             if not ok then
                 return false, "failed to validate the 'case' expression: " .. err
             end
-            setmetatable(rule, {
-                __index = function(_, key)
-                    if key == "__expr" then
-                        return expr
-                    end
+            local mt = getmetatable(rule) or {}
```
membphis: pls take a look at this style:

```
local my_table = {}

local meta = getmetatable(my_table)
if not meta then
    meta = {}
    setmetatable(my_table, meta) 
end

meta.a = "a"

File: apisix/plugins/workflow.lua [hunk 5]
```diff
@@ -116,10 +120,20 @@ function _M.check_schema(conf)
...
+            local expr, err = expr.new(rule.case)
             if not ok then
                 return false, "failed to validate the 'case' expression: " .. err
             end
+            local mt = getmetatable(rule)
+            if not mt then
+                mt = {}
+                setmetatable(rule, mt)
+            end
+            mt.__index = function (_, key)
```
membphis: this way is much better to read:

```
local my_table = {b="b"}

local meta = getmetatable(my_table)
if not meta then
    meta = {}
    meta.__index = meta
    setmetatable(my_table, meta)
end
meta.a = "a"

print(my_table.a)

**Review by membphis — DISMISSED** [review]

**Review by nic-6443 — DISMISSED** [review]

File: apisix/plugins/workflow.lua [hunk 6]
```diff
@@ -143,12 +154,14 @@ end
 
 
 function _M.access(conf, ctx)
-    for _, rule in ipairs(conf.rules) do
+    ctx._workflow_cache = ctx._workflow_cache or {}
+    for idx, rule in ipairs(conf.rules) do
         local match_result = true
         if rule.case then
-            local expr, _ = expr.new(rule.case)
+            local expr = rule.__expr
```
bzp2010, line 161: Need a check to assert that `__expr` must exist.

apisix/plugins/workflow.lua
```diff
@@ -157,5 +170,18 @@ function _M.access(conf, ctx)
...
 end
 
+function _M.log(conf, ctx)
+    for idx, rule in ipairs(conf.rules) do
+        local match_result = ctx._workflow_cache[idx]
+        if match_result then
+            -- only one action is currently supported
+            local action = rule.actions[1]
+            local log_handler = support_action[action[1]].log_handler
+            if log_handler then
```
bzp2010, line 182: This if seems to be useless? We can directly return nil to the plugin framework?

apisix/plugins/workflow.lua
```diff
@@ -116,10 +120,17 @@ function _M.check_schema(conf)
...
          if rule.case then
-            local ok, err = expr.new(rule.case)
+            local expr, err = expr.new(rule.case)
             if not ok then
                 return false, "failed to validate the 'case' expression: " .. err
             end
+            local mt = getmetatable(rule)
+            if not mt then
+                mt = {}
+                mt.__index = mt
```
bzp2010, line 130:
```suggestion
                mt = {
                    __index = mt
                }
```

apisix/plugins/limit-count.lua
```diff
@@ -40,11 +40,11 @@ end
...
 function _M.workflow_handler()
     workflow.register(plugin_name,
-    function (conf, ctx)
-        return limit_count.rate_limit(conf, ctx, plugin_name, 1)
-    end,
     function (conf)
         return limit_count.check_schema(conf)
+    end,
+    function (conf, ctx)
+        return limit_count.rate_limit(conf, ctx, plugin_name, 1)
```
nic-chen, line 47: Are there any other plugins that need to be modified like this?

apisix/plugins/workflow.lua
Revolyssup: Not all plugins might pass log_handler callback. This handler's action only needs to be run when log_handler is passed.

apisix/plugins/limit-count.lua
Revolyssup: No these are the only plugins

apisix/plugins/workflow.lua
Revolyssup: done

apisix/plugins/workflow.lua
Revolyssup: mt.__index can only be set after mt is defined.

bzp2010 (approved):

membphis (approved):
LGTM

PR #12509, opened 2025-08-11
feat(ai-proxy-multi): add support for healthcheck

Revolyssup:

Fixes #

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -1234,6 +1234,26 @@ function _M.run_plugin(phase, plugins, api_ctx)
     return api_ctx, plugin_run
 end
 
+function _M.set_plugins_meta_parent(plugins, parent)
+    if not plugins then
+        return
+    end
+    for _, plugin_conf in pairs(plugins) do
+        if not plugin_conf._meta then
+            plugin_conf._meta = {}
+        end
+        if not plugin_conf._meta.parent then
+            local mt_table = getmetatable(plugin_conf._meta)
+            if mt_table then
+                mt_table.parent = parent
+            else
+                plugin_conf._meta = setmetatable(plugin_conf._meta,
+                                                    { __index = {parent = parent} })
+            end
+        end
+    end
+end
+
 
 function _M.run_global_rules(api_ctx, global_rules, phase_name)
     if global_rules and #global_rules > 0 then
--- a/apisix/http/service.lua
+++ b/apisix/http/service.lua
@@ -17,6 +17,7 @@
 local core   = require("apisix.core")
 local apisix_upstream = require("apisix.upstream")
 local plugin_checker = require("apisix.plugin").plugin_checker
+local plugin = require("apisix.plugin")
 local services
 local error = error
 
@@ -45,6 +46,9 @@ local function filter(service)
     if not service.value then
         return
     end
+
+
+    plugin.set_plugins_meta_parent(service.value.plugins, service)
 
     apisix_upstream.filter_upstream(service.value.upstream, service)
 
--- a/apisix/plugins/ai-proxy-multi.lua
+++ b/apisix/plugins/ai-proxy-multi.lua
@@ -19,6 +19,10 @@ local core = require("apisix.core")
 local schema = require("apisix.plugins.ai-proxy.schema")
 local base   = require("apisix.plugins.ai-proxy.base")
 local plugin = require("apisix.plugin")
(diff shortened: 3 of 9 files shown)


apisix/plugin.lua
```diff
@@ -1234,6 +1236,26 @@ function _M.run_plugin(phase, plugins, api_ctx)
...
+        return
+    end
+    for _, plugin_conf in pairs(plugins) do
+        if not plugin_conf._meta then
+            plugin_conf._meta = {}
+        end
+        if not plugin_conf._meta.parent then
+            local mt_table = getmetatable(plugin_conf._meta)
+            if mt_table then
+                mt_table.parent = parent
```
membphis: https://github.com/apache/apisix/pull/12426

this PR has beed merge

so the parent should be a resource path, do you think?

Revolyssup: yes you're correct. I missed this, let me correct.

Revolyssup: the healthcheck_manager module is tied only to manage the checkers on the upstream configuration. I don't know if we can use that mechanism here because here we are creating custom healthchecker using the   `checks` configuration in the plugin configuration.

membphis: yes, i think we have to use the `healthcheck_manager` way

the old way `parent` is bad way which was deprecated

apisix/healthcheck_manager.lua
```diff
@@ -206,6 +210,15 @@ function _M.upstream_version(index, nodes_ver)
 end
 
 
+local function get_plugin_name(path)
+    -- Extract JSON path (after '#') or use full path
+    local json_path = path:match("#(.+)$") or path
+    -- Match plugin name in the JSON path segment
```
nic-6443, line 223: For plugin names that contain hyphens, do we must to enclose them in quotation marks?

Revolyssup: yes. like `plugins["ai-proxy-multi"].`
`plugins.ai-proxy-multi.` will break

apisix/healthcheck_manager.lua
```diff
@@ -60,6 +61,9 @@ end
 
 
 local function fetch_latest_conf(resource_path)
+    -- if resource path contains json path, extract out the prefix
+    -- for eg: extracts /routes/1 from /routes/1#plugins.abc
+    resource_path = resource_path:match("^(.-)#") or resource_path
```
membphis: use `local res = ngx.re.split(...)`

the `res[1]` is resource path, `res[2]` is sub resource path

we can add a new function to parse the path

apisix/healthcheck_manager.lua
```diff
@@ -224,7 +237,17 @@ local function timer_create_checker()
             if not res_conf then
                 goto continue
             end
-            local upstream = res_conf.value.upstream or res_conf.value
+            local upstream
+            local plugin_name = get_plugin_name(resource_path)
+            if plugin_name and plugin_name ~= "" then
+                local json_path = "$." .. (resource_path:match("#(.+)$") or "")
+                local tab = jp.value(res_conf.value, json_path)
```
membphis: need a meaning name

apisix/plugin.lua
```diff
@@ -1234,6 +1236,30 @@ function _M.run_plugin(phase, plugins, api_ctx)
     return api_ctx, plugin_run
 end
 
+function _M.set_plugins_meta_parent(plugins, parent)
+    if not plugins then
+        return
```
membphis, line 1241: return err

apisix/plugin.lua
```diff
@@ -1234,6 +1236,30 @@ function _M.run_plugin(phase, plugins, api_ctx)
...
+            }
+            local mt_table = getmetatable(plugin_conf._meta)
+            if mt_table then
+                mt_table.parent = parent_info
+            else
+                plugin_conf._meta = setmetatable(plugin_conf._meta,
+                                                    { __index = {parent = parent_info} })
+            end
+        end
+    end
```
membphis, line 1260: return true

apisix/plugin.lua
Revolyssup: This is not an error. `plugins` may or may not exist on the route, if they dont then nothing needs to be attached and this logic can be silently skipped

nic-6443 (approved):

membphis (approved):

AlinsRan (approved):

PR #12515, opened 2025-08-12
feat(ai-proxy): add support for pushing logs in ai-proxy plugins

Revolyssup:
Defines log format and add support for pushing ai-proxy request/response in logs to any logger.

The information about logging is already in the doc https://github.com/apache/apisix/blob/master/docs/en/latest/plugins/ai-proxy.md
### Checklist


apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -34,6 +34,7 @@ local pairs = pairs
 local type  = type
 local ipairs = ipairs
 local setmetatable = setmetatable
+local table  = table
```
shreemaan-abhishek: [nit] two spaces after table.

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -97,7 +99,7 @@ local function read_response(ctx, res)
             ngx_print(chunk)
             ngx_flush(true)
 
-            local events, err = ngx_re.split(chunk, "\n")
+            local events, err = ngx_re.split(chunk, "\n\n")
```
shreemaan-abhishek: why split by two new lines?

t/sse_server_example/main.go
```diff
@@ -54,5 +71,5 @@ func main() {
 	port := os.Args[1]
 	// Start the server
 	log.Println("Starting server on :", port)
-	log.Fatal(http.ListenAndServe(":" + port, nil))
+	log.Fatal(http.ListenAndServe(":"+port, nil))
```
shreemaan-abhishek: no need to make this change.

Revolyssup: It happened due to autoformatting and i think its better to have no spaces by lint rules so its okay.

apisix/plugins/ai-drivers/openai-base.lua
Revolyssup: chunks here are actually separated by `\n\n` and logic was wrong previously but not caught because we were extracting chunk by chunk.

nic-6443 (approved):

AlinsRan (approved):

PR #12523, opened 2025-08-14
fix(etcd): load full data from etcd while worker restart

nic-6443:
### Description

When using etcd as the config server, the apisix master process performs a full data load during the init phase and saves the result in `loaded_configuration`. After the master forks other worker processes, these worker processes will directly initialize configurations based on `loaded_configuration`:
https://github.com/apache/apisix/blob/1b0bb84ccf343bd1aedc4bfff1afa9fcefb7c4d5/apisix/core/config_etcd.lua#L1023-L1031
Since `loaded_configuration` in the master process is not continuously updated, after running for some time, `loaded_configuration` may become outdated. If a worker restarts at this point, it will use an outdated configuration for initialization. If any requests are sent to this worker at that time, these requests may be processed by incorrect confiurations.

### Benchmark

Because the issue addressed by this PR only becomes apparent when APISIX is continuously handling high QPS requests, I am unable to provide effective automated integration tests. 
So I conducted a continuous  one hour stability test, here are the testing steps and results.

#### Steps
1. start APISIX with two workers
```
nginx_config:
  worker_processes: 2
```
2. create 1000 routes after APISIX startup
```
#!/usr/bin/env bash
set -euo pipefail

APISIX_HOST="http://127.0.0.1:9180"
ADMIN_API_KEY="${ADMIN_API_KEY:-edd1c9f034335f136f87ad84b625c8f1}"

for i in $(seq 1 1000); do
  echo ">>> creating upstream $i"
  curl -s -o /dev/null -X PUT "$APISIX_HOST/apisix/admin/upstreams/$i" \
    -H "X-API-KEY: $ADMIN_API_KEY" \
    -H "Content-Type: application/json" \
    -d "{
      \"id\": $i,
      \"nodes\": {\"127.0.0.1:1980\": 1},
      \"type\": \"roundrobin\"
    }"

  echo ">>> creating route $i"
  curl -s -o /dev/null -X PUT "$APISIX_HOST/apisix/admin/routes/$i" \
    -H "X-API-KEY: $ADMIN_API_KEY" \
    -H "Content-Type: application/json" \
    -d "{
      \"id\": $i,
      \"uri\": \"/$i\",
      \"upstream_id\": $i
    }"
done

echo ">>> All routes and upstreams created"
```
3. kill one worker process randomly every 60s
```
#!/usr/bin/env bash
set -euo pipefail

MASTER_PID="${NGINX_MASTER_PID:?must set NGINX_MASTER_PID}"

while true; do
  worker_pids=$(pgrep -P "$MASTER_PID" -f "nginx: worker process" || true)

  if [ -n "$worker_pids" ]; then
    pid=$(echo "$worker_pids" | shuf -n 1)
    echo ">>> killing worker $pid (master $MASTER_PID)"
    kill "$pid"
  else
    echo ">>> no worker process found for master $MASTER_PID"
  fi

  sleep 60
done
```
4. use wrk to continuously request APISIX
```
wrk -t2 -c100 -R6000 -d1h -s request.lua http://127.0.0.1:9080
```

#### Result
```
❯ wrk -t2 -c100 -R6000 -d1h -s request.lua http://127.0.0.1:9080

Running 60m test @ http://127.0.0.1:9080
  2 threads and 100 connections
  Thread calibration: mean lat.: 1.313ms, rate sampling interval: 10ms
  Thread calibration: mean lat.: 1.331ms, rate sampling interval: 10ms
    Latency     1.60ms    9.28ms 337.15ms   99.31%
    Req/Sec     2.92k   508.56    18.00k    74.18%
  19835460 requests in 60.00m, 3.31GB read
  Socket errors: connect 0, read 2623, write 0, timeout 14660
Requests/sec:   5509.85
Transfer/sec:      0.94MB
```

Fixes #

### Checklist

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -74,6 +74,7 @@ if not is_http then
 end
 local created_obj  = {}
 local loaded_configuration = {}
+local configuration_loaded_time
 local watch_ctx
 
 
@@ -1125,6 +1126,22 @@ local function create_formatter(prefix)
 end
 
 
+local function init_loaded_configuration()
+    loaded_configuration = {}
+    local etcd_cli, prefix, err = etcd_apisix.new_without_proxy()
+    if not etcd_cli then
+        return "failed to start a etcd instance: " .. err
+    end
+
+    local res, err = readdir(etcd_cli, prefix, create_formatter(prefix))
+    if not res then
+        return err
+    end
+
+    configuration_loaded_time = ngx_time()
+end
+
+
 function _M.init()
     local local_conf, err = config_local.local_conf()
     if not local_conf then
@@ -1135,14 +1152,8 @@ function _M.init()
         return true
     end
 
-    -- don't go through proxy during start because the proxy is not available
-    local etcd_cli, prefix, err = etcd_apisix.new_without_proxy()
-    if not etcd_cli then
-        return nil, "failed to start a etcd instance: " .. err
-    end
-
-    local res, err = readdir(etcd_cli, prefix, create_formatter(prefix))
-    if not res then
+    local err = init_loaded_configuration()
+    if err then
         return nil, err
     end
 
@@ -1157,8 +1168,14 @@ function _M.init_worker()
         return nil, err
     end
 
-    if table.try_read_attr(local_conf, "apisix", "disable_sync_configuration_during_start") then
-        return true
+    -- if the startup time of a worker differs significantly from that of the master process,
+    -- we consider it to have restarted, and at this point,
+    -- it is necessary to reload the full configuration from etcd.
+    if configuration_loaded_time and ngx_time() - configuration_loaded_time > 60 then
+        local err = init_loaded_configuration()
+        if err then
+            return nil, err
+        end
     end
 
     return true
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -1168,10 +1168,14 @@ function _M.init_worker()
         return nil, err
     end
 
+    local threshold = table.try_read_attr(local_conf, "apisix",
+                                    "worker_startup_time_threshold") or 60
     -- if the startup time of a worker differs significantly from that of the master process,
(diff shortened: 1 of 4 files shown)


apisix/core/config_etcd.lua
```diff
@@ -1157,8 +1168,14 @@ function _M.init_worker()
         return nil, err
     end
 
-    if table.try_read_attr(local_conf, "apisix", "disable_sync_configuration_during_start") then
-        return true
+    -- if the startup time of a worker differs significantly from that of the master process,
+    -- we consider it to have restarted, and at this point,
+    -- it is necessary to reload the full configuration from etcd.
+    if configuration_loaded_time and ngx_time() - configuration_loaded_time > 60 then
```
shreemaan-abhishek: why hardcode it as 60 seconds? could it not be made a configurable parameter?

nic-6443: This is an empirical value. I think 60 seconds is a sufficient duration to distinguish between the worker's first-time and restart scenarios. It can be made into a configuration parameter, but I don't think users would know how to set this configuration parameter properly.

shreemaan-abhishek: > but I don't think users would know how to set this configuration parameter properl

makes sense 👍🏼

apisix/core/config_etcd.lua
```diff
@@ -74,6 +74,7 @@ if not is_http then
 end
 local created_obj  = {}
 local loaded_configuration = {}
+local configuration_loaded_time
```
membphis, line 77: I can’t sure if we can use Lua global var to cache the value. If I do this, will store it in sharedict 
Pls add test it to confirm

nic-6443: Okay, I will add a test case, but this one is obviously working. You can see another variable above `loaded_configuration`, which uses the same mechanism to allow worker processes to reuse the etcd data loaded when the master process starts.

membphis (approved):

nic-chen (approved):

PR #12526, opened 2025-08-14
feat: add last modified and digest metadata to standalone API

bzp2010:
### Description

1. Add a new metadata response header to the Standalone API. This includes `X-Last-Modified` and `X-Digest`. 
They are used to report to a cluster of multiple APISIX instances, which instance was last updated, and the configuration digest currently in use.
The `X-Digest` is passed in by the `PUT` request and recorded in memory, and then output in the response header of a `GET` or `HEAD` request. This value is transparent to APISIX, which does not parse or generate the value in any way; it is completely controlled by the client. There is a simple judgment here; if the digest input for the PUT request matches what is currently stored in memory, the API will return 204, and the configuration update will not actually occur.

2. Add support for the `HEAD` method, which returns only metadata but not configuration.

### Checklist

I explicitly point out that this is not backwards compatible. Since this feature is clearly marked as experimental and users have been warned not to rely on it, we can comfortably make the changes we need.


SkyeYoung:
Please fix the errors reported in the CI.

apisix/admin/standalone.lua
```diff
@@ -233,11 +252,17 @@ local function update(ctx)
         end
     end
 
+    -- write metadata
+    apisix_yaml[METADATA_LAST_MODIFIED] = ngx_time()
```
AlinsRan, line 256: You might consider storing `METADATA_LAST_MODIFIED` and `METADATA_DIGEST` in a separate shared memory key rather than mixing them into apisix_yaml, so as not to affect the diff logic and to keep the config structure cleaner.

apisix/admin/standalone.lua
```diff
@@ -286,15 +311,37 @@ local function get(ctx)
...
 
+local function head(ctx)
+    local config, err = get_config()
+    if not config then
+        if err ~= "not found" then
+            core.log.error("failed to get config from shared dict: ", err)
+            return core.response.exit(500, {
+                error_msg = "failed to get config from shared dict: " .. err
+            })
+        end
```
SkyeYoung, line 329: add some logs for other types of errors?

apisix/admin/standalone.lua
bzp2010: This is intentional, I want to make sure that the record is atomic, succeeding at the same time or failing at the same time. If it partially succeeds, this will lead to accidents.

apisix/admin/standalone.lua
bzp2010: Only `not found` is skipped by this if, which is expected to exist, so it seems fine.

apisix/admin/standalone.lua
```diff
@@ -138,9 +143,16 @@ end
 
 
 local function update(ctx)
-    local content_type = core.request.header(nil, "content-type") or "application/json"
+    -- check digest header existence
+    local digest = core.request.header(nil, METADATA_DIGEST)
+    if not digest then
```
membphis, line 148: I don't know why this "digest" is not optional?

It seems to be a break change

bzp2010: I clearly stated this in the PR description.

1. This is a break change
2. This is necessary to ensure global consistency in new releases.
3. I'm not obligated to guarantee backward compatibility of an experimental feature. Upon my request, the feature was documented as experimental in the 3.13 documentation.

membphis: I reviewed again, agree with you

PR #12539, opened 2025-08-20
fix: basic auth scheme supports case insensitivity

### LINKED ISSUE #11837: bug: Basic authentication requests are case-insensitive both in scheme and parameters, but this isn't supported by APISIX
### Current Behavior

Lines like https://github.com/apache/apisix/blob/master/apisix/plugins/basic-auth.lua#L81 prescribe case-sensitive header keys and values but according to https://datatracker.ietf.org/doc/html/rfc7617#page-3 (scroll down a bit):

```
Note that both scheme and parameter names are matched case-
insensitively.
```

APISIX thus does not appear to follow the http spec.

Products such as Elasticsearch have addressed [this issue](https://github.com/elastic/elasticsearch/issues/31486)

Thank you.

### Expected Behavior

APISIX should listen to (and unit-test for)

- `authorization: basic` [etc.]
- `Authorization: basic` [etc.]
- `authorization: Basic` [etc.]
Even:
- `AuThOrIZaTIOn: bAsIC`
is a valid authorization header string

### Error Logs

_No response_

### Steps to Reproduce

I'm unsure on what to write here. The code is in the default branch. 

### Environment

This doesn't appear to be a version-specific thing.

SkyeYoung:
### Description

The parsing part is reimplemented with reference to the following rfcs:

* https://datatracker.ietf.org/doc/html/rfc7235
* https://datatracker.ietf.org/doc/html/rfc9110
* https://datatracker.ietf.org/doc/html/rfc5234

It should be noted that the original Http header field name part can handle case-insensitivity, so this PR focuses on processing the Http header field value part.

Fixes #11837 

### Checklist


apisix/plugins/ldap-auth.lua
```diff
@@ -66,23 +67,27 @@ function _M.check_schema(conf, schema_type)
     return ok, err
 end
 
-local function extract_auth_header(authorization)
+local function extract_auth_header(auth)
     local obj = { username = "", password = "" }
+    -- Check format of Authorization header.
+    local matches, err = ngx_re.split(auth, "\\s+", nil, nil, 2)
 
-    local m, err = ngx.re.match(authorization, "Basic\\s(.+)", "jo")
```
bzp2010, line 72: Why not use regex to do the pattern matching and data extraction directly?
`Basic` is short, and in the worst case we only need to assert the uppercase and lowercase format of each letter.
There doesn't seem to be a significant means from using split.

`[Bb][Aa][Ss][Ii][Cc]\\s(.*)`
`(?i:basic)\\s(.*)`

t/plugin/basic-auth.t
```diff
@@ -252,7 +288,7 @@ GET /t
 
 
 
-=== TEST 13: invalid schema, not field given
+=== TEST 16: invalid schema, not field given
```
bzp2010: Do these tests need to be added in the middle? Can it be in a new file or at the end of the old one?

SkyeYoung: The reason I did this was to put similar test cases together. 🤔

apisix/plugins/ldap-auth.lua
SkyeYoung: The current implementation refers to https://github.com/apache/apisix/blob/12b0824e23e82ee097dcdd3fa7faafe35dc674c8/apisix/plugins/openid-connect.lua#L393-L407.

But as you said, it is indeed much simpler in terms of syntax.

I may need to try it to determine their performance.

moonming:
Totally agree with @bzp2010

apisix/plugins/ldap-auth.lua
SkyeYoung: done

t/plugin/basic-auth.t
SkyeYoung: done

bzp2010 (approved):
LGTM

PR #12029, opened 2025-03-09
feat: extend chaitin-waf plugin functionalities #9945

AlyHKafoury:
### Description

Fixed #9945

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #12551, opened 2025-08-27
feat: use `trusted_addresses` to handle `X-Forwarded-*` headers

SkyeYoung:

Fixes #

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -45,6 +45,7 @@ local xrpc            = require("apisix.stream.xrpc")
 local ctxdump         = require("resty.ctxdump")
 local debug           = require("apisix.debug")
 local pubsub_kafka    = require("apisix.pubsub.kafka")
+local trusted_addresses_util = require("apisix.utils.trusted-addresses")
 local ngx             = ngx
 local get_method      = ngx.req.get_method
 local ngx_exit        = ngx.exit
@@ -159,6 +160,8 @@ function _M.http_init_worker()
     -- To ensure that all workers related to Prometheus metrics are initialized,
     -- we need to put the initialization of the Prometheus plugin here.
     plugin.init_prometheus()
+
+    trusted_addresses_util.init_worker()
 end
 
 
@@ -599,6 +602,19 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
 end
 
 
+local function handle_x_forwarded_headers(api_ctx)
+    if trusted_addresses_util.is_trusted(api_ctx.var.remote_addr) then
+        api_ctx.var.x_forwarded_proto = core.request.header(api_ctx, "X-Forwarded-Proto") or api_ctx.var.scheme
+        api_ctx.var.x_forwarded_host = core.request.header(api_ctx, "X-Forwarded-Host") or api_ctx.var.host
+        api_ctx.var.x_forwarded_port = core.request.header(api_ctx, "X-Forwarded-Port") or api_ctx.var.server_port
+    else
+        api_ctx.var.x_forwarded_host = api_ctx.var.host
+        api_ctx.var.x_forwarded_proto = api_ctx.var.scheme
+        api_ctx.var.x_forwarded_port = api_ctx.var.server_port
+    end
+end
+
+
 function _M.http_access_phase()
     -- from HTTP/3 to HTTP/1.1 we need to convert :authority pesudo-header
     -- to Host header, so we set upstream_host variable here.
@@ -647,6 +663,8 @@ function _M.http_access_phase()
     -- the original request_uri can be accessed via var.real_request_uri
     api_ctx.var.real_request_uri = api_ctx.var.request_uri
     api_ctx.var.request_uri = api_ctx.var.uri .. api_ctx.var.is_args .. (api_ctx.var.args or "")
+
+    handle_x_forwarded_headers(api_ctx)
 
     router.router_http.match(api_ctx)
 
--- a/apisix/utils/trusted-addresses.lua
+++ b/apisix/utils/trusted-addresses.lua
@@ -1,0 +1,59 @@
+local require         = require
+local core            = require("apisix.core")
+local next            = next
+local ipairs          = ipairs
+
+local trusted_addresses_matcher
+
+local _M = {}
+
+
+local function validate_trusted_addresses(trusted_addresses)
+    if trusted_addresses and next(trusted_addresses) then
+        for _, cidr in ipairs(trusted_addresses) do
+            if not core.ip.validate_cidr_or_ip(cidr) then
+                core.log.error("Invalid IP/CIDR '", cidr, "' exists in trusted_addresses")
+                return false
+            end
+        end
+    end
+    return true
+end
+
+
+function _M.init_worker()
+    local local_conf = core.config.local_conf()
+    local trusted_addresses = core.table.try_read_attr(local_conf, "apisix", "trusted_addresses")
+
+    if not trusted_addresses then
+        return
+    end
+    
+    if not core.table.isarray(trusted_addresses) then
+        core.log.error("trusted_addresses '", trusted_addresses, "' is not an array, please check your configuration")
+        return
+    end
+
+    if not validate_trusted_addresses(trusted_addresses) then
+        return
+    end
+
+    local matcher, err = core.ip.create_ip_matcher(trusted_addresses)
+    if not matcher then
+        core.log.error("failed to create ip matcher for trusted_addresses: ", err)
+        return
+    end
+    
+    trusted_addresses_matcher = matcher
+end
+
+
+function _M.is_trusted(address)
+    if not trusted_addresses_matcher then
+        core.log.info("trusted_addresses_matcher is not initialized, skipping subsequent parsing.")
+        return false
+    end
+    return trusted_addresses_matcher(address)
+end
+
+return _M


apisix/init.lua
```diff
@@ -290,21 +293,6 @@ end
...
-
-    local x_forwarded_host = api_ctx.var.http_x_forwarded_host
-    if x_forwarded_host then
-        api_ctx.var.var_x_forwarded_host = x_forwarded_host
-    end
-
-    local port = api_ctx.var.http_x_forwarded_port
-    if port then
-        api_ctx.var.var_x_forwarded_port = port
-    end
```
SkyeYoung, line 309: https://github.com/apache/apisix/blob/09f6e3682c5da1437f951273200d44cc18b13a28/apisix/plugins/ai.lua#L277-L282

Because the code above will replace `handle_upstream` with `ai_upstream` below:

https://github.com/apache/apisix/blob/09f6e3682c5da1437f951273200d44cc18b13a28/apisix/plugins/ai.lua#L115-L117

This will cause this deleted part of the code to be skipped.

Now that we need to handle `X-Forwarded-*`, we need to update these `var_x_forwarded_*`, so we need to extract this part of the code as `update_var_x_forwarded_headers`.

https://github.com/apache/apisix/blob/3260931f301f741d5551af895cbfd6b12d1aff31/apisix/cli/ngx_tpl.lua#L835-L838

apisix/init.lua
```diff
@@ -599,6 +587,53 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
 end
 
 
+local function handle_x_forwarded_headers(api_ctx)
+    local addr_is_trusted = trusted_addresses_util.is_trusted(api_ctx.var.realip_remote_addr)
```
membphis, line 591: it seems wrong, pls confirm should we use `var.remote_addr` here

conf/config.yaml.example
```diff
@@ -141,6 +141,8 @@ apisix:
                                   # or (standalone mode) the config isn't loaded yet either via file or Admin API.
   # disable_upstream_healthcheck: false # A global switch for healthcheck. Defaults to false.
                                         # When set to true, it overrides all upstream healthcheck configurations and globally disabling healthchecks.
+# trusted_addresses:              # When enabled, APISIX will trust the `X-Forwarded-*` Headers
```
membphis: add more explain about the default process way

apisix/init.lua
membphis: I made a test, confirmed you are right

membphis (approved):

conf/config.yaml.example
SkyeYoung: done

t/plugin/dubbo-proxy/upstream.t
```diff
@@ -51,6 +51,8 @@ _EOC_
 apisix:
     node_listen: 1984
     enable_admin: false
+    trusted_addresses:
+        - "127.0.0.1"
```
moonming: why change this test case?

t/plugin/proxy-rewrite2.t
```diff
@@ -27,6 +27,8 @@ add_block_preprocessor(sub {
     my $yaml_config = $block->yaml_config // <<_EOC_;
 apisix:
     node_listen: 1984
+    trusted_addresses:
```
moonming, line 30: ditto

t/plugin/sls-logger.t
```diff
@@ -267,6 +267,8 @@ apisix:
         enable_encrypt_fields: true
         keyring:
             - edd1c9f0985e76a2
+    trusted_addresses:
```
moonming: ditto

t/core/trusted-addresses.t
```diff
@@ -0,0 +1,219 @@
+#
```
moonming, line 1: Need to test the situation where trusted-addresses contain multiple IP addresses

conf/config.yaml.example
```diff
@@ -141,6 +141,10 @@ apisix:
...
+#   - 127.0.0.1                   # passed in requests from the IP/CIDR in the list.
+#   - 0.0.0.0/0                   # CAUTION: When not configured, APISIX will remove `X-Forwarded-*` headers
```
moonming: No IP restriction is a bad example

conf/config.yaml.example
```diff
@@ -141,6 +141,10 @@ apisix:
                                   # or (standalone mode) the config isn't loaded yet either via file or Admin API.
   # disable_upstream_healthcheck: false # A global switch for healthcheck. Defaults to false.
                                         # When set to true, it overrides all upstream healthcheck configurations and globally disabling healthchecks.
+# trusted_addresses:              # When enabled, APISIX will trust the `X-Forwarded-*` Headers
```
moonming:
```suggestion
# trusted_addresses:              # When enabled, APISIX will trust the `X-Forwarded-*` Headers
```
how to disabled?

t/core/trusted-addresses.t
moonming: and need test cases in plugins.

moonming (changes requested):

conf/config.yaml.example
SkyeYoung: If not configured, it is disabled by default.

This statement references https://github.com/apache/apisix/blob/1e3950e43168097d09227da79dc69fb044f4a1ee/conf/config.yaml.example#L138

moonming: write it in this file and docs

t/plugin/proxy-rewrite2.t
SkyeYoung: In order to pass the original tests.

https://github.com/apache/apisix/blob/197ce75259b729717f4a5f7cbca898e4f73bc4fc/t/plugin/proxy-rewrite2.t#L154-L235

conf/config.yaml.example
SkyeYoung: done

conf/config.yaml.example
SkyeYoung: done

t/core/trusted-addresses.t
SkyeYoung: done

t/plugin/dubbo-proxy/upstream.t
SkyeYoung: Added in batches, removed

t/plugin/sls-logger.t
SkyeYoung: Added in batches, removed

t/core/trusted-addresses.t
SkyeYoung: Other than `chaitin-waf`, [the original PR](https://github.com/apache/apisix/pull/12029) didn't find any tests covering the `real_client_ip=true` scenario, and after spending time researching, I also couldn't find the actual purpose of this parameter.

SkyeYoung:
@moonming ping

apisix/init.lua
```diff
@@ -599,6 +587,53 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
...
+
+    if not addr_is_trusted then
+        -- store the original x-forwarded-* headers for later process
```
moonming: where is the `later process`?

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,79 @@
...
+
+local trusted_addresses_matcher
+
+local _M = {}
+
+
+local function validate_trusted_addresses(trusted_addresses)
+    for _, cidr in ipairs(trusted_addresses) do
+        if not core.ip.validate_cidr_or_ip(cidr) then
+            core.log.error("Invalid IP/CIDR '", cidr, "' exists in trusted_addresses")
```
moonming: test case

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,79 @@
...
+function _M.init_worker()
+    local local_conf = core.config.local_conf()
+    local trusted_addresses = core.table.try_read_attr(local_conf, "apisix", "trusted_addresses")
+
+    if not trusted_addresses then
+        return
+    end
+
+    if not core.table.isarray(trusted_addresses) then
+        core.log.error("trusted_addresses '", trusted_addresses,
```
moonming: test case

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,79 @@
...
+    end
+    return true
+end
+
+
+function _M.init_worker()
+    local local_conf = core.config.local_conf()
+    local trusted_addresses = core.table.try_read_attr(local_conf, "apisix", "trusted_addresses")
+
+    if not trusted_addresses then
```
moonming, line 29: ditto

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,79 @@
...
+        return
+    end
+
+    if not validate_trusted_addresses(trusted_addresses) then
+        return
+    end
+
+    local matcher, err = core.ip.create_ip_matcher(trusted_addresses)
+    if not matcher then
+        core.log.error("failed to create ip matcher for trusted_addresses: ", err)
```
moonming, line 36: ditto

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,79 @@
...
+        return
+    end
+
+    trusted_addresses_matcher = matcher
+end
+
+
+function _M.is_trusted(address)
+    if not trusted_addresses_matcher then
+        core.log.info("trusted_addresses_matcher is not initialized")
```
moonming, line 46: ditto

conf/config.yaml.example
```diff
@@ -141,6 +141,10 @@ apisix:
                                   # or (standalone mode) the config isn't loaded yet either via file or Admin API.
   # disable_upstream_healthcheck: false # A global switch for healthcheck. Defaults to false.
                                         # When set to true, it overrides all upstream healthcheck configurations and globally disabling healthchecks.
+# trusted_addresses:              # When configured, APISIX will trust the `X-Forwarded-*` Headers
+#   - 127.0.0.1                   # passed in requests from the IP/CIDR in the list.
+#   - 172.18.0.0/16               # CAUTION: When not configured, APISIX will remove `X-Forwarded-*` headers
```
moonming: override or remove?

docs/en/latest/plugins/chaitin-waf.md
```diff
@@ -70,6 +70,10 @@ The response headers are listed below:
 | config.keepalive_timeout | integer       | false    | 60000         | Idle connection timeout, in milliseconds.                                                                                    |
 | config.real_client_ip    | boolean       | false    | true          | Specifies whether to use the `X-Forwarded-For` as the client IP (if present). If `false`, uses the direct client IP from the connection. |
 
+:::note
+Only `X-Forwarded-*` headers sent from addresses in the `apisix.trusted_addresses` configuration (supports IP and CIDR) will be trusted and passed to plugins or upstream. If `apisix.trusted_addresses` is not configured or the IP is not within the configured address range, all `X-Forwarded-*` headers will be discarded.
```
moonming: discarded or overwrite?

moonming (changes requested):

apisix/init.lua
SkyeYoung: There is no follow-up processing at present, just to allow some plugins or processes to use it in the future.

I will adjust the description for precise expression.

apisix/utils/trusted-addresses.lua
SkyeYoung: This situation is already covered by `TEST 1`. I will add an info-level log for assertion.

apisix/utils/trusted-addresses.lua
SkyeYoung: After research, I found that this part of the check should be placed in `cli/schema.lua`, and relevant tests should be conducted.

apisix/utils/trusted-addresses.lua
SkyeYoung: After research, I found that this part of the check should be placed in cli/schema, and relevant tests should be conducted.

SkyeYoung: I found that this part overlaps with the functionality provided by `core.ip.create_ip_matcher`, so I removed it.

apisix/utils/trusted-addresses.lua
SkyeYoung: done

apisix/init.lua
SkyeYoung: done

conf/config.yaml.example
SkyeYoung: override. changed

docs/en/latest/plugins/chaitin-waf.md
SkyeYoung: done

apisix/utils/trusted-addresses.lua
SkyeYoung: This is just defensive programming.

The validation of ip/cidr has already been handled in `validate_trusted_addresses`. Currently, no other cases have been found that require assertion here.

Therefore, only `validate_trusted_addresses` has been tested.

SkyeYoung: You might ask why verification is needed upfront. This is because I just discovered that https://github.com/api7/lua-resty-ipmatcher has the following issue that needs to be fixed:

Image

This library considers `1.0.0.0/33` to be a valid CIDR.

apisix/utils/trusted-addresses.lua
SkyeYoung: done

apisix/utils/trusted-addresses.lua
SkyeYoung: done

docs/zh/latest/plugins/chaitin-waf.md
```diff
@@ -89,6 +89,11 @@ description: chaitin-waf 插件与长亭雷池 WAF 集成，以检测和阻止
 
 继续操作之前，请确保您已安装 [长亭雷池 WAF](https://docs.waf.chaitin.com/en/GetStarted/Deploy)。
 
+
```
SkyeYoung:
```suggestion
```

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,67 @@
...
+    if not trusted_addresses then
+        core.log.info("trusted_addresses is not configured")
+        return
+    end
+
+    if not validate_trusted_addresses(trusted_addresses) then
+        return
+    end
+
+    local matcher, err = core.ip.create_ip_matcher(trusted_addresses)
```
nic-6443, line 49: we need a lru cache for `matcher` like:
https://github.com/apache/apisix/blob/bb71dfc8c62d41915a4c66f53682d1bbb18ab297/apisix/plugins/real-ip.lua#L79-L80

apisix/cli/schema.lua
```diff
@@ -259,6 +259,14 @@ local config_schema = {
                     default = false,
                     description = "a global switch to disable upstream health checks",
                 },
+                trusted_addresses = {
```
nic-6443, line 263: ref: https://github.com/apache/apisix/blob/bb71dfc8c62d41915a4c66f53682d1bbb18ab297/apisix/plugins/real-ip.lua#L32-L36
we can add more restrction to schema like this.

apisix/init.lua
```diff
@@ -599,6 +587,54 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
...
+        core.request.set_header(api_ctx, "X-Forwarded-Proto", proto)
+        core.request.set_header(api_ctx, "X-Forwarded-Host", host)
+        core.request.set_header(api_ctx, "X-Forwarded-Port", port)
+        -- later processed in ngx_tpl by `$proxy_add_x_forwarded_for`
+        core.request.set_header(api_ctx, "X-Forwarded-For", nil)
+    end
+end
+
+
+local function update_var_x_forwarded_headers(api_ctx)
```
nic-6443: I suggest renaming this function to `set_upstream_x_forwarded_headers` because the ultimate purpose of updating these `ctx.var` is to update the headers sent upstream. Just like the previous function named `set_upstream_headers`, using a name like `update_var...` makes it hard to know that it's for updating headers sent upstream without looking at specific variable usage.

nic-6443 (approved):

apisix/cli/schema.lua
SkyeYoung: Adding core.schema.ip_def doesn't seem to work.

https://github.com/apache/apisix/blob/8bb6802bd2b2ccd4dabe78e90beb8f87ec046035/apisix/schema_def.lua depends on https://github.com/apache/apisix/blob/0fd582bc299a9d90893d3843cd3ec21f217cb40e/apisix/core/schema.lua

Using it will cause an error `/usr/local/openresty//luajit/bin/luajit: ./apisix/core/lrucache.lua:22: module '\''resty.lrucache'\'' not found`:

image
image

apisix/init.lua
SkyeYoung: done.

nic-6443 (dismissed):

t/core/trusted-addresses.t
```diff
@@ -0,0 +1,416 @@
...
+    uri: /old_uri
+    upstream:
+        nodes:
+            "127.0.0.1:1980": 1
+        type: roundrobin
+#END
+--- request
+GET /old_uri
+--- error_log
+invalid IP/CIDR '1.0.0' exists in trusted_addresses
```
moonming: we also need to check the value of trusted_addresses when apisix starts, not only at runtime.

apisix/cli/schema.lua
```diff
@@ -259,6 +259,14 @@ local config_schema = {
...
+                    type = "array",
+                    minItems = 1,
+                    items = {
+                        type = "string",
```
moonming: "oneOf": [
            {"format": "ipv4"},
            {"format": "ipv6"}
        ]

apisix/init.lua
```diff
@@ -599,6 +587,54 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
...
+        local proto = api_ctx.var.scheme
+        local host = api_ctx.var.host
+        local port = api_ctx.var.server_port
+
+        api_ctx.var.http_x_forwarded_proto = proto
+        api_ctx.var.http_x_forwarded_host = host
+        api_ctx.var.http_x_forwarded_port = port
+        api_ctx.var.http_x_forwarded_for = nil
+
+        -- override the x-forwarded-* headers to the trusted ones
```
moonming: In the `if not trusted` logic of this code, where the `trusted ones` come from?

apisix/init.lua
```diff
@@ -599,6 +587,54 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
...
+        core.request.set_header(api_ctx, "X-Forwarded-For", nil)
+    end
+end
+
+
+local function set_upstream_x_forwarded_headers(api_ctx)
+    local proto = api_ctx.var.http_x_forwarded_proto
+    if proto then
+        api_ctx.var.var_x_forwarded_proto = proto
+    end
```
moonming, line 649: why not set `X-Forwarded-Proto` as `http_x_forwarded_proto`?

apisix/utils/trusted-addresses.lua
```diff
@@ -0,0 +1,67 @@
...
+local require       = require
+local core          = require("apisix.core")
+local ipairs        = ipairs
+
+local trusted_addresses_matcher
+
+local _M = {}
+
+
+local function validate_trusted_addresses(trusted_addresses)
```
moonming: we also need to check this when apisix starts.

moonming (changes requested):

apisix/cli/schema.lua
SkyeYoung: There are a few problems with doing this:

1. Need CIDR support
2. As discussed in https://github.com/apache/apisix/pull/12551#discussion_r2347756462, try to use the existing schema_def. A feasible solution is to copy the corresponding schema again.

apisix/init.lua
SkyeYoung: https://github.com/apache/apisix/blob/33b9632cb893d150b54c5c52efe02bc0adbe7dc3/apisix/cli/ngx_tpl.lua#L830-L837

Referenced the implementation of APISIX.

apisix/init.lua
SkyeYoung: I don't quite understand.

1. This function is extracted from the following code without any adjustments.

https://github.com/apache/apisix/pull/12551/files/528116c496c9c5ea3ef2b3030e1da6596bac67cb#diff-7439ffe94cc4bcad12066d5d83f6c1004b8e9b3ff5097b645fa7c700eb3b3743L293-L307

2. `X-Forwarded-Proto` in the header == `http_x_forwarded_proto`

apisix/cli/schema.lua
moonming: > Using it will cause an error /usr/local/openresty//luajit/bin/luajit: ./apisix/core/lrucache.lua:22: module '\''resty.lrucache'\'' not found:

What is the reason for this error？

apisix/cli/schema.lua
SkyeYoung: Currently, I will try to copy a schema from schema_def to implement this function.

apisix/init.lua
```diff
@@ -599,6 +587,54 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
 end
 
 
+local function handle_x_forwarded_headers(api_ctx)
```
moonming, line 590: Please add some comments to explain the purpose of each code segment

apisix/cli/schema.lua
```diff
@@ -66,6 +68,24 @@ local etcd_schema = {
     required = {"prefix", "host"}
 }
 
+-- copy from apisix/schema_def.lua
```
moonming: please do not copy. we should find the root reason

SkyeYoung: My suggestion is to extract this schema separately.

t/core/trusted-addresses.t
SkyeYoung: done

apisix/utils/trusted-addresses.lua
SkyeYoung: done

apisix/cli/schema.lua
nic-6443: The reason for this error is that there is no openresty environment in the apisix cli, which makes it impossible to use the `resty.lru` library. 
I have considered the dependency relationship between `schema_def.lua` and `schema.lua`, and I think their dependency relationship is reversed. It should be that `schema.lua` depends on `schema_def.lua`, allowing `schema_def.lua` to retain only the definition of schema, in line with the naming of this file. 
Added a new commit to this PR to fix this dependency issue: https://github.com/apache/apisix/pull/12551/commits/30ce9c26e65f9e8ecfc063c08abe9b94b75bd9ba

apisix/cli/schema.lua
SkyeYoung: This part has been handled by @nic-6443 . Thanks to nic for the assistance.

apisix/init.lua
SkyeYoung: done

apisix/core/schema.lua
```diff
@@ -68,4 +70,9 @@ end
 
 _M.valid = get_validator
 
+setmetatable(_M, {
+    __index = schema_def,
+    __newindex = function() error("no modification allowed") end,
```
bzp2010, line 75: I'm not sure what its purpose is, but this approach cannot prevent `schema_def` from being modified. The `__newindex` is only called when the key to be modified does not exist in the original table; it is not invoked for fields that already exist.

apisix/core.lua
```diff
@@ -48,7 +48,7 @@ return {
     request     = require("apisix.core.request"),
     response    = require("apisix.core.response"),
     lrucache    = require("apisix.core.lrucache"),
-    schema      = require("apisix.schema_def"),
+    schema      = require("apisix.core.schema"),
```
bzp2010, line 51: What's the purpose? To unify things?

SkyeYoung: @bzp2010 Both were brought by [`30ce9c2` (#12551)](https://github.com/apache/apisix/pull/12551/commits/30ce9c26e65f9e8ecfc063c08abe9b94b75bd9ba) and the original purpose was to solve this problem: https://github.com/apache/apisix/pull/12551#discussion_r2347932417

apisix/core/schema.lua
SkyeYoung: replied below

nic-6443 (approved):

PR #12563, opened 2025-08-29
fix(ai-proxy): catch malformed override endpoint in schema validation

Revolyssup:
I noticed that when invalid url is passed in override endpoint like "http//localhost:8080" (here : is missing). The ai configuration is still accepted.
But later error comes in user request path and user gets 500
Screenshot 2025-08-29 at 12 04 55 PM

This PR adds endpoint validation during schema check.

### DIFF
--- a/apisix/plugins/ai-proxy-multi.lua
+++ b/apisix/plugins/ai-proxy-multi.lua
@@ -30,6 +30,7 @@ local ipairs = ipairs
 local type = type
 
 local priority_balancer = require("apisix.balancer.priority")
+local endpoint_regex = "^(https?)://([^:/]+):?(%d*)/?.*$"
 
 local pickers = {}
 local lrucache_server_picker = core.lrucache.new({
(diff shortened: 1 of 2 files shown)


apisix/plugins/ai-proxy-multi.lua
```diff
@@ -30,6 +30,7 @@ local ipairs = ipairs
 local type = type
 
 local priority_balancer = require("apisix.balancer.priority")
+local endpoint_regex = "^(https?)://([^:/]+):?(%d*)/?.*$"
```
bzp2010, line 33: Are you sure this is correct?

1. The `/` appears to be unescaped. If the Lua string `"xxx"` can avoid this issue, please disregard.
2. `%d` seems to cause backtracking and result in a failed match? Whereas `\d` matches correctly.

I'm using https://regex101.com PCRE2 mode to check it.

AlinsRan (approved):

apisix/plugins/ai-proxy-multi.lua
Revolyssup: I guess its parsed differently in the website vs the lua code.
1. is no issue. its parsed correctly as can be confirmed by tests.
2. It does match otherwise tests would be failing.

PR #12571, opened 2025-09-01
feat: add fallback mechanism for specific error codes in ai-proxy-multi

Revolyssup:
429/5xx response codes can enter the retry logic and improve the request success rate.

### Checklist


docs/en/latest/plugins/ai-rate-limiting.md
```diff
@@ -413,9 +413,9 @@ X-AI-RateLimit-Reset-deepseek-instance: 0
 
 ### Configure Instance Priority and Rate Limiting
 
-The following example demonstrates how you can configure two models with different priorities and apply rate limiting on the instance with a higher priority. In the case where `fallback_strategy` is set to `instance_health_and_rate_limiting`, the Plugin should continue to forward requests to the low priority instance once the high priority instance's rate limiting quota is fully consumed.
+The following example demonstrates how you can configure two models with different priorities and apply rate limiting on the instance with a higher priority. In the case where `fallback_strategy` is set to `["rate_limiting"]`, the Plugin should continue to forward requests to the low priority instance once the high priority instance's rate limiting quota is fully consumed.
```
kayx23, line 416: 1. Why is the example changed to use the array format? The string is also supported.
2. The same example is also in the `ai-proxy-multi` doc. If it needs to be changed, both should have the same change?

Revolyssup: Yes string is also supported but using array is recommended.

Revolyssup: okay i will change the other doc as well.

docs/en/latest/plugins/ai-proxy-multi.md
```diff
@@ -51,7 +51,7 @@ In addition, the Plugin also supports logging LLM request information in the acc
...
 |------------------------------------|----------------|----------|-----------------------------------|--------------|-------------|
-| fallback_strategy                  | string         | False    | instance_health_and_rate_limiting | instance_health_and_rate_limiting | Fallback strategy. When set, the Plugin will check whether the specified instance’s token has been exhausted when a request is forwarded. If so, forward the request to the next instance regardless of the instance priority. When not set, the Plugin will not forward the request to low priority instances when token of the high priority instance is exhausted. |
+| fallback_strategy                  | string or array         | False    |  | "instance_health_and_rate_limiting" "http_429", "http_5xx" or ["rate_limiting", "http_429", "http_5xx"] | Fallback strategy. When set, the Plugin will check whether the specified instance’s token has been exhausted when a request is forwarded. If so, forward the request to the next instance regardless of the instance priority. When not set, the Plugin will not forward the request to low priority instances when token of the high priority instance is exhausted. |
```
kayx23: The valid value format is slightly confusing, since this doc uses `[ ]` to list all the configurable values for strings as well. How about:

```suggestion
| fallback_strategy                  | string or array         | False    |  | string: "instance_health_and_rate_limiting", "http_429", "http_5xx"<br>array: ["rate_limiting", "http_429", "http_5xx"] | Fallback strategy. When set, the Plugin will check whether the specified instance’s token has been exhausted when a request is forwarded. If so, forward the request to the next instance regardless of the instance priority. When not set, the Plugin will not forward the request to low priority instances when token of the high priority instance is exhausted. |
```

kayx23: Maybe mention the option `instance_health_and_rate_limiting` is kept for backward compatibility and functionally, it is the same as `rate_limiting` as well?

kayx23:
```suggestion
| fallback_strategy                  | string or array         | False    |  | string: "instance_health_and_rate_limiting", "http_429", "http_5xx"<br>array: ["rate_limiting", "http_429", "http_5xx"] | Fallback strategy. The option `instance_health_and_rate_limiting` is kept for backward compatibility and is functionally the same as `rate_limiting`. With `rate_limiting` or `instance_health_and_rate_limiting`, when the current instance’s quota is exhausted, the request is forwarded to the next instance regardless of priority. With `http_429`, if an instance returns status code 429, the request is retried with other instances. With `http_5xx`, if an instance returns a 5xx status code, the request is retried with other instances. If all instances fail, the plugin returns the last error response code. When not set, the Plugin will not forward the request to low priority instances when token of the high priority instance is exhausted. |
```

docs/en/latest/plugins/ai-rate-limiting.md
kayx23: Sounds good. Just a reminder that there's more than one example to update.

kayx23 (approved):
doc changes lgtm

PR #2592, opened 2020-11-01
feat: implement `error-log-logger` plug-in

dabue:
### What this PR does / why we need it:

add a plugin to collect error log and report to remote TCP server .

fix #2410

you can set the plugin's config in config-default.yaml, likes below:
plugins:                          # plugin list
  - error-log-logger

and  update it links below:
step 1: update the attributes of the plugin

curl http://127.0.0.1:9080//apisix/admin/plugin_metadata/error-log-logger -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
  "host": "127.0.0.1",
  "port": 1999,
  "inactive_timeout": 1
}'

PR #12603, opened 2025-09-09
feat: admin api no longer populates default values when writing

SkyeYoung:
### Description

This PR removes the logic of populating default values before writing in the admin API.

This solves the following problems:

1. Modifying the user-provided configuration can easily lead to user
confusion, which is clearly not best practice.
2. This also affects the diff logic in the adc that the apisix ingress
controller depends on.

Since when using GET to retrieve data from the admin API, it will be read directly from etcd, the changes introduced by this PR are compatible with the old version.

Fixes #

### Checklist


apisix/admin/utils.lua
```diff
@@ -106,7 +106,7 @@ function _M.decrypt_params(decrypt_func, body, schema_type)
     -- metadata
     if schema_type == core.schema.TYPE_METADATA then
         local conf = body.node and body.node.value
-        decrypt_func(conf.name, conf, schema_type)
```
SkyeYoung, line 109: There is an amazing problem here:

In `error-log-logger.lua`, the following code exists:

https://github.com/apache/apisix/blob/36b2b83653adfa1ad0f45b8d4bb33686a6550550/apisix/plugins/error-log-logger.lua#L130

This is a property not marked in the document.

The intention of its design cannot be seen from the original PR https://github.com/apache/apisix/pull/2592.

At the same time, it is also the only one that contains `encrypt_fields` in `metadata_schema`.

https://github.com/apache/apisix/blob/36b2b83653adfa1ad0f45b8d4bb33686a6550550/apisix/plugins/error-log-logger.lua#L149

This causes this logic in the actual `decrypt_params` to take effect only for `error-log-logger.lua` when `schema_type == core.schema.TYPE_METADATA`.

However, due to the cancellation of the default value padding, this logic no longer takes effect.

Anyway, **I think this is wrong. And it needs to be repaired in subsequent PRs.** 

The reasons are as follows:

1. There should not be a hidden property in the schema that is used for plugin-unrelated logic.
2. https://github.com/apache/apisix/pull/11452 implements a standardized `id`, which can be consistent with other resources.

apisix/plugins/request-id.lua
```diff
@@ -51,7 +51,10 @@ local schema = {
                     default = "abcdefghijklmnopqrstuvwxyzABCDEFGHIGKLMNOPQRSTUVWXYZ0123456789"
                 }
             },
-            default = {}
+            default = {
+                length = 16,
+                char_set = "abcdefghijklmnopqrstuvwxyzABCDEFGHIGKLMNOPQRSTUVWXYZ0123456789"
+            }
```
SkyeYoung, line 57: The tests revealed that this is the only way to properly populate default values when `range_id` doesn't exist.

The original one should be because it's filled again after writing `default={}`.

t/admin/global-rules.t
```diff
@@ -504,3 +503,40 @@ GET /t
 {"deleted":"1","key":"/apisix/global_rules/1"}
 --- request
 GET /t
+
+
+
+=== TEST 15: not unwanted data, PUT with use_real_request_uri_unsafe
```
SkyeYoung, line 509: Add tests, when including property `use_real_request_uri_unsafe`, it will be filled normally.

t/admin/plugin-configs.t
```diff
@@ -521,3 +520,91 @@ passed
     }
 --- response_body
 passed
+
+
+
+=== TEST 15: PUT with limit-count policy=local
```
SkyeYoung, line 526: Add tests, when including property `policy`, it will be filled normally.

SkyeYoung: Because I found that most of the tests covered non-required fields, I didn't add similar tests later.

t/plugin/jwt-auth.t
```diff
@@ -1001,25 +1001,20 @@ base64_secret required but the secret is not in base64 format
...
-                            "key": "exp-not-set",
-                            "secret": "my-secret-key"
-                        }
-                    }
-                }]]
-            )
-
-            res = require("toolkit.json").decode(res)
-            assert(res.value.plugins["jwt-auth"].exp == 86400)
+            local core = require("apisix.core")
```
SkyeYoung, line 1004: Because the original method can no longer detect whether the `exp` variable exists, it's changed to direct validation through the schema, according to the internal logic.

apisix/admin/plugin_metadata.lua
```diff
@@ -45,13 +56,8 @@ local function check_conf(plugin_name, conf)
...
     end
 
-    if not plugin_object.metadata_schema then
-        plugin_object.metadata_schema = {
-            type = "object",
-            ['$comment'] = injected_mark,
-            properties = {},
-        }
-    end
+    inject_metadata_schema(plugin_object)
```
bzp2010, line 59: Although it is not introduced by your PR, it is strange to actively inject an empty schema and use ordinary schema or metadata_schema according to certain conditions for configuration verification for some non-plugin metadata.

If a plugin does not indicate that it supports plugin metadata configuration by actively declaring metadata_schema, it should not be configured in this way. Specifically, if this happens, it may be more appropriate to reject this configuration directly. For example, key-auth does not have the meaning of plugin metadata configuration.

apisix/admin/resource.lua
```diff
@@ -123,7 +124,12 @@ function _M:check_conf(id, conf, need_id, typ, allow_time)
         core.log.info("schema: ", core.json.delay_encode(self.schema))
     end
 
-    local ok, err = self.checker(id, conf, need_id, self.schema, {secret_type = typ})
+    local conf_for_check = tbl_deepcopy(conf)
+    local ok, err = self.checker(id, conf_for_check, need_id, self.schema, {secret_type = typ})
```
bzp2010, line 128: I'm wondering if it's necessary to keep a switch enabled by a specific querystring to retain the old behavior?

For example, when `fillin_default=true`, still fill in the default value? This may have some meaning for debugging, otherwise we have to rely only on the correctness of the document. Historical experience tells me that this is usually difficult.

SkyeYoung: This is actually a question I've thought about. Let's wait for other people's comments.

apisix/admin/plugin_metadata.lua
SkyeYoung: Let's wait for what others think.

apisix/admin/consumer_group.lua
```diff
@@ -38,6 +39,11 @@ local function check_conf(id, conf, need_id, schema)
 end
 
 
+local function encrypt_conf(id, conf)
+    plugins_encrypt_conf(conf.plugins)
```
membphis, line 43: pls confirm it never failed

for me, I prefer `return plugins_encrypt_conf(conf.plugins)`

membphis (approved):

apisix/admin/consumer_group.lua
SkyeYoung: CleanShot 2025-09-16 at 18 52 42@2x


It looks like they are the same?

SkyeYoung: CleanShot 2025-09-16 at 19 07 07@2x
CleanShot 2025-09-16 at 19 09 13@2x

Regarding the calling method, it references the original `local schema_plugin = require(“apisix.admin.plugins”).check_schema`. And in the current scenario, no issues have been detected.

SkyeYoung: In the current implementation, it will never fail:

https://github.com/apache/apisix/pull/12603/files#diff-00090cde519e7280adbc73e830b92089c4e62a22057c1b7577c77dc1456e72e5

apisix/admin/resource.lua
nic-6443: I can't think of any issues that require specifying `fillin_default=true` to troubleshoot, but adding this option doesn't really have any drawbacks either, making it difficult to decide.

apisix/admin/plugin_metadata.lua
nic-6443: agree with @bzp2010 , it's strange to inject a empty metadata schema for plugins that don't need plugin metadata, look like we need a refactor for this, I think we can deal with it in another PR.

apisix/admin/resource.lua
AlinsRan: In terms of debugging purposes only, this PR should not be blocked, and the default value can be confirmed through the version of apisix.

Issue #12556, opened 2025-08-28
bug: stream routes in the standalone API cannot be loaded correctly


bzp2010:
### Current Behavior

When submitting configurations containing stream routes via the standalone API, all configurations except the stream route load correctly, but the stream route does not.

### Expected Behavior

The submitted configuration can take effect correctly.

### Error Logs

N/A

### Steps to Reproduce

1. Enable the stream mode in APISIX and create a TCP proxy port for listening.
2. Use standalone mode with Admin API enabled.
3. Submit a configuration containing the stream route.
4. Access that port, and the proxy will not be effective.

---

The standalone API, as part of the Admin API, operates within the HTTP subsystem. When a worker receives an HTTP request, it notifies other workers via an event broadcasting mechanism. Upon receiving the broadcast, other workers load the new configuration.
The current event broadcasting mechanism correctly notifies the HTTP subsystems in other workers and loads configurations on their Lua VMs. However, the stream subsystem uses a different Lua VM, preventing configuration data sharing.
Additionally, the HTTP subsystem's event broadcasting and the stream subsystem's event broadcasting communicate over separate Unix domain sockets, inherently isolating them.
This results in the stream subsystem failing to receive notifications prompting it to load new configurations, thereby preventing stream routing from taking effect.

---

To address this issue and further ensure the reliability of the event notification mechanism, I believe we should introduce a background timer loop. This would allow the configuration refresh mechanism to passively take effect even if event broadcasting fails.

Specifically, we should add a timer to the standalone mode. This timer would periodically retrieve the latest configuration from shdict and check whether it is newer than the locally loaded configuration. If it is newer, the configuration would be reloaded.
Given our existing `X-Last-Modified` record, comparing differences requires no deep inspection—simply matching timestamps suffices.
This mechanism is lightweight and serves solely as a fallback when event notifications fail or when we intentionally update configurations from HTTP to the stream subsystem. We can run it every few hundred milliseconds or once per second. It will operate on both subsystems within each worker.

### Environment

- APISIX version (run `apisix version`): master
- Operating system (run `uname -a`): N/A
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): N/A
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): N/A
- APISIX Dashboard version, if relevant: N/A
- Plugin runner version, for issues related to plugin runners: N/A
- LuaRocks version, for installation issues (run `luarocks --version`): N/A

PR #12604, opened 2025-09-09
fix(standalone): support stream_route in admin api mode

nic-6443:

Fixes https://github.com/apache/apisix/issues/12556

### Checklist

### DIFF
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -32,8 +32,23 @@ local events       = require("apisix.events")
 local core         = require("apisix.core")
 local config_yaml  = require("apisix.core.config_yaml")
 local tbl_deepcopy = require("apisix.core.table").deepcopy
+local timers       = require("apisix.timers")
+local constants    = require("apisix.constants")
+
+-- combine all resources that using in http and stream substreams as one constant
+local CONF_VERSION_KEY_SUFFIX = "_conf_version"
+local ALL_RESOURCE_KEYS = {}
+for dir in pairs(constants.HTTP_ETCD_DIRECTORY) do
+    local key = str_sub(dir, 2)
+    ALL_RESOURCE_KEYS[key] = key .. CONF_VERSION_KEY_SUFFIX
+end
+for dir in pairs(constants.STREAM_ETCD_DIRECTORY) do
+    local key = str_sub(dir, 2)
+    ALL_RESOURCE_KEYS[key] = key .. CONF_VERSION_KEY_SUFFIX
+end
 
 local EVENT_UPDATE = "standalone-api-configuration-update"
+local NOT_FOUND_ERR = "not found"
 -- do not use the HTTP standard Last-Modified header to prevent affecting
 -- the caching implementation in the client
 local METADATA_LAST_MODIFIED = "X-Last-Modified"
@@ -79,7 +94,7 @@ end
 local function get_config()
     local config = shared_dict:get("config")
     if not config then
-        return nil, "not found"
+        return nil, NOT_FOUND_ERR
     end
 
     local err
@@ -180,7 +195,7 @@ local function update(ctx)
 
     local config, err = get_config()
     if not config then
-        if err ~= "not found" then
+        if err ~= NOT_FOUND_ERR then
             core.log.error("failed to get config from shared dict: ", err)
             return core.response.exit(500, {
                 error_msg = "failed to get config from shared dict: " .. err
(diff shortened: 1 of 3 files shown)


t/admin/standalone.t
```diff
@@ -91,7 +94,7 @@ X-Digest: t2
 
 === TEST 3: get config
 --- config
-    location /t {
+    location /config {
```
nic-6443, line 97: I modified the URL of this location because when `stream_enable` is used, test-nginx will occupy the `/t` URL to initiate stream requests.

apisix/admin/standalone.lua
```diff
@@ -420,6 +430,29 @@ function _M.init_worker()
     end
     events:register(update_config, EVENT_UPDATE, EVENT_UPDATE)
 
+    -- due to the event module can not broadcast events between http and stream subsystems,
+    -- we need to poll the shared dict to keep the config in sync
+    local last_sync_time = ngx_time()
```
AlinsRan: This timer is only applicable in stream mode. Should we check if stream is enabled?

nic-6443: ref from https://github.com/apache/apisix/issues/12556
> To address this issue and further ensure the reliability of the event notification mechanism, I believe we should introduce a background timer loop. This would allow the configuration refresh mechanism to passively take effect even if event broadcasting fails.

Since we check `METADATA_LAST_MODIFIED` and `last_sync_time` first, this timer is lightweight. I think we can run it in the HTTP subsystem too, ensuring config sync even if event broadcasting fails (not only for stream subsystem).

AlinsRan (dismissed):

apisix/admin/standalone.lua
```diff
@@ -420,6 +430,29 @@ function _M.init_worker()
...
+        if not config then
+            if err ~= NOT_FOUND_ERR then
+                core.log.error("failed to get config: ", err)
+            end
+        else
+            if config[METADATA_LAST_MODIFIED] > last_sync_time then
+                update_config(config)
+            end
+        end
+        last_sync_time = now
```
bzp2010: I thought maybe we could just store `METADATA_LAST_MODIFIED` in this local variable.
Then use the rule whether the shdict stored config is higher than the local timestamp to check if the config should be sync again.
We only need to track last modified and no longer need to record a new timestamp additionally or use the difference between last sync and now to check if it should be changed.

bzp2010: Event-triggered updates should also update the last sync time on each worker, so the last sync should be a file-level variable rather than in init_worker.
Each timer loop only needs to check whether the last modification stored in shdict is consistent with the modification time of the current worker to determine whether the timer needs to update again.
Since the worker is single-threaded, we don't need to worry about races. If an update is already triggered by an event, the timer loop will never intervene and repeated updates will not occur. Since UDS communication is almost always faster than our fixed-interval timer, most updates will be triggered and executed by events.
Only in rare cases, such as when a new update occurs immediately after the UDS event broadcast and the worker has not had time to receive it, will the timer detect the difference between the last sync time and shdict and perform an update. This is expected to be rare, and even if it happens, our design is sufficient to handle it correctly.

Therefore, we can assume that in most normal cases event broadcasting will take effect, and the last sync time on the worker will always be overwritten by the update triggered by the event and consistent with the shdict, so the timer loop will not actually work.
This timer only works when event broadcasting is abnormal or when broadcasting to stream from http side. In this case, since the last sync time is not overwritten by the event triggered update, the timer loop will actually update the configuration. In most cases, it is just a fallback.

nic-6443: > I thought maybe we could just store METADATA_LAST_MODIFIED in this local variable.
Then use the rule whether the shdict stored config is higher than the local timestamp to check if the config should be sync again.

However, if we remove `last_sync_time`, we will not be able to control this timer to run once per second. This means that apisix will loop through the execution of the two functions `shareddict:get` and `json.decode`. I think this will consume more computing resources than comparing two local variables.

bzp2010: > if config[METADATA_LAST_MODIFIED] > last_sync_time then

This introduces two distinct time coordinate systems and compares them. Although both are epoch timestamps, their starting points may be unreliable. For example, NTP might occasionally cause system time to roll back.
Each instance may have multiple workers, any of which could restart at any time. I believe this creates the potential for time inconsistencies across workers, which we can entirely avoid.

What I mean is that we should ideally use a single time coordinate system and avoid numerical comparisons whenever possible. That is, if we all use LAST_MODIFIED, I can simply perform equality checks. Timestamps should only ever be written by a single worker; other workers should not generate another timestamp for comparison. This is likely a more robust design.

---

I don't understand how `last_sync_time` affects the timer. Shouldn't we be able to simply register a fixed timer using `ngx.timer.every`? 

Regarding the overhead of `shdict:get`, this is primarily memory-related, with the overhead occurring in the locking portion. If you believe these timer-driven `get` operations might also incur unexpected overhead due to lock waits, I think we could introduce some random delays before `shdict:get`. This would spread out the `get` calls as much as possible, striving to avoid the lock being held at that exact moment.

Regarding the overhead of JSON decode, if this is a concern, we can change the encoding method for shdict to use text encoding like `<LAST_MODIFIED>|<DIGEST>|<JSON>`. This way, data extraction can be done using PCRE, which can be JIT optimized.

nic-6443: got it, `ngx.timer.every` is a good idea, applied your suggestion, please review again

AlinsRan (approved):

Revolyssup (approved):

PR #12605, opened 2025-09-09
fix(limit-req): Make Redis path atomic via EVAL + use hash key with TTL

### LINKED ISSUE #12592: bug: limit-req plugin does not work correctly when configuring redis or redis-cluster
### Current Behavior

When we use limit-req plugin with `policy:local` it works as expected.
BUT when we configure `policy:redis` or `policy:redis-cluster` the results are totally random and unexpected.

### Expected Behavior

I expect that limit-req plugin works in the same way when using `policy:local`, `policy:redis` and `policy:redis-cluster`.

### Error Logs

_No response_

### Steps to Reproduce

1. Run Apisix with the docker image apisix:3.12.0
2. Run 1 redis instance  (or a redis cluster) using Helm chart https://artifacthub.io/packages/helm/bitnami/redis or https://artifacthub.io/packages/helm/bitnami/redis-cluster
3. Create a consumer (in this case anonymous but it also fails with authenticated consumers) with limit-req plugin (that uses redis)
```
{
  "username": "anonymous",
  "desc": "",
  "plugins": {
    "limit-req": {
      "_meta": {
        "disable": false
      },
      "allow_degradation": false,
      "burst": 0,
      "key": "remote_addr",
      "key_type": "var",
      "policy": "redis",
      "rate": 1,
      "redis_database": 0,
      "redis_host": "redis-master.ingress-apisix.svc.cluster.local",
      "redis_timeout": 1000,
      "rejected_code": 429,
      "rejected_msg": "Too many requests in time window"
    }
  }
}
```
4. Configure an upstream
5. Configure a route using JWT plugin:
```
{
  "uris": [
    "/test/*"
  ],
  "name": "test-service",
  "desc": "",
  "plugins": {
    "jwt-auth": {
      "anonymous_consumer": "anonymous",
      "cookie": "jwt",
      "header": "authorization",
      "hide_credentials": false,
      "key_claim_name": "key",
      "query": "jwt",
      "store_in_ctx": false
    },
    }
  },
  "labels": {
    "managed-by": "apisix-ingress-controller"
  }
}
```
7. Send requests to the defined route and check how the RPS are not being limited to 1 RPS, so 🔴 
8. After checking point 7, change the consumer to use limit-req plugin with policy: local`:
```
{
  "username": "anonymous",
  "desc": "",
  "plugins": {
    "limit-req": {
      "_meta": {
        "disable": false
      },
      "allow_degradation": false,
      "burst": 0,
      "key": "remote_addr",
      "key_type": "var",
      "policy": "local",
      "rate": 1,
      "rejected_code": 429,
      "rejected_msg": "Too many requests in time window"
    }
  }
}
```
9. Send requests to the defined route and check how the RPS are now being limited to 1 RPS, so 🟢 

### Environment

- APISIX version (run `apisix version`): apisix:3.12.0

falvaradorodriguez:
### Description

The current `limit-req` Redis implementation uses two separate keys (excess and last) and updates them with multiple GET/SET operations.

- Under concurrent load, this leads to race conditions:
1. Several workers may read stale values in parallel and overwrite each other.
2. As a result, the plugin allows more requests than expected, effectively bypassing the intended rate limit.

- On Redis Cluster, the current approach is also problematic: atomic EVAL cannot be executed across two different keys located on different slots.

**Solution**

- Store both values (excess and last) under a single Redis hash key, so the state is managed as one unit.

- Use a single EVAL script that performs read → compute → write atomically inside Redis, removing race conditions. This approach is consistent with how the `limit-count` plugin already works.

- Add a TTL to the key to avoid buildup of stale state.

- Preserve existing semantics: the first request with no prior state does not consume tokens.

Fixes #12592

### Checklist


Baoyuantop:
‌Hi @falvaradorodriguez, we need to add the test case for this fix.

Baoyuantop:
Hi @falvaradorodriguez, there are failed CI that need fixing.

falvaradorodriguez:
> Hi @falvaradorodriguez, there are failed CI that need fixing.

Hi @Baoyuantop!

The failures are not related to the current changes. Tests of the modified plugin seems to be fine.

Please, Can you review?

Thanks

```
[15:23:12] t/plugin/limit-req-redis-cluster.t ......... ok    15865 ms ( 0.01 usr  0.01 sys +  0.91 cusr  2.03 csys =  2.96 CPU)
[15:23:29] t/plugin/limit-req-redis.t ................. ok    16682 ms ( 0.02 usr  0.00 sys +  0.95 cusr  2.23 csys =  3.20 CPU)
2025/09/22 15:23:37 Processed 0 requests
[15:23:41] t/plugin/limit-req.t ....................... ok    11856 ms ( 0.02 usr  0.00 sys +  0.83 cusr  1.78 csys =  2.63 CPU)
[15:23:46] t/plugin/limit-req2.t ...................... ok     5010 ms ( 0.01 usr  0.00 sys +  0.81 cusr  0.42 csys =  1.24 CPU)
[15:23:47] t/plugin/limit-req3.t ...................... ok     1290 ms ( 0.00 usr  0.00 sys +  0.39 cusr  0.22 csys =  0.61 CPU)
```

apisix/plugins/limit-req/util.lua
```diff
@@ -14,60 +14,66 @@
...
+    new_excess = 0
+  end
+
+  if new_excess > burst then
+    return {0, new_excess}
+  end
+
+  if commit == 1 then
+    redis.call("HMSET", state_key, "excess", new_excess, "last", now)
+    redis.call("EXPIRE", state_key, 60)
```
Baoyuantop: Can TTL be dynamically configured based on the rate-limiting window period, or allow user configuration?

falvaradorodriguez: In the limit-req plugin, there is no time window like in the limit-count plugin. 

This configuration is simply done to prevent keys from remaining dead in redis. Currently, if a consumer stops making requests, their key remains in redis until it is deleted by an action independent of Apisix.

It could be made configurable, but in my opinion, it would not add much value to this plugin. It would also be an extra feature, not related with this fix.

In my opinion, since limit-req is always for 1 second, with a 1 minute of margin in redis, it is acceptable. If the consumer makes requests after 1 minute, the flow management would be the same as for the first request.

luarx:
When will this PR be merged? I am having the same issue that this PR fixes 🙏

Baoyuantop:
I will promptly urge other community maintainers to review.

apisix/plugins/limit-req/util.lua
```diff
@@ -14,60 +14,66 @@
...
 
-    local excess, err = red:get(excess_key)
-    if err then
-        return nil, err
-    end
-    local last, err = red:get(last_key)
-    if err then
+    local commit_flag = commit and "1" or "0"
+
+    local res, err = red:eval(script, 1, state_key,
```
membphis: we can use `evalsha`, avoid to repeat pass the `script` every time

official website: https://redis.io/docs/latest/commands/evalsha

here is a demo:

```lua
local redis = require "resty.redis"

-- Create redis instance
local red = redis:new()

-- Connect to Redis
local ok, err = red:connect("127.0.0.1", 6379)
if not ok then
    ngx.say("failed to connect: ", err)
    return
end

-- Step 1: Load script and get SHA1
local script = "return redis.call('get', KEYS[1])"
local sha1, err = red:script("load", script)
if not sha1 then
    ngx.say("failed to load script: ", err)
    return
end

ngx.say("Script SHA1: ", sha1)

-- Step 2: Set a test key
red:set("mykey", "hello " .. ngx.time())

-- Step 3: Use evalsha to execute script
-- evalsha <sha1> <numkeys> <key1> [key2 ... ] [arg1 arg2 ...]
local res, err = red:evalsha(sha1, 1, "mykey")
if not res then
    ngx.say("failed to evalsha: ", err)
    return
end

ngx.say("Result: ", res)  -- Output: hello

-- Put connection back to the pool
local ok, err = red:set_keepalive(10000, 100)
if not ok then
    ngx.say("failed to set keepalive: ", err)
    return
end
```

we can cache the `sha1` result

falvaradorodriguez: Thanks @membphis, makes all sense, updated https://github.com/apache/apisix/pull/12605/changes/8263caa4b8831212189b0bda73856e8d16b53bfb

Baoyuantop:
Hi @falvaradorodriguez, it appears there are test failures in CI that need to be fixed.

falvaradorodriguez:
> Hi @falvaradorodriguez, it appears there are test failures in CI that need to be fixed.

There was a node management issue with Redis.cluster. It seems that loading the script does not guarantee that it will be available in the next sha1 evaluation. I have changed the fallback to eval to ensure that the script is executed. Once executed with eval, it can be checked with sha1.

Baoyuantop:
Hi @falvaradorodriguez, could you please fix the lint bug?

apisix/plugins/limit-req/util.lua
```diff
@@ -14,61 +14,97 @@
...
+    new_excess = 0
+  end
+
+  if new_excess > burst then
+    return {0, new_excess}
+  end
+
+  if commit == 1 then
+    redis.call("HMSET", state_key, "excess", new_excess, "last", now)
+    local ttl = math.ceil(burst / rate) + 1
```
shreemaan-abhishek: nice!

apisix/plugins/limit-req/util.lua
```diff
@@ -14,61 +14,97 @@
...
+
+
+local function is_noscript_error(err)
+    if not err then
+        return false
+    end
+
+    local s
+    if type(err) == "table" then
+        s = tostring(err[1] or err.err or err.message or err.msg or err)
```
shreemaan-abhishek: last time I checked, the type of `err` was always a string. Did you notice something different that you added `err.err or err.message or err.msg or err`?

apisix/plugins/limit-req/util.lua
```diff
@@ -14,61 +14,97 @@
...
+    local res, err = red:evalsha(script_sha1, 1, state_key,
+                                rate, now, self.burst, commit_flag)
 
-        if excess > self.burst then
-            return nil, "rejected"
-        end
-    else
-        excess = 0
+    -- If the script isn't cached on this Redis node, fall back to EVAL.
+    if not res and is_noscript_error(err) then
```
shreemaan-abhishek: I observed that reliability of redis evalsha with redis-cluster was very low.
https://github.com/apache/apisix/pull/12872#discussion_r2727073505

So I decided to not use `evalsha` when the `policy` is set to `rediscluster`. You can refer the PR I just linked to see how I did it.

Let me know if you have any better ideas tho.

shreemaan-abhishek (commented):
please resolve the conflicts too.

apisix/plugins/limit-req/util.lua
falvaradorodriguez: ~~Thanks for the note. In resty.redis the error is indeed typically returned as a plain string.~~

~~However, since this code path is meant to work with both resty.redis and resty.rediscluster, I added a small defensive handling for the case where errors may be propagated as a structured table (e.g. containing err/message fields) instead of a raw string.~~

~~This doesn’t change the behavior for the common case, but makes the NOSCRIPT detection more robust across Redis standalone and Redis Cluster setups, especially during redirects/failovers where error formats may differ.~~

apisix/plugins/limit-req/util.lua
falvaradorodriguez: Makes sense, and it is also good practice to keep the logic of the plugins aligned. I have tried to replicate the functionality you mention here: https://github.com/apache/apisix/pull/12605/changes/05657f67151416b4af9bd37937bc3b76e9fa5b27

apisix/plugins/limit-req/util.lua
falvaradorodriguez: With the changes you mentioned [here](https://github.com/apache/apisix/pull/12605#discussion_r2734766405), what I wrote above doesn't makes sense. I have replicated the way you handle the error you use in the `limit-conn` plugin.

moonming (changes requested):
Hi @falvaradorodriguez, thank you for making the Redis limit-req path atomic via EVAL!

Using a Lua script with EVAL to ensure atomicity of the rate limiting operation is the correct approach — the current non-atomic multi-command flow can have race conditions under high concurrency. With 12 reviews, this has been thoroughly discussed.

**To confirm readiness**:
1. Are all 12 review comments addressed?
2. Has the EVAL script been tested under concurrent load to verify it resolves the race condition?
3. The hash key with TTL approach — does it handle key expiration correctly for sliding windows?

This is an important correctness fix. Let's get it finalized! Thank you.

Baoyuantop:
Hi @falvaradorodriguez, I see no problems with the code here. Could you please fix the failing test?

falvaradorodriguez:
Hi @moonming.

I’m replying on your comment:

> Hi @falvaradorodriguez, thank you for making the Redis limit-req path atomic via EVAL!
> 
> Using a Lua script with EVAL to ensure atomicity of the rate limiting operation is the correct approach — the current non-atomic multi-command flow can have race conditions under high concurrency. With 12 reviews, this has been thoroughly discussed.
> 
> **To confirm readiness**:
> 
> 1. Are all 12 review comments addressed?

Yes, they have all been checked and corrected where necessary.

> 2. Has the EVAL script been tested under concurrent load to verify it resolves the race condition?


Yes, in fact, in my case it has been validated in an environment with a high volume of requests

> 3. The hash key with TTL approach — does it handle key expiration correctly for sliding windows?

Yes, also validated.

> This is an important correctness fix. Let's get it finalized! Thank you.

Thank you for the review!

falvaradorodriguez:
> Hi @falvaradorodriguez, I see no problems with the code here. Could you please fix the failing test?

Hi @Baoyuantop!

As you can see [here](https://github.com/apache/apisix/actions/runs/21439970632/job/64626319434?pr=12605), the tests that were failing were not related to this PR. 

Screenshot 2026-03-31 at 13 21 33
Screenshot 2026-03-31 at 13 21 22

Could you check that again?

Thanks

Baoyuantop:
Okay, I'll check it after CI finishes executing.

Baoyuantop:
It seems the failed CI is unrelated to this PR; I'll try to fix it.

apisix/plugins/limit-req/util.lua
```diff
@@ -14,61 +14,103 @@
...
+  end
+
+  if new_excess > burst then
+    return {0, new_excess}
+  end
+
+  if commit == 1 then
+    redis.call("HMSET", state_key, "excess", new_excess, "last", now)
+    local ttl = math.ceil(burst / rate) + 1
+    redis.call("EXPIRE", state_key, ttl)
```
nic-6443, line 51: TTL is too short when `burst=0` with sub-1 request rates.

With `rate=0.1, burst=0` (scaled: `rate=100, burst=0`): `TTL = ceil(0/100) + 1 = 1 second`. But the leaky bucket drain time is `1000 / rate = 10s`. After 1 second the key expires, resetting the limiter — effectively allowing 1 req/s instead of the configured 0.1 req/s (10× the intended rate).

The fix accounts for the per-request excess increment (`+1000` in scaled units):
```lua
local ttl = math.ceil((burst + 1000) / rate) + 1
```
For `rate=0.1, burst=0`: TTL = `ceil(1000/100) + 1 = 11s` — correctly covers the drain window.
For `rate=1, burst=0`: TTL = `ceil(1000/1000) + 1 = 2s` — still correct.
For `rate=1, burst=10000`: TTL = `12s` vs old `11s` — negligible difference.

This is pre-existing (same formula in old code), but since the PR's goal is fixing rate-limiting correctness, worth addressing here.

falvaradorodriguez: Fixed by @Baoyuantop here https://github.com/apache/apisix/pull/12605/commits/4deb878f622068dcd6db800aa2fc4fb70b0cb9a5

AlinsRan:
Closing this — the underlying issue (#12592) was fixed and merged in #13467, which takes the same approach: single hash key + one atomic `EVAL` for the read-decay-check-write cycle, with a TTL and first-request admission preserved.

The one thing this PR adds on top is the `EVALSHA` fast path with `NOSCRIPT` fallback. That's a nice bandwidth optimization but not needed for correctness — happy to see it as a small follow-up PR if you want to pursue it.

Thanks for the work here!

PR #12607, opened 2025-09-10
fix: healthcheck manager missing runtime information

Revolyssup:
- In case of service discovery, the node_ver information is not attached to the service received from etcd (due to deepcopy) therefore it's lost and healthchecker is not recreated.
- In case of dns, dns_value is not attached to the data from etcd for similar reason.

Solution:
- Since we use the data from etcd in healthcheck, we need the runtime information like nodes_ver and dns_value to be stored. This PR refactors the code so that those runtime updated values are persisted for future requests via healthcheck module. 
### Checklist


apisix/core.lua
```diff
@@ -36,15 +39,96 @@ end
...
+    upstream.nodes = nodes
+end
 
 return {
     version     = require("apisix.core.version"),
     log         = log,
     config      = config,
     config_util = require("apisix.core.config_util"),
     sleep       = utils.sleep,
+    fetch_latest_conf = fetch_latest_conf,
```
membphis: in my opinion:

create a new `module`, it's name can be `resource`

then we can use it by call `core.resource.get_nodes_ver`, what do you think? @Revolyssup

apisix/core/config_util.lua
```diff
@@ -215,5 +214,11 @@ function _M.parse_time_unit(s)
     return size
 end
 
+function _M.upstream_version(index, nodes_ver)
```
membphis: move this function to `upstream` lua module

apisix/core.lua
Revolyssup: done

apisix/core/config_util.lua
Revolyssup: done. moved to utils/upstream

nic-6443 (approved):

AlinsRan (approved):

PR #12611, opened 2025-09-12
change(jwt-auth): when algorithm is not `RS256` or `ES256`, require the user to fill in `secret`

SkyeYoung:
### Description

This PR is a part of #12603 and is separated out due to breaking changes.

Currently, the jwt-auth plugin generates a random value for
`conf.secret` in the `check_schema` function when `conf.algorithm ~=
"RS256" and conf.algorithm ~= "ES256" and not conf.secret`.

Modifying the user-provided configuration can easily lead to user
confusion, which is clearly not best practice.

This also affects the diff logic in the adc that the apisix ingress
controller depends on.

I believe this generation behavior should be removed. To solve this problem, This PR removed the corresponding code and
instead return an error message, requiring users to fill in the
corresponding configuration themselves.

Fixes #

### Checklist

	- https://lists.apache.org/thread/0q7f4llo5ofd8zg6oblsbrxq80obdcyn

### DIFF
--- a/t/plugin/jwt-auth.t
+++ b/t/plugin/jwt-auth.t
@@ -1242,7 +1242,7 @@ hello world
                 algorithm = "HS256",
             }, core.schema.TYPE_CONSUMER)
             assert(not ok, "secret should be required when algorithm is HS256")
-            
+
             ok, err = plugin.check_schema({
                 key = "123",
                 algorithm = "HS512",


apisix/plugins/jwt-auth.lua
```diff
@@ -145,7 +143,8 @@ function _M.check_schema(conf, schema_type)
     end
 
     if conf.algorithm ~= "RS256" and conf.algorithm ~= "ES256" and not conf.secret then
-        conf.secret = ngx_encode_base64(resty_random.bytes(32, true))
+        err = "property \"secret\" is required when \"algorithm\" is not \"RS256\" or \"ES256\""
```
nic-6443: for this plugin, we already use json schema to define relationship between `algorithm` and `public_key` field, I think better to use same way for `secret` too:
https://github.com/apache/apisix/blob/69564959d99e42ec2095eab2c35b3c73bc31b259/apisix/plugins/jwt-auth.lua#L95-L117

SkyeYoung: Let me check it

SkyeYoung: CleanShot 2025-09-12 at 11 37 13@2x

https://github.com/apache/apisix/pull/12611/files#diff-9b9604832706062cf84d2e9b7a13754c0f21a9a0b34cf271a4eb059f4bffbdaa TEST 52

After trying to modify it, I found that it would incorrectly pass the test when going through the default value logic.

The original code should have also been implemented through code for this reason.

apisix/plugins/jwt-auth.lua
```diff
@@ -144,8 +142,9 @@ function _M.check_schema(conf, schema_type)
         return false, err
     end
 
-    if conf.algorithm ~= "RS256" and conf.algorithm ~= "ES256" and not conf.secret then
-        conf.secret = ngx_encode_base64(resty_random.bytes(32, true))
+    if (conf.algorithm == "HS256" or conf.algorithm == "HS512") and not conf.secret then
```
SkyeYoung, line 145: Adjust to be consistent with the document

https://github.com/apache/apisix/pull/12611/files#diff-3cd5fc846614f1e59961d065c6f32dca0b2d6cc3ef549e7a961dd4b3694a74bc

Image

apisix/plugins/jwt-auth.lua
```diff
@@ -144,8 +142,9 @@ function _M.check_schema(conf, schema_type)
...
+        err = "property \"secret\" is required when \"algorithm\" is \"HS256\" or \"HS512\""
```
bzp2010: There seems to be no point in using a variable to store it; returning the string directly should suffice.

t/plugin/jwt-auth.t
```diff
@@ -1222,3 +1222,71 @@ hello world
...
+            local plugin = require("apisix.plugins.jwt-auth")
+            -- default algorithm is HS256
+            local ok, err = plugin.check_schema({
+                key = "123",
+            }, core.schema.TYPE_CONSUMER)
+            if not ok then
+                ngx.say(err)
+            else
+                ngx.say("done")
+            end
```
bzp2010: Just a suggestion: using assert is enough.

```
assert(ok, "HS256 but secret is not required")
```

apisix/plugins/jwt-auth.lua
SkyeYoung: done

t/plugin/jwt-auth.t
SkyeYoung: done

SkyeYoung:
> change

@bzp2010 This is not very common. I usually refer to https://github.com/TrigenSoftware/simple-release/blob/main/GUIDE.md

Since this feature did not exist before, it should be `feat` 🤔

bzp2010:
If the `change` prefix is not used, special attention must be paid to these changes during release and they must be marked as non-backward-compatible modifications.
This may cause trouble in the next release process.

PR #12629, opened 2025-09-23
fix: redact encrypted fields from error log

Revolyssup:
Currently sensitive encrypted fields are directly logged in plugins. This PR removes from all the places where sensitive information from plugins could be logged.
### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -47,7 +47,6 @@ local debug           = require("apisix.debug")
 local pubsub_kafka    = require("apisix.pubsub.kafka")
 local resource        = require("apisix.resource")
 local trusted_addresses_util = require("apisix.utils.trusted-addresses")
-local redact_encrypted = require("apisix.core.utils").redact_encrypted
 local ngx             = ngx
 local get_method      = ngx.req.get_method
 local ngx_exit        = ngx.exit


apisix/balancer.lua
```diff
@@ -194,7 +195,20 @@ end
...
+        local schema
+        if plugin.type == "auth" then
+            schema = plugin.consumer_schema
+        else
+            schema = plugin.schema
+        end
+        redacted_ctx.var._ctx.consumer.plugins[name] = redact_encrypted(conf, schema)
+        local redacted_auth = redact_encrypted(redacted_ctx.var._ctx.consumer.auth_conf, schema)
+        redacted_ctx.var._ctx.consumer.auth_conf = redacted_auth
+    end
```
nic-6443: this code snippet repeated multiple times, we should put it into a util function to reuse it.

apisix/balancer.lua
```diff
@@ -194,7 +195,20 @@ end
...
+    core.log.info("ctx: ", core.json.delay_encode(redacted_ctx, true))
```
nic-6443: we should avoid to redact the log that will not be printed like `json.delay_encode`, my suggestion:
```
local max_delay_encode_items = 16
local delay_tab_idx = 0
local delay_tab_arr = {}
for i = 1, max_delay_encode_items do
    delay_tab_arr[i] = setmetatable({data = "", force = false}, {
        __tostring = function(self)
            if self.redact_func then
                self.data = deepcopy(self.data)
                self.redact_func(self.data)
            end
            local res, err = encode(self.data, self.force)
            if not res then
                ngx.log(ngx.WARN, "failed to encode: ", err,
                        " force: ", self.force)
            end

            return res
        end
    })
end


function _M.delay_encode(data, force, redact_func)
    delay_tab_idx = delay_tab_idx+1
    if delay_tab_idx > max_delay_encode_items then
        delay_tab_idx = 1
    end
    delay_tab_arr[delay_tab_idx].data = data
    delay_tab_arr[delay_tab_idx].force = force
    delay_tab_arr[delay_tab_idx].redact_func = redact_func
    return delay_tab_arr[delay_tab_idx]
end
```
only execute `redact_func` when this line of log be printed actually.

apisix/admin/resource.lua
```diff
@@ -117,8 +118,23 @@ function _M:check_conf(id, conf, need_id, typ, allow_time)
         end
     end
 
-    core.log.info("conf  : ", core.json.delay_encode(conf))
```
membphis, line 120: we can remove this line directly

which is easy and quick

t/cli/test_etcd_sync_event_handle.sh
```diff
@@ -127,7 +127,7 @@ cat logs/error.log | grep "watchdir err: has no healthy etcd endpoint available"
 ##   }
 ## }
 ## After check, it only appears when watch recovers and returns events in bulk.
-cat logs/error.log | grep "}, {" || (echo "failed: Log case 2 unexpected"; exit 1)
+cat logs/error.log | grep "etcd events sent in bulk" || (echo "failed: Log case 2 unexpected"; exit 1)
```
Revolyssup, line 130: Since the entire res object is no longer logged. We programmatically assert in code and log this specific condition to make this test case pass.

apisix/core/config_etcd.lua
```diff
@@ -698,8 +699,9 @@ local function sync_data(self)
 
     local dir_res, err = waitdir(self)
     log.info("waitdir key: ", self.key, " prev_index: ", self.prev_index + 1)
-    log.info("res: ", json.delay_encode(dir_res, true), ", err: ", err)
-
```
Revolyssup, line 702: This log contained sensitive information and was being used to assert if there is a bulk operation. So the entire log is replaced with just the log about whether etcd events were sent in bulk.

apisix/consumer.lua
```diff
@@ -244,11 +245,25 @@ local function fill_consumer_secret(consumer)
 end
 
 
-function create_consume_cache(consumers_conf, key_attr)
+function create_consume_cache(plugin_name, consumers_conf, key_attr)
     local consumer_names = {}
 
     for _, consumer in ipairs(consumers_conf.nodes) do
-        core.log.info("consumer node: ", core.json.delay_encode(consumer))
```
membphis, line 251: can we just remove this line? which is easier to implement

apisix/core/json.lua
```diff
@@ -22,6 +22,7 @@
 local cjson = require("cjson.safe")
 local json_encode = cjson.encode
 local clear_tab = require("table.clear")
+local table = require("apisix.core.table")
```
membphis: we do not need to update this library if we only remove some `core.log(....)`

apisix/init.lua
```diff
@@ -267,7 +269,27 @@ local function parse_domain_in_route(route)
     resource.set_nodes_ver_and_nodes(route.value.upstream.resource_key,
                                                     nodes_ver, new_nodes)
     core.log.info("parse route which contain domain: ",
-                  core.json.delay_encode(route, true))
+                core.json.delay_encode(route, true, function (route)
```
membphis: same issue, we can remove this `core.log.info`

apisix/plugin.lua
```diff
@@ -733,14 +735,33 @@ local function merge_consumer_route(route_conf, consumer_conf, consumer_group_co
...
 
-    core.log.info("merged conf : ", core.json.delay_encode(new_route_conf))
     return new_route_conf
 end
 
 
 function _M.merge_consumer_route(route_conf, consumer_conf, consumer_group_conf, api_ctx)
     core.log.info("route conf: ", core.json.delay_encode(route_conf))
-    core.log.info("consumer conf: ", core.json.delay_encode(consumer_conf))
+    core.log.info("consumer conf: ", core.json.delay_encode(consumer_conf, false,
```
membphis: ditto

apisix/plugin.lua
```diff
@@ -885,8 +906,19 @@ end
...
+        if plugin.type == "auth" then
+            schema = plugin.consumer_schema
+        else
+            schema = plugin.schema
+        end
+        redacted_plugin_conf = redact_encrypted(plugin_conf, schema)
+    end
     core.log.info("check plugin schema, name: ", name, ", configurations: ",
-        core.json.delay_encode(plugin_conf, true))
+        core.json.delay_encode(redacted_plugin_conf, true))
```
membphis: ditto

apisix/plugins/basic-auth.lua
```diff
@@ -160,7 +161,7 @@ end
 
 
 function _M.rewrite(conf, ctx)
-    core.log.info("plugin access phase, conf: ", core.json.delay_encode(conf))
+    core.log.info("plugin access phase, conf: ", core.json.delay_encode(conf, false))
```
membphis: ditto

apisix/plugins/hmac-auth.lua
```diff
@@ -111,13 +111,16 @@ end
...
+    local conf_to_log = conf
     if schema_type == core.schema.TYPE_CONSUMER then
-        return core.schema.check(consumer_schema, conf)
+        cur_schema = consumer_schema
+        conf_to_log = core.utils.redact_encrypted(conf, consumer_schema)
     else
-        return core.schema.check(schema, conf)
+        cur_schema = schema
     end
+    core.log.info("input conf: ", core.json.delay_encode(conf_to_log, false))
```
membphis: ditto

apisix/plugins/hmac-auth.lua
```diff
@@ -130,8 +133,11 @@ local function get_consumer(key_id)
     if not cur_consumer then
         return nil, err or "Invalid key_id"
     end
-    core.log.info("consumer: ", core.json.delay_encode(consumer, true))
 
+    core.log.info("consumer: ", core.json.delay_encode(consumer, true, function (cur_consumer)
```
membphis: ditto

apisix/plugins/jwe-decrypt.lua
```diff
@@ -174,8 +175,16 @@ local function get_consumer(key)
     if not consumers then
         return nil
     end
-    core.log.info("consumers: ", core.json.delay_encode(consumers))
-    return consumers[key]
+    local consumer = consumers[key]
```
membphis: ditto

apisix/init.lua
```diff
@@ -266,8 +268,12 @@ local function parse_domain_in_route(route)
     route.value.upstream.nodes = new_nodes
     resource.set_nodes_ver_and_nodes(route.value.upstream.resource_key,
                                                     nodes_ver, new_nodes)
+    -- remove plugin before logging to avoid logging sensitive info
+    local route_log = core.table.deepcopy(route)
```
Revolyssup, line 271: this log is needed for test so only plugins are removed.

membphis (approved):
LGTM

nic-6443 (approved):

apisix/plugin.lua
```diff
@@ -733,14 +734,12 @@ local function merge_consumer_route(route_conf, consumer_conf, consumer_group_co
         new_route_conf.value.plugins[name] = conf
     end
 
-    core.log.info("merged conf : ", core.json.delay_encode(new_route_conf))
     return new_route_conf
 end
 
 
 function _M.merge_consumer_route(route_conf, consumer_conf, consumer_group_conf, api_ctx)
     core.log.info("route conf: ", core.json.delay_encode(route_conf))
```
AlinsRan, line 742: Does route_conf not print sensitive data?

Revolyssup: Before merging, consumer_conf has the sensitive data. it would have been caught in test if route conf had sensitive data,.

PR #12631, opened 2025-09-23
feat: Implement etcd-logger plugin

mdnfiras:
### Description

This PR implements a new plugin: `etcd-logger`.
APISIX already uses ETCD, it would be very convenient to have the possibility log to ETCD as well.

Note: I tested the plugin manually, but generated the tests in this PR using Copilot.

none

### Checklist


Baoyuantop:
Hi @mdnfiras, do you need to use etcd as a logger? 

Technically speaking, etcd is not suitable for log storage.

Writes to etcd must undergo Raft consensus, requiring majority node confirmation for all write operations. This results in high write latency and limited throughput (suitable for configuration/metadata but unsuitable for high-frequency writes like logs). Log systems typically require thousands to millions of writes per second, which etcd cannot handle.

When etcd's data volume becomes excessively large, performance degrades sharply and may even compromise cluster stability. The official recommendation is to keep data volumes at the GB level, whereas logs routinely reach TB/PB scales.

etcd only supports range queries based on keys and lacks the search, filtering, and aggregation capabilities required for log analysis.

Could you elaborate on your proposal?

mdnfiras:
That makes sense.
My goal was to implement audit logging for the Apisix dashboard. And for this, I'm routing Apisix dashboard requests through Apisix gateway itself, and using this `etcd-logger` to log to ETCD. So in my case, there won't be a huge amount of logs.

But perhaps it makes more sense to properly implement audit logging for Apisix Admin API, right? Instead of working around this by routing request through the API gateway then using plugins, it would be better to have the option to just enable audit logging for POST/PUT/DELETE/PATCH requests to the admin API. These audit logs should not be as large as regular request logs, and can possibly fit in ETCD. What do you think?

mdnfiras:
Should I close this PR and open a discussion about implementing audit logs?

Baoyuantop:
Hi @mdnfiras, I'm afraid the community won't be able to accept the etcd logging part of this PR, as it's not a very useful scenario.

> But perhaps it makes more sense to properly implement audit logging for Apisix Admin API, right?

Yes, I agree with this point of view. Maybe we can implement an audit plugin for apisix to record the call information of admin api or other APIs and save this information to any logging configured by the user.

We can create an issue to discuss this proposal in detail.

mdnfiras:
closing this, and opening a new issue for audit logging: https://github.com/apache/apisix/issues/12640

PR #12634, opened 2025-09-24
fix: correct handling of endpointSlices in Kubernetes service discovery

ChuanFF:
### Description

Currently, the Kubernetes service discovery implementation in APISIX incorrectly handles EndpointSlices, treating each individual slice as the complete set of endpoints for a service. This contradicts the official Kubernetes documentation.

According to the [Kubernetes EndpointSlice documentation](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoint-slice-v1/):
> "For a given service there may be multiple EndpointSlice objects which must be joined to produce the full set of endpoints; you can find all of the slices for a given service by listing EndpointSlices in the service's namespace whose `kubernetes.io/service-name` label contains the service's name."

The correct behavior should be to combine all EndpointSlices sharing the same `kubernetes.io/service-name` label to form the complete endpoint set for a service.

#### 1. Add EndpointSlices Cache
- Implemented `endpoint_slices_cache` to store all EndpointSlices with the structure: 
  `endpoint_slices_cache["k8s_id/namespace/k8s_service_name:port_name"] = {slice1, slice2, ...}`
- Any changes to EndpointSlices trigger updates to this cache
- The cache is then used to assemble the complete node list for each service, which is subsequently updated in shared memory

#### 2. Code Refactoring
- Renamed variables to better align with Kubernetes EndpointSlices data structure terminology
- Improved code readability and maintainability

#### 3. Some Bug Fixes
- **Fixed incorrect handling of `addresses` field**: The `addresses` field contains a list of IP strings, not objects with `.ip` attributes
- **Optimized node sorting logic**: Simplified the nested loop structure in `endpoint_buffer` processing:
  ```lua
  -- Before (incorrect):
  for _, ports in pairs(endpoint_buffer) do
      for _, nodes in pairs(ports) do
          core.table.sort(nodes, sort_nodes_cmp)
      end
  end
  
  -- After (correct):
  for _, nodes in pairs(endpoint_buffer) do
      core.table.sort(nodes, sort_nodes_cmp)
  end
  ```
  The `endpoint_buffer` uses ports as keys and node lists as values, requiring only a single iteration level.

### Checklist

### DIFF
--- a/apisix/discovery/kubernetes/init.lua
+++ b/apisix/discovery/kubernetes/init.lua
@@ -125,7 +125,7 @@ local function on_endpoint_slices_modified(handle, endpoint_slice, operate)
         return
     end
 
-    core.log.debug(operate .. " endpoint_slice: ", core.json.delay_encode(endpoint_slice))
+    core.log.debug("get endpoint_slice: ", core.json.delay_encode(endpoint_slice))
     --record nodes to every port in service
     local port_to_nodes = {}
 


Baoyuantop:
Hi @ChuanFF, you can resolve the conflict by merging the code in the main branch.

apisix/discovery/kubernetes/init.lua
```diff
@@ -51,77 +51,172 @@ local function sort_nodes_cmp(left, right)
...
+local function update_endpoint_dict(handle, endpoints, endpoint_key)
+    local endpoint_content = core.json.encode(endpoints, true)
     local endpoint_version = ngx.crc32_long(endpoint_content)
-
     local _, err
     _, err = handle.endpoint_dict:safe_set(endpoint_key .. "#version", endpoint_version)
     if err then
         core.log.error("set endpoint version into discovery DICT failed, ", err)
-        return
+        return false, err
```
nic-6443: This return value is not used. Additionally, when this function executes normally, there is a missing return value.

apisix/discovery/kubernetes/init.lua
```diff
@@ -51,77 +51,172 @@ local function sort_nodes_cmp(left, right)
...
+                core.json.delay_encode(endpoint_slice))
+        return
+    end
+    if not endpoint_slice.metadata.namespace then
+        core.log.error("endpoint_slice has no metadata.namespace, endpointSlice: ",
+                core.json.delay_encode(endpoint_slice))
+        return
+    end
+    if not endpoint_slice.metadata.labels
+            or not endpoint_slice.metadata.labels["kubernetes.io/service-name"] then
```
nic-6443: Define `"kubernetes.io/service-name"` as a constant

apisix/discovery/kubernetes/init.lua
```diff
@@ -51,77 +51,172 @@ local function sort_nodes_cmp(left, right)
...
+        core.log.error("endpoint_slice has no service-name, endpointSlice: ",
+                core.json.delay_encode(endpoint_slice))
+        return
```
nic-6443: This part of the code is completely duplicated with that in `on_endpoint_slices_modified` and can be extracted into an independent function.

apisix/discovery/kubernetes/init.lua
ChuanFF: done, i check the return value at the call site; if it's false, I log an error and return.

apisix/discovery/kubernetes/init.lua
ChuanFF: done

apisix/discovery/kubernetes/init.lua
ChuanFF: done

ChuanFF:
The code changes have been completed; please help review the code again.

apisix/discovery/kubernetes/init.lua
```diff
@@ -51,69 +52,133 @@ local function sort_nodes_cmp(left, right)
...
+    if not ok then
+        core.log.error("endpoint_slice validation fail: ", err)
+        return
+    end
+    if handle.namespace_selector and
+            not handle:namespace_selector(endpoint_slice.metadata.namespace) then
+        return
+    end
+
+    core.log.debug(core.json.delay_encode(endpoint_slice))
```
membphis: we can add more text to explain this error message
make it easier to understand

apisix/discovery/kubernetes/init.lua
```diff
@@ -122,6 +187,36 @@ local function on_endpoint_slices_modified(handle, endpoint, operate)
...
+        core.log.error("endpoint_slice validation fail: ", err)
+        return
+    end
+
+    if handle.namespace_selector and
+            not handle:namespace_selector(endpoint_slice.metadata.namespace) then
+        return
+    end
+
+    core.log.debug(core.json.delay_encode(endpoint_slice))
```
membphis: ditto

apisix/discovery/kubernetes/init.lua
ChuanFF: done

nic-chen (approved):

PR #12656, opened 2025-10-08
chore: add test for verifying lua-resty-openssl bug fix

Revolyssup:
### Description
Previous version of lua-resty-openssl introduced a bug in jwt verification in how the Ed448 public key is being handled by lua-resty-openssl (base64 encode-decode). The newer version of APISIX uses recent version of lua-resty-openssl which doesn't have this bug. The following added test just confirms the fix.

With version lua-resty-openssl 1.6.1
Screenshot From 2025-10-08 22-43-33

### Checklist


Revolyssup:
Failng test unrelated/flaky,

PR #12682, opened 2025-10-16
refactor(improvement): use secret URI as key for cache and refactor lrucache

Revolyssup:
1. Secret module changes: Currently the cache key is the whole conf which is inefficient because multiple conf might be referencing the same secret.
So new requirement is to refactor apisix/secret.lua so that cache key is the url and we only cache the result of fetch(uri) function.

2. LRUCache module changes: Allow LRU cache to have different underlying cache for success and failure, so that each of them can be configured differently.

NOTE: The fetch_secrets() API is simplified to not take key and version from the caller because they were redundant anyways and moreover now the key is resolved internally.

### Checklist

### DIFF
--- a/t/core/lrucache2.t
+++ b/t/core/lrucache2.t
@@ -25,7 +25,6 @@ run_tests;
 
 __DATA__
 
-
 === TEST 1: negative cache basic functionality
 --- config
     location /t {


apisix/secret.lua
```diff
@@ -136,7 +136,7 @@ local function parse_secret_uri(secret_uri)
 end
 
 
-local function fetch_by_uri(secret_uri)
+local function fetch_by_uri_secret(secret_uri)
```
nic-6443, line 139: Is this change necessary?

Revolyssup: I am using a new local function with same name here https://github.com/apache/apisix/pull/12682/files#diff-50c5241a560560f8d61637b60caa3e9f2dce27139863265cfa8ffb63a8537803R212 so this function name got changed to be more specific to avoid conflict.

membphis (approved):

AlinsRan (approved):

PR #12696, opened 2025-10-23
feat(log): add custom variables for enhanced logging capabilities

felpasl:
#### Introduced a comprehensive set of custom logging variables as functions, including:

- latency: Total request latency in milliseconds
- upstream_latency: Time spent communicating with upstream servers
- apisix_latency: APISIX processing time (total latency minus upstream latency)
- client_ip: Remote client IP address
- start_time: Request start time in milliseconds
- hostname: Server hostname
- version: APISIX version
- request_method: HTTP method
- request_headers: Request headers
- request_querystring: Query parameters
- request_body: Request body (with size limits)
- response_status: HTTP response status
- response_headers: Response headers
- response_size: Response size in bytes
- response_body: Response body
- upstream: Upstream server address
- url: Full request URL
- consumer_username: Consumer username
- service_id: Associated service ID
- route_id: Associated route ID

Fixes #

### Checklist

### DIFF
--- a/apisix/utils/log-util.lua
+++ b/apisix/utils/log-util.lua
@@ -92,7 +92,6 @@ local log_vars = {
         local upstream_latency =
             ctx.var.upstream_response_time and ctx.var.upstream_response_time * 1000 or 0
         local apisix_latency = latency - upstream_latency
-        
         -- The latency might be negative, as Nginx uses different time measurements in
         -- different metrics.
         -- See https://github.com/apache/apisix/issues/5146#issuecomment-928919399


Baoyuantop:
We have some built-in variables and all nginx variables are supported. Can you explain in detail why this change is needed?

https://apisix.apache.org/docs/apisix/apisix-variable/
https://nginx.org/en/docs/varindex.html

felpasl:
The goal here is to give users more freedom and control over how logs are handled.

Right now, the default log includes useful fields but it also bundles headers and other request/response data that I don’t want to store due to privacy concerns and storage overhead.

I want to be able to:

- Send audit logs to Loki
- Stream consumption data to Kafka

Customize log formats based on destination and purpose

To do that, I need those key fields (latency, hostname, method, version, client IP, etc.) to be available in log_format, so I can use them in custom_log without relying on the default log structure. This change enables that flexibility while keeping observability intact and avoiding unnecessary data.

Let me know if you have concerns or suggestions

Baoyuantop:
Hi @felpasl, we support customizing log formats directly in the log plugin, and support all built-in variables. Does this not meet your needs?

https://apisix.apache.org/docs/apisix/plugins/kafka-logger/#metadata
https://apisix.apache.org/docs/apisix/plugins/loki-logger/#plugin-metadata

felpasl:
Thanks for the follow-up. Just to clarify: the change includes methods and calculations that aren't available through existing NGINX or APISIX variables.

For example:

``` lua
    latency = function() return (ngx_now() - ngx.req.start_time()) * 1000 end,
    upstream_latency = function(ctx)
        return ctx.var.upstream_response_time and ctx.var.upstream_response_time * 1000 or nil
    end,
    apisix_latency = function(ctx)
        local latency = (ngx_now() - ngx.req.start_time()) * 1000
        local upstream_latency =
            ctx.var.upstream_response_time and ctx.var.upstream_response_time * 1000 or 0
        local apisix_latency = latency - upstream_latency
        if apisix_latency < 0 then apisix_latency = 0 end
        return apisix_latency
    end,
    hostname = function() return core.utils.gethostname() end,
    version = function() return core.version.VERSION end,
```
These aren't just references to existing variables — they encapsulate logic that calculates values like total latency, APISIX processing time, and upstream latency. My proposal is to expose these as named variables so users can access them directly in log_format, without needing to replicate the logic themselves.

While many fields (like client_ip, request_method, etc.) are already covered by NGINX variables but with other names and having them available as variables would simplify custom logging and improve consistency of naming variables between default log and log_format across plugins.

test-nginx
Baoyuantop, line 1: Is this change necessary?

apisix/utils/log-util.lua
```diff
@@ -82,23 +82,74 @@ local function gen_log_format(format)
...
+        return string.format("%s://%s:%s%s",
+            var.scheme,
+            var.host,
+            var.server_port,
+            var.request_uri)
+    end,
+    consumer_username = function(ctx)
+        return ctx.consumer and ctx.consumer.username or nil
+    end,
+    service_id = function(ctx)
```
Baoyuantop, line 133: In the original get_full_log logic, if no route is matched (matched_route is empty), service_id will fall back to var.host.

apisix/utils/log-util.lua
```diff
@@ -82,23 +82,74 @@ local function gen_log_format(format)
     return log_format
 end
 
+local log_vars = {
+    latency = function() return (ngx_now() - ngx.req.start_time()) * 1000 end,
```
Baoyuantop, line 86: The calculation logic for latency, upstream_latency, and apisix_latency has been reimplemented in log_vars. This is completely duplicated from the existing latency_details_in_ms function logic.

Baoyuantop:
Hi @felpasl, following up on the previous review comments. Let us know if you have an update. Thanks!

Baoyuantop:
Due to a lack of activity, this PR will be closed for now. We will reopen it if there are any questions.

PR #12714, opened 2025-10-30
fix(performance): move the ipv6 check to schema validation

Revolyssup:
There is a big performance difference between APISIX 3.10 and 2.13 when no plug-ins are enabled.
After comparing the flame graphs of APISIX 3.10 and 2.13 versions, it can be clearly seen that the biggest difference is in the pick_server function. After comparing the function codes of the two versions, it is found that the code of version 3.10 has added logic to traverse all upstream nodes, and when there are more upstream nodes during performance testing, so the difference is more obvious.

After backtracking the changes, it was found that the changes traversing all upstream nodes were introduced by this pr https://github.com/apache/apisix/pull/7594.

change point
1. Remove https://github.com/apache/apisix/blob/73618d48f32e263228f1a0a0df71fbc3ec5d9a54/apisix/balancer.lua#L198-L202 from the request path, and escape it to upstream's `check_schema` function to implement this logic
2. Illegal ipv6 addresses are directly rejected in the admin api

Note: This change is backwards compatible. When detecting data from etcd, it modifies the host for compatibility
```log
2025/10/30 16:17:45 [warn] 2277252#2277252: *1 [lua] upstream.lua:408: checker(): PREV CONF.NODES : { {
    host = "::1",
    port = 80,
    priority = 0,
    weight = 1
  },
  <metatable> = {}
}, context: init_worker_by_lua*
2025/10/30 16:17:45 [warn] 2277252#2277252: *1 [lua] upstream.lua:414: checker(): AFTER CONF.NODES : { {
    host = "[::1]",
    port = 80,
    priority = 0,
    weight = 1
  },
  <metatable> = {}
```

### DIFF
--- a/t/node/upstream-ipv6.t
+++ b/t/node/upstream-ipv6.t
@@ -115,7 +115,7 @@ hello world
                         {
                             "weight": 100,
                             "priority": 0,
-                            "host": "::1",
+                            "host": "[::1]",
                             "port": 1980
                         }
                     ],
@@ -221,7 +221,7 @@ passed
 GET /hello
 --- error_code: 502
 --- error_log
-connect() to [::0.1.25.128]:80 failed
+failed to set server peer [::1:1980:80] err: invalid port while connecting to upstream
 
 
 
@@ -237,13 +237,13 @@ connect() to [::0.1.25.128]:80 failed
                         {
                             "weight": 100,
                             "priority": 0,
-                            "host": "::1",
+                            "host": "[::1]",
                             "port": 1980
                         },
                         {
                             "weight": 100,
                             "priority": 0,
-                            "host": "::1",
+                            "host": "[::1]",
                             "port": 1980
                         }
                     ],


apisix/upstream.lua
```diff
@@ -326,6 +327,13 @@ end
 
 
 function _M.check_schema(conf)
+    if conf.nodes then
+        for _, node in ipairs(conf.nodes) do
```
nic-6443:
```suggestion
        for _, node in ipairs(conf.nodes or {}) do
```

apisix/upstream.lua
```diff
@@ -326,6 +327,13 @@ end
...
+            if core.utils.parse_ipv6(node.host) and str_byte(node.host, 1) ~= str_byte("[") then
+                return false, "IPv6 address must be enclosed with '[' and ']'"
```
nic-6443: We not only need to reject such configurations in the admin API, but also consider executing data correction code when watching for such configurations from etcd:
```
node.host = '[' .. node.host .. ']'
```

membphis (commented):
more test cases:

call Admin API, to confirm the `schema` check is working right

apisix/upstream.lua
Revolyssup: done

Revolyssup:
> more test cases:
> 
> call Admin API, to confirm the `schema` check is working right

done

apisix/upstream.lua
Revolyssup: done

membphis (approved):

apisix/upstream.lua
```diff
@@ -357,6 +363,11 @@ end
 
 local function check_upstream_conf(in_dp, conf)
     if not in_dp then
+        for _, node in ipairs(conf.nodes or {}) do
+            if core.utils.parse_ipv6(node.host) and str_byte(node.host, 1) ~= str_byte("[") then
+                return false, "IPv6 address must be enclosed with '[' and ']'"
+            end
+        end
         local ok, err = core.schema.check(core.schema.upstream, conf)
```
nic-6443:
```suggestion
        local ok, err = check_schema(conf)
```

nic-6443 (dismissed):

PR #12718, opened 2025-10-30
feat: add validate API to standalone mode

Revolyssup:
Provide a validate API for standalone mode that only validate input like admin api but does not save input to memory.
Required for ingress controller to validate plugin configurations that convert from CRD.

### Checklist

### DIFF
--- a/apisix/admin/standalone.lua
+++ b/apisix/admin/standalone.lua
@@ -158,10 +158,8 @@ end
 
 
 local function validate_configuration(req_body, collect_all_errors)
-    local validation_results = {
-        valid = true,
-        errors = {}
-    }
+    local is_valid = true
+    local validation_results = {}
 
     for key, conf_version_key in pairs(ALL_RESOURCE_KEYS) do
         local items = req_body[key]
@@ -173,8 +171,8 @@ local function validate_configuration(req_body, collect_all_errors)
             if not collect_all_errors then
                 return false, conf_version_key .. " must be a number"
             end
-            validation_results.valid = false
-            table_insert(validation_results.errors, {
+            is_valid = false
+            table_insert(validation_results, {
                 resource_type = key,
                 error = conf_version_key .. " must be a number, got " .. type(new_conf_version)
             })
@@ -196,8 +194,8 @@ local function validate_configuration(req_body, collect_all_errors)
                     if not collect_all_errors then
                         return false, error_msg
                     end
-                    validation_results.valid = false
-                    table_insert(validation_results.errors, {
+                    is_valid = false
+                    table_insert(validation_results, {
                         resource_type = key,
                         index = index - 1,
                         error = error_msg
@@ -210,8 +208,8 @@ local function validate_configuration(req_body, collect_all_errors)
                     if not collect_all_errors then
                         return false, dup_err
                     end
-                    validation_results.valid = false
-                    table_insert(validation_results.errors, {
+                    is_valid = false
+                    table_insert(validation_results, {
                         resource_type = key,
                         index = index - 1,
                         error = dup_err
@@ -222,10 +220,10 @@ local function validate_configuration(req_body, collect_all_errors)
     end
 
     if collect_all_errors then
-        return validation_results.valid, validation_results
-    else
-        return validation_results.valid, nil
+        return is_valid, validation_results
     end
+
+    return is_valid, nil
 end
 
 local function validate(ctx)
@@ -260,7 +258,7 @@ local function validate(ctx)
     if not valid then
         return core.response.exit(400, {
             error_msg = "Configuration validation failed",
-            errors = validation_results.errors
+            errors = validation_results
         })
     end
 


apisix/admin/standalone.lua
```diff
@@ -157,6 +158,104 @@ local function check_conf(checker, schema, item, typ)
     })
 end
 
+local function validate(ctx)
```
nic-6443: can we reuse this new `validate` function in exist `update` function to reduce code duplicate between those two functions?

apisix/admin/standalone.lua
```diff
@@ -211,54 +333,44 @@ local function update(ctx)
...
-                local valid, err = check_conf(item_checker, item_schema, item_temp, key)
-                if not valid then
-                    local err_prefix = "invalid " .. key .. " at index " .. (index - 1) .. ", err: "
-                    local err_msg = type(err) == "table" and err.error_msg or err
-                    core.response.exit(400, { error_msg = err_prefix .. err_msg })
-                end
+            for _, item in ipairs(items) do
                 -- prevent updating resource with the same ID
                 -- (e.g., service ID or other resource IDs) in a single request
                 local duplicated, err = check_duplicate(item, key, id_set)
```
AlinsRan: The logic of `check_deuplicate` exists twice, the first time in `validate_configuration`, and the second time is redundant.

apisix/admin/standalone.lua
```diff
@@ -347,20 +457,24 @@ local function head(ctx)
     return core.response.exit(200)
 end
 
-
 function _M.run()
```
membphis, line 430: another Lua style:

```lua
function _M.run()
    local ctx = ngx.ctx.api_ctx
    local method = str_lower(get_method())
    if method == "put" then
        return update(ctx)
    end

    if method == "post" then
        local path = ctx.var.uri
        if path == "/apisix/admin/configs/validate" then
            return validate(ctx)
        else
            return core.response.exit(404, {error_msg = "Not found"})
        end
    end

    if method == "head" then
        return head(ctx)
    end

    -- default is get
    return get(ctx)
end

Comment (AlinsRan) [reply apisix/admin/standalone.lua:430]: fixed.

**Comment by AlinsRan** [comment]
The CI failure is not related to the current PR modification.

File: apisix/admin/standalone.lua [hunk 4]
```diff
@@ -158,6 +157,128 @@ local function check_conf(checker, schema, item, typ)
...
+                        resource_type = key,
+                        index = index - 1,
+                        error = dup_err
+                    })
+                end
+            end
+        end
+    end
+
+    if collect_all_errors then
```
membphis, line 222: code style

```lua
    if collect_all_errors then
        return validation_results.valid, validation_results
    end
    
    return validation_results.valid

File: apisix/admin/standalone.lua [hunk 5]
```diff
@@ -158,6 +157,128 @@ local function check_conf(checker, schema, item, typ)
...
+    end
+
+    if err then
+        core.log.error("invalid request body: ", req_body, " err: ", err)
+        return core.response.exit(400, {error_msg = "invalid request body: " .. err})
+    end
+
+    local valid, validation_results = validate_configuration(data, true)
+
+    if valid then
```
membphis:
```lua
    if not valid then
        return core.response.exit(400, {
            error_msg = "Configuration validation failed",
            valid = false,
            errors = validation_results.errors
        })
    end
        
    return core.response.exit(200, {
          message = "Configuration is valid",
          valid = true
      })

**Review by membphis — APPROVED** [review]
LGTM, only some minor code style issues

File: apisix/admin/standalone.lua (hunk 4 above)
Comment (AlinsRan) [reply apisix/admin/standalone.lua:222]: fixed.

File: apisix/admin/standalone.lua (hunk 5 above)
Comment (AlinsRan) [reply apisix/admin/standalone.lua]: fixed.

File: apisix/admin/standalone.lua [hunk 6]
```diff
@@ -158,6 +157,120 @@ local function check_conf(checker, schema, item, typ)
...
+            error_msg = "Configuration validation failed",
+            valid = false,
+            errors = validation_results.errors
+        })
+    end
+
+    return core.response.exit(200, {
+        message = "Configuration is valid",
+        valid = true
+    })
```
bzp2010: When the check passes, neither the `message` nor the `valid` field carries any meaning; an HTTP 200 response is sufficient to imply this outcome. In practice, neither of these return fields will be utilised anywhere. **Remove them.**

When the check fails, the `valid` field similarly holds no significance. An HTTP 400 response can indicate this error. **Remove it.** The `error_msg` should maintain a consistent style, **using lowercase initial letters.**

t/admin/standalone.spec.ts
```diff
@@ -194,6 +194,86 @@ describe('Admin - Standalone', () => {
...
+
+    it('validate config does not persist changes', async () => {
+      // First validate a configuration
+      const validateResp = await client.post(`${ENDPOINT}/validate`, config1);
+      expect(validateResp.status).toEqual(200);
+
+      // Then check that the configuration was not persisted
+      const getResp = await client.get(ENDPOINT);
+      expect(getResp.data.routes).toBeUndefined();
+    });
```
bzp2010: I recommend moving these validate API tests to a separate describe block, which will help achieve a clearer separation.

t/admin/standalone.spec.ts
```diff
@@ -194,6 +194,86 @@ describe('Admin - Standalone', () => {
...
+        message: 'Configuration is valid',
+        valid: true,
+      });
+    });
+
+    it('validate config with consumer credentials', async () => {
+      const resp = await client.post(`${ENDPOINT}/validate`, credential1);
+      expect(resp.status).toEqual(200);
+      expect(resp.data).toEqual({
+        message: 'Configuration is valid',
```
bzp2010: This `message` does not need assertion, as checking such constant strings is meaningless. You may switch to the `toMatchObject` API to assert fields other than `message`. Only fields expected to exist in the configuration will be checked.

If no further assertion of the response is required, then only assert the response status code.

t/admin/standalone.spec.ts
```diff
@@ -641,5 +721,228 @@ describe('Admin - Standalone', () => {
           'invalid routes at index 0, err: invalid configuration: failed to match pattern "^((uri|server_name|server_addr|request_uri|remote_port|remote_addr|query_string|host|hostname|mqtt_client_id)|arg_[0-9a-zA-z_-]+)$" with "args_invalid"',
       });
     });
+    it('validate config (duplicate route id)', async () => {
```
bzp2010, line 730: What is the purpose of these newly added tests? Did we previously have duplicate ID checking functionality without test coverage?

t/admin/standalone.spec.ts
AlinsRan: fixed.

t/admin/standalone.spec.ts
AlinsRan: fixed.

apisix/admin/standalone.lua
AlinsRan: fixed.

t/admin/standalone.spec.ts
AlinsRan: The target API for testing is different, here we are testing the validate API.

PR #12720, opened 2025-11-01
fix: make protocol_name optional in mqtt-proxy plugin

### LINKED ISSUE #11589: help request: why configuring `protocol_name` in the `mqtt-proxy` plugin
### Description

Plugin doc: https://apisix.apache.org/docs/apisix/plugins/mqtt-proxy/

Currently `protocol_name` is a required parameter and the doc says it's _generally_ set to `MQTT`. Upon checking the test cases and code, it is still not clear to me why this is a configurable parameter. Why shouldn't it be hard-coded and what other value than `MQTT` could it be?

### Environment

- APISIX version (run `apisix version`): 3.10.0

tysoncung:
This PR makes the `protocol_name` parameter optional in the mqtt-proxy plugin, as requested in #11589.

Currently, `protocol_name` is required even though it's almost always set to `MQTT`. This creates unnecessary configuration overhead for users.

### Changes:
- Made `protocol_name` optional in the plugin schema with default value `MQTT`
- Updated validation logic to use the default when not specified
- Updated English documentation to reflect optional status
- Updated Chinese documentation to reflect optional status

### Pre-submission checklist:
- [x] Did you explain what problem does this PR solve? Or what new features have been added?
- [x] Have you added corresponding test cases?
- [x] Have you modified the corresponding document?
- [x] Is this PR backward compatible?

### Backward Compatibility:
✅ This change is fully backward compatible. Existing configurations that specify `protocol_name` will continue to work exactly as before.

Fixes #11589

### DIFF
--- a/apisix/stream/plugins/mqtt-proxy.lua
+++ b/apisix/stream/plugins/mqtt-proxy.lua
@@ -29,10 +29,10 @@ end)
 local schema = {
     type = "object",
     properties = {
-        protocol_name = {type = "string"},
+        protocol_name = {type = "string", default = "MQTT"},
         protocol_level = {type = "integer"}
     },
-    required = {"protocol_name", "protocol_level"},
+    required = {"protocol_level"},
 }
 
 
@@ -156,8 +156,9 @@ function _M.preread(conf, ctx)
         return 503
     end
 
-    if res.protocol and res.protocol ~= conf.protocol_name then
-        core.log.error("expect protocol name: ", conf.protocol_name,
+    local protocol_name = conf.protocol_name or "MQTT"
+    if res.protocol and res.protocol ~= protocol_name then
+        core.log.error("expect protocol name: ", protocol_name,
                        ", but got ", res.protocol)
         return 503
     end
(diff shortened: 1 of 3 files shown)


apisix/stream/plugins/mqtt-proxy.lua
```diff
@@ -156,8 +156,9 @@ function _M.preread(conf, ctx)
         return 503
     end
 
-    if res.protocol and res.protocol ~= conf.protocol_name then
-        core.log.error("expect protocol name: ", conf.protocol_name,
+    local protocol_name = conf.protocol_name or "MQTT"
```
Baoyuantop, line 159: No default value is needed here since we've already defined the default value in the schema.

Baoyuantop: Hi @tysoncung, please take a look

tysoncung:
Sorry for the delay on this. You're right that tests are needed for the optional parameter case.

I haven't had time to work on this yet - if someone else wants to take it over or if you want to add the test cases directly, feel free. Otherwise I can try to get to it but can't promise a timeline right now.

Appreciate the patience!

tysoncung:
Closing this PR as I don't have capacity to add the requested tests in a reasonable timeframe. The change itself is straightforward (making protocol_name optional), but I understand the importance of proper test coverage.

If anyone wants to pick this up, the change is just updating the schema and docs - the tests would need to verify the plugin works correctly when protocol_name is omitted.

Thanks for the review and apologies for not being able to complete this.

PR #12751, opened 2025-11-16
feat: ai rate limiting redis support

### LINKED ISSUE #12482: help request: How to config ai-rate-limiting plugin's  counter to redis, as  limit-count plugin
### Description

I have 2 apisix replicas , so I can not counter the tokens in local memory, please  help, thank you!

### Environment

beardnick:

Fixes #12482

##### Notice

- I have updated the `limit-count-redis.lua` and `limit-count-redis-cluster.lua` files to ensure they now support rate-limiting during the log phase.  
- I referred to the `limit-conn-redis.lua` file as a guide while implementing rate-limiting in the log phase.  
- To ensure a clean testing environment, I added the `require("lib.test_redis").flush_all()` function to every Redis rate-limiting test case. Without this addition, the tests could fail unpredictably.  
- I have adjusted the expected results in `limit-req-redis.t` and `limit-req-redis-cluster.t` test cases. After thorough verification, the correct behavior is `[200, 403, 403, 403]`. Previously, the results appeared as `[403, 403, 403, 403]` due to Redis not being properly cleaned beforehand.  

### Checklist


beardnick:
@Baoyuantop PTAL

membphis:
@beardnick many thx for your contribution, some CI tests are failed, you can fix them

apisix/plugins/limit-count/util.lua
```diff
@@ -0,0 +1,85 @@
+--
```
membphis, line 1: `apisix/plugins/limit-count/util.lua`

I checked `_M.incoming` and `_M.log_phase_incoming`, and they both work for `redis`.

We should add `redis` to the filename of the function name.

apisix/plugins/limit-count/init.lua
```diff
@@ -275,11 +277,24 @@ function _M.rate_limit(conf, ctx, name, cost, dry_run)
     key = gen_limit_key(conf, ctx, key)
     core.log.info("limit key: ", key)
 
+    local phase = get_phase()
+    if phase == "log" then
+        if not conf.policy or conf.policy == "local" then
+            lim:incoming(key, not dry_run, conf, cost)
```
AlinsRan: This conditional statement is unnecessary, as an identical check is performed again a few lines later:
```lua
local delay, remaining, reset
    if not conf.policy or conf.policy == "local" then
        delay, remaining, reset = lim:incoming(key, not dry_run, conf, cost)
    else
        delay, remaining, reset = lim:incoming(key, cost, dry_run)
    end
```

And why do we need to perform two incoming operations during the log phase?

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -57,26 +47,33 @@ function _M.new(plugin_name, limit, window, conf)
...
+    return util.redis_incoming(self, self.red_cli, key, commit, cost)
+end
 
-    if err then
-        return nil, err, ttl
-    end
 
-    local remaining = res[1]
-    ttl = res[2]
+local function log_phase_incoming_thread(premature, self, key, cost)
```
nic-6443, line 50: This code should be placed in the ai-rate-limiting plugin, not in the limit-count plugin itself, because this code is only useful for ai-rate-limiting.

apisix/plugins/limit-count/util.lua
```diff
@@ -0,0 +1,87 @@
...
+    if ttl < 0 then
+        return {limit - cost, tonumber(ARGV[2])}
+    end
+    local current = redis.call('get', KEYS[1])
+    if not current then
+        return {limit - cost, tonumber(ARGV[2])}
+    end
+    return {tonumber(current) - cost, ttl}
+]=])
+
```
nic-6443: Is it necessary to use a separate Redis script? Can we just remove 
```
assert(tonumber(ARGV[3]) >= 1, "cost must be at least 1")
```
from the `commit_script` directly?

t/plugin/limit-conn-redis-cluster.t
```diff
@@ -51,6 +51,7 @@ add_block_preprocessor(sub {
 
     location /test_concurrency {
         content_by_lua_block {
+            require("lib.test_redis").flush_all()
```
nic-6443: Why were we able to pass the tests without this flush before?

nic-6443 (commented):
Look like the changes in this PR should be limited to the `ai-rate-limiting` plugin and minor tweaks to the `limit-count` plugin. The current PR modifies too much code and test cases for other rate limiting plugins, expanding its scope. We should minimize unnecessary code changes.

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -57,26 +47,33 @@ function _M.new(plugin_name, limit, window, conf)
...
-    key = self.plugin_name .. tostring(key)
+function _M.incoming(self, key, cost, dry_run)
+    local commit = true
+    if dry_run ~= nil then
+        commit = not dry_run
+    end
 
-    local ttl = 0
-    local res, err = red:eval(script, 1, key, limit, window, cost or 1)
+    return util.redis_incoming(self, self.red_cli, key, commit, cost)
```
nic-6443: Can we directly check `ngx.phase` here? This way, the APIs of `limit-count-redis-cluster.lua` and `limit-count-redis.lua` do not need to be changed, still only providing an `incoming` function without introducing a new `log_phase_incoming`.

t/plugin/limit-conn-redis-cluster.t
beardnick: The tests succeed due to luck. They won't be successful every time. I've tested this numerous times and the results are inconsistent. You can confirm this by checking the GitHub Actions history.
https://github.com/apache/apisix/actions/runs/19564429993/job/56023083356
image
https://github.com/apache/apisix/actions/runs/19129422988/job/54666333550
image

beardnick:
> Look like the changes in this PR should be limited to the `ai-rate-limiting` plugin and minor tweaks to the `limit-count` plugin. The current PR modifies too much code and test cases for other rate limiting plugins, expanding its scope. We should minimize unnecessary code changes.

I agree, there have been many changes. I'll focus on including only the essential modifications in this PR and will submit a separate PR to enhance the tests for the other rate-limiting plugins.

apisix/plugins/limit-count/limit-count-redis-cluster.lua
beardnick: Yes, this is a better solution.

apisix/plugins/limit-count/init.lua
beardnick: These changes is not needed now. I moved these logic to limit-count-redis.lua and limit-count-redis-cluster.lua

apisix/plugins/limit-count/util.lua
beardnick: Yes, we can do this. We should change this assertion to `assert(tonumber(ARGV[3]) >= 0, "cost must be at least 0")`

apisix/plugins/limit-count/limit-count-redis-cluster.lua
beardnick: I disagree. Similar logic also appears in the limit-conn.
https://github.com/apache/apisix/blob/7e907a565b9302e70daf294398722377fd64c482/apisix/plugins/limit-conn/limit-conn-redis.lua#L60-L81

 If I put this logic into ai-rate-limiting plugin, it will make the plugin too complicated. I have to copy a lot of logic from limit-count. Currently, the ai-rate-limiting is just a simple wrapper of limit-count.

Or, do you think I need to rewrite the ai-rate-limiting to something like limit-ai-redis.lua and limit-ai-redis-cluster.lua?

beardnick: This PR is quite large and hard to review. If you agree, I can split it into 3 separate PRs:  
1. A PR to improve the tests for the rate-limiting plugins.  
2. A PR to add support for the log phase.  
3. A PR to add Redis support to the ai-rate-limiting.

t/plugin/ai-rate-limiting.t
```diff
@@ -984,3 +984,223 @@ passed
 Authorization: Bearer token
 --- error_code eval
 [200, 200, 200, 200, 200, 200, 200, 503, 503]
+
+
+
+=== TEST 21: set route with Redis policy
```
AlinsRan: It is recommended to use multi worker testing for Redis related tasks in the new test file `ai-rate-limiting-redis.t`.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -52,37 +44,66 @@ function _M.new(plugin_name, limit, window, conf)
...
+    if not red then
+        return red, err, 0
+    end
+
+    local commit = true
+    if dry_run ~= nil then
+        commit = not dry_run
+    end
+
+    local delay, remaining, ttl = util.redis_incoming(self, red, key, commit, cost)
```
AlinsRan:
```suggestion
    local delay, remaining, ttl = util.redis_incoming(self, red, key, not commit, cost)
```

apisix/plugins/limit-count/util.lua
```diff
@@ -0,0 +1,79 @@
...
+function _M.redis_incoming(self, red, key, commit, cost)
+    local limit = self.limit
+    local window = self.window
+    key = self.plugin_name .. tostring(key)
+
+    local requested_cost = cost or 1
+    local script_cost = commit and requested_cost or 0
+    local res, err = red:eval(commit_script, 1, key, limit, window, script_cost)
```
AlinsRan:
```suggestion
    local res, err = red:eval(commit_script, 1, key, limit, window, commit and cost or 0)
```

apisix/plugins/limit-count/util.lua
```diff
@@ -0,0 +1,79 @@
...
+        stored_remaining = limit - script_cost
+    end
+    local ttl = tonumber(res[2]) or window
+
+    local remaining
+    if commit then
+        remaining = stored_remaining
+    else
+        remaining = stored_remaining - requested_cost
+    end
```
AlinsRan:
```suggestion
    local remaining = stored_remaining - (commit and 0 or cost)
```

apisix/plugins/limit-count/util.lua
```diff
@@ -0,0 +1,79 @@
...
+
+    if remaining < 0 then
+        return nil, "rejected", ttl
+    end
+
+    return 0, remaining, ttl
+end
+
+function _M.redis_log_phase_incoming(self, red, key, cost)
```
AlinsRan, line 63: I think it's redundant. We can call redis_incoming.

t/plugin/ai-rate-limiting.t
```diff
@@ -984,3 +984,223 @@ passed
...
+[
+    "POST /ai\n" . "{ \"messages\": [ { \"role\": \"system\", \"content\": \"You are a mathematician\" }, { \"role\": \"user\", \"content\": \"What is 1+1?\"} ] }",
+    "POST /ai\n" . "{ \"messages\": [ { \"role\": \"system\", \"content\": \"You are a mathematician\" }, { \"role\": \"user\", \"content\": \"What is 1+1?\"} ] }",
+    "POST /ai\n" . "{ \"messages\": [ { \"role\": \"system\", \"content\": \"You are a mathematician\" }, { \"role\": \"user\", \"content\": \"What is 1+1?\"} ] }",
+    "POST /ai\n" . "{ \"messages\": [ { \"role\": \"system\", \"content\": \"You are a mathematician\" }, { \"role\": \"user\", \"content\": \"What is 1+1?\"} ] }",
+]
+--- more_headers
+Authorization: Bearer token
+--- error_code eval
+[200, 200, 200, 503]
```
AlinsRan: I think it's necessary to check the response header:
```
--- response_headers eval
[
    "X-AI-RateLimit-Remaining-ai-proxy-openai: 29",
    "X-AI-RateLimit-Remaining-ai-proxy-openai: 19",
    "X-AI-RateLimit-Remaining-ai-proxy-openai: 9",
    "X-AI-RateLimit-Remaining-ai-proxy-openai: 0",
]
```

t/plugin/ai-rate-limiting.t
```diff
@@ -984,3 +984,223 @@ passed
...
```
AlinsRan: ditto

t/plugin/limit-count-redis.t
```diff
@@ -51,6 +51,7 @@ __DATA__
 --- config
     location /t {
         content_by_lua_block {
+            require("lib.test_redis").flush_all()
```
AlinsRan: It can be globally initialized and reset without the need to call it in every test chunk, and the same applies to other test files.
```
add_block_preprocessor(sub {
    my ($block) = @_;
    my $extra_init_worker_by_lua = <<_EOC_;
        require("lib.test_redis").flush_all()
_EOC_

    $block->set_value("extra_init_worker_by_lua", $extra_init_worker_by_lua);
});
```

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -52,37 +44,66 @@ function _M.new(plugin_name, limit, window, conf)
...
+        return nil, err
     end
 
-    local remaining = res[1]
-    ttl = res[2]
+    return ok
+end
+
+
+function _M.incoming(self, key, cost, dry_run)
```
AlinsRan, line 100: We should be consistent with local and recommend using the `commit` parameter instead of `dry_run`.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -52,37 +44,66 @@ function _M.new(plugin_name, limit, window, conf)
...
+    if get_phase() == "log" then
```
AlinsRan: I think we should use a new function to handle the log phase instead of dealing with it within the incoming function. This is because we need to mock parameters like remaining, which isn't reasonable, and the log phase doesn't require a status code.
https://github.com/apache/apisix/blob/896d3c389196e7a06fddeb5c03006bf300fa4704/apisix/plugins/limit-count/init.lua#L301-L322

Baoyuantop:
Hi @beardnick, please check these comments.

apisix/plugins/limit-count/limit-count-redis.lua
AlinsRan: It seems that the modification of the `init.lua` file was overlooked, and the caller did not pass the `dry_run` parameter.
https://github.com/apache/apisix/blob/896d3c389196e7a06fddeb5c03006bf300fa4704/apisix/plugins/limit-count/init.lua#L282

Baoyuantop:
Hi @beardnick, following up on the previous review comments. Please let us know if you have any updates. Thank you.

apisix/plugins/limit-count/limit-count-redis.lua
beardnick: Ok, I'll simplify this

t/plugin/limit-count-redis.t
beardnick: I updated the unit tests, but this approach doesn't work for some of them because they rely on the state set by previous tests. For example
https://github.com/apache/apisix/blob/c0c906eb4d62886ffa4b0938b017dc4810cfe639/t/plugin/limit-count-redis-cluster.t#L124-L188
In these tests, test 5 expects `[200, 503, 503]` because test 4 consumed one quota, and the request limit is 2.

beardnick: In my opinion, the best approach is to update these tests so that they can run independently.

moonming (changes requested):
Hi @beardnick, thank you for the AI rate limiting Redis support! This has received 26 reviews, showing significant community interest.

**Since this has extensive review history**:
1. Could you rebase on the latest master?
2. Please confirm all 26 review comments have been addressed
3. Provide a summary of the key design decisions made during the review process

The feature itself is important — token-based rate limiting for AI workloads needs a distributed backend like Redis. Looking forward to getting this across the finish line!

Baoyuantop:
Hi @beardnick, for all the review comments above, please reply or mark them as resolved if you have already addressed them.

apisix/plugins/limit-count/limit-count-redis.lua
beardnick: Agreed, I split this logic into `log_phase_incoming`.

t/plugin/ai-rate-limiting.t
beardnick: I added a multi-worker test.

Baoyuantop:
Hi @beardnick, I re-run CI, but some tests are still failing. Please fix it.

beardnick:
> Hi @beardnick, I re-run CI, but some tests are still failing. Please fix it.

The failed tests don't seem related to my PR.

apisix/plugins/limit-count/util.lua
beardnick: I attempted to replace `redis_log_phase_incoming` with `redis_incoming`, but it ended up complicating the error handling during the log phase. So, instead, I decided to extract the shared logic between these two functions into a new function called `eval_limit`.

beardnick:
> Hi @beardnick, thank you for the AI rate limiting Redis support! This has received 26 reviews, showing significant community interest.
> 
> **Since this has extensive review history**:
> 
> 1. Could you rebase on the latest master?
> 2. Please confirm all 26 review comments have been addressed
> 3. Provide a summary of the key design decisions made during the review process
> 
> The feature itself is important — token-based rate limiting for AI workloads needs a distributed backend like Redis. Looking forward to getting this across the finish line!

1. I squashed my commits and rebased on master.
2. They are all addressed.
3. This PR updates the limit-count-redis plugin to support recording limits during the log phase. However, it will not add response headers as it does in the access phase.

Baoyuantop (dismissed):

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -52,37 +43,79 @@ function _M.new(plugin_name, limit, window, conf)
     return setmetatable(self, mt)
 end
 
-function _M.incoming(self, key, cost)
+
+local function release_connection(red, conf)
+    local ok, err = red:set_keepalive(conf.redis_keepalive_timeout, conf.redis_keepalive_pool)
+    if ok then
+        return true
+    end
```
nic-6443, line 51: Errors from this timer callback are silently swallowed. In OpenResty, return values from `ngx.timer.at` callbacks are discarded — nobody reads `return red, err` or `return nil, keepalive_err` or `return res, incoming_err`.

For `ai-rate-limiting`, the log phase is where actual LLM token usage gets deducted from the counter. If Redis is unreachable, the deduction silently fails with **zero visibility** — tokens are consumed but never counted, and operators have no indication.

The caller in `init.lua` only catches `ngx.timer.at()` creation failure (extremely rare), not the actual Redis operation failure inside the timer.

Same issue exists in `limit-count-redis-cluster.lua`'s `log_phase_incoming_thread`.

Should add `core.log.error(...)` for each error path inside the callback:
```lua
local red, err = redis.new(conf)
if not red then
    core.log.error("failed to create redis in log phase: ", err)
    return
end
...
if incoming_err then
    core.log.error("failed to deduct tokens in log phase: ", incoming_err)
end
```

beardnick:
I noticed that this has been implemented in #13670. It's a pity that I didn't manage to merge this PR. I'll find other opportunities to contribute to APISIX. Thanks for your review — I really learned a lot from it. I'll close this PR.

PR #12754, opened 2025-11-18
fix: ensure UI directory is created when building debian-dev image

sihyeonn:
### Description

This PR fixes a build failure when running `make build-on-debian-dev`.

The debian-dev image build failed at:

[2/2] STEP 9/21: COPY --chown=nobody:root ui/ /usr/local/apisix/ui/
Error: building at STEP "COPY --chown=nobody:root ui/ /usr/local/apisix/ui/": checking on sources under "/var/tmp/libpod_builder971763380/build": copier: stat: "/ui": no such file or directorybecause the `ui` directory does not exist in the build context.

To resolve this, the build script now ensures that an (empty) `ui` directory is created before building the debian-dev image, so the `COPY ui/ /usr/local/apisix/ui/` step always succeeds even when no UI assets are present.

Fixes #

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -427,7 +427,7 @@ release-src: compress-tar
 	gpg --batch --yes --armor --detach-sig $(project_release_name).tgz
 	shasum -a 512 $(project_release_name).tgz > $(project_release_name).tgz.sha512
 
-	$(call func_check_folder,release)
+	$(call func_check_folder, release)
 	mv $(project_release_name).tgz release/$(project_release_name).tgz
 	mv $(project_release_name).tgz.asc release/$(project_release_name).tgz.asc
 	mv $(project_release_name).tgz.sha512 release/$(project_release_name).tgz.sha512
@@ -496,6 +496,7 @@ ci-env-stop:
 .PHONY: build-on-debian-dev
 build-on-debian-dev:
 	@$(call func_echo_status, "$@ -> [ Start ]")
+	$(call func_check_folder, ui)
 	$(ENV_DOCKER) build -t $(ENV_APISIX_IMAGE_TAG_NAME)-debian-dev \
 		--build-arg TARGETARCH=$(ENV_OS_ARCH) \
 		--build-arg CODE_PATH=. \


Baoyuantop (approved):

Baoyuantop:
Hi @sihyeonn, you can merge the latest main branch to make CI pass.

sihyeonn:
> Hi @sihyeonn, you can merge the latest main branch to make CI pass.

Thank you for your review! But I still need 2 more reviews to merge 🥲 Merging is blocked!

Baoyuantop:
The failed test appears to be unrelated to this change and has already been fixed in the main branch.

sihyeonn:
Hi @SkyeYoung, thanks for the review!

Yes, this is for use cases where the UI is not needed.

I use `make build-on-debian-dev` locally when testing just the gateway functionality without the UI, and the build fails at the COPY step when the ui directory doesn't exist.

Moving the UI build from the workflow into the Makefile as you suggested is one approach, but it would increase build time and complexity for development environments where the UI isn't necessary.

The current fix just creates an empty directory to allow builds to succeed without UI assets - it's a minimal change. Do you think there's a better approach to handle this?

Baoyuantop:
Hi @sihyeonn, thank you for your contribution, but we are unsure if this requirement aligns with the needs of most users, so we need more discussion on this topic within the community.

sihyeonn:
Hi @Baoyuantop, I think the direction here is appropriate and consistent with the existing pattern. It’s just a bit disappointing to hear that it doesn’t fully align with the current approach, since applying the same func_check_folder behavior to ui seems like a natural extension of how it’s already used (e.g., for release).

Makefile
```diff
@@ -427,7 +427,7 @@ release-src: compress-tar
 	gpg --batch --yes --armor --detach-sig $(project_release_name).tgz
 	shasum -a 512 $(project_release_name).tgz > $(project_release_name).tgz.sha512
 
-	$(call func_check_folder,release)
+	$(call func_check_folder, release)
```
Baoyuantop, line 430: Please retain spaces.

Makefile
```diff
@@ -496,6 +496,7 @@ ci-env-stop:
 .PHONY: build-on-debian-dev
 build-on-debian-dev:
 	@$(call func_echo_status, "$@ -> [ Start ]")
+	$(call func_check_folder, ui)
```
Baoyuantop, line 499:
```suggestion
	$(call func_check_folder,ui)
```

moonming:
Hi @sihyeonn, thank you for looking into the build/Makefile issue!

After reviewing the discussion thread (7 comments), it seems there wasn't a clear consensus on the right approach for the `ui/` directory handling. Given that:

1. This is a build configuration change with differing opinions on the correct solution
2. The change is small (3 lines) but touches infrastructure that affects all build environments
3. The discussion hasn't converged on a direction

I think we should close this PR for now and revisit the `ui/` directory question as part of a broader build system review. If you'd like to continue this discussion, opening an issue might be a better venue to reach consensus first before implementing.

Thank you for the contribution, and I hope you'll continue contributing to APISIX! 🙏

PR #12756, opened 2025-11-18
feat(limit-count): add sliding window support for redis-based policies

sihyeonn:
### Description

This PR introduces **sliding window** and **approximate sliding window** support to the limit-count plugin for Redis- and Redis Cluster–based policies, while preserving the existing fixed window behavior by default.

#### Key Changes

1. **New `window_type` option**: Adds a configurable `window_type` parameter with three modes:
   - `"fixed"` (default): Traditional fixed window rate limiting
   - `"sliding"`: Exact sliding window using Redis sorted sets for precise per-request tracking
   - `"approximate_sliding"`: Memory-efficient approximate sliding window using counter-based algorithm

2. **Policy restriction**: Enforces that `window_type` other than `"fixed"` is only allowed when the policy is `redis` or `redis-cluster` to avoid unsupported combinations.

3. **Implementation**:
   - **Exact sliding window** (`"sliding"`): Uses Redis sorted sets (ZADD, ZCARD, ZREMRANGEBYSCORE) to track individual requests within the time window, providing 100% accuracy but higher memory usage (~100 bytes per request).
   - **Approximate sliding window** (`"approximate_sliding"`): Based on [Cloudflare's algorithm](https://blog.cloudflare.com/counting-things-a-lot-of-different-things/), uses two window counters with weighted calculation. Provides ~99.997% accuracy (tested with 400B+ requests) with significantly lower memory footprint (~16 bytes per key).

4. **Test coverage**: Adds comprehensive test cases for both sliding window modes in Redis and Redis Cluster configurations:
   - `t/plugin/limit-count-redis-sliding.t`
   - `t/plugin/limit-count-redis-cluster-sliding.t`
   - `t/plugin/limit-count-redis-approximate-sliding.t`
   - `t/plugin/limit-count-redis-cluster-approximate-sliding.t`

5. **Documentation**: Updates English and Chinese limit-count plugin documentation to describe the new `window_type` option, behavior differences, and performance/accuracy trade-offs.

#### Backward Compatibility

The default behavior remains unchanged for existing configurations. Configurations that do not specify `window_type` or explicitly use `"fixed"` will continue to work as before.

#### Performance Considerations

- **Exact sliding window**: Higher accuracy, higher memory usage, suitable for strict rate limiting requirements
- **Approximate sliding window**: Near-perfect accuracy with minimal memory overhead, suitable for high-traffic scenarios where slight approximation is acceptable

Fixes #

### Checklist


apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -40,12 +40,61 @@ local script = core.string.compress_script([=[
...
+            reset = earliest[2] + window - now
+            if reset < 0 then
+                reset = 0
+            end
+        end
+        return {-1, reset}
+    end
+
+    for i = 1, cost do
+        redis.call('ZADD', KEYS[1], now, now .. ':' .. i)
```
Baoyuantop: In high-concurrency scenarios, uniqueness may not be guaranteed. I think we can utilize ngx.var.request_id

Baoyuantop: Hi @sihyeonn, any updates?

sihyeonn: Sorry to late and also thanks for the review and the suggestion about using ngx.var.request_id for uniqueness under high concurrency.
I’ve updated both Redis and Redis Cluster sliding window implementations so that the Lua scripts now always use ngx.var.request_id as part of the ZSET member value (i.e. "$request_id:<i>").

Baoyuantop (dismissed):

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -39,7 +40,56 @@ local script = core.string.compress_script([=[
...
+                reset = 0
+            end
+        end
+        return {-1, reset}
+    end
+
+    for i = 1, cost do
+        local member = req_id .. ':' .. i
+        redis.call('ZADD', KEYS[1], now, member)
+    end
```
nic-6443, line 76: In the current implementation, each request needs to record an entry in the Redis set. This puts too much pressure on Redis and is probably not a good solution.

sihyeonn: @nic-6443 

Thanks for raising this concern!

In this sliding-window implementation, we intentionally trade some Redis work per request for *exact* “N requests per rolling window” semantics. However, the memory and per-key load are bounded:

- on every request we first call `ZREMRANGEBYSCORE` to drop all events outside of the current window and then enforce `current + cost <= limit` before inserting new members,
- rejected requests do **not** insert new entries,
- so for each key the ZSET cardinality is always `<= limit`, and Redis memory for this key is bounded by that limit.

Because of this, even though we do one `ZADD` per allowed request, we don’t end up with unbounded growth like in some TTL-only designs – the sliding window state per key stays O(limit). Also, `window_type = "sliding"` is an opt-in mode and the docs explicitly call out that it is more expensive than the default fixed window, so it’s intended for use cases that really need accurate rolling windows and can accept the extra Redis cost.

nic-6443: I suggest you use the approximate sliding window mentioned in this blog https://blog.cloudflare.com/counting-things-a-lot-of-different-things/ to implement this function. You can refer to the code here: https://github.com/ElvinEfendi/lua-resty-global-throttle/blob/main/lib/resty/global_throttle/sliding_window.lua

Baoyuantop: Hi @sihyeonn, following up on the previous review comments. Please let us know if you have any updates. Thank you.

sihyeonn: @spacewander @Baoyuantop Thank you so much for the excellent suggestion and providing both the Cloudflare blog reference and the implementation code! This was incredibly helpful.


I've implemented both approaches in this PR to give users flexibility based on their requirements:

**Sliding window goal**

Fixed window allows burst traffic at window boundaries. For example, with `limit=100, time_window=60s`, a user could make 100 requests at 11:00:59 and another 100 at 11:01:00 — effectively 200 requests in 2 seconds, defeating the rate limit purpose.
Sliding window prevents this by enforcing the limit over any rolling time period.

1. Exact sliding window (`window_type: "sliding"`)
- 100% accuracy using sorted sets
- ~100 bytes per request
- Best for: Payment APIs, security-sensitive endpoints where burst prevention is critical

2. Approximate sliding window (`window_type: "approximate_sliding"`)
- ~99.997% accuracy (0.003% error, validated with 400B+ requests per Cloudflare)
- ~16 bytes per key (counter-based: `prev_count × (1 - rate) + curr_count`)
- Best for: High-traffic APIs, CDN edges, cost-sensitive deployments

Test cases have been added for both modes. Please let me know if you'd like any adjustments!

moonming (changes requested):
Hi @sihyeonn, thank you for the sliding window rate limiting implementation!

At 1816 lines, this is a significant feature addition. Sliding window is more accurate than fixed window for rate limiting, and using Redis sorted sets is a well-known approach.

**Key concerns**:
1. **Redis overhead**: Each request requires multiple Redis commands (ZADD, ZREMRANGEBYSCORE, ZCARD, EXPIRE). Have you benchmarked the latency impact compared to the current fixed window approach?
2. **Memory usage**: Sorted sets store individual timestamps, which can consume significant Redis memory under high traffic. What's the memory footprint for, say, 10K requests/second with a 60-second window?
3. **Atomicity**: Are the Redis operations wrapped in a pipeline or Lua script to ensure atomicity?
4. **Migration path**: How do existing users migrate from fixed window to sliding window? Is it a simple config change?

Please provide performance benchmarks comparing the two approaches. This will help us make an informed decision about including it. Thank you for the thorough implementation!

sihyeonn:
Addressed all review feedback. Thanks for the reviews!

apisix/plugins/limit-count/limit-count-redis-cluster.lua
sihyeonn: Sorry for the late response! Pushed a fix addressing the remaining feedback — CROSSSLOT prevention with hash tags, TTL unit normalization (ms to seconds), request_id fallback, docs for approximate_sliding, and test improvements. Please take a look when you get a chance.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -74,14 +186,15 @@ function _M.incoming(self, key, cost)
     local remaining = res[1]
     ttl = res[2]
```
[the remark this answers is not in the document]
sihyeonn: Good catch, fixed — sliding/approximate_sliding now convert the reset value from ms to seconds before returning, so X-RateLimit-Reset is consistent across all modes.

apisix/plugins/limit-count/limit-count-redis-cluster.lua
```diff
@@ -39,7 +40,101 @@ local script = core.string.compress_script([=[
...
+    -- Calculate window IDs
+    local window_id = math.floor(now / window)
+    local prev_window_id = window_id - 1
+
+    -- Get counts from current and previous windows
+    local curr_key = KEYS[1] .. ':' .. window_id
+    local prev_key = KEYS[1] .. ':' .. prev_window_id
+
+    local curr_count = tonumber(redis.call('GET', curr_key) or 0)
+    local prev_count = tonumber(redis.call('GET', prev_key) or 0)
```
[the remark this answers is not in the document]
sihyeonn: Fixed — wrapping the base key with hash tags ({KEYS[1]}:window_id) so derived keys always land in the same slot.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -59,13 +156,28 @@ function _M.incoming(self, key, cost)
...
+        local window = self.window * 1000
+        local req_id = ngx_var.request_id
+
+        res, err = red:eval(script_sliding, 1, key, now, window, limit, c, req_id)
+    elseif self.window_type == "approximate_sliding" then
+        local now = ngx.now() * 1000
+        local window = self.window * 1000
+
+        res, err = red:eval(script_approximate_sliding, 1, key, now, window, limit, c)
+    else
```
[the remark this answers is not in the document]
sihyeonn: Added a fallback using worker id + timestamp + random number when request_id isn't available.

docs/en/latest/plugins/limit-count.md
```diff
@@ -46,6 +51,7 @@ You may see the following rate limiting headers in the response:
...
+| window_type             | string  | False                                     | fixed         | ["fixed","sliding"]          | The window behavior type. `fixed` uses a fixed window algorithm. `sliding` uses a sliding window algorithm to enforce an exact number of requests per rolling window. `sliding` is only supported when `policy` is `redis` or `redis-cluster`.                                                                                                                                                                                                                                               |
```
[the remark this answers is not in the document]
sihyeonn: Added — both English and Chinese docs now include approximate_sliding with a description.

apisix/plugins/limit-count/limit-count-redis.lua
```diff
@@ -40,12 +41,108 @@ local script = core.string.compress_script([=[
...
 
 
-function _M.new(plugin_name, limit, window, conf)
+local script_sliding = core.string.compress_script([=[
+    assert(tonumber(ARGV[3]) >= 1, "cost must be at least 1")
+
+    local now = tonumber(ARGV[1])
+    local window = tonumber(ARGV[2])
+    local limit = tonumber(ARGV[3])
+    local cost = tonumber(ARGV[4])
```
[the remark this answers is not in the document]
sihyeonn: In the fixed script, ARGV order is (limit, window, cost), so ARGV[3] is actually cost — the assert is correct as-is.

t/plugin/limit-count-redis-sliding.t
```diff
@@ -0,0 +1,501 @@
...
+
+
+=== TEST 10: sliding window with redis-cluster policy
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            -- Note: This test requires redis-cluster to be available
+            -- It validates schema but may fail at runtime if cluster unavailable
+            local code, body = t('/apisix/admin/routes/6',
```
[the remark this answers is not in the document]
sihyeonn: Moved Redis Cluster test cases behind a skip_all check that verifies cluster availability before running.

t/plugin/limit-count-redis-sliding.t
```diff
@@ -0,0 +1,501 @@
...
+                    r.req, r.status, r.limit or "nil", r.remaining or "nil", tostring(r.has_reset)))
+            end
+        }
+    }
+--- response_body_like
+req 1: status=404, limit=10, remaining=9, has_reset=true
+req 2: status=404, limit=10, remaining=8, has_reset=true
+req 3: status=404, limit=10, remaining=7, has_reset=true
+req 4: status=404, limit=10, remaining=6, has_reset=true
+req 5: status=404, limit=10, remaining=5, has_reset=true
```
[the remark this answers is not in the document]
sihyeonn: Changed to \d+ pattern so the test focuses on rate limit headers.

sihyeonn:
@Baoyuantop @nic-6443 Addressed the remaining feedback — CROSSSLOT fix with hash tags, TTL unit normalization, request_id fallback, docs for approximate_sliding, and test improvements. Would appreciate another review when you have time.

Baoyuantop:
Hi @sihyeonn, please resolve the code conflict.

sihyeonn:
Resolved the conflicts with the latest master.\n\nWhile resolving them, I noticed the upstream change that optimizes the fixed-window Redis Lua script path with EVALSHA and a NOSCRIPT fallback. This PR now preserves that behavior for the fixed-window path.\n\nI'll follow up with a separate PR to apply the same EVALSHA optimization to the new sliding and approximate_sliding Redis scripts as well.

sihyeonn:
Hi @moonming, thank you for the careful review.

Your concerns make sense, especially around Redis overhead and memory usage. Let me clarify the intended trade-offs.

For the exact `sliding` mode, yes, it is more expensive than the current fixed-window implementation. It uses a Redis sorted set and stores request events within the active window, so both Redis CPU and memory usage increase with traffic. I think this mode is useful when users need stricter rate-limit accuracy near window boundaries, but it is not the mode I would recommend for very high-throughput traffic by default.

That is why this PR also includes `approximate_sliding`. This implementation is the same approach we already use internally in production. It keeps counters for the current and previous windows instead of storing every request timestamp, so the memory profile is much closer to fixed window while still smoothing the boundary burst issue that fixed windows have.

On the specific points:

1. Redis overhead

The Redis operations are wrapped in Lua scripts, so each check/update is a single atomic Redis script execution from the APISIX side.

For `sliding`, the script does the sorted-set cleanup, count, insert, expire, and reset calculation.

For `approximate_sliding`, the script reads the current and previous counters, calculates the weighted count, increments the current counter, and updates expiration. This is the path we use for production traffic internally.

2. Memory usage

For exact `sliding`, memory usage is proportional to the number of request events kept in the active window:

```text
number_of_limited_keys * requests_per_key_within_window
```

So for a single hot key at `10K RPS` with a `60s` window, it can keep up to `600K` sorted-set entries. That is the main trade-off of exact sliding window.

For `approximate_sliding`, it only keeps counter keys for the current and previous windows, so the memory usage is much closer to the existing fixed-window mode.

3. Atomicity

Yes, the Redis operations are atomic because they are executed inside Lua scripts.

4. Migration path

Existing users are not affected. The default remains fixed window.

Users can opt in with a config-only change:

```json
"window_type": "sliding"
```

or:

```json
"window_type": "approximate_sliding"
```

I will also add benchmark results comparing `fixed`, `sliding`, and `approximate_sliding` for Redis and Redis Cluster before asking for another review.

PR #12773, opened 2025-11-24
fix: change consul event module to shared dict

### LINKED ISSUE #12398: bug: 503 errors occur after restarting APISIX when using Consul and APISIX both in Docker
### Current Behavior

Restarting APISIX while it is handling continuous requests leads to intermittent 503 Service Unavailable responses.

I have prepared a minimal reproducible setup. The file structure looks like this:

```
├── apisix
│   ├── apisix.yaml
│   └── config.yaml
├── nginx
│   └── default.conf
├── init.sh
├── restart.sh
└── test.sh

```

Once unpacked, the issue can be reproduced easily on any machine that has Docker and curl installed.

1. Run init.sh to launch Consul and the upstream NGINX container that APISIX will route traffic to.
2. Run restart.sh to start APISIX.
3. Run test.sh, which sends continuous requests to APISIX. At first, all responses should return HTTP status code 200.
4. While test.sh is running, repeatedly execute restart.sh to restart APISIX.
5. After restarting, even when the new APISIX instance is up and running normally, you will still see intermittent 503 errors in the output of test.sh.

This indicates that APISIX fails to route traffic correctly after a restart, despite the upstream and service discovery being healthy.

### Expected Behavior

After restarting APISIX, once it has successfully started and is healthy, it should be able to route traffic to upstream services registered in Consul without returning 503 errors.

### Error Logs

When test.sh starts to show 503 responses, you can check the APISIX container logs to see detailed error messages. These logs provide insight into why the requests failed, such as upstream discovery or routing issues.

### Steps to Reproduce

1. Unpack the minimal test setup files.
2. Run: ./init.sh
3. Start APISIX: ./restart.sh
4. Start testing: ./test.sh
5. In a separate terminal, repeatedly run: ./restart.sh
6. Watch test.sh output — it will gradually show 503 errors.

### Environment

[apisix-debug.zip](https://github.com/user-attachments/files/21035829/apisix-debug.zip)

Baoyuantop:
### Description

Configuring Consul service discovery in APISIX and restarting it while APISIX is continuously receiving traffic will result in frequent 503 errors.

In the design of Consul service discovery, only worker 0 directly pulls nodes from Consul and updates data. Other workers rely on `events:register` to receive broadcasts. If, when APISIX restarts, any worker has not yet completed `events:register`, but the service list data broadcast has already been sent, these workers will miss receiving data, and requests sent to that worker will return a 503 error.

This problem can be avoided by changing the event module to shared dict.

Fixes #12398

### Checklist

### DIFF
--- a/t/plugin/mcp-bridge.t
+++ b/t/plugin/mcp-bridge.t
@@ -71,8 +71,8 @@ property "args" validation failed: wrong type: expected array, got string
                         "plugins": {
                             "mcp-bridge": {
                                 "base_uri": "/mcp",
-                                "command": "npx",
-                                "args": ["-y", "@modelcontextprotocol/server-filesystem@2025.7.1", "/"]
+                                "command": "pnpm",
+                                "args": ["dlx", "@modelcontextprotocol/server-filesystem@2025.7.1", "/"]
                             }
                         },
                         "uri": "/mcp/*"


apisix/discovery/consul/init.lua
```diff
@@ -61,25 +65,62 @@ local watch_type_catalog = 1
...
+    end
+
+    local decoded, err = core.json.decode(data)
+    if not decoded then
+        log.error("failed to decode consul services from shared dict: ", err)
+        return
+    end
+
+    all_services = decoded
+end
```
nic-6443, line 107: Both functions writing to and reading from shared memory through `all_services` seem very dangerous. It may lead to re-writing the data just read from shared memory instead of getting the latest data from consul due to unexpected execution order. It is recommended to remove the use of the `all_services` variable, simplifying it so that only a privileged agent starts a timer to periodically fetch data from consul, while other workers only load data from shared memory.

jizhuozhi:
LGTM, I have merged this PR into my internal release, and it's already resolved my problem

PR #12779, opened 2025-11-26
fix(batch-requests): the number of sub-responses does not match that of sub-requests

hachi029:
### Description

when there is any timeout sub request, the number of sub responses does not match the number of sub requests。

for example: 

 the batch-request with 2 sub requests
`{"headers":{},"pipeline":[{"path":"/ok","method":"GET"},{"path":"/timeout","method":"GET"}]}`

there are 3  sub responses in the batch-response,  with an extra empty json object '{}'
`[{"status":200,"body":"ok","reason":"OK","headers":{"Connection":"keep-alive"}},{"status":504,"reason":"upstream timeout"},{}]`

### Checklist

### DIFF
--- a/t/plugin/batch-requests2.t
+++ b/t/plugin/batch-requests2.t
@@ -454,7 +454,7 @@ Content-Type: application/json
             local cjson = require("cjson.safe")
             local core = require("apisix.core")
             local t = require("lib.test_admin").test
-            local code, body, res_body, res_header = t('/apisix/batch-requests',
+            local _, _, batch_responses_str = t('/apisix/batch-requests',
                  ngx.HTTP_POST,
                  [=[{
                     "headers": {
@@ -469,10 +469,11 @@ Content-Type: application/json
                       "method": "GET"
                     }]
                 }]=])
-            ngx.status = code
-            -- print the number of sub-responses.
-            -- the number is expected to be the same as that of the sub-requests.
-            ngx.say(#cjson.decode(res_body))
+            local batch_responses = cjson.decode(batch_responses_str)
+            -- there are expected to be only 2 responses in batch_responses
+            for idx, response in ipairs(batch_responses) do
+                ngx.say(idx, "th response code: ", response.status)
+            end
         }
     }
 
@@ -492,4 +493,5 @@ GET /aggregate
 --- error_log
 timeout
 --- response_body
-2
+1th response code: 200
+2th response code: 504


t/plugin/batch-requests2.t
```diff
@@ -444,3 +444,52 @@ POST /apisix/batch-requests
...
+            ngx.sleep(1)
+            ngx.print("timeout")
+        }
+    }
+--- request
+GET /aggregate
+--- error_log
+timeout
+--- response_body
+2
```
Baoyuantop: It is recommended to check the specific content, such as the status.

t/plugin/batch-requests2.t
```diff
@@ -444,3 +444,52 @@ POST /apisix/batch-requests
...
+                    "pipeline":[
+                    {
+                        "path": "/ok",
+                        "method": "GET"
+                    },{
+                      "path": "/timeout",
+                      "method": "GET"
+                    }]
+                }]=])
+            -- core.log.warn("res_body:", res_body)
```
Baoyuantop: Do we need this line?

hachi029: Ok, I will delete it.

t/plugin/batch-requests2.t
hachi029: the original response contains variables such as  Date header like : `[{"body":"ok","headers":{"Server":"openresty/1.27.1.2","Content-Type":"text/plain","Date":"Thu,` 27 Nov 2025 06:46:49 GMT","Connection":"keep-alive","Transfer-Encoding":"chunked"},"status":200,"reason":"OK"},{"status":504,"reason":"upstream timeout"}]', so it is not easy to compare the original content.

Baoyuantop:
Hi @hachi029, could you merge the latest main branch to make CI pass?

hachi029:
@Baoyuantop merged

t/plugin/batch-requests2.t
membphis: for me, i think we can print two status which is much clear.

like this:
```
1th response code: 200
2th response code: 504

membphis (approved):
LGTM, only one minor issue I have leave comment

t/plugin/batch-requests2.t
Baoyuantop: Hi @hachi029, please check this comment, thanks.

hachi029: @membphis @Baoyuantop Fixed.

Baoyuantop:
@membphis  The comment has been resolved. Please review it again.

Issue #12704, opened 2025-10-27
bug: wrong plugin execution order


sina-devel:
### Current Behavior

The controller creates multiple global rules per plugin.
The resulting plugin execution order depends on the creation timestamp of each global rule, not the order defined in the APISIX Documentation.

The effective execution becomes:
```
ip-restriction -> real-ip
```

which causes `ip-restriction` to evaluate the wrong client IP.

### Expected Behavior

According to the [APISIX Documentation](https://apisix.apache.org/docs/apisix/terminology/plugin/#plugins-execution-order)
, the plugin order should follow:

Plugins in global rules
1. rewrite phase
2. access phase 

plugins bound to other objects
1. rewrite phase
2. access phase

Therefore, the expected execution order in this case is:
```
real-ip -> ip-restriction
```

since `real-ip` (rewrite phase) should run before `ip-restriction` (access phase).

### Steps to Reproduce

1. Create an initial `ApisixGlobalRule`:
```
apiVersion: apisix.apache.org/v2
kind: ApisixGlobalRule
metadata:
  name: global
  namespace: apisix
spec:
  plugins:
  - name: ip-restriction
    enable: true
    config:
      whitelist:
        - ...

```
2. Apply the manifest.
3. Later, update it to include another plugin:
```
apiVersion: apisix.apache.org/v2
kind: ApisixGlobalRule
metadata:
  name: global
  namespace: apisix
spec:
  plugins:
  - name: real-ip
    enable: true
    config:
      source: http_x_real_ip
  - name: ip-restriction
    enable: true
    config:
      whitelist:
        - ...

```
4. Apply the manifest again.

### Environment

- APISIX version (run `apisix version`): 3.13
- APISIX Ingress Controller version: 2.0.0-rc3

Baoyuantop:
> The resulting plugin execution order depends on the creation timestamp of each global rule, not the order defined in the APISIX Documentation.

Hi @sina-devel, how did you come to this conclusion? You can check the access log of apisix to see if the correct IP address is passed.

sina-devel:
> > The resulting plugin execution order depends on the creation timestamp of each global rule, not the order defined in the APISIX Documentation.
> 
> Hi [@sina-devel](https://github.com/sina-devel), how did you come to this conclusion? You can check the access log of apisix to see if the correct IP address is passed.

Hi @Baoyuantop, I verified this using the APISIX access logs.

[...]

PR #12800, opened 2025-12-08
fix: disallow creating duplicate plugins in global rules

shreemaan-abhishek:
### Description

When the same plugin is configured across different global rules, there will be no guarantee on which plugin config will be executed first. Additionally, the current architecture does not guarantee plugin execution order according to the priority.

We address this problem by disallowing configuring same plugin across multiple global_rules and sorting the plugins to their priority and executing in order.

Fixes: #12704

### Checklist

### DIFF
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -1276,7 +1276,7 @@ local function merge_global_rules(global_rules, conf_version)
         if global_rule.value and global_rule.value.plugins then
             for plugin_name, plugin_conf in pairs(global_rule.value.plugins) do
                 if seen_plugin[plugin_name] then
-                    core.log.warn("Found ", plugin_name,
+                    core.log.error("Found ", plugin_name,
                                   " configured across different global rules.",
                                   " Removing it from execution list")
                     plugins_hash[plugin_name] = nil


apisix/admin/global_rules.lua
```diff
@@ -31,6 +35,33 @@ local function check_conf(id, conf, need_id, schema)
         return nil, {error_msg = err}
     end
 
+    -- Check for plugin conflicts with existing global rules
+    if conf.plugins then
+        local global_rules = global_rules_mod.global_rules()
+        core.log.info("dibag global_rules: ", core.json.encode(global_rules))
```
nic-6443: temporary debug log should be removed

apisix/plugin.lua
```diff
@@ -1283,6 +1297,22 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
...
+                if duplicate_plugins[plugin_name] then
+                    -- Remove duplicate plugin and its config
+                    core.table.remove(plugins, i)
+                    core.table.remove(plugins, i)
+                else
+                    -- Move to next plugin pair
+                    i = i + 2
+                end
+            end
+
```
nic-6443: We not only need to remove duplicate plugins that appear in the global rule, but also merge the remaining plugins into a `plugins` table before executing them. Otherwise, the execution order of the plugins will depend on the order of the `values` array rather than plugin priority, which is unintended.。
```
for _, global_rule in config_util.iterate_values(values) do
     ...
end
```

apisix/plugin.lua
```diff
@@ -1288,37 +1288,52 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
...
+            value = {
+                updated_time = ngx.time(),
+                plugins = all_plugins,
+                created_time = ngx.time(),
+                id = 1,
+            },
+            createdIndex = createdIndex,
+            modifiedIndex = modifiedIndex,
+            clean_handlers = {},
+        }
```
nic-6443: I think we need to cache the result of this merge. We can use `global_rules.conf_version` as the cache key and the final `modifiedIndex` to avoid executing the merge code repeatedly for each request.
https://github.com/apache/apisix/blob/896d3c389196e7a06fddeb5c03006bf300fa4704/apisix/global_rules.lua#L45

apisix/plugin.lua
```diff
@@ -1273,22 +1329,26 @@ function _M.run_global_rules(api_ctx, global_rules, phase_name)
             api_ctx.global_rules = global_rules
         end
 
+        local dummy_global_rule = merge_global_rule_lrucache(conf_version, global_rules, merge_global_rules, global_rules)
+        if not dummy_global_rule then
+            core.log.error("failed to get merged global rules form cache, using merge_global_rules instead")
+            dummy_global_rule = merge_global_rules(global_rules)
```
nic-6443: This is unnecessary, lruchche will automatically execute `merge_global_rules` when the request key does not exist.

apisix/plugin.lua
```diff
@@ -1263,7 +1267,59 @@ function _M.set_plugins_meta_parent(plugins, parent)
...
+        end
+    end
+
+    -- remove duplicate plugins
+    for plugin_name, _ in pairs(duplicate_plugins) do
+        all_plugins[plugin_name] = nil
+    end
+
+    local dummy_global_rule = {
+        key = "/apisix/global_rules/1",
```
nic-6443:
```suggestion
        key = "/apisix/global_rules/dummy",
```

nic-6443: give it a special name for easy  distinguish

nic-6443 (dismissed):

membphis (dismissed):

apisix/admin/global_rules.lua
```diff
@@ -31,6 +35,32 @@ local function check_conf(id, conf, need_id, schema)
...
+                                    "' already exists in global rule with id '" ..
+                                    existing_rule.value.id .. "'"
+                                }
+                            end
+                        end
+                    end
+                end
+            end
+        end
+    end
```
bzp2010, line 69: I suspect this might be a problem. If an operation has already been executed but the etcd watch hasn't yet received the update event to refresh the in-memory cache, this validation could be unexpectedly bypassed.
Shouldn't we call the API here to fetch the latest list of global rules?

shreemaan-abhishek: good idea, let me fix this 👍🏼

apisix/plugin.lua
```diff
@@ -1263,7 +1267,59 @@ function _M.set_plugins_meta_parent(plugins, parent)
...
+            if global_rule.modifiedIndex > modifiedIndex then
+                modifiedIndex = global_rule.modifiedIndex
+                createdIndex = global_rule.createdIndex
+            end
+        end
+    end
+
+    -- remove duplicate plugins
+    for plugin_name, _ in pairs(duplicate_plugins) do
+        all_plugins[plugin_name] = nil
```
AlinsRan: We should warn which plugins are not working.

shreemaan-abhishek: you are right, thanks. I pushed a new commit that contains:
```lua
        core.log.warn("found ", plugin_name, " configured across different global rules ",
                      " it won't get executed")
```

AlinsRan: I think `duplicate_plugins` are redundant.
Can we merge three loops into one?

shreemaan-abhishek: done

nic-6443 (approved):

membphis:
lots of CI failed

apisix/plugin.lua
```diff
@@ -1263,7 +1267,45 @@ function _M.set_plugins_meta_parent(plugins, parent)
...
+local function merge_global_rules(global_rules, conf_version)
+    -- First pass: identify duplicate plugins across all global rules
+    local plugins_hash = {}
+    local seen_plugin = {}
+    local values = global_rules
+    for _, global_rule in config_util.iterate_values(values) do
+        if global_rule.value and global_rule.value.plugins then
+            for plugin_name, plugin_conf in pairs(global_rule.value.plugins) do
+                if seen_plugin[plugin_name] then
+                    core.log.warn("Found ", plugin_name,
```
membphis: we should use `error` log level

membphis: it is an important error message

membphis: if we used one plugin about 3 or 5 times

we should throw an error in this case

current way seems wrong

Baoyuantop (approved):

rgalex2034:
Maybe it's too late to consider, but I've been testing around this change, and it does not consider that multiple global rules may be defined with different _meta.filter configuration that do not overlap.

On the place I work for, we use this behaviour to define multiple traversal limit-count configurations where this limit changes based on the request.

It make sense to consider this situation to allow multiple definitions that ensure only one is executed?

It would require to traverse and discard the plugins by checking it's filtering first. If it's doable and make sense to allow this, I'm willing to try an implementation.

If it sounds interesting, I can open a new issue.

Thank you all :)

Baoyuantop:
Hi @rgalex2034, welcome to create an issue with your use case and configuration examples, and we will continue the discussion in your issue.

PR #12805, opened 2025-12-10
fix: repairing failed CI

Baoyuantop:

1. Due to the issue mentioned in https://github.com/apache/apisix/issues/12806, we have temporarily skipped the mcp-bridge test in this PR.
2. The httpbin.org service became unstable, preventing tests from running. All httpbin requests in the tests were replaced with a locally deployed httpbin service.

### Checklist

### DIFF
--- a/docker/compose/docker-compose-master.yaml
+++ b/docker/compose/docker-compose-master.yaml
@@ -48,10 +48,10 @@ services:
       - apisix
 
   httpbin:
-    image: kennethreitz/httpbin
+    image: ghcr.io/mccutchen/go-httpbin
     restart: always
     ports:
-      - "8280:80/tcp"
+      - "8280:8080/tcp"
     networks:
       - apisix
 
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

docker/compose/docker-compose-master.yaml
```diff
@@ -47,6 +47,14 @@ services:
     networks:
       - apisix
 
+  httpbin:
+    image: ghcr.io/mccutchen/go-httpbin
```
AlinsRan, line 51: Why does `kennethreitz/httpbin` need to use different image versions?

moonming (approved):

docker/compose/docker-compose-master.yaml
Baoyuantop: This configuration serves the workflow at https://github.com/apache/apisix/blob/master/.github/workflows/push-dev-image-on-commit.yml

This workflow runs on both amd and arm platforms. `kennethreitz/httpbin` does not provide an arm image, causing it to fail. Therefore, I used this image here.

PR #12807, opened 2025-12-12
feat(docker): improve debian-dev Dockerfile build caching

sihyeonn:
### Description

Optimize the debian-dev Dockerfile to leverage Docker layer caching more effectively.
Previously, any code change would invalidate the cache and trigger a full rebuild including `make deps` and brotli compilation.

#### What this PR does:

1. **Separate dependency files from source code**: Copy only `Makefile`, `rockspec`, `.requirements`, and `install-dependencies.sh` before running `make deps`, so dependencies are cached unless these files change.

2. **Move brotli build to build stage**: Previously, brotli was built in the runtime stage, requiring installation of ~80 build packages (gcc, cmake, etc.) every time. Now it's built once in the build stage and only the binary is copied to runtime.

#### Performance improvement:

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| Source code change only | 4m 37s | 18s | **~15x faster** |

Fixes #

### Checklist

### DIFF
--- a/docker/debian-dev/Dockerfile
+++ b/docker/debian-dev/Dockerfile
@@ -68,7 +68,11 @@ COPY --from=build /usr/bin/apisix /usr/bin/apisix
 COPY --from=build /usr/local/brotli /usr/local/brotli
 COPY --chown=nobody:root ui/ /usr/local/apisix/ui/
 
-RUN echo /usr/local/brotli/lib | tee /etc/ld.so.conf.d/brotli.conf \
+RUN if [ -d "/usr/local/brotli/lib64" ]; then \
+        echo /usr/local/brotli/lib64 | tee /etc/ld.so.conf.d/brotli.conf; \
+    else \
+        echo /usr/local/brotli/lib | tee /etc/ld.so.conf.d/brotli.conf; \
+    fi \
     && ldconfig \
     && ln -sf /usr/local/brotli/bin/brotli /usr/bin/brotli
 


docker/debian-dev/Dockerfile
```diff
@@ -54,11 +65,12 @@ RUN apt-get -y update --fix-missing \
 COPY --from=build /usr/local/apisix /usr/local/apisix
 COPY --from=build /usr/local/openresty /usr/local/openresty
 COPY --from=build /usr/bin/apisix /usr/bin/apisix
+COPY --from=build /usr/local/brotli /usr/local/brotli
 COPY --chown=nobody:root ui/ /usr/local/apisix/ui/
 
-COPY ${INSTALL_BROTLI} /install-brotli.sh
-RUN chmod +x /install-brotli.sh \
-    && cd / && ./install-brotli.sh && rm -rf /install-brotli.sh
+RUN echo /usr/local/brotli/lib | tee /etc/ld.so.conf.d/brotli.conf \
```
Baoyuantop: Please note the handling of the lib64 directory in the original install-brotli.sh script. https://github.com/apache/apisix/blob/master/docker/debian-dev/install-brotli.sh#L28-L32

Baoyuantop (commented):
Is there a significant change in the size of the image?

docker/debian-dev/Dockerfile
sihyeonn: Thanks for your review! I handled it.

sihyeonn:
@Baoyuantop 
No significant change in the final image size. 

The main benefit of this PR is improved build caching - brotli is now built once in the build stage and reused across rebuilds, rather than being rebuilt every time in the runtime stage. The image size remains similar because we're copying the same brotli binaries, just from a different stage.

Baoyuantop:
Hi @sihyeonn, please help to resolve the code conflicts.

PR #12824, opened 2025-12-19
fix(docker): Modify permissions for the docker file to run in Openshift without anyuid

### LINKED ISSUE #11714: feat: Consider fixing APISIX and APISIX-ingress containers to work without anyuid SCC
### Description

On OpenShift, it shouldn't be necessary for the user to add 'anyuid' permissions to SA that will be running APISIX.

The issues that are currently solved by using `anyuid` (https://github.com/apache/apisix/issues/10480, https://github.com/apache/apisix/issues/11425) are most likely due to `/usr/local/apisix` being owned by apisix:apisix (uid 636). This could be solved at build-time for APISIX images by following "Support arbitrary user ids" from [OpenShift docs](https://docs.openshift.com/container-platform/4.16/openshift_images/create-images.html).

Vishva066:
This pull request makes a minor change to the Docker build process to improve file permissions for the `/usr/local/apisix` directory. This helps ensure that the directory is accessible to users in group 0, which can be important for running containers with non-root users.

- Dockerfile permissions update:
  * Modified the `docker/debian-dev/Dockerfile` to recursively change the group ownership of `/usr/local/apisix` to group 0 and set group permissions to match user permissions.

This PR closes #11714

### DIFF
--- a/docker/debian-dev/Dockerfile
+++ b/docker/debian-dev/Dockerfile
@@ -58,7 +58,9 @@ COPY --chown=nobody:root ui/ /usr/local/apisix/ui/
 
 COPY ${INSTALL_BROTLI} /install-brotli.sh
 RUN chmod +x /install-brotli.sh \
-    && cd / && ./install-brotli.sh && rm -rf /install-brotli.sh
+    && cd / && ./install-brotli.sh && rm -rf /install-brotli.sh \
+    && chgrp -R 0 /usr/local/apisix \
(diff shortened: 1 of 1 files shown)


Baoyuantop (approved):
This file is responsible for building the apisix dev image. We also need to update all versions of the image at https://github.com/apache/apisix-docker/pull/612

membphis (approved):

nic-6443 (approved):

PR #9692, opened 2023-06-19
feat: increment route update for radixtree host uri, radixtree uri and radi…

ranxuxin001:
This PR extends apisix to modify a route without create the whole radixtree everytime.

refer to #9334

Issue #9140, opened 2023-03-22
help request:  send Put method admin request frequently costs apisix cpu high(about 100%) when more than 15 thousand routes exist in apisix


ranxuxin001:
### Description

Hello everyone,
    My apisix 2.9 has more than 15 thousand routes currently. And when I use the script to update large amount of routes frequently(sleep 0.5 to 5 seconds), my apisix costs cpu near 100%(user). It leads to request apisix slowly. 

   I've read some source code of apisix. When update a route, the apisix firstly search relevant node in radix tree. And secondly insert node in radix tree. But most of my routes has the same uri "/*" . The radix tree should has a few nodes, but it has an array with many routes which has the same uri. After that it updates the configuration of the matched route according to host, uri, header etc. So I doubt finding matched route in a long array causes much more cpu time than updating the node in the radix tree. By the way, it there any cpu performance tool suggusted? How about openresty xray?

   Is there any strategy to improve the apisx's performance for updating or creating routes frequently while large amount of routes already exist in the apisix, please? Any suggest is appreciated.

![apisixCpu](https://user-images.githubusercontent.com/33952968/226799280-f4b2e0fe-dab3-4c85-add4-18931d9a902e.png)

Best Regards,
Ranxuxin   

### Environment

- APISIX version (run `apisix version`):    apisix-2.9-0.el7.x86_64
- Operating system (run `uname -a`):   Linux  3.10.0-862.el7.x86_64 #1 SMP Fri Apr 20 16:44:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):    nginx version: openresty/1.19.9.1
built by gcc 9.3.1 20200408 (Red Hat 9.3.1-2) (GCC) 
built with OpenSSL 1.1.1k  25 Mar 2021
TLS SNI support enabled
[...]

PR #12826, opened 2025-12-19
feat: incremental compilation routers

apaloleg:
### Description
This commit introduces incremental route updates in the radixtree_host_uri, radixtree_uri, radixtree_uri_with_parameter routers implementation, eliminating the need to rebuild the entire radix tree on every configuration change via the Admin API.
Refer to #9692.

This feature depends on changes in lua-resty-radixtree lib: https://github.com/api7/lua-resty-radixtree/pull/156

Fixes (https://github.com/apache/apisix/issues/9140)

### Checklist

### DIFF
--- a/apisix/http/router/radixtree_host_uri.lua
+++ b/apisix/http/router/radixtree_host_uri.lua
@@ -50,8 +50,8 @@ local function get_host_radixtree_route(host_rev, sub_router)
 end
 
 
-local function push_host_router(route, host_routes, only_uri_routes, mode)
-    if route == nil or (type(route) ~= "table" or  route.value == nil) then
+local function push_host_router(route, host_routes, only_uri_routes)
+    if route == nil or (type(route) ~= "table" or route.value == nil) then
         return
     end
 
@@ -260,8 +260,7 @@ end
 
 local function incremental_operate_radixtree(routes)
     if ar.need_create_radixtree then
-        core.log.notice("create object of radixtree host uri after load_full_data or init. ",
-                        #routes)
+        core.log.notice("create object of radixtree host uri after load_full_data or init.")
         create_radixtree_router(routes)
         ar.need_create_radixtree = false
         core.table.clear(ar.sync_tb)
--- a/apisix/http/router/radixtree_host_uri.lua
+++ b/apisix/http/router/radixtree_host_uri.lua
@@ -25,6 +25,7 @@ local type = type
 local tab_insert = table.insert
 local loadstring = loadstring
 local pairs = pairs
+local ar = require("apisix.router")
 local cached_router_version
 local cached_service_version
 local host_router
@@ -34,8 +35,23 @@ local only_uri_router
 local _M = {version = 0.1}
 
 
-local function push_host_router(route, host_routes, only_uri_routes)
-    if type(route) ~= "table" then
+local function get_host_radixtree_route(host_rev, sub_router)
+    return {
+        id = host_rev,
+        paths = host_rev,
+        sub_router = sub_router,
+        filter_fun = function(vars, opts, ...)
+            return sub_router:dispatch(vars.uri, opts, ...)
+        end,
+        handler = function (api_ctx, match_opts)
+            api_ctx.real_curr_req_matched_host = match_opts.matched._path
+        end
+    }
+end
+
+
+local function push_host_router(route, host_routes, only_uri_routes, mode)
+    if route == nil or (type(route) ~= "table" or  route.value == nil) then
         return
     end
 
@@ -70,6 +86,7 @@ local function push_host_router(route, host_routes, only_uri_routes)
     end
 
     local radixtree_route = {
+        id = route.value.id,
         paths = route.value.uris or route.value.uri,
         methods = route.value.methods,
         priority = route.value.priority,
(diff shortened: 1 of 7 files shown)


apaloleg:
Hi!

Thank you @ranxuxin001 for the previous attempt to implement incremental updates for APISIX! The core idea was really solid and greatly inspired this solution here.

Unfortunately, when I tried to build on that solution for `radixtree_host_uri`, I couldn't quite get it working properly in my tests — it's very possible I didn't fully understand the original intent. So, I decided to take a slightly different path by directly passing and retrieving sub-routers in the host routes.

This allowed me to achieve stable incremental updates, but I'm well aware there might be a more elegant or cleaner way to do it. If you or anyone else has alternative ideas or suggestions, I'd genuinely love to hear them and am happy to adjust the implementation!

Additionally, in this PR was also: 

1. Added proper handling of the route `status` field (0/1) during create/update/delete operations. 
2. Removed storing changes in `sync_tb` in the privileged agent to prevent unnecessary data accumulation inside process.
3. Some changes according prev review.

apisix/http/router/radixtree_host_uri.lua
```diff
@@ -34,8 +35,23 @@ local only_uri_router
...
+            return sub_router:dispatch(vars.uri, opts, ...)
+        end,
+        handler = function (api_ctx, match_opts)
+            api_ctx.real_curr_req_matched_host = match_opts.matched._path
+        end
+    }
+end
+
+
+local function push_host_router(route, host_routes, only_uri_routes, mode)
```
Baoyuantop: What is the meaning of the `mode` parameter? It's not used in the function.

apisix/router.lua
```diff
@@ -17,33 +17,130 @@
...
+        sync_tb[key] = {op = "delete", last_route = route}
+    elseif sync_tb[key]["op"] == "create" then
+        sync_tb[key] = nil
+    elseif sync_tb[key]["op"] == "update" then
+        sync_tb[key] = {op = "delete", last_route = sync_tb[key]["last_route"]}
+    end
+end
+
+
+local function sync_tb_update(sync_tb,  pre_route, route)
```
Baoyuantop:
```suggestion
local function sync_tb_update(sync_tb, pre_route, route)
```

apisix/http/router/radixtree_host_uri.lua
```diff
@@ -34,8 +35,23 @@ local only_uri_router
...
+    if route == nil or (type(route) ~= "table" or  route.value == nil) then
```
Baoyuantop:
```suggestion
    if route == nil or (type(route) ~= "table" or route.value == nil) then
```

apisix/http/router/radixtree_host_uri.lua
apaloleg: Fixed

apisix/router.lua
apaloleg: Done

apisix/http/router/radixtree_host_uri.lua
apaloleg: Done

Baoyuantop:
Hi @apaloleg, thanks for your contribution. I would like to ask if you have encountered the problem described in this PR while actually using APISIX? When evaluating the implementation solutions for this PR, we found that introducing incremental updates might alleviate the problem to some extent, but incremental updates are unreliable, and their reliability is lower than that of full updates after long-term operation. Therefore, I would like to hear about the problems you actually encountered to determine a reasonable solution.

apaloleg:
> Hi @apaloleg, thanks for your contribution. I would like to ask if you have encountered the problem described in this PR while actually using APISIX? When evaluating the implementation solutions for this PR, we found that introducing incremental updates might alleviate the problem to some extent, but incremental updates are unreliable, and their reliability is lower than that of full updates after long-term operation. Therefore, I would like to hear about the problems you actually encountered to determine a reasonable solution.

Hi!

In our environment, we currently have over 40,000 routes (and this number is steadily growing over time), and we're using the `radixtree_host_uri` router.

We have frequently routes updates (add, update, delete) — these can be small operations (1–5 routes) or large ones (50–100+ routes). A series of such operations can last from 3-5s to 10+ minutes. At the same time, we have constant traffic on the gateway: around 200–300 RPS on average, and up to 900–1000 RPS during peak hours.

During these update periods, the CPU usage of APISIX pods spikes to 100%, and the latency of incoming requests increased significantly because requests had to wait for the full rebuild of the radix tree. This was a critical issue for us in production.

The patch from this PR completely resolved our problem. We've been running it in our production environment for about a month now, and we no longer experience these CPU spikes or increased latency during this updates period.

Baoyuantop:
Hi @apaloleg, may I ask if this solution is still working well?

apaloleg:
> Hi @apaloleg, may I ask if this solution is still working well?

Hi! Yes, we're still using this solution since mid-November - no problems whatsoever so far.

The high CPU usage we were seeing during admin operations before this patch hasn't come back at all.

We also haven't noticed any issues with create/update/delete operations or when toggling route status from 0 to 1 (and vice versa).

Baoyuantop:
cc @membphis

moonming (changes requested):
Hi @apaloleg, thank you for the incremental router compilation work!

This is an ambitious and important optimization — rebuilding the entire radix tree on every route change is a known performance bottleneck, especially at scale.

**Context**: There's also #13073 which takes a simpler approach (throttling rebuilds with a minimum interval). The two approaches complement each other:
- Your incremental compilation: Fundamentally reduces rebuild cost
- Throttling (#13073): Quick win that reduces rebuild frequency

**Questions**:
1. Does this require changes to `lua-resty-radixtree` upstream? If so, what's the status of those changes?
2. What's the performance improvement in your benchmarks? (e.g., time to add/remove a route with 10K existing routes)
3. How does it handle batch route updates (e.g., syncing 100 routes from etcd at once)?

This is a strategic improvement we'd love to land. Let's discuss the path forward — we might merge the throttling approach first as a quick win while this more fundamental optimization matures.

Thank you for the significant engineering effort!

PR #12830, opened 2025-12-22
feat: allow grpc web in non prefix based routes

shreemaan-abhishek:
### Description

Previously we only allowed routes with prefix matched based URI to be used with grpc-web, this is a limitation are there are valid use cases where grpc-web can be used to directly use the path configured in the route. Example: the grpc-web client library structures URI exactly in the format specified in the route URI. This limitation has been removed in this PR.

### Checklist

### DIFF
--- a/t/plugin/grpc-web.t
+++ b/t/plugin/grpc-web.t
@@ -178,6 +178,7 @@ Access-Control-Allow-Origin: *
 Content-Type: application/grpc-web
 
 
+
 === TEST 9: set route (with cors plugin)
 --- config
     location /t {


nic-6443 (approved):

membphis (approved):

docs/en/latest/plugins/grpc-web.md
```diff
@@ -68,14 +68,6 @@ curl http://127.0.0.1:9180/apisix/admin/routes/1 -H "X-API-KEY: $admin_key" -X P
 }'
 ```
 
-:::info IMPORTANT
-
-While using the `grpc-web` Plugin, always use a prefix matching pattern (`/*`, `/grpc/example/*`) for matching Routes. This is because the gRPC Web client passes the package name, the service interface name, the method name and other information in the proto in the URI. For example, `/path/a6.RouteService/Insert`.
```
kayx23, line 73: Need to remove this note from zh doc as well

PR #12836, opened 2025-12-23
ci: fix flaky tests

shreemaan-abhishek:
### Description

The sls logger failure got fixed by just waiting for a while to let the workers receive update from etcd.

Whereas, no clear fix was found for lago tests failures as most failures were due to `client socket timed out`.
So we modify the `--- exec` directive of APISIX.pm to timeout one second before the exec request times out there by appropriately reporting the stdout and stderr of the exec request instead of an empty `''` message.

This will help us understand the underlying reason of lago test failure and fix it in future.

Here is the verification that the test will output stdout/stderr properly when client socket times out:

image

### Checklist

### DIFF
--- a/t/plugin/lago.spec.mts
+++ b/t/plugin/lago.spec.mts
@@ -19,7 +19,7 @@ import { existsSync } from 'node:fs';
 import { readFile, rm, writeFile } from 'node:fs/promises';
 import { promisify } from 'node:util';
 
-import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { afterAll, beforeAll, describe, expect, it, jest } from '@jest/globals';
 import axios from 'axios';
 import * as compose from 'docker-compose';
 import { gql, request } from 'graphql-request';
@@ -69,6 +69,7 @@ const launchLago = async () => {
     cwd: LAGO_PATH,
     log: true,
     env: {
+      ...process.env,
       LAGO_RSA_PRIVATE_KEY: Buffer.from(privateKey).toString('base64'),
       FRONT_PORT: `${LAGO_FRONT_PORT}`, // avoiding conflicts, tests do not require a front-end
       API_PORT: `${LAGO_API_PORT}`,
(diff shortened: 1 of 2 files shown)


t/APISIX.pm
```diff
@@ -380,9 +380,9 @@ _EOC_
             location /exec_request {
                 content_by_lua_block {
                     local shell = require("resty.shell")
-                    local ok, stdout, stderr, reason, status = shell.run([[ $exec_snippet ]], $stdin, @{[$timeout*1000]}, $max_size)
+                    local ok, stdout, stderr, reason, status = shell.run([[ $exec_snippet ]], $stdin, @{[($timeout-1)*1000]}, $max_size)
                     if not ok then
-                        ngx.log(ngx.WARN, "failed to execute the script with status: " .. status .. ", reason: " .. reason .. ", stderr: " .. stderr)
+                        ngx.log(ngx.WARN, "failed to execute the script with status: " .. (status or "nil   ") .. ", reason: " .. (reason or "nil") .. ", stderr: " .. (stderr or "nil"))
```
nic-6443: Please manually simulate a timeout error locally, such as adding sleep in a test case executed by pnpm. I want to confirm that after making this change, the output of external programs like pnpm is indeed obtained when the timeout is triggered. Then put the test results in the PR description.

shreemaan-abhishek: Got it, I think we can add such test in the test suite itself where in it will verify itself?

nic-6443: Good idea, we can verify pnpm's output by asserting the error log.

t/APISIX.pm
```diff
@@ -380,9 +380,9 @@ _EOC_
             location /exec_request {
                 content_by_lua_block {
                     local shell = require("resty.shell")
-                    local ok, stdout, stderr, reason, status = shell.run([[ $exec_snippet ]], $stdin, @{[$timeout*1000]}, $max_size)
+                    local ok, stdout, stderr, reason, status = shell.run([[ $exec_snippet ]], $stdin, @{[($timeout-1)*1000]}, $max_size)
```
nic-6443, line 384: add a comment to explain why we minus 1 from total timeout

t/APISIX.pm
shreemaan-abhishek: well it turns out that even if we assert the response body, due to client socket timing out the test would still fail. 

so I added the my manual test record in the PR comment.

t/APISIX.pm
shreemaan-abhishek: done

nic-6443 (dismissed):

Baoyuantop (approved):

moonming (approved):

PR #12837, opened 2025-12-23
fix: Adding request-id header in case of empty header value in request

shubhammxv:
### Description

1. Adding request-id header in case header value comes empty in original request

### Checklist

### DIFF
--- a/t/plugin/request-id.t
+++ b/t/plugin/request-id.t
@@ -506,6 +506,8 @@ X-Request-ID: 123
 --- response_body
 true
 
+
+
 === TEST 16: check for request id in response header when request id is empty in request
 --- config
     location /t {


Baoyuantop:
Hi @shubhammxv, could you explain in detail why this modification was necessary? What is your specific scenario?

shubhammxv:
Hi @Baoyuantop 

In all the scenarios, we return `x-request-id` in response headers when `include_in_response` is true except when it comes as empty.
When it is passed as empty, then no header comes back in response (empty value header gets omitted) and it gives perception as if plugin is not working.

Also, the plugin's main functionality is to track the requests, But empty value will not really help track it further.

apisix/plugins/request-id.lua
```diff
@@ -102,7 +102,7 @@ end
 function _M.rewrite(conf, ctx)
     local headers = ngx.req.get_headers()
     local uuid_val
-    if not headers[conf.header_name] then
+    if not headers[conf.header_name] or headers[conf.header_name] == "" then
```
Baoyuantop: We can switch to a more readable style:

```
local header_req_id = headers[conf.header_name]
if not header_req_id or header_req_id == "" then
   ...
end
```

Baoyuantop: Hi @shubhammxv, there's only one minor code style suggestion; the rest is LGTM.

Baoyuantop:
```suggestion
    local header_req_id = headers[conf.header_name]
    if not header_req_id or header_req_id == "" then
```

apisix/plugins/request-id.lua
```diff
@@ -120,7 +120,7 @@ function _M.header_filter(conf, ctx)
     end
 
     local headers = ngx.resp.get_headers()
-    if not headers[conf.header_name] then
+    if not headers[conf.header_name] or headers[conf.header_name] == "" then
```
Baoyuantop:
```suggestion
    local header_req_id = headers[conf.header_name]
    if not header_req_id or header_req_id == "" then
```

nic-6443 (dismissed):

t/plugin/request-id.t
```diff
@@ -505,3 +505,29 @@ X-Request-ID: 123
 --- wait: 5
 --- response_body
 true
+
```
Baoyuantop, line 508:
```suggestion



```

nic-6443 (approved):

Baoyuantop (approved):

PR #12841, opened 2025-12-25
fix: metrics reporting when LLM plugin is not enabled

qiqizjl:
## Summary
Fix incorrect metrics reporting when LLM plugin is not enabled.

## Problem
When the LLM plugin is not enabled, the nginx variables for LLM metrics were defaulting to '0', which caused the metrics to be incorrectly reported as active even when no LLM functionality was being used.

## Solution
Changed the default values from '0' to '' (empty string) for the following nginx variables:
- `llm_time_to_first_token`
- `llm_prompt_tokens`
- `llm_completion_tokens`

Also updated the condition check in `openai-base.lua` to properly detect when LLM functionality is not active.

## Changes
- `apisix/cli/ngx_tpl.lua`: Changed default values from '0' to ''
- `apisix/plugins/ai-drivers/openai-base.lua`: Updated condition check

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -816,10 +816,10 @@ http {
             set $request_type               'traditional_http';
             set $request_llm_model              '';
 
-            set $llm_time_to_first_token        '0';
+            set $llm_time_to_first_token        '';
             set $llm_model                      '';
-            set $llm_prompt_tokens              '0';
-            set $llm_completion_tokens          '0';
+            set $llm_prompt_tokens              '';
+            set $llm_completion_tokens          '';
 
 
             access_by_lua_block {
--- a/apisix/plugins/ai-drivers/openai-base.lua
+++ b/apisix/plugins/ai-drivers/openai-base.lua
@@ -100,7 +100,7 @@ local function read_response(ctx, res)
                 return
             end
 
-            if ctx.var.llm_time_to_first_token == "0" then
+            if ctx.var.llm_time_to_first_token == "" then
                 ctx.var.llm_time_to_first_token = math.floor(
                                                 (ngx_now() - ctx.llm_request_start_time) * 1000)
             end


Baoyuantop:
Hi @qiqizjl, we made this change in https://github.com/apache/apisix/pull/12549. Could you explain the original requirements of this PR? Changing the default value from '0' to '' may affect users' existing JSON log format.

qiqizjl:
> Hi @qiqizjl, we made this change in #12549. Could you explain the original requirements of this PR? Changing the default value from '0' to '' may affect users' existing JSON log format.

We found in [apisix/plugins/prometheus/exporter.lua#L376-L394](https://github.com/apache/apisix/blob/master/apisix/plugins/prometheus/exporter.lua#L376-L394) that metrics are reported when the value is not empty. When the LLM plugin is not enabled, the default value is 0, which results in invalid metrics.

qiqizjl:
Another possible solution is to check if `request_type` is `ai_stream` or `ai_chat`, and report accordingly. I tried changing the check to report if `ctx.var.llm_time_to_first_token != 0`, but this might affect abnormal requests. If possible, I can continue to refine this issue based on the discussion.

flearc:
> Another possible solution is to check if `request_type` is `ai_stream` or `ai_chat`, and report accordingly. I tried changing the check to report if `ctx.var.llm_time_to_first_token != 0`, but this might affect abnormal requests. If possible, I can continue to refine this issue based on the discussion.


```
    if llm_time_to_first_token ~= "0" then
        metrics.llm_latency:observe(tonumber(llm_time_to_first_token),
            gen_arr(route_id, service_id, consumer_name, balancer_ip,
            vars.request_type, vars.request_llm_model, vars.llm_model,
            unpack(extra_labels("llm_latency", ctx))))
    end
    if vars.llm_prompt_tokens ~= "0" then
        metrics.llm_prompt_tokens:inc(tonumber(vars.llm_prompt_tokens),
            gen_arr(route_id, service_id, consumer_name, balancer_ip,
            vars.request_type, vars.request_llm_model, vars.llm_model,
            unpack(extra_labels("llm_prompt_tokens", ctx))))
    end
    if vars.llm_completion_tokens ~= "0" then
        metrics.llm_completion_tokens:inc(tonumber(vars.llm_completion_tokens),
            gen_arr(route_id, service_id, consumer_name, balancer_ip,
            vars.request_type, vars.request_llm_model, vars.llm_model,
            unpack(extra_labels("llm_completion_tokens", ctx))))
    end
```
Adjust the if statement might help

Baoyuantop:
Hi @qiqizjl, the proposed changes in this PR are incomplete and will cause a regression. We recommend creating an issue first to discuss a solution for this problem.

Baoyuantop:
The current PR implementation is incorrect, so I'm closing it for now. You can resubmit the PR according to the discussion in the issue.

PR #12843, opened 2025-12-25
fix(proxy-rewrite): query string discarded when use_real_request_uri_…

### LINKED ISSUE #12623: bug: proxy-rewrite discard the query string when use_real_request_uri_unsafe is true
### Current Behavior

Using the proxy-rewrite plugin to modify the request uri, if `use_real_request_uri_unsafe` is set to `true`, the queryS tring of original request will be discarded. And returned to normal when `use_real_request_uri_unsafe`  is set to `false`(default), 

### Expected Behavior

Whether or not `use_real_request_uri_unsafe` is enabled, the queryString should not be discarded.

### Error Logs

_No response_

### Steps to Reproduce

1. Add route with proxy-rewrite plugin:
```
curl -XPUT -H "x-api-key:${ADMIN_KEY}" http://localhost:9180/admin/apisix/routes \
-d '{
    "id": "httpbin-get",
    "methods": ["GET"],
    "uri": "/fakeget",
    "plugins": {
      "proxy-rewrite": {
        "uri": "/get",
        "use_real_request_uri_unsafe": true
      }
    },
    "upstream": {
      "type": "roundrobin",
      "pass_host": "node",
      "nodes": {
        "httpbin.org:80": 1
      }
    }
  }'
```
2. Send httpbin GET request with query param `curl -XGET "localhost:9080/fakeget?foo=bar"`

Image

3. Update proxy-rewrite plugin setting:
```
curl -XPUT -H "x-api-key:${ADMIN_KEY}" http://localhost:9180/admin/apisix/routes \
-d '{
    "id": "httpbin-get",
    "methods": ["GET"],
    "uri": "/fakeget",
    "plugins": {
      "proxy-rewrite": {
        "uri": "/get",
      }
    },
    "upstream": {
      "type": "roundrobin",
      "pass_host": "node",
      "nodes": {
        "httpbin.org:80": 1
      }
    }
  }'
```
4. Send the same request `curl -XGET "localhost:9080/fakeget?foo=bar"`

Image

### Environment

- APISIX version (run `apisix version`): 3.13.0
- Operating system (run `uname -a`): Debian GNU/Linux 11

YapWC:
…unsafe is true

When use_real_request_uri_unsafe is set to true, url that was requested with query string, arguments is not parsed. Expected to parse the arguments just like same default behavior when use_real_request_uri_unsafe is set to false.

The real issue lies in the fact that when use_real_request_uri_unsafe is set to true and conf.uri is provided, the core.utils.resolve_var output back an upstream uri that does not include query string. Therefore we need to concatenate the query string for the case if conf.uri is true.

Fixes #12623 

### Checklist


Baoyuantop (commented):
If the plugin only configures "use_real_request_uri_unsafe": true, it will cause the query to be concatenated repeatedly.

YapWC:
Hi @Baoyuantop alright I think I understand what you mean. I think it is better for me to debug it line by line in this case. For your side normally how do you run debugging/printing code output?

apisix/plugins/proxy-rewrite.lua
```diff
@@ -275,15 +275,27 @@ function _M.rewrite(conf, ctx)
         end
     end
 
+    -- normalized uri (without query string)
     local upstream_uri = ctx.var.uri
     local separator_escaped = false
+
+    local query_string = ""
+    -- When use_real_request_uri_unsafe is true, we use real_request_uri directly
```
Baoyuantop: Please simplify the comments.

Baoyuantop:
Hi @YapWC, following up on the previous review comments. Please let us know if you have any updates. Thank you.

moonming (changes requested):
Hi @YapWC, thank you for fixing the query string issue in proxy-rewrite!

Query strings being discarded when using `use_real_request_uri` is a real bug that affects production use. 

**Please**:
1. Rebase on the latest master
2. Add edge case tests: empty query strings, query strings with special characters, multiple `?` in the URI
3. Verify the fix works correctly with both `use_real_request_uri_unsafe` variants

Looking forward to the update!

Baoyuantop:
Hi @YapWC, following up on the previous review comments. Please let us know if you have any updates. Thank you.

t/plugin/proxy-rewrite.t
```diff
@@ -1231,3 +1231,95 @@ GET /hello
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+passed
+
+
+
+=== TEST 46: rewrite uri args
```
nic-6443, line 1274: The earlier revision of this PR double-appended the query when only `use_real_request_uri_unsafe: true` was set (no `uri`) — that is fixed now since `query_string` is only appended in the `conf.uri` branch, but none of the tests actually pin that scenario down. TEST 8/9 in proxy-rewrite3.t cover unsafe mode without a query string, and the new tests here all set `uri` together with the flag. Since this exact case already regressed once during review, it would be good to add a test with just `use_real_request_uri_unsafe: true` (no `uri`/`regex_uri`) and a request like `/hello?q=apisix`, asserting the query reaches the upstream exactly once.

shreemaan-abhishek (approved):

nic-6443 (approved):

membphis (approved):

moonming (approved):

PR #12856, opened 2026-01-01
fix: maintain node_version for independent upstream

shreemaan-abhishek:
### Description

The new decoupled healthcheck_manager logic has missing code to update the version of upstream conf in the case where routes are configured via upstream_id reference.

This version update is necessary for the `healtcheck_manager` as it relies on that to know that the nodes have changed in order to create correct health checkers for new nodes.

### Checklist

### DIFF
--- a/t/node/healthchecker-independent-upstream.t
+++ b/t/node/healthchecker-independent-upstream.t
@@ -106,7 +106,7 @@ GET /t
 --- response_body
 passed
 --- grep_error_log eval
-qr/create new checker: table: /
+qr/create new checker: table:/
 --- grep_error_log_out
 create new checker: table:
 create new checker: table:


membphis:
some ci failed, pls fix them first

apisix/utils/upstream.lua
```diff
@@ -117,11 +118,15 @@ function _M.parse_domain_in_up(up)
...
     end
 
-    if not up.orig_modifiedIndex then
-        up.orig_modifiedIndex = up.modifiedIndex
+    local nodes_ver = resource.get_nodes_ver(up.value.resource_key)
+    if not nodes_ver then
+        nodes_ver = 0
     end
-    up.modifiedIndex = up.orig_modifiedIndex .. "#" .. ngx_now()
+    nodes_ver = nodes_ver + 1
```
membphis, line 125: can we delete this line? then use the `node_ver` directly

shreemaan-abhishek: I don't understand what you mean.

membphis: my question:

can we delete this line `nodes_ver = nodes_ver + 1`? it seems useless

why do we have to `+1`, I do not understand this

shreemaan-abhishek: After dns resolution, the nodes have changed, that's why we need to increment the version. This is in fact the crux of this PR.

PR #12857, opened 2026-01-02
feat(ai-proxy): add path_mode option to override.endpoint 

sihyeonn:
### Description

Add `override.path_mode` option to `ai-proxy` and `ai-proxy-multi` plugins to control how the upstream request path is determined.

**Background (Problem)**
Currently, when `override.endpoint` is set, the URL path is fixed to either the endpoint's path or the driver's default path. This causes issues when:
- Using a single route for multiple API endpoints (e.g., `/v1/chat/completions` and `/v1/responses`)
- Wanting to preserve the original request URI path while only overriding the host

**Solution**
Add `path_mode` option with three modes:
- `fixed` (default): Existing behavior - uses endpoint path or driver default
- `preserve`: Uses the original request URI path (with query string)
- `append`: Appends original request URI path to the endpoint path

This allows using a single route with the same model for multiple API endpoints, reducing route explosion in multi-model deployments.

Fixes #

### Checklist


Baoyuantop:
Hi @sihyeonn, thank you for your contribution. Could you please describe your original requirements in detail? What problems did you encounter while using the current version, and why was this change necessary?

sihyeonn:
Hi @Baoyuantop ,

I'm running an LLM gateway. Many models expose multiple endpoints from the same base URL:
- `/v1/chat/completions`
- `/v1/responses`
- `/v1/embeddings`
- `/v1/moderations`

Currently, I must create separate routes for each endpoint because `override.endpoint` replaces the entire URI:
```yaml
# Example Current: 4 routes per model
routes:
  - uri: /v1/chat/completions
    plugins:
      ai-proxy:
        override:
          endpoint: https://api.example.com/models/gpt-4/v1/chat/completions

  - uri: /v1/responses
    plugins:
      ai-proxy:
        override:
          endpoint: https://api.example.com/models/gpt-4/v1/responses
  # ... 2 more routes
```

This creates unnecessary duplication:
- 4+ routes per model
- Same base URL repeated across routes
- Duplicate plugin configs (healthcheck, auth, timeout)

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -239,7 +239,36 @@ function _M.request(self, ctx, conf, request_table, extra_opts)
...
+            path = req_path
+        else
+            path = prefix .. req_path
+            path = path:gsub("//+", "/")
+        end
+        if ctx.var.is_args and ctx.var.args and #ctx.var.args > 0 then
+            local req_args_tab = core.string.decode_args(ctx.var.args)
+            if type(req_args_tab) == "table" then
+                core.table.merge(query_params, req_args_tab)
+            end
```
Baoyuantop: It can be extracted into a common function.

Baoyuantop: It is recommended to use `ctx.var.is_args == "?"` to check.

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -239,7 +239,36 @@ function _M.request(self, ctx, conf, request_table, extra_opts)
...
+        end
+    else
+        path = (endpoint_path and endpoint_path ~= "" and endpoint_path) or self.path
```
Baoyuantop: It needs to be confirmed whether the behavior remains consistent when endpoint_path is an empty string.

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -89,6 +89,15 @@ local ai_instance_schema = {
                         type = "string",
                         description = "To be specified to override the endpoint of the AI Instance",
                     },
+                    path_mode = {
```
Baoyuantop: It is recommended to extract it as a common schema variable to avoid duplication.

sihyeonn:
@Baoyuantop Thanks for your review. I handled it.

Baoyuantop:
Hi @sihyeonn, please help resolve the code conflicts.

apisix/plugins/ai-drivers/openai-base.lua
```diff
@@ -45,8 +45,26 @@ local HTTP_INTERNAL_SERVER_ERROR = ngx.HTTP_INTERNAL_SERVER_ERROR
...
+        local req_args_tab = core.string.decode_args(ctx.var.args)
+        if type(req_args_tab) == "table" then
+            core.table.merge(query_params, req_args_tab)
+        end
+    end
+end
+
+
+function _M.new(opts)
+    local self = {
```
Baoyuantop: Is this change necessary? It breaks the original transparent pass-through mechanism—now only explicitly listed fields are passed through.

Baoyuantop: Hi @sihyeonn, following up on the previous review comments. Please let us know if you have any updates. Thank you.

moonming (changes requested):
Hi @sihyeonn, thanks for the path_mode feature for ai-proxy!

Supporting `preserve_path` and `override_path` modes in the endpoint override is useful for scenarios where a single route needs to hit different API endpoints based on the original request path.

**Please address**:
1. Rebase on latest master (this has been waiting for a while)
2. Ensure the schema validation clearly documents the available modes and their behavior
3. Add test cases covering both modes, including edge cases like empty paths and paths with query strings

Thank you!

sihyeonn:
Addressed all review feedback. Thanks for the reviews!

apisix/plugins/ai-drivers/openai-base.lua
sihyeonn: Done — extracted the path logic into a build_path() helper.

sihyeonn: Already using ctx.var.is_args in the merge_request_query_params function, kept as is.

apisix/plugins/ai-drivers/openai-base.lua
sihyeonn: Handled — build_path falls back to the driver default path when endpoint_path is nil or empty string.

apisix/plugins/ai-proxy/schema.lua
sihyeonn: path_mode_schema is already defined once and reused across both ai_instance_schema and ai_proxy_schema.

apisix/plugins/ai-drivers/openai-base.lua
sihyeonn: You're right, that was unintentional. Reverted _M.new back to the original setmetatable(opt, mt) so all fields are passed through transparently. Sorry for the delay on this.

sihyeonn:
@Baoyuantop Sorry for the delay. Addressed all your feedback — restored transparent pass-through in _M.new, extracted build_path helper, and handled empty endpoint_path. Ready for another look.

Baoyuantop (commented):
1. Please supplement the documentation for the newly added configuration.
2. What will happen if a user sets `path_mode: preserve` but doesn't set `override.endpoint`?

PR #12864, opened 2026-01-06
feat: auth plugins respond with `www-authenticate` header with realm

shreemaan-abhishek:
### Description

This PR adds a realm configuration option to several authentication plugins (basic-auth, key-auth, hmac-auth, jwt-auth, ldap-auth). This allows users to customize the realm value returned in the WWW-Authenticate header when an authentication request fails with a 401 Unauthorized status.

This is in line with standard concept in the HTTP authentication protocol (RFC 7235). It indicates to the client which "protection area" the currently requested resource belongs to. This helps the client (such as a browser or API client) determine which set of credentials to use to re-initiate the request.

### Checklist


apisix/plugins/basic-auth.lua
```diff
@@ -157,15 +160,17 @@ end
 
 
 function _M.rewrite(conf, ctx)
-    local cur_consumer, consumer_conf, err = find_consumer(ctx)
+    local cur_consumer, consumer_conf, err = find_consumer(ctx, conf)
     if not cur_consumer then
         if not conf.anonymous_consumer then
+            core.response.set_header("WWW-Authenticate", "Basic realm='" .. (conf.realm or "basic") .. "'")
```
nic-6443: I think we can remove `or "basic"` in here because the default value is already specified in the schema.

apisix/plugins/basic-auth.lua
```diff
@@ -32,6 +32,10 @@ local schema = {
         hide_credentials = {
             type = "boolean",
             default = false,
+        },
+        realm = {
+            type = "string",
```
membphis: it can not contain any `'`, `"`

maybe we should limit it

apisix/plugins/ldap-auth.lua
```diff
@@ -28,7 +28,8 @@ local schema = {
         ldap_uri = { type = "string" },
         use_tls = { type = "boolean", default = false },
         tls_verify = { type = "boolean", default = false },
-        uid = { type = "string", default = "cn" }
+        uid = { type = "string", default = "cn" },
+        realm = { type = "string", default = "ldap" }
```
membphis: we can add a `,` at the end of this line always

nic-6443 (approved):

membphis (dismissed):

nic-6443 (approved):

PR #12872, opened 2026-01-08
fix(limit-conn): implement configurable redis key expiry

### LINKED ISSUE #12823: help request: plugin limit-conn with redis counter leaks in some condition
### Description

We use plugin limit-conn to limit the request concurrency of each client. We noticed that when a partial replicas of apisix failed while processing requests, the counter on redis will not be decreased properly, which will result in rejecting coming requests on the rest replicas.
We also noticed that the if the decrease of redis counter failed, it will not be retried, which will leads to counter leaks in an unstable network environment.

How can I solve the above problems?

### Environment

- APISIX version (run `apisix version`): 3.14.1
- Operating system (run `uname -a`): Debian GNU/Linux 11 (bullseye)
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.27.1.2

shreemaan-abhishek:
### Description

When using limit-conn, when an APISIX node crashes and is unable to perform the "decrement count" operation, this key remains permanently in Redis as there is not expiry configured, causing the user's concurrency quota to be permanently occupied or even the user to be blocked. 

UPDATE:

Thanks to @ChuanFF's comment: https://github.com/apache/apisix/pull/12872#issuecomment-3738318512, we realised that the existing method of counting connections would not yield accurate and correct connection limiting in distributed deployment scenario.

The correct solution to ensure keys expire timely would be when an expiry was tied to each individual increment (each connection). To achieve this we use redis sorted set to count each connection by a `request_id`, assign a ttl to each `request_id` and remove timely remove all `request_ids` that have exceeded the ttl.

### Checklist

Fixes #12823


apisix/plugins/limit-req/util.lua
```diff
@@ -68,6 +68,10 @@ function _M.incoming(self, red, key, commit)
         if not ok then
             return nil, err
         end
+
+        local ttl = math.ceil(self.burst / self.rate) + 1
+        red:expire(excess_key, ttl)
+        red:expire(last_key, ttl)
```
nic-6443: I think fiexed one second is enough for limit-req

nic-6443: and can we set ttl by `redis:set` api directly? otherwise we introduce another two redis calls per request now.

shreemaan-abhishek: done

apisix/plugins/limit-conn/util.lua
```diff
@@ -28,9 +28,20 @@ function _M.incoming(self, red, key, commit)
 
     local conn, err
     if commit then
-        conn, err = red:incrby(key, 1)
-        if not conn then
-            return nil, err
+        if self.conf.key_ttl then
+            red:init_pipeline()
```
moonming: The PR introduces two different methods for setting a TTL:

For maintainability and predictability, we should use a single, consistent approach across all plugins that share a common logic. The SET ... EX approach is generally preferred for its atomicity and efficiency. Why was a different approach chosen for limit-conn?

•limit-conn: Uses a Redis pipeline to issue separate INCRBY and EXPIRE commands.

•limit-req: Uses the atomic SET ... EX command.

apisix/plugins/limit-req/util.lua
```diff
@@ -57,14 +57,16 @@ function _M.incoming(self, red, key, commit)
     end
 
     if commit then
-        local ok
-        local err
-        ok, err = red:set(excess_key, excess)
+        -- limit-req does rate limiting per second, so we set the ttl to 2 seconds
+        local ttl = 2
```
moonming: The TTL for limit-req is hardcoded to 2 seconds. This feels arbitrary and lacks flexibility.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -44,6 +44,7 @@ The `limit-conn` Plugin limits the rate of requests by the number of concurrent
...
 | key_type        | string  | False      | var   | ["var","var_combination"] | The type of key. If the `key_type` is `var`, the `key` is interpreted a variable. If the `key_type` is `var_combination`, the `key` is interpreted as a combination of variables.    |
 | key       | string  | False      | remote_addr |   | The key to count requests by. If the `key_type` is `var`, the `key` is interpreted a variable. The variable does not need to be prefixed by a dollar sign (`$`). If the `key_type` is `var_combination`, the `key` is interpreted as a combination of variables. All variables should be prefixed by dollar signs (`$`). For example, to configure the `key` to use a combination of two request headers `custom-a` and `custom-b`, the `key` should be configured as `$http_custom_a $http_custom_b`. |
+| key_ttl   | integer | False      | 60          |   | The TTL of the Redis key in seconds. Used when `policy` is `redis` or `redis-cluster`. |
```
kayx23: 1. update `limit-req` doc for the parameter
2. add the corresponding zh docs

apisix/utils/redis-schema.lua
```diff
@@ -78,4 +78,14 @@ local _M = {
     schema = policy_to_additional_properties
 }
 
+
+function _M.ttl_policy_schema(kind, ttl)
+    local schema = policy_to_additional_properties[kind]
```
moonming: The ttl_policy_schema function modifies a shared table, is this safe?

t/plugin/limit-conn-redis-cluster-ttl.t
```diff
@@ -0,0 +1,334 @@
+#
```
moonming: Need to add a scenario where a key's TTL expires. The test should verify that the counter is effectively gone and that a new request can proceed without being blocked by a stale counter.

•Example: Set a short TTL (e.g., 2s), make a request, wait for 3s, and then confirm that a subsequent request does not inherit the old counter value.

apisix/plugins/limit-conn/util.lua
shreemaan-abhishek: Unlike `limit-req`, `limit-conn` uses `incrby`, and `incrby` does not support passing an `expiry` parameter.

Using redis pipeline to perform incrby and expire over a single call instead of two different incrby and expire calls is more performant.

> •limit-req: Uses the atomic SET ... EX command.

apisix/utils/redis-schema.lua
shreemaan-abhishek: No it is not. Good catch. Let me fix this.

apisix/plugins/limit-conn/util.lua
nic-6443: I think using the pipeline is a bit too complicated. For limit-req, I support you continuing to use the additional `expire` call. In fact, my previous suggestion was only for limit-conn.

membphis (commented):
I think:

you'd better split this PR to two PRs

One PR for one plugin, eg: limit-count, limit-req and limit-conn

Smaller PR is much easier to review

apisix/plugins/limit-conn/util.lua
shreemaan-abhishek: > I think using the pipeline is a bit too complicated.

why? Even if it is complicated, it should not stop us from implementing measures for performance optimisation right?

ChuanFF:
> when an APISIX node crashes and is unable to perform the "decrement count" operation, this key remains permanently in Redis as there is not expiry configured, causing the user's concurrency quota to be permanently occupied or even the user to be banned.

Setting a TTL for Redis keys does not seem to solve the scenario where a node in an APISIX cluster goes down under high traffic. When the traffic is heavy, APISIX keeps performing Redis operations, which continuously refresh the TTL. As a result, the counter associated with the failed node in Redis cannot be reset through TTL expiration.

shreemaan-abhishek:
> As a result, the counter associated with the failed node in Redis cannot be reset through TTL expiration.

I am not sure if I fully understand what you are trying to convey. The counter key will be same for all apisix instances in a cluster. Am I missing something 🤔

t/plugin/limit-conn-redis-cluster-ttl.t
shreemaan-abhishek: done

apisix/utils/redis-schema.lua
```diff
@@ -74,8 +74,20 @@ local policy_to_additional_properties = {
     },
 }
 
+local limit_conn_redis_cluster_schema = policy_to_additional_properties["redis-cluster"]
+limit_conn_redis_cluster_schema.properties.key_ttl = {
+    type = "integer", default = 60,
```
membphis: I think the default value can be `3600`, means 1hour

it should be enough

shreemaan-abhishek: makes sense, I have set the default ttl to be 3600.

membphis (dismissed):
LGTM

nic-6443 (dismissed):

ChuanFF:
@shreemaan-abhishek 

**Example:** an APISIX cluster with two nodes, **A** and **B**.  
Configuration: `limit_key = limit_conn`, `conf.key_ttl = 10s`, `conf.conn = 10`, `conf.burst = 0`.

- **t = 1s:**  
  Five requests arrive at node A. Redis `INCR` is called with `5`, resulting in  
  `limit_conn = 5`, `TTL = 10s`.

- **t = 5s:**  
  Five requests arrive at node B. Redis `INCR` is called with `5`, resulting in  
  `limit_conn = 10`, `TTL = 10s`.

- **t = 10s:**  
  Node A crashes. At this moment, the Redis key remains  
  `limit_conn = 10` with `TTL = 5s`.

- **t = 14s:**  
  Five requests finish on node B. Redis is updated, resulting in  
  `limit_conn = 5`, `TTL = 10s`.

- **t = 20s:**  
  Six requests finish on node B. Redis is updated to  
  `limit_conn = 11`, `TTL = 10s`.  
  Since `11 > conf.conn`, one request is rejected.

Therefore, in scenarios with continuously ongoing requests, it is difficult to rely on TTL expiration to clear the counters of nodes that have gone down.

apisix/utils/redis-schema.lua
```diff
@@ -74,8 +74,20 @@ local policy_to_additional_properties = {
     },
 }
 
+local limit_conn_redis_cluster_schema = policy_to_additional_properties["redis-cluster"]
+limit_conn_redis_cluster_schema.properties.key_ttl = {
+    type = "integer", default = 3600,
```
nic-6443, line 79: If a default value is set for ttl, then the check in the program when `self.conf.key_ttl` is nil becomes meaningless.

docs/en/latest/plugins/limit-conn.md
shreemaan-abhishek: done

apisix/utils/redis-schema.lua
shreemaan-abhishek: fixed

shreemaan-abhishek:
@ChuanFF I have updated the redis script to ensure each redis key has a ttl assigned to it. So the issue that you described should not occur.

apisix/plugins/limit-conn/util.lua
```diff
@@ -18,36 +18,66 @@
...
-        conn, err = red:incrby(key, 1)
-        if not conn then
+        local req_id = ngx.ctx.request_id or uuid.generate_v4()
+        if not ngx.ctx.limit_conn_req_ids then
+            ngx.ctx.limit_conn_req_ids = {}
+        end
+        ngx.ctx.limit_conn_req_ids[raw_key] = req_id
+
+        local now = ngx_time()
+        local res, err = red:eval(redis_incoming_script, 1, key,
```
membphis: you can take a look at this example, which is better performance

we can cache the `sha1` of Lua script, then we call `evalsha`

```lua
local redis = require "resty.redis"
local red = redis:new()

-- 1. Define your fixed script and pre-calculate its SHA1 locally
-- You can use an online SHA1 tool or calculate it via code
local script_content = "return redis.call('GET', KEYS[1])"
local script_sha1 = "6bce267422d7793d3e0034d7c99602029c5737e6"

red:set_timeout(1000) -- 1 sec
local ok, err = red:connect("127.0.0.1", 6379)
if not ok then
    ngx.log(ngx.ERR, "failed to connect: ", err)
    return
end

--- Function to execute script with EVALSHA optimization
local function optimized_eval(red, sha1, script, num_keys, ...)
    -- Try EVALSHA first
    local res, err = red:evalsha(sha1, num_keys, ...)
    
    -- If the script is missing in Redis (NOSCRIPT error)
    if not res and err and string.find(err, "NOSCRIPT") then
        ngx.log(ngx.INFO, "Script not found in cache, loading now...")
        
        -- Load the script to get the SHA1 (or verify it)
        local new_sha, load_err = red:script("LOAD", script)
        if not new_sha then
            return nil, "failed to load script: " .. (load_err or "")
        end
        
        -- Retry EVALSHA with the loaded script
        return red:evalsha(new_sha, num_keys, ...)
    end
    
    return res, err
end

-- 2. Call the function
local res, err = optimized_eval(red, script_sha1, script_content, 1, "my_key")

if not res then
    ngx.say("Error: ", err)
else
    ngx.say("Result: ", res)
end

-- Put it back in the connection pool
red:set_keepalive(10000, 100)

**Review by nic-6443 — DISMISSED** [review]

File: apisix/plugins/limit-conn/util.lua (hunk 9 above)
Comment (shreemaan-abhishek) [reply apisix/plugins/limit-conn/util.lua]: done

File: apisix/plugins/limit-conn/util.lua [hunk 10]
```diff
@@ -18,36 +18,76 @@
...
-        conn, err = red:incrby(key, 1)
-        if not conn then
+        local req_id = ngx.ctx.request_id or uuid.generate_v4()
+        if not ngx.ctx.limit_conn_req_ids then
+            ngx.ctx.limit_conn_req_ids = {}
+        end
+        ngx.ctx.limit_conn_req_ids[raw_key] = req_id
+
+        local now = ngx_time()
+        local res, err = red:evalsha(redis_incoming_script_sha, 1, key,
```
membphis: another example, gen the sha1 by `redis:script`:

```lua
local function fetch_latest_redis_script_sha1()
    if redis_incoming_script_sha then
        return redis_incoming_script_sha
    end
    
    local sha1, load_err = red:script("LOAD", lua_script)
    if load_err then
        return nil, load_err
    end

    redis_incoming_script_sha = sha1
    return redis_incoming_script_sha
end

local sha1, err = fetch_latest_redis_script_sha1()
if not sha1 then
    ngx.log(ngx.ERR, "Failed to load Lua script into Redis: ", err)
    return nil, err
end

local res, err = red:evalsha(sha1, ...)
if err and core.string.has_prefix(err, "NOSCRIPT") then
    redis_incoming_script_sha = nil
end

membphis: it is more safer than we generate it by `ngx api`

shreemaan-abhishek: got it, I have fixed this too now.

FYI, `lua-resty-redis-cluster` didn't work with `evalsha` so I ensured a conditional usage of plain redis eval by rediscluster policy and evalsha usage by `redis` policy.

membphis (approved):

nic-6443 (approved):

AlinsRan (approved):

PR #12565, opened 2025-08-29
feat: add support for azure-ai driver

Revolyssup:
```bash
❯ curl http://127.0.0.1:9080/v1/chat/completions  -XPOST -H "Content-Type: application/json"  -d '{
        "messages": [
            { "role": "system", "content": "You are a mathematician" },
            { "role": "user", "a": 1, "content": "Correct the spelling in lostu" }
        ], "stream": false
    }' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1442    0  1231  100   211    212     36  0:00:05  0:00:05 --:--:--   324
{
  "choices": [
    {
      "content_filter_results": {
        "hate": {
          "filtered": false,
          "severity": "safe"
        },
        "protected_material_code": {
          "filtered": false,
          "detected": false
        },
        "protected_material_text": {
          "filtered": false,
          "detected": false
        },
        "self_harm": {
          "filtered": false,
          "severity": "safe"
        },
        "sexual": {
          "filtered": false,
          "severity": "safe"
        },
        "violence": {
          "filtered": false,
          "severity": "safe"
        }
      },
      "finish_reason": "stop",
      "index": 0,
      "logprobs": null,
      "message": {
        "content": "It looks like you meant to write \"lost.\"",
        "refusal": null,
        "role": "assistant"
      }
    }
  ],
  "created": 1745912572,
  "id": "chatcmpl-BRZu83059FcmuuMYD8mqdDk9kJ4Yv",
  "model": "gpt-4o-2024-05-13",
  "object": "chat.completion",
  "prompt_filter_results": [
    {
      "prompt_index": 0,
      "content_filter_results": {
        "hate": {
          "filtered": false,
          "severity": "safe"
        },
        "jailbreak": {
          "filtered": false,
          "detected": false
        },
        "self_harm": {
          "filtered": false,
          "severity": "safe"
        },
        "sexual": {
          "filtered": false,
          "severity": "safe"
        },
        "violence": {
          "filtered": false,
          "severity": "safe"
        }
      }
    }
  ],
  "system_fingerprint": "fp_ee1d74bde0",
  "usage": {
    "completion_tokens": 11,
    "completion_tokens_details": {
      "accepted_prediction_tokens": 0,
      "audio_tokens": 0,
      "reasoning_tokens": 0,
      "rejected_prediction_tokens": 0
    },
    "prompt_tokens": 22,
    "prompt_tokens_details": {
[...]

PR #12878, opened 2026-01-09
feat: add support for openrouter

SkyeYoung:
### Description

Refer to the changes in https://github.com/apache/apisix/pull/12565/ to add support for openrouter

Already tested in real world:

```shell
young@ubuntu-20:~/apisix$ apisix restart
/usr/local/openresty//luajit/bin/luajit ./apisix/cli/apisix.lua restart
configuration test is successful
nginx: [warn] [lua] config_yaml.lua:198: read_apisix_config(): config file /home/young/apisix/conf/apisix.yaml reloaded.
young@ubuntu-20:~/apisix$ curl -X POST http://localhost:9080/anything \
>   -H "Content-Type: application/json" \
>   -d '{"messages": [{"role": "user", "content": "Hello"}]}'
data: {"id":"gen-1767948065-l9Xfdy3Lv0vOkdZq5gxE","provider":"OpenAI","model":"openai/gpt-4o","object":"chat.completion.chunk","created":1767948065,"choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":null,"native_finish_reason":null,"logprobs":null}],"system_fingerprint":"fp_deacdd5f6f"}

data: {"id":"gen-1767948065-l9Xfdy3Lv0vOkdZq5gxE","provider":"OpenAI","model":"openai/gpt-4o","object":"chat.completion.chunk","created":1767948065,"choices":[{"index":0,"delta":{"role":"assistant","content":"Hello"},"finish_reason":null,"native_finish_reason":null,"logprobs":null}],"system_fingerprint":"fp_deacdd5f6f"}

data: {"id":"gen-1767948065-l9Xfdy3Lv0vOkdZq5gxE","provider":"OpenAI","model":"openai/gpt-4o","object":"chat.completion.chunk","created":1767948065,"choices":[{"index":0,"delta":{"role":"assistant","content":"!"},"finish_reason":null,"native_finish_reason":null,"logprobs":null}],"system_fingerprint":"fp_deacdd5f6f"}

data: {"id":"gen-1767948065-l9Xfdy3Lv0vOkdZq5gxE","provider":"OpenAI","model":"openai/gpt-4o","object":"chat.completion.chunk","created":1767948065,"choices":[{"index":0,"delta":{"role":"assistant","content":" How"},"finish_reason":null,"native_finish_reason":null,"logprobs":null}],"system_fingerprint":"fp_deacdd5f6f"}

[...]

PR #12881, opened 2026-01-09
feat: support anthropic openai api

SkyeYoung:
### Description

Refer to the changes in https://github.com/apache/apisix/pull/12565 to add support for anthropic

Already tested in real world:

```shell
ubuntu-20:~/apisix$ curl -X POST http://localhost:9080/anything   -H "Content-Type: application/json"   -d '{"messages": [{"role": "user", "content": "Hello"}]}'
data: {"id":"msg_01Nsk7mHmmCAxsEpWwp2t2db","choices":[{"index":0,"delta":{"role":"assistant"}}],"created":1767952346,"model":"claude-sonnet-4-5-20250929","object":"chat.completion.chunk"        }

data: {"id":"msg_01Nsk7mHmmCAxsEpWwp2t2db","choices":[{"index":0,"delta":{"content":"Hello! How can I help you today"}}],"created":1767952346,"model":"claude-sonnet-4-5-20250929","object":"chat.completion.chunk"  }

data: {"id":"msg_01Nsk7mHmmCAxsEpWwp2t2db","choices":[{"index":0,"delta":{"content":"?"}}],"created":1767952346,"model":"claude-sonnet-4-5-20250929","object":"chat.completion.chunk"        }

data: {"id":"msg_01Nsk7mHmmCAxsEpWwp2t2db","choices":[{"index":0,"delta":{},"finish_reason":"stop"}],"created":1767952346,"model":"claude-sonnet-4-5-20250929","object":"chat.completion.chunk"      }

data: [DONE]
```

Fixes #

### Checklist


SkyeYoung:
@moonming 

1. pls check https://platform.claude.com/docs/en/api/openai-sdk. Just like Azure has its own API but also provides OpenAI compatible API, anthropic also provides a similar compatibility layer. 
2. The reason is the same as above. Currently apisix does not use azure’s own api, but azure compatible API(so it's `azure-ai` but not `azure`). I think it can be consistent with the current azure-openai at this point. The testing part also completely refers to the existing practices of the project. If you need to use real API requests, please add the relevant API KEY secret in the project so that I can test it.
3. This seems to be an issue caused by me adjusting the document locally, I will adjust that.

SkyeYoung:
waiting merge of https://github.com/apache/apisix/pull/12878

moonming (dismissed):

membphis (dismissed):

apisix/plugins/ai-drivers/anthropic.lua
bzp2010, line 1: @moonming @membphis 

Has this naming been confirmed? Do we need to add an explicit suffix hint for "using compatible API"? This seems inconsistent with our expectations. This holds true even if you intend to express a "vendor-product branch."

I believe that regardless of the AI we use internally, whether it's in compatibility mode or not has no impact on users. Since all APIs are now exposed in OpenAI format, users don't need to know the format used upstream—they just need the functionality to work.

Therefore, `anthropic` suffices as this provider value.

SkyeYoung: My initial idea came from the naming of azure-openai, which has already been expressed above https://github.com/apache/apisix/pull/12881#issuecomment-3734876551. The advantage of doing this is to leave room for the Anthropic API.

But it seems you already have different ideas.

bzp2010 (changes requested):
Temporarily block the merge until the provider name is confirmed.

apisix/plugins/ai-drivers/anthropic.lua
membphis: agree to use `anthropic` as provider name

nic-6443 (approved):

PR #12885, opened 2026-01-09
fix(cli): remove stale key after resolving env var in config keys

### LINKED ISSUE #10505: help request: Failed to parse domain
### Description

Context: I'm trying to build an application with an Auth Service which should only validate the tokens, and admin service which sould be protected. 

I am running APIsix from docker container, but I keep getting errors about failing to parse the domain.
When runing echo $variable_name inside the container, it echoes perfectly. But I still get these errors inside the log file.

The calls to the Auth service to check for the API are actually working, and the call to Admin service as well. But I still get these errors in the logs

On another hand, when effectuating these API calls from a front-end application, I get a CORS error, and I noted that the authorization header is not present inside the request sent with forward-auth plugin (even though I've added it inside the request_headers) 

Here are the logs

```

2023/11/16 11:08:54 [error] 55#55: *160063 [lua] resolver.lua:80: parse_domain(): failed to parse domain: ${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}}, error: failed to query the DNSserver: dns server error: 3 name error, client: 172.21.0.1, server: _, request: "OPTIONS /auth/greet HTTP/1.1", host: "localhost:8080", referrer: "http://localhost:4200/"
2023/11/16 11:08:54 [error] 55#55: *160063 [lua] upstream.lua:79: parse_domain_for_nodes(): dns resolver domain: ${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}} error: failed to query the DNS server: dns server error: 3 name error, client: 172.21.0.1, server: _, request: "OPTIONS /auth/greet HTTP/1.1", host: "localhost:8080", referrer: "http://localhost:4200/"
2023/11/16 11:08:54 [error] 55#55: *160063 [lua] resolver.lua:80: parse_domain(): failed to parse domain: ${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}}, error: failed to query the DNSserver: dns server error: 3 name error, client: 172.21.0.1, server: _, request: "GET /auth/greet HTTP/1.1", host: "localhost:8080", referrer: "http://localhost:4200/"
2023/11/16 11:08:54 [error] 55#55: *160063 [lua] upstream.lua:79: parse_domain_for_nodes(): dns resolver domain: ${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}} error: failed to query the DNS server: dns server error: 3 name error, client: 172.21.0.1, server: _, request: "GET /auth/greet HTTP/1.1", host: "localhost:8080", referrer: "http://localhost:4200/"
2023/11/16 11:08:55 [warn] 55#55: *160063 [lua] plugin.lua:1107: run_plugin(): forward-auth exits with http status code 401, client: 172.21.0.1, server: _, request: "OPTIONS /admin/users HTTP/1.1", host: "localhost:8080", referrer: "http://localhost:4200/"
```

apisix.yaml
```

routes:
  - uri: /auth/*
    plugins:
      proxy-rewrite:
        regex_uri: ["^/auth/(.*)","/$1"]

    upstream:
      nodes:
        "${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}}": 1
      type: roundrobin
    
  - uri: /admin/*
    plugins:  
      proxy-rewrite:
        regex_uri: ["^/admin/(.*)","/$1"]
      forward-auth:
        uri: http://${{AUTH_SERVICE_IP:=}}:${{AUTH_SERVICE_PORT:=3000}}/validate
        ssl_verify: false
        request_headers: ["Authorization",authorization]
    upstream:
      nodes:
        "${{ADMIN_SERVICE_IP:=192.168.1.76}}:${{ADMIN_SERVICE_PORT:=3001}}": 1
    
#END

```

### Environment

APISIX: 3.6 dockerized
etcd:3.4.15 dockerized

### LINKED ISSUE #12884: bug: pulling env vars in yaml keys works, but results in error message in logs
### Current Behavior

When using ENV vars in yaml keys, for example in 
```
    upstream:
      nodes:
          "${{HOST}}": 1
```
The actual routing is works, but log messages appears: 
> failed to parse domain: ${{HOST}}, error: failed to query the DNS server: dns server error: 3 name error

But the routing works even with this error message. 

### Expected Behavior

Routing should work without error message referring failed dns query on unresolved variable name.

### Error Logs

2026/01/09 16:02:05 [error] 35#35: *798 [lua] resolver.lua:80: parse_domain(): failed to parse domain: ${{HOST}}, error: failed to query the DNS server: dns server error: 3 name error, client: 172.18.0.1, server: _, request: "GET /img/favicon.svg HTTP/2.0", host: "local.example.ru", referrer: "https://local.example.ru/"
2026/01/09 16:02:05 [error] 35#35: *798 [lua] upstream.lua:97: parse_domain_for_nodes(): dns resolver domain: ${{HOST}} error: failed to query the DNS server: dns server error: 3 name error, client: 172.18.0.1, server: _, request: "GET /img/favicon.svg HTTP/2.0", host: "local.example.ru", referrer: "https://local.example.ru/"

### Steps to Reproduce

1. Configure apisix in standalone mode. 
2. Add route 
```
routes:   
  - uri: /*
    upstream:
      nodes:
          "${{HOST}}": 1
      type: roundrobin
```
3. Add environment variable 
`export HOST=localhost:8080`
4. Start apisix
5. Request resource that matches configured route.
6. Observe requested resource (routing works)
7. Observe log message with an error "failed to parse domain: ${{HOST}}"

### Environment

- APISIX version (run `apisix version`): 3.14.1
- Operating system (run `uname -a`): docker image apache/apisix:3.14.1-debian
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): Not applicable
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): Not applicable
- APISIX Dashboard version, if relevant: Not applicable
- Plugin runner version, for issues related to plugin runners: Not applicable
- LuaRocks version, for installation issues (run `luarocks --version`): Not applicable

pronchakov:
### Description

There is a little bug in function `local function resolve_conf_var(conf)` in `cli/file.lua`
`conf.key = nil` deletes entry from conf with name "key"
but  after variable substitution, old entry with name contained in variable `key` should be removed.
so it needs to be rewritten to `conf[key] = nil`

`conf.key = nil` results in adding entry with resolved variable to conf, but doesn't remove entry with unresolved variable name.
In case of using it in routes->upstream->nodes it results in two node entries. one is correct and one with unresolved variable name in it, that results in dns error in logs. I suppose that another places will result in different consequences. 

Fixes #12884

### Checklist

Fixes #10505

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -105,7 +105,7 @@ local function resolve_conf_var(conf)
             end
             if new_key ~= key then
                 new_keys[new_key] = "dummy" -- we only care about checking the key
-                conf.key = nil
+                conf[key] = nil
                 conf[new_key] = val
                 key = new_key
             end
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -91,25 +91,8 @@ end
 
 
 local function resolve_conf_var(conf)
-    local new_keys = {}
+    local renamed_keys
     for key, val in pairs(conf) do
-        -- avoid re-iterating the table for already iterated key
-        if new_keys[key] then
-            goto continue
-        end
-        -- substitute environment variables from conf keys
-        if type(key) == "string" then
-            local new_key, _, err = var_sub(key)
-            if err then
-                return nil, err
-            end
-            if new_key ~= key then
-                new_keys[new_key] = "dummy" -- we only care about checking the key
-                conf[key] = nil
-                conf[new_key] = val
-                key = new_key
-            end
-        end
         if type(val) == "table" then
             local ok, err = resolve_conf_var(val)
             if not ok then
@@ -135,7 +118,27 @@ local function resolve_conf_var(conf)
 
             conf[key] = new_val
         end
-        ::continue::
+
+        -- substitute environment variables from conf keys. The rename is
+        -- deferred because inserting a new key while iterating with pairs()
+        -- is undefined behavior.
+        if type(key) == "string" then
+            local new_key, _, err = var_sub(key)
+            if err then
+                return nil, err
+            end
+            if new_key ~= key then
+                renamed_keys = renamed_keys or {}
+                renamed_keys[key] = new_key
+            end
+        end
+    end
+
+    if renamed_keys then
+        for key, new_key in pairs(renamed_keys) do
+            conf[new_key] = conf[key]
+            conf[key] = nil
+        end
     end
 
     return true
--- a/t/cli/test_http_config.sh
+++ b/t/cli/test_http_config.sh
@@ -39,6 +39,31 @@ echo "passed: define custom shdict"
 
 git checkout conf/config.yaml
 
+# resolve environment variable used as a config key: the resolved key must be
+# rendered and the stale unresolved key must be removed from the config table
+echo '
+nginx_config:
+  http:
+    custom_lua_shared_dict:
+      "${{DICT_NAME}}": 1m
+' > conf/config.yaml
+
+DICT_NAME=env_dict make init
+
+if ! grep "lua_shared_dict env_dict 1m;" conf/nginx.conf > /dev/null; then
+    echo "failed: resolve env var used as config key"
+    exit 1
+fi
+
+if grep -F '${{DICT_NAME}}' conf/nginx.conf > /dev/null; then
+    echo "failed: stale unresolved env var key should be removed"
+    exit 1
+fi
+
+echo "passed: resolve env var used as config key"
+
+git checkout conf/config.yaml
+
 echo "
 plugins:
     - ip-restriction
--- a/t/config-center-json/route-upstream.t
+++ b/t/config-center-json/route-upstream.t
@@ -21,6 +21,8 @@ log_level('info');
 no_root_location();
 no_shuffle();
 
+$ENV{TEST_UPSTREAM_NODE} = '127.0.0.1:1980';
+
 run_tests();
 
 __DATA__
@@ -242,3 +244,29 @@ GET /hello
 test: one
 --- error_log
 proxy request to 127.0.0.1:1980
+
+
+
+=== TEST 8: env var used as a node key is resolved, stale key is removed
+--- yaml_config eval: $::yaml_config
+--- apisix_json
+{
+  "routes": [
+    {
+      "id": 1,
+      "uri": "/hello",
+      "upstream": {
+        "nodes": {
+          "${{TEST_UPSTREAM_NODE}}": 1
+        },
+        "type": "roundrobin"
+      }
+    }
+  ]
+}
+--- request
+GET /hello
+--- response_body
+hello world
+--- no_error_log
+failed to parse domain


Baoyuantop:
Hi @pronchakov, can you add corresponding tests for this fix?

pronchakov:
> Hi @pronchakov, can you add corresponding tests for this fix?

Ok, I'll add some tests.

moonming (changes requested):
Hi @pronchakov, thank you for catching this bug!

This is a clear and correct fix — `conf.key = nil` should be `conf[key] = nil`. The current code sets a literal field named `key` to nil instead of deleting the dynamic key. Great catch!

**One request**: Could you please add a test case that verifies this behavior? Something like:
- Define a route/config where a key name contains an environment variable reference (e.g., `\${{ENV_KEY_NAME}}`)
- Verify that after resolution, the old key is properly removed and the new key exists

This is a one-line fix with clear correctness, so once the test is added, this should be ready to merge. Thank you!

Baoyuantop:
Hi @pronchakov, following up on the previous review comments. Please let us know if you have any updates. Thank you.

apisix/cli/file.lua
```diff
@@ -105,7 +105,7 @@ local function resolve_conf_var(conf)
             end
             if new_key ~= key then
                 new_keys[new_key] = "dummy" -- we only care about checking the key
-                conf.key = nil
+                conf[key] = nil
```
nic-6443: The fix itself looks right. One heads-up for the test that was requested: since #13078, standalone `apisix.yaml` substitutes env vars in the raw text *before* YAML parsing (`resolve_conf_var_in_text`), so the original repro from #12884 (env var as an upstream node key in apisix.yaml) no longer reaches this function on master. A test modeled on the issue's repro steps won't cover this line. The remaining callers of `resolve_conf_var` are `config.yaml` (this file, in `read_yaml_conf`) and JSON standalone mode (`apisix.json` in `core/config_yaml.lua`), so the test needs to target one of those.

An easy way is a shell test for an env var used as a `config.yaml` key, e.g. in `t/cli/test_http_config.sh` next to the existing custom shdict case:

```sh
echo '
nginx_config:
  http:
    custom_lua_shared_dict:
      "${{DICT_NAME}}": 1m
' > conf/config.yaml

DICT_NAME=test_dict make init

if ! grep "lua_shared_dict test_dict 1m;" conf/nginx.conf > /dev/null; then
    echo "failed: resolve env var in config key"
    exit 1
fi

if grep "DICT_NAME" conf/nginx.conf > /dev/null; then
    echo "failed: stale unresolved env var key should be removed"
    exit 1
fi
```

The second grep fails without this fix (the stale `${{DICT_NAME}}` entry stays in the table and gets rendered as an extra `lua_shared_dict` line) and passes with it.

AlinsRan: Done — went with the `config.yaml` custom_lua_shared_dict approach as you suggested. Used `grep -F '${{DICT_NAME}}'` for the stale-key check instead of `grep DICT_NAME`, since APISIX emits an `env DICT_NAME;` directive that would match the latter regardless of the fix.

AlinsRan:
@pronchakov I pushed two commits to your branch — one reshapes the fix, so here's why.

**The one-char fix isn't enough.** `conf[key] = nil` is right, but the delete and the insert (`conf[new_key] = val`) both stay inside the `pairs()` loop. Inserting into a table you're iterating with `pairs()` is undefined in Lua — LuaJIT may raise `invalid key to 'next'`, skip entries, or revisit them. `new_keys` only papered over one symptom (the new key being visited again).

`new_keys` also has a bug of its own: the guard `if new_keys[key]` is keyed on the *resolved* name, so if a resolved key collides with another key already in the same table, that entry gets skipped wholesale and its own placeholders never get substituted.

**New shape:** collect renames during the traversal, apply them after. No insert while iterating, so `new_keys` and the `goto` both disappear.

```lua
    if renamed_keys then
        for key, new_key in pairs(renamed_keys) do
            conf[new_key] = conf[key]
            conf[key] = nil
        end
    end
```


**Tests.** Yours covers the `config.yaml` caller. I added `t/config-center-json/route-upstream.t` TEST 8: env var as an upstream node key in standalone `apisix.json`, the closest repro of #12884 that still reaches `resolve_conf_var()` on master (the `apisix.yaml` path stopped calling it after #13078).

Worth noting what it exposes — with the pre-fix code the request doesn't just log a DNS error, it **times out**. Roundrobin hits the stale unresolved node and hangs. The impact in standalone JSON is heavier than the issue suggests.

Verified: luacheck clean, TEST 8 passes with the fix and fails without it, `t/core/config.t` still green. I also retitled the PR.

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

moonming (approved):

PR #12900, opened 2026-01-13
fix(limit-count): use meta parent to identify plugin source

### LINKED ISSUE #12896: bug: ai-rate-limiting is applied globally instead of per consumer
### Current Behavior

When I add a consumer (e.g. georg) with an ai-rate-limitin configuration as shown in the example, and then create a second consumer (e.g. martin) with the same plugin configuration but a different key-auth API key, both consumers end up sharing the same rate limit for an instance.
Once georg reaches the configured token limit, request from martin are also rejected with 429 "Configured rate limit reached", even though martin has his own consumer entry and API key.
The rate limitin appears to be applied globally per model, rather than per consumer, as described in the documentation. 
```
{
  "username":"georg",
  "plugins":{
    "key-auth":{
      "key":"Bearer "
    },
    "ai-rate-limiting":{
      "instances":[
        {
          "name":"gpt-oss-120b",
          "limit_strategy":"prompt_tokens",
          "time_window":60,
          "limit":400
        },
        {
          "name":"bge-m3",
          "limit_strategy":"prompt_tokens",
          "time_window":60,
          "limit":10000
        },
        {
          "name":"gpt-oss-120b",
          "limit_strategy":"completion_tokens",
          "time_window":60,
          "limit":400
        }
      ],
      "rejected_code":429,
      "rejected_msg":"Configured rate limit reached",
      "show_limit_quota_header":true
    }
  }
}
```
### Expected Behavior

Each consumer should have an independent rate limit quota.
When multiple consumers (e.g. georg and marting) are configured  with the same ai-rate-limiting plugin settings but different key-auth API keys, the token limit should be enforced per consumer, not shared globally.
If georg reaches his configured token limit, only request from georg should be rejected with 429 ..., while martin should still be able to make requests within his own quota.

### Error Logs

_No response_

### Steps to Reproduce

- Create a Consumer named georg with key-auth enabled and configure the ai-rate-limiting plugin with a token limit (e.g. 400 tokens per 60 seconds for an instance).
- Create a second Consumer named martin with the same ai-rate-limiting configuration, but with a different key-auth API key.
- Send requests using georg’s API key until the configured token limit is reached and requests start returning: "429 Configured rate limit reached"
- Immediately send requests using martin’s API key.
- Observe that martin’s requests are also rejected with: "429 Configured rate limit reached"

### Environment

  3.14.1

  Linux (Kubernetes container, official APISIX Docker image)

  OpenResty (bundled with APISIX Docker image)

  v3.6.0 (self-deployed in Kubernetes)

  Not used

  Not used (using a serverless-post-function)

  Not applicable (using official Docker image)

shreemaan-abhishek:
### Description

When configuring rate limiting plugin in different consumers, quota usage for one consumer would affect another consumer's quota too. This happened because the rate limiting key being constructed would be unaware of the source of the rate limiting plugin hence the rate limiting key for different consumer would produce same key for a given route.

To fix this we ensure that parent information is filled in to plugin configurations for plugins configured in a consumer.

Fixes #12896

### Checklist

### DIFF
--- a/t/node/plugin-configs.t
+++ b/t/node/plugin-configs.t
@@ -411,7 +411,7 @@ hello world
 
 
 
-=== TEST 7: change plugin config will cause the conf_version change
+=== TEST 7: configure plugin configs with limit-count to check parent association
 --- config
     location /t {
         content_by_lua_block {


apisix/plugins/limit-count/init.lua
```diff
@@ -205,8 +205,21 @@ local function gen_limit_key(conf, ctx, key)
...
-    local conf_type = ctx.conf_type_without_consumer or ctx.conf_type
-    local conf_id = ctx.conf_id_without_consumer or ctx.conf_id
+    local parent = conf._meta and conf._meta.parent
+    if not parent or not parent.resource_key or not parent.resource_version then
+        core.log.error("failed to generate key invalid parent: ", core.json.encode(parent))
+        return nil
+    end
+
+    local match, err = ngx.re.match(parent.resource_key, [[/apisix/([^/]+)/([^/]+)]])
+
```
membphis: remove this useless line

apisix/plugins/limit-count/init.lua
```diff
@@ -205,8 +205,21 @@ local function gen_limit_key(conf, ctx, key)
...
+    if not match then
+        core.log.error("failed to generate key invalid parent resource key, err: ", err)
```
membphis: I think we can remove the `etcd_prefix`, then it should be enough

we can always use it directly, eg: `routes/r1`

```yaml
etcd:
  host:
    - "http://127.0.0.1:2379" # Replace with your etcd cluster host(s)
  prefix: "/apisix" # The default prefix, modify as needed
  timeout: 30

File: apisix/plugins/limit-count/init.lua [hunk 3]
```diff
@@ -205,8 +205,21 @@ local function gen_limit_key(conf, ctx, key)
...
+        return nil
+    end
+
+    local conf_type = match[1]
+    local conf_id = match[2]
     local new_key = conf_type .. conf_id .. ':' .. apisix_plugin.conf_version(conf)
```
nic-6443:
```suggestion
    local parent = conf._meta and conf._meta.parent

    local new_key = parent.resource_key .. ':' .. apisix_plugin.conf_version(conf)
```

nic-6443: I think it's okay to directly use  parent.resource_key   as the prefix for rate limiting, without any additional processing. Currently, splitting out  conf_type   and  conf_id   and then concatenating them into the key's prefix is completely unnecessary.

membphis (approved):

AlinsRan (approved):

PR #12930, opened 2026-01-21
change: remove lua-resty-worker-events

shreemaan-abhishek:
### Description

https://github.com/apache/apisix/pull/10550 implemented support for `lua-resty-events`, it's been long since; so we can remove the old events module lua-resty-worker-events.

### Checklist

### DIFF
--- a/t/plugin/ai-proxy-multi3.t
+++ b/t/plugin/ai-proxy-multi3.t
@@ -928,51 +928,84 @@ POST /ai
                 return original_parse_domain(host)
             end
             -- Create a route with health check that uses the domain
-            local code, body = t('/apisix/admin/routes/1',
-                 ngx.HTTP_PUT,
-                 [[{
-                    "uri": "/ai",
-                    "plugins": {
-                        "ai-proxy-multi": {
-                            "instances": [
-                                {
-                                    "name": "openai-test",
-                                    "provider": "openai",
-                                    "weight": 1,
-                                    "priority": 1,
-                                    "auth": {
-                                        "header": {
-                                            "Authorization": "Bearer token"
-                                        }
-                                    },
-                                    "options": {
-                                        "model": "gpt-4"
-                                    },
-                                    "override": {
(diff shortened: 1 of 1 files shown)


t/plugin/ai-proxy-multi3.t
```diff
@@ -933,51 +928,84 @@ POST /ai
...
-                                            "healthy": {
-                                                "interval": 1,
-                                                "successes": 1
-                                            },
-                                            "unhealthy": {
-                                                "interval": 1,
-                                                "http_failures": 1
-                                            }
+            local core = require("apisix.core")
+            local route_config = {
```
shreemaan-abhishek, line 932: I had to convert json config to lua table to get rid of this error in CI:

Image

t/node/healthcheck-passive.t
```diff
@@ -1,373 +0,0 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-BEGIN {
-    if ($ENV{TEST_EVENTS_MODULE} ne "lua-resty-worker-events") {
-        $SkipReason = "Only for lua-resty-worker-events events module";
-    }
-}
```
nic-6443, line 22: Why can this test case only run using lua-resty-worker-events?

shreemaan-abhishek: I think this was mistakenly copied and has no real meaning as the tests pass when using lua-resty-events module too

shreemaan-abhishek: the deleted test cases have their counterparts that can run under lua-resty-events.

Baoyuantop:
Should we discuss this change on the mailing list? It appears to be a breaking change.

nic-6443:
> Should we discuss this change on the mailing list? It appears to be a breaking change.

These two modules are fully compatible, and `lua-resty-events` has been the default for some time now.

PR #12949, opened 2026-01-29
fix(docker): support valid YAML variations in standalone mode

### LINKED ISSUE #12427: bug: docker image standalone check fails on valid config files
### Current Behavior

In the apisix docker image, the `/check_standalone_config.sh`  checks that the config is in "standalone mode". However it uses hard-coded greps for strings in the yaml such as `role: data_plane`. This means the check will fail if the yaml is formated even slightly different, e.g.
```yaml
deployment:
  role: "data_plane"
```
or 
```yaml
deployment: {"role": "data_plane"}
```

Also, the check does not respect the `APISIX_PROFILE` environment variable and so will be looking in the wrong config file entirely if you are using multiple profiles.

### Expected Behavior

The docker image should accept valid YAML files.

### Error Logs
```
Error: /usr/local/apisix/conf/config.yaml does not contain 'role: data_plane'. Deployment role must be set to 'data_plane' for standalone mode.
Please refer to the APISIX documentation for deployment modes: https://apisix.apache.org/docs/apisix/deployment-modes/
```

### Steps to Reproduce

Run apisix via the docker image, with config set to something like this:
```yaml
 role: "data_plane"
  role_data_plane:
    config_provider: yaml
```

### Environment

Apisix docker image, tag `3.12.0`

janiussyafiq:
### Description

This PR fixes a bug in the Docker standalone configuration check script (check_standalone_config.sh) where valid YAML configurations would fail validation due to rigid pattern matching. and the case where `APISIX_PROFILE` does not respected. This PR also included necessary test cases.

Fixes #12427 

### Checklist

### DIFF
--- a/t/cli/test_standalone_docker.sh
+++ b/t/cli/test_standalone_docker.sh
@@ -28,8 +28,6 @@ standalone() {
 DOCKER_IMAGE="${DOCKER_IMAGE:-apache/apisix:master-debian-dev}"
 trap standalone EXIT
 
-make build-on-debian-dev
-
 echo 'routes: []
 #END' > conf/apisix.yaml
 
--- a/.github/workflows/docker-standalone.yml
+++ b/.github/workflows/docker-standalone.yml
@@ -34,6 +34,30 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v5
 
+      - name: Build APISIX Dashboard
+        run: |
+          # install node.js and pnpm
+          sudo n lts
+          corepack enable pnpm
+
+          # prepare apisix-dashboard source code
+          source .requirements
+          git clone --revision=${APISIX_DASHBOARD_COMMIT} --depth 1 https://github.com/apache/apisix-dashboard.git
+          pushd apisix-dashboard
+
+          # compile
+          pnpm install --frozen-lockfile
+          pnpm run build
+          popd
+
+          # copy the dist files to the ui directory
+          mkdir ui
+          cp -r apisix-dashboard/dist/* ui/
+          rm -r apisix-dashboard
+
+          # build Docker image
+          make build-on-debian-dev
+
       - name: Run Docker standalone test
         run: |
           ./t/cli/test_standalone_docker.sh


docker/utils/check_standalone_config.sh
```diff
@@ -16,20 +16,30 @@
...
+if ! grep -E -q '["'\'']?role_data_plane["'\'']?:' "$CONF_FILE"; then
+    echo "Error: $CONF_FILE does not contain 'role_data_plane:'."
     echo "Please refer to the APISIX documentation for deployment modes: https://apisix.apache.org/docs/apisix/deployment-modes/"
     exit 1
 fi
 
-if ! grep -q 'config_provider: yaml' "${PREFIX}/conf/config.yaml"; then
-    echo "Error: ${PREFIX}/conf/config.yaml does not contain 'config_provider: yaml'. Config provider must be set to 'yaml' for standalone mode."
+if ! grep -E -q '["'\'']?config_provider["'\'']?:\s*["'\'']?yaml["'\'']?' "$CONF_FILE"; then
+    echo "Error: $CONF_FILE does not contain 'config_provider: yaml'. Config provider must be set to 'yaml' for standalone mode."
```
shreemaan-abhishek: please add tests that cover the changes you made in this file.

janiussyafiq: could you provide more context on what tests are needed since i've written a few? do you mean the $CONF_FILE returning the correct file name?

docker/utils/check_standalone_config.sh
```diff
@@ -16,20 +16,30 @@
 # limitations under the License.
 #
 
-if ! grep -q 'role: data_plane' "${PREFIX}/conf/config.yaml"; then
-    echo "Error: ${PREFIX}/conf/config.yaml does not contain 'role: data_plane'. Deployment role must be set to 'data_plane' for standalone mode."
+CONF_FILE="${PREFIX}/conf/config.yaml"
```
nic-6443: I suggest that we should not validate `config.yaml` through the shell, as it is prone to errors. We can move the checking logic to ops.lua, so that we can validate `config.yaml` in a semantic way:
https://github.com/apache/apisix/blob/adf2ea1573a17dee97ff8a4ad2df835f3ba8a862/apisix/cli/ops.lua#L179-L186
Here, check the environment variables and fields in the `yaml_conf` object.

docker/debian-dev/docker-entrypoint.sh
shreemaan-abhishek, line 1: Image

you have changed the mode of this file, why?

janiussyafiq: might've changed it when i did local testing, will revert it back

t/cli/test_standalone_docker.sh
```diff
@@ -0,0 +1,216 @@
...
+if ! run_docker_test "true"; then
+    echo "failed: single quoted \"role: 'data_plane'\" was rejected"
+    exit 1
+fi
+
+echo "passed: single-quoted format accepted"
+
+# flow syntax
+echo '
+deployment: {"role": "data_plane", "role_data_plane": {"config_provider": "yaml"}}
```
shreemaan-abhishek, line 117: this is neither a valid json nor a valid yaml, why would using this config not fail?

t/cli/test_standalone_docker.sh
```diff
@@ -0,0 +1,216 @@
...
+standalone() {
+    clean_up
+    docker rm -f apisix-test-standalone 2>/dev/null || true
+    git checkout conf/apisix.yaml
+}
+
+DOCKER_IMAGE="${DOCKER_IMAGE:-apache/apisix:master-debian-dev}"
+trap standalone EXIT
+
+make build-on-debian-dev
```
shreemaan-abhishek: This is going to significantly increase the duration of the CI. We can take one of the following actions to avoid this:

- use prebuilt image (image built from previous push) instead of building image while running the test.
- run this test only on master branch
  - but this will require further more changes (e.g: a separate GitHub actions workflow)
  - and we won't be able to detect regressions before commit
- run this test on a different GitHub actions workflow and only run them if any of the files related to this feature (ops.lua, docker-entrypoint.sh, etc) change.

shreemaan-abhishek: 3rd option sounds good to me

t/cli/test_standalone_docker.sh
janiussyafiq: https://www.yamllint.com/ tried the syntax here, and it works - copy this code block
`deployment: {"role": "data_plane", "role_data_plane": {"config_provider": "yaml"}}
apisix:
    node_listen: 9080`
i think this also addressed here: https://github.com/apache/apisix/issues/12427

t/cli/test_standalone_docker.sh
janiussyafiq: opt 1: won't work, i've discussed with @nic-6443, since the older image still rely on the script that is decided to be removed `check_standalone_config.sh`, so test will fail
as for opt 2 and 3, can try those solution, i'm more leaning towards opt 3 wdyt?

janiussyafiq: will try implementing 3rd option thanks!

t/cli/test_standalone_docker.sh
shreemaan-abhishek: interesting..

nic-6443 (approved):

Baoyuantop (approved):

PR #12967, opened 2026-02-04
feat: support configuring variables in limit-conn, limit-count and ai-rate-limiting

shreemaan-abhishek:
### Description

Allowing to use variables in limiting configuration unlocks prospects for dynamic rate limiting strategies.

### Checklist

### DIFF
--- a/apisix/plugins/limit-count/init.lua
+++ b/apisix/plugins/limit-count/init.lua
@@ -231,7 +231,6 @@ local function create_limit_obj(conf, ctx, plugin_name)
 end
 
 
-
 local function gen_limit_key(conf, ctx, key)
     if conf.group then
         return conf.group .. ':' .. key
@@ -297,7 +296,6 @@ function _M.rate_limit(conf, ctx, name, cost, dry_run)
         key = ctx.var["remote_addr"]
     end
 
-    core.log.info("key bef: ", key, ". conf: ", core.json.encode(conf), ". ctx: ", type(ctx))
     key = gen_limit_key(conf, ctx, key)
     core.log.info("limit key: ", key)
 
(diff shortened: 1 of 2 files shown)


apisix/plugins/limit-count/init.lua
```diff
@@ -276,7 +271,32 @@ function _M.rate_limit(conf, ctx, name, cost, dry_run)
         key = ctx.var["remote_addr"]
     end
 
-    key = gen_limit_key(conf, ctx, key)
+    -- copied from gen_limit_key
```
nic-6443: Why copy the code from the `gen_limit_key` function? This will make `rate_limit` very long and hard to read. It doesn't seem like a necessary change.

nic-6443 (dismissed):

ronething (approved):
lgtm

apisix/plugins/limit-count/init.lua
```diff
@@ -276,6 +297,7 @@ function _M.rate_limit(conf, ctx, name, cost, dry_run)
         key = ctx.var["remote_addr"]
     end
 
+    core.log.info("key bef: ", key, ". conf: ", core.json.encode(conf), ". ctx: ", type(ctx))
```
AlinsRan:
```suggestion
    core.log.info("key bef: ", key, ". conf: ", core.json.delay_encode(conf), ". ctx: ", type(ctx))
```

AlinsRan: The type(ctx) seems to have no useful information.

shreemaan-abhishek: my bad, I had added this log just for debugging. This whole log line itself does not serve any meaningful purpose so I have removed it. 👍🏼

PR #12986, opened 2026-02-06
feat(openidc): support redis for session storage

### LINKED ISSUE #12495: help request: OIDC Login Loop & Refresh Token Failures with openid-connect Plugin for SPA + API Gateway Pattern
### Description

Hello APISIX Team,

I am facing a persistent issue trying to secure a Single-Page Application (SPA) and its backend API using the **openid-connect** plugin with Keycloak as the IdP. My goal is to have all security managed by the APISIX platform, without any OIDC logic in the frontend application itself.

## Architecture:
- APISIX as the API Gateway.
- Keycloak as the OIDC Identity Provider.
- Frontend: A static SPA served by an Nginx container.
- Backend: A REST/OData API served by another container.
- Both frontend and backend need to be protected, as the frontend needs user identity to render correctly.

## The Problem:
I cannot find a working configuration for the openid-connect plugin that avoids either 
a) token refresh failures, or 
b) login loops.

### Separate Routes, Separate OIDC Plugins, Unified Keycloak Client:

- Route 1 (/*, for frontend) with openid-connect plugin.
- Route 2 (/api/*, for backend) with an identical openid-connect plugin configuration.
- **Result**: Login works. Claim propagation works. However, after the access token expires, the token refresh fails. Keycloak logs show the error invalid_grant: "Token client and authorized client don't match". This happens because the refresh is initiated by the plugin instance on the /api/ route, but the token was originally obtained by the plugin instance on the /* route. The context (likely due to different redirect_uri configurations in the plugins) does not match.

### Shared Callback Route:

- To solve the context mismatch, I configured both openid-connect plugins to use the exact same redirect_uri (e.g., https://.../auth/callback).
- A dedicated route for /auth/callback handles the redirect back to the application.
- **Result**: This leads to a login loop. The OIDC flow is never completed because the /auth/callback route does not have the openid-connect plugin to handle the code-for-token exchange.

### Unified redirect_uri on Main Route:

- I configured both plugin instances to use the application root (https://.../) as the redirect_uri.
- **Result**: This causes the openid-connect plugin to fail with a 500 Internal Server Error on the initial visit to /. The logs show unhandled request to the redirect_uri: /. The plugin cannot handle when the application entry point is the same as the configured redirect_uri.

## Conclusion:
It seems impossible to have two separate routes protected by the openid-connect plugin in _bearer_only: false_ mode, as the session/refresh context cannot be reliably shared. At the same time, using a single route to protect both also presents challenges.

Is there a recommended, official architecture for this common use case (platform-managed security for an SPA)? Am I missing a configuration option, or is this a limitation of the current plugin?

Thank you for your help.

### Environment

- APISIX version (run `apisix version`): 3.11.0
- Operating system (run `uname -a`): Ubuntu 24.04.2 (kernel  6.8.0-64-generic)
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx v1 (docker image: nginx:1-alpine)

shreemaan-abhishek:
### Description

Fixes #12495

### Checklist

### DIFF
--- a/t/plugin/openid-connect-redis.t
+++ b/t/plugin/openid-connect-redis.t
@@ -289,7 +289,7 @@ property "port" validation failed: wrong type: expected integer, got string
 
             local httpc = http.new()
 
-            -- extract cookie value by name from Set-Cookie heade
+            -- extract cookie value by name from Set-Cookie header
             local function get_cookie(headers, name)
                 local cookies = headers["Set-Cookie"]
                 if not cookies then return nil end
@@ -316,7 +316,7 @@ property "port" validation failed: wrong type: expected integer, got string
                 return
             end
 
-            -- extract state from the Location URL egample: http://.../authorize?client_id=...&state=...&nonce=...
+            -- extract state from the Location URL example: http://.../authorize?client_id=...&state=...&nonce=...
             local loc = res.headers["Location"]
             local state = string.match(loc, "state=([^&]+)")
             if not state then
@@ -345,7 +345,7 @@ property "port" validation failed: wrong type: expected integer, got string
                 ngx.say("failed to get authenticated session cookie after callback. status: ", res.status)
                 return
             end
-            ngx.log(ngx.INFO, "dibag auth_cookie: ", auth_cookie)
+            ngx.log(ngx.INFO, "auth_cookie: ", auth_cookie)
 
             -- wait for token expiry as our mock idp issues tokens with 'expires_in: 1'
             ngx.sleep(2)


apisix/plugins/openid-connect.lua
```diff
@@ -84,6 +84,58 @@ local schema = {
...
+                    properties = {
+                        host = {
+                            type = "string", minLength = 2, default = "127.0.0.1"
+                        },
+                        port = {
+                            type = "integer", minimum = 1, default = 6379,
+                        },
+                        socket = {
+                            type = "string"
+                        },
```
nic-6443: Does this configuration mean connecting to redis via a linux domain socket? If so, I think it can be removed because it won't be used in actual situations.

apisix/plugins/openid-connect.lua
```diff
@@ -84,6 +84,58 @@ local schema = {
                             description = "it holds the cookie lifetime in seconds in the future",
                         }
                     }
+                },
+                storage = {
+                    type = "string",
+                    enum = {"cookie", "redis"},
+                    default = "cookie",
+                },
+                redis = {
```
nic-6443, line 93: It is necessary to add a dependency relationship between the `storage` and `redis` fields.

apisix/plugins/openid-connect.lua
```diff
@@ -84,6 +84,58 @@ local schema = {
...
+                        },
+                        password = {
+                            type = "string", minLength = 0,
+                        },
+                        database = {
+                            type = "integer", minimum = 0, default = 0,
+                        },
+                        prefix = {
+                            type = "string", default = "sessions"
+                        },
```
nic-6443, line 116: Some fields need to have a `description` field; otherwise, their meanings are hard to understand.

docs/en/latest/plugins/openid-connect.md
```diff
@@ -67,6 +67,17 @@ The `openid-connect` Plugin supports the integration with [OpenID Connect (OIDC)
...
+| session.redis        | object   | False    |     |             |   Redis configuration when `storage` is `redis`.    |
+| session.redis.host   | string   | False    | 127.0.0.1 |             |   Redis host.    |
+| session.redis.port   | integer   | False    | 6379 |             |   Redis port.    |
+| session.redis.password | string   | False    |     |             |   Redis password.    |
+| session.redis.username | string   | False    |     |             |   Redis username.    |
+| session.redis.database | integer   | False    | 0 |             |   Redis database index.    |
+| session.redis.prefix | string   | False    | sessions |             |   Redis key prefix.    |
+| session.redis.ssl    | boolean   | False    | false |             |   Enable SSL for Redis connection.    |
+| session.redis.server_name | string   | False    |     |             |   Redis server name for SNI.    |
+| session.redis.auth   | string   | False    |     |             |   Alias for `session.redis.password`.    |
```
nic-6443: Some fields are missing in the document, such as `*_timeout`

t/plugin/openid-connect-redis.t
```diff
@@ -0,0 +1,382 @@
...
+                return
+            end
+
+            local initial_cookie = get_cookie(res.headers, "session")
+            if not initial_cookie then
+                ngx.say("failed to get initial session cookie")
+                return
+            end
+
+            -- extract state from the Location URL egample: http://.../authorize?client_id=...&state=...&nonce=...
```
nic-6443:
```suggestion
            -- extract state from the Location URL example: http://.../authorize?client_id=...&state=...&nonce=...
```

t/plugin/openid-connect-redis.t
```diff
@@ -0,0 +1,382 @@
...
+                return
+            end
+
+            -- After callback, we get the FINAL authenticated session cookie.
+            local auth_cookie = get_cookie(res.headers, "session")
+            if not auth_cookie then
+                ngx.say("failed to get authenticated session cookie after callback. status: ", res.status)
+                return
+            end
+            ngx.log(ngx.INFO, "dibag auth_cookie: ", auth_cookie)
```
nic-6443:
```suggestion
            ngx.log(ngx.INFO, "auth_cookie: ", auth_cookie)
```

t/plugin/openid-connect-redis.t
```diff
@@ -0,0 +1,382 @@
...
+            )
+
+            if code >= 300 then
+                ngx.say("setup route 2 failed")
+                return
+            end
+
+            local httpc = http.new()
+
+            -- extract cookie value by name from Set-Cookie heade
```
nic-6443:
```suggestion
            -- extract cookie value by name from Set-Cookie header
```

apisix/plugins/openid-connect.lua
shreemaan-abhishek: fixed

apisix/plugins/openid-connect.lua
shreemaan-abhishek: fixed

apisix/plugins/openid-connect.lua
shreemaan-abhishek: done

membphis (approved):

apisix/plugins/openid-connect.lua
bzp2010, line 1: Essentially, this is just defining a new schema to expose features already supported by the library and add documentation?

shreemaan-abhishek: yes, the underlying lua-resty-session library is very feature rich and mature.

PR #12990, opened 2026-02-09
fix(opentelemetry): validate x-request-id before using it as trace_id

### LINKED ISSUE #12976: bug: opentelemetry trace_id_source=x-request-id crashes with request-id plugin
### Current Behavior

When `opentelemetry` metadata `trace_id_source` is set to `x-request-id` and the `request-id` plugin is enabled (default UUID), requests with `X-Request-Id` present trigger a Lua runtime error. The request fails and logs show a `bad argument` error.

Related issues that might be similar:
- #11969
- #10389

### Expected Behavior

The plugin should validate that `x-request-id` is a 32-hex trace_id (16 bytes). If invalid (UUID with hyphens, ksuid/nanoid/range_id, empty string, etc.), it should fall back to random trace_id or ignore the header. It should never crash.

We can also consider supporting the generation of a valid trace ID in the `request-id` plugin.

### Error Logs

2025-02-14 14:10:24 2025/02/14 14:10:24 [error] 53#53: *537 lua entry thread aborted: runtime error: bad argument #1 to '?' (invalid value)
2025-02-14 14:10:24 stack traceback:
2025-02-14 14:10:24 coroutine 0:
2025-02-14 14:10:24 [C]: in function 'gsub'
2025-02-14 14:10:24 ...s/share/lua/5.1/opentelemetry/trace/exporter/encoder.lua:22: in function 'hex2bytes'
2025-02-14 14:10:24 ...s/share/lua/5.1/opentelemetry/trace/exporter/encoder.lua:66: in function 'for_otlp'
2025-02-14 14:10:24 ...deps/share/lua/5.1/opentelemetry/trace/exporter/otlp.lua:102: in function 'export_spans'
2025-02-14 14:10:24 ...are/lua/5.1/opentelemetry/trace/batch_span_processor.lua:45: in function 'process_batches'
2025-02-14 14:10:24 ...are/lua/5.1/opentelemetry/trace/batch_span_processor.lua:77: in function

### Steps to Reproduce

1. Enable the `request-id` plugin (default header `X-Request-Id`, algorithm `uuid`).
2. Set `opentelemetry` plugin metadata `trace_id_source` to `x-request-id`.
3. Send a request
4. Observe the runtime error and failed request.

### Environment

prasunsrivastav123-lang:
### Description

This PR fixes a crash in the OpenTelemetry plugin when trace_id_source is set to x-request-id and the request ID is not a valid 32-hex trace ID (for example, a UUID).
The plugin now validates the header and gracefully falls back to the default trace ID generator instead of passing invalid values to the OpenTelemetry encoder.

Fixes #12976 
Fixes #12976 

### Checklist


prasunsrivastav123-lang:
@indrekj @markokocic @huacnlee pls review this pr

Baoyuantop:
Hi @prasunsrivastav123-lang, please add a test case for this fix.

prasunsrivastav123-lang:
hey @Baoyuantop i added a test case pls review this pr thank you !

Baoyuantop:
Hi @prasunsrivastav123-lang, please fix failed CI

prasunsrivastav123-lang:
Hi @Baoyuantop  I’ve added the requested test case to cover the invalid X-Request-Id scenario and ensured the route is properly recreated to avoid upstream issues.if any adjustment need pls tell .

prasunsrivastav123-lang:
@Baoyuantop Addressed the review feedback:
Added validation for all-zero trace IDs
Prevented multiple id_generator overrides
Added a test to verify valid fallback trace generation
Kindly requesting a re-review. Thanks!

prasunsrivastav123-lang:
@Baoyuantop I’ve addressed the review comments and updated the tests accordingly.
Could you please take another look?

apisix/plugins/opentelemetry.lua
```diff
@@ -29,6 +29,8 @@ local exporter_client_new = require("opentelemetry.trace.exporter.http_client").
 local otlp_exporter_new = require("opentelemetry.trace.exporter.otlp").new
 local batch_span_processor_new = require("opentelemetry.trace.batch_span_processor").new
 local id_generator = require("opentelemetry.trace.id_generator")
+local original_new_ids = id_generator.new_ids
+local id_generator_overridden = false
```
Baoyuantop: The `id_generator_overridden` flag, once set to true, is never reset. If an administrator changes `trace_id_source` back to `random` from `x-request-id`, all worker processes will continue to attempt to extract trace IDs from `x-request-id` because `id_generator.new_ids` has been replaced with a custom function and will never be restored. This could lead to a silent bug where configuration changes don't take effect (although this issue existed in the original code, the new implementation has solidified it through a global flag).

The global flag is redundant. `original_new_ids` already captures the original function reference during module loading; multiple executions of `id_generator.new_ids = function()...end` simply replace the function reference and do not introduce nested wrapping issues. Introducing the global flag solves a problem that doesn't actually exist, but introduces a real bug.

In fact, the correct and simplest fix is ​​to remove `id_generator_overridden` and the module-level `original_new_ids`, and directly capture and verify the original function reference within the existing `if` block:

```
if plugin_info.trace_id_source == "x-request-id" then
    local _original_new_ids = id_generator.new_ids
    id_generator.new_ids = function()
        local trace_id = core.request.headers()["x-request-id"] or ngx_var.request_id
        if is_valid_trace_id(trace_id) then
            return trace_id, id_generator.new_span_id()
        end
        return _original_new_ids()
    end
end
```

apisix/plugins/opentelemetry.lua
```diff
@@ -230,12 +250,23 @@ end
 
 
 local function create_tracer_obj(conf, plugin_info)
-    if plugin_info.trace_id_source == "x-request-id" then
-        id_generator.new_ids = function()
-            local trace_id = core.request.headers()["x-request-id"] or ngx_var.request_id
-            return trace_id, id_generator.new_span_id()
+   if plugin_info.trace_id_source == "x-request-id"
```
Baoyuantop: Please fix the code format

prasunsrivastav123-lang:
@Baoyuantop  i just Removed the global flag and simplified the id_generator override as suggested. 
Could you please recheck? Thanks!

prasunsrivastav123-lang:
@Baoyuantop can you review this

Baoyuantop (dismissed):

Baoyuantop:
Hi @prasunsrivastav123-lang, please fix failed CI

prasunsrivastav123-lang:
Hi @Baoyuantop , I’ve fixed the CI issue by reindexing the tests and pushed the update. Could you please take another look? Thanks!

moonming (changes requested):
Hi @prasunsrivastav123-lang, thank you for fixing the x-request-id validation in OpenTelemetry (#12976)!

Validating the x-request-id format before using it as a trace_id is the right approach — invalid formats could cause issues with tracing backends.

The implementation with a fallback to the default generator looks clean. Could you confirm that the tests cover:
1. Valid 32-character hex trace IDs
2. Invalid formats (too short, non-hex characters, empty string)
3. Missing x-request-id header (should use default generator)

If all tests pass, this should be ready. Thank you!

prasunsrivastav123-lang:
@moonming @Baoyuantop Thanks! Yes, I’ve added tests covering valid 32-character hex IDs, multiple invalid formats (UUID, all-zero, malformed), and the missing header fallback case. All tests are passing now. Please take another look.

Baoyuantop (dismissed):

prasunsrivastav123-lang:
@moonming if there need any  other changes?

apisix/plugins/opentelemetry.lua
```diff
@@ -231,9 +249,17 @@ end
...
-            local trace_id = core.request.headers()["x-request-id"] or ngx_var.request_id
-            return trace_id, id_generator.new_span_id()
+            local trace_id = core.request.headers()["x-request-id"]
+                            or ngx_var.request_id
+
+            if is_valid_trace_id(trace_id) then
+                return trace_id, id_generator.new_span_id()
+            end
+
+            return _original_new_ids()
```
AlinsRan: The lrucache has a 24-hour TTL After expiry, create_tracer_obj is called again. At that point `id_generator.new_ids` is already the closure from the previous call, so `_original_new_ids` captures the already-overridden function. Each 24h expiry adds another wrapper layer.

The pre-PR code was idempotent (it overwrote the same closure). This PR makes it strictly worse.

prasunsrivastav123-lang: You're right — the current implementation wraps `id_generator.new_ids` multiple times when the tracer is recreated, which breaks idempotency and can lead to stacked wrappers over time.

My intention was to validate the x-request-id and fallback safely, but I see that the current approach needs adjustment to avoid repeated wrapping.

I’ll update the implementation to ensure the override happens only once (idempotent behavior) and push a fix shortly

Baoyuantop:
Hi @prasunsrivastav123-lang, there's a conflict in a code file. Could you help me resolve it?

prasunsrivastav123-lang:
sure 😃

prasunsrivastav123-lang:
@Baoyuantop is it ok now ?

apisix/plugins/opentelemetry.lua
Baoyuantop: Hi @prasunsrivastav123-lang, has this comment been resolved?

Baoyuantop:
Hi @prasunsrivastav123-lang, please fix the failed CI

prasunsrivastav123-lang:
@Baoyuantop Updated based on feedback. Added full validation and test coverage.

Baoyuantop:
Hi @prasunsrivastav123-lang, there were some formatting issues when the code was merged; I’ve corrected them.

prasunsrivastav123-lang:
@Baoyuantop Thanks for fixing the formatting!

I’ve updated the implementation to ensure trace_id validation is handled cleanly:
- Only valid 32-character hex trace IDs are used
- Invalid formats (non-hex, incorrect length, all-zero) fall back to the default generator
- Missing x-request-id also correctly falls back

Test cases have been added to cover all these scenarios.

Please let me know if any further changes are needed.

apisix/plugins/opentelemetry.lua
```diff
@@ -232,9 +253,24 @@ end
...
-            local trace_id = core.request.headers()["x-request-id"] or ngx_var.request_id
-            return trace_id, id_generator.new_span_id()
+        if not id_generator._wrapped then
+            local _original_new_ids = id_generator.new_ids
+
+            id_generator.new_ids = function()
+                local trace_id = core.request.headers()["x-request-id"]
+                                or ngx_var.request_id
+
+                trace_id = trace_id and string_lower(trace_id)
```
nic-6443: A duplicated `X-Request-Id` header still crashes this: `core.request.headers()` is `ngx.req.get_headers()`, which returns a Lua table when the same header appears more than once, so `string_lower(trace_id)` throws `bad argument`. And since plugin phase handlers are not pcall-wrapped, that error now fails the request with a 500 on every such request (before this patch the same input crashed asynchronously in the export timer). Given the whole point of this fix is to validate untrusted header input, it should survive this case too — a `type(trace_id) ~= "string"` check (e.g. at the top of `is_valid_trace_id`, before any `string_lower`) would cover it.

t/plugin/opentelemetry.t
```diff
@@ -488,3 +489,143 @@ opentracing
...
+=== TEST 26: all-zero x-request-id should not be used as trace id
+--- request
+GET /opentracing
+--- more_headers
+X-Request-Id: 00000000000000000000000000000000
+--- wait: 2
+--- exec
+tail -n 1 ci/pod/otelcol-contrib/data-otlp.json
+--- response_body eval
+qr/"traceId"\s*:\s*"[0-9a-f]{32}"/i
```
nic-6443: This assertion can't fail: the all-zero id itself matches `[0-9a-f]{32}`, so the test passes even if the validation it's supposed to verify is removed. It also matches whatever span an *earlier* test exported, since `tail -n 1` on the collector file is shared state across blocks. Something like `qr/"traceId"\s*:\s*"(?!0{32})[0-9a-f]{32}"/` would at least pin down the rejection here; the same vacuous-match concern applies to TESTs 25 and 28-31.

Unrelated nit while in this file: the first hunk is a whitespace-only change before TEST 20 (4 blank lines between blocks instead of 3) — please drop it.

t/plugin/opentelemetry.t
```diff
@@ -488,3 +489,143 @@ opentracing
...
+
+
+
+=== TEST 27: uppercase x-request-id should still generate a valid trace id
+--- request
+GET /opentracing
+--- more_headers
+X-Request-Id: 550E8400-E29B-41D4-A716-446655440000
```
nic-6443: This value still contains hyphens, so it's rejected for the same reason as TEST 24/25 — the case-normalization branch (invalid as-is, valid after `string_lower`) is never exercised. To test what the title says, send a 32-char uppercase hex value like `550E8400E29B41D4A716446655440000` and assert the exported traceId is exactly the lowercased form.

Relatedly, none of the new tests cover the happy path: a valid 32-hex header being used verbatim as the traceId (TEST 16 sends one but only greps the URL). One test that sends a fixed valid value and greps for that exact traceId would pin down both the core feature and the new lowercasing behavior.

apisix/plugins/opentelemetry.lua
AlinsRan: Good catch — fixed. `new_ids` now only uses the header when it is a plain string (`type(trace_id) == "string"`), so a duplicated `X-Request-Id` (which `get_headers()` returns as a table) falls through to the default generator instead of reaching `string_lower`. Added a duplicated-header regression test.

t/plugin/opentelemetry.t
AlinsRan: Right, that assertion could not fail. Split the request and exec into separate blocks, and the all-zero case now asserts `(?!0{32})[0-9a-f]{32}`, so it fails if the zero id is ever reused.

t/plugin/opentelemetry.t
AlinsRan: Fixed — the uppercase test now sends a real 32-char hex (`550E8400E29B41D4A716446655440000`) and asserts the exact lowercased trace id, so the case-normalization branch is actually exercised.

apisix/plugins/opentelemetry.lua
AlinsRan: Addressed: the default generator is now captured once at module load (`original_new_ids`) instead of inside the closure, so re-creating the tracer no longer stacks wrappers.

nic-6443 (dismissed):

membphis (dismissed):
LGTM

apisix/plugins/opentelemetry.lua
```diff
@@ -230,12 +250,23 @@ end
...
+
+        if is_valid_trace_id(header_trace_id) then
+            return header_trace_id, id_generator.new_span_id()
         end
+
+        -- fallback to default generator for invalid values (e.g. UUID)
+        return original_new_ids()
     end
+    id_generator_overridden = true
+end
```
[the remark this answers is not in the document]
AlinsRan: Fixed. `trace_id_source` lives in plugin metadata (global), so there is no per-route mode to preserve — `create_tracer_obj` now resets `id_generator.new_ids` to the captured default first and only wraps it for the `x-request-id` mode. The lrucache is keyed on `metadata.modifiedIndex`, so switching back to `random` rebuilds the tracer and takes effect without a worker restart. Added a regression test that a valid `X-Request-Id` is ignored once `trace_id_source` is `random`.

membphis:
P2 follow-up: the current commit history still contains AI co-author trailers such as `Co-authored-by: Copilot ...`. Before merge, please clean those trailers from the final commit history (for example via squash/rewrite), so the project history does not carry AI attribution markers.

membphis (approved):
LGTM

shreemaan-abhishek (approved):

nic-6443 (approved):

moonming (approved):

PR #12998, opened 2026-02-10
fix(plugin): merge consumer group plugins when consumer has no direct plugins

### LINKED ISSUE #12912: bug:  consumer group rate limiting not working when configuring consumer credentials using /credentials endpoint
### Current Behavior

Consumer group rate limiting rules are not being applied when consumer credentials are configured using the `/consumers/{username}/credentials` endpoint. However, rate limiting roles work correctly when authentication plugins are configured directly on the consumer.

### Expected Behavior

Consumer group rate limiting rules should work correctly when consumer credentials are configured using the `/consumers/{username}/credentials` endpoint. 

#### 1. Use` /credentials` endpoint (rate limiting not working)

Create consumer groups:

```shell
curl "http://127.0.0.1:9180/apisix/admin/consumer_groups" -X PUT -d '{
  "id": "basic_plan",
  "plugins": {
    "limit-count": {
      "count": 2,
      "time_window": 30,
      "rejected_code": 429
    }
  }
}'

curl "http://127.0.0.1:9180/apisix/admin/consumer_groups" -X PUT -d '{
  "id": "premium_plan", 
  "plugins": {
    "limit-count": {
      "count": 20,
      "time_window": 30,
      "rejected_code": 429
    }
  }
}'
```

Create consumers:

```shell
curl "http://127.0.0.1:9180/apisix/admin/consumers" -X PUT -d '{
  "username": "JohnDoe",
  "group_id": "basic_plan"
}'

curl "http://127.0.0.1:9180/apisix/admin/consumers" -X PUT -d '{
  "username": "FetchBot", 
  "group_id": "premium_plan"
}'
```

Configure credentials:

```shell
curl "http://127.0.0.1:9180/apisix/admin/consumers/JohnDoe/credentials" -X PUT -d '{
  "id": "cred-john-key-auth",
  "plugins": {
    "key-auth": {
      "key": "john-key"
    }
  }
}'

curl "http://127.0.0.1:9180/apisix/admin/consumers/FetchBot/credentials" -X PUT -d '{
  "id": "cred-bot-key-auth", 
  "plugins": {
    "key-auth": {
      "key": "bot-key"
    }
  }
}'
```

Create a route:

```shell
curl "http://127.0.0.1:9180/apisix/admin/routes" -X PUT -d '{
  "id": "rate-limiting",
  "uri": "/internal",
  "plugins": {
    "key-auth": {}
  },
  "upstream": {
    "nodes": {
      "mock.api7.ai:443": 1
    },
    "pass_host": "node",
    "scheme": "https"
  }
}'
```

Test rate limiting:

```shell
resp=$(seq 10 | xargs -I{} curl "http://127.0.0.1:9080/internal" -H 'apikey: john-key' -o /dev/null -s -w "%{http_code}\n") && \
  count_200=$(echo "$resp" | grep "200" | wc -l) && \
  count_429=$(echo "$resp" | grep "429" | wc -l) && \
  echo "200": $count_200, "429": $count_429

# Result: 200: 10, 429: 0
# SHOULD BE: 200: 2, 429: 8
```

#### 2. Configure auth credentials directly on the consumers (rate limiting works)

Create the same consumer groups as above.

Create consumers with auth plugins configured directly:

```shell
curl "http://127.0.0.1:9180/apisix/admin/consumers" -X PUT -d '{
  "username": "JohnDoe",
  "group_id": "basic_plan",
  "plugins": {
    "key-auth": {
      "key": "john-key"
    }
  }
}'

curl "http://127.0.0.1:9180/apisix/admin/consumers" -X PUT -d '{
  "username": "FetchBot",
  "group_id": "premium_plan", 
  "plugins": {
    "key-auth": {
      "key": "bot-key"
    }
  }
}'
```

Test rate limiting:

```shell
resp=$(seq 10 | xargs -I{} curl "http://127.0.0.1:9080/internal" -H 'apikey: john-key' -o /dev/null -s -w "%{http_code}\n") && \
  count_200=$(echo "$resp" | grep "200" | wc -l) && \
  count_429=$(echo "$resp" | grep "429" | wc -l) && \
  echo "200": $count_200, "429": $count_429

# Result: 200: 2, 429: 8 (CORRECT)
```

### Environment

APISIX version: 3.14.1

janiussyafiq:
### Description

This PR fixes a bug where consumer group plugins are not applied when consumers are configured using the credentials endpoint. The issue occurs because the `merge_consumer_route()` function in `plugin.lua` returns early when a consumer has no direct plugins, without checking if the consumer belongs to a consumer group that has plugins.

When using the credentials endpoint (`/apisix/admin/consumers/{consumer}/credentials/{credential_id}`), the credential object stores the authentication plugin (e.g., `key-auth`), while the parent consumer may have no direct plugins but references a consumer group via `group_id`. The bug is in `apisix/plugin.lua` lines 706-712: 
https://github.com/apache/apisix/blob/184678ccc1b3ca97bb0a9c5df2dfc92286d04514/apisix/plugin.lua#L708-L714

This early return prevents consumer group plugins from being merged into the request context.

Solution

The fix modifies merge_consumer_route() to:
1. Check both consumer_conf.plugins and consumer_group_conf for plugins before early return
2. Merge consumer group plugins first, then consumer plugins (maintaining correct precedence)
3. Only return early if neither consumer nor consumer group has plugins

This ensures consumer group plugins are applied regardless of whether the consumer has direct plugins.

Fixes #12912 

### Checklist


t/plugin/limit-count-consumer-group-credentials.t
```diff
@@ -0,0 +1,555 @@
...
+            end
+            ngx.say(body)
+        }
+    }
+--- response_body
+passed
+
+
+
+=== TEST 5: verify rate limiting works (all 3 requests in one test)
```
shreemaan-abhishek, line 152: check `request_eval` in limit count test files, using that you can test this same behaviour in a much more simpler way.

janiussyafiq: will do

t/plugin/limit-count-consumer-group-credentials.t
```diff
@@ -0,0 +1,399 @@
...
+    "GET /hello"
+]
+--- more_headers
+apikey: auth-override
+--- error_code eval
+[200, 200, 200, 200, 200, 429]
+
+
+
+=== TEST 14: cleanup
```
shreemaan-abhishek: get rid of this test

shreemaan-abhishek (commented):
go through all the test cases and check if we actually need all of those.

Baoyuantop (approved):

PR #13004, opened 2026-02-11
feat: support header prefix in limit-count rules

shreemaan-abhishek:
### Description

We need a way to distinguish the rate limiting status with different headers when multiple rules are hit simultaneously. We can do this by allowing users to configure a header_prefix, the default value for this would be the index of the rule being executed.

### Checklist


docs/en/latest/plugins/limit-count.md
```diff
@@ -49,7 +49,8 @@ You may see the following rate limiting headers in the response:
...
+| rules.header_prefix     | string  | False                                     |               |                            | Prefix for rate limit headers. If configured, the response will include `X-{header_prefix}-RateLimit-Limit`, `X-{header_prefix}-RateLimit-Remaining`, and `X-{header_prefix}-RateLimit-Reset` headers. If not configured, the rule index (1-based) will be used as the prefix.                                                                                                                                                                                                                                                                   |
```
kayx23:
```suggestion
| rules.header_prefix     | string  | False                                     |               |                            | Prefix for rate limit headers. If configured, the response will include `X-{header_prefix}-RateLimit-Limit`, `X-{header_prefix}-RateLimit-Remaining`, and `X-{header_prefix}-RateLimit-Reset` headers. If not configured, the index of the rule in the rules array is used as the prefix. For example, headers for the first rule will be `X-1-RateLimit-Limit`, `X-1-RateLimit-Remaining`, and `X-1-RateLimit-Reset`.                                                                                                                                                                                                                                                                  |
```

nic-6443 (approved):

Baoyuantop (approved):

PR #13019, opened 2026-02-18
fix(limit-req): use parent resource key for consumer isolation

### LINKED ISSUE #12946: bug: limit-req: conf_type/conf_version suffix makes per-consumer rate limits effectively per-route
### Current Behavior

When the limit-req plugin is configured only at the Consumer level, the effective rate limit is still scoped per route, not per consumer.

Internally, limit-req always appends ctx.conf_type and ctx.conf_version to the generated key. At runtime, ctx.conf_type includes both route and consumer (e.g. route&consumer), and conf_version includes the route config ID. As a result, a separate Redis counter is created for each route, even though the plugin is not configured on the routes themselves.

This can be observed in logs such as:

```
conf type: route&consumer
conf version: <route_version_id>&<consumer_version>
limit key: <base_key>route&consumer<conf_version>
```

Because of this behavior, a consumer configured with N req/sec can effectively make N req/sec per route, which bypasses the expected global per-consumer rate limit.

There is currently no configuration option in limit-req to prevent the conf_type/conf_version suffix from being added, making it impossible to implement a true global per-consumer leaky-bucket rate limit using limit-req.

### Expected Behavior

When the limit-req plugin is configured at the Consumer level, the rate limit should apply globally to that consumer, regardless of which route is being accessed.

In this scenario, all requests made by the same consumer should share a single rate-limit bucket, so that a consumer configured with N req/sec can only make N req/sec in total, even when calling multiple routes.

More generally, users should be able to control the scope of the rate-limit key (for example, per route vs. per consumer), or at least have a way to prevent route-specific information from being implicitly included in the rate-limit key when the plugin is defined on a consumer.

This would allow limit-req to be used for true global per-consumer rate limiting, without requiring custom plugin overrides or switching to a different rate-limiting plugin.

### Error Logs

The following logs were captured by adding debug prints in limit-req.lua while sending requests from the same consumer to different routes.

```
  key = key .. ctx.conf_type .. ctx.conf_version
  core.log.error("limit key: ", key)
  core.log.error("conf type: ", ctx.conf_type)
  core.log.error("conf version: ", ctx.conf_version)
```

Even though limit-req is configured only on the Consumer, the plugin reports a combined configuration context (route&consumer), and the generated key includes route-specific information:

```
2026/01/27 13:11:59 [error] 204#204: *53475 [lua] limit-req.lua:162: phase_func(): conf version: 2008&3, client: 192.168.65.1, server: _, request: "GET /test HTTP/1.1", host: "localhost:9080"

2026/01/27 13:12:00 [error] 194#194: *53481 [lua] limit-req.lua:160: phase_func(): limit key: 192.168.65.1route&consumer2008&3, client: 192.168.65.1, server: _, request: "GET /test HTTP/1.1", host: "localhost:9080"

2026/01/27 13:12:00 [error] 194#194: *53481 [lua] limit-req.lua:161: phase_func(): conf type: route&consumer, client: 192.168.65.1, server: _, request: "GET /test HTTP/1.1", host: "localhost:9080"
```

Redis keys:

```
127.0.0.1:6379> keys *
1) "limit_req:192.168.65.1route&consumer2008&3excess"
2) "limit_req:192.168.65.1route&consumer2031&3last"
3) "limit_req:192.168.65.1route&consumer2031&3excess"
4) "limit_req:192.168.65.1route&consumer2008&3last"
```

These logs show that the rate-limit key is effectively namespaced by both the route and the consumer, resulting in separate rate-limit counters per route.

### Steps to Reproduce

1. Start APISIX 3.12 and Redis

Run APISIX 3.12 with Redis enabled (local or via Docker).
Ensure Redis is reachable from APISIX.

2. Create a Consumer with key-auth and limit-req

Configure limit-req only at the Consumer level:

```
curl -X PUT http://127.0.0.1:9180/apisix/admin/consumers/test-consumer \
  -H 'X-API-KEY: <ADMIN_API_KEY>' \
  -d '{
    "username": "test-consumer",
    "plugins": {
      "key-auth": {
        "key": "test-api-key"
      },
      "limit-req": {
        "rate": 2,
        "burst": 1,
        "key": "remote_addr",
        "key_type": "var",
        "policy": "redis",
        "redis_host": "redis",
        "redis_port": 6379,
        "redis_database": 0
      }
    }
  }'
```

3. Create two routes using key-auth

Route A:

```
curl -X PUT http://127.0.0.1:9180/apisix/admin/routes/1 \
  -H 'X-API-KEY: <ADMIN_API_KEY>' \
  -d '{
    "uri": "/route-a",
    "plugins": {
      "key-auth": {}
    },
    "upstream": {
      "type": "roundrobin",
      "nodes": {
        "httpbin.org:80": 1
      }
    }
  }'
```

Route B:

```
curl -X PUT http://127.0.0.1:9180/apisix/admin/routes/2 \
  -H 'X-API-KEY: <ADMIN_API_KEY>' \
  -d '{
    "uri": "/route-b",
    "plugins": {
      "key-auth": {}
    },
    "upstream": {
      "type": "roundrobin",
      "nodes": {
        "httpbin.org:80": 1
      }
    }
  }'
```

4. Send requests to both routes using the same API key

```
curl http://127.0.0.1:9080/route-a -H "apikey: test-api-key"
curl http://127.0.0.1:9080/route-b -H "apikey: test-api-key"
```

Repeat the requests rapidly (more than 2 requests per second in total).

5. Observe the behavior

Requests to /route-a and /route-b are rate-limited independently

The consumer can exceed the configured rate by distributing requests across routes

Redis shows separate rate-limit keys per route, even though limit-req is not configured on the routes

### Environment

- APISIX version (run `apisix version`): APISIX version: 3.12.0
- Operating system (run `uname -a`): Linux (Docker-based setup)

janiussyafiq:
### Description

Previously, when limit-req was configured at the consumer level, rate limits were applied per route instead of globally per consumer. This occurred because the key generation concatenated `ctx.conf_type` and `ctx.conf_version`, which included route-specific information. Now consumer-level rate limits are properly shared across all routes accessed by the same consumer by using the `conf._meta.parent` approach.

Changes:
- Used conf._meta.parent.resource_key for key generation
- Added consumer isolation test

Fixes #12946 

### Checklist

### DIFF
--- a/apisix/plugins/limit-req.lua
+++ b/apisix/plugins/limit-req.lua
@@ -21,6 +21,7 @@ local policy_to_additional_properties   = redis_schema.schema
 local plugin_name                       = "limit-req"
 local sleep                             = core.sleep
 local apisix_plugin                     = require("apisix.plugin")
+local error                             = error
 
 local redis_single_new
 local redis_cluster_new


apisix/plugins/limit-req.lua
```diff
@@ -122,6 +123,23 @@ local function create_limit_obj(conf)
...
 
 
+local function gen_limit_key(conf, ctx, key)
+    local parent = conf._meta and conf._meta.parent
+    if not parent or not parent.resource_key then
+        core.log.error("failed to generate key invalid parent: ", core.json.encode(parent))
+        return nil
+    end
+
+    local new_key = parent.resource_key .. ':' .. apisix_plugin.conf_version(conf)
```
shreemaan-abhishek: previous logic use `ctx.conf_version` instead of `apisix_plugin.conf_version`, the value of the former stays the same even if the consumer is recreated, however the value of conf_version changes in the latter method. This is the reason why limit-req test is failing as it expects 403 in the first request itself.

apisix/plugins/limit-req.lua
```diff
@@ -122,6 +123,23 @@ local function create_limit_obj(conf)
...
+                    .. ':' .. key
+    if conf._vid then
```
shreemaan-abhishek: Do we need this? I think limit-req is not supported by workflow plugin.

janiussyafiq: will remove

apisix/plugins/limit-req.lua
janiussyafiq:
```
2026/02/24 01:10:38 [error] 59465#59465: *1486928 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6619&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "1dd149afeab7a65c85aea2fee2d7799e"
2026/02/24 01:10:39 [error] 59472#59472: *1486944 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6618&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /headers HTTP/1.1", host: "127.0.0.1:9080", request_id: "8cace6edff2ebd5229e9f3484f605eb1"
2026/02/24 01:10:40 [error] 59465#59465: *1487019 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6619&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "3eee5322201e0cce7097ed51577b938f"
2026/02/24 01:10:40 [error] 59479#59479: *1487026 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6618&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /headers HTTP/1.1", host: "127.0.0.1:9080", request_id: "20afdffb38c3fbe13ac7ed1c809c8888"
2026/02/24 01:11:35 [error] 59478#59478: *1491539 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6619&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "34285dd4f1ba2aef6eec46b78ac5855f"
2026/02/24 01:11:36 [error] 59465#59465: *1491610 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6618&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /headers HTTP/1.1", host: "127.0.0.1:9080", request_id: "69fc39676a1bcb7c666ffb2eddd3a7ce"
2026/02/24 01:11:36 [error] 59470#59470: *1491623 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6619&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "71c611319270e13f3fa76e69a2eddb8f"
2026/02/24 01:11:37 [error] 59466#59466: *1491692 [lua] limit-req.lua:180: phase_func(): limit key: /apisix/consumers/jack:6618&34:127.0.0.1, client: 127.0.0.1, server: _, request: "GET /headers HTTP/1.1", host: "127.0.0.1:9080", request_id: "147ddf0f672c9c67798e6d2bd6952aa1"
```
this is the result when we used the `ctx.conf_version`, we can see that for those two routes, the keys differ hence failing the `limit-req-shared-counter.t` test that I made.

t/plugin/limit-req.t
```diff
@@ -452,7 +452,7 @@ passed
 --- more_headers
 apikey: auth-jack
 --- error_code eval
-[403, 403, 403, 403]
+[200, 403, 403, 403]
```
Baoyuantop, line 455: It is necessary to explain why this change is needed.

janiussyafiq: https://github.com/openresty/lua-resty-limit-traffic/blob/2a834b193ccf1cd94af6c7f2856eb2dce9fc5a39/lib/resty/limit/req.lua#L105
Since the first request key is always new, hence the request will fall onto the else condition where `excess = 0` hence making the first request always passing. This condition is verified by adding a test case manually into the repo within this file (https://github.com/openresty/lua-resty-limit-traffic/blob/master/t/req.t).

```
=== TEST 8: very low rate and burst (0.1) - first request always accepted
--- http_config eval
"
$::HttpConfig
    lua_shared_dict store 1m;
"
--- config
    location = /t {
        content_by_lua '
            local limit_req = require "resty.limit.req"
            ngx.shared.store:flush_all()
            local lim = limit_req.new("store", 0.1, 0.1)
            local key = "test_key"
            -- First request should always be accepted with delay=0
            local delay1, err1 = lim:incoming(key, true)
            if not delay1 then
                ngx.say("1st request failed: ", err1)
            else
                ngx.say("1st request delay: ", delay1)
            end
            -- Second request
            local delay2, err2 = lim:incoming(key, true)
            if not delay2 then
                ngx.say("2nd request failed: ", err2)
            else
                ngx.say("2nd request delay: ", delay2)
            end
            -- Third request
            local delay3, err3 = lim:incoming(key, true)
            if not delay3 then
                ngx.say("3rd request failed: ", err3)
            else
                ngx.say("3rd request delay: ", delay3)
            end
        ';
    }
--- request
GET /t
--- response_body
1st request delay: 0
2nd request failed: rejected
3rd request failed: rejected
--- no_error_log
[error]
[lua]
```

Response I get is:
```
root@bf0f4dba9671:/workspace# FLUSH_ETCD=1 prove -I../test-nginx/lib -I./ -r -s lua-resty-limit-traffic/t/req.t
lua-resty-limit-traffic/t/req.t .. ok     
All tests successful.
Files=1, Tests=64,  7 wallclock secs ( 0.01 usr  0.01 sys +  0.12 cusr  0.09 csys =  0.23 CPU)
Result: PASS
```
Hence confirming this behavior is expected.
This also reflected here:
https://github.com/apache/apisix/blob/c8248a69381c02c0f1e6743b65a80cc02f22c155/t/plugin/limit-req.t#L166-L170

nic-6443: @janiussyafiq I can understand your explanation, but why could the previous code pass this test case? The previous key included `ctx.conf_version`. When the consumer is updated, `ctx.conf_version` will change, so the key becomes a brand-new key. According to your explanation, the first request should also be allowed too.

apisix/plugins/limit-req.lua
```diff
@@ -122,6 +123,20 @@ local function create_limit_obj(conf)
 end
 
 
+local function gen_limit_key(conf, ctx, key)
+    local parent = conf._meta and conf._meta.parent
+    if not parent or not parent.resource_key then
+        core.log.error("failed to generate key invalid parent: ", core.json.encode(parent))
```
membphis: we can call `error`, throw a panic error
it should never happen

apisix/plugins/limit-req.lua
```diff
@@ -122,6 +123,20 @@ local function create_limit_obj(conf)
...
+        return nil
+    end
+
+    local new_key = parent.resource_key .. ':' .. apisix_plugin.conf_version(conf)
```
membphis:
```suggestion
    return parent.resource_key .. ':' .. apisix_plugin.conf_version(conf) .. ':' .. key
```

apisix/plugins/limit-req.lua
membphis: if you copied those code from plugin `limit-count`
welcome to submit a new PR, fix it too

https://github.com/apache/apisix/blob/afda19403df926556d19cee0f4f8f5f9dee95426/apisix/plugins/limit-count/init.lua#L245-L246

membphis (changes requested):

t/plugin/limit-req.t
janiussyafiq: That is indeed the case and I even double check it manually to confirm that the key gets updated after we updated the `limit-req` config,
```
2026/02/25 02:19:48 [error] 98621#98621: *3824971 [lua] limit-req.lua:161: phase_func(): limit key: jackroute&consumer6896&1, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "b7365540e7c2ea83e69ddbfc56d1dcb9"
2026/02/25 02:30:16 [error] 98622#98622: *3876200 [lua] limit-req.lua:161: phase_func(): limit key: jackroute&consumer6896&2, client: 127.0.0.1, server: _, request: "GET /get HTTP/1.1", host: "127.0.0.1:9080", request_id: "a01763218daf2f8e4911c208a2da8b8d"
```
but upon running the test, I can confirm that the key indeed persisted even after plugin update (see limit key):
```
#   Failed test 't/plugin/limit-req.t TEST 14: not exceeding the burst - pattern "[error]" should not match any line in error.log but matches line "2026/02/25 02:48:48 [error] 12944#12944: *1217 [lua] limit-req.lua:161: phase_func(): limit key: new_consumerroute&consumer7060&1, client: 127.0.0.1, server: localhost, request: \"GET /hello HTTP/1.1\", host: \"localhost\", request_id: \"55af5f6e801120afb962f643141ae8b1\"" (req 0)
#   Failed test 't/plugin/limit-req.t TEST 16: exceeding the burst - pattern "[error]" should not match any line in error.log but matches line "2026/02/25 02:48:51 [error] 13097#13097: *1405 [lua] limit-req.lua:161: phase_func(): limit key: new_consumerroute&consumer7060&1, client: 127.0.0.1, server: localhost, request: \"GET /hello HTTP/1.1\", host: \"localhost\", request_id: \"a6bb80f372f3f39eb8f792296c485198\"" (req 0)
```
this is an unexpected behavior and i am currently not sure what causing it. One of the function mentioned upon running this test is `merge_consumer_route`; maybe it originated from there (?).
i tried changing the `key` value and the test is still passing i.e. key persisted. however when i change the `username` within consumer config, the first request is now passing. I think this is important to take note on.

apisix/plugins/limit-req.lua
janiussyafiq: fixed

apisix/plugins/limit-req.lua
janiussyafiq: fixed

t/plugin/limit-req.t
nic-6443: Good job on the troubleshooting. It is recommended that you continue troubleshooting because this may mean there are other bugs hidden here.

shreemaan-abhishek (approved):

PR #13022, opened 2026-02-19
fix(schema): correct minLength type for anonymous consumer

vijayalakshmipokala:
The anonymous_consumer_schema defined minLength as a string instead of a number.
This patch fixes the schema type to ensure correct validation behavior.

### DIFF
--- a/apisix/schema_def.lua
+++ b/apisix/schema_def.lua
@@ -26,7 +26,7 @@ local plugins_schema = {
 
 _M.anonymous_consumer_schema = {
     type = "string",
-    minLength = "1"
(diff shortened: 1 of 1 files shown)


vijayalakshmipokala:
This patch fixes the incorrect use of a string for minLength in anonymous consumer schema.
minLength should be a number to match JSON schema expectations and correct validation behavior.

vijayalakshmipokala:
This patch fixes the incorrect use of a string for minLength in anonymous consumer schema.
minLength should be a number to match JSON schema expectations and correct validation behavior.

PR #12951, opened 2026-01-29
feat: allow to use secrets in clickhouse-logger plugin

vlad-vinogradov-47:
### Description

This will allow to store credential information like user/password in environment variables or HashiCorp Vault.

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #13025, opened 2026-02-24
chore(elasticsearch-logger): allow to use $env://MY_ENV_VAR in auth.password

### LINKED ISSUE #11366: feat: support env var for secret config of plugins like kafka-logger, elasticsearch-logger
### Description

I am using helm to deploy a set of APISIX, Dashboard and etcd.
I defined the extraEnvVars in the values.yaml and set K8s secret, and able to print the ENV VAR key/value in the apisix-gateway pod.

However, when I set the plugin config json following this guide, it is not able to use the ENV VAR value.
https://apisix.apache.org/docs/apisix/terminology/secret/#use-environment-variables-to-manage-secrets

Sample of simplied config as below:
```javascript
{
    "plugins": {
        "kafka-logger": {
            "brokers": [
                {
                    "sasl_config": {
                        "mechanism": "PLAIN",
                        "password": "$ENV://KAFKA_LOGGER_PW",
                        "user": "user"
                    }
                }
            ]
        },
        "elasticsearch-logger": {
            "auth": {
                "password": "$ENV://ELASTICSEARCH_LOGGER_PW",
                "username": "user"
            }
        }
    }
}
``` 

I also tried with the mocking plugin but with no luck, is the ENV VAR mechanism not working on every plugin?

### Environment
Kubernetes Version: v1.27.12 +rke2r1
APISIX Helm Charts Version: 2.8.0
APISIX version: 3.9.1
Operating system: Linux apisix-797b96d476-vtkg2 5.14.0-362.8.1.el9_3.x86_64
OpenResty / Nginx version: openresty/1.25.3.1
etcd version: 3.5.10

jr-instantsystem:
### Description

Allow to reference elasticsearch password from environment variable, instead of storing in clear in configuration source files.
Tested & already used.

(Sorry guys, my first PR on apisix, not sure it's very clean...)

### Checklist


Baoyuantop:
Hi @jr-instantsystem, please add a test case for this PR

Baoyuantop:
Hi @jr-instantsystem, following up on the previous review comments. Please let us know if you have any updates. Thank you.

jr-instantsystem:
Sorry, I don't have anything to launch or debug "test cases".
We have implemented this fix from DEV to Production directly in apisix, everything is ok on our side.
Maybe someone can add the test, should be quick, I just can't do it...

Baoyuantop (commented):
Hi @jr-instantsystem, thanks for the contribution! Here's a test case you can add to `t/plugin/elasticsearch-logger.t` (append after the existing TEST 26):

```perl
=== TEST 27: set route (auth.password from env variable via $env://)
--- main_config
env ES_PASSWORD=123456;
--- config
    location /t {
        content_by_lua_block {
            local code, body = t('/apisix/admin/routes/1', ngx.HTTP_PUT, {
                uri = "/hello",
                upstream = {
                    type = "roundrobin",
                    nodes = {
                        ["127.0.0.1:1980"] = 1
                    }
                },
                plugins = {
                    ["elasticsearch-logger"] = {
                        endpoint_addr = "http://127.0.0.1:9201",
                        field = {
                            index = "services"
                        },
                        auth = {
                            username = "elastic",
                            password = "$env://ES_PASSWORD"
                        },
                        batch_max_size = 1,
                        inactive_timeout = 1
                    }
                }
            })

            if code >= 300 then
                ngx.status = code
            end
            ngx.say(body)
        }
    }
--- response_body
passed



=== TEST 28: test route (auth.password from env variable, auth success)
--- main_config
env ES_PASSWORD=123456;
--- request
GET /hello
--- wait: 2
--- response_body
hello world
--- error_log
Batch Processor[elasticsearch-logger] successfully processed the entries
```

**What this tests:**
- **TEST 27** configures a route with `elasticsearch-logger` where `auth.password` uses `$env://ES_PASSWORD` to reference the password from an environment variable instead of hardcoding it. The `--- main_config` directive exposes the `ES_PASSWORD` env var to the nginx worker process.
- **TEST 28** sends a request to the route and verifies the elasticsearch-logger plugin successfully resolves the `$env://` reference via `fetch_secrets`, constructs the correct Basic auth header, and sends logs to Elasticsearch (verified by the `successfully processed the entries` log line).

This follows the same pattern as TEST 4/5 (auth with hardcoded password) but validates the `$env://` secret reference path added in this PR.

moonming (changes requested):
Hi @jr-instantsystem, thank you for working on secret support for elasticsearch-logger!

The use case is valid — storing ES credentials as environment variables or secrets is a common security practice. However, I have a few concerns with the current implementation:

1. **Performance concern**: Calling `fetch_secrets(conf)` inside `send_to_elasticsearch()` means secret resolution happens on every log batch send. This could add latency and unnecessary overhead. A better approach would be to resolve secrets during the `rewrite` or `check_schema` phase, or cache the resolved config.

2. **Missing test cases**: This PR needs test coverage to verify:
   - `\$env://MY_PASSWORD` resolves correctly in `auth.password`
   - `\$secret://vault/...` references work
   - Invalid/missing secret references are handled gracefully

3. **Consistency with other plugins**: Could you check how other logger plugins (e.g., `clickhouse-logger`, `kafka-logger`) handle secrets? Following the same pattern would be ideal. PR #12951 (clickhouse-logger secrets) is a good reference.

The change itself is small (2 lines), but getting the architecture right matters. Looking forward to the updates!

jr-instantsystem:
Hi @moonming,

I don't know how to cache the resolved conf after `fetch_secrets(conf)`, tried to look on other plugins but I don't get how to do it... But, they all do `fetch_secrets()` in `_M.access` or  `_M.rewrite` as you said, so I just moved the line there.

Plus, this fix an issue linked to my change for version 3.13+ (I'm running on 3.12)

Plus, I had to implement elasticsearch datastream support for our internal needs, it's pushed on this PR too...
Could be moved on an other ticket/PR if required, or we change the desc on this issue/PR, I don't know, you tell me :)

Baoyuantop:
Hi @jr-instantsystem, Please submit a separate PR to handle this feature, ensuring that each PR addresses only one task.

apisix/plugins/elasticsearch-logger.lua
```diff
@@ -306,6 +330,10 @@ function _M.log(conf, ctx)
     local metadata = plugin.plugin_metadata(plugin_name)
     local max_pending_entries = metadata and metadata.value and
                                 metadata.value.max_pending_entries or nil
+
+    -- resolve secrets & env vars
```
Baoyuantop: The behavior of `fetch_secrets` (lines 251-258 of `apisix/secret.lua`) is to execute `core.table.deepcopy(refs)` on every call to create a deep copy of the configuration, then iterate through and replace the secret references. The second parameter `true` indicates that the secret value itself should be looked up using an LRU cache, but the deep copy operation is performed every time.

It's recommended to refer to the clickhouse-logger pattern, where calling `fetch_secrets` only once in `send_to_elasticsearch` should cover all scenarios. Also, ensure that `fetch_and_update_es_version` in `_M.access` can use the parsed password—if needed, call it once in `_M.access` as well (just for version detection), but the call in `_M.log` is redundant.

Baoyuantop:
Please refer to https://github.com/apache/apisix/pull/13025#pullrequestreview-3941424735 and include this test code in your pull request.

apisix/plugins/elasticsearch-logger.lua
jr-instantsystem: What to you think about removing the call to `fetch_and_update_es_version` inside `send_to_elasticsearch` ?
As it's called in `_M.access`, it should already been executed before any execution of `send_to_elasticsearch`, no ?

jr-instantsystem:
> Please refer to [#13025 (review)](https://github.com/apache/apisix/pull/13025#pullrequestreview-3941424735) and include this test code in your pull request.

done

Baoyuantop (changes requested):
Hi @jr-instantsystem, thanks for the update. The direction is right, but there is an issue with the current `fetch_secrets` placement that will cause a regression. Let me explain.

`fetch_secrets(conf, true)` internally does a **deep copy** of the conf table and returns a brand new table with secrets resolved (see [`apisix/secret.lua` L256](https://github.com/apache/apisix/blob/master/apisix/secret.lua#L256)). It does **not** modify the original conf in place.

This causes two problems in the current PR:

**1) `fetch_secrets` inside `fetch_and_update_es_version()` breaks the version caching mechanism**

```lua
local function fetch_and_update_es_version(conf)
    if conf._version then
        return
    end

    -- resolve secrets & env vars
    conf = fetch_secrets(conf, true)   -- conf is now a NEW deep-copied table

    ...
    conf._version = major_version      -- writes to the copy, not the original
end
```

After this function returns, the caller still holds the original `conf` which never got `_version` set. So the `if conf._version then return end` guard will never be true, and **every request will send an HTTP GET to Elasticsearch to fetch the version** -- this is a significant performance regression.

**2) `fetch_and_update_es_version(conf)` was removed from `send_to_elasticsearch()`**

The batch processor calls `send_to_elasticsearch` in a background timer. If the version was not cached during `_M.access` (e.g. ES was temporarily unreachable), the original code had a fallback `fetch_and_update_es_version(conf)` call inside `send_to_elasticsearch`. The current PR removes this fallback.

**Suggested fix:**

The simplest correct approach is: only call `fetch_secrets` at the beginning of `send_to_elasticsearch`, keep `fetch_and_update_es_version` as it was, and do **not** add `fetch_secrets` inside `fetch_and_update_es_version`. Like this:

```lua
local function send_to_elasticsearch(conf, entries)
    local httpc, err = http.new()
    if not httpc then
        return false, str_format("create http error: %s", err)
    end

    -- resolve secrets & env vars (returns a deep copy, does not modify original conf)
    conf = fetch_secrets(conf, true)

    fetch_and_update_es_version(conf)

    local selected_endpoint_addr
    ...
```

This is the same pattern used by `clickhouse-logger` in PR #12951 -- call `fetch_secrets` once at the top of the send function and pass the resolved conf to the rest of the logic.

No changes needed to `fetch_and_update_es_version` itself.

jr-instantsystem:
@Baoyuantop Thanks for the details, didn't know about the deep copy :)

as fetch_secrets is making a copy of the object, would this be ok in `fetch_and_update_es_version` ?
```
    -- resolve secrets & env vars
    local conf_resolved = fetch_secrets(conf, true)

    ...
    local major_version, err = get_es_major_version(selected_endpoint_addr, conf_resolved)
    ...

    conf._version = major_version
```
Because we absolutly need to resolv secrets when calling `get_es_major_version`, to prevent any 401 to happen.
Or this could be moved inside `get_es_major_version` directly too, maybe better there

Baoyuantop:
Yes, you're right. Here's an example:

```lua
local function fetch_and_update_es_version(conf)
     if conf._version then
         return
     end

     -- resolve secrets for the HTTP request (deep copy, does not modify original)
     local conf_resolved = fetch_secrets(conf, true)

     local selected_endpoint_addr
     if conf_resolved.endpoint_addr then
         selected_endpoint_addr = conf_resolved.endpoint_addr
     else
         selected_endpoint_addr = conf_resolved.endpoint_addrs[
                                     math_random(#conf_resolved.endpoint_addrs)]
     end
     local major_version, err = get_es_major_version(selected_endpoint_addr,
                                                      conf_resolved)
     if err then
         core.log.error("failed to get Elasticsearch version: ", err)
         return
     end
     -- write to ORIGINAL conf to preserve caching across requests
     conf._version = major_version
 end
```

jr-instantsystem:
pushed

PR #13039, opened 2026-02-26
fix(ai-proxy): support Anthropic token field names in openai-base driver

iakuf:
## Summary

When using `openai-compatible` provider with Anthropic-format endpoints
(e.g. DeepSeek's `/anthropic/v1/messages`), the upstream returns
`input_tokens`/`output_tokens` instead of `prompt_tokens`/`completion_tokens`.

This causes `$llm_prompt_tokens` and `$llm_completion_tokens` to always be 0.

## Changes

Add fallback support for Anthropic field names in both streaming and
non-streaming paths of `openai-base.lua`:

- `prompt_tokens = usage.prompt_tokens or usage.input_tokens or 0`
- `completion_tokens = usage.completion_tokens or usage.output_tokens or 0`

This is fully backward compatible — OpenAI-format responses are unaffected.

### DIFF
--- a/apisix/plugins/ai-drivers/openai-base.lua
+++ b/apisix/plugins/ai-drivers/openai-base.lua
@@ -131,12 +131,12 @@ local function read_response(conf, ctx, res, response_filter)
                         core.log.info("got token usage from ai service: ",
                                             core.json.delay_encode(data.usage))
                         ctx.llm_raw_usage = data.usage
+                        local pt = data.usage.prompt_tokens or data.usage.input_tokens or 0
+                        local ct = data.usage.completion_tokens or data.usage.output_tokens or 0
                         ctx.ai_token_usage = {
-                            prompt_tokens = data.usage.prompt_tokens
-                                            or data.usage.input_tokens or 0,
-                            completion_tokens = data.usage.completion_tokens
-                                               or data.usage.output_tokens or 0,
-                            total_tokens = data.usage.total_tokens or 0,
+                            prompt_tokens = pt,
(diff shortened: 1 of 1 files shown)


iakuf:
This PR fixes token usage statistics when using openai-compatible provider 
with Anthropic-compatible endpoints (e.g. DeepSeek's /anthropic/v1/messages).
Anthropic format uses input_tokens/output_tokens instead of 
prompt_tokens/completion_tokens, causing $llm_prompt_tokens to always be 0.

Baoyuantop:
Hi @iakuf, test cases need to be added for the current changes.

moonming (changes requested):
Hi @iakuf, thanks for this small but useful fix!

Supporting Anthropic's token field names (`input_tokens`/`output_tokens`) alongside OpenAI's (`prompt_tokens`/`completion_tokens`) in the openai-base driver is reasonable — it provides a fallback when the response uses Anthropic's naming convention.

**One question**: Does this change affect how tokens are counted in Prometheus metrics? We want to make sure the `apisix_ai_llm_token_usage` metric correctly captures tokens regardless of which field name the upstream uses.

**Also**: The 12-line change is clean, but could you add a brief test case showing that Anthropic-style token fields are correctly parsed?

Thanks!

Baoyuantop:
Hi @iakuf, following up on the previous review comments. Please let us know if you have any updates. Thank you.

Baoyuantop:
fixed in https://github.com/apache/apisix/pull/13170

PR #13042, opened 2026-02-27
test(aws): correct secret path and test

janiussyafiq:
### Description

Fixed incorrect secret paths in AWS secrets manager tests that were causing test failures.

**Problem**
The test file `t/secret/aws.t` had incorrect secret reference paths in TEST 8. The paths included an extra `/key` suffix that doesn't exist in the actual secret structure and missing the secret's name `apisix-key`.
https://github.com/apache/apisix/blob/3ccbfd6e4678834585af3b4a7198982c5d020034/t/secret/aws.t#L269-L271

According to the CI setup script, LocalStack creates the secret with this structure:
  ```json
  {"jack":"value"}
```
  https://github.com/apache/apisix/blob/3ccbfd6e4678834585af3b4a7198982c5d020034/ci/init-common-test-service.sh#L25

To access the value for jack, the correct path should be:
- `$secret://aws/mysecret/apisix-key/jack` instead of `$secret://aws/mysecret/jack/key` (incorrect - has extra /key and missing /apisix-key)

Changes:
1. Fixed secret reference paths in TEST 8 (lines 270, 282, 297, 308) by removing the invalid /key suffix
2. Corrected the expected response body to match actual test behavior (the second fetch happens after the secret is deleted, so it returns nil)

This ensures the tests correctly validate AWS Secrets Manager integration functionality.

Fixes #

### Checklist

### DIFF
--- a/t/secret/aws.t
+++ b/t/secret/aws.t
@@ -267,7 +267,7 @@ GET /t
                     "username": "jack",
                     "plugins": {
                           "key-auth": {
-                            "key": "$secret://aws/mysecret/jack/key"
+                            "key": "$secret://aws/mysecret/apisix-key/jack"
                         }
                     }
                 }]]
@@ -279,7 +279,10 @@ GET /t
 
 
             local secret = require("apisix.secret")
-            local value = secret.fetch_by_uri("$secret://aws/mysecret/jack/key")
+            local value, err = secret.fetch_by_uri("$secret://aws/mysecret/apisix-key/jack")
+            if value then
+                ngx.say("secret value: ", value)
+            end
 
 
             local code, body = t('/apisix/admin/secrets/aws/mysecret', ngx.HTTP_DELETE)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: t/secret/aws.t [hunk 1]
```diff
@@ -305,12 +308,14 @@ GET /t
             end
 
             local secret = require("apisix.secret")
-            local value = secret.fetch_by_uri("$secret://aws/mysecret/jack/key")
-            if value then
-                ngx.say("secret value: ", value)
+            local value, err = secret.fetch_by_uri("$secret://aws/mysecret/apisix-key/jack/key")
+            if err then
+                ngx.say(err)
```
shreemaan-abhishek, line 313:
```suggestion
                ngx.say(err)
                return
```

if we detect an error we should return so that `ngx.say("all done")` is not executed. This is a common mistake we make when writing tests using test::nginx.

janiussyafiq: @shreemaan-abhishek however the err there was expected since we deleted the mysecret hence value should be nil, or is there a cleaner approach?

shreemaan-abhishek: oh okay my bad

Baoyuantop (approved):

PR #13048, opened 2026-02-28
feat(ai-proxy): support automatic protocol translation from Claude to OpenAI

moonming:
## Why this PR?

This PR introduces an Automatic Protocol Translation feature to the `ai-proxy` plugin. This feature enables the APISIX API Gateway to automatically detect AI requests that are directed to the `/v1/messages` endpoint (which is the standard Claude API format), translate the request into the OpenAI format (`/v1/chat/completions`), forward the request to an OpenAI-compatible upstream provider, and lastly translate the response back into the Claude API format.

This allows AI clients, extensions, and tooling that strictly rely on the Claude API format to seamlessly work with any backend models exposed through OpenAI-compatible APIs behind APISIX, expanding integration possibilities without requiring changes on the client side.

## What this PR does?

- Added a new module `apisix/plugins/ai-proxy/converter/claude_to_openai.lua` containing the core logic to convert JSON requests, JSON responses, and SSE streaming chunks between Claude and OpenAI formats.
- Updated `apisix/plugins/ai-proxy/base.lua` to automatically detect requests targeting `/v1/messages` and set a context flag.
- Updated `apisix/plugins/ai-drivers/openai-base.lua` to invoke the converter functions before sending the request and during response processing (for both standard and streaming paths) when the flag is present.
- Wrote an end-to-end `test::nginx` test suite in `t/plugin/ai-proxy-claude.t`, covering:
  - Standard Chat Requests
  - Complex System Prompts (Array format)
  - Streaming SSE Responses
  - Abnormal Requests (e.g. missing messages array)
- Added documentation for the new feature in `docs/en/latest/plugins/ai-proxy.md`, including a Mermaid architecture sequence diagram and concrete usage examples.

### DIFF
--- a/apisix/plugins/ai-drivers/openai-base.lua
+++ b/apisix/plugins/ai-drivers/openai-base.lua
@@ -16,6 +16,8 @@
 --
 local _M = {}
 
+local claude_converter = require("apisix.plugins.ai-proxy.converter.claude_to_openai")
+
 local mt = {
     __index = _M
 }
@@ -60,6 +62,10 @@ function _M.validate_request(ctx)
         local request_table, err = core.request.get_json_request_body_table()
         if not request_table then
             return nil, err
+        end
+
+        if ctx.ai_client_protocol == "claude" then
+            request_table = claude_converter.convert_request(request_table)
         end
 
         return request_table, nil
@@ -147,6 +153,7 @@ local function read_response(conf, ctx, res, response_filter)
                 ::CONTINUE::
             end
 
+            if ctx.ai_client_protocol == "claude" then chunk = claude_converter.convert_sse_events(ctx, chunk) end
             plugin.lua_response_filter(ctx, res.headers, chunk)
         end
     end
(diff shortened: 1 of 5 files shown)


moonming:
**Architectural Note (self-review)**:

This PR (#13048) and #13050 (anthropic-native driver by @iakuf) address the same need — Anthropic/Claude support — but with different approaches:

- **This PR**: Protocol translation (Claude ↔ OpenAI format conversion) — more general, single code path
- **#13050**: Native Anthropic driver — more efficient, preserves Anthropic-specific features

**My current thinking**: Protocol translation as the default approach, with native drivers as optional optimizations for providers that have significantly different capabilities (like Anthropic's tool use format).

Need to finalize this direction before either PR can move forward. Will update both PRs once the decision is made.

Baoyuantop:
The issues in this PR have been resolved in https://github.com/apache/apisix/pull/13170 and https://github.com/apache/apisix/pull/13181

PR #13050, opened 2026-02-28
feat(ai-proxy-multi): add anthropic-native provider for native Anthropic Messages API

iakuf:
## Summary

Add a new `anthropic-native` driver that speaks the native Anthropic Messages API protocol (`/v1/messages`) directly.

The existing `anthropic` provider uses the OpenAI-compatible driver, which works for basic use cases but cannot handle Anthropic-specific protocol details. The new `anthropic-native` provider addresses this.

## Key differences from `openai-base` driver

- **SSE event types**: handles `message_start`, `content_block_delta`, `message_delta`, `message_stop` (no `[DONE]` sentinel)
- **Token usage fields**: reads `input_tokens`/`output_tokens` instead of `prompt_tokens`/`completion_tokens`
- **Request format**: removes `stream_options` (not supported by Anthropic), injects `anthropic-version` header automatically
- **Response format**: extracts text from `content[].text` array instead of `choices[].message.content`
- **System prompt**: accepts top-level `system` field and flexible `content` types (string or array of content blocks)

## Changes

- `apisix/plugins/ai-drivers/anthropic-native.lua`: new driver implementation
- `apisix/plugins/ai-drivers/schema.lua`: register `anthropic-native` provider with its own request schema
- `t/assets/anthropic-native-response.json`: mock response fixture for tests
- `t/plugin/ai-proxy-multi.anthropic-native.t`: 26 test cases covering non-streaming, streaming, token usage, TTFT, error handling, and schema validation
- `docs/en/latest/plugins/ai-proxy-multi.md`: document the new provider with a usage example

## Testing

All 26 tests pass:

### DIFF
--- a/apisix/plugins/ai-drivers/anthropic-native.lua
+++ b/apisix/plugins/ai-drivers/anthropic-native.lua
@@ -1,0 +1,399 @@
+--
+-- anthropic-native.lua
+-- A driver for the Anthropic Messages API native protocol (/v1/messages).
+-- Handles Anthropic-specific SSE event types, response format, and token usage fields.
+-- Compatible with any endpoint that speaks the native Anthropic protocol,
+-- e.g. api.anthropic.com/v1/messages or api.deepseek.com/anthropic/v1/messages.
+--
+-- Differences from openai-base:
+--   Request:  removes stream_options (not supported), adds anthropic-version header
+--   Response: content[].text  (not choices[].message.content)
+--   SSE text: event=content_block_delta, delta.type=text_delta, delta.text
+--   SSE token: message_start -> input_tokens; message_delta -> output_tokens
+--   SSE end:  event=message_stop  (no [DONE] sentinel)
+--   Token fields: input_tokens / output_tokens  (not prompt_tokens / completion_tokens)
+--
+
+local _M = {}
+
+local mt = { __index = _M }
+
+local CONTENT_TYPE_JSON = "application/json"
+local ANTHROPIC_VERSION = "2023-06-01"
+
+local core    = require("apisix.core")
+local plugin  = require("apisix.plugin")
+local http    = require("resty.http")
+local url     = require("socket.url")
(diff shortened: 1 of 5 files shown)


apisix/plugins/ai-drivers/anthropic-native.lua
```diff
@@ -0,0 +1,399 @@
...
+        if not ok then
+            core.log.warn("failed to keepalive connection: ", ka_err)
+        end
+    end
+
+    return code, body
+end
+
+
+return _M
```
Baoyuantop, line 399: Following the pattern of anthropic.lua or vertex-ai.lua, make the module entry point return openai-base.new(...) or an equivalent instance object.

apisix/plugins/ai-drivers/anthropic-native.lua
```diff
@@ -0,0 +1,399 @@
+--
+-- anthropic-native.lua
```
Baoyuantop, line 2: Requires the inclusion of the Apache Software Foundation standard license header.

moonming (changes requested):
Hi @iakuf, thank you for the substantial work on the anthropic-native provider! This is a significant contribution (1233 lines).

**Context**: There's an ongoing architectural discussion about how to support Anthropic's API:
- **This PR (#13050)**: Adds a native Anthropic driver that speaks the Anthropic Messages API directly
- **#13048**: Takes a protocol translation approach (Claude ↔ OpenAI format conversion)

Both approaches have merit:
- Native driver: More efficient, preserves Anthropic-specific features (tool use, vision, etc.)
- Protocol translation: Simpler to maintain, single code path for request/response handling

**Before we can proceed**, we need to decide on the technical direction. I'll make a decision on this and update both PRs. In the meantime, could you:

1. Address any existing review comments
2. Ensure tests cover the key Anthropic API features (streaming, tool use, vision)
3. Rebase if needed

Thank you for your patience while we sort out the architecture! Your implementation quality is appreciated.

PR #13054, opened 2026-03-01
feat(discovery): add kubernetes discovery data warm-up in master process

ChuanFF:
### Description

This PR introduces a data warm-up mechanism for the Kubernetes discovery module. Previously, APISIX worker processes might start handling traffic before the Kubernetes endpoints cache was fully populated, potentially causing initial request failures due to missing upstream nodes.

With this change, the master process now performs an immediate, synchronous list operation against the Kubernetes API server during initialization (init phase), ensuring that endpoint data is pre-loaded into shared memory before worker processes are spawned.

The warm-up execution is synchronous and blocking but only runs once in the master process during startup. Since it performs a single list request (not watch), the latency impact is minimal and acceptable for initialization phase.

Other types of service discovery, such as Consul, can also implement init() to support the data warm-up.

### Checklist

### DIFF
--- a/apisix/discovery/init.lua
+++ b/apisix/discovery/init.lua
@@ -37,6 +37,16 @@ function discovery.init_worker()
     end
 end
 
+function discovery.init()
+    if discovery_type then
+        for discovery_name, _ in pairs(discovery_type) do
+            if discovery[discovery_name].init then
+                discovery[discovery_name].init()
+            end
+        end
+    end
+end
+
 return {
     version = 0.1,
     discovery = discovery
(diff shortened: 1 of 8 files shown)


ChuanFF:
threre is a bug :https://github.com/apache/apisix/issues/13055, i modify order of `KUBERNETES_CLIENT_TOKEN_FILE` and `KUBERNETES_CLIENT_TOKEN` to avoid bug.

moonming (commented):
Hi @ChuanFF, thank you for the Kubernetes discovery data warm-up!

Pre-loading Kubernetes endpoints in the master process to avoid cold-start 404s in workers is an excellent improvement for Kubernetes deployments. I'll review the implementation in detail shortly.

PR #13061, opened 2026-03-02
fix: Support secrets for client ca

cybersec-jochenarnold:
### Description

Allow secrets like env or secret for client ca in ssl config. 
Useful to fetch ca cert directly from e.g. Vault.

Fixes #13021

### Checklist

### DIFF
--- a/t/node/ssl.t
+++ b/t/node/ssl.t
@@ -30,6 +30,7 @@ BEGIN {
     set_env_from_file('TEST_KEY', 't/certs/apisix.key');
     set_env_from_file('TEST2_CERT', 't/certs/test2.crt');
     set_env_from_file('TEST2_KEY', 't/certs/test2.key');
+    set_env_from_file('TEST_CA_CERT', 't/certs/mtls_ca.crt');
 }
 
 use t::APISIX 'no_plan';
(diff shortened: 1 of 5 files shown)


t/node/ssl.t
janiussyafiq, line 1: I think you can move the test case to `client-mtls.t` https://github.com/apache/apisix/blob/16b9d7e5b611a729c6eb99d3fbca6c8bce71f494/t/node/client-mtls.t, since the test cases involve mTLS (client side verification)

Baoyuantop:
Hi @cybersec-jochenarnold, following up on the previous review comments. Please let us know if you have any updates. Thank you.

moonming (changes requested):
Hi @cybersec-jochenarnold, thank you for this fix!

This addresses the same gap as #13062 — secret references not being resolved in SSL configurations. The client CA case is equally important.

**Suggestion**: To streamline the review process, could you merge this change into #13062? Having both upstream SSL and client CA secret resolution in a single PR would be easier to review and test holistically.

If you'd prefer to keep them separate, that's fine too — just please add test cases for the client CA secret resolution path.

Thank you!

cybersec-jochenarnold:
@moonming Good point—this might also make testing easier, since the two fixes are very similar.

Will be fixed with #13062

PR #13062, opened 2026-03-02
fix: fetch secrets from upstream ssl config

### LINKED ISSUE #13013: bug: nginx env variable for private key is not resolved in upstream with mtls
### Current Behavior

Using mTLS for HTTPS Upstream with $env:// for Certificate and Private Key does not work. I think it does not resolve the variable "$env://API_UPSTREAM_KEY]". It tries to decrypt it instead.

I haven't been able to figure out where the variables for Upstream TLS are resolved in the code yet.

I also configured upstream tls client directly (not via extra SSL ID) - this works as expected. Maybe the resolving of the env vars is missing if the SSL config is loaded via get_by_id(cert_id)?

I will try to add some debug logs in the next days.

init.lua: handle_upstream()
```
    if api_ctx.matched_upstream and api_ctx.matched_upstream.tls and
        api_ctx.matched_upstream.tls.client_cert_id then

        local cert_id = api_ctx.matched_upstream.tls.client_cert_id
        local upstream_ssl = router.router_ssl.get_by_id(cert_id)
        if not upstream_ssl or upstream_ssl.type ~= "client" then
            local err  = upstream_ssl and
                "ssl type should be 'client'" or
                "ssl id [" .. cert_id .. "] not exits"
            core.log.error("failed to get ssl cert: ", err)

            if is_http then
                return core.response.exit(502)
            end

            return ngx_exit(1)
        end

        core.log.info("matched ssl: ",
                  core.json.delay_encode(upstream_ssl, true))
        api_ctx.upstream_ssl = upstream_ssl
    end
```

### Expected Behavior

Using nginx env variable should be working for secrets like private key.

### Error Logs

ssl.lua:162: aes_decrypt_pkey(): base64 decode ssl key failed. key[$env://API_UPSTREAM_KEY] ...

### Steps to Reproduce

1. Run APISIX binary on Ubuntu
2. Create Service with Upstream and refer to a SSL ID
3. Create SSL with type=client

Used ADC but curl to Admin API does not work too.

```
services:
  - id: apigw-service
    name: apigw-service
    upstream:
      name: apigw-upstream
      type: roundrobin
      scheme: https
      pass_host: "node"
      nodes:
        - host: "example.com"
          port: 443
          weight: 1
      tls:
        client_cert_id: "UPSTREAM_CLIENT_CERT_ID"
    routes:
      - name: apigw-route
        plugins:
          proxy-rewrite:
            headers:
              set:
                apikey: "REDACTED"
        uris:
          - "/api/example-api"
        methods:
          - GET
```

```
ssls:
  - id: "UPSTREAM_CLIENT_CERT_ID"
    certificates:
      - certificate: "$env://API_UPSTREAM_CERT"
        key: "$env://API_UPSTREAM_KEY"
    snis:
      - "example.com"
    ssl_protocols:
      - TLSv1.2
      - TLSv1.3
    type: client
```

### Environment

- Operating system (run `uname -a`): Ubuntu
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): -
- APISIX Dashboard version, if relevant: not used
- Plugin runner version, for issues related to plugin runners: -
- LuaRocks version, for installation issues (run `luarocks --version`): -

### LINKED ISSUE #13020: bug: Difference between ADC and APISIX schemas regarding secrets in client CA for upstream mTLS
### Current Behavior

Schema does not allow using secret/env variables in SSL Client CA value.

```
curl http://127.0.0.1:9180/apisix/admin/ssls/1 -H "X-API-KEY: REDACTED" -X PUT -d '
{
     "cert" : "$env://SERVER_CERT",
     "key": "$env://SERVER_KEY",
     "snis": ["example.com"],
     "client": {
        "ca": "$env://CA_CERT"
     },
     "ssl_protocols": [
         "TLSv1.2",
         "TLSv1.3"
     ]
}'

{"error_msg":"invalid configuration: property \"client\" validation failed: property \"ca\" validation failed: string too short, expected at least 128, got 18"}
```

https://github.com/apache/apisix/blob/2500db7a4ae0c047606e8233821f92d318fb589e/apisix/schema_def.lua#L841

### Expected Behavior

ADC schema allows secret var (secret, env) in this property. 

I would expect APISIX to process this value as well. Current implementation via secret.fetch_secrets should already allow this.

https://github.com/api7/adc/blob/main/schema.json#L1129

```
          "client": {
            "type": "object",
            "properties": {
              "ca": {
                "anyOf": [
                  {
                    "type": "string",
                    "minLength": 128,
                    "maxLength": 65536
                  },
                  {
                    "type": "string",
                    "pattern": "^\\$(secret|env):\\/\\/"
                  }
                ]
              },
              ...
```

APISIX should allow certificate_scheme and secret_uri_schema.

https://github.com/apache/apisix/blob/2500db7a4ae0c047606e8233821f92d318fb589e/apisix/schema_def.lua#L771

```
{
            type = "array",
            items = {
                oneOf = {
                    certificate_scheme,
                    secret_uri_schema
                }
            }
        },
```

### Error Logs

{"error_msg":"invalid configuration: property \"client\" validation failed: property \"ca\" validation failed: string too short, expected at least 128, got 18"}

### Steps to Reproduce

Try to create SSL object with "$env://" as client ca value.

```
curl http://127.0.0.1:9180/apisix/admin/ssls/1 -H "X-API-KEY: REDACTED" -X PUT -d '
{
     "cert" : "$env://SERVER_CERT",
     "key": "$env://SERVER_KEY",
     "snis": ["example.com"],
     "client": {
        "ca": "$env://CA_CERT"
     },
     "ssl_protocols": [
         "TLSv1.2",
         "TLSv1.3"
     ]
}'
```

### Environment

- Operating system (run `uname -a`): Ubuntu
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): penresty/1.27.1.2
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): -
- APISIX Dashboard version, if relevant: -
- Plugin runner version, for issues related to plugin runners: -
- LuaRocks version, for installation issues (run `luarocks --version`): -

cybersec-jochenarnold:
### Description

This PR fixes an issue where $ENV:// and $secret:// certificate references
in the upstream configuration with mTLS were not resolved when the SSL configuration was linked via an ID.

Added fetch_secrets function to handle_upstream in init.lua: Line 578:
` api_ctx.upstream_ssl = apisix_secret.fetch_secrets(upstream_ssl, true)`

Fixes #13013

### Checklist

Fixes #13020


Baoyuantop:
Hi @cybersec-jochenarnold, following up on the previous review comments. Please let us know if you have any updates. Thank you.

moonming (changes requested):
Hi @cybersec-jochenarnold, thank you for fixing secret resolution in upstream SSL! This is a real security gap — mTLS upstream configs using `\$ENV://` or `\$secret://` references should definitely be resolved.

**The approach looks correct**: Adding `fetch_secrets()` in `handle_upstream()` before assigning `api_ctx.upstream_ssl` is the right location.

**Suggestions**:

1. **Combine with #13061**: I see you also have #13061 (secrets for client CA) from the same area. Would it be possible to combine both fixes into a single PR? This would make review easier and ensure the changes are tested together.

2. **Error handling**: The nil check on `new_upstream_ssl` with a 502 response is appropriate. Good defensive coding!

3. **Test approach**: The Vault-based tests in `upstream-mtls.t` are thorough. The `set_env_from_file` helper in the BEGIN block is a nice pattern for loading cert content into env vars.

If you can merge the client CA fix (#13061) into this PR, I think we can move this forward efficiently. Thank you for the security-conscious contribution! 🔒

apisix/init.lua
```diff
@@ -552,9 +552,22 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
             return ngx_exit(1)
         end
 
+        local new_upstream_ssl = apisix_secret.fetch_secrets(upstream_ssl, true)
+
+        if not new_upstream_ssl then
```
Baoyuantop: `fetch_secrets` only returns `nil` if the input is not a table. Since `upstream_ssl` here is an SSL object (table) obtained from `router.router_ssl.get_by_id()`, `fetch_secrets` will never return `nil`.

When a single secret query fails, the logic in `retrieve_refs` is `refs[k] = fetch(v, use_cache) or v`—that is, preserving the original `$secret://...` or `$env://...` string, instead of returning `nil`. This means:

- The `nil` check indicates the branch is unreachable, giving a false sense of security that "the error has been handled".

- The actual failure scenario (secret parsing failure, preserving an unparsed URI string) will not be intercepted and will be passed downstream, causing TLS handshake/decryption failures and generating incomprehensible error messages.

It is recommended to adopt a forgiving mode consistent with radixtree_sni.lua—remove nil checks and use `or upstream_ssl` for fallback.

t/node/ssl.t
```diff
@@ -63,7 +63,7 @@ __DATA__
 
 === TEST 1: store two certs and keys in vault
 --- exec
-VAULT_TOKEN='root' VAULT_ADDR='http://0.0.0.0:8200' vault kv put kv/apisix/ssl \
+VAULT_TOKEN='root' VAULT_ADDR='http://172.17.0.1:8200' vault kv put kv/apisix/ssl \
```
Baoyuantop: Why does this file need to be modified?

apisix/cli/file.lua
```diff
@@ -328,7 +328,7 @@ function _M.read_yaml_conf(apisix_home)
             -- Therefore we need to check the absolute version instead
             local cert_path = pl_path.abspath(apisix_ssl.ssl_trusted_certificate)
             if not pl_path.exists(cert_path) then
-                util.die("certificate path", cert_path, "doesn't exist\n")
+                util.die("certificate path ", cert_path, " doesn't exist\n")
```
Baoyuantop, line 345: Is this change related to this PR?

t/node/ssl.t
cybersec-jochenarnold: I had some problems with testing vault using my local dev container. To speed things up, I manually started the vault container. I didn't want to push for that change.

apisix/cli/file.lua
cybersec-jochenarnold: While I was fixing this bug, I noticed that there are missing spaces in the log message for broken links.

apisix/schema_def.lua
```diff
@@ -838,7 +838,12 @@ _M.ssl = {
         client = {
             type = "object",
             properties = {
-                ca = certificate_scheme,
+                ca = {
+                    oneOf = {
+                        certificate_scheme,
+                        secret_uri_schema
```
nic-6443: This hunk should be dropped when you merge master (the PR is currently conflicting): `secret_uri_schema` was removed in #13312, which made schema validation skip `$secret://`/`$env://` string values centrally (`skip_validation = secret.is_secret_ref` in `apisix/core/schema.lua`). So on current master `client.ca = "$env://..."` already passes the schema check without any schema change. Worse, git auto-merges this file, so after merging master this would reference an undefined variable and the `oneOf` silently degrades to a no-op.

The part of #13020 that still needs fixing on master is the unconditional `validate(conf.client.ca, nil)` in `check_ssl_conf` — your ssl.lua change covers exactly that, so the fix stays intact without this hunk.

apisix/init.lua
```diff
@@ -552,9 +552,22 @@ function _M.handle_upstream(api_ctx, route, enable_websocket)
             return ngx_exit(1)
         end
 
+        local new_upstream_ssl = apisix_secret.fetch_secrets(upstream_ssl, true)
```
nic-6443: `fetch_secrets()` does a `core.table.deepcopy` of the whole SSL object on every request through this path, even when the config contains no secret refs at all — the `true` arg only caches the individual URI lookups (`secrets_cache` in `apisix/secret.lua` is keyed by URI), it doesn't avoid the per-request copy. Unlike the SNI path in `radixtree_sni.lua` which runs per handshake, this runs for every proxied request to an upstream with `tls.client_cert_id`.

How about guarding it with `secret.has_secret_ref(upstream_ssl)` (added in #13312, available once you merge master)? That keeps the zero-copy assignment for the common case where the SSL object holds plain cert/key.

apisix/schema_def.lua
AlinsRan: Right, dropped that hunk when merging master. `secret_uri_schema` is gone and `skip_validation = secret.is_secret_ref` already lets `$env://` / `$secret://` through the schema check. Only the `check_ssl_conf` guard in ssl.lua is left.

apisix/init.lua
AlinsRan: Done. Master refactored this into `resolve_upstream_client_cert()`, so the resolve now lives there and is gated on `secret.has_secret_ref(upstream_ssl)` — plain cert/key keeps the zero-copy path, and the stream subsystem gets the fix too since it shares that helper.

apisix/init.lua
AlinsRan: Agreed, the nil check was dead code. Removed it, now it matches radixtree_sni.lua: `fetch_secrets(...) or upstream_ssl`.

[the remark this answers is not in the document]
AlinsRan: Removed the nil check entirely — it was unreachable. The resolve is now `fetch_secrets(...) or upstream_ssl`, gated on `secret.has_secret_ref()`.

t/node/upstream-mtls.t
```diff
@@ -783,3 +809,125 @@ passed
...
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin")
+            local json = require("toolkit.json")
+
+            local data = {
+                type = "client",
+                cert = "$secret://vault/test/ssl/mtls_client.crt",
+                key = "$secret://vault/test/ssl/mtls_client.key"
+            }
```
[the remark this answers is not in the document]
AlinsRan: Vault backend is configured now — TEST 1 writes the cert/key into Vault and TEST 2 creates `/apisix/admin/secrets/vault/test`.

t/node/upstream-mtls.t
```diff
@@ -783,3 +809,125 @@ passed
 GET /hello
 --- response_body
 hello world
+
+=== TEST 24: get cert by tls.client_cert_id with secrets using secrets
```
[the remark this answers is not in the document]
AlinsRan: Renamed to `get cert by tls.client_cert_id with $secret:// refs`.

membphis:
P2 follow-up: `resolve_upstream_client_cert()` now calls `apisix_secret.fetch_secrets(..., true)` from `handle_upstream`. Cache hits are cheap, but a cold/expired cache can still call the secret manager on the proxied request path. Please consider moving upstream SSL secret resolution to config-change/background refresh, or caching a resolved SSL object by SSL resource version, so request handling only reads already-resolved cert/key material.

membphis (approved):
LGTM

nic-6443 (approved):

moonming (approved):

cybersec-jochenarnold:
@AlinsRan Thanks for circling back on this. I didn’t manage to look into it myself, sadly.

PR #13066, opened 2026-03-04
fix: use shdict instead of events module for nodes data exchange

### LINKED ISSUE #12398: bug: 503 errors occur after restarting APISIX when using Consul and APISIX both in Docker

shreemaan-abhishek:
### Description

Fixes #12398

### Checklist

### DIFF
--- a/apisix/discovery/consul/init.lua
+++ b/apisix/discovery/consul/init.lua
@@ -55,10 +55,16 @@ local skip_service_map = core.table.new(0, 1)
 local dump_params
 
 local consul_services
--- per-worker LRU cache: avoids JSON-decoding on every _M.nodes() call and
--- returns the same table instance while the shared dict value is unchanged,
--- so that compare_upstream_node() fast-path (old_t == new_t) works
-local nodes_cache = core.lrucache.new({ttl = 300, count = 1024})
+-- Per-worker LRU cache: avoids shared dict access on every request.
+-- neg_ttl caches unknown services. invalid_stale ensures expired
+-- entries are refreshed from the shared dict instead of re-cached.
+local nodes_cache = core.lrucache.new({
+    ttl = 1,
+    count = 1024,
+    invalid_stale = true,
+    neg_ttl = 1,
+    neg_count = 64,
+})
 
 local default_skip_services = {"consul"}
 local default_random_range = 5
@@ -71,6 +77,22 @@ local max_retry_time = 256
 local _M = {
     version = 0.3,
 }
+
+
+local function fetch_node_from_shdict(service_name)
+    local value = consul_dict:get(service_name)
+    if not value then
+        return nil, "consul service not found: " .. service_name
+    end
+
+    local nodes, err = core.json.decode(value)
+    if not nodes then
+        return nil, "failed to decode nodes for service: "
+                    .. service_name .. ", error: " .. (err or "")
+    end
+
+    return nodes
+end
 
 
 function _M.all_nodes()
@@ -96,15 +118,8 @@ end
 
 
 function _M.nodes(service_name)
-    local value = consul_dict:get(service_name)
-    if not value then
-        log.error("consul service not found: ", service_name, ", return default service")
-        return default_service and {default_service}
-    end
-
-    -- use the raw JSON string as version: same string = cache hit (same table
-    -- instance returned), different string = re-decode and cache new table
-    local nodes, err = nodes_cache(service_name, value, core.json.decode, value)
+    local nodes, err = nodes_cache(service_name, nil,
+                                   fetch_node_from_shdict, service_name)
     if not nodes then
         log.error("fetch nodes failed by ", service_name, ", error: ", err)
         return default_service and {default_service}


apisix/discovery/consul/init.lua
```diff
@@ -149,14 +162,21 @@ local function read_dump_services()
...
-    log.info("load dump file into memory success")
+    for k, v in pairs(entity.services) do
+        local content, json_err = core.json.encode(v)
+        if content then
+            consul_dict:set(k, content)
+        else
+            log.error("failed to encode dump service: ", k, ", error: ", json_err)
+        end
+    end
+    log.info("load dump file into shared dict success")
```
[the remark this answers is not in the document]
shreemaan-abhishek: since the values are set using service keys, I don't think this should be of concern.

apisix/discovery/consul/init.lua
```diff
@@ -66,53 +73,73 @@ local _M = {
...
+    return services
 end
 
 
 function _M.nodes(service_name)
-    if not all_services then
-        log.error("all_services is nil, failed to fetch nodes for : ", service_name)
-        return
+    local value = consul_dict:get(service_name)
+    if not value then
```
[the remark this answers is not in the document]
shreemaan-abhishek: we don't want stale values, so we can ignore this.

apisix/discovery/consul/init.lua
```diff
@@ -611,35 +637,32 @@ end
...
     end
+
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    -- flush stale data that may persist across reloads,
+    -- since consul_services is re-initialized empty
+    consul_dict:flush_all()
+
```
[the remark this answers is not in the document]
shreemaan-abhishek: the `stale` data mentioned in the comment does not mean "data that has expired", it means data that is no longer up to date according to consuls state. So we should avoid flushing only expired data to obtain a clean slate upon restart.

apisix/discovery/consul/init.lua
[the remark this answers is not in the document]
shreemaan-abhishek: we don't want expired shdict data so we should skip this.

apisix/discovery/consul/init.lua
```diff
@@ -611,35 +637,32 @@ end
...
         default_service.weight = default_weight
     end
+
+    if process.type() ~= "privileged agent" then
+        return
+    end
+
+    -- flush stale data that may persist across reloads,
+    -- since consul_services is re-initialized empty
+    consul_dict:flush_all()
```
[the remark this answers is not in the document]
shreemaan-abhishek: in this context `stale data` does not mean expired shdict data. Stale means data that is not up to date according to the state in consul. The idea behind doing `flush_all` is to start with a clean slate.

apisix/discovery/consul/init.lua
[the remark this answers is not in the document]
shreemaan-abhishek: this bug was only reproducible when apisix was deployed in docker.

apisix/discovery/consul/init.lua
```diff
@@ -149,14 +176,21 @@ local function read_dump_services()
...
+    log.info("load dump file into shared dict success")
 end
 
 
 local function write_dump_services()
     local entity = {
-        services = all_services,
+        services = _M.all_nodes(),
         last_update = ngx.time(),
         expire = dump_params.expire, -- later need handle it
```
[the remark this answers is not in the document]
shreemaan-abhishek: done

apisix/discovery/consul/init.lua
```diff
@@ -66,53 +73,82 @@ local _M = {
...
-        ", all services: ", json_delay_encode(all_services, true))
-end
-
-
 function _M.all_nodes()
-    return all_services
+    local keys = consul_dict:get_keys(0)
+    local services = core.table.new(0, #keys)
+    for _, key in ipairs(keys) do
+        local value = consul_dict:get(key)
```
membphis, line 102: After calculating every 10~20 keys, call `ngx.sleep(0)`

apisix/discovery/consul/init.lua
```diff
@@ -66,53 +73,82 @@ local _M = {
...
 
 
 function _M.nodes(service_name)
-    if not all_services then
-        log.error("all_services is nil, failed to fetch nodes for : ", service_name)
-        return
+    local value = consul_dict:get(service_name)
+    if not value then
+        log.error("consul service not found: ", service_name, ", return default service")
+        nodes_cache[service_name] = nil
```
membphis: `nodes_cache`

This cache may contain some useless, expired data. To prevent it from growing indefinitely, it needs to be cleaned up periodically (deleting useless cached data).

Another way,  an LRU (Local Redirect) approach can be used for easier and more direct control over the cache size.

membphis (changes requested):

apisix/cli/ngx_tpl.lua
```diff
@@ -78,6 +78,7 @@ lua {
     lua_shared_dict status-report {* meta.lua_shared_dict["status-report"] *};
     {% end %}
     lua_shared_dict nacos 10m;
+    lua_shared_dict consul 10m;
```
membphis: 10m is not enough for everyone

Allow users to customize configurations

apisix/discovery/consul/init.lua
```diff
@@ -66,53 +73,80 @@ local _M = {
...
-    return resp_list
+    return nodes
 end
 
 
 local function update_all_services(consul_server_url, up_services)
-    -- clean old unused data
+    -- write new/updated values first so readers never see a missing service
+    for k, v in pairs(up_services) do
+        local content, err = core.json.encode(v)
```
membphis, line 139: need the same way too

```
if i % 100 == 0 then
            ngx.sleep(0)
        end
```

membphis (dismissed):

Baoyuantop (approved):

apisix/discovery/consul/init.lua
```diff
@@ -66,53 +73,85 @@ local _M = {
...
+    return services
 end
 
 
 function _M.nodes(service_name)
-    if not all_services then
-        log.error("all_services is nil, failed to fetch nodes for : ", service_name)
-        return
+    local value = consul_dict:get(service_name)
+    if not value then
```
bzp2010: Each request using service discovery will directly retrieve from shdict, which I suspect may cause lock contention.
Why not adopt the following pattern:

1. Retrieve the node for the service name from lrucache. If present, it will return quickly without contention.
2. If service does not exist in lrucache, retrieve authoritative data prepared by privileged processes via the initialization function from shdict. Lrucache will then cache and return these.
3. If the shdict also doesn't contain it, it's best to log the "not found" result in lrucache to prevent an unexpected service name from consistently bypassing the fast path cache. It will be cached via `neg_ttl`, https://github.com/apache/apisix/blob/master/apisix/core/lrucache.lua#L130.

This ensures most requests complete on the fast path without accessing shdict. Only the next cache hit after a miss will fetch from shdict.
The way currently being adopted is exactly the opposite.

bzp2010: cc @membphis WDYT?

shreemaan-abhishek: > The way currently being adopted is exactly the opposite.

I don't think so, the lrucache usage helps cache the the json decode operation.

If we use an lrucache to cache the node for a given service name we don't have a good way to invalidate the lrucache when service data changes. Since privileged_agent is the only one that fetches data from consul a shdict is the only way to let other workers know that consul data has changed. Hence, we would still need to retrieve data from shdict if we want to use lrucache like you said.

bzp2010: What I'm referring to is a unidirectional process where each layer relies on its own TTL to control data expiration. Lower levels (privileged process, shdict) do not proactively invalidate higher-level (lrucache) caches. As long as the TTL is sufficiently small, changes take effect quickly. This process is entirely controlled by hard-coded values or configuration settings.

On the fast path, JSON decoding never occurs because we never fetch the raw JSON-encoded text from the shdict each time. We always use the already-cached table.

If we set the lrucache TTL to 1s, service node updates will take effect within 1 second. During this 1-second cache period, shdict is not read, eliminating lock contention. Moreover, since the data is read and parsed from shdict only once, JSON decoding does not occur (this is also an advantage of the worker-level LRU cache, whose scope is confined to a single worker. Consequently, there is no concurrency in populating the LRU cache, thereby saving the decoding overhead for each service_name). Of course, we can also set it to 0.1 seconds or 60 seconds—it's a balance between response speed and overhead.

Therefore, I have never considered JSON decoding to be an issue.

Issue #12240, opened 2025-05-21
bug: ext-plugin process rewrite:args logic error


wanglihui:
### Current Behavior

when i use go runner to write rewrite plugin, if i did't return new args, the original args will lost.
Example:
http://example.org/api/test?x=1&y=2 use go-plugin-runner rewrite the path to /api/test2 the questring args x,y will lost.

### Expected Behavior

expect the querystring args operate like the lua proxy-rewrite plugin process args. combine the original querystring to rewrited upstream.
var.upstream_uri = path + args...

maybe modify the code logic ext-plugin/init.lua could fix it. 
now code logic
```
            local len = rewrite:ArgsLength()
            if len > 0 then
                local changed = {}
                for i = 1, len do
                    local entry = rewrite:Args(i)
                    local name = entry:Name()
                    local value = entry:Value()
                    if value == nil then
                        args[name] = nil

                    else
                        if changed[name] then
                            if type(args[name]) == "table" then
                                core.table.insert(args[name], value)
                            else
                                args[name] = {args[name], entry:Value()}
                            end
                        else
                            args[name] = entry:Value()
                        end

                        changed[name] = true
                    end
                end

                core.request.set_uri_args(ctx, args)

                if path then
                    var.upstream_uri = path .. '?' .. var.args
                end
            end
```
modify logic, chang the combine path and args to `if` outline:
 ```
 local len = rewrite:ArgsLength()
            if len > 0 then
                local changed = {}
                for i = 1, len do
                    local entry = rewrite:Args(i)
                    local name = entry:Name()
                    local value = entry:Value()
                    if value == nil then
                        args[name] = nil

                    else
                        if changed[name] then
                            if type(args[name]) == "table" then
                                core.table.insert(args[name], value)
                            else
                                args[name] = {args[name], entry:Value()}
                            end
                        else
                            args[name] = entry:Value()
                        end

                        changed[name] = true
                    end
                end
            end
            core.request.set_uri_args(ctx, args)

            if path then
                    var.upstream_uri = path .. '?' .. var.args
            end
```

### Steps to Reproduce

1. write a go plugin, use RequestFilter rewrite the request path
2. the upstream can't recevie args, the args will lost
3. example like https://example.org/api/test?x=1&y=2  the plugin rewrite the path r.SetPath([]byte("/api/test2"))

### Environment

- APISIX version (run `apisix version`):
- Operating system (run `uname -a`):
[...]

PR #13080, opened 2026-03-09
fix: preserve uri args if path has not been modified

shreemaan-abhishek:
### Description

when the path is rewritten but no args are modified, we still need to append the original query string to upstream_uri.
Fixes https://github.com/apache/apisix/issues/12240

### Checklist

### DIFF
--- a/apisix/plugins/ext-plugin/init.lua
+++ b/apisix/plugins/ext-plugin/init.lua
@@ -709,11 +709,7 @@ local rpc_handlers = {
             end
 
             if path then
-                if var.args and var.args ~= "" and not core.string.find(path, "?") then
-                    var.upstream_uri = path .. "?" .. var.args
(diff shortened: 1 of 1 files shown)


apisix/plugins/ext-plugin/init.lua
```diff
@@ -706,9 +706,13 @@ local rpc_handlers = {
                 end
 
                 core.request.set_uri_args(ctx, args)
+            end
 
-                if path then
-                    var.upstream_uri = path .. '?' .. var.args
+            if path then
+                if var.args and var.args ~= "" and not core.string.find(path, "?") then
```
membphis: can we use this variable `var.is_args`?

https://nginx.org/en/docs/http/ngx_http_core_module.html#var_is_args

PR #13081, opened 2026-03-10
fix(openidc): include request port in `redirect_uri`

### LINKED ISSUE #12970: bug: APISIX 3.14.0+ OpenID Connect redirect drops request port (regression vs 3.13.0)
### Current Behavior

Starting from Apache APISIX 3.14.0, the openid-connect plugin generates redirect / callback URLs without the original request port, even when the client accesses APISIX directly (no reverse proxy or load balancer in front).

The same configuration works correctly on APISIX 3.13.0, where the redirect URL preserves the port.

This is a regression and breaks OIDC flows when APISIX is exposed on a non-default port (anything other than 80/443).

### Example

```bash
$ curl -v http://localhost:8888 2>&1 | grep -i '< Location'
< Location: http://10.126.166.49:8888/auth/realms/warehouse-automation/protocol/openid-connect/auth?scope=openid%20profile&client_id=apisix&redirect_uri=http%3A%2F%2Flocalhost%2F.apisix%2Fredirect&state=f44c18832740fc4486c099a6be0c1329&response_type=code
```

As you can see the redirect_uri parameter is missing the original request port

### Expected Behavior

```bash
$ curl -v http://localhost:8888 2>&1 | grep -i '< Location'
< Location: http://10.126.166.49:8888/auth/realms/warehouse-automation/protocol/openid-connect/auth?state=d15ffd15ae044c3d0f1c0583ce9f203e&response_type=code&scope=openid%20profile&redirect_uri=http%3A%2F%2Flocalhost%3A8888%2F.apisix%2Fredirect&client_id=apisix
```

### Error Logs

_No response_

### Steps to Reproduce

1. Run APISIX 3.14.0+ on port 8888 and configure OIDC plugin
2. Try to make a curl to a protected route

### Environment

- APISIX: 3.14.1-debian

shreemaan-abhishek:
### Description

This PR: https://github.com/apache/apisix/pull/12551, introduced a change that would overwrite X-Forwarded-Host with ctx.var.host which is just the hostname without the port.

When lua-resty-openidc sees a redirect_uri starting with /, it tries to reconstruct the full URL by checking headers in this order:

  1. Forwarded header → usually not set
  2. X-Forwarded-Host → now set to just localhost (no port) by handle_x_forwarded_headers
  3. ngx.var.http_host → never reached

Before v3.14.0, X-Forwarded-Host wasn't forcibly overwritten, so openidc would fall through to ngx.var.http_host
  which includes the port (e.g., localhost:8888), and everything worked.
#### Fix:
  Instead of handing a relative path to lua-resty-openidc and letting it reconstruct the URL (where it can lose the
  port), we now build the full absolute redirect_uri directly in the plugin using ctx.var.http_host. This is the
  original Host header from the client, which preserves the port as the client sees it critical for Docker port
  mappings where the external port differs from the container port.

Fixes #12970

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -20,7 +20,6 @@ local ngx_re            = require("ngx.re")
 local openidc           = require("resty.openidc")
 local fetch_secrets     = require("apisix.secret").fetch_secrets
 local jsonschema        = require('jsonschema')
-local trusted_addr      = require("apisix.utils.trusted-addresses")
 local string            = string
 local ngx               = ngx
 local ipairs            = ipairs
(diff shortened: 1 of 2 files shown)


Baoyuantop:
Hi @shreemaan-abhishek, could you describe the root cause of the problem and the current solution?

shreemaan-abhishek:
@Baoyuantop, I have added that in the description.

apisix/plugins/openid-connect.lua
```diff
@@ -622,6 +623,73 @@ local function validate_claims_in_oidcauth_response(resp, conf)
...
+    local forwarded = ctx.var.http_forwarded
+    if not forwarded then
+        return nil
+    end
+    -- take only the first proxy entry (before any comma)
+    local first = forwarded:match("^([^,]+)")
+    if not first then
+        return nil
+    end
+    for part in first:gmatch("[^;]+") do
```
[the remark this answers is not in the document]
shreemaan-abhishek: we actually need client values so this can be ignored.

apisix/plugins/openid-connect.lua
```diff
@@ -622,6 +623,88 @@ local function validate_claims_in_oidcauth_response(resp, conf)
...
+            redirect_path = string.sub(uri, 1, -2) .. suffix
+        else
+            redirect_path = uri .. suffix
+        end
+    end
+
+    local scheme
+    local host
+
+    if trusted_addr.is_trusted(ctx.var.realip_remote_addr) then
```
Baoyuantop: Can we directly use ctx.var.http_x_forwarded_host, ctx.var.http_x_forwarded_proto, and ctx.var.http_x_forwarded_port (which have already been correctly handled in handle_x_forwarded_headers) without having to repeatedly check the trusted/untrusted branch?

shreemaan-abhishek: I have followed the method used by lua-resty-openidc library, i.e try to extract from `Forward` header first, then look at xfh otherwise fallback to `http_host` value. You can find that here:

https://github.com/zmartzone/lua-resty-openidc/blob/fabf64e29062f5e7658de1cee52f572221203b1e/lib/resty/openidc.lua#L271-L275

nic-6443: I think this logic is duplicated with `handle_x_forwarded_headers`. For the current issue, I believe we should fix it in `handle_x_forwarded_headers` by adding the port to `X-Forwarded-Host`. The port can be omitted for 80/443, while other non-standard ports should be included in `X-Forwarded-Host`. I've checked the relevant information, and this is the conventional practice.

nic-6443: Otherwise, we need to figure out how to correctly obtain the external access address of the gateway for every function involving the acquisition of the gateway access address, and this is exactly what the `X-Forwarded-Host` is for. We should fix the issue at the root cause instead of making fixes wherever the `X-Forwarded-Host` is used.

apisix/plugins/openid-connect.lua
```diff
@@ -622,6 +622,7 @@ local function validate_claims_in_oidcauth_response(resp, conf)
     return core.schema.check(conf.claim_schema, data)
 end
 
+
```
membphis, line 625: useless?

Issue #12953, opened 2026-01-29
bug: Prometheus reports LLM metrics even when LLM plugin is not enabled


qiqizjl:
## Current Behavior

When the LLM plugin is **not enabled**, the Prometheus exporter still reports LLM metrics (`apisix_llm_latency`, `apisix_llm_prompt_tokens`, `apisix_llm_completion_tokens`) with value `0`.

This happens because:
1. The nginx variables (`llm_time_to_first_token`, `llm_prompt_tokens`, `llm_completion_tokens`) have a default value of `"0"` (set in `apisix/cli/ngx_tpl.lua`)
2. The condition check in `apisix/plugins/prometheus/exporter.lua` (L376-L394) only checks if the value is not empty (`~= ""`)

As a result, all requests (even non-LLM requests) generate LLM metrics with `0` values, polluting the metrics data.

## Expected Behavior

LLM metrics should **only** be reported when the LLM plugin is actually used and processing requests.

### Solution 1: Change default value from `"0"` to `""`
**Pros:** Simple change
**Cons:** May affect users' existing JSON log format (as mentioned in #12841)

### Solution 2: Check `~= "0"` instead of `~= ""`
**Pros:** No impact on log format
**Cons:** If an LLM request fails or returns 0 tokens (e.g., error response), the request won't be counted in metrics. This loses visibility into failed/abnormal LLM requests.

### Solution 3: Add a dedicated flag variable (e.g., `llm_request_active`)
Set this flag to `"1"` only when LLM plugin processes the request, then check this flag before reporting metrics.
**Pros:** Accurate detection, no side effects
**Cons:** Requires adding a new nginx variable

### Solution 4: Check `request_type` variable
Only report LLM metrics when `request_type` is `"ai_stream"` or `"ai_chat"`.
**Pros:** Uses existing variable
**Cons:** Need to verify if `request_type` is reliably set

I would appreciate guidance on the preferred approach before submitting a PR.

Related PR: #12841

## Steps to Reproduce

1. Start APISIX with Prometheus plugin enabled (without any LLM plugin configured)
2. Send a normal HTTP request to any route
3. Check Prometheus metrics endpoint
4. Observe that LLM metrics are reported with value `0`

## Environment

- APISIX version: master branch
- Related code:
  - `apisix/cli/ngx_tpl.lua` (default variable values)
  - `apisix/plugins/prometheus/exporter.lua` (L376-L394)

Baoyuantop:
[...]

PR #13088, opened 2026-03-13
fix: remove llm related metrics for non llm requests

shreemaan-abhishek:
Fixes https://github.com/apache/apisix/issues/12953

### Checklist

### DIFF
--- a/t/plugin/prometheus-ai-proxy2.t
+++ b/t/plugin/prometheus-ai-proxy2.t
@@ -218,27 +218,3 @@ hello world
 GET /t
 --- response_body
 success
-
-
-
-=== TEST 4: send a chat request authenticated as test-consumer
---- request
-POST /chat
-{"messages":[{"role":"user","content":"What is 1+1?"}], "model": "gpt-3"}
---- more_headers
-apikey: test-key-for-prometheus
---- error_code: 200
-
-
-
-=== TEST 5: assert consumer label is suppressed in llm metrics but visible in http metrics
---- request
-GET /apisix/prometheus/metrics
---- response_body_like eval
-qr/apisix_http_latency_bucket\{.*consumer="test-consumer".*\}/
---- response_body_like eval
-qr/apisix_llm_prompt_tokens\{.*consumer="".*\}/
---- response_body_like eval
-qr/apisix_llm_completion_tokens\{.*consumer="".*\}/
---- response_body_like eval
-qr/apisix_llm_latency_count\{.*consumer="".*\}/


moonming (commented):
Hi @shreemaan-abhishek, thank you for fixing the LLM metrics for non-LLM requests (#12953)!

Removing LLM-related metrics (token counts, etc.) from non-LLM requests is the correct behavior — these metrics should only be populated for AI/LLM plugin routes. I'll review the code in detail shortly.

apisix/plugins/prometheus/exporter.lua
```diff
@@ -355,24 +355,26 @@ function _M.http_log(conf, ctx)
...
-            unpack(extra_labels("llm_prompt_tokens", ctx))))
-    end
-    if vars.llm_completion_tokens ~= "" then
-        metrics.llm_completion_tokens:inc(tonumber(vars.llm_completion_tokens),
-            gen_arr(route_id, service_id, consumer_name, balancer_ip,
-            vars.request_type, vars.request_llm_model, vars.llm_model,
-            unpack(extra_labels("llm_completion_tokens", ctx))))
+    if vars.request_type == "ai_stream" or vars.request_type == "ai_chat" then
+        local llm_time_to_first_token = vars.llm_time_to_first_token
+        if llm_time_to_first_token ~= "" then
```
nic-6443: It seems that these judgments about token variables are incorrect; they will never be equal to an empty string:
https://github.com/apache/apisix/blob/4990927937280037602e81bb1b9554a784afa076/apisix/cli/ngx_tpl.lua#L816-L819

t/plugin/prometheus-ai-proxy2.t
```diff
@@ -0,0 +1,138 @@
...
+
+            if not metric_resp then
+                ngx.say("failed to fetch metrics")
+                return
+            end
+
+            local body = metric_resp.body
+            local has_error = false
+            for line in string.gmatch(body, "[^\r\n]+") do
+                ngx.log(ngx.WARN, "dibag: ", line)
```
nic-6443: This is a temporary debug code line that should be removed.

t/plugin/prometheus-ai-proxy2.t
```diff
@@ -0,0 +1,138 @@
...
+            local data = {
+                {
+                    url = "/apisix/admin/routes/2",
+                    data = [[{
+                        "plugins": {
+                            "prometheus": {},
+                            "serverless-pre-function": {
+                                "phase": "rewrite",
+                                "functions": [
+                                    "return function(conf, ctx) ngx.var.llm_time_to_first_token = '100'; ngx.var.llm_prompt_tokens = '10'; ngx.var.llm_completion_tokens = '20' end"
```
nic-6443: I think this test case is incorrect. In fact, only AI requests will record variables such as `llm_time_to_first_token`. We don't need to simulate scenarios where non-AI requests have the `llm_time_to_first_token` variable; we only need to confirm that the metrics of ordinary HTTP requests do not contain labels related to `llm_*`.

Baoyuantop (approved):

Baoyuantop (approved):

PR #13091, opened 2026-03-17
fix: preserve config.yaml when auto-generating admin key

### LINKED ISSUE #12170: bug: APISIX rewrites yaml configuration file to cause restarting errors
### Current Behavior

APISIX generates and writes a configuration file when the Admin API key is empty, which causes the configuration file to be exported from memory and encoded in YAML and then written. 
https://github.com/apache/apisix/blob/master/apisix/core/id.lua#L124-L133
During this process, comments in the configuration file will be lost, and some will be generated containing `...` ending, which will result in an error on the next startup.

### Expected Behavior

Configuration files should not be rewritten in any way. If the Admin API key is empty, we need to write it to the logs and also warn the user there that you need to configure your own key, otherwise every reboot will cause the key to change.

### Error Logs

No need

### Steps to Reproduce

1. Set `deployment.admin.admin_key` in the configuration file to the empty string
2. Start APISIX
3. Check config.yaml

### Environment

- Operating system (run `uname -a`): unrelated
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): unrelated
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): unrelated
- APISIX Dashboard version, if relevant: unrelated
- Plugin runner version, for issues related to plugin runners: unrelated
- LuaRocks version, for installation issues (run `luarocks --version`): unrelated

janiussyafiq:
### Description

This PR fixes a bug where APISIX rewrites the entire `conf/config.yaml` file when `admin_key` is empty, causing all user comments and formatting to be permanently stripped from the configuration file.

Fixes #12170 

#### What was the problem                                                                                                                     
When deployment.admin.admin_key is left empty, APISIX auto-generates a random key and writes it back to `conf/config.yaml` using `lyaml.dump()`. This serialization process operates on the fully merged in-memory config table and has no awareness of the original file causing:
- All comments stripped — lyaml.dump() serializes a Lua table, which has no concept of YAML comments. All inline documentation in the original file is permanently lost on first boot
- Config file bloated — the written-back file contains the full merge of config.yaml and system defaults, replacing the user's minimal config with a large auto-generated one
- The core issue is that APISIX should never destructively overwrite a user-managed configuration file.                                    
                  
#### What was changed                                                                                                                              
Instead of serializing and rewriting the entire config file, the fix performs a targeted string substitution — only replacing the empty `key: ''` value in the raw file content with the generated key, leaving everything else untouched.
                                                                                                                                           
Key changes in apisix/core/id.lua:                                                                                                       
- Removed the `generate_yaml()` call and full file rewrite from `_M.init()`
- Modified `autogenerate_admin_key()` to return the generated key value as a third return                                                  
- Added a targeted `gsub` to substitute only the key field in the raw file
- Added a log.warn() that surfaces the generated key so users are aware and can set a permanent one                                      

If the key field is not found in the raw file (e.g. it came from merged defaults and was never explicitly in config.yaml), APISIX logs the generated key and instructs the user to set it manually — no file write is attempted.

### Checklist

### DIFF
--- a/t/cli/test_admin.sh
+++ b/t/cli/test_admin.sh
@@ -32,6 +32,10 @@ deployment:
         admin_api_mtls:
             admin_ssl_cert: '../t/certs/apisix_admin_ssl.crt'
             admin_ssl_cert_key: '../t/certs/apisix_admin_ssl.key'
+        admin_key:
+            - name: admin
+              key: ''
+              role: admin
 " > conf/config.yaml
 
 make init
@@ -64,6 +68,10 @@ deployment:
     admin_listen:
       ip: 127.0.0.2
       port: 9181
+    admin_key:
+      - name: admin
+        key: ""
+        role: admin
 ' > conf/config.yaml
 
 make init
@@ -276,6 +284,10 @@ deployment:
   admin:
     admin_listen:
       port: 9180
+    admin_key:
+      - name: admin
+        key: ""
+        role: admin
 ' > conf/config.yaml
 
 rm logs/error.log
@@ -321,6 +333,12 @@ echo "passed: Admin API can only be used with etcd config_provider"
 echo '
 apisix:
   enable_admin: false
+deployment:
+  admin:
+    admin_key:
+      - name: admin
+        key: ""
+        role: admin
 ' > conf/config.yaml
 
 rm logs/error.log
@@ -351,6 +369,12 @@ plugins:
   - node-status
 nginx_config:
   error_log_level:  info
+deployment:
+  admin:
+    admin_key:
+      - name: admin
+        key: ""
+        role: admin
 ' > conf/config.yaml
 
 rm logs/error.log
@@ -416,6 +440,12 @@ echo "pass: ignore changes to /apisix/plugins/ due to init_etcd successfully"
 
 # accept changes to /apisix/plugins when enable_admin is false
 echo '
+deployment:
+  admin:
+    admin_key:
+      - name: admin
+        key: ""
+        role: admin
 apisix:
   enable_admin: false
 plugins:
--- a/t/cli/test_admin_key.sh
+++ b/t/cli/test_admin_key.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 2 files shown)


t/cli/test_admin.sh
```diff
@@ -32,6 +32,10 @@ deployment:
         admin_api_mtls:
             admin_ssl_cert: '../t/certs/apisix_admin_ssl.crt'
             admin_ssl_cert_key: '../t/certs/apisix_admin_ssl.key'
+        admin_key:
+            - name: admin
+              key: ''
+              role: admin
```
janiussyafiq, line 38: @Baoyuantop need advice on this, my current implementation requires that the config.yaml file to have this block in order to substitute empty string on key with the auto_generated admin_key since previously the entire config file would get rewritten from mem. Hence, this broke many cli test cases since the call to search the admin_key failed:
`admin_key=$(yq '.deployment.admin.admin_key[0].key' conf/config.yaml | sed 's/"//g')`
-- would result in null if the block was not declared.
In your opinion, do i need to make the large refactor towards affected test cases or would there be any alternative to this?

Baoyuantop: I've discovered a contradiction in the current solution: gsub's targeted replacement requires the file to contain the key `key: ''` to work, but in many scenarios, the user's (or test's) `config.yaml` doesn't have this field; the key comes from the merged `config-default.yaml`.

This means the gsub solution has an implicit prerequisite: the user must explicitly declare an empty `admin_key` in `config.yaml`; otherwise, the automatically generated key only exists in memory and cannot be written back to the file. This prerequisite holds true for the default `conf/config.yaml` (which already has `key: ''`), but it doesn't hold true for user-defined, streamlined config files.

janiussyafiq: image
This is what causes it, which is the main root cause for this issue since it rewrites the whole config file with the in-mem config, when the `changed = true` where we auto-generate the password. is this intended? i removed it (same as the previous PR approach) to avoid config file getting rewritten, causing it to break lots of test cases. this is the tradeoff that needs to be considered since the expected behavior requires that the config file should not be rewritten in any way. however, things to note is that the first-time user would still be able to run apisix without configuring anything on config file.

PR #13096, opened 2026-03-18
fix(ai-prompt-template): fix malformed JSON error message

sihyeonn:
### Description

The error message in `get_request_body_table()` had two issues:

1. **Grammatical error**: `"could not get parse"` instead of `"could not parse"`
2. **String concatenation bug**: `err` was added as a separate table element (`{ message = "...: ", err }`) instead of being concatenated to the message string (`{ message = "...: " .. err }`), causing the error detail to be silently dropped from the response.

The correct pattern already exists in `ai-prompt-decorator.lua:76`.

### Before

```lua
return nil, { message = "could not get parse JSON request body: ", err }
```

### After

```lua
return nil, { message = "could not parse JSON request body: " .. err }
```

### DIFF
--- a/apisix/core/request.lua
+++ b/apisix/core/request.lua
@@ -343,7 +343,7 @@ function _M.get_json_request_body_table()
 
     local body_tab, err = json.decode(body)
     if not body_tab then
-        return nil, { message = "could not get parse JSON request body: " .. err }
+        return nil, { message = "could not parse JSON request body: " .. err }
     end
 
     return body_tab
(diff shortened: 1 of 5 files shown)


Baoyuantop:
Please update the code in all relevant places.

Baoyuantop:
Hi @sihyeonn,  following up on the previous review comments. Please let us know if you have any updates. Thank you.

sihyeonn:
Hi, thanks for the follow-up. The fix has been applied — the error message now correctly reflects the malformed JSON case. Ready for review whenever you get a chance.

Baoyuantop:
Hi @sihyeonn, I mean, there are several similar errors; could you fix them all at once?

nic-6443 (approved):

Baoyuantop (approved):

shreemaan-abhishek (approved):

PR #13097, opened 2026-03-18
fix(ai-request-rewrite): return 400 when request body is missing

sihyeonn:
### Description

In the `access` handler of `ai-request-rewrite`, when `core.request.get_body()` returns `nil` without an error (i.e., the body is simply empty), the function returns without a status code. This causes the request to continue through the pipeline, where downstream code expects `client_request_body` to be present and fails with confusing errors.

The fix returns `HTTP_BAD_REQUEST` (400) to reject the request early, which is consistent with the error handling path directly above it (lines 190-193).

### Before

```lua
if not client_request_body then
    core.log.warn("missing request body")
    return
end
```

### After

```lua
if not client_request_body then
    core.log.warn("missing request body")
    return HTTP_BAD_REQUEST
end
```

### DIFF
--- a/apisix/plugins/ai-request-rewrite.lua
+++ b/apisix/plugins/ai-request-rewrite.lua
@@ -179,7 +179,7 @@ function _M.access(conf, ctx)
 
     if not client_request_body then
         core.log.warn("missing request body")
-        return
+        return HTTP_BAD_REQUEST
     end
 
     -- Determine provider protocol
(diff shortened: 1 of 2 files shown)


Baoyuantop:
This part comes from https://github.com/apache/apisix/pull/12036. @LiteSun  Can you confirm this?

Baoyuantop (commented):
Please add a test for this change.

Baoyuantop:
Hi @sihyeonn, there are some errors in the tests that aren't related to the PR. Could you merge the main branch?

nic-6443 (approved):

Baoyuantop (approved):

shreemaan-abhishek (approved):

PR #13099, opened 2026-03-18
fix(core): use CSPRNG for admin key auto-generation

### LINKED ISSUE #13092: feat: As a user, I want admin key auto-generation to use a cryptographically secure random source, so that auto-generated keys are not predictable or brute-forceable
### Description

The current implementation of admin key auto-generation in `apisix/core/id.lua` uses `math.random()` to generate the key:
https://github.com/apache/apisix/blob/4990927937280037602e81bb1b9554a784afa076/apisix/core/id.lua#L108-L112

#### This has two security concerns:
1. `math.random` is a predictable PRNG                                                                                                     
`math.random` is not a cryptographically secure pseudo-random number generator (CSPRNG). It is seeded predictably, meaning an attacker with knowledge of the seed or timing information could reconstruct or predict the generated key.                                             
                  
2. Limited character set reduces entropy
The generated key only uses A–Z and a–z (52 characters), giving approximately 5.7 bits of entropy per character.
Over 32 characters this yields ~182 bits of theoretical entropy, but the effective entropy is significantly lower due to the use of a predictable PRNG.          
   
#### Suggested improvement:                                                                                                                   
Both lua-resty-random and lua-resty-string are already available in the APISIX codebase (used in `apisix/patch.lua` and various plugins). A more secure implementation could leverage these existing dependencies:                                                                  
```lua         
local resty_random = require("resty.random")
local resty_str    = require("resty.string")
local key = resty_str.to_hex(resty_random.bytes(16)) -- 32 hex chars, 128 bits of true entropy                                           
```                                                                                                  
`resty.random.bytes()` is backed by OpenSSL's RAND_bytes, making it a proper CSPRNG. Combined with hex encoding via `resty.string.to_hex()`, this produces a 32-character alphanumeric key with 128 bits of true entropy — a significant improvement over the current approach with no new dependencies required.                                                                                                                              

This issue was identified during the fix for #12170  . Since it is a separate concern, it is being tracked here independently.

Credits: inspired by @fekitibi from https://github.com/apache/apisix/issues/12170#issuecomment-3242146618

sihyeonn:
### Description

Closes #13092

The `autogenerate_admin_key()` function in `apisix/core/id.lua` used `math.random()` to generate admin API keys character by character. `math.random()` relies on a Lua PRNG that is predictable if the seed can be inferred, which is a security concern for authentication keys.

### Changes

- Replace `math.random()` with `resty.random.bytes(32, true)` which calls OpenSSL `RAND_bytes` for cryptographically secure randomness
- Falls back to `resty.random.bytes(32)` (non-strong) if the entropy pool is insufficient
- Hex-encode the output via `resty.string.to_hex()`, producing a 64-character key

Both `resty.random` and `resty.string` are already available in the OpenResty distribution and are used elsewhere in APISIX (e.g., `apisix/patch.lua`, `apisix/plugins/csrf.lua`).

### DIFF
--- a/apisix/core/id.lua
+++ b/apisix/core/id.lua
@@ -31,7 +31,6 @@ local smatch           = string.match
 local open             = io.open
 local type             = type
 local ipairs           = ipairs
-local string           = string
 local prefix           = ngx.config.prefix()
 local pairs            = pairs
 local ngx_exit         = ngx.exit
--- a/apisix/core/id.lua
+++ b/apisix/core/id.lua
@@ -25,12 +25,13 @@ local profile          = require("apisix.core.profile")
 local log              = require("apisix.core.log")
 local uuid             = require("resty.jit-uuid")
 local lyaml            = require("lyaml")
+local resty_random     = require("resty.random")
+local resty_str        = require("resty.string")
 local smatch           = string.match
 local open             = io.open
 local type             = type
 local ipairs           = ipairs
 local string           = string
-local math             = math
 local prefix           = ngx.config.prefix()
 local pairs            = pairs
 local ngx_exit         = ngx.exit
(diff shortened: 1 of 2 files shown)


Baoyuantop (commented):
Please add a test case for this PR.

Baoyuantop:
Hi @sihyeonn, following up on the previous review comments. Please let us know if you have any updates. Thank you.

Baoyuantop (approved):

Issue #13101, opened 2026-03-18
bug: ai-proxy-multi health checker creation always fails in timer context due to missing _dns_value


Baoyuantop:
### Description

The `ai-proxy-multi` plugin's health check mechanism has a structural bug: the `construct_upstream` function is called from both request context and timer context (`healthcheck_manager`), but only works correctly in request context.

In timer context, `construct_upstream` always returns `nil` because the `_dns_value` runtime field does not exist on instance configs read from etcd. This causes a Lua runtime error on the subsequent line (`upstream.resource_key`), breaking health check management.

**There are two crash points** — both lack nil checks on the `construct_upstream` return value:

1. **`timer_create_checker`** (line 180): Crashes when creating new checkers if `_dns_value` is missing. In practice, the first creation usually succeeds because the request thread sets `_dns_value` on the shared config cache object before the timer fires.

2. **`timer_working_pool_check`** (line 242/238): Crashes when validating existing checkers after a **config update**. When the route config is updated via Admin API, etcd watch triggers a config cache refresh with a new in-memory object that does NOT have `_dns_value`. The timer calls `construct_upstream` with this fresh config object and crashes. **This is the primary reproducible crash point.**

### Reproduction Steps

**Environment**: Docker (tested on APISIX 3.15.0 and 3.16.0), etcd 3.5

**Step 1**: Create a route with `ai-proxy-multi` and health checks enabled:
```bash
curl http://127.0.0.1:9180/apisix/admin/routes/1 \
  -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' \
  -X PUT -d '
{
  "uri": "/ai/*",
  "plugins": {
    "ai-proxy-multi": {
      "instances": [
        {
          "name": "healthy-backend",
          "provider": "openai",
          "weight": 1,
          "auth": {"header": {"Authorization": "Bearer test-key"}},
          "endpoint": "http://<healthy-host>:80/v1",
          "checks": {
            "active": {
              "type": "http",
              "http_path": "/",
              "healthy": {"interval": 1, "successes": 1},
              "unhealthy": {"interval": 1, "http_failures": 1}
            }
          }
        },
        {
          "name": "unhealthy-backend",
          "provider": "openai",
          "weight": 1,
          "auth": {"header": {"Authorization": "Bearer test-key"}},
          "endpoint": "http://<unhealthy-host>:80/v1",
          "checks": {
            "active": {
              "type": "http",
              "http_path": "/",
[...]

PR #13136, opened 2026-03-31
fix: ai-proxy-multi health checker creation always fails in timer context due to missing _dns_value #13101

guilongyang:
[##] Description

hi, @Baoyuantop .   Based on the 3.15.0-ubuntu Docker image, I have verified the changes in this PR and confirmed that the issue has been resolved.

Fixes # https://github.com/apache/apisix/issues/13101

### Checklist


Baoyuantop:
The current PR does not fix the root cause as described in the original issue. Please modify the technical solution.

Baoyuantop:
Please revise your technical solution and reopen the PR. Thank you for your contribution.

guilongyang:
sorry for late.
# Technical Solution: Fix ai-proxy-multi Health Check Mechanism

## Root Cause

The `ai-proxy-multi` plugin's health check mechanism has a structural bug:
1. `construct_upstream` in timer context (reading from etcd) cannot access the `_dns_value` field
2. This field is only set by `resolve_endpoint` in request context
3. When health checker creation fails, the resource is permanently removed from `waiting_pool`, causing it to never be retried
4. Original code: `fetch_checker` returns nil, resource is removed from waiting_pool but never retried

---

## Solution

### I. Changes to `healthcheck_manager.lua`

#### 1. New SHM Proxy Checker (Lines 127-174)

```lua
-- Create a proxy checker object that reads health status directly from SHM.
-- This is used when the checker is not available in the current worker's local working_pool.
local function create_shm_proxy_checker(resource_path)
    -- Returns a proxy object that reads health status directly from shared memory
    -- Uses internal key format of lua-resty-healthcheck
end

function _M.get_target_status_by_path(resource_path, ip, port, hostname)
    -- Query health status based on resource_path
end
```

**Purpose**:
- Returns a proxy checker instead of nil when the checker is not yet created
- Proxy reads health status directly from SHM, avoiding caller blocking while waiting

#### 2. Modified `fetch_checker` Function (Lines 103-113)

```lua
-- Check if this is a known waiting resource with matching version
if waiting_pool[resource_path] == resource_ver then
    -- A checker is being created, but we need immediate response
    -- Return a proxy that reads from SHM while waiting
    return create_shm_proxy_checker(resource_path)
end

-- First time requesting this resource, add to waiting pool
-- Don't return anything useful yet - the real checker is being created
waiting_pool[resource_path] = resource_ver
return nil
```

**Purpose**:
- When resource is in waiting_pool, subsequent requests can immediately read status from SHM
- Prevents callers from blocking due to checker not being created yet

#### 3. Added Error Logging (Lines 293-298)

```lua
upstream = plugin.construct_upstream(upstream_constructor_config)
if not upstream then
    core.log.error("[healthcheck] default upstream is nil | resource_path: ", resource_path, 
                   "upstream_constructor_config: ", core.json.encode(upstream_constructor_config))
end
```

---

### II. Changes to `ai-proxy-multi.lua`

#### 1. Modified `fetch_health_instances` Function (Lines 236-282)

```lua
local function fetch_health_instances(conf, checkers)
    local instances = conf.instances
    local new_instances = core.table.new(0, #instances)

    for i, ins in ipairs(instances) do
        local checker = checkers and checkers[ins.name]
        if checker then
            -- Checker exists: get health status via fetch_node_status
            local node = ins._dns_value
            local ok, err = healthcheck_manager.fetch_node_status(checker, node.host, port or node.port, host)
            if ok then
                transform_instances(new_instances, ins)
            end
        elseif checker == nil then
            -- Checker is being created: read health status from SHM
            local node = ins._dns_value
            local idx = ins._index or (i - 1)
            local resource_path = conf._meta.parent.resource_key ..
                                  "#plugins['ai-proxy-multi'].instances[" .. idx .. "]"
            local ok = healthcheck_manager.get_target_status_by_path(resource_path, node.host, port or node.port, host)
            if ok then
                transform_instances(new_instances, ins)
            end
        else
            -- No checker configured: include instance by default
            transform_instances(new_instances, ins)
        end
    end
    -- ...
end
```

**Three states handled**:
| State | Handling Method |
|-------|-----------------|
| Checker created | `fetch_node_status` to get status |
| Checker being created | `get_target_status_by_path` from SHM |
| No checker configured | Include by default |

#### 2. Modified `get_checkers_status_ver` Function (Lines 223-232)

```lua
local function get_checkers_status_ver(checkers)
    local status_ver_total = 0
    for _, checker in pairs(checkers) do
        -- Only count status_ver from real checkers (proxy checkers don't have this attribute)
        if checker.status_ver then
            status_ver_total = status_ver_total + checker.status_ver
        end
    end
    return status_ver_total
end
```

**Reason**: Proxy checker doesn't have `status_ver` attribute, needs type check

#### 3. Modified `pick_target` Function (Lines 356-359)

```lua
instances[i]._dns_value = instance._dns_value
instances[i]._nodes_ver = instance._nodes_ver
instances[i]._index = i - 1  -- Store index for later use
```

**Key points**:
- `_dns_value`: Syncs node info computed in request context to config object
- `_index`: Stores instance array index for building correct resource_path

---

## Core Mechanism: Why `_dns_value` Can Be Correctly Accessed in Timer Context

```
Request Context:
  resolve_endpoint(instance) → Sets instance._dns_value
  ↓
  instances[i]._dns_value = instance._dns_value  [Syncs to etcd copy]
  ↓
  res_conf.value.plugins['ai-proxy-multi'].instances[i]._dns_value is set

Timer Context:
  res_conf.value fetched from etcd
  ↓
  jp.value(res_conf.value, json_path) extracts instance config
  ↓
  upstream_constructor_config._dns_value exists ← because it was synced
  ↓
  construct_upstream works correctly
```

**Key Understanding**:
- On first request, `_dns_value` is computed by `resolve_endpoint` and set
- Immediately afterward, `instances[i]._dns_value = instance._dns_value` **writes the value to the config object**
- When timer triggers, the config read from etcd **already contains `_dns_value`**
- The entire flow is "**lazy initialization + state persistence**", not recalculation each time

---

## Summary

| Problem | Solution |
|---------|----------|
| `_dns_value` missing in timer context | Sync computed value to config object in request context, timer reads the synced value |
| Health checker lost after creation failure | Add SHM proxy, can read status even if checker not created |
| Caller blocks waiting for checker | Returns proxy checker immediately usable |
| `status_ver` attribute missing | Add type check to skip proxy checker |

guilongyang:
> Please revise your technical solution and reopen the PR. Thank you for your contribution.

I don't have permission to reopen the PR. What should I do next?

Baoyuantop:
Hi @guilongyang, I've further refined the relevant code

guilongyang:
> Hi @guilongyang, I've further refined the relevant code

I fixed two test cases

nic-6443:
The ai-proxy-multi half of this PR is superseded on master: #13322 made `construct_upstream` fall back to `resolve_endpoint` when the runtime DNS field is nil, and #13441 reworked the whole DNS/health-check path (`_dns_value` is gone, replaced by the plural `_dns_nodes` model), so this diff no longer applies.

The healthcheck_manager nil-check part is still worth having though — master still indexes the `construct_upstream` return value without a nil check (`healthcheck_manager.lua` lines 180 and 238), which can produce the per-second error loop when an instance is removed from the route while its checker is in the working pool. I'd suggest rebasing this down to just the healthcheck_manager hardening, plus a test if feasible.

PR #13142, opened 2026-04-01
feat: Add debug logging for OPA responses

vdegans:
## Description

This PR introduces an optional debug logging configuration for Open Policy Agent (OPA) responses to improve observability during development and troubleshooting.

---

### When `debug` is enabled in apisix:

* Logs the raw OPA response, including HTTP status and body.
* Logs the evaluated decision result (`allow`) along with the full decoded response payload.
* Logging is performed using `core.log.debug`, ensuring no impact on production environments unless explicitly enabled.

---

## Motivation

OPA policy evaluation can be difficult to debug due to limited visibility into the responses returned by the OPA server. This change provides developers with detailed insights into:

* The raw response returned by OPA
* The parsed decision output used by APISIX

This significantly improves the ability to diagnose misconfigurations, policy errors, or unexpected authorization behavior during local development.

---

## Checklist

* [x] I have explained the need for this PR and the problem it solves
* [x] I have explained the changes or the new features added to this PR
* [ ] I have added tests corresponding to this change
* [x] I have updated the documentation to reflect this change
* [x] I have verified that this change is backward compatible (If not, please discuss on the APISIX mailing list first)

### DIFF
--- a/apisix/plugins/opa.lua
+++ b/apisix/plugins/opa.lua
@@ -51,6 +51,11 @@ local schema = {
         with_route = {type = "boolean", default = false},
         with_service = {type = "boolean", default = false},
         with_consumer = {type = "boolean", default = false},
+        debug = {
+            type = "boolean",
+            default = false,
+            description = "enable debug logging of OPA responses (only for local development)"
+        },
     },
     required = {"host", "policy"}
 }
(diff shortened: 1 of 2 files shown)


apisix/plugins/opa.lua
```diff
@@ -51,6 +51,11 @@ local schema = {
         with_route = {type = "boolean", default = false},
         with_service = {type = "boolean", default = false},
         with_consumer = {type = "boolean", default = false},
+        debug = {
```
Baoyuantop: I don't think this configuration option is necessary. When the log level is set to debug, debug logs will already be printed.

vdegans: I have removed the config to let the apisix config take over this functionality

PR #13146, opened 2026-04-02
fix(opentelemetry): coerce additional_attributes values to string

### LINKED ISSUE #9921: help request: OpenTelemetry plugin has problems handling Nginx variables of non-string type
### Description

Hello,

I'm encountering an issue when using the OpenTelemetry plugin with Apache APISIX. I want to add some Nginx variables (such as `$request_time` and `$upstream_response_time`) as additional attributes in my trace data. 

Here is my configuration:

```json
{
  "plugins": {
      "opentelemetry": {
          "sampler": {
              "name": "always_on"
          },
          "additional_attributes": [
              "upstream_response_time",
              "request_time",
              "host"
          ]
      }
  }
}
```
However, I get the following error when APISIX tries to process a request:
```
2023/07/27 20:31:33 [error] 50#50: *127339 lua entry thread aborted: runtime error: ...deps/share/lua/5.1/opentelemetry/trace/exporter/otlp.lua:67: bad argument #2 to 'encode' (string expected for field 'string_value', got number)
```
It seems that OpenTelemetry's OTLP exporter expects all attribute values to be strings, but the values of Nginx variables like $request_time and $upstream_response_time are not strings.

Is there a recommended way to convert these non-string values to strings before adding them as additional attributes? Or could the OpenTelemetry plugin be modified to automatically handle non-string attribute values?

Thank you for your help.

### Environment

- Operating system (run `uname -a`): Linux apisix-test-6cd7f9b885-cpq8b 5.15.0-60-generic
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx version: openresty/1.21.4.1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): "etcd:3.5.4-debian-11-r14" image in k8s
- LuaRocks version, for installation issues (run `luarocks --version`): 3.8.0

janiussyafiq:
### Description

When using the `opentelemetry` plugin with `additional_attributes` that reference Nginx variables returning numeric values (e.g. `request_time`, `upstream_response_time`, `bytes_sent`), APISIX crashes with the following error during span export:
```
lua entry thread aborted: runtime error: .../opentelemetry/trace/exporter/otlp.lua: bad argument #2 to 'encode' (string expected for field 'string_value', got number)
```
                                                                                                                                           
This happens because `inject_attributes()` always calls `attr.string(key, val)` regardless of the actual Lua type of the value. The `resty.ngxvar` library returns Lua numbers for certain Nginx variables, and the OTLP protobuf encoder strictly expects a string for `string_value`, causing it to abort.

The fix wraps all values with `tostring()` before passing them to `attr.string()`. This is correct because Nginx variables are conceptually always strings — `tostring()` simply restores that invariant that `resty.ngxvar` breaks for numeric variables.

Fixes #9921

### Checklist

### DIFF
--- a/t/plugin/opentelemetry.t
+++ b/t/plugin/opentelemetry.t
@@ -480,8 +480,6 @@ GET /opentracing
 --- wait: 2
 --- response_body
 opentracing
---- no_error_log
-bad argument #2 to 'encode'
 
 
 
@@ -489,4 +487,4 @@ bad argument #2 to 'encode'
 --- exec
 tail -n 1 ci/pod/otelcol-contrib/data-otlp.json
 --- response_body eval
-qr/.*\/opentracing.*/
+qr/.*opentelemetry-lua.*/


shreemaan-abhishek (approved):

t/plugin/opentelemetry.t
```diff
@@ -434,3 +434,59 @@ HEAD /specific_status
...
+
+
+=== TEST 21: trigger opentelemetry, numeric nginx variables must not cause encode error
+--- request
+GET /opentracing
+--- wait: 2
+--- response_body
+opentracing
+--- no_error_log
+bad argument #2 to 'encode'
```
nic-6443: Remove this specific error message and simply assert that there are no error logs at all. Because if a specific error message is set, problems may fail to be detected due to changes in the error message.

janiussyafiq: good catch! fixed

PR #13147, opened 2026-04-02
fix(env): fix os.getenv prefix collision bug in init_by_lua phase

### LINKED ISSUE #13055: bug: Environment variable with common prefix returns incorrect value in init_by_lua
### Current Behavior

**Description:**  
When reading environment variables in the `init_by_lua` phase, if two variables share a common prefix (e.g., `KUBERNETES_CLIENT_TOKEN` and `KUBERNETES_CLIENT_TOKEN_FILE`), the value of the shorter-named variable may be erroneously returned when the longer-named one is requested. The issue appears to be sensitive to the order in which the variables are defined in the configuration.

### Expected Behavior

**Expected behavior:**  
`os.getenv("KUBERNETES_CLIENT_TOKEN_FILE")` should always return the value of the environment variable named exactly `KUBERNETES_CLIENT_TOKEN_FILE`, regardless of the definition order of the `env` directives.

### Error Logs

_No response_

### Steps to Reproduce

**Steps to reproduce:**

1. Create a test configuration similar to `t/kubernetes/discovery/kubernetes2.t` that defines two environment variables with a common prefix, e.g.:
   ```
   env KUBERNETES_CLIENT_TOKEN=some-token;
   env KUBERNETES_CLIENT_TOKEN_FILE=/path/to/token;
   ```
2. In `init_by_lua`, attempt to read `KUBERNETES_CLIENT_TOKEN_FILE` using `os.getenv("KUBERNETES_CLIENT_TOKEN_FILE")`.
3. Observe that the returned value is `some-token` (the value of the shorter variable) instead of `/path/to/token`.
4. Swap the order of the two `env` directives so that the longer name (`KUBERNETES_CLIENT_TOKEN_FILE`) appears first:
   ```
   env KUBERNETES_CLIENT_TOKEN_FILE=/path/to/token;
   env KUBERNETES_CLIENT_TOKEN=some-token;
   ```
5. In `init_by_lua`, again read `KUBERNETES_CLIENT_TOKEN_FILE` – this time it returns the correct value `/path/to/token`.

### Environment

master branch

currenjin:
## Problem

In OpenResty's `init_by_lua` phase, `os.getenv()` uses nginx's internal environment mechanism which performs prefix-based matching when `env NAME=VALUE` directives share a common prefix. This causes `os.getenv()` to return the shorter-named variable's value when the longer-named one is requested.

Example: with both `KUBERNETES_CLIENT_TOKEN` and `KUBERNETES_CLIENT_TOKEN_FILE` defined as `env NAME=VALUE;` directives, calling `os.getenv("KUBERNETES_CLIENT_TOKEN_FILE")` returns `KUBERNETES_CLIENT_TOKEN`'s value when the shorter name appears first in the config.

Fixes #13055

## Root cause

OpenResty's `os.getenv` in `init_by_lua` does not perform exact-match lookup when `env NAME=VALUE;` directives have a common prefix — the shorter name is matched first. In contrast, reading `ffi.C.environ` directly with Lua string operations (as `_M.init()` already does) produces a correctly keyed table.

## Fix

1. **`apisix/core/env.lua`**: After `_M.init()` populates `apisix_env_vars` from `ffi.C.environ`, override `os.getenv` with a wrapper that does exact table lookup. Falls back to the original `os.getenv` for variables set dynamically after startup (e.g., via `core.os.setenv`).

2. **`apisix/init.lua`**: Move `core.env.init()` to be the **first** call in `http_init()`, ensuring the override is applied before any other code in `init_by_lua` can call `os.getenv`.

## Tests

- **TEST 10**: Verifies `os.getenv` returns correct values when shorter-prefix variable is declared first in nginx config.
- **TEST 11**: Verifies `os.getenv` called inside `init_by_lua` (via `extra_init_by_lua`) returns correct values for prefix-colliding variables.

## Risk / Compatibility

- `os.getenv` is monkey-patched globally after `_M.init()`. Existing callers continue to work — the semantics are identical for non-colliding variable names.
- Variables set dynamically via `core.os.setenv` fall back to the original `os.getenv`, so those are unaffected.
- The ordering change in `http_init` is safe: `core.env.init()` has no dependencies on `resolver` or `id` initialization.

### DIFF
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -26,6 +26,7 @@ local upper       = string.upper
 local find        = string.find
 local sub         = string.sub
 local str         = ffi.string
+local os_getenv   = os.getenv
 
 local ENV_PREFIX = "$ENV://"
 
@@ -57,6 +58,19 @@ function _M.init()
         end
 
         i = i + 1
+    end
+
+    -- Override os.getenv to use exact lookup from apisix_env_vars, avoiding
+    -- the prefix collision bug in OpenResty's init_by_lua phase where
+    -- env NAME=VALUE directives with a common prefix cause os.getenv to
+    -- return the shorter-named variable's value for the longer-named one.
+    os.getenv = function(name)
+        local val = apisix_env_vars[name]
+        if val ~= nil then
+            return val
+        end
+        -- Fall back for vars set dynamically after startup (e.g., via core.os.setenv)
+        return os_getenv(name)
     end
 end
 
--- a/apisix/core/env.lua
+++ b/apisix/core/env.lua
@@ -64,7 +64,7 @@ function _M.init()
     -- the prefix collision bug in OpenResty's init_by_lua phase where
     -- env NAME=VALUE directives with a common prefix cause os.getenv to
     -- return the shorter-named variable's value for the longer-named one.
-    os.getenv = function(name)
+    os.getenv = function(name) -- luacheck: ignore
         local val = apisix_env_vars[name]
         if val ~= nil then
             return val
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -89,9 +89,9 @@ local _M = {version = 0.4}
 
 
 function _M.http_init(args)
+    core.env.init()
     core.resolver.init_resolver(args)
     core.id.init()
-    core.env.init()
 
     local process = require("ngx.process")
     local ok, err = process.enable_privileged_agent()
--- a/t/core/env.t
+++ b/t/core/env.t
@@ -179,3 +179,47 @@ env ngx_env=apisix-nice;
 GET /t
 --- response_body
 apisix-nice
+
+
+
+=== TEST 10: os.getenv prefix collision - shorter prefix first should not shadow longer name
+--- main_config
+env TEST_PREFIX=prefix-value;
+env TEST_PREFIX_LONGER=longer-value;
+--- config
+    location /t {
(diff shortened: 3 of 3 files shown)


Baoyuantop:
Hi @currenjin, please fix the lint error

currenjin:
Fixed the lint error by adding `-- luacheck: ignore` to suppress the read-only field warning on the `os.getenv` override.

nic-6443:
I looked into the underlying issue (#13055) and I don't think this approach can work, unfortunately. The collision happens in init_by_lua, where lua-resty-core has replaced `os.getenv` with a shim backed by `ngx_http_lua_ffi_get_conf_env` — and at that point nginx hasn't applied `env` directives to the real environ yet (that happens in worker init), so the `ffi.C.environ` snapshot this PR builds doesn't contain `env NAME=VALUE` directive vars at all. The lookup misses and falls back to the captured `os.getenv`, which is the buggy shim itself — I verified this by simulating the override on openresty 1.29.2.4, and it matches the CI failures here.

The real bug is in lua-nginx-module's `ngx_http_lua_ffi_get_conf_env`: it matches entries with `ngx_strncmp(name, var[i].data, var[i].len)` without checking that the queried name ends at `var[i].len`, so the first prefix-colliding entry wins. That needs an exact-length check upstream (plus a runtime patch until it's released); an APISIX-side Lua fix can detect the wrong entry but can't reach the correct one, since the FFI only returns the first match.

Issue #13163, opened 2026-04-06
bug: proxy-rewrite silently drops duplicate keys in headers.add due to JSON object limitation


janiussyafiq:
### Current Behavior

When configuring `headers.add` in the `proxy-rewrite` plugin with multiple values for the same header name, only the last value is forwarded to upstream. All earlier values are silently dropped with no warning, error log, or validation message.

This is caused by a fundamental JSON limitation: duplicate keys in an object are undefined behavior by spec. In practice, Lua's cjson parser (used by APISIX/OpenResty) only retains the last occurrence of a duplicate key when deserializing the plugin config.

For example, the following config:
```json
"headers": {
  "add": {
    "foo": "bar2",
    "foo": "bar3"
  }
}
```
Results in only foo: bar3 being added. The value bar2 is lost entirely at config parsing time, before the plugin even executes.

### Expected Behavior

Either:
1. The schema should support multi-value headers via an array syntax, e.g. "foo": ["bar1", "bar2"], so users can express the intent without relying on duplicate JSON keys.
2. Or, the admin API should reject configs with duplicate keys and return a validation error, rather than silently accepting and dropping values.

### Error Logs

No error logs are produced.

### Steps to Reproduce

1. Create a route with the following config:

```bash
curl -X PUT 'http://127.0.0.1:9180/apisix/admin/routes/1' \
  -H 'X-API-KEY: <admin_key>' \
  -H 'Content-Type: application/json' \
  -d '{
    "uri": "/anything",
    "plugins": {
      "proxy-rewrite": {
        "headers": {
          "add": {
            "foo": "bar2",
            "foo": "bar3"
          }
        }
      }
    },
    "upstream": {
      "type": "roundrobin",
      "nodes": { "httpbin.org:80": 1 }
    }
  }'
```

2. Send a test request:

```bash
curl -X GET -H 'foo: bar1' 'http://127.0.0.1:9080/anything'
```

3. Observe the echoed headers in the response — only foo: bar3 is present. foo: bar2 is missing with no error.

```bash
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 423
Connection: keep-alive
Date: Mon, 06 Apr 2026 04:09:23 GMT
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Server: APISIX/3.15.0

{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
[...]

PR #13164, opened 2026-04-06
fix(proxy-rewrite): correct header operation order to remove > set > add

### LINKED ISSUE #11587: bug: proxy-rewrite header priority wrong
### Current Behavior

The current priority for header operations is: `add > remove > set`. This makes it impossible to handle the use-case where you want to add multiple headers with the same name but first remove all the headers with that name. The following configuration would make sense in my opinion:

```
"proxy-rewrite": {
   "headers": {
       "remove: ["foo"]
       "add": {
              "foo": "bar1",
              "foo": "bar2",
        }
    }
}
```
But due to the header priority `add > remove > set` this leads to no foo headers, as first they are added and then all removed again.

### Expected Behavior

The expected behaviour would be to have both foo headers added and all earlier foo headers removed. For this to work the priority order should be `remove > set > add`. The `set` operation is basically a shortcut for `remove, add` but is confusing when you need a header multiple times, which set does not allow.

### Error Logs

_No response_

### Steps to Reproduce

1. Run APISIX via docker image
2. Create a route:
```
{
  "uri": "/*",
  "name": "httpbin-demo",
  "host": "httpbin.org",
  "upstream": {
    "scheme": "https",
    "type": "roundrobin",
    "nodes": {
      "httpbin.org:443": 1
    }
  },
  "plugins": {
    "proxy-rewrite": {
      "headers": {
        "remove": ["foo"],
        "add": {
          "foo": "bar2",
          "foo": "bar3"
        }
      }
    }
  }
}
```
3. Test using `curl -H"foo: bar1" -H"Host: httpbin.org" http://localhost:9080/anything

The resulting json is missing the expected `foo: bar1` and `foo: bar2` headers.

### Environment

- APISIX version (run `apisix version`): Stock docker container with version 3.10.0
- Operating system (run `uname -a`): Not relevant, but Ubuntu 24.04 LTS
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): Not relevant
- APISIX Dashboard version, if relevant: Not relevant
- Plugin runner version, for issues related to plugin runners: Not relevant
- LuaRocks version, for installation issues (run `luarocks --version`): Not relevant

janiussyafiq:
### Description

The `proxy-rewrite` plugin executed header operations in the wrong order: `add → set → remove`. This made it impossible to remove existing upstream headers and re-add them with new values in the same plugin config — the `remove` would always run last and wipe out anything added.

This PR fixes the execution order to `remove → set → add`, which is the only order that makes semantic sense: clear first, overwrite next, then append new values.

  **Example that now works correctly:**
  ```json
  "proxy-rewrite": {
      "headers": {
          "remove": ["foo"],
          "add": {"foo": "bar2"}
      }
  }
```
Previously, sending a request with foo: bar1 would result in no foo header forwarded to upstream. With this fix, foo: bar2 is correctly forwarded.

#### Additional Notes
- added a new test file for proxy-rewrite since `proxy-rewrite3.t` reached ~1000 lines.
- updated a previous test case to reflect the changes made

Fixes #11587

### Checklist

### CODE REVIEW

File: t/plugin/proxy-rewrite3.t [hunk 1]
```diff
@@ -454,7 +454,7 @@ passed
 --- request
 GET /echo HTTP/1.1
 --- response_headers
-test: test_in_set
+test: test_in_set, test_in_add
```
shreemaan-abhishek, line 457: why this change?

t/plugin/proxy-rewrite4.t
```diff
@@ -0,0 +1,87 @@
...
+            local code, body = t('/apisix/admin/routes/1',
+                ngx.HTTP_PUT,
+                [[{
+                    "methods": ["GET"],
+                    "plugins": {
+                        "proxy-rewrite": {
+                            "headers": {
+                                "remove": ["foo"],
+                                "add": {
+                                    "foo": "bar2"
```
shreemaan-abhishek, line 55: I don't understand the significance of this PR, if the user want's to remove "foo" and then add "foo", they can simply use "set", no?

janiussyafiq: yes, that is right but according to #11587 it fails the use case where user wants to add multiple headers since `set` can only handle a single value. the more appropriate test case would be:
```
"headers": {
        "remove": ["foo"],
        "add": {
          "foo": "bar2",
          "foo": "bar3"
        }
      }
```
desired behavior is to remove all foo headers then add `foo: bar2` and `foo: bar3` but with the current order, foo headers will get wiped out entirely. but there is a current caveat i just explained in a new issue where only `foo: bar3` would get added: see #13163

t/plugin/proxy-rewrite3.t
janiussyafiq: with the current header operation order `remove -> set -> add`, both headers would be in the expected result, see https://github.com/apache/apisix/pull/13164/changes#diff-794cd8c7c23c7d0356c4a6f2892b7995a8892cf1fd19966eb2ac65336e9b61feL427-L428

PR #13165, opened 2026-04-06
feat(plugin): add dpop plugin for RFC 9449 proof-of-possession

### LINKED ISSUE #11219: feat: support DPoP in OpenID/OAuth2 plugin
### Description

Demonstrating Proof of Possession (DPoP, RFC9449) is a new proposed standard (https://datatracker.ietf.org/doc/html/rfc9449) for OAuth2 that constraints who can use the access token for when calling an OAuth2-protected endpoint. It's an important standard for improving the security of OAuth2 tokens when migrating to from mTLS to OAuth2 deployments.

It works by the client sharing a JWK thumbprint when requesting an access token. The client then signs a DPoP proof (containing request URL and HTTP method) and provides it as HTTP header when accessing the OAuth2-protected API.
The resource server (or APISIX in this case) then uses the `cnf` claim from the OAuth2 Token Introspection response to verify the proof signature. It can then reassure the OAuth2 access token wasn't stolen by a MITM, but is in fact used by the very same party that requested the token. Usage is not enforced by the server, but up to the OAuth2 client to enable.

Are there plans to support DPoP?

Alnyli07:
## Description

Add `dpop` plugin for [RFC 9449 DPoP (Demonstrating Proof of Possession)](https://datatracker.ietf.org/doc/html/rfc9449) validation. This plugin enforces sender-constrained access tokens at the API gateway level, preventing token theft and replay attacks with zero changes to upstream services.

Fixes https://github.com/apache/apisix/issues/11219

### Why a Separate Plugin?

DPoP is a **proof-of-possession layer**, not an authentication mechanism. It complements existing auth plugins (`openid-connect`, `jwt-auth`, `multi-auth`) rather than replacing them. It runs at priority **2601** to rewrite `Authorization: DPoP` → `Authorization: Bearer` before downstream auth plugins process the token.

### Features

- Full RFC 9449 DPoP proof validation (`typ`, `alg`, `htm`, `htu`, `iat`, `ath`, `jti`)
- DPoP proof signature verification (ES256/384/512, RS256/384/512, PS256/384/512)
- JWK Thumbprint binding (`cnf.jkt`) per RFC 7638
- Access token signature verification via JWKS (OIDC discovery or direct URI)
- Token introspection fallback (RFC 7662) for opaque tokens
- Distributed JTI replay protection (memory / Redis / Infinispan, configurable fallback strategy)
- Bearer downgrade detection
- `DPoP` → `Bearer` header rewrite + `DPoP-Thumbprint` header injection
- Selective enforcement via `uri_allow`
- Structured audit logging with OpenTelemetry trace_id correlation

A Go WASM version (proxy-wasm-go-sdk v0.24.0) is also available at [Keymate-io/keymate-apisix-dpop-plugin](https://github.com/Keymate-io/keymate-apisix-dpop-plugin).

### Files Changed

| File | Description |
|------|-------------|
| `apisix/plugins/dpop.lua` | Plugin implementation (1384 LOC) |
| `conf/config.yaml.example` | Plugin registration (priority 2601) |
| `docs/en/latest/plugins/dpop.md` | English documentation |
| `t/plugin/dpop.t` | 15 Test::Nginx test cases |

### Tested With

- Apache APISIX 3.11.0
- Keycloak 26.0 (any RFC 9449-compliant IdP)
- 58 E2E checks (k6), 49 Postman requests in our upstream repo
- Dependencies: only libraries already in APISIX (`resty.openssl`, `resty.http`, `resty.redis`, `resty.lrucache`)

## Checklist

### DIFF
--- a/apisix/plugins/dpop.lua
+++ b/apisix/plugins/dpop.lua
@@ -1,3 +1,19 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
 local core = require("apisix.core")
 local cjson = require("cjson.safe")
 local resty_sha256 = require("resty.sha256")
@@ -5,6 +21,13 @@ local http = require("resty.http")
 local openssl_pkey = require("resty.openssl.pkey")
 local lrucache = require("resty.lrucache")
 local ngx = ngx
+local pairs = pairs
+local ipairs = ipairs
+local tostring = tostring
+local type = type
+local string = string
+local math = math
+local require = require
 
 local plugin_name = "dpop"
 
@@ -209,6 +232,10 @@ local schema = {
         token_issuer = {
             type = "string",
             default = "",
+        },
+        ssl_verify = {
+            type = "boolean",
+            default = true,
         },
     },
     additionalProperties = false,
@@ -454,7 +481,10 @@ local function resolve_jwks_uri(conf)
 
     local httpc = http.new()
     httpc:set_timeout(5000)
-    local res, err = httpc:request_uri(conf.discovery, { method = "GET", ssl_verify = false })
+    local res, err = httpc:request_uri(conf.discovery, {
+        method = "GET",
+        ssl_verify = conf.ssl_verify,
+    })
     if not res then
         return nil, "discovery fetch failed: " .. (err or "unknown")
     end
@@ -480,10 +510,13 @@ local function resolve_jwks_uri(conf)
 end
 
 
-local function fetch_jwks(uri)
+local function fetch_jwks(uri, ssl_verify)
     local httpc = http.new()
     httpc:set_timeout(5000)
-    local res, err = httpc:request_uri(uri, { method = "GET", ssl_verify = false })
+    local res, err = httpc:request_uri(uri, {
+        method = "GET",
+        ssl_verify = ssl_verify,
+    })
     if not res then
         return nil, "JWKS fetch failed: " .. (err or "unknown")
     end
@@ -535,7 +568,7 @@ local function get_jwk_for_kid(conf, kid)
         return nil, "kid '" .. kid .. "' not found in JWKS (rate limited, retry later)"
     end
 
-    local keys_by_kid, fetch_err = fetch_jwks(jwks_uri)
+    local keys_by_kid, fetch_err = fetch_jwks(jwks_uri, conf.ssl_verify)
     if not keys_by_kid then
         return nil, fetch_err
     end
@@ -857,9 +890,11 @@ local function jti_check(jti, conf)
             )
             if auth_hdr then headers["Authorization"] = auth_hdr end
         end
-        local res, err = httpc:request_uri(url,
-            { method = "POST", body = "1", headers = headers, ssl_verify = false }
-        )
+        local res, err = httpc:request_uri(url, {
+            method = "POST", body = "1",
+            headers = headers,
+            ssl_verify = conf.ssl_verify,
+        })
         -- Digest auth: on 401, parse nonce from WWW-Authenticate and retry
         if res and res.status == 401 and has_creds then
             local www_auth = res.headers
@@ -878,9 +913,11 @@ local function jti_check(jti, conf)
                     headers["Authorization"] = auth_hdr
                     httpc = http.new()
                     httpc:set_timeout(3000)
-                    res, err = httpc:request_uri(url,
-                        { method = "POST", body = "1", headers = headers, ssl_verify = false }
-                    )
+                    res, err = httpc:request_uri(url, {
+                        method = "POST", body = "1",
+                        headers = headers,
+                        ssl_verify = conf.ssl_verify,
+                    })
                 end
             end
         end
@@ -1007,7 +1044,7 @@ local function call_introspection(access_token, conf)
         method = "POST",
         body = body,
         headers = req_headers,
-        ssl_verify = false,
+        ssl_verify = conf.ssl_verify,
     })
     if not res then
         return nil, "introspection request failed: " .. (err or "unknown")
--- a/apisix/plugins/dpop.lua
+++ b/apisix/plugins/dpop.lua
@@ -1,0 +1,1384 @@
+local core = require("apisix.core")
+local cjson = require("cjson.safe")
+local resty_sha256 = require("resty.sha256")
+local http = require("resty.http")
+local openssl_pkey = require("resty.openssl.pkey")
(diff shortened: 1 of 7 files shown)


apisix/plugins/dpop.lua
```diff
@@ -0,0 +1,1384 @@
...
+    properties = {
+        allowed_algs = {
+            type = "array",
+            items = {
+                type = "string",
+                enum = {
+                    "ES256", "ES384", "ES512",
+                    "RS256", "RS384", "RS512",
+                    "PS256", "PS384", "PS512",
+                },
```
Baoyuantop, line 111: But does verify_dpop_proof_signature() only implement ES256 and RS256/RS384/RS512, a total of 4?

apisix/plugins/dpop.lua
```diff
@@ -0,0 +1,1384 @@
...
+    local now = ngx.time()
+    local cached = _discovery_cache[conf.discovery]
+    local ttl = conf.jwks_cache_ttl or 86400
+    if cached and (now - cached.fetched_at) < ttl then
+        return cached.jwks_uri
+    end
+
+    local httpc = http.new()
+    httpc:set_timeout(5000)
+    local res, err = httpc:request_uri(conf.discovery, { method = "GET", ssl_verify = false })
```
Baoyuantop: It is recommended to add the ssl_verify configuration item (default is true), or at least refer to the approach of the openid-connect plugin to allow users to control whether to skip TLS verification.

apisix/plugins/dpop.lua
Alnyli07: Thanks for flagging this.

Our primary test environment uses Keycloak which defaults to ES256 for DPoP                                                                      
proofs, and RS256/384/512 for access token signing — so those were the                                                                           
algorithms we had full E2E coverage for. The schema was populated from the                                                                       
full RFC 9449 recommended algorithm list for completeness, but the                                                                               
verification implementation didn't keep pace with the schema. That's on us.                                                                      
                                                                                                                                                  
We'll add ES384, ES512, PS256, PS384, and PS512 support — `resty.openssl.pkey`                                                                   
already handles these, it's mainly a matter of mapping the correct hash                                                                          
algorithm and padding parameters.                                                                                                                
                                                                                                                                                  
While we're at it — are there any other algorithms you'd like to see                                                                             
supported, or is the current RFC 9449 set (EC + RSA + RSA-PSS families)                                                                          
sufficient?

apisix/plugins/dpop.lua
Alnyli07: Valid point. This was left from our development setup and should definitely be configurable.

We'll add an `ssl_verify` attribute (boolean, default `true`) to the schema
and propagate it to all HTTP calls (discovery, JWKS, introspection). 
Consistent with how `openid-connect` handles it.

Alnyli07: Landed in **7b6c419**.  The hardcoded `false` was a leftover from local
development.

The schema gains an `ssl_verify` boolean (default `true`), propagated
to every outbound HTTP call the plugin issues:

- `resolve_jwks_uri` (OIDC discovery fetch)
- `fetch_jwks` (JWKS fetch; the function signature now takes
  `ssl_verify` so the call site can pass `conf.ssl_verify` through)
- `jti_check` (Infinispan replay-cache backend, including the
  digest-auth retry path)
- `call_introspection` (RFC 7662 introspection endpoint)

Default `true` keeps existing deployments behaving the same as before;
the opt-out is only available when operators set it explicitly. Same
shape as `openid-connect`'s `ssl_verify`.

apisix/plugins/dpop.lua
Alnyli07: **75f2d87** : the verification path now dispatches all nine asymmetric algs declared
in the schema: ES256/384/512, RS256/384/512, PS256/384/512.

Two implementation notes:

1. ECDSA proofs carry signatures in raw R||S (JWS Compact). Converting
   to the DER form OpenSSL expects needs long-form SEQUENCE length once
   R+S exceeds 127 bytes (every ES512 signature). The `raw_ecdsa_to_der`
   helper now emits long-form (`0x81`/`0x82`).
2. RSA-PSS verification under OpenSSL 3's provider API requires
   `RSA_PKCS1_PSS_PADDING` to be set before any saltlen/MGF1 parameter.
   The plugin now passes padding via the dedicated argument and uses
   the named `pss_saltlen` setter (digest length per RFC 7518 §3.5),
   so parameters are applied in the order the provider enforces.

**0028e83** : TEST 31 (ES512) and TEST 32-35 (RS384/RS512/PS384/PS512) cover the
newly-added algorithms. The full suite has 35 cases / 105 assertions
and passes against the standard upstream test runner.

Alnyli07:
@Baoyuantop both threads above are now addressed. Brief summary of what
landed on top of the initial PR:

- `ssl_verify` config option (default `true`), propagated to every
  outbound HTTP call the plugin issues.
- Verification path now covers all nine RFC 7518 asymmetric algorithms
  declared in the schema (ES256/384/512, RS256/384/512, PS256/384/512),
  with the long-form DER and OpenSSL 3 PSS adjustments noted in the
  thread.
- Test::Nginx suite covering schema validation, route enforcement, the
  full crypto flow per algorithm, and proof-claim negatives. 35 cases /
  105 assertions, all passing.

Alnyli07:
Hi everyone, 

Just bumping this PR. All feedback from @Baoyuantop has been fully addressed, and the 35 E2E test cases are passing cleanly.

Could we get additional reviews to help meet the 3-approval requirement? Thanks! 🙏

Baoyuantop:
Thanks for continuing to improve this DPoP plugin. The current revision does address the two earlier review points: `ssl_verify` is now configurable and propagated to outbound discovery/JWKS calls, and proof signature verification now covers the ES/RS/PS algorithm families with additional tests.

However, I do not think this is mergeable yet. There are a few blocking issues that should be fixed first:

1. A malformed proof without `alg` can hit a Lua runtime error. `verify_dpop_proof_signature` eventually concatenates the nil `alg` value when no algorithm branch matches, and `validate_proof` also calls `h.alg:sub(...)` directly. This kind of input should consistently return `invalid_dpop_proof`, not fall into a 500/error path. Please add explicit header structure validation and negative tests for missing/non-string `alg`.

2. `require_nonce` is accepted by the schema, while the documentation says it is “not yet implemented”, and the implementation does not validate a DPoP nonce. A security option should not be accepted as a no-op. Please either implement the nonce semantics with tests, or remove the option for now.

3. With the default `strict_htu=false`, the plugin only compares the path and ignores scheme/host/port. This compatibility mode can weaken the URI binding semantics from RFC 9449. Please re-evaluate the default behavior, or at least document it clearly and add tests for strict/full-URL and path-only boundaries.

Please address these points and request another review.

Alnyli07:
Thanks for the review again @Baoyuantop. All three points addressed:

1. Malformed proof / missing alg: Added header/payload structure validation that runs before signature verification and validate_proof, rejecting a missing/non-string alg (and malformed header/payload/jwk) with invalid_dpop_proof. The nil-concat and h.alg:sub() paths are now unreachable. Added negative tests for missing alg, non-string alg, and non-object header/payload (each expects 401, no error log).

2. **require_nonce**:  Removed from schema and docs, since the nonce semantics aren't implemented (no longer a no-op option).

3. **strict_htu** default Kept false because strict_htu=true requires public_base_url, so defaulting to true would break existing routes. The path-only behavior and its weakening of RFC 9449 §4.3 URI binding is now documented explicitly, with tests for full-URL match (200), host/scheme mismatch (401), and path-only default (200). Happy to flip the default if you'd prefer secure-by-default.

Ready for another look.

Alnyli07:
Hi @Baoyuantop and everyone! 👋 All review feedback has been addressed and all tests are passing. Could we get a couple more reviews to help move this forward? Thanks! 🙏

PR #13166, opened 2026-04-07
docs: fix misleading "types of Plugin" wording in lifecycle diagram

CrispOvO:
The diagram shows plugin execution **phases**, not "types" of plugins. Updated the description to avoid confusion.
  - Phases (init, rewrite, access, etc.) are lifecycle stages
  - Each plugin implements callbacks for the phases it needs
  - The diagram shows the order of phase execution, not plugin categories


Baoyuantop:
Hi @CrispOvO, for the document fixes, we need to commit to the main branch. I have already switched branches, please merge and resolve the conflicts.

PR #13167, opened 2026-04-07
docs: update ai-aliyun-content-moderation plugin documentation

Yilialinn:
## Summary

- Update EN and ZH plugin docs for `ai-aliyun-content-moderation` based on API7 docs source
- Add canonical link to API7 docs
- Add comprehensive attribute table with all fields from plugin Lua schema, including `keepalive`, `keepalive_pool`, `keepalive_timeout` fields
- Add detailed descriptions and chunking behavior for length limits
- Add two rich examples: Moderate Request Content Toxicity and Adjust Risk Level Threshold
- Replace hardcoded `ADMIN_API_KEY` with `admin_key` variable and add the note block
- Fix attribute types and defaults to match plugin source code (e.g. `deny_message` type was wrong in API7 config, `ssl_verify` is optional not required)
- Follow APISIX doc conventions (`True`/`False` for Required, lowercase types)


docs/en/latest/plugins/ai-aliyun-content-moderation.md
```diff
@@ -27,103 +30,646 @@ description: This document contains information about the Apache APISIX ai-aliyu
...
-| ssl_verify                   | No           | Boolean   | Enable SSL certificate verification. Default: `true`.                                                                                                                       |
-
-## Example usage
-
-First initialise these shell variables:
+The `ai-aliyun-content-moderation` Plugin integrates with [Aliyun Machine-Assisted Moderation Plus](https://help.aliyun.com/document_detail/2671445.html) to check request and response content for risk level when proxying to LLMs, such as profanity, hate speech, insult, harassment, violence, and more, rejecting requests if the evaluated outcome exceeds the configured threshold.
+
+Please ensure that the `access_key_secret` is correctly configured in the Plugin. If misconfigured, all requests will bypass the Plugin to be directly forwarded to the LLM Upstream, and you will see a `Specified signature is not matched with our calculation` error in the gateway's error log from the Plugin.
```
kayx23: The implementation does log the Aliyun signature error and then allows the request to continue when moderation fails, but this wording feels stronger than what the code actually guarantees. It is not really an intentional “bypass” mode tied specifically to `access_key_secret`; it is a moderation-call failure path. Suggest softening this to say the moderation check fails and the request may still be forwarded upstream, rather than stating that all requests will bypass the plugin directly.

Yilialinn: Fixed — softened the wording to say the moderation check fails and the request may still be forwarded upstream, rather than implying it is an intentional bypass mode.

kayx23 (approved):

moonming (approved):

shreemaan-abhishek (approved):

PR #13239, opened 2026-04-17
docs(opa): re-port with Admin API, ADC, and Ingress Controller tabs

Yilialinn:
Re-ports the opa plugin documentation with Admin API, ADC, and Ingress Controller tab structure, updated attribute table including send_headers_upstream, updated OPA image to 1.6.0, modern Rego syntax, and canonical link.


docs/en/latest/plugins/opa.md
```diff
@@ -109,219 +119,1267 @@ The JSON below shows the response from the OPA service to APISIX:
...
+```
 
-:::note
-You can fetch the `admin_key` from `config.yaml` and save to an environment variable with the following command:
+Send a GET request to the Route with the JWT of `john`:
 
-```bash
-admin_key=$(yq '.deployment.admin.admin_key[0].key' conf/config.yaml | sed 's/"//g')
+```shell
+curl -i "http://127.0.0.1:9080/get" -H "Authorization: ${john_jwt_token}"
```
kayx23: These verification requests seem to be missing the `Bearer ` prefix. `jwt-auth` expects the token in the `Authorization` header as a bearer token, so `Authorization: ${john_jwt_token}` / `${jane_jwt_token}` would fail as written. It should be `Authorization: Bearer ${...}` here (and in the later `jane` examples as well).

Yilialinn: Fixed — added the `Bearer ` prefix to all four JWT Authorization headers (john GET/POST and jane GET/POST) in both EN and ZH files.

kayx23 (approved):

moonming (approved):

shreemaan-abhishek (approved):

PR #13242, opened 2026-04-17
docs(hmac-auth): re-port with Admin API, ADC, and Ingress Controller tabs

Yilialinn:
## Summary

Re-ports the `hmac-auth` plugin documentation with the following updates:

- Added canonical link in `<head>` pointing to API7 docs
- Updated Consumer/Credential attribute table to 6-column format with `key_id` field (replacing incorrect `access_key`)
- Added ADC and Ingress Controller (Gateway API + APISIX CRD) tabs to all four examples
- Added separate Implementation section describing how HMAC verification works
- Fixed `X-Credential-Identifier` header name typo (`Indentifier` → `Identifier`)
- Updated admin_key retrieval path to `/usr/local/apisix/conf/config.yaml`
- Removed highlight annotations and enterprise-only content
- Updated Chinese translation with full tab structure and fixed field names


docs/en/latest/plugins/hmac-auth.md
kayx23, line 1: In the anonymous-consumer rate-limit example, the route is configured on `/get`, but the verification requests near the end use `/anything`. As written, those requests will miss the configured route, so the expected `200`/`429` result here will not match unless the request path is updated to `/get` (or the route is changed to `/anything`).

Yilialinn: Fixed. Both verification requests (the authenticated `john` request and the anonymous request) now use `/get` to match the configured route URI. Note the HMAC signature was already computed for `/get` (the Python script uses `request_path = "/get"`), so only the curl URLs needed updating. Same fix applied to the ZH doc.

kayx23 (approved):

moonming (approved):

shreemaan-abhishek (approved):

PR #10887, opened 2024-01-30
feat: support openresty-1.25.3.1

zll600:
### Description

Fixes https://github.com/apache/apisix/issues/6403

related PRs:
- https://github.com/api7/ngx_multi_upstream_module/pull/12
- https://github.com/api7/ngx_multi_upstream_module/pull/13
- https://github.com/api7/grpc-client-nginx-module/pull/73
- https://github.com/api7/wasm-nginx-module/pull/162
- https://github.com/api7/apisix-nginx-module/pull/83
- https://github.com/api7/apisix-build-tools/pull/367

### Checklist

- [ ] I have explained the need for this PR and the problem it solves
- [ ] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #13243, opened 2026-04-17
fix(ext-plugin-post-resp): avoid H2/H3 Content-Length check when read…

### LINKED ISSUE #13237: bug: 502 error when using ext-plugin-post-resp due to HTTP/2/HTTP/3 request without Content-Length header
### Current Behavior

 1. Enable http2 on the APISIX gateway listener.
  2. Configure a route or Ingress that uses the ext-plugin-post-resp plugin.
  3. Make sure the request reaches the ext-plugin-post-resp code path.
     For example, attach an ApisixPluginConfig with ext-plugin-post-resp enabled to an Ingress.
  4. Send an HTTP/2 GET request that has no request body and no Content-Length header.
Example:

  curl -k -i --http2 'https://(host)/xx/'

  5. Observe that APISIX returns 502.

  Relevant code path

  - apisix/plugins/ext-plugin-post-resp.lua
  - apisix/core/request.lua

  The failure is triggered by logic equivalent to:

  if (var.server_protocol == "HTTP/2.0" or var.server_protocol == "HTTP/3.0")
      and not content_length then
      return nil, "HTTP2/HTTP3 request without a Content-Length header"
  end
  

### Expected Behavior

The request should not fail.
  For an HTTP/2 or HTTP/3 GET request without a body, APISIX should treat the request body as empty and continue normally. Content-Length should not be required in this case.

### Error Logs

  [lua] ext-plugin-post-resp.lua:153: phase_func(): failed to request: HTTP2/HTTP3 request without a Content-Length header
  [lua] plugin.lua:1198: common_phase(): ext-plugin-post-resp exits with http status code 502

### Steps to Reproduce

Refer to the Current Behavior section

### Environment

Goend:
### Description

  This PR fixes an issue in `ext-plugin-post-resp` when handling HTTP/2 and HTTP/3 requests without a `Content-Length` header.

  Currently, `ext-plugin-post-resp` replays the original client request to upstream and reads the request body via `core.request.get_body()`. After the behavior introduced in #10887, `core.request.get_body()` rejects
  HTTP/2 and HTTP/3 requests that do not contain a `Content-Length` header.

  This causes valid body-less requests such as normal `GET` requests over HTTP/2 to fail unexpectedly with `502` when `ext-plugin-post-resp` is enabled.

  This PR avoids using `core.request.get_body()` in `ext-plugin-post-resp` and reads the request body directly from the Nginx request APIs instead. As a result, body-less HTTP/2 and HTTP/3 requests are handled
  correctly, while requests that actually contain a body can still be forwarded as expected.

  Fixes #13237

  ### Checklist

### DIFF
--- a/apisix/core/request.lua
+++ b/apisix/core/request.lua
@@ -290,15 +290,6 @@ function _M.get_body(max_size, ctx)
         end
     end
 
-    -- check content-length header for http2/http3
-    do
-        local var = ctx and ctx.var or ngx.var
-        local content_length = tonumber(var.http_content_length)
-        if (var.server_protocol == "HTTP/2.0" or var.server_protocol == "HTTP/3.0")
-            and not content_length then
-            return nil, "HTTP2/HTTP3 request without a Content-Length header"
-        end
-    end
     req_read_body()
 
     local req_body = req_get_body_data()
--- a/t/plugin/azure-functions.t
+++ b/t/plugin/azure-functions.t
@@ -501,10 +501,9 @@ passed
 
 
 
-=== TEST 15: http2 failed to check response body and headers
+=== TEST 15: http2 success to without Content-Length
 --- http2
 --- request
 GET /azure
---- error_code: 400
---- error_log
-HTTP2/HTTP3 request without a Content-Length header,
+--- response_body
+faas invoked


apisix/core/request.lua
```diff
@@ -290,15 +290,6 @@ function _M.get_body(max_size, ctx)
         end
     end
 
-    -- check content-length header for http2/http3
```
Baoyuantop, line 293: This modification is too extensive, we recommend:

1. **Narrow the fix to `ext-plugin-post-resp.lua`**: bypass `core.request.get_body()` and use `ngx.req.read_body()` + `ngx.req.get_body_data()` directly for the upstream replay request
2. **Or refine the check in `get_body()`**: skip the Content-Length check for request methods that typically have no body (GET, HEAD, DELETE, OPTIONS), while keeping it for POST/PUT/PATCH
3. **Add targeted tests**: add H2 test cases specifically for `ext-plugin-post-resp` to cover the reported scenario, and verify no regression in other H2-dependent plugins (e.g., `grpc-web`, `batch-requests`)

Goend: @Baoyuantop Thanks for the review. I initially took the same approach of only bypassing the check in ext-plugin-post-resp, but after digging into the upstream lua-nginx-module, I believe removing the check from get_body() is actually the correct fix.                                                                                                                                                    This Content-Length validation was originally added to align with lua-nginx-module's documented limitation (see https://github.com/openresty/lua-nginx-module/pull/2237). However, upstream has since  relaxed this restriction: - https://github.com/openresty/lua-nginx-module/commit/e0d19f787e74ce3ffbb9b9c5abad02fd6ea81f41 partially lifted it - https://github.com/openresty/lua-nginx-module/commit/6e29c1a96e641ea3e3498b1524baeaa28d3ab58c further relaxed it, and the docs no longer state that ngx.req.read_body() requires Content-Length for  H2/H3                                                                                                                                                                                                 Since the underlying layer no longer enforces this restriction, keeping the check in get_body() makes APISIX stricter than the runtime — it blocks valid H2/H3 requests across all 20+ plugins that    call get_body(), not just ext-plugin-post-resp. Removing it here is not "too extensive" — it's removing a guard that no longer serves its purpose.
 I'll add targeted H2 tests for ext-plugin-post-resp as suggested. Let me know your thoughts.

nic-6443:
The root cause was removed on master by #13428 — the H2/H3 Content-Length guard in `core.request.get_body()` is gone entirely, so ext-plugin-post-resp no longer 502s on HTTP/2 requests without a body, and a plugin-local `read_body()` workaround isn't needed anymore (it would also bypass `get_body()`'s max_size handling). I think this can be closed; the fix ships in the next release after 3.16.0.

AlinsRan:
Fixed: #13428. I will close this PR.

PR #13244, opened 2026-04-17
docs(forward-auth): re-port with Admin API, ADC, and Ingress Controller tabs

Yilialinn:
Re-ports the forward-auth plugin documentation with Admin API, ADC, and Ingress Controller tab structure, updated attribute table, and canonical link.


docs/en/latest/plugins/forward-auth.md
```diff
@@ -28,265 +28,1087 @@ description: This document contains information about the Apache APISIX forward-
...
-    "methods": ["GET"],
-    "uri": "/hello",
-    "plugins": {},
-    "upstream": {
-        "type": "roundrobin",
-        "nodes": {
-            "127.0.0.1:1980": 1
-        }
-    }
+  "tenant_id": "000"
```
kayx23: This failing example looks inconsistent with the mechanism used above. The route extracts `tenant_id` via `$post_arg.tenant_id`, which is resolved from form-style POST args, but this request sends a JSON body and does not set `Content-Type: application/x-www-form-urlencoded`. As written, the auth service would not receive `tenant_id=000`, so the example/result here is likely misleading unless this request is changed to a form body (or the docs switch to a JSON-compatible extraction path).

Yilialinn: The comment is accurate. Fixed: the failing example now sends `tenant_id=000` as a form-encoded body with `Content-Type: application/x-www-form-urlencoded`, consistent with the `$post_arg.tenant_id` extraction used in the route configuration. The same fix was applied to the ZH doc.

kayx23 (approved):

moonming (approved):

shreemaan-abhishek (approved):

PR #13248, opened 2026-04-17
docs(jwt-auth): re-port with Admin API, ADC, and Ingress Controller tabs

Yilialinn:
Re-ports the jwt-auth plugin documentation with Admin API, ADC, and Ingress Controller tab structure, updated attribute table, and canonical link.

### DIFF
--- a/docs/en/latest/plugins/jwt-auth.md
+++ b/docs/en/latest/plugins/jwt-auth.md
@@ -56,7 +56,6 @@ The following attributes are available for configurations on Consumers or Creden
 | exp | integer | False | 86400 | >=1 | Expiry time of the token in seconds. If you are not using APISIX to sign the JWT, this parameter is ignored and you should specify the expiration in the payload when signing the JWT. |
 | base64_secret | boolean | False | false | | Set to true if the secret is base64 encoded. |
 | lifetime_grace_period | integer | False | 0 | >=0 | Grace period in seconds. Used to account for clock skew between the server generating the JWT and the server validating the JWT. |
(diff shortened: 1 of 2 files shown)


docs/en/latest/plugins/jwt-auth.md
```diff
@@ -32,57 +32,59 @@ description: The jwt-auth Plugin supports the use of JSON Web Token (JWT) as a m
...
 The examples below demonstrate how you can work with the `jwt-auth` Plugin for different scenarios.
 
 :::note
 
-You can fetch the `admin_key` from `config.yaml` and save to an environment variable with the following command:
+You can fetch the `admin_key` from `/conf/config.yaml` and save to an environment variable with the following command:
 
 ```bash
-admin_key=$(yq '.deployment.admin.admin_key[0].key' conf/config.yaml | sed 's/"//g')
+admin_key=$(yq '.deployment.admin.admin_key[0].key' /conf/config.yaml | sed 's/"//g')
```
kayx23: `/conf/config.yaml` looks like a container-internal absolute path. In the rest of the APISIX docs, these setup snippets use `conf/config.yaml`, and using `/conf/...` on a normal checkout will fail because that file does not exist at the filesystem root. Could we switch this back to `conf/config.yaml` here (and in the mirrored examples) so the command works in local/source installs?

kayx23 (commented):
Found one user-facing issue in the updated examples.

kayx23 (approved):

guoqqqi (approved):

moonming (approved):

PR #13249, opened 2026-04-17
feat(ai-proxy): support aws bedrock

shreemaan-abhishek:
## Add Amazon Bedrock support to ai-proxy and ai-proxy-multi

Adds a new `bedrock` provider that proxies requests to Amazon Bedrock's [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html), with full AWS SigV4 request signing.

### What's new

**Provider & protocol:**
- New `bedrock` provider that signs requests with SigV4 and routes them to `bedrock-runtime.{region}.amazonaws.com/model/{modelId}/converse`
- New `bedrock-converse` protocol module that parses Bedrock's native message/content-block format
- Detection: requests with body containing `messages` array sent to a URI ending with `/converse`

**Authentication:**
- New `auth.aws` block with `access_key_id`, `secret_access_key` (required), and `session_token` (optional, for temporary credentials from STS / assumed roles)
- Sensitive fields (`secret_access_key`, `session_token`) are encrypted at the schema level
- All credential headers (including `x-amz-security-token`) are redacted from logs

**Region & endpoint:**
- `provider_conf.region` is required for `bedrock` (no env-var fallback or hardcoded default — fails fast on misconfiguration)
- `override.endpoint` supported for AWS PrivateLink, custom proxies, etc.

**Model identifiers (`options.model`):**
- Foundation model IDs (e.g., `anthropic.claude-3-5-sonnet-20240620-v1:0`)
- Cross-region inference profile IDs (e.g., `us.anthropic.claude-3-5-sonnet-20240620-v1:0`)
- Application inference profile ARNs (e.g., `arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/abc123`)

ARN model IDs are URL-encoded automatically so reserved characters (`:`, `/`) stay as a single path segment.

### Example

```bash
curl http://127.0.0.1:9180/apisix/admin/routes/1 -X PUT \
  -H 'X-API-KEY: ...' -d '{
    "uri": "/ai/converse",
    "plugins": {
      "ai-proxy": {
        "provider": "bedrock",
        "auth": {
          "aws": {
            "access_key_id": "AKIA...",
            "secret_access_key": "..."
          }
        },
        "provider_conf": { "region": "us-east-1" },
        "options": {
          "model": "anthropic.claude-3-5-sonnet-20240620-v1:0"
        }
      }
    }
  }'
```

```bash
curl http://127.0.0.1:9080/ai/converse -X POST \
  -H 'Content-Type: application/json' -d '{
    "messages": [{"role": "user", "content": [{"text": "What is 1+1?"}]}],
    "inferenceConfig": {"maxTokens": 256}
  }'
```

### Implementation highlights

- AWS SigV4 signing uses `resty.aws.request.sign` (already a dependency)
- Path encoding handles AWS's [double-encoding requirement](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) for canonical URIs (e.g., `:` → `%3A` on the wire → `%253A` in canonical URI)
- `path:gsub` preserves slash structure including trailing slashes and empty segments
- Header values from the signer are written back using lowercase keys to match `construct_forward_headers()` and avoid duplicates

### Limitations / future work

- **Non-streaming only.** Bedrock's `ConverseStream` (which uses AWS EventStream binary protocol, not SSE) is not yet supported.
- Bedrock is the only AWS service currently — other AWS services (e.g., SageMaker) would need separate provider modules.

### Checklist


apisix/plugins/ai-protocols/bedrock-converse.lua
```diff
@@ -0,0 +1,283 @@
...
+
+local _M = {}
+
+
+--- Detect whether the request matches the Bedrock Converse API format.
+-- Uses URI suffix (/converse) and body (valid JSON table with messages).
+function _M.matches(body, ctx)
+    local uri = ctx.var and ctx.var.uri
+    return uri and string_sub(uri, -9) == "/converse"
+        and type(body) == "table" and type(body.messages) == "table"
```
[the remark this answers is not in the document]
shreemaan-abhishek: nah this is unnecessary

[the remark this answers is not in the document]
shreemaan-abhishek: A user sending an image-only or tool-only message (valid Bedrock requests) would have content blocks that don't
contain text, and the tightened matches() would return false. Their request would then fall through to
openai-chat, producing the same "provider does not support openai-chat" error this change is trying to avoid —
except now it'd hit valid users instead of misconfigured ones.

The mis-detection scenario this addresses requires a user to (1) explicitly route to a /converse URI on the
bedrock provider, and (2) send a body using OpenAI's string-content format. That's a user error, and the current
error message ("provider bedrock does not support openai-chat protocol") clearly tells them their body doesn't
match Bedrock format.

If we wanted to tighten without breaking multimodal/tool use, the safer check would be "content is an array of
objects" (distinguishes from OpenAI's string content). But since /converse URI is an explicit opt-in for Bedrock,
I think the current behavior is appropriate.

apisix/plugins/ai-providers/base.lua
```diff
@@ -187,6 +187,20 @@ function _M.build_request(self, ctx, conf, request_body, opts)
...
+        local region = opts.conf and opts.conf.region
+        if not region then
+            return nil, "missing region for AWS SigV4 signing "
+                .. "(provider_conf.region required for bedrock)"
+        end
+        local sign_err = auth_aws.sign_request(params, auth.aws, region)
+        if sign_err then
+            return nil, "failed to sign AWS request: " .. sign_err
+        end
+    end
```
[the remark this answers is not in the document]
membphis: If we can follow this way, it is much better @shreemaan-abhishek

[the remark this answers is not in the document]
shreemaan-abhishek: already done: 
image

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -234,7 +308,58 @@ _M.ai_proxy_schema = {
...
+                },
+            },
+            ["then"] = {
+                properties = {
+                    options = { required = { "model" } },
+                },
+                required = { "options" },
+            },
+        },
+    },
```
nic-6443: I think the logic of this if-else block is overly complicated. It would be sufficient to throw an error directly when a request is made if the necessary auth configuration is missing.

apisix/plugins/ai-protocols/bedrock-converse.lua
```diff
@@ -0,0 +1,283 @@
...
+-- Streaming is not supported in this phase.
+function _M.is_streaming(body)
+    return false
+end
+
+
+--- Prepare the outgoing request body for the target provider.
+-- Remove fields Bedrock doesn't accept.
+function _M.prepare_outgoing_request(body)
+    body.stream = nil
```
nic-6443, line 53: This should be a TODO item, not that Bedrock doesn't accept it?

apisix/plugins/ai-providers/bedrock.lua
```diff
@@ -0,0 +1,81 @@
...
+        scheme = "https",
+        host = get_host(get_region(instance_conf)),
+        port = 443,
+    }
+end
+
+
+-- Map override.request_body fields to Bedrock Converse API format.
+-- Bedrock uses inferenceConfig.maxTokens (camelCase, nested) for max output tokens.
+local function rewrite_chat_request_body(body, override, force)
```
nic-6443:
```suggestion
local function rewrite_converse_request_body(body, override, force)
```

nic-6443: Since the bedrock provider may support other AI protocols in the future, such as openai-chat and anthropic-message, naming needs to be done according to the protocol here

apisix/plugins/ai-proxy/schema.lua
```diff
@@ -348,7 +359,42 @@ _M.ai_proxy_multi_schema = {
...
+        if not (conf.auth and conf.auth.aws) then
+            return false, "bedrock requires auth.aws"
+        end
+        if not has_override then
+            local has_model = conf.options and conf.options.model
+            if not has_model then
+                return false, "bedrock requires options.model when "
+                    .. "override.endpoint is not set"
+            end
+        end
```
nic-6443: In this case, users need to configure a separate plugin for each model supported by bedrock, which feels extremely inconvenient to use. We need to investigate how other AI gateways handle this situation. For example, downstream requests should be allowed to specify the model in a certain way.

t/fixtures/bedrock/converse-basic.json
```diff
@@ -0,0 +1,14 @@
...
+            "content": [{"text": "1 + 1 = 2."}]
+        }
+    },
+    "stopReason": "end_turn",
+    "usage": {
+        "inputTokens": 10,
+        "outputTokens": 8,
+        "totalTokens": 18
+    }
+}
```
nic-6443, line 22: From now on, we will no longer create fake data for testing; instead, we will directly record real responses from the Bedrock API.

apisix/plugins/ai-proxy/schema.lua
shreemaan-abhishek: we can specify the model in request body, I have added the logic for this now.

nic-6443 (approved):

membphis (approved):

AlinsRan (approved):

elsodiop:
Hello,
ai-proxy accepts $secret:// references in auth.aws, but the Bedrock provider passes access_key_id directly to the SigV4 signer without resolving the secret first. The signer therefore builds the credential scope using the literal secret URI.
I created a Bedrock toute with:
        - name: ai-proxy
          enable: true
          config:
            provider: bedrock
            auth:
              aws:
                access_key_id: "$secret://aws/xxxxxxxx/secret-name/access-key-id"
                secret_access_key: "$secret://aws/xxxx/secret-name/secret-access-key"
            provider_conf:
              region: us-east-1
            options:
              model: anthropic.claude-3-5-sonnet-20240620-v1:0
When i test my route i get {"message":"Credential must have exactly 5 slash-delimited elements, e.g. keyid/date/region/service/term, got '$secret://aws/xxxx/secret-name/access-key-id/20260723/us-east-1/bedrock/aws4_request'"}

shreemaan-abhishek:
Hello @elsodiop 

The bedrock provider doesn't do its own secret resolution, and neither does any other plugin: since #13312 the whole plugin conf is resolved centrally in plugin.filter (apisix/plugin.lua:97) before any plugin handler runs, so auth.aws is already plain text when it reaches the SigV4 signer.

What your error shows is that the reference failed to resolve: secret.lua falls back to the literal string when a fetch fails (refs[k] = fetch(v, use_cache) or v), so the raw URI ends up in the credential scope. Could you confirm
- your APISIX version
- and that the /apisix/secrets/aws/<id> resource exists and its credentials can read that secret?

Separately, this did surface a real problem: when resolution fails on the cached path there is no error log at all, so the misconfiguration is invisible. I'll open a fix for that.

elsodiop:
Thanks @shreemaan-abhishek  for your feedback. 
I resolve my problem as you mention with iam user credentials. 
Am using the APISIX 3.17 version.
Best regards

PR #13268, opened 2026-04-21
feat(forward-auth): support attaching consumer by response header

panxiaoquan:
### Description

  This PR extends the `forward-auth` plugin to support attaching an existing APISIX Consumer from a
  successful authorization response header.

  Today, `forward-auth` can forward request headers to the auth service and propagate selected response
  headers back to the upstream or client, but it cannot turn a successful auth response into APISIX
  Consumer context. That means Consumer-scoped plugins and Consumer Group plugins cannot participate in
  requests authenticated by an external auth service.

  This change adds an opt-in `consumer_header` attribute to `forward-auth`. When it is configured and the
  auth service returns a `2xx` response, APISIX reads the configured response header as a Consumer
  username, looks up the existing Consumer locally, and attaches it to the current request. This allows
  downstream Consumer-related plugins to work with `forward-auth`-authenticated traffic.

  The implementation also moves the auth execution into `rewrite` and uses `ctx.forward_auth_processed` to
  avoid running the auth flow twice in `access`. This keeps the request authenticated once while making the
  attached Consumer available early enough for Consumer plugins to take effect.

  Behavior on invalid auth-to-consumer mapping is explicit:
  - If the configured response header is missing, APISIX returns `403`.
  - If the referenced Consumer does not exist, APISIX returns `403`.

  Additional changes included in this PR:
  - Expose a helper in `apisix.consumer` to fetch an existing Consumer by name from local cache.
  - Add English and Chinese documentation for the new `consumer_header` attribute and example usage.
  - Add plugin tests covering schema validation, successful Consumer attachment, execution of Consumer
  plugins after attachment, missing header, and unknown Consumer.
  - Add a `.luarc.json` workspace config to improve local LuaLS path/global resolution for development.

  ### Checklist

  - [x] I have verified that this change is backward compatible (If not, please discuss on the [APISIX
  mailing list](https://github.com/apache/apisix/tree/master#community) first)

### DIFF
--- a/apisix/plugins/forward-auth.lua
+++ b/apisix/plugins/forward-auth.lua
@@ -18,16 +18,17 @@
 local ipairs   = ipairs
 local core     = require("apisix.core")
 local http     = require("resty.http")
+local consumer = require("apisix.consumer")
 local pairs    = pairs
 local type     = type
 local tostring = tostring
 
-local schema = {
+local schema   = {
     type = "object",
     properties = {
-        uri = {type = "string"},
-        allow_degradation = {type = "boolean", default = false},
-        status_on_error = {type = "integer", minimum = 200, maximum = 599, default = 403},
+        uri = { type = "string" },
+        allow_degradation = { type = "boolean", default = false },
+        status_on_error = { type = "integer", minimum = 200, maximum = 599, default = 403 },
         ssl_verify = {
             type = "boolean",
             default = true,
@@ -35,13 +36,13 @@ local schema = {
         request_method = {
             type = "string",
             default = "GET",
-            enum = {"GET", "POST"},
+            enum = { "GET", "POST" },
             description = "the method for client to request the authorization service"
         },
         request_headers = {
             type = "array",
             default = {},
-            items = {type = "string"},
+            items = { type = "string" },
             description = "client request header that will be sent to the authorization service"
         },
         extra_headers = {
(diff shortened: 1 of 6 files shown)


apisix/plugins/forward-auth.lua
```diff
@@ -184,14 +218,31 @@ function _M.access(conf, ctx)
...
+        return code, body
+    end
+
+    -- append headers that need to be get from the auth response header
     for _, header in ipairs(conf.upstream_headers) do
         local header_value = res.headers[header]
-        -- if header_value is nil, the client header's value will be removed if it exists
-        core.request.set_header(ctx, header, header_value)
+        if header_value then
+            core.request.set_header(ctx, header, header_value)
```
Baoyuantop: **`upstream_headers` security regression**: The original code intentionally clears client-supplied headers when the auth response does not include a configured upstream header (prevents header spoofing). The new `if header_value then` guard in the loop breaks this behavior — a malicious client can now inject headers that the upstream trusts (e.g., `X-User-Id`). Please revert to the original unconditional `set_header` logic.

apisix/plugins/forward-auth.lua
```diff
@@ -184,14 +218,31 @@ function _M.access(conf, ctx)
...
+        end
     end
 end
 
 
+function _M.rewrite(conf, ctx)
+    return do_auth(conf, ctx)
```
Baoyuantop, line 236: **Moving auth to `rewrite` phase**: Running the full HTTP auth subrequest in `rewrite` is unconventional for APISIX plugins. Please clarify: (a) Why can't this be done in `access`? (b) How does this interact with other `rewrite`-phase plugins like `proxy-rewrite`? (c) What happens when another auth plugin (e.g., `key-auth`) is also configured on the same route — does it overwrite the Consumer set by forward-auth?

apisix/plugins/forward-auth.lua
panxiaoquan: Fixed in latest commit.

apisix/plugins/forward-auth.lua
panxiaoquan: (a) The reason for moving `forward-auth` into `rewrite` is specifically
  the `consumer_header` path.

  For this feature, we do not just need to authenticate the request. We
  need the auth result early enough to attach an APISIX Consumer and allow
  Consumer / Consumer Group plugins to participate in the same request.

  In APISIX, route-level `rewrite` runs first, then APISIX merges Consumer
  plugins, and only after that does it enter `access` (`apisix/init.lua`).
  If `forward-auth` stayed in `access`, it could still attach
  `ctx.consumer`, but that would be too late for Consumer-scoped plugins to
  be merged for the current request. Since the Consumer identity comes from
  the auth response header, the auth HTTP call itself has to complete
  before that merge point.

  (b) With the current priorities, `forward-auth` runs before
  `proxy-rewrite` in the `rewrite` phase (`forward-auth`: 2002,
  `proxy-rewrite`: 1008). That means the auth service sees the original
  client-facing request state, and `proxy-rewrite` only mutates the
  request that goes to the upstream afterwards.

  This does change the interaction with rewrite-phase mutations compared
  with the old access-phase behavior. 

  (c) If another auth plugin on the same route also attaches a Consumer,
  the later plugin wins. `consumer.attach_consumer()` overwrites
  `ctx.consumer` rather than merging. With `key-auth` specifically,
  `key-auth` runs first (priority 2500) and `forward-auth` runs later
  (2002), so `forward-auth` would overwrite the Consumer set by `key-auth`
  if `consumer_header` is configured.

  So the current behavior is last-writer-wins when multiple auth plugins
  attach a Consumer on the same route. If we keep this feature, we should
  either document that limitation clearly or add an explicit guard for
  conflicting Consumer attachment.

PR #13274, opened 2026-04-21
docs: update skywalking and skywalking-logger plugin docs

Yilialinn:
## Summary

- Fix broken admonition block in `skywalking` ZH doc (orphan closing `:::` tag and stray text were inserted before the `admin_key` note block)
- Update APISIX version in `skywalking-logger` JSON response examples from `3.11.0` to `3.13.0` (EN and ZH)
- Add missing `max_req_body_bytes` and `max_resp_body_bytes` attributes to `skywalking-logger` ZH attributes table

## Changes

- `docs/zh/latest/plugins/skywalking.md` — fix malformed admonition
- `docs/en/latest/plugins/skywalking-logger.md` — update version in example output
- `docs/zh/latest/plugins/skywalking-logger.md` — update version + add missing attributes


kayx23 (commented):
The `skywalking-logger` example JSON now looks malformed in both EN and ZH docs: inside `response.headers`, the indentation shifts after the `server` field and the example ends up with a trailing comma / broken object structure. Since this block is presented as JSON output, could you reformat it so the `headers` object stays valid and copy-pasteable?

kayx23 (approved):

Baoyuantop (approved):

shreemaan-abhishek (approved):

PR #13283, opened 2026-04-23
docs(proxy-rewrite): fix regex_uri example

xuruidong:

Fixes #

### Checklist


docs/en/latest/plugins/proxy-rewrite.md
```diff
@@ -42,7 +42,7 @@ The `proxy-rewrite` Plugin offers options to rewrite requests that APISIX forwar
...
+| regex_uri                   | array[string] | False    |         |                                                                                                                                        | Regular expressions used to match the URI path from client requests and compose a new Upstream URI path. When both `uri` and `regex_uri` are configured, `uri` has a higher priority. The array should contain one or more **key-value pairs**, with the key being the regular expression to match URI against and value being the new Upstream URI path. For example, with `["^/iresty/(.*)/(.*)", "/$1-$2", "^/theothers/*", "/theothers"]`, if a request is originally sent to `/iresty/hello/world`, the Plugin will rewrite the Upstream URI path to `/hello-world`; if a request is originally sent to `/theothers/hello/world`, the Plugin will rewrite the Upstream URI path to `/theothers`. |
```
Baoyuantop: The key evidence is in `apisix/plugins/proxy-rewrite.lua:294-308`: after `regex_uri` matches, APISIX runs `re_sub(upstream_uri, pattern, replacement, "jo")` on the whole upstream URI. That means only the matched part is replaced, while any unmatched suffix is preserved. The existing test `t/plugin/proxy-rewrite.t:572-602` shows this behavior as well: when the pattern `^/test/(.*)/(.*)/(.*)` covers the full path, `/test/plugin/proxy/rewrite` becomes exactly `/plugin_proxy_rewrite`.

In this PR, the docs still use `"^/theothers/*" -> "/theothers"` and say `/theothers/hello/world` will be rewritten to `/theothers`. That pattern does not consume the trailing `hello/world` segment, so it cannot produce the documented result. In other words, this revision fixes the first wrong example, but the second wrong example still remains in both the English and Chinese docs.

Please update the second example to a pattern that consumes the remaining path as well (for example, if the intent is to collapse any `/theothers/...` request to `/theothers`, something like `^/theothers(/.*)?$` would be closer to the actual behavior), and then I can take another quick look.

kayx23:
Please resolve conflicts

kayx23 (approved):

Baoyuantop (approved):

xuruidong:
ping @shreemaan-abhishek

xuruidong:
Hi @Yilialinn , could you please help me review this PR?

PR #13307, opened 2026-04-27
feat(ai-proxy): support Bedrock ConverseStream streaming

shreemaan-abhishek:
## Summary

Phase 2 of the Bedrock provider (Phase 1 was #13249). Routes requests with `body.stream = true` to `/model/<model>/converse-stream` and forwards the AWS EventStream binary response unchanged.

- New `ai-transport.eventstream` codec parses/encodes EventStream frames (12-byte prelude + headers + payload + trailing CRC, validated via `ngx.crc32_long`); same API surface as `ai-transport.sse` so providers pick a framing module by name.
- `ai-providers.base.parse_streaming_response` selects `sse` vs `aws-eventstream` framing via `provider.streaming_framing`.
- Bedrock provider declares `streaming_framing = \"aws-eventstream\"` and routes `ctx.var.request_type == \"ai_stream\"` to `/converse-stream`.
- `bedrock-converse` protocol decodes EventStream events: `contentBlockDelta` -> texts; `metadata` -> `usage_and_done`; `:message-type=exception` -> `done` with warning.

The wire format is decoded in pure Lua against the spec at https://docs.aws.amazon.com/lexv2/latest/dg/event-stream-encoding.html. Streaming is opt-in (`body.stream = true`); existing non-streaming routes keep going to `/converse`.

## Test plan

- [x] New streaming test cases (4 in `t/plugin/ai-proxy-bedrock.t`, 3 in `t/plugin/ai-proxy-bedrock-single.t`) cover: route hits `/converse-stream`, SigV4 still validates, Content-Type is `application/vnd.amazon.eventstream`, raw bytes contain expected event ordering, prompt/completion tokens extracted from `metadata`, and a non-streaming control on the same route.
- [x] Test fixture (`t/fixtures/bedrock/bedrock-converse-streaming.bin`) is a real recorded `/converse-stream` response (6 frames: messageStart, contentBlockDelta×2, contentBlockStop, messageStop, metadata).
- [x] All 114 existing Bedrock tests still pass (60 multi + 54 single).
- [x] Live verification against real AWS Bedrock (`anthropic.claude-3-haiku-20240307-v1:0` in `ap-southeast-1`): streaming, system prompt, per-request `body.model`, `ai-proxy-multi`, non-streaming after streaming on the same route, and error paths (missing model, invalid JSON, wrong method) — all work.

## Docs

- `docs/{en,zh}/latest/plugins/ai-proxy.md` and `ai-proxy-multi.md` — added `stream` to the Bedrock request format table; replaced the \"not yet supported\" note with a streaming usage section.


membphis (approved):

apisix/plugins/ai-providers/base.lua
```diff
@@ -25,13 +25,11 @@ local mt = {
...
 }
 
--- Maximum SSE buffer size per request (1 MB).
-local MAX_SSE_BUF_SIZE = 1024 * 1024
-
 local core = require("apisix.core")
 local plugin = require("apisix.plugin")
 local url  = require("socket.url")
 local sse  = require("apisix.plugins.ai-transport.sse")
+local eventstream = require("apisix.plugins.ai-transport.eventstream")
```
nic-6443: The naming of `eventstream` is too generic and should include the aws keyword.

shreemaan-abhishek: Done in 3b667a854 — renamed the file to `aws-eventstream.lua`, the require local to `aws_eventstream`, and the log prefix strings to `aws-eventstream:` so the AWS scope is unambiguous.

nic-6443 (approved):

t/lib/server.lua
```diff
@@ -1036,6 +1036,114 @@ function _M.bedrock_converse()
 end
 
 
+-- Mock for Bedrock /converse-stream. Reuses the same SigV4 + body shape
```
membphis, line 1039: Currently, all AI upstream services are simulated using this fake server method.
I'm worried that the difference between the fake server and the real LLM request here is too big.

Should we introduce a container specifically for LLM Fake Server?

membphis (approved):

t/lib/server.lua
shreemaan-abhishek: this can be added as a follow up.

AlinsRan (approved):

PR #13311, opened 2026-04-28
feat(chaitin-waf): add allow_degradation option for fail-open behavior

jlongWang:
Add `allow_degradation` configuration option to allow requests to pass through when the WAF service is unavailable (e.g., connection timeout, no healthy nodes). This prevents the WAF service from becoming a single point of failure.

Changes:
- Add `allow_degradation` boolean field to plugin schema (default: false)
- Implement degradation logic for WAF server unavailability
- Implement degradation logic for WAF detection errors/timeout
- Update documentation (zh/en)
- Add test cases for degradation behavior

Fixes #

### Checklist


walkerxiong:
Not stale — please keep this open. We are running APISIX + chaitin-waf in
production and this PR addresses a real availability gap that currently has no
workaround at the configuration level.

**Production impact.** With `mode: block`, the SafeLine detector is a hard
availability dependency of the gateway. In our fault-injection tests:

- detector killed → every request burns `connect_timeout` (default 1s) and
  returns 500; a WAF outage becomes a full gateway outage
- detector accepting but hanging → every request burns `read_timeout`, then 500
- even a 2% detector connection-failure rate translates directly into 2%
  user-facing 5xx

**Why this can't be fixed outside the plugin.** The 500 is produced by the
plugin in the `access` phase, before the request reaches any upstream —
`retries` / `proxy-next-upstream` / response-phase plugins cannot intercept it.
The only operational workaround today is switching `mode` to `monitor` via the
Admin API during an incident, which gives up blocking entirely. Detection
action (block/monitor) and failure policy (fail-open/fail-closed) are
orthogonal concerns; this PR correctly decouples them, and `allow_degradation`
is consistent with the existing option of the same name in `limit-count` /
`limit-req` / `limit-conn`. Comparable products (AWS WAF, Cloudflare, NGINX
App Protect `failure_mode_action`) all expose this switch.

**Two small gaps worth addressing (happy to help):**

1. `do_access` has a third 500 path that the PR doesn't cover: missing plugin
   metadata / no `nodes` (`chaitin-waf.lua` ~L306-310). For consistency,
   `allow_degradation` should apply there too — notably, even `monitor` mode
   returns 500 on that path today.
2. `allow_degradation` is only added to `plugin_schema`. Adding it to
   `metadata_schema` as well (same precedence as `mode`:
   route > metadata > default) would let operators set a global default
   instead of repeating it on every route.

@jlongWang are you still available to update this? If not, I'm willing to
pick it up in a follow-up PR keeping your authorship.

PR #13319, opened 2026-04-29
fix: add fetch_secrets support for ai-rate-limiting, limit-conn, and limit-req plugins

shreemaan-abhishek:
### Description

While testing distributed rate limiting with secret references, I noticed that `ai-rate-limiting`, `limit-conn`, and `limit-req` were passing literal strings such as `"$ENV://REDIS_HOST"` to Redis instead of resolving them, even though `limit-count` already supported this. The root cause is that those three plugins never invoked `apisix.secret.fetch_secrets` on their config before using it. This change adds the missing `fetch_secrets` calls in the relevant access/log/instance-status entry points so that `$ENV://` and `$secret://` references in fields like `redis_host` are resolved consistently across all rate-limiting plugins.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/ai-rate-limiting.lua
+++ b/apisix/plugins/ai-rate-limiting.lua
@@ -24,6 +24,7 @@ local load = load
 local math_floor = math.floor
 local math_huge = math.huge
 local core = require("apisix.core")
+local fetch_secrets = require("apisix.secret").fetch_secrets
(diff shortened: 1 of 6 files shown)


nic-6443:
@shreemaan-abhishek We have a new plan for the secret feature #13312

PR #13326, opened 2026-04-30
fix(loki-logger): prevent log_labels cross-request leakage

Adarsh879:
### Description

`loki-logger` mutates `conf.log_labels` in place when substituting
`$variable` placeholders, then captures that shared table in the
batch-processor flush closure. The early return at `add_entry` causes
all subsequent requests on the same worker to skip the resolution
block, so every later entry flushes with the **first** request's
resolved labels. Symptom: log entries where `log_labels` and
`log_format` disagree on the same line (#12485), or a stream label
that carries a value from a different request than the one whose
body it accompanies.

The bug affects `loki-logger` at any scope where multiple requests
share the same conf and a `$variable` in `log_labels` resolves to
different values per request — so it surfaces in real deployments
both at global-rule and at service scope.

#### Root cause (pre-fix)

```lua
if batch_processor_manager:add_entry(conf, entry, max_pending_entries) then
    return                          -- subsequent requests bypass everything below
end

local labels = conf.log_labels      -- alias, NOT a copy
for key, value in pairs(labels) do
    local new_val, err, n_resolved = core.utils.resolve_var(value, ctx.var)
    if not err and n_resolved > 0 then
        labels[key] = new_val       -- mutates conf.log_labels
    end
end

local func = function(entries)      -- closure captures the shared table
    local data = { streams = { { stream = labels, values = ... } } }
    ...
end
```

#### Fix

- Resolve labels **per request** into a fresh table; never mutate
  `conf.log_labels`.
- Attach the resolved labels to each entry (`entry.loki_labels`).
- At flush time, group entries by their resolved label set and emit
  one Loki stream per group within the push (Loki's `streams` API
  already supports this, so batching is preserved).

The change is fully contained to `apisix/plugins/loki-logger.lua` —
nothing in `plugin.lua`, `batch-processor-manager.lua`, or other
plugins is touched.

#### Test

`t/plugin/loki-logger.t` TEST 17–19 (modeled on the existing TEST
12–14 pattern):

1. Configure a route with `log_labels: { custom_label: "$arg_X" }`
   and `batch_max_size: 1`.
2. Hit `/hello?X=alpha` then `/hello?X=beta`.
3. Query Loki for `{custom_label="beta"}`. Without the fix the
   second request's value never reaches Loki (the worker resolved
   `"alpha"` once and reused it forever), so the assertion fails.
   With the fix, both stream labels exist.

Verified locally:

| Plugin source                          | Result   |
|----------------------------------------|----------|
| Pre-fix (master before this PR)        | TEST 19 fails: `no entries with custom_label=beta` |
| Post-fix (this PR)                     | All 54 subtests pass |

Fixes #12485

### Checklist

- [x] I have updated the documentation to reflect this change *(no doc change needed; bug fix, no surface change)*

### DIFF
--- a/apisix/plugins/loki-logger.lua
+++ b/apisix/plugins/loki-logger.lua
@@ -202,6 +202,20 @@ function _M.body_filter(conf, ctx)
 end
 
 
+local function resolve_labels(conf_labels, ctx)
+    local labels = new_tab(0, 4)
+    for key, value in pairs(conf_labels) do
+        local new_val, err, n_resolved = core.utils.resolve_var(value, ctx.var)
+        if not err and n_resolved > 0 then
+            labels[key] = new_val
+        else
+            labels[key] = value
+        end
+    end
+    return labels
+end
+
+
 function _M.log(conf, ctx)
     local metadata = plugin.plugin_metadata(plugin_name)
     local max_pending_entries = metadata and metadata.value and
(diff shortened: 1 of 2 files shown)


Baoyuantop (commented):
Could you add one regression test with `batch_max_size > 1`, for example sending `/hello?X=alpha` and `/hello?X=beta` so they are processed in the same batch flush, then asserting that both label streams are present in Loki or that the push payload contains two streams? That would cover the most important new grouping logic introduced by this fix.

Adarsh879:
Thanks for the review! Added in 9e79f92e35c3d43fddf02c4f89af09ae5f7387de.

`TEST 20–22` now covers the batch grouping path:

- Route configured with `log_labels: { batch_label: "$arg_X" }` and **`batch_max_size: 2`**
- Sends `/hello?X=gamma` then `/hello?X=delta` so both entries land in the **same batch flush**
- Asserts Loki ends up with **two distinct streams** (`batch_label="gamma"` and `batch_label="delta"`)

I verified it both ways locally (Loki 2.8 + `-auth.enabled`, matching CI):

- **With the fix:** all subtests pass
- **Reverting just the plugin change:** `TEST 22` fails — the buggy code produces a single frozen `gamma` stream holding both entries

PR #13331, opened 2026-05-05
feat(cas-auth): sign request URI cookie and tighten cookie attributes

shreemaan-abhishek:
### Description

This change updates the `cas-auth` plugin so that the `CAS_REQUEST_URI` cookie — which the plugin uses to remember where the user was headed before being bounced to the CAS IdP — is signed with HMAC-SHA256 and verified on the IdP callback using a constant-time comparison.

After successful CAS ticket validation, the recovered redirect target is also validated as a same-origin path before being applied; if the cookie is missing, malformed, fails signature verification, or carries something that is not a relative path, the plugin falls back to `/` instead of using the cookie value verbatim.

A new required `cookie` object is added to the plugin schema:

- `cookie.secret` (string, required, minLength 32) — HMAC-signing key. Must match across all APISIX nodes so cookies issued on one node remain verifiable on another. Encrypted at rest via `schema.encrypt_fields`.
- `cookie.secure` (boolean, default `true`) — whether to set the `Secure` attribute on issued cookies. Set to `false` only for deployments where the protected route is not served over HTTPS.
- `cookie.samesite` (string, default `"Lax"`, enum `"Lax"`/`"None"`) — value for the `SameSite` cookie attribute. `"Strict"` is intentionally not exposed because it suppresses the cookie on the IdP→SP top-level redirect when the IdP is on a different site. `samesite="None"` requires `secure=true` (rejected at `check_schema` time) because browsers refuse to store `SameSite=None` cookies without `Secure`.

`HttpOnly` continues to be set unconditionally on the issued cookies. `t/lib/keycloak_cas.lua` is updated so existing integration tests continue to pass; new tests in `t/plugin/cas-auth.t` exercise the schema constraints, the safe-redirect predicate, and the HMAC sign/verify roundtrip including tampering, wrong-secret, and malformed-cookie cases.

### Breaking change

**This PR is not backward compatible for existing `cas-auth` plugin configurations.** A new field `cookie.secret` is added as a required field of the schema, and a route that has `cas-auth` enabled without it will fail `check_schema` and be rejected at admin-API time.

**Why it has to be required:** the cookie integrity guarantee this PR introduces depends on every issued cookie being signed and every received cookie being verified. Making the secret optional would mean some deployments sign and some do not — the verification path could not assume a signature exists, the safe-redirect fallback could not distinguish "unsigned cookie" from "tampered cookie", and the protection would be defeated in exactly the deployments that did not opt in. Requiring the secret keeps the invariant "every cookie issued by this plugin is signed" simple and enforceable.

**Migration for existing configurations:** add a `cookie.secret` of at least 32 characters to every route whose `plugins` contains `cas-auth`. The same value must be configured on every APISIX node in the deployment so that cookies issued on one node remain verifiable on another. Generate one with e.g. `openssl rand -base64 48`.

Before:
```json
{
  "plugins": {
    "cas-auth": {
      "idp_uri": "https://cas.example.com",
      "cas_callback_uri": "/cas/callback",
      "logout_uri": "/cas/logout"
    }
  }
}
```

After:
```json
{
  "plugins": {
    "cas-auth": {
      "idp_uri": "https://cas.example.com",
      "cas_callback_uri": "/cas/callback",
      "logout_uri": "/cas/logout",
      "cookie": {
        "secret": "<32+ char random string>"
      }
    }
  }
}
```

`cookie.secure` and `cookie.samesite` are optional and default to `true` and `"Lax"` respectively, which match the previous unconditional behavior introduced earlier in this PR.

**Operational impact:** during a rolling upgrade, any node already running the new code will reject old-format configs from etcd until they are updated. Recommended order: update all `cas-auth` route configurations to include `cookie.secret` first, then roll out the new APISIX binary.

N/A

### Checklist

### DIFF
--- a/apisix/plugins/cas-auth.lua
+++ b/apisix/plugins/cas-auth.lua
@@ -16,12 +16,16 @@
 ----
 local core = require("apisix.core")
 local http = require("resty.http")
+local hmac = require("resty.openssl.hmac")
+local bit = require("bit")
 local ngx = ngx
 local ngx_re_match = ngx.re.match
+local ngx_encode_base64 = ngx.encode_base64
+local ngx_decode_base64 = ngx.decode_base64
 
 local CAS_REQUEST_URI = "CAS_REQUEST_URI"
 local COOKIE_NAME = "CAS_SESSION"
-local COOKIE_PARAMS = "; Path=/; HttpOnly"
+local COOKIE_PARAMS = "; Path=/; HttpOnly; Secure; SameSite=Lax"
 local SESSION_LIFETIME = 3600
 local STORE_NAME = "cas_sessions"
 
@@ -35,9 +39,10 @@ local schema = {
         idp_uri = {type = "string"},
         cas_callback_uri = {type = "string"},
         logout_uri = {type = "string"},
+        cookie_secret = {type = "string", minLength = 32},
     },
     required = {
-        "idp_uri", "cas_callback_uri", "logout_uri"
+        "idp_uri", "cas_callback_uri", "logout_uri", "cookie_secret"
     }
 }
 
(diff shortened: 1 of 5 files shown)


membphis (changes requested):
This is a breakchange
This point must be clearly stated in the PR description.

shreemaan-abhishek:
@membphis it is mentioned at the end of the PR description that it is not backward compatible.

membphis:
> @membphis it is mentioned at the end of the PR description that it is not backward compatible.

only this is not enough, pls write some detail doc about the `not backward compatible`

image

shreemaan-abhishek:
@membphis updated the PR description with a **Breaking change** section that:

- Calls out the incompatibility explicitly at the top
- Explains why `cookie.secret` is required rather than optional (the integrity invariant only holds if *every* issued cookie is signed; making it optional would defeat the guarantee for unconfigured routes)
- Provides a concrete before/after JSON migration example
- Notes the rolling-upgrade operational impact (update configs first, then roll out the binary)

Happy to add anything else if there's a specific deployment concern you want covered.

docs/en/latest/plugins/cas-auth.md
```diff
@@ -64,7 +68,10 @@ curl http://127.0.0.1:9180/apisix/admin/routes/cas1 -H "X-API-KEY: $admin_key" -
           "cas-auth": {
               "idp_uri": "http://127.0.0.1:8080/realms/test/protocol/cas",
               "cas_callback_uri": "/anything/cas_callback",
-              "logout_uri": "/anything/logout"
+              "logout_uri": "/anything/logout",
+              "cookie": {
+                  "secret": "please-replace-with-a-32+-char-random-secret"
```
nic-6443: This example still runs the protected route over `http://127.0.0.1`, but the new default is `cookie.secure=true`. Browsers will ignore both `CAS_REQUEST_URI` and `CAS_SESSION` when they are set with `Secure` over HTTP, so following this example would either fall back to `/` after login or fail to keep the session cookie. Can we either add `"secure": false` here for the local HTTP example, or switch the example to HTTPS?

nic-6443 (approved):

apisix/plugins/cas-auth.lua
```diff
@@ -35,23 +38,54 @@ local schema = {
...
         cas_callback_uri = {type = "string"},
         logout_uri = {type = "string"},
+        cookie = {
+            type = "object",
+            properties = {
+                secret = {type = "string", minLength = 32},
+                secure = {type = "boolean", default = true},
+                samesite = {type = "string", enum = {"Lax", "None"}, default = "Lax"},
+            },
+            required = {"secret"},
```
AlinsRan, line 48: Perhaps we can consider automatically generating when `cookie.secret` is empty to avoid destructive changes?

shreemaan-abhishek: It doesn't avoid the breaking change in practice 🤔. In multi-node deployment scenario, the cookie would be set on the pre-redirect request but verified on the IdP callback, which can land on a different node. If each node generates its own secret, verification fails across nodes and on every reload, so CAS login silently breaks instead of the config being loudly rejected.

Also how can we persist such autogenerated cookie?

AlinsRan (approved):

membphis (approved):

PR #13334, opened 2026-05-06
feat(elasticsearch-logger): support dynamic index with time and variable resolution

nic-6443:
The `field.index` configuration now supports dynamic patterns:
- Date/time patterns in curly brackets, e.g. `services-{%Y.%m.%d}`
- APISIX variables prefixed with \$, e.g. `services-$host`
- Both combined, e.g. `services-$host-{%Y.%m.%d}`

The index is resolved per-request without mutating the plugin configuration object, so it stays dynamic across requests (e.g. rotates correctly when the date changes).

This was previously documented in the attribute table but not actually implemented in the open-source codebase.


apisix/plugins/elasticsearch-logger.lua
```diff
@@ -200,11 +202,36 @@ local function get_es_major_version(uri, conf)
 end
 
 
-local function get_logger_entry(conf, ctx)
+local function resolve_index_vars(index, var)
+    local function replace_time(m)
```
membphis: To avoid temporary functions like `replace_time`

```lua
local function replace_time(m)
 ... ...
end

local function resolve_index_vars(index, var)
 ... ...
 -- same as current
 local new_index, _, err = ngx_re.gsub(index, "(?<!\\$){([^}]*)}", replace_time, "jo")
    if not new_index then
        core.log.error("failed to substitute time format: ", err)
    end
  ... ...
end

membphis (dismissed):

shreemaan-abhishek (dismissed):

AlinsRan (dismissed):

apisix/plugins/elasticsearch-logger.lua
nic-6443: Done, moved `replace_time` to module level.

membphis (approved):

AlinsRan (approved):

shreemaan-abhishek (approved):

PR #13336, opened 2026-05-07
docs: update plugin docs with ADC and Ingress Controller examples

Yilialinn:
## Summary

Update plugin documentation for aws-lambda, public-api, real-ip, authz-keycloak, and serverless plugins with comprehensive configuration examples for different deployment methods.

## Changes

- **Add ADC (API7 Declarative Config) examples**: Show how to configure each plugin using declarative YAML configuration
- **Add Ingress Controller examples**: Provide both Gateway API and APISIX CRD examples for Kubernetes deployments
- **Improve Admin API examples**: Update curl commands to use `admin_key` environment variable for better security
- **Add multi-tab interface**: Organize examples using tabs to make it easier for users to find their preferred configuration method
- **Enhance example descriptions**: Improve clarity and add more context for each example

## Related PRs

- api7/docs#1554 (aws-lambda, public-api, real-ip, serverless-functions)
- api7/docs#1556 (authz-keycloak)

## Testing

All examples have been verified for syntax correctness and formatting. The documentation builds successfully with the changes.


docs/en/latest/plugins/aws-lambda.md
```diff
@@ -200,46 +477,186 @@ Next, select **Create resource** to create a sub-path:
...
 
 Select the allowed HTTP method in the dropdown. For the purpose of demonstration, this example continues to use the same Lambda function as the triggered action when the path is requested:
 
-![create method and lambda function](https://static.api7.ai/uploads/2024/04/26/vni7yS2q_create%20method%202.png)
+<div style={{textAlign: 'center'}}>
+<img
+  src="https://static.api7.ai/uploads/2024/04/26/vni7yS2q_create%20method%202.png"
+  alt="create method and lambda function"
+  width="70%"
+/>
```
kayx23, line 499: The new sub-path walkthrough still only matches `/aws-lambda` in the ADC, Gateway API, and APISIX CRD examples, but the verification step sends `GET /aws-lambda/api7-docs`. That request will never hit those routes as written, so the example is broken outside the Admin API tab. These variants need a prefix/wildcard match (for example `/aws-lambda/*` or the equivalent controller syntax) to actually demonstrate sub-path forwarding.

kayx23 (approved):

shreemaan-abhishek (approved):

guoqqqi (approved):

PR #13339, opened 2026-05-07
fix(chaitin-waf): use trusted client IP source for WAF backend

shreemaan-abhishek:
### Description

The chaitin-waf plugin populates the `client_ip` field sent to the Chaitin WAF backend by reading `ctx.var.http_x_forwarded_for` directly when `config.real_client_ip = true` (the default). This bypasses nginx's `real_ip` module: any external client can supply an arbitrary `X-Forwarded-For` header and have it forwarded verbatim to the WAF, making attribution and policy decisions on that field attacker-controllable.

This change replaces the raw-header read with `core.request.get_remote_client_ip(ctx)` (which returns `$remote_addr` after `real_ip` processing) when `real_client_ip` is enabled, and `core.request.get_ip(ctx)` (direct TCP peer, ignoring realip rewrites) when disabled. The toggle's intent is preserved:

- `real_client_ip = true` (default): forwards the realip-resolved client IP — the forwarded header is honored only when the request comes from a trusted proxy listed in `set_real_ip_from`. Spoofed headers from untrusted sources are ignored.
- `real_client_ip = false`: forwards the actual TCP peer address.

This applies the same fix shape used for wolf-rbac in #13329.

An `info`-level log line records the chosen `client_ip` for auditability and to anchor the regression tests.

Fixes #

### Checklist


membphis (commented):
pls update the doc too(the process to fetch real ip, is different as before)

nic-6443 (dismissed):

AlinsRan (dismissed):

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):

PR #13345, opened 2026-05-09
fix: split global rules phase execution for client-control compatibility

nic-6443:
## Description

Split global rules phase execution so that route/service rewrite plugins run between global rule rewrite and global rule access phases. This fixes a conflict where `client-control` (route rewrite, priority 22000) cannot set the nginx body size override via FFI before a global rule access-phase plugin (e.g., a logger with `include_req_body`) calls `ngx.req.read_body()`.

### Before
```
global_rules rewrite + access → route/service rewrite → route/service access
```
A global rule logger reading the body in access phase triggers nginx body size check before `client-control` can set the FFI override → 413.

### After
```
global_rules rewrite → route/service rewrite → global_rules access → route/service access
```
`client-control` sets the FFI override in route rewrite, then the global rule access phase reads the body using the overridden limit.

### Changes
- `apisix/init.lua`: Split `run_global_rules(nil)` into separate `"rewrite"` and `"access"` calls in both the no-route and matched-route paths
- `apisix/plugin.lua`: Simplified `run_global_rules()` — removed nil phase_name handling, always uses explicit phase name with phase-specific tracer span naming
- `t/plugin/client-control.t`: Added regression tests with 31MB body exceeding nginx's 30M limit, verifying client-control FFI override works when global rule reads body in access phase

### Behavioral Changes
1. Global rule access now runs after route/service rewrite (more correct phase ordering)
2. `consumer-restriction` on global rule now works correctly (consumer info available after merge)
3. If route/service rewrite terminates request (401/403), global rule access won't run — but `log` phase always runs

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -806,12 +806,15 @@ function _M.http_access_phase()
     api_ctx.route_id = route.value.id
     api_ctx.route_name = route.value.name
 
-    -- run global rule
+    -- Split global rule execution: run rewrite first so route/service plugins
+    -- can set overrides (e.g., client-control FFI) before global rule access
+    -- phase runs (e.g., logger body collection).
     local global_rules, conf_version = apisix_global_rules.global_rules()
-    plugin.run_global_rules(api_ctx, global_rules, conf_version, nil)
+    plugin.run_global_rules(api_ctx, global_rules, conf_version, "rewrite")
 
     if route.value.script then
         script.load(route, api_ctx)
+        plugin.run_global_rules(api_ctx, global_rules, conf_version, "access")
         script.run("access", api_ctx)
 
     else
@@ -850,6 +853,7 @@ function _M.http_access_phase()
                 plugin.run_plugin(phase, api_ctx.plugins, api_ctx)
             end
         end
+        plugin.run_global_rules(api_ctx, global_rules, conf_version, "access")
         plugin.run_plugin("access", plugins, api_ctx)
     end
     span:finish(ngx_ctx)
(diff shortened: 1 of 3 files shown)


apisix/init.lua
```diff
@@ -850,6 +853,7 @@ function _M.http_access_phase()
                 plugin.run_plugin(phase, api_ctx.plugins, api_ctx)
             end
         end
+        plugin.run_global_rules(api_ctx, global_rules, conf_version, "access")
         plugin.run_plugin("access", plugins, api_ctx)
```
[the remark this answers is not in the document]
nic-6443: The expensive part (merging global rules) is already LRU-cached. The `filter()` call is lightweight — it just iterates through plugins and builds a priority-sorted list. The overhead of calling it twice is negligible compared to the actual plugin execution. Adding a cache for the filtered list would introduce complexity (cache invalidation on plugin changes) without meaningful perf benefit.

AlinsRan (dismissed):

apisix/init.lua
nic-chen: It is recommended to process all phases with nil names by calling rewrite and access separately, which would make things clearer.

membphis (commented):
I think it time to remove `route.value.script`, `service.value.script`

those code: https://github.com/apache/apisix/pull/13345/changes#diff-7439ffe94cc4bcad12066d5d83f6c1004b8e9b3ff5097b645fa7c700eb3b3743R815-R818

anyway, this is another question, we can do it in a new PR

apisix/init.lua
membphis: agree

nic-6443: Done — split the no-route path into explicit rewrite + access calls, and removed the nil handling from `run_global_rules()` entirely.

shreemaan-abhishek (approved):

nic-chen (approved):

membphis (approved):

AlinsRan (approved):

PR #13350, opened 2026-05-11
feat(proxy-cache): add consumer_isolation and cache_set_cookie options

shreemaan-abhishek:
### Description

This PR adds two new options to the `proxy-cache` plugin and tightens how the in-memory cache strategy treats upstream `Cache-Control` response directives.

**New options**

- `consumer_isolation` (boolean, default: `true`). When the request resolves to an APISIX consumer (`ctx.consumer_name`) or carries a remote user (`ctx.var.remote_user`), the identity is prepended to the effective cache key so each consumer gets its own cache namespace. The augmentation is skipped when the user-supplied `cache_key` already references an identity-bearing variable (`$consumer_name`, `$consumer_group_id`, `$remote_user`, or `$http_authorization`). Set to `false` for routes where consumers should share cached responses.
- `cache_set_cookie` (boolean, default: `false`). The in-memory and on-disk strategies refuse to cache responses that include a `Set-Cookie` header unless explicitly opted in. `Set-Cookie` is per-recipient and not generally safe for a shared cache to store.

**Behavior change**

The in-memory strategy now always honors upstream `Cache-Control: private`, `no-store`, and `no-cache` directives, regardless of the existing `cache_control` flag. The `cache_control` flag continues to govern request-side semantics (client `Cache-Control` request directives such as `max-age` / `min-fresh` / `only-if-cached`) and TTL derivation from `max-age` / `s-maxage`. This matches the behavior of NGINX's native `proxy_cache` (used by the on-disk strategy) and brings the two strategies into closer agreement.

**Backward compatibility**

The new defaults change behavior in three ways for users who have proxy-cache enabled today:

1. Routes that combine `proxy-cache` with an authentication plugin will start partitioning the cache by consumer. Existing cache entries become unreachable; new entries accumulate per consumer. Set `consumer_isolation: false` to opt out.
2. The in-memory strategy will no longer cache responses that include `Set-Cookie`. Set `cache_set_cookie: true` to opt back in.
3. The in-memory strategy will no longer cache responses whose upstream `Cache-Control` includes `private`, `no-store`, or `no-cache`, irrespective of the route-level `cache_control` flag.

Each is independent; users can mix and match opt-outs.

N/A

### Checklist


membphis (approved):

apisix/plugins/proxy-cache/init.lua
```diff
@@ -103,6 +123,14 @@ local schema = {
             minimum = 1,
             default = 300,
         },
+        consumer_isolation = {
+            type = "boolean",
+            default = true,
+        },
```
nic-6443, line 129: Any variable can be configured in the existing `cache_key` field. Why is it necessary to handle the consumer separately in a special way?

apisix/plugins/proxy-cache/memory_handler.lua
```diff
@@ -170,14 +170,27 @@ local function cacheable_response(conf, ctx, cc)
         end
     end
 
-    if conf.cache_control and (cc["private"] or cc["no-store"] or cc["no-cache"]) then
+    -- Always honor upstream Cache-Control directives that mark the response as
+    -- non-shared/non-storable, regardless of the conf.cache_control flag. The
+    -- flag governs request-side semantics; upstream response directives are a
+    -- safety contract the application uses to mark personalized content.
+    if cc["private"] or cc["no-store"] or cc["no-cache"] then
```
nic-6443, line 177: Didn't the previous `cache_control` exist exactly to provide an option for users to control whether to follow upstream directives? Why remove this feature?

shreemaan-abhishek: `cache_control` behaviour like this is only in memory_handler, rest of the caching strategies by default honor the upstream's cache_control header.

moreover, the default `false` value guarantees that upstream's decision on cache_control is not honored.

thus we should make the caching behaviour consistent across handlers, and be secure by default.

after removing this `cache_control` field will only interfere with the request-side behaviour.

apisix/plugins/proxy-cache/init.lua
shreemaan-abhishek: we should follow the approach of "secure by default" and ensure the cache is segregated across consumers.

nic-6443 (approved):

AlinsRan (approved):

PR #13358, opened 2026-05-12
fix(dns): correct error() call in dns client initialization

### LINKED ISSUE #13357: fix(limit-count): incorrect error() call in gen_limit_key passes string as level argument
### Current Behavior

In `apisix/plugins/limit-count/init.lua` at line 262, the `error()` 
function is called with incorrect arguments:

error("failed to generate key invalid parent: ", core.json.encode(parent))

Lua's error(message, level) expects the second argument to be an integer 
(stack level: 0, 1, or 2), not a string.

Passing core.json.encode(parent) as the level causes LuaJIT to raise:
bad argument #2 to 'error' (number expected, got string)

This crashes the request with an unhelpful LuaJIT internal error instead 
of the intended error message about the invalid parent configuration.

### Expected Behavior

When conf._meta.parent is nil or missing resource_key, APISIX should log 
a clear error message containing the JSON of the invalid parent object, 
helping developers debug the issue quickly.

### Error Logs

bad argument #2 to 'error' (number expected, got string)

### Steps to Reproduce

1. Open apisix/plugins/limit-count/init.lua
2. Go to line 262
3. Observe the error() call:
   error("failed to generate key invalid parent: ", core.json.encode(parent))
4. The second argument core.json.encode(parent) is a string, but Lua's 
   error() expects an integer as the second argument (stack level).
5. This path is triggered when conf._meta.parent is nil or 
   parent.resource_key is missing during limit-count plugin execution.

### Environment

APISIX version: master (3.16.0)
OS: Linux
OpenResty version: 1.29.2.3
apisix/plugins/limit-count/init.lua
Line: 262

suryaparua-official:
## Summary

Fixes #13357

## Changes

In `apisix/plugins/limit-count/init.lua` at line 262, the `error()` 
function was called with incorrect arguments:

```lua
-- Before (wrong)
error("failed to generate key invalid parent: ", core.json.encode(parent))

-- After (correct)
error("failed to generate key invalid parent: " .. core.json.encode(parent))
```

Lua's `error(message, level)` expects the second argument to be an 
integer (stack level: 0, 1, or 2), not a string. The previous code 
would cause LuaJIT to raise:
> bad argument #2 to 'error' (number expected, got string)

## Test

Added a test case in `t/plugin/limit-count5.t` to verify the error 
message is correct when `_meta.parent` is missing.

### DIFF
--- a/apisix/discovery/dns/init.lua
+++ b/apisix/discovery/dns/init.lua
@@ -78,7 +78,7 @@ function _M.init_worker()
 
     local client, err = core.dns_client.new(opts)
     if not client then
-        error("failed to init the dns client: ", err)
+        error("failed to init the dns client: " .. (err or "unknown"))
         return
     end
 


t/plugin/limit-count5.t
```diff
@@ -307,3 +307,45 @@ peek3: 3
 commit2: 0
 peek4: 0
 commit3: rejected
+
```
shreemaan-abhishek: we need three empty lines in between tests.

hachi029:
A similar error occurs in apisix/discovery/dns/init.lua:81

suryaparua-official:
Hi @hachi029, thank you for pointing that out! I will fix the similar issue in apisix/discovery/dns/init.lua:81 and include it in this PR.

t/plugin/limit-count5.t
```diff
@@ -307,3 +307,47 @@ peek3: 3
...
+            }
+            local ctx = {
+                var = {remote_addr = "127.0.0.1"},
+                conf_version = 1,
+                conf_type = "route"
+            }
+            local ok, err = pcall(init.rate_limit, conf, ctx, "limit-count", 1)
+            if not ok then
+                -- error message should contain the JSON of parent (null)
+                if err:find("failed to generate key invalid parent") then
```
nic-6443: The comment on the line above says the message should contain the JSON of parent, but the assertion only checks the prefix `failed to generate key invalid parent` — which would pass even if the `.. core.json.encode(parent)` part got lost. Since gluing the encoded parent onto the message is exactly what this PR changes, I'd assert the full message, e.g. `err:find("failed to generate key invalid parent: null", 1, true)`, so the test also locks in that `core.json.encode(nil)` ends up in the output.

AlinsRan: This is outdated now — I slimmed the PR down to just the DNS `error()` fix, so the only file left is `apisix/discovery/dns/init.lua`. The limit-count change and TEST 9 this comment refers to were dropped, nothing to assert here anymore. Resolving.

nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13372, opened 2026-05-14
feat(plugin): add graphql-limit-count plugin

AlinsRan:
## What does this PR do?

Add a new `graphql-limit-count` plugin that limits GraphQL request rates based on the depth of the query Abstract Syntax Tree (AST) within a given time window.

## Why is this change needed?

Standard rate limiting counts every request equally, regardless of its computational cost. This is insufficient for GraphQL endpoints where the cost of execution scales with query complexity. A single deeply nested query can be far more expensive than dozens of simple ones:

```graphql
# Shallow query: depth 1, cheap to execute
query { users }

# Deep query: depth 5, potentially expensive (N+1 queries, large result sets)
query { users { orders { items { product { category { name } } } } } }
```

Without depth-aware limiting, a client could exhaust server resources with a single request while staying well within a request-count limit.

The `graphql-limit-count` plugin solves this by using the **GraphQL query AST depth as the cost** for each request, powered by the existing `limit-count` infrastructure. This lets operators configure a single depth budget per time window and have it consumed proportionally by query complexity.

## How does it work?

1. The plugin intercepts `POST` requests with `application/json` (containing a `query` field) or `application/graphql` content type.
2. It parses the request body to extract the GraphQL query string.
3. It calls `graphql.parse()` to build the AST and recursively counts the maximum `selections` nesting depth.
4. Fragment spreads (`...FragName`) and inline fragments (`... on Type { ... }`) are fully expanded before depth is computed, with cycle detection to prevent infinite loops.
5. It calls `limit_count.rate_limit(conf, ctx, plugin_name, depth)`, passing the depth as the cost.
6. If the accumulated cost within the time window exceeds `count`, the request is rejected with the configured `rejected_code`.

## Example

Configure a route to allow a cumulative query depth of 10 per minute per client IP:

```shell
curl http://127.0.0.1:9180/apisix/admin/routes -X PUT \
  -H "X-API-KEY: ${admin_key}" \
  -d '{
    "id": "graphql-rate-limited",
    "uri": "/graphql",
    "plugins": {
      "graphql-limit-count": {
        "count": 10,
        "time_window": 60,
        "rejected_code": 429,
        "key": "remote_addr"
      }
    },
    "upstream": {
      "type": "roundrobin",
      "nodes": { "127.0.0.1:1980": 1 }
    }
  }'
```

A depth-4 query consumes 4 out of 10. After the budget is exhausted:

```shell
curl -i http://127.0.0.1:9080/graphql \
  -d '{"query": "query { foo { bar { baz { id } } } }"}'

# HTTP/1.1 200 OK
# X-RateLimit-Limit: 10
# X-RateLimit-Remaining: 6

# (after quota exhaustion)
# HTTP/1.1 429 Too Many Requests
```

Fragment spreads are expanded before computing depth, so the following query has the same depth (5) as the equivalent inline query:

```graphql
fragment DeepFields on User {
  orders { items { product { category { name } } } }
}
query { user { ...DeepFields } }
```

Chained fragments (A spreading B spreading C) are also handled correctly.

## Changes

| File | Change |
|---|---|
| `apisix/plugins/graphql-limit-count.lua` | New plugin (priority 1004) with AST depth calculation, fragment expansion, cycle detection, and configurable max body size |
| `t/plugin/graphql-limit-count.t` | 26 test cases covering normal flow, fragment spreads, inline fragments, cycle detection, charset content-type, max body size, and Redis/Redis-cluster policies |
| `apisix/cli/config.lua` | Register plugin; add `plugin-graphql-limit-count` and `plugin-graphql-limit-count-reset-header` shared dict defaults |
| `apisix/cli/ngx_tpl.lua` | Conditional shared dict declarations for the plugin |
| `conf/config.yaml.example` | Register plugin and shared dicts |
| `docs/en/latest/plugins/graphql-limit-count.md` | English documentation with examples |
| `docs/zh/latest/plugins/graphql-limit-count.md` | Chinese documentation with examples |
| `docs/en/latest/config.json` | Add to Traffic sidebar |
| `docs/zh/latest/config.json` | Add to 流量控制 sidebar |
| `t/admin/plugins.t` | Register plugin in admin test list |


nic-6443 (dismissed):

membphis (dismissed):

membphis:
`node_depth` currently walks the raw AST and takes the maximum `selections` nesting it sees, but it does not expand a `FragmentSpread` at the point where the fragment is used. This lets clients move deeply nested fields into a fragment and reference it from a shallow query, causing the calculated cost to be lower than the actual query depth.

Example:

```graphql
query {
  user {
    ...Deep
  }
}

fragment Deep on User {
  posts {
    comments {
      author {
        id
      }
    }
  }
}
```

The plugin should first build a fragment definition map, then calculate depth from the operation selection set. When it encounters a `FragmentSpread`, it should expand it in place; when it encounters an `InlineFragment`, it should continue calculating from that selection set; and it should guard against fragment cycles. Please also add tests for depth accounting with fragment spreads and inline fragments.

shreemaan-abhishek (approved):

nic-6443 (approved):

membphis (approved):

PR #13373, opened 2026-05-14
feat: enable upstream PROXY protocol per TCP port

### LINKED ISSUE #12273: help request:Is there a way to selectively disable the proxy protocol for stream ports while keeping it enabled globally
### Description

Hi,
I'm working with the following configuration:

proxy_protocol:   
  enable_tcp_pp_to_upstream: true

stream_proxy:                  
  only: false                 
  tcp:                         
    - addr: 25
    - addr: 465

My goal is to enable the proxy protocol for most ports, but disable it specifically for port 465.
Is there a way to selectively disable the proxy protocol for stream ports while keeping it enabled globally?
Any guidance would be appreciated!
Thanks in advance.

### Environment

- APISIX version (run `apisix version`): v3.3
- Operating system (run `uname -a`): Linux reg.wps.lan 3.10.0-1127.el7.x86_64 #1 SMP Tue Mar 31 23:36:51 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): nginx/1.26.3

azambranogalbis:
### Description

Feature to allow enabling upstream PROXY protocol specifically for a selection TCP ports (stream_tcp).
In order to do so, the LUA script creating nginx.conf has been modified accordingly.

Fixes https://github.com/apache/apisix/issues/12273

### Checklist

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -217,15 +217,48 @@ stream {
             require("resty.events.compat").run()
         }
     }
-
+    {% for _, item in ipairs(stream_proxy.tcp or {}) do %}
     server {
-        {% for _, item in ipairs(stream_proxy.tcp or {}) do %}
         listen {*item.addr*} {% if item.tls then %} ssl {% end %} {% if enable_reuseport then %} reuseport {% end %} {% if proxy_protocol and proxy_protocol.enable_tcp_pp then %} proxy_protocol {% end %};
-        {% end %}
-        {% for _, addr in ipairs(stream_proxy.udp or {}) do %}
-        listen {*addr*} udp {% if enable_reuseport then %} reuseport {% end %};
+        
+        {% if tcp_enable_ssl then %}
+        ssl_certificate      {* ssl.ssl_cert *};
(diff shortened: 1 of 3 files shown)


Baoyuantop (commented):
Issue #12273 asks for keeping `proxy_protocol.enable_tcp_pp_to_upstream: true` globally while disabling upstream PROXY protocol for a specific stream TCP port. In the current template the condition is effectively `enable_tcp_pp_to_upstream or item.proxy_protocol`, so once the global option is enabled every TCP server block will still get `proxy_protocol on;`. That does not fix the "globally enabled, one port disabled" path.

Please define the per-port override semantics explicitly. For example, a TCP item could be allowed to set `proxy_protocol: false` to override the global setting, or the design could switch to a fully per-port enablement model with clear migration documentation. Also please add `proxy_protocol` to the `stream_proxy.tcp` item schema in `apisix/cli/schema.lua`, update `conf/config.yaml.example`, and add regression tests for `enable_tcp_pp_to_upstream: true` with multiple TCP ports where one port is disabled/overridden. The current tests only count the number of `proxy_protocol on` directives, which is not enough to prove the directive is generated in the correct per-port server block.

PR #13375, opened 2026-05-15
fix(log-rotate): reopen logs after partial rotation

### LINKED ISSUE #13368: bug: The logrotate plugin causes custom plugin logs not to be printed
### Current Behavior

After configuring the logrotate plugin, after the logs are cut, the core.log.info of the custom plugin cannot output logs, and access.log also has no log output

Image

### Expected Behavior

_No response_

### Error Logs

_No response_

### Steps to Reproduce

After restarting apisix, the log starts printing again

### Environment

juzhiyuan:
### Description

Fixes #13368.

This updates `log-rotate` so a partial rotation failure does not prevent Nginx from reopening the log files that were rotated successfully. The regression test covers the reported behavior where a custom plugin writes through `core.log.info` after rotation and access logs should continue landing in the current `access.log`.

### Root cause

`rotate_file()` iterates over the configured log files in `apisix/plugins/log-rotate.lua`. Before this patch, if `rename_file()` returned `nil` for any configured file, `rotate_file()` returned immediately from the loop before reaching the USR1 reopen step.

That means this sequence could happen:

1. `error.log` is renamed to a timestamped file.
2. `access.log` is missing, so `rename_file()` returns `nil`.
3. `rotate_file()` returns before sending USR1 to the master process.
4. The master keeps writing through stale file descriptors, so later plugin logs and access logs do not appear at the canonical current log paths.

### Fix

- Continue rotating the remaining configured files when one file is missing or cannot be renamed.
- Track only files that actually have a rotated target.
- Send USR1 if at least one log file was rotated.
- Keep the degenerate both-files-missing case as a no-op by skipping USR1 when there are no rotated files.

### Tests

- Added a regression test in `t/plugin/log-rotate.t` using three pipelined requests:
  - `/setup`: configure a log-phase `serverless-post-function`, remove `access.log`, wait for partial rotation, then disable `log-rotate` to stabilize assertions.
  - `/hello`: trigger both plugin error-log output and access-log output.
  - `/verify`: assert the current `error.log` contains the plugin marker and current `access.log` contains the request.
- The test explicitly avoids a single-request assertion shape because log-phase plugins run after the response has been sent.

Validation run:

```bash
PATH=/usr/local/openresty/bin:/usr/local/openresty/nginx/sbin:/usr/local/openresty/luajit/bin:$PATH \
FLUSH_ETCD=1 prove -Itest-nginx/lib -I. t/plugin/log-rotate.t
```

Result after the fix:

```text
t/plugin/log-rotate.t .. ok
All tests successful.
Files=1, Tests=23, Result: PASS
```

Additional manual check:

- Temporarily changed the regression setup to remove `error.log` while leaving `access.log` present; the same command passed all 23 tests after the fix.

### DIFF
--- a/t/plugin/log-rotate.t
+++ b/t/plugin/log-rotate.t
@@ -28,6 +28,12 @@ add_block_preprocessor(sub {
     my $extra_yaml_config = <<_EOC_;
 plugins:                          # plugin list
   - log-rotate
+  - serverless-post-function
+
+nginx_config:
+  error_log_level: info
+  http:
+    access_log_buffer: 1
 
 plugin_attr:
   log-rotate:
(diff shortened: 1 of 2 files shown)


t/plugin/log-rotate.t
```diff
@@ -216,3 +222,116 @@ true
...
+  admin_key: null
+plugins:
+  - serverless-post-function
+nginx_config:
+  error_log_level: info
+  http:
+    access_log_buffer: 1
+            ]]
+            require("lib.test_admin").set_config_yaml(data)
+            code, _, body = t('/apisix/admin/plugins/reload',
```
nic-6443, line 283: The fix itself looks right to me, but I think this disable step can flake on slow runners: `plugins/reload` just posts an event and returns (`post_reload_plugins` in `admin/init.lua`), it does not wait for the privileged agent — where the log-rotate timer actually runs — to process the event and unregister the timer. With `interval: 1`, a late rotation tick can still fire after `/setup` returns; if it lands between `/hello`'s log-phase write and `/verify`'s read (that window spans a 1s boundary thanks to the 0.5s sleep), both logs get rotated away to fresh empty files and `/verify` fails on the missing marker. TEST 4 in this file tolerates exactly this uncertainty with a `<= 1` assertion. Maybe make `/setup` wait until rotation has provably stopped before returning, e.g. record the set of `__error.log` files after the reload, sleep just past one interval, and loop until no new rotated file shows up.

t/plugin/log-rotate.t
```diff
@@ -28,6 +28,12 @@ add_block_preprocessor(sub {
     my $extra_yaml_config = <<_EOC_;
 plugins:                          # plugin list
   - log-rotate
+  - serverless-post-function
+
+nginx_config:
```
nic-6443: These two `nginx_config` entries do not actually take effect in tests: the test server's nginx.conf is generated by `t/APISIX.pm`, which hardcodes an unbuffered `access_log logs/access.log main;` and already sets `log_level('info')`. Nothing at runtime reads `error_log_level` / `access_log_buffer` from config.yaml either — they are only consumed by the CLI template, which the test framework bypasses. The test works because the framework's access log is unbuffered to begin with. I would drop them (here and from the `set_config_yaml` payload in `/setup`) so they do not read as load-bearing.

t/plugin/log-rotate.t
AlinsRan: Fixed. /setup now waits until rotation has provably stopped before returning: it counts __error.log files and only returns once that count stays unchanged for two full intervals, so a stray tick can't land between /hello and /verify anymore. Confirmed stable across repeated local runs.

t/plugin/log-rotate.t
AlinsRan: Dropped both here and in the /setup payload. You're right — t/APISIX.pm hardcodes log_level(info) and an unbuffered access_log, and config.yaml's nginx_config only feeds the CLI template which the test framework bypasses, so these keys never took effect.

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13379, opened 2026-05-15
feat(jwt-auth): support key_claim_name in JWT header with tests and docs

esistfarouk:
# Description

### Summary
Currently, the `jwt-auth` plugin strictly searches for the `key_claim_name` (used to identify the Consumer) within the **JWT Payload**. However, many identity providers and standard JWT implementations place the Key ID (`kid`) or issuer (`iss`) solely in the **JWT Header**.

This PR enhances the plugin to fall back to the JWT **Header** when the configured key_claim_name claim is absent from the **Payload**, making jwt-auth compatible with standard OIDC and third-party JWT providers without requiring payload modifications.

### Why this is needed
In many security architectures, the **Header** contains the metadata required to identify which key or secret should be used to verify the signature. By restricting the lookup to the Payload, APISIX currently forces developers to either:
1. **Non-standardly duplicate** the `kid` into the Payload.
2. **Write custom serverless functions** or modify core code to handle standard JWTs.

This change makes the `jwt-auth` plugin more robust and "plug-and-play" for modern authentication flows.

### Changes
- **Plugin Logic:** Modified `find_consumer` in `apisix/plugins/jwt-auth.lua`.
- **Fallback Mechanism:** Updated the `user_key` lookup to check the JWT Header if the claim is missing from the Payload.
- **Backward Compatibility:** Maintained existing behavior by prioritizing the Payload if the claim exists in both locations.
- **Tests:** Added test cases to `t/plugin/jwt-auth.t` covering:
  * claim in JWT Header only -> Consumer identified, request succeeds
  * claim in JWT Payload only -> backward compatibility confirmed
  * claim in both Header and Payload -> Payload value takes precedence
  * claim missing from both -> original error message preserved
- **Docs:** Updated key_claim_name attribute description and added a new
  example section in both English and Chinese documentation.

### Verification Results
- [x] **Test Case 1:** JWT with `kid` in Header only (HS512) -> **Passed** (Consumer identified, Signature verified).
- [x] **Test Case 2:** JWT with claim in Payload only -> **Passed** (Backward compatibility confirmed).
- [x] **Test Case 3:** JWT with claim in both Header and Payload -> **Passed** (Payload prioritized).

### DIFF
--- a/apisix/plugins/jwt-auth.lua
+++ b/apisix/plugins/jwt-auth.lua
@@ -306,7 +306,7 @@ local function find_consumer(conf, ctx)
                    ", key claim name: ", conf.key_claim_name)
 
     local key_claim_name = conf.key_claim_name
-    local user_key = jwt.payload and jwt.payload[key_claim_name]
+    local user_key = (jwt.payload and jwt.payload[key_claim_name]) or (jwt.header and jwt.header[key_claim_name])
     if not user_key then
         return nil, nil, "missing user key in JWT token"
     end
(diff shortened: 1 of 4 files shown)


Baoyuantop (commented):
This changes the consumer lookup semantics of `jwt-auth`. Today the code only reads `key_claim_name` from the JWT payload, and the docs also describe it as a payload claim. After changing this to payload-first plus header fallback, the new behavior needs to be defined in the docs and covered by committed regression tests, not only mentioned in the PR description.

Could you please add at least these test cases?

1. `key_claim_name = "kid"` with the key only in the JWT header authenticates successfully.
2. The existing payload-only behavior still works.
3. When both payload and header contain the same claim with different values, payload priority is tested.
4. When both payload and header miss the claim, the existing `missing user key in JWT token` behavior is preserved.
5. Update the English and Chinese `jwt-auth` docs to explain the lookup locations and priority, and clarify that the header value is still used to look up an APISIX `jwt-auth` consumer `key`; this is not full OIDC/JWKS `kid` key discovery.

PR #13380, opened 2026-05-17
feat(plugin): add error-page plugin

AlinsRan:
## Summary

This PR introduces the `error-page` plugin, which allows customizing the response body and content type for HTTP error responses generated by APISIX itself (e.g., when no route matches or when the upstream is unreachable). Responses from upstream services are not intercepted.

## Configuration

The plugin is configured via plugin metadata (global config, no per-route attributes needed). Set `enable: true` and configure one or more status codes:

```json
{
    "enable": true,
    "error_404": {
        "body": "<html><body><h1>404 - Not Found</h1></body></html>",
        "content_type": "text/html"
    },
    "error_500": {
        "body": "{\"code\": 500, \"message\": \"Internal Server Error\"}",
        "content_type": "application/json"
    }
}
```

## Key design points

- **Supported status codes**: `error_404`, `error_500`, `error_502`, `error_503` with built-in default HTML pages
- **Global on/off**: controlled by `enable` in plugin metadata (default: `false`)
- **Custom content-type**: each status code can have its own `body` and `content_type`
- **Upstream responses unaffected**: only APISIX/nginx-generated error responses (where `upstream_status` is not set) are intercepted
- **Graceful fallback**: if `enable` is `false` or no custom page is configured for a status code, the default nginx error page is used
- **Opt-in plugin**: not enabled by default; users must add it to their `conf/config.yaml`

## Changes

- `apisix/plugins/error-page.lua`: plugin implementation
- `conf/config.yaml.example`: add `error-page` as a commented-out opt-in entry
- `t/plugin/error-page.t`: 18 test cases
- `docs/en/latest/plugins/error-page.md`: English documentation
- `docs/zh/latest/plugins/error-page.md`: Chinese documentation
- `docs/en/latest/config.json`: add to sidebar under Transformation
- `docs/zh/latest/config.json`: add to sidebar under Transformation


nic-6443 (dismissed):

membphis:
`get_metadata` currently skips customization whenever `ctx.var.upstream_status` is set, but `$upstream_status` can also be set for NGINX proxy-layer failures such as connection refused or upstream timeout. Those are not real upstream application responses.

As a result, `error-page` will miss common gateway-generated 502/503 responses, even though the PR description explicitly lists “upstream is unreachable” as a supported scenario.

Could we use `core.response.get_response_source(ctx)` here and only skip when the source is `"upstream"`? This would still avoid rewriting real upstream responses, while allowing APISIX/NGINX-generated errors to be customized. Please also add a 502/503 test with an unreachable or timeout upstream.

shreemaan-abhishek (approved):

nic-6443 (approved):

membphis (approved):

PR #13390, opened 2026-05-19
fix: correct loggly syslog emergency severity

moying2026:
### Description

This PR corrects the syslog emergency severity name used by the `loggly` plugin and the RFC5424 utility from `EMEGR` to `EMERG`.

`EMERG` is the standard syslog severity name for severity value `0`. The current typo makes `severity: "EMERG"` fail schema validation and the docs currently advertise the misspelled value.

### Checklist

- [x] Correct `loggly` severity enum from `EMEGR` to `EMERG`
- [x] Correct `apisix.utils.rfc5424` severity map from `EMEGR` to `EMERG`
- [x] Update English and Chinese loggly plugin docs
- [x] Add a regression case to ensure `EMERG` is accepted and the old misspelling is rejected

### Tests

- [x] `git diff --check`
- [ ] `make test TEST_FILE=t/plugin/loggly.t` (not run successfully in my local environment because OpenResty is not installed: `WARNING: OpenResty not found`)


moying2026:
补充一下背景：这个 PR 修正了 loggly 插件和 rfc5424 工具里 syslog emergency 级别的拼写错误（EMEGR → EMERG），同时更新了中英文文档，并补了一个回归测试。本地因为没装 OpenResty，没能跑完整套测试，但改动范围很小，只涉及字符串修正和文档，不影响其他逻辑。麻烦有空时帮忙 review，谢谢。

Baoyuantop (commented):
1. The diff does not only add `EMERG`; it removes the existing `EMEGR` key. Since both `severity` and `severity_map` in the loggly plugin reuse the same severity enum, users who configured `EMEGR` according to the current docs would hit schema validation failures after upgrading. I suggest keeping `EMEGR` as a deprecated alias while documenting and recommending `EMERG`. If we intentionally want to reject the old spelling, the PR should explicitly call out this compatibility impact.
2. The new test currently covers only the top-level `severity` field. Please also add coverage for `severity_map` with `EMERG`, and add a regression case for `apisix.utils.rfc5424.encode()` using `EMERG`, so the utility-level change is covered as well.

moying2026:
👋 Friendly ping — this PR has been open for a while without any review activity. Would any maintainer be able to take a look when they have a chance? Thank you!

PR #13391, opened 2026-05-19
perf(ai-proxy): optimize SSE decoder - remove PCRE, add decode_buf, fix comment lines

AlinsRan:
## What

Optimize the AI proxy SSE streaming hot path: fewer allocations, less CPU per chunk, and a new interval-flush mode for bursty LLM streams.

## Changes

| # | Change | Approximate Benefit |
|---|--------|---------------------|
| 1 | Replace `ngx_re.split` (PCRE) with `string.find` in SSE decoder | Eliminates PCRE JIT compilation on every chunk (~1–5 µs/chunk saved) |
| 2 | New `decode_buf(buf)` — combines `split_buf` + `decode` in one forward pass | ~50% fewer byte-scan iterations per chunk on the hot path |
| 3 | Fix SSE comment lines (`:`) being emitted as empty events | Correctness fix; providers using `: keep-alive` pings no longer produce spurious events |
| 4 | Replace `buf = buf .. chunk` with a 2-element table + `table.concat` | O(n) → O(1) allocation per chunk; a 100-chunk stream drops from ~250 KB of copies to ~5 KB |
| 5 | Add `streaming_flush_interval_ms` config option (default `10` ms) | Reduces `ngx.flush` syscalls by ~95% on bursty streams (100 chunks → 2–4 flushes at 10 ms) |
| 6 | Add `no_flush` param to `lua_response_filter` | Enables background-thread flush ownership (required by #5) |

## `streaming_flush_interval_ms` example

```json
{
  "provider": "openai",
  "streaming_flush_interval_ms": 10
}
```

| Value | Behavior |
|-------|----------|
| `10` (default) | Background thread calls `ngx.flush(false)` every 10 ms — batches output on bursty streams |
| `> 0` (e.g. `50`) | Background thread flushes every 50 ms; `lua_response_filter` skips `ngx.flush` per chunk |
| `0` | Background thread disabled; each chunk flushed synchronously via `ngx.flush(true)` |

The background thread only flushes when there is new data (via a `needs_flush` flag), so empty intervals cost nothing.


apisix/plugins/ai-providers/base.lua
```diff
@@ -392,7 +392,12 @@ function _M.parse_streaming_response(self, ctx, res, target_proto, converter, co
     local body_reader = res.body_reader
     local contents = {}
     local sse_state = { is_first = true }
-    local sse_buf = ""
+    -- SSE framing buffer: accumulate chunks with table.insert to avoid
+    -- allocating a new string on every append; reset to {remainder} after
+    -- each split so the table never grows beyond two elements.
+    -- Initialized with "" so the fast-path (sse_parts[1] == "") activates
+    -- immediately on the first chunk, avoiding an unnecessary table.concat.
+    local sse_parts = {""}
```
shreemaan-abhishek, line 400: very nice 👍🏼

nic-6443 (dismissed):

membphis:
@AlinsRan 

Changing the default to 10ms looks fine, but a few places still need to be updated so the behavior is documented consistently:

The PR description still says 0 (default); it should say the default is 10ms, and 0 disables background interval flushing.
The comment in base.lua still says == 0 (default), but the schema default is now 10.
The configuration docs should explicitly state what 0 means: streaming_flush_interval_ms = 0 uses per-chunk synchronous flushing, i.e. ngx.flush(true); values > 0 use the background thread to flush periodically with ngx.flush(false).
The tests currently cover explicit 0 and 50; it would be good to add one case where the field is omitted and the 10ms default is used.

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

PR #13398, opened 2026-05-20
feat: support filtering the key and value of labels

okaybase:
### Description
admin-api, only support filtering labels' key currently, however, in more scenarios, filtering is required based on both key and value, so support filtering the key and value of labels.

### Checklist

### DIFF
--- a/apisix/admin/v3_adapter.lua
+++ b/apisix/admin/v3_adapter.lua
@@ -174,8 +174,8 @@ local function filter(body, args, resource)
         if args.label then
             label_matched = false
             if body.list[i].value.labels then
-                for k, _ in pairs(body.list[i].value.labels) do
-                    if k == args.label then
+                for k, v in pairs(body.list[i].value.labels) do
+                    if k == args.label or k .. ":" .. v == args.label then
                         label_matched = true
                         break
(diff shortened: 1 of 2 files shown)


Baoyuantop (commented):
First, this changes a public Admin API query semantic, but the documentation is not updated. `docs/en/latest/admin-api.md` and `docs/zh/latest/admin-api.md` currently only say that `label` filters by resource labels. They do not describe the new `label=key:value` format, whether value-only matching is still unsupported, or how key-only and key/value matching interact.

Second, the implementation uses `k .. ":" .. v == args.label` to represent key/value matching, but the label schema does not forbid `:` in either label keys or label values. This makes a query such as `label=a:b:c` ambiguous, and it can also overlap with the existing key-only matching space. Please define the query grammar first: if `:` is the delimiter, the docs and tests should specify how `:` in keys/values is handled; if `:` is allowed, we may need parsing or escaping rules instead of plain string concatenation.

Please add documentation and boundary tests before merging. At minimum, the tests should cover a wrong value not matching, value-only matching remaining unsupported, the old key-only behavior still working, and the expected behavior when keys or values contain `:`.

PR #13399, opened 2026-05-20
docs: sign Digest header in hmac-auth body validation example

### LINKED ISSUE #13395: docs: hmac-auth body validation example should sign the Digest header
### Description

This is a small, self-contained documentation improvement for the `hmac-auth` plugin docs.

**File:** `docs/en/latest/plugins/hmac-auth.md`
**Location:** the body-validation example block (`hmac-sig-digest-header-gen.py`, currently around lines 871-921).

### Current example

The example demonstrates the `validate_request_body: true` configuration. In the Python helper:

- The `signing_string` is built from three lines only:
  ```
  {key_id}\n
  {request_method} {request_path}\n
  date: {gmt_time}\n
  ```
- The `Authorization` header lists `headers="@request-target date"`.
- A `Digest` header (`SHA-256=<base64 body digest>`) is computed and sent in the request.

So the `Digest` header is sent, but it never appears in the signed `headers="..."` list and is not part of the `signing_string`. As written, the example produces a setup where the `Digest` header is not bound to the HMAC signature.

### Why this is worth fixing

A reader copying this example as a starting point for body validation ends up with an example that is incomplete and slightly misleading: it looks like it demonstrates end-to-end body integrity, but the `Digest` header it sends is not actually covered by the signature. The example should reflect the intended usage, where the body digest is part of what gets signed.

### Proposed fix

Update the example so it signs the `Digest` header:

1. Compute the SHA-256 body digest first.
2. Append `digest: SHA-256=<base64 body digest>\n` to the `signing_string` (lowercased key, `<key>:<space><value>` format, consistent with the existing `date:` line).
3. Add `digest` to the `headers="..."` list in the `Authorization` header, e.g. `headers="@request-target date digest"`.
4. Update the printed example output accordingly.

It would also help to add a short note in the prose explaining that `validate_request_body` only checks the `Digest` header against the request body, so operators who want the body itself bound to the signature should also include the `Digest` header in the signed headers list (or enforce it via the `signed_headers` config option).

### Good first issue

This is a good first issue: it is a small, contained docs-only change to a single example block plus a short prose note. No code changes are required.

moying2026:
The body validation example under the `hmac-auth` plugin currently computes a Digest header but does not sign it. This leads to a setup where the body integrity is not bound to the HMAC signature, which can mislead users.

Changes:
- Move body digest computation before signature generation.
- Append the `digest` header to the signing string.
- Include `digest` in the `headers` list of the Authorization header.
- Update the printed example output accordingly.

This ensures the example correctly demonstrates end‑to‑end body integrity, where the body digest is part of the signed material.

Resolves #13395

### DIFF
--- a/docs/en/latest/plugins/hmac-auth.md
+++ b/docs/en/latest/plugins/hmac-auth.md
@@ -617,6 +617,10 @@ You should see an `HTTP/1.1 200 OK` response and notice the `Authorization` head
 
 The following example demonstrates how to enable body validation to ensure the integrity of the request body.
 
+:::note
+`validate_request_body` only checks the `Digest` header against the request body; it does not automatically bind the body to the HMAC signature. To ensure the body itself is covered by the signature (preventing tampering without detection), you **must** include the `Digest` header in the signed headers list (the `headers` field in the `Authorization` header). This example shows how to do that.
+:::
+
 <Tabs
 groupId="api"
 defaultValue="admin-api"
@@ -907,6 +911,7 @@ body_digest = hashlib.sha256(body.encode('utf-8')).digest()
 body_digest_base64 = base64.b64encode(body_digest).decode('utf-8')
 
 # construct the request headers
+# Note: To ensure the body is bound to the signature, include the Digest header in the signed headers list.
 headers = {
     "Date": gmt_time,
     "Digest": f"SHA-256={body_digest_base64}",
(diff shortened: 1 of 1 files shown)


moying2026:
请问有空 review 吗？🙏

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？ 🙏

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

moying2026:
请问有空 review 吗？

PR #13406, opened 2026-05-21
perf(ai): reuse raw request body when unchanged

nic-6443:
This avoids re-encoding the ai-proxy request body when the outgoing body is unchanged. If protocol conversion, provider preparation, model options, request body overrides, or provider cleanup changes the body, the existing encoded table path is still used.

It also covers ai-proxy-multi through the shared provider build path and keeps retry behavior from falling back to the original raw body after an earlier attempt has rewritten the parsed body. The raw body is read lazily in the unchanged-body branch and cached on the request context, so retries do not read it again.

The sidecar LLM request path used by ai-request-rewrite is explicitly marked as generated, so it continues to send the JSON-encoded LLM request instead of reusing the original client body.

Tests run:

```
luacheck apisix/plugins/ai-proxy/base.lua apisix/plugins/ai-providers/base.lua apisix/plugins/ai-protocols/openai-chat.lua apisix/plugins/ai-protocols/bedrock-converse.lua apisix/plugins/ai-protocols/passthrough.lua apisix/plugins/ai-request-rewrite.lua
perl -I. -c t/plugin/ai-proxy-request-body-override.t
git diff --check
prove -I. t/plugin/ai-proxy-request-body-override.t
```

### DIFF
--- a/t/plugin/ai-proxy-request-body-override.t
+++ b/t/plugin/ai-proxy-request-body-override.t
@@ -174,7 +174,129 @@ __DATA__
 
 
 
-=== TEST 3: llm_options: openai provider maps max_tokens to max_completion_tokens
+=== TEST 3a: ai-proxy forwards the original body when it is not rewritten
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                    "uri": "/chat",
+                    "plugins": {
+                        "ai-proxy": {
+                            "provider": "openai-compatible",
+                            "auth": { "header": { "Authorization": "Bearer t" } },
+                            "override": {
+                                "endpoint": "http://localhost:6732"
+                            },
+                            "ssl_verify": false
+                        }
+                    }
+                }]]
+            )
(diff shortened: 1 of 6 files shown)


membphis (approved):

t/plugin/ai-proxy-request-body-override.t
```diff
@@ -174,7 +174,129 @@ __DATA__
...
+            local body = cjson.decode(res.body)
+            ngx.say(body.choices[1].message.content == raw and "same body" or "body changed")
+        }
+    }
+--- response_body
+same body
+
+
+
+=== TEST 4b: ai-proxy-multi forwards the original body when it is not rewritten
```
shreemaan-abhishek, line 217: I think lint will fail here.

shreemaan-abhishek (approved):

t/plugin/ai-proxy-request-body-override.t
shreemaan-abhishek: strange that lint didn't fail here, I remember doing something similar and lint failed, lol.

AlinsRan:
```
===TEST 4a
...
===TEST 4b 
```
At this point, Lint will fail.

AlinsRan (approved):

PR #13414, opened 2026-05-21
docs(hmac-auth): sign Digest header in body validation example (#13395)

moying2026:
### Description\n\nThis PR improves the body validation example in the hmac-auth plugin documentation.\n\n**Issue:** #13395\n\n**Changes:**\n- Include the Digest header in the signing string construction\n- Add `digest` to the signed headers list in Authorization header\n- Ensure the request body digest is covered by the HMAC signature, demonstrating complete body integrity validation\n\n**Testing:**\n- Verified the updated Python example generates the correct HMAC signature\n- Confirmed with the example cURL command that requests are properly authenticated\n\n### Related\n\nFixes #13395

### DIFF
--- a/docs/en/latest/plugins/hmac-auth.md
+++ b/docs/en/latest/plugins/hmac-auth.md
@@ -895,6 +895,7 @@ signing_string = (
     f"{key_id}\n"
     f"{request_method} {request_path}\n"
     f"date: {gmt_time}\n"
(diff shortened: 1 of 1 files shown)


Baoyuantop:
Thanks for working on this. We already have #13396 covering the same issue with a smaller and more complete patch, so let's keep the review focused there and close this as a duplicate.

PR #13435, opened 2026-05-25
feat(plugin): add graphql-proxy-cache plugin

AlinsRan:
## What does this PR do?

This PR adds a new `graphql-proxy-cache` plugin that caches GraphQL query responses on disk or in memory, reusing the existing `proxy-cache` infrastructure.

### Changes

- `apisix/plugins/graphql-proxy-cache.lua` — plugin implementation
- `t/plugin/graphql-proxy-cache/graphql.t` — schema and request validation tests (19 cases)
- `t/plugin/graphql-proxy-cache/disk.t` — disk cache hit/miss tests (11 cases)
- `t/plugin/graphql-proxy-cache/memory.t` — memory cache hit/miss and consumer isolation tests (15 cases)
- `docs/en/latest/plugins/graphql-proxy-cache.md` — English documentation
- `docs/zh/latest/plugins/graphql-proxy-cache.md` — Chinese documentation
- Plugin registered in the default plugin list at priority 1009

## Motivation

GraphQL APIs have different caching semantics from plain HTTP APIs:

- The request body (query + variables) determines the response, not just the URL
- `mutation` operations have side effects and must never be cached
- Consumers authenticating to the same route should by default have isolated cache namespaces to prevent data leakage

The existing `proxy-cache` plugin works at the HTTP level and cannot distinguish query operations from mutations, nor does it handle GraphQL-over-GET (where the query is in the query string). This plugin fills that gap.

### Cache key

```
md5(conf_version + "\1" + host + "\1" + route_id + "\1" + service_id + "\1" + identity + "\1" + body)
```

- `conf_version` — invalidates the cache automatically when the plugin configuration changes
- `host` + `route_id` + `service_id` — ensures two routes with the same query body never share cache entries
- `identity` — when `consumer_isolation` is enabled (default), this is the consumer name or remote user; otherwise it is an empty string
- `body` — the normalized GraphQL query body

### Mutation bypass

After parsing the GraphQL AST, if any top-level operation is a `mutation`, the plugin sets `Apisix-Cache-Status: BYPASS` and skips the cache entirely.

### Caching infrastructure

The plugin delegates to the same `proxy-cache` disk and memory handlers (`apisix.plugins.proxy-cache.disk_handler`, `apisix.plugins.proxy-cache.memory_handler`), so all existing cache zone configuration in `config.yaml` is reused without duplication.

## Configuration

| Name               | Type    | Required | Default        | Valid values       | Description                                                                                  |
|--------------------|---------|----------|----------------|--------------------|----------------------------------------------------------------------------------------------|
| cache_strategy     | string  | No       | disk           | ["disk", "memory"] | Use `disk` for NGINX `proxy_cache`, or `memory` for a shared dict.                          |
| cache_zone         | string  | No       | disk_cache_one |                    | Must match a zone name defined in `config.yaml`.                                             |
| cache_ttl          | integer | No       | 300            | >= 1               | TTL in seconds for the `memory` strategy. Disk TTL follows upstream `Cache-Control`/`Expires`.|
| consumer_isolation | boolean | No       | true           |                    | Partition the cache by authenticated consumer identity.                                      |
| cache_set_cookie   | boolean | No       | false          |                    | Cache responses that include a `Set-Cookie` header (memory strategy only).                  |

## Static configuration

Configure at least one cache zone in `config.yaml` before enabling this plugin:

```yaml
apisix:
  proxy_cache:
    cache_ttl: 10s
    zones:
      - name: disk_cache_one
        memory_size: 50m
        disk_size: 1G
        disk_path: /tmp/disk_cache_one
        cache_levels: 1:2
      - name: memory_cache
        memory_size: 50m
```

### Cache GraphQL queries on disk (default)

```shell
# Create a route with graphql-proxy-cache
curl http://127.0.0.1:9180/apisix/admin/routes/1 \
  -H "X-API-KEY: $admin_key" -X PUT -d '
{
  "uri": "/graphql",
  "plugins": {
    "graphql-proxy-cache": {}
  },
  "upstream": {
    "type": "roundrobin",
    "nodes": { "127.0.0.1:8080": 1 }
  }
}'

# First request — cache miss
curl -s -D - http://127.0.0.1:9080/graphql \
  -H "Content-Type: application/json" \
  -d '{"query": "query { persons { name } }"}'
# Apisix-Cache-Status: MISS
# APISIX-Cache-Key: <cache-key>

# Second identical request — cache hit
curl -s -D - http://127.0.0.1:9080/graphql \
  -H "Content-Type: application/json" \
  -d '{"query": "query { persons { name } }"}'
# Apisix-Cache-Status: HIT
```

### Mutation operations always bypass the cache

```shell
curl -s -D - http://127.0.0.1:9080/graphql \
  -H "Content-Type: application/json" \
  -d '{"query": "mutation { addPerson(name: \"Alice\") { id } }"}'
# Apisix-Cache-Status: BYPASS
```

### Cache in memory with a short TTL

```shell
curl http://127.0.0.1:9180/apisix/admin/routes/1 \
  -H "X-API-KEY: $admin_key" -X PUT -d '
{
  "uri": "/graphql",
  "plugins": {
    "graphql-proxy-cache": {
      "cache_strategy": "memory",
      "cache_zone": "memory_cache",
      "cache_ttl": 60
    }
  },
  "upstream": {
    "type": "roundrobin",
    "nodes": { "127.0.0.1:8080": 1 }
  }
}'
```

### Purge a cached response

Expose the purge endpoint via the `public-api` plugin:

```shell
curl http://127.0.0.1:9180/apisix/admin/routes/graphql-purge \
  -H "X-API-KEY: $admin_key" -X PUT -d '
{
  "uri": "/apisix/plugin/graphql-proxy-cache/*",
  "plugins": { "public-api": {} }
}'
```

Then purge by strategy, route ID, and cache key:

```shell
# Purge a disk-cached entry for route 1
curl -X PURGE http://127.0.0.1:9080/apisix/plugin/graphql-proxy-cache/disk/1/<cache-key>
# HTTP 200 on success, HTTP 404 if the entry does not exist
```

### Consumer isolation

With `consumer_isolation: true` (default), each consumer gets an isolated cache namespace.  
Two consumers sending the same query will each get their own `MISS` on the first request
and their own `HIT` on subsequent requests — they never share each other's cached responses:

```shell
# alice: first request → MISS, second → HIT
curl -H "apikey: alice-key" -H "Content-Type: application/json" \
  -d '{"query":"query{persons{id}}"}' http://127.0.0.1:9080/graphql
# Apisix-Cache-Status: MISS

curl -H "apikey: alice-key" -H "Content-Type: application/json" \
  -d '{"query":"query{persons{id}}"}' http://127.0.0.1:9080/graphql
# Apisix-Cache-Status: HIT

# bob: first request → MISS (isolated from alice)
curl -H "apikey: bob-key" -H "Content-Type: application/json" \
  -d '{"query":"query{persons{id}}"}' http://127.0.0.1:9080/graphql
# Apisix-Cache-Status: MISS
```

Set `consumer_isolation: false` to let all consumers share the same cache, useful when the upstream response is not user-specific.

## Checklist

- [x] Code changes follow the existing code style
- [x] Tests added for new functionality
- [x] Documentation added (English + Chinese)
- [x] Lint passes
- [x] All CI checks pass

### DIFF
--- a/apisix/cli/config.lua
+++ b/apisix/cli/config.lua
@@ -244,6 +244,7 @@ local _M = {
     "ai-aws-content-moderation",
     "ai-aliyun-content-moderation",
     "proxy-mirror",
+    "graphql-proxy-cache",
     "proxy-rewrite",
     "workflow",
     "api-breaker",
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -679,7 +679,7 @@ http {
 
     {% if deployment_role ~= "control_plane" then %}
 
-    {% if enabled_plugins["proxy-cache"] then %}
+    {% if enabled_plugins["proxy-cache"] or enabled_plugins["graphql-proxy-cache"] then %}
     # for proxy cache
     {% for _, cache in ipairs(proxy_cache.zones) do %}
     {% if cache.disk_path and cache.cache_levels and cache.disk_size then %}
@@ -858,9 +858,8 @@ http {
             proxy_set_header   X-Forwarded-Host     $var_x_forwarded_host;
             proxy_set_header   X-Forwarded-Port     $var_x_forwarded_port;
 
-            {% if enabled_plugins["proxy-cache"] then %}
+            {% if enabled_plugins["proxy-cache"] or enabled_plugins["graphql-proxy-cache"] then %}
             ###  the following configuration is to cache response content from upstream server
-
             set $upstream_cache_zone            off;
             set $upstream_cache_key             '';
             set $upstream_cache_bypass          '';
--- a/apisix/plugins/graphql-proxy-cache.lua
+++ b/apisix/plugins/graphql-proxy-cache.lua
@@ -370,7 +370,7 @@ local function purge_hander()
 
     local conf = find_graphql_proxy_cache_conf(route_id)
     if not conf then
-        core.log.error("failed to find graph-proxy-cache conf, route_id: ", route_id)
+        core.log.error("failed to find graphql-proxy-cache conf, route_id: ", route_id)
         return core.response.exit(404)
     end
 
@@ -384,13 +384,13 @@ local function purge_hander()
             ngx.var.upstream_cache_key)
 
         if not util.file_exists(filename) then
-            core.log.error("failed to purge graphql cache, file not exits: ", filename)
+            core.log.error("failed to purge graphql cache, file not exists: ", filename)
             return core.response.exit(404)
         end
         os.remove(filename)
     else
-        local memory_hanler = memory_strategy({shdict_name = conf.cache_zone})
-        memory_hanler:purge(ngx_var.upstream_cache_key)
+        local memory_handler = memory_strategy({shdict_name = conf.cache_zone})
+        memory_handler:purge(ngx_var.upstream_cache_key)
     end
 
     return core.response.exit(200)
--- a/apisix/plugins/graphql-proxy-cache.lua
+++ b/apisix/plugins/graphql-proxy-cache.lua
@@ -222,13 +222,13 @@ function _M.access(conf, ctx)
 
     local ok, res = pcall(gq_parse, query_or_err)
     if not ok then
-        core.log.error("failed to parse graphql: ", res, ", body: ", body)
+        core.log.error("failed to parse graphql: ", res, ", body_size: ", #body)
         return 400, {message = "Invalid graphql request: failed to parse graphql query"}
     end
 
     local n = #res.definitions
     if n == 0 then
-        core.log.error("failed to parse graphql: empty query, body: ", body)
+        core.log.error("failed to parse graphql: empty query, body_size: ", #body)
         return 400, {message = "Invalid graphql request: empty graphql query"}
     end
 
--- a/apisix/plugins/graphql-proxy-cache.lua
+++ b/apisix/plugins/graphql-proxy-cache.lua
@@ -1,0 +1,399 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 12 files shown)


nic-6443 (dismissed):

shreemaan-abhishek (dismissed):

membphis:
Review notes from merge-risk check:

### [P1] `graphql-proxy-cache` can be enabled without the required NGINX proxy-cache variables/directives

- Problem: the new plugin writes/uses `upstream_cache_key`, `upstream_cache_zone`, `upstream_cache_bypass`, `upstream_no_cache`, and disk purge relies on `upstream_cache_zone_info`, but the OSS NGINX template only defines these variables and `proxy_cache*` directives when `enabled_plugins["proxy-cache"]` is true.
- Why this blocks merge: `graphql-proxy-cache` is registered as an independent plugin and the tests/docs show it being enabled directly. In custom deployments that include `graphql-proxy-cache` but not `proxy-cache`, the plugin can hit undefined NGINX variables or run without the required disk cache configuration.
- Impact: runtime failures or non-functional disk cache/purge for users with a minimal custom plugin list.
- Trigger: configure `plugins:` with `graphql-proxy-cache` but without `proxy-cache`, then enable the plugin on a route.
- Evidence: `apisix/plugins/graphql-proxy-cache.lua` sets `ctx.var.upstream_cache_key` and purge sets `ngx_var.upstream_cache_key`; `apisix/cli/ngx_tpl.lua` wraps cache variable/directive generation in `enabled_plugins["proxy-cache"]`; `t/plugin/graphql-proxy-cache/graphql.t` enables only `graphql-proxy-cache` and `public-api` in its minimal plugin list.
- Suggested fix: treat `graphql-proxy-cache` as a proxy-cache user in the template and startup validation, e.g. generate the cache variables/directives and require `apisix.proxy_cache` when either `proxy-cache` or `graphql-proxy-cache` is enabled. Add a startup/runtime test that enables only `graphql-proxy-cache`.

### [P2] Parse-error logging still writes the full GraphQL request body to error log

- Problem: malformed GraphQL requests and empty query paths still log the raw `body` value.
- Why this blocks merge: GraphQL bodies frequently contain variables or business data. A bad query should not cause request payloads to be written to `error.log`.
- Impact: sensitive request data can leak into logs for any route using this plugin.
- Trigger: send an invalid GraphQL query or an empty query body.
- Evidence: `apisix/plugins/graphql-proxy-cache.lua` logs `failed to parse graphql: ..., body: ...` and `failed to parse graphql: empty query, body: ...`.
- Suggested fix: log parser error plus `body_size`/request id only, not the raw body; add a regression test that malformed input is rejected without body content appearing in logs.

membphis (approved):

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13442, opened 2026-05-26
fix(upstream): use `cert` and `key` instead of stale `ok` in mTLS error checks

okaybase:
### Description

The branch uses a stale `ok` variable from `fill_node_info()` to check whether `fetch_cert` and `fetch_pkey` succeeded, instead of checking the `cert` and `key` variables they actually return. Since `ok` was already `true` at that point, both error branches were unreachable — if either fetch returned `false`, the function would silently proceed with an invalid value rather than returning a 503 error.

### Checklist

- [x]  I have added tests corresponding to this change


Baoyuantop (commented):
The remaining blocker is test coverage. The existing `t/node/upstream-mtls.t` covers several upstream mTLS schema validation and successful handshake paths, but it does not directly cover the runtime path in `apisix/upstream.lua` where `fetch_cert` or `fetch_pkey` fails and `set_by_route` should return 503 instead of passing an invalid value to the later TLS setup logic.

nic-6443 (approved):

AlinsRan (approved):

Baoyuantop (approved):

PR #13443, opened 2026-05-26
feat(limit-count): add Redis Sentinel backend, sliding window and delayed sync support

AlinsRan:
## Background / Motivation

The `limit-count` plugin currently supports fixed-window counting over `local`, `redis`, and `redis-cluster` backends. Three gaps come up repeatedly in production rate-limiting setups:

1. **No high-availability Redis option.** A standalone `redis` backend is a single point of failure, and `redis-cluster` is operationally heavy for users who run Redis with Sentinel-based failover.
2. **Fixed windows allow boundary bursts.** A client can spend its full quota at the end of one window and again at the start of the next, briefly doubling the intended rate.
3. **Every request is a Redis round-trip.** On high-traffic routes the per-request `INCR` to Redis becomes a latency and throughput bottleneck.

This PR addresses all three with optional, backward-compatible additions to `limit-count`.

## What's included

| Feature | New config | Applies to |
|---|---|---|
| Redis Sentinel backend | `policy: redis-sentinel` (+ `redis_sentinels`, `redis_master_name`, …) | new policy |
| Sliding window | `window_type: sliding` | `local`, `redis`, `redis-cluster`, `redis-sentinel` |
| Delayed sync | `sync_interval` | `redis`, `redis-cluster`, `redis-sentinel` |

All three are independent and composable — any combination can be enabled on a single route.

---

## Feature 1 — Redis Sentinel backend

A new `redis-sentinel` policy connects to Redis through one or more Sentinel nodes:

- Automatic master discovery and failover handling via Sentinel.
- Reuses the full existing Redis option set (database selection, ACL username/password, SSL, timeouts, keepalive).
- Separate `sentinel_username` / `sentinel_password` for Sentinel ACL, independent from the Redis data-node credentials.
- Falls back to local counting when Redis is temporarily unavailable (consistent with the other Redis backends).

### Use case: HA rate limiting shared across gateway instances

Multiple APISIX nodes enforce one global quota backed by a Sentinel-managed Redis, surviving master failover without config changes.

```jsonc
{
  "uri": "/api/*",
  "plugins": {
    "limit-count": {
      "count": 100,
      "time_window": 60,
      "key": "remote_addr",
      "rejected_code": 429,
      "policy": "redis-sentinel",
      "redis_sentinels": [
        { "host": "10.0.0.1", "port": 26379 },
        { "host": "10.0.0.2", "port": 26379 },
        { "host": "10.0.0.3", "port": 26379 }
      ],
      "redis_master_name": "mymaster",
      "redis_password": "redis-secret",
      "sentinel_password": "sentinel-secret"
    }
  },
  "upstream": { "nodes": { "127.0.0.1:1980": 1 }, "type": "roundrobin" }
}
```

---

## Feature 2 — Sliding window algorithm

A new `window_type` field selects the counting algorithm (default `fixed`, i.e. current behavior):

- **`fixed`** — the existing fixed-window counter; the count resets at each window boundary.
- **`sliding`** — a **sliding-window-counter** algorithm: time is divided into fixed slots of `time_window`, and the effective count is the current slot's count plus a time-weighted fraction of the previous slot's count. This smooths enforcement and removes the double-rate burst at window boundaries, without storing per-request timestamps.

It is available for both the local and all Redis-based backends. The local backend uses a shared dict for slot counters; the Redis backends use `INCRBY` with a per-slot TTL (no sorted sets, so memory and command cost stay constant per request).

### Use case: smooth limiting that avoids boundary bursts

```jsonc
{
  "uri": "/checkout",
  "plugins": {
    "limit-count": {
      "count": 60,
      "time_window": 60,
      "key": "remote_addr",
      "window_type": "sliding"
    }
  },
  "upstream": { "nodes": { "127.0.0.1:1980": 1 }, "type": "roundrobin" }
}
```

With `fixed`, a client could send 60 requests at second 59 and another 60 at second 61 (120 in ~2s). With `sliding`, the weighted previous-slot contribution keeps the observed rate close to 60/60s across the boundary.

---

## Feature 3 — Delayed sync (`sync_interval`)

For Redis-based policies, `sync_interval` (seconds) batches counter synchronization instead of going to Redis on every request:

- Between sync intervals, counting is served from a local shared dict; deltas are reconciled with Redis once per interval.
- Cuts Redis round-trips and tail latency on high-QPS routes, trading exactness for throughput (the global count may briefly lag by up to one interval's local delta).
- Validation: must be `>= 0.1` and smaller than `time_window`. Set to `-1` (or omit) to disable and keep the per-request sync behavior.

### Why a dedicated lock shared dict (`plugin-limit-count-lock`)

A single delayed sync is a **non-atomic** sequence: read the accumulated local delta from the shared dict, flush it to Redis, then write the returned remote quota back to the local dict. If two workers — or two concurrent requests on the same worker — synced the same key at the same time, the delta could be flushed twice or the cached remote quota clobbered, corrupting the count.

To prevent this, each sync grabs a **per-key `lua-resty-lock`** before running and releases it afterwards (the sync body is wrapped in `pcall`, so the lock is never left held if the sync errors). `lua-resty-lock` manages the lifecycle of its own entries and must not share storage with counter data, so it requires a **separate** shared dict, named by convention `<counter dict>-lock` — hence `plugin-limit-count-lock`.

This dict is provisioned by default (`10m`) and is only touched when `sync_interval` is enabled, so typical deployments need no manual configuration. It only holds short-lived lock entries, so it can stay small.

### Use case: high-throughput per-consumer quota

A busy API enforces a large per-consumer quota where occasional sub-second lag is acceptable in exchange for far fewer Redis calls.

```jsonc
{
  "uri": "/v1/*",
  "plugins": {
    "limit-count": {
      "count": 100000,
      "time_window": 60,
      "key": "consumer_name",
      "key_type": "var",
      "policy": "redis",
      "redis_host": "10.0.0.10",
      "redis_port": 6379,
      "sync_interval": 1
    }
  },
  "upstream": { "nodes": { "127.0.0.1:1980": 1 }, "type": "roundrobin" }
}
```

---

## Combined use case

The three features compose. Example: HA Redis via Sentinel, smooth sliding-window enforcement, and batched sync for throughput.

```jsonc
{
  "uri": "/api/*",
  "plugins": {
    "limit-count": {
      "count": 5000,
      "time_window": 60,
      "key": "remote_addr",
      "window_type": "sliding",
      "policy": "redis-sentinel",
      "redis_sentinels": [
        { "host": "10.0.0.1", "port": 26379 },
        { "host": "10.0.0.2", "port": 26379 }
      ],
      "redis_master_name": "mymaster",
      "sync_interval": 2
    }
  },
  "upstream": { "nodes": { "127.0.0.1:1980": 1 }, "type": "roundrobin" }
}
```

## New configuration fields

| Field | Type | Default | Valid values | Description |
|---|---|---|---|---|
| `window_type` | string | `fixed` | `fixed`, `sliding` | Counting algorithm. |
| `sync_interval` | number | (disabled) | `-1` or `>= 0.1` | Delayed-sync interval in seconds for Redis policies; `-1` disables. Must be `< time_window`. |
| `redis_sentinels` | array[object] | — | `[{host, port}, …]` | Sentinel nodes. Required for `redis-sentinel`. |
| `redis_master_name` | string | — | | Master name monitored by Sentinel. Required for `redis-sentinel`. |
| `redis_role` | string | `master` | `master`, `slave` | Role to resolve via Sentinel. |
| `sentinel_username` | string | — | | Sentinel ACL username (optional). |
| `sentinel_password` | string | — | | Sentinel ACL password (optional). |

(`redis_database`, `redis_username`, `redis_password`, and the timeout/keepalive options behave as in the existing `redis` policy.)

### Why the Redis counting changed (remaining → consumed)

Previously, for Redis policies the counter stored the **remaining quota**: on key creation it did `SET key = count - cost` and each request applied a *negative* `INCRBY`, so `count` was baked into the stored value at the moment the key was first created in a window. That scheme cannot support a `count` that varies within a window — which this PR introduces via variable `count`/`time_window` (resolved from nginx variables per request) and `rules`. Once the key existed, a later change to `count` could not be reflected, because the stored "remaining" was computed against the old `count`.

The counter now stores the **consumed** count instead: `SET key = cost` on creation, a *positive* `INCRBY` per request, and `remaining = count - consumed` computed at read time against the **current** `count`. Storing an absolute consumed value (rather than a `count`-relative remaining) is exactly what lets a changing `count` be honored correctly.

### Upgrade isolation (versioned key)

Because the stored meaning flipped (remaining → consumed) on the same Redis key namespace, a freshly upgraded node could otherwise read a pre-upgrade key with the new meaning and briefly over- or under-limit existing routes. To prevent this, the fixed-window Redis key now embeds a storage-format version, rendered as `:v1:`:

```
plugin-limit-count<key>  ->  plugin-limit-count:v1:<key>
```

Pre-upgrade keys carry no `:v1:` segment, so new code never reads an old-format value; old keys simply expire via their own TTL. The net effect is a one-time, bounded counter reset on upgrade (predictable and safe) rather than a window of misread values. Only a major bump of the version (e.g. `:v1:` → `:v2:`) would rotate keys again, so unrelated releases leave live counters intact.

### Defaults & local policy

All new fields are optional. `window_type` defaults to `fixed` (the existing fixed-window behavior) and `sync_interval` defaults to `-1` (delayed sync disabled) — consistent with every other optional `limit-count` field. The `local` policy is unaffected (its shared-dict counter resets on restart regardless).

## Tests

- `t/plugin/limit-count-redis-sentinel.t` — Sentinel parameter validation, counting, multi-route isolation, Sentinel ACL (with/without auth), connection-failure fallback.
- `t/plugin/limit-count-sliding.t` — local sliding window, header precision, and `redis-sentinel` + sliding + delayed-sync combination.
- `t/plugin/limit-count-redis-delayed-sync.t` — delayed sync across `redis`, `redis-cluster`, and `redis-sentinel`.
- Existing `limit-count` suites pass unchanged, except `t/control/services.t`, whose expected response now includes the `window_type`/`sync_interval` defaults.


membphis:
[P2] `sync_interval` validation should reject invalid window relationships, including equality and dynamic windows.

The docs say delayed sync must use a `sync_interval` smaller than `time_window`, but the current check only rejects `sync_interval > time_window`. This allows `sync_interval == time_window`, which means local deltas can be delayed for the whole window and weakens the global Redis-backed limit.

There is also a dynamic case: when `time_window` is configured as a string/template variable, schema-time validation skips the comparison entirely, so a request can resolve to a `time_window` that is less than or equal to `sync_interval` and still run delayed sync.

Please reject `sync_interval >= time_window` for static numeric windows, and add request-time validation after resolving rule/count variables so dynamic `time_window` values also satisfy `sync_interval < time_window`.

nic-6443 (dismissed):

membphis:
I think this needs a compatibility guard before merge.

The Redis fixed-window counter format changes from “remaining quota” to “consumed count”, but the Redis key namespace appears to stay the same. Existing `policy=redis` / `redis-cluster` keys created before upgrade store `limit - cost` and then decrement with negative `INCRBY`; the new implementation stores `cost` and increments positively.

During rolling upgrades, or until existing Redis keys expire, the new code can read old keys with the new meaning. That may temporarily over-limit or under-limit existing routes.

Could we either use a new Redis key namespace/version for the new format, detect/migrate old-format keys safely, or document and enforce a migration strategy that avoids reusing the same keys across the format change?

AlinsRan:
Good catch — addressed in c844c2aec.

I went with your first option (a versioned key namespace). The fixed-window Redis counter key now embeds a storage-format version, rendered as `:v1:`:

```
plugin-limit-count<key>  ->  plugin-limit-count:v1:<key>
```

Pre-upgrade keys never carry the `:v1:` segment, so the new code can never read an old-format value (remaining quota) with the new meaning (consumed count). Old keys are simply left to expire via their existing TTL (<= `time_window`) — no detection/migration step needed, and no over-/under-limit window.

nic-6443 (dismissed):

membphis (dismissed):

AlinsRan:
Follow-up: the latest commit 17a6117 fixes a global under-counting bug that affected the **sliding-window + delayed-sync** combination.

### Problem

In delayed-sync mode, requests are admitted locally and the accumulated delta is later flushed to the shared store. The flush reused the sliding-window `incoming()` path, but `incoming()` rejects **before** incrementing once the counter is at/over the limit:

```
incoming(): if over limit -> return "rejected"   (no store:incr)
```

So when a node had already locally permitted requests that pushed the window over the limit, the flush of that already-consumed delta was silently dropped instead of written to Redis. Those requests really happened, so the shared counter permanently under-records them, and the cluster-wide quota drifts above the configured limit over time.

This is the same reason the fixed-window backend never had the bug: its Redis script *always* increments first and only then decides whether to report `rejected`, so a delayed-sync flush is never lost.

### Fix

Add a `commit()` method to the sliding-window backend that unconditionally increments the counter and reports the resulting remaining, skipping the pre-increment rejection. The delayed syncer now prefers it when available:

```lua
local flush = self.limiter.commit or self.limiter.incoming
```

`commit()` mirrors the fixed-window semantics: an already-permitted delta always reaches the shared store, even when the window is already over the limit. Backends without a `commit()` fall back to `incoming()`, so behavior is unchanged for them.

Regression test added in `t/plugin/limit-count-sliding.t`: fill the window via `incoming()`, confirm a further `incoming()` is rejected with no increment, then confirm `commit()` still flushes the delta and drives `remaining` negative as expected.

nic-6443 (dismissed):

AlinsRan:
Follow-up: the latest commit 92b3460 closes a cross-plugin isolation gap in the **sliding-window Redis** counter key.

### Problem

The fixed-window Redis path prefixes its counter key with `plugin_name` (and a storage-format version), so two plugins that reuse this module never collide. The sliding-window Redis path did not — `get_counter_key()` built the key only from `gen_limit_key()` output plus the window id:

```
<resource_key>:<conf_version>:<key>.<wid>.counter
```

`gen_limit_key()` does not contain the plugin name, and `conf_version` is `crc32(conf)`. So when two plugins reuse this module on the **same resource with identical config** — `limit-count` and `graphql-limit-count` are the real-world pair — they produce the same key and share one Redis counter, double-counting each other's requests.

This only affects the Redis-backed stores (`redis` / `redis-cluster` / `redis-sentinel`). The local store is already namespaced by its per-plugin shared dict, so it was never exposed.

### Fix

Prefix the sliding-window Redis counter key with `plugin_name`, mirroring the fixed-window backend:

```lua
if self.plugin_name then
    return string_format("%s:%s.%s.counter", self.plugin_name, key, wid)
end
return string_format("%s.%s.counter", key, wid)
```

`plugin_name` is wired onto the limiter instance right after construction, the same way `fallback_limiter` already is, so no constructor signatures change. The local store passes no name and keeps its original key format unchanged.

Regression test added in `t/plugin/limit-count-sliding.t`: two limiters with different `plugin_name` but the same key — exhausting one plugin's quota leaves the other's quota intact, proving the counters are independent.

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Approved.

Issue #13456, opened 2026-05-30
bug: ai-proxy-multi plugin fails with plugin_config_id when multiple instances are configured


chryancclee-boop:
### Current Behavior

When using plugin_config_id to reference an external PluginConfig containing ai-proxy-multi with multiple instances, APISIX returns 503 with error "failed to fetch the parent config". Single instance works fine, and embedding multiple instances directly in the Route also works.

### Expected Behavior

Multi-instance PluginConfig should work via plugin_config_id reference

### Error Logs

*138335 [lua] resource.lua:63: fetch_latest_conf(): unsupported resource type: plugin_configs, client: 172.22.0.1, server: _, request: "POST /v1/chat/completions/test-multi HTTP/1.1", host: "127.0.0.1:9080", request_id: "3b90a7b3b15772d42b8b57fed5b6350e"

### Steps to Reproduce

1. Create PluginConfig with ai-proxy-multi containing 2+ instances
2. Create Route referencing it via plugin_config_id
3. Send request → 503 "failed to fetch the parent config"

### Environment

- APISIX version (run `apisix version`): 3.16.0
- Operating system (run `uname -a`): Linux 198ca90b5de7 6.8.0-110-generic #110-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 19 15:09:20 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

Baoyuantop:
Thanks for the detailed report. This looks like a real bug in `ai-proxy-multi` when it is used through `plugin_config_id`.

The reported behavior matches the current code path:

1. When `ai-proxy-multi` has only one instance, it directly selects that instance and does not fetch the parent config, so the single-instance case works.
2. When it has multiple instances, it enters the balancer path and calls `resource.fetch_latest_conf(conf._meta.parent.resource_key)`.
3. For a plugin configured via `plugin_config_id`, the parent resource key is `/plugin_configs/<id>`.
[...]

PR #13458, opened 2026-05-30
fix(resource): add plugin_configs support for fetch_latest_conf, fix ai-proxy-multi plugin with multi instance not support plugin_config_id

chryancclee-boop:
When the ai-proxy-multi plugin uses plugin_config_id to reference a PluginConfig containing multiple instances, APISIX returns a 503 error: "fetch_latest_conf(): unsupported resource type: plugin_configs".

Root cause:
- pick_target() in ai-proxy-multi.lua calls resource.fetch_latest_conf() to fetch the parent configuration for load balancing and health checking.
- When referencing via plugin_config_id, the _meta.parent.resource_key is set to "/plugin_configs/<id>".
- resource.fetch_latest_conf() previously only supported upstreams, routes, services, and stream_routes, causing it to reject "plugin_configs".

Why single instance works:
- Single instance scenarios bypass pick_target() entirely since no load balancing is needed, so resource.fetch_latest_conf() is never invoked.

Fix:
- #13456 
- Add "plugin_configs" to the supported resource types in resource.lua.

Verification:
- Single instance + plugin_config_id: 200 (unchanged)
- Multi-instance + plugin_config_id: 503 -> 200 (fixed)
- Multi-instance + embedded plugins: 200 (unchanged)

### Checklist

### DIFF
--- a/apisix/resource.lua
+++ b/apisix/resource.lua
@@ -59,6 +59,8 @@ local function fetch_latest_conf(resource_path)
         key = "/services"
     elseif resource_type == "stream_routes" then
         key = "/stream_routes"
+    elseif resource_type == "plugin_configs" then
+        key = "/stream_routes"
     else
         log.error("unsupported resource type: ", resource_type)
         return nil
--- a/apisix/resource.lua
+++ b/apisix/resource.lua
@@ -59,6 +59,8 @@ local function fetch_latest_conf(resource_path)
         key = "/services"
     elseif resource_type == "stream_routes" then
         key = "/stream_routes"
+    elseif resource_type == "plugin_configs" then
+        key = "/plugin_configs"
     else
         log.error("unsupported resource type: ", resource_type)
         return nil


Baoyuantop (commented):
Please provide the corresponding test cases.

apisix/resource.lua
```diff
@@ -59,6 +59,8 @@ local function fetch_latest_conf(resource_path)
         key = "/services"
     elseif resource_type == "stream_routes" then
         key = "/stream_routes"
+    elseif resource_type == "plugin_configs" then
```
nic-6443, line 62: This fixes `plugin_configs`, but the same 503 still exists for the other resource types that attach plugins with `_meta.parent`: `apisix/global_rules.lua`, `apisix/consumer_group.lua` and `apisix/consumer.lua` all call `plugin.set_plugins_meta_parent` too, so e.g. ai-proxy-multi with multiple instances in a global_rule fails with the exact same `unsupported resource type` error. Since every supported prefix here just maps to itself, maybe turn this if/else chain into a lookup table covering all resource types that can carry plugins, instead of adding them one at a time.

apisix/resource.lua
```diff
@@ -59,6 +59,8 @@ local function fetch_latest_conf(resource_path)
...
+        key = "/plugin_configs"
```
nic-6443, line 63: To make Baoyuantop's ask concrete: none of the `t/plugin/ai-proxy-multi*.t` files currently exercise `plugin_config_id`, so this bug had no coverage. A minimal regression block would fit in `t/plugin/ai-proxy-multi.balancer.t`: create a plugin_config with ai-proxy-multi and 2 instances, create a route referencing it via `plugin_config_id`, then assert the request succeeds — before this fix it returns 503 with `failed to fetch the parent config`. The existing multi-instance setup in that file can be reused almost as-is.

neu-hsc:
Hi, I noticed this PR already contains the minimal `resource.lua` fix for `plugin_configs`.

I prepared a local regression test for the `ai-proxy-multi + plugin_config_id + multiple instances` case discussed in #13456. The test uses `apisix_yaml` to avoid Admin API/etcd setup flakiness and the existing local AI fixture, so it does not call any external LLM service.

Local focused test result:

```bash
TEST_NGINX_BINARY=/usr/local/openresty/bin/openresty \
prove --timer -I/opt/test-nginx/lib -I. -I./t \
t/plugin/ai-proxy-multi-plugin-config.t
```

Result: `All tests successful. Files=1, Tests=4, Result: PASS`.

Would it be helpful if I open a small test-only PR based on this PR's fix, or would you prefer me to share the test patch here?

AlinsRan:
Thanks for the thorough root-cause writeup — the `fetch_latest_conf()` gap on `plugin_configs` (only reached via `pick_target()` in the multi-instance path) was diagnosed exactly right.

We ended up fixing this more broadly in #13663 (merged): rather than adding just `plugin_configs`, it covers all plugin-bearing resource types (`plugin_configs`, `global_rules`, `consumers`, `consumer_groups`) in `fetch_latest_conf` in one place, plus a regression test — which closes the same gap this PR targets and then some. #13456 is resolved by it.

Closing as superseded by #13663. Appreciate the clear diagnosis and repro.

PR #13467, opened 2026-06-03
fix(limit): atomic redis commits and resolved-var validation

### LINKED ISSUE #12592: bug: limit-req plugin does not work correctly when configuring redis or redis-cluster

shreemaan-abhishek:
### Description

This PR fixes two issues in the redis-backed rate limiting plugins.

**limit-req: non-atomic redis commit lets concurrent requests exceed the configured rate**

The redis policy commit path in `limit-req/util.lua` ran `GET excess` / `GET last` / limit check / `SET excess` / `SET last` as separate commands. Under concurrent load, multiple requests can read the same stale `excess`, each conclude they are within limits, and all be admitted before any `SET` lands, so clients can exceed the configured rate.

The commit path now runs a single redis Lua script (`EVAL`) that performs the read, leaky-bucket decay, limit check, and write atomically. The `excess`/`last` keys now share a `{...}` hash tag so they always map to the same Redis Cluster slot, which `EVAL` with multiple keys requires (`limit-req/util.lua` is shared by the redis-cluster policy). The dry-run (non-commit) path keeps two plain `GET`s since it writes nothing back. First-request admission (no prior state, `excess = 0`) is preserved.

Note on upgrades: the redis key format changes from `limit_req:<key>excess` to `{limit_req:<key>}excess`. Old keys are simply ignored and expire via their existing TTL (a few seconds), so the rate-limit state resets once when upgrading.

**limit-conn: variable-resolved conn/burst values are not validated**

`conn` and `burst` can be resolved from request variables (e.g. a header). The resolved value was only passed through `tonumber()`: values like `0`, `-1`, `1.5`, or values above 2^53-1 were accepted and fed into `floor((conn - 1) / max)`, producing wrong delay calculations. Resolved values are now rejected unless they are positive integers within the safe integer range, and error messages report the offending input.

Also adds a warning log when the redis-backed limit-conn rejects a connection, noting that connections outliving `key_ttl` expire from the tracking ZSET and can lead to undercounting of long-lived connections (e.g. WebSocket, gRPC streaming).

N/A

### Checklist

Fixes #12592

### DIFF
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -43,14 +43,29 @@ local _M = {}
 
 local function resolve_var(ctx, value)
     if type(value) == "string" then
+        local original = value
         local err, _
         value, err, _ = core.utils.resolve_var(value, ctx.var)
         if err then
-            return nil, "could not resolve var for value: " .. value .. ", err: " .. err
+            return nil, "could not resolve var for value: " .. original .. ", err: " .. err
         end
+        local resolved = value
         value = tonumber(value)
         if not value then
-            return nil, "resolved value is not a number: " .. tostring(value)
+            return nil, "resolved value is not a number: " .. tostring(resolved)
+        end
+        if value <= 0 then
+            return nil, "resolved value must be a positive number, got: " .. tostring(value)
+        end
+        -- conn/burst are used as integer operands in floor((conn - 1) / max);
+        -- fractional values produce wrong delay calculations
+        if value ~= math.floor(value) then
+            return nil, "resolved value must be an integer, got: " .. tostring(value)
+        end
+        -- LuaJIT doubles lose integer precision above 2^53 (9007199254740992)
+        if value > 9007199254740991 then
+            return nil, "resolved value exceeds safe integer range (2^53-1), got: "
+                        .. tostring(value)
         end
     end
     return value
--- a/apisix/plugins/limit-conn/init.lua
+++ b/apisix/plugins/limit-conn/init.lua
@@ -42,7 +42,7 @@ end
 local _M = {}
 
 
-local function resolve_var(ctx, value)
+local function resolve_var(ctx, value, allow_zero)
     if type(value) == "string" then
         local original = value
         local err, _
@@ -55,7 +55,14 @@ local function resolve_var(ctx, value)
         if not value then
             return nil, "resolved value is not a number: " .. tostring(resolved)
         end
-        if value <= 0 then
+        -- conn must be positive; burst may be zero, mirroring the schema where
+        -- conn has exclusiveMinimum 0 and burst has minimum 0
+        if allow_zero then
+            if value < 0 then
+                return nil, "resolved value must be a non-negative number, got: "
+                            .. tostring(value)
+            end
+        elseif value <= 0 then
             return nil, "resolved value must be a positive number, got: " .. tostring(value)
         end
         -- conn/burst are used as integer operands in floor((conn - 1) / max);
@@ -79,7 +86,7 @@ local function get_rules(ctx, conf)
         if err then
             return nil, err
         end
-        local burst, err2 = resolve_var(ctx, conf.burst)
+        local burst, err2 = resolve_var(ctx, conf.burst, true)
         if err2 then
             return nil, err2
         end
@@ -99,7 +106,7 @@ local function get_rules(ctx, conf)
         if err then
             goto CONTINUE
         end
-        local burst, err2 = resolve_var(ctx, rule.burst)
+        local burst, err2 = resolve_var(ctx, rule.burst, true)
         if err2 then
             goto CONTINUE
         end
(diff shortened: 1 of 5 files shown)


membphis (dismissed):

apisix/plugins/limit-conn/util.lua
```diff
@@ -102,6 +102,11 @@ function _M.incoming(self, red, key, commit)
         conn = res[2]
 
         if allowed == 0 then
+            core.log.warn("limit-conn: connection limit (", max, ") reached; note that ",
```
AlinsRan: I think this is expected and logs should not be printed.
And the `limit-req` does not print logs

nic-6443: agree, this warning log is printed repeatedly when the connection limiting threshold is exceeded.

shreemaan-abhishek: done

nic-6443 (dismissed):

AlinsRan (dismissed):

apisix/plugins/limit-conn/init.lua
```diff
@@ -43,14 +44,29 @@ local _M = {}
...
-            return nil, "could not resolve var for value: " .. value .. ", err: " .. err
+            return nil, "could not resolve var for value: " .. original .. ", err: " .. err
         end
+        local resolved = value
         value = tonumber(value)
         if not value then
-            return nil, "resolved value is not a number: " .. tostring(value)
+            return nil, "resolved value is not a number: " .. tostring(resolved)
+        end
+        if value <= 0 then
```
membphis: Could we split validation between `conn` and `burst` here? `burst = 0` is valid in the schema/docs (`integer`, `minimum = 0`), but this helper is also used for `conf.burst` and `rule.burst`, so a config like `burst = "${http_burst ?? 0}"` will pass schema and then fail requests with 500 once it resolves to 0 (or skip the matching rule in `rules`). `conn` should stay positive, but dynamic `burst` should allow 0 while still rejecting negative, fractional, and unsafe integer values.

shreemaan-abhishek: Good catch, you are right. `burst` allows `minimum = 0` in the schema, so a dynamic `burst` resolving to `0` should be admitted, while only `conn` needs to stay strictly positive (`exclusiveMinimum = 0`). I split the validation: `resolve_var` now takes an `allow_zero` flag, passed `true` for both `conf.burst` and `rule.burst`, so `burst = 0` is accepted while negative, fractional and unsafe-integer values are still rejected. Added tests for dynamic `burst: 0` (admitted) and `burst: -1` (rejected). Pushed in b9365968b.

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):

PR #13468, opened 2026-06-03
change(auth): require configured jwt claims, harden empty claims_to_verify and key-auth anonymous fallback

shreemaan-abhishek:
### Description

This PR contains hardening fixes in the auth plugins, each with tests.

**1. `jwt-auth`: treat an empty `claims_to_verify` array as unset**

`verify_claims()` falls back to the default claims (`exp`/`nbf`) when `claims_to_verify` is not configured. The guard used `if not claims then`, but an explicitly empty array (`"claims_to_verify": []`) is truthy in Lua, so it skipped the fallback and the validation loop iterated over nothing, meaning an expired token was accepted when `claims_to_verify` was set to `[]`. The schema allows an empty array (no `minItems`), so this configuration is reachable.

The guard now treats an empty array the same as an unset value (`if not claims or #claims == 0 then`), restoring the default `exp`/`nbf` checks. Behaviour for tokens that legitimately omit `exp`/`nbf` is unchanged (they are validated only if present).

**2. `jwt-auth`: require explicitly configured claims to be present (behavior change)**

Previously `verify_claims()` validated a configured claim only if it was present in the token (`if claim then ...`), so a route with e.g. `claims_to_verify: ["exp"]` would still accept a token that omitted `exp`. With this change, when `claims_to_verify` is explicitly configured the listed claims are **required**: a token missing a configured claim is rejected. The default path (no `claims_to_verify`, or an empty array) is unchanged and still validates `exp`/`nbf` only if present.

> [!WARNING]
> This is a behavior change. A route configured with `claims_to_verify: ["exp"]` (or any claim) will now reject a token that omits that claim, where it was previously accepted. It only affects routes that explicitly set `claims_to_verify`; the default behavior is unchanged. `jwt-auth-more-algo.t` TEST 13 is updated to reflect this (a token missing the configured `nbf` claim is now rejected). Happy to discuss on the dev mailing list if preferred.

**3. `key-auth`: strip credentials on anonymous-consumer fallback**

When a request carried an invalid API key and key-auth fell back to the configured `anonymous_consumer`, the (invalid) credential was still forwarded upstream even with `hide_credentials` enabled: `find_consumer()` only strips credentials on the successful-auth path. The fix strips the credential before falling back to the anonymous consumer when `hide_credentials` is true. Since a request can carry the key in both the header and the query string, both are cleaned up.

Fixes #

### Checklist

- [ ] I have verified that this change is backward compatible (item 2 is an intentional behavior change for explicitly-configured `claims_to_verify`; called out above)

### DIFF
--- a/apisix/plugins/key-auth.lua
+++ b/apisix/plugins/key-auth.lua
@@ -108,6 +108,21 @@ function _M.rewrite(conf, ctx)
             core.response.set_header("WWW-Authenticate", "apikey realm=\"" .. conf.realm .. "\"")
             return 401, { message = err}
         end
+        -- Strip credentials before falling back to the anonymous consumer when
+        -- hide_credentials is enabled. find_consumer() only strips on the
+        -- successful-auth path, so without this an invalid credential would be
+        -- forwarded upstream during anonymous fallback. A request may carry the
+        -- credential in both the header and the query string, so clean up both.
+        if conf.hide_credentials then
+            if core.request.header(ctx, conf.header) then
+                core.request.set_header(ctx, conf.header, nil)
+            end
+            local args = core.request.get_uri_args(ctx) or {}
+            if args[conf.query] then
+                args[conf.query] = nil
+                core.request.set_uri_args(ctx, args)
+            end
+        end
         consumer, consumer_conf, err = consumer_mod.get_anonymous_consumer(conf.anonymous_consumer)
         if not consumer then
             err = "key-auth failed to authenticate the request, code: 401. error: " .. err
--- a/apisix/plugins/jwt-auth/parser.lua
+++ b/apisix/plugins/jwt-auth/parser.lua
@@ -256,6 +256,9 @@ function _M.verify_claims(self, claims, conf)
     -- required: they must exist in the payload and be valid.
     for _, claim_name in ipairs(claims) do
         local claim = self.payload[claim_name]
+        if claim == nil then
+            return false, "claim " .. claim_name .. " is missing"
+        end
         local checker = claims_checker[claim_name]
         if type(claim) ~= checker.type then
             return false, "claim " .. claim_name .. " is not a " .. checker.type
--- a/apisix/plugins/jwt-auth/parser.lua
+++ b/apisix/plugins/jwt-auth/parser.lua
@@ -229,7 +229,11 @@ end
 
 
 function _M.verify_claims(self, claims, conf)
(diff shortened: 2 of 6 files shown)


membphis (dismissed):

AlinsRan (dismissed):

apisix/plugins/key-auth.lua
```diff
@@ -108,6 +108,21 @@ function _M.rewrite(conf, ctx)
...
+        if conf.hide_credentials then
+            if core.request.header(ctx, conf.header) then
+                core.request.set_header(ctx, conf.header, nil)
+            end
+            local args = core.request.get_uri_args(ctx) or {}
+            if args[conf.query] then
+                args[conf.query] = nil
+                core.request.set_uri_args(ctx, args)
+            end
+        end
```
nic-6443, line 125: This code is completely duplicated from that inside `find_consumer`. It can be moved to the end of this function to handle both cases in one place.

shreemaan-abhishek: Thanks @nic-6443. The two blocks look duplicated but they intentionally implement different behaviors, so collapsing them into one place would change documented behavior.

- `find_consumer` (auth succeeded) strips **only** the credential location that was actually used (`from_header ? header : query`) and leaves the other location untouched. This is asserted by `key-auth.t` TEST 19 ("when apikey both in header and query string, verify apikey request header is hidden but request args is NOT hidden") and TEST 23 (the mirror for the query case).
- The `rewrite` anonymous-fallback path strips **both** locations, because auth failed: there is no single "used" credential, and the request may carry a credential-looking value in both the header and the query, either of which would otherwise be forwarded upstream under the anonymous identity (the leak this PR closes).

Concretely, for `GET /echo?apikey=Q` with header `apikey: H` and `hide_credentials=true`:
- success (`H` valid): strip header, keep `?apikey=Q` (TEST 19);
- fallback (`H` invalid): strip both.

Same request, opposite handling of the query location. Moving the logic to the end of `rewrite` and stripping both in one place would drop `?apikey=Q` on the success path and break TEST 19; conversely, stripping only the used location everywhere would re-introduce the leak on the fallback path. So a single shared block cannot satisfy both without a deliberate behavior change. I am keeping the two paths separate for that reason.

membphis (approved):

nic-6443 (approved):

AlinsRan (approved):

PR #13472, opened 2026-06-04
feat(openid-connect): make client_secret optional for local JWT verification modes

### LINKED ISSUE #10563: change: remove the requirement of `client_secret` in `openid-connect` plugin
### Description

`client_secret` is currently a required attribute to use the `openid-connect` plugin. However, some use cases supported by this plugin seem to not require `client_secret`:

1. when using authorization code grant with PKCE (`brearer_only` to `false` and `use_pkce` to `true`), `client_secret` is not needed.
2. when using `private_key_jwt` as the `token_endpoint_auth_method`, `client_secret` is not needed and `client_rsa_private_key` is used instead.

This issue is a discussion thread that evaluates whether we should toggle `client_secret` requirement to false and add relevant test cases if needed. Thoughts welcomed.

### Environment

- APISIX version (run `apisix version`): 3.7.0

### LINKED ISSUE #13397: openid-connect: claim_schema is not enforced in the bearer-token authentication path
### Description

The `openid-connect` plugin enforces the `claim_schema` option only in the session/callback authentication path, not in the bearer-token path.

`claim_schema` is validated by `validate_claims_in_oidcauth_response()` in `apisix/plugins/openid-connect.lua`, and that function has a single call site, inside the `openidc.authenticate()` branch (session/browser flow). The bearer branch that handles `bearer_only` / `introspection_endpoint` / `public_key` / `use_jwks` never calls it.

As a result, when a route accepts both authentication modes and a `claim_schema` is configured:

- requests authenticated via the session/callback flow are checked against `claim_schema`
- requests authenticated via a bearer token are not checked against `claim_schema`

### Why it happens

`claim_schema` was added together with the OIDC claim validator (#11824) and was wired into the session-flow response, which is the structured object `{ user, access_token, id_token }` returned by `openidc.authenticate()`. The bearer/introspection/JWKS path returns a flat introspection or JWT payload with a different shape (no `id_token`, no `user` wrapper), so it was never routed through the same validation function.

This is plausibly intentional scoping rather than an oversight: the function is named for the `openidc.authenticate()` response, and `claim_schema` validates a `{ user, access_token, id_token }` object that only exists in the session flow. The bearer path also already has its own claim controls (`claim_validator`, `required_scopes`).

### Expected behavior

The behavior of `claim_schema` across the two authentication paths should be consistent and clearly documented. Either:

1. document explicitly that `claim_schema` applies only to the session/callback flow, or
2. apply an equivalent claim validation step to the bearer/introspection/JWKS path.

**A docs-only fix (option 1) is an acceptable resolution for this issue.** The current docs describe `claim_schema` as "JSON schema of OIDC response claim" with no mention of which authentication flow it applies to, which is what invites the wrong assumption. Clarifying that it covers the session/callback flow only is sufficient to close this.

If option 2 is chosen instead, note that the bearer-path response has a different structure than the session-flow response, so the validation object for bearer mode would need to be built accordingly (and the per-flow key differences documented).

### Steps to reproduce

1. Configure a route with the `openid-connect` plugin, `bearer_only: false`, an `introspection_endpoint`, and a `claim_schema`.
2. Access the route without credentials and complete the browser login flow: the response is checked against `claim_schema`.
3. Access the same route with an `Authorization: Bearer <token>` header: the response is not checked against `claim_schema`.

### Environment

- APISIX version: present since 3.14.0 (the release that first carried #11824); reproducible on current master.

AlinsRan:
## Background

Closes #10563
Closes #13397

In service-to-service scenarios where the gateway only validates an incoming Bearer token locally (via a configured public key or JWKS endpoint), `client_secret` plays no role — no call is made to the IdP's token or introspection endpoint. However, the plugin currently requires `client_secret` unconditionally, forcing users to supply a dummy value as a workaround.

### `apisix/plugins/openid-connect.lua`

- Remove `client_secret` from the schema's `required` array.
- Add conditional enforcement in `check_schema`: `client_secret` is still required for all flows that need it (session/callback flow, introspection), but is now optional when:
  - `bearer_only=true` + `public_key`: local JWT verification with a configured public key
  - `bearer_only=true` + `use_jwks=true`: local JWT verification via JWKS endpoint
  - `token_endpoint_auth_method=private_key_jwt`: RSA private key replaces `client_secret`
  - `use_pkce=true` (non-bearer): public-client PKCE flow

- Fix `claim_schema` not being enforced in the bearer-token path (#13397): the schema is now applied directly to the flat JWT payload / introspection response in the bearer branch.

### `t/plugin/openid-connect.t`

Add TEST 42–47 covering:
- `bearer_only=true` + `public_key` → no `client_secret` required
- `bearer_only=true` + `use_jwks=true` → no `client_secret` required
- `bearer_only=true` + introspection endpoint (no local key) → `client_secret` still required
- `token_endpoint_auth_method=private_key_jwt` → no `client_secret` required
- `use_pkce=true` → no `client_secret` required
- Session flow without special auth method → `client_secret` still required

## Backward Compatibility

All existing configurations remain valid. The change only relaxes the requirement for specific scenarios; any config that previously worked continues to work unchanged.

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -403,7 +403,15 @@ local schema = {
     },
     encrypt_fields = {"client_secret", "client_rsa_private_key",
                       "session.secret", "session.redis.password"},
-    required = {"client_id", "discovery"}
+    required = {"client_id", "discovery"},
+    anyOf = {
+        {required = {"client_secret"}},
+        {properties = {public_key = {type = "string", minLength = 1}}, required = {"public_key"}},
+        {properties = {use_jwks   = {["const"] = true}}},
+        {properties = {use_pkce   = {["const"] = true}}},
+        {properties = {introspection_endpoint_auth_method = {["const"] = "private_key_jwt"}}},
+        {properties = {token_endpoint_auth_method         = {["const"] = "private_key_jwt"}}},
+    },
 }
 
 
@@ -422,25 +430,6 @@ function _M.check_schema(conf)
 
     if not conf.bearer_only and not conf.session then
         return false, "property \"session.secret\" is required when \"bearer_only\" is false"
-    end
-
-    -- client_secret is not required in certain authentication modes:
-    --   bearer_only=true + public_key/use_jwks: local JWT verification, no IdP call needed
-    --   bearer_only=true + introspection_endpoint_auth_method=private_key_jwt: introspection
-    --     endpoint authenticates via signed JWT instead of client_secret
-    --   token_endpoint_auth_method=private_key_jwt (non-bearer): token endpoint uses signed
-    --     JWT; this exemption applies only to the session/callback flow, not bearer mode
-    --   use_pkce=true (non-bearer): public-client PKCE flow needs no client_secret
-    local client_secret_optional
-    if conf.bearer_only then
-        client_secret_optional = (conf.public_key or conf.use_jwks)
-            or (conf.introspection_endpoint_auth_method == "private_key_jwt")
-    else
-        client_secret_optional = (conf.token_endpoint_auth_method == "private_key_jwt")
-            or conf.use_pkce
-    end
-    if not client_secret_optional and not conf.client_secret then
-        return false, "property \"client_secret\" is required"
     end
 
     local check = {"discovery", "introspection_endpoint", "redirect_uri",
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -407,10 +407,12 @@ local schema = {
     anyOf = {
         {required = {"client_secret"}},
         {properties = {public_key = {type = "string", minLength = 1}}, required = {"public_key"}},
-        {properties = {use_jwks   = {["const"] = true}}},
-        {properties = {use_pkce   = {["const"] = true}}},
-        {properties = {introspection_endpoint_auth_method = {["const"] = "private_key_jwt"}}},
-        {properties = {token_endpoint_auth_method         = {["const"] = "private_key_jwt"}}},
+        {properties = {use_jwks   = {["const"] = true}}, required = {"use_jwks"}},
+        {properties = {use_pkce   = {["const"] = true}}, required = {"use_pkce"}},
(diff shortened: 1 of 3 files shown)


membphis:
[P2] The `client_secret` alternatives need to be scoped to the OIDC flow they actually replace.

The final schema-level `anyOf` accepts any of `public_key`, `use_jwks`, `use_pkce`, `introspection_endpoint_auth_method=private_key_jwt`, or `token_endpoint_auth_method=private_key_jwt` as a global replacement for `client_secret`. That is broader than the intended behavior: some alternatives only apply to bearer/local JWT verification, some only to bearer introspection, and some only to the non-bearer token endpoint flow.

For example, a bearer introspection config without `client_secret` can currently pass schema validation if it sets `token_endpoint_auth_method=private_key_jwt`, even though that field is not the auth method used for introspection. Similarly, a non-bearer session flow can be accepted via an alternative that is only meaningful for bearer validation.

Please either restore the mode-aware validation in `check_schema`, or express the condition in JSON Schema with `bearer_only`/flow-specific branches. It would also be good to add negative tests for cross-flow combinations, e.g. `bearer_only=true` + introspection + `token_endpoint_auth_method=private_key_jwt` without `client_secret`, and non-bearer session flow with an introspection-only alternative.

AlinsRan:
@membphis Thanks for catching this. You're right that the global `anyOf` was too broad — it accepted cross-flow combinations where the alternative does not apply to the actual flow.

I've reverted the schema-level `anyOf` and restored the mode-aware validation in `check_schema`, scoping each exemption to its flow:
- `bearer_only=true`: exempt only via `public_key`/`use_jwks` (local JWT verification) or `introspection_endpoint_auth_method=private_key_jwt` (introspection auth)
- non-bearer (session/callback): exempt only via `token_endpoint_auth_method=private_key_jwt` or `use_pkce`

Also added negative tests for the cross-flow cases you mentioned:
- `bearer_only=true` + introspection + `token_endpoint_auth_method=private_key_jwt` without `client_secret` → now rejected
- non-bearer session flow with a bearer-only alternative (introspection `private_key_jwt`) without `client_secret` → now rejected

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

PR #13478, opened 2026-06-08
feat(hmac-auth): add max_req_body_size to bound client request body during validation

shreemaan-abhishek:
### Description

The `hmac-auth` plugin reads the entire client request body into memory to verify the digest when `validate_request_body` is `true`. Today that read (`core.request.get_body()`) has no upper bound, so a client can force a worker to buffer an arbitrarily large body. The only existing backstop is the global nginx `client_max_body_size`, which operators routinely raise or disable for upload routes.

This adds a `max_req_body_size` option (integer, default `67108864` = 64 MiB, minimum `1`) to `hmac-auth`, so that when body validation is enabled, oversized requests are rejected with `413` before the body is buffered:

- `hmac-auth`: caps the body read for digest validation (passed to `core.request.get_body()`); enforced only on the `validate_request_body = true` path.

This complements the body-size hardening series started in #13466, which added `max_req_body_size` to `forward-auth`, `ai-proxy`, and `ai-proxy-multi`. That PR deliberately excluded `hmac-auth`; this PR closes that gap with the same default, the same `413` rejection behavior, and the same schema/docs conventions.

Docs and e2e tests are included.

### ⚠️ Default behavior change

`hmac-auth` now defaults `max_req_body_size` to **64 MiB**. The change is observable only when **all** of the following hold:

1. the route uses `hmac-auth` with `validate_request_body = true`; **and**
2. the operator has raised nginx `client_max_body_size` above 64 MiB (or set it to `0`); **and**
3. a client sends a request body larger than 64 MiB.

In that case the request is now rejected with `413` instead of being buffered in full. Under the default `client_max_body_size` (1 MiB) there is **no** change. The limit is configurable via `max_req_body_size` to restore prior behavior.

Fixes #

### Checklist

- [x] I have verified that this change is backward compatible (see the documented default behavior change above; configurable via `max_req_body_size`)


nic-6443 (approved):

membphis:
One test needs to be updated before merge. This PR adds `max_req_body_size` to the `hmac-auth` route/service schema, but `t/plugin/hmac-auth2.t` still asserts the exact old response body for `/apisix/admin/schema/plugins/hmac-auth`. Once the admin schema includes the new property, that test should fail deterministically.

Please update the expected schema JSON, or narrow the assertion to the relevant schema fields, so it includes the new `max_req_body_size` property.

shreemaan-abhishek:
`hmac-auth2.t` TEST 2 isn't an exact-match assertion. `test_admin.test` uses `com_tab`, which only checks the expected pattern is a *subset* of the response, ignoring extra keys. The expected JSON already omits `signed_headers.default`, `realm`, and `anonymous_consumer` and still passes. `max_req_body_size` is just another un-asserted extra key, so the test stays green — confirmed by the passing `t/plugin/[a-k]*.t` CI job.

membphis (approved):

AlinsRan (approved):

PR #13483, opened 2026-06-08
fix(xrpc): bound redis command-line preallocation size

shreemaan-abhishek:
### Description

The redis xRPC protocol parser reads the RESP array header (`*<n>\r\n`) from the
downstream client and passes the declared argument count straight into
`core.tablepool.fetch("xrpc_redis_cmd_line", narg, 0)`. `table.new(n, 0)` eagerly
allocates `n` array slots, so a large declared count forces a large upfront
allocation before any argument is actually read, regardless of how much data the
client sends.

This PR bounds and validates the declared argument count:

- The declared count is validated to be a positive integer. Previously a negative
  or non-integer value (e.g. `*-1\r\n`) flowed into `table.new(-1, 0)` and raised
  an uncaught Lua error; it is now rejected cleanly with
  `invalid argument number: <n>`.
- The declared count is now used only as a bounded preallocation hint
  (clamped to `MAX_CMD_LINE_PREALLOC = 64`). The command-line table still grows
  naturally as the real arguments are read, so legitimate large commands are
  unaffected; only the eager preallocation is capped.

This is a behaviour-hardening change to the redis xRPC command parser.

Fixes #

### Checklist

### DIFF
--- a/apisix/stream/xrpc/protocols/redis/init.lua
+++ b/apisix/stream/xrpc/protocols/redis/init.lua
@@ -173,6 +173,10 @@ local function read_req(session, sk)
         return nil, err
     end
 
+    if n < 0 then
+        return nil, str_fmt("invalid bulk length: %s", n)
+    end
+
     local p, err = sk:read(n + 2)
     if not p then
         return nil, err
@@ -204,6 +208,10 @@ local function read_req(session, sk)
         local n, err = read_len(sk)
         if not n then
             return nil, err
+        end
+
+        if n < 0 then
+            return nil, str_fmt("invalid bulk length: %s", n)
         end
 
         local s
--- a/apisix/stream/xrpc/protocols/redis/init.lua
+++ b/apisix/stream/xrpc/protocols/redis/init.lua
@@ -28,6 +28,7 @@ local DONE = ngx.DONE
 local sleep = ngx.sleep
 local str_byte = string.byte
 local str_fmt = string.format
+local str_match = string.match
 local ipairs = ipairs
 local tonumber = tonumber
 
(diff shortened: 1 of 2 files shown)


membphis:
I think this still leaves one malformed RESP path unsafe.

The new validation protects the array argument count (`*<n>`), but `read_len()` is also used for bulk-string lengths. Since the new pattern still accepts negative decimal tokens (`^-?%d+$`), a request such as:

```text
*1\r\n$-1\r\n
```

can return `n = -1` for the command bulk length. `read_req()` then reaches `sk:read(n + 2)` and later `ffi_str(p, n)` without rejecting `n < 0`. Locally, LuaJIT reports `ffi.string(ptr, -1)` as `string length overflow`, so a malformed client request can still hit an uncaught runtime-error path.

Suggested fix: reject negative bulk-string lengths before every bulk `sk:read(n + 2)` in `read_req()` (both command and argument lengths), and add regression tests for negative command bulk length and negative argument bulk length.

shreemaan-abhishek:
Good catch, thanks. You're right that `read_len()` is shared and the bulk-length path still accepted negative tokens. Pushed e332f8eaf: `read_req()` now rejects `n < 0` right before each bulk `sk:read(n + 2)` — both the command bulk length and the per-argument bulk length — so a negative length can no longer reach `ffi_str(p, n)`. `read_len()` stays as the lexical check (a decimal integer token) while `read_req()` does the semantic range checks (`narg >= 1`, bulk `n >= 0`). Added regression tests TEST 16 (negative command bulk length, `*1\r\n$-1\r\n`) and TEST 17 (negative argument bulk length, `*2\r\n$3\r\nGET\r\n$-1\r\n`).

membphis (approved):

nic-6443 (approved):

nic-6443 (approved):

AlinsRan (approved):

PR #13487, opened 2026-06-08
feat(prometheus): add built-in LLM histograms for TTFT and token distribution

AlinsRan:
### Description

This PR enriches the AI Gateway Prometheus metrics.

**1. Time to first token (TTFT), via a `type` label on `apisix_llm_latency`.**

`apisix_llm_latency` previously observed `llm_time_to_first_token` for every AI request. For non-streaming that value is the total response time, but for streaming it is only the TTFT, so the metric mixed two different quantities. This PR makes the meaning explicit with a `type` label, mirroring `apisix_http_latency{type="request"|"upstream"|"apisix"}`:

- `type="total"` — full response latency (`ai_chat` and `ai_stream`). For streaming this now uses `apisix_upstream_response_time` (time until the whole response completes) instead of the TTFT.
- `type="ttft"` — time to first token (`ai_stream` only; non-streaming responses have no first-token moment).

This removes the previous overlap and lets you track streaming TTFT separately from total latency. Both kinds share `llm_latency_buckets`.

**2. Token distribution histograms.**

- `apisix_llm_prompt_tokens_dist` / `apisix_llm_completion_tokens_dist` — histograms of prompt/completion tokens per request, complementing the existing `apisix_llm_prompt_tokens` / `apisix_llm_completion_tokens` counters with a distribution so quantiles (e.g. p95 prompt size) can be computed. Buckets configurable via `plugin_attr.prometheus` (`llm_prompt_tokens_buckets` / `llm_completion_tokens_buckets`, unit: token); defaults are tuned to real-world token ranges up to 1M.

**Compatibility note:** `apisix_llm_latency` gains a `type` label and, for streaming requests, now records two samples (`total` + `ttft`). Existing queries should add a `type="total"` selector to keep their previous meaning.

Fixes #

### Checklist

- [x] I have verified that this change is backward compatible (If not, please discuss on the APISIX mailing list first)


apisix/plugins/prometheus/exporter.lua
```diff
@@ -366,16 +416,36 @@ function _M.http_log(conf, ctx)
...
+
+            -- Only streaming requests expose a real TTFT; for non-streaming the
+            -- var holds the total response time, which would pollute the TTFT
+            -- distribution, so record llm_ttft for ai_stream only.
+            if vars.request_type == "ai_stream" then
+                metrics.llm_ttft:observe(tonumber(llm_time_to_first_token),
+                    gen_arr(route_id, service_id, consumer_name, balancer_ip,
+                        vars.request_type, vars.request_llm_model, vars.llm_model,
+                        unpack(extra_labels("llm_ttft", ctx))))
+            end
```
nic-6443: > `apisix_llm_ttft` — LLM time to first token (milliseconds), observed for streaming (ai_stream) requests only. The existing apisix_llm_latency mixes streaming TTFT and non-streaming total latency in one series; this dedicated metric keeps the TTFT distribution semantically consistent so it can be used for streaming latency SLOs.

Then we should also adjust the existing `apisix_llm_latency`. In stream request, set it to the time when the entire response is completed. Otherwise, `apisix_llm_ttft` and `apisix_llm_latency` will have overlapping functions.

membphis (approved):

apisix/plugins/prometheus/exporter.lua
AlinsRan: `apisix_llm_latency`:
- `type="total"` — full response latency (`ai_chat` and `ai_stream`). For streaming this now uses `apisix_upstream_response_time` (time until the whole response completes) instead of the TTFT.
- `type="ttft"` — time to first token (`ai_stream` only; non-streaming responses have no first-token moment).

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13489, opened 2026-06-09
feat(ai-plugins): add fail_mode for Consumer-bound protocol handling

AlinsRan:
### Description

When a content-security AI plugin is bound at the **Consumer** or **Service** level, it can receive requests it cannot inspect — plain HTTP traffic, non-JSON bodies, or requests that never passed through `ai-proxy`. Today the three plugins behave inconsistently in that situation:

| Plugin | Behavior on an unrecognized / non-AI request |
|--------|----------------------------------------------|
| `ai-aliyun-content-moderation` | returns `500 no ai instance picked` |
| `ai-prompt-guard` | silently allows the request (a security gap) |
| `ai-aws-content-moderation` | moderates the raw body |

This PR adds a shared helper `apisix/plugins/ai-protocols/binding.lua` exposing a `fail_mode` attribute (`skip` \| `warn` \| `error`, default `skip`) and a unified handler, and wires it into the three plugins:

- **`skip`** (default): the request passes through unchecked, logged at info level.
- **`warn`**: same as `skip`, plus a warning log.
- **`error`**: the request is rejected (fail-closed) — for bindings where every request must be an AI request.

With the default `skip`, a Consumer-bound AI plugin no longer breaks plain-HTTP / `multipart/form-data` traffic. The log reason is sanitized (control characters stripped, length capped) to avoid log forging, and Content-Type matching is case-insensitive.

#### Compatibility note

`fail_mode` is a new optional attribute, but its default (`skip`) intentionally changes the behavior on non-AI requests so the three plugins become consistent:

- `ai-aliyun-content-moderation`: previously `500` → now passes through (fixes the reported breakage).
- `ai-prompt-guard`: previously allowed → unchanged.
- `ai-aws-content-moderation`: previously moderated the raw body → now passes through.

Operators who need the old fail-closed behavior can set `"fail_mode": "error"`.

Fixes #

### Checklist

- [x] I have verified that this change is backward compatible (new optional attribute; default-behavior change is documented above)

### DIFF
--- a/apisix/plugins/ai-aliyun-content-moderation.lua
+++ b/apisix/plugins/ai-aliyun-content-moderation.lua
@@ -28,6 +28,7 @@ local core      = require("apisix.core")
 local http      = require("resty.http")
 local uuid      = require("resty.jit-uuid")
 local protocols = require("apisix.plugins.ai-protocols")
+local binding   = require("apisix.plugins.ai-protocols.binding")
 local sse       = require("apisix.plugins.ai-transport.sse")
 
 local schema = {
@@ -57,6 +58,7 @@ local schema = {
         region_id = {type ="string", minLength = 1},
         access_key_id = {type = "string", minLength = 1},
         access_key_secret = {type ="string", minLength = 1},
+        fail_mode = binding.schema_property("skip"),
         check_request = {type = "boolean", default = true},
         check_response = {type = "boolean", default = false},
         request_check_service = {type = "string", minLength = 1, default = "llm_query_moderation"},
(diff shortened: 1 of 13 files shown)


nic-6443 (dismissed):

membphis (commented):
The new `fail_mode=skip` path does not seem to cover non-JSON/plain HTTP bodies in `ai-prompt-guard`.

`ai-prompt-guard.access` still returns `400` immediately when `core.json.decode(body)` fails, before it reaches `protocols.detect()` and `binding.on_unsupported()`. That means a Consumer-bound `ai-prompt-guard` will still reject plain form/multipart/non-JSON traffic by default, even though the PR describes the default behavior as passing non-AI/plain HTTP requests through unchecked.

Impact: users who bind `ai-prompt-guard` at Consumer/Service level can still break ordinary non-JSON traffic with the default configuration.

Suggested fix: route the JSON decode error branch through `binding.on_unsupported(conf.fail_mode, ...)` as well. Keep the current 400 behavior for `fail_mode="error"`, but let `skip`/`warn` return without rejecting. It would also be good to add a non-JSON passthrough test for `ai-prompt-guard`, similar to the Aliyun/AWS coverage.

AlinsRan:
@membphis good catch, you're right. `ai-prompt-guard.access` returned `400` on the JSON decode error *before* `fail_mode` was consulted, so a Consumer/Service-bound prompt guard still rejected plain form / multipart / non-JSON traffic under the default `skip` — exactly the case this PR is meant to fix.

Fixed in 9e2dcf6: the decode-error branch now goes through `binding.on_unsupported(conf.fail_mode, ...)` as well — `error` keeps the 400, while `skip`/`warn` let the request pass through. Also added a non-JSON (`multipart/form-data`) passthrough test for `ai-prompt-guard` mirroring the Aliyun/AWS coverage. PTAL.

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

PR #13502, opened 2026-06-10
fix: remove credential-leaking debug logs in logger plugins

AlinsRan:
### Description

Several logger plugins print the full serialized log payload at `info`/`debug` level. The payload is the serialized log entries, which by default include request/response headers and bodies, so it can leak credentials (e.g. `Authorization`/`Cookie` headers) into the error log:

- `elasticsearch-logger.lua` — logged `uri` and the full bulk `body`.
- `syslog/init.lua` — logged the full RFC5424 payload and every buffered entry.
- `sls-logger.lua` — logged the full RFC5424 payload before sending it (debug level).
- `kafka-logger.lua` / `rocketmq-logger.lua` — logged the full serialized batch payload right before sending it.

This PR removes those statements.

For `elasticsearch-logger`, `syslog` and `sls-logger`, no test relies on the removed lines. For `kafka-logger` and `rocketmq-logger`, the existing tests used the removed log line to observe what was queued for delivery, so the same observability is reproduced with a **test-only** hook that logs each batch entry from `batch-processor-manager` (the entry is the exact object that gets sent, so content-sensitive assertions — including the `no_error_log` body-filter cases — keep working). Production code no longer logs the payload.

### Checklist

### DIFF
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -294,8 +294,6 @@ local function send_to_elasticsearch(conf, entries)
         end
     end
 
-    core.log.info("uri: ", uri, ", body: ", body)
-
     httpc:set_timeout(conf.timeout * 1000)
     local resp, err = httpc:request_uri(uri, {
         ssl_verify = conf.ssl_verify,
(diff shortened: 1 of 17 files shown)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Approved.

PR #13504, opened 2026-06-10
fix: remove credential-leaking debug log in kafka-logger and rocketmq-logger

AlinsRan:
### Description

`kafka-logger` and `rocketmq-logger` log the full serialized batch payload right before sending it:

```lua
core.log.info("send data to kafka: ", data)     -- kafka-logger
core.log.info("send data to rocketmq: ", data)  -- rocketmq-logger
```

`data` is the serialized log entries, which by default include request/response headers and bodies, so it can leak credentials (e.g. `Authorization`/`Cookie` headers) into the error log when the log level is set to `info`.

This PR removes those statements. The existing tests used this log line to observe what was queued for delivery, so the same observability is reproduced with a **test-only** hook that logs each batch entry from `batch-processor-manager` (the entry is the exact object that gets sent, so content-sensitive assertions — including the `no_error_log` body-filter cases — keep working). Production code no longer logs the payload.

### Checklist

### DIFF
--- a/apisix/plugins/kafka-logger.lua
+++ b/apisix/plugins/kafka-logger.lua
@@ -292,8 +292,6 @@ function _M.log(conf, ctx)
             return false, 'error occurred while encoding the data: ' .. err
         end
 
-        core.log.info("send data to kafka: ", data)
-
         return send_kafka_data(conf, data, prod)
     end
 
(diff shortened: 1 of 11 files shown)


AlinsRan:
Folded into #13502, which now covers all four logger plugins in one change.

PR #13509, opened 2026-06-10
feat: add Kubernetes Secret Manager for $secret://kubernetes/ references

### LINKED ISSUE #13493: bug: authz-keycloak encrypt_fields causes service load failure when used with Kubernetes Secret Manager ($secret://kubernetes)
## Environment
- APISIX version: 3.16.0 (debian image)
- Deployment: Kubernetes via apisix-ingress-controller 2.1.0 + ADC 0.26.0
- Gateway API mode (HTTPRoute + PluginConfig v1alpha1)
## Description
Two related bugs in the `authz-keycloak` plugin when used with the Kubernetes Secret Manager (`$secret://kubernetes/...`):
### Bug 1: `encrypt_fields` corrupts services on startup
When `client_secret` is stored in etcd as a `$secret://kubernetes/...` reference and the Kubernetes Secret Manager resolves it at runtime, APISIX encrypts the resolved value (because `client_secret` is in `encrypt_fields`) **before storing it back to etcd** via the Admin API. The resulting encrypted value (AES-256-CBC base64, ~128-152 chars) exceeds the `maxLength: 100` defined in the plugin's JSON schema for `client_secret`.
On next startup, `config_etcd.lua:load_full_data()` validates the schema against the already-encrypted value and **fails**, causing every service that uses `authz-keycloak` to be skipped entirely:
lua config_etcd.lua:612: load_full_data(): failed to check item data of
/apisix/services err: failed to check the configuration of plugin
authz-keycloak err: property "client_secret" validation failed: string too
long, expected at most 100, got 128
This results in 404 for all routes using the affected services.
**Root cause:** Schema validation runs against the post-encryption value instead of the pre-encryption plaintext. The `maxLength: 100` constraint is designed for the user-supplied credential, not the AES-encrypted output.
### Bug 2: `authz-keycloak` plugin schema `maxLength: 100` is too restrictive
Even without encryption, real-world OAuth2 `client_secret` values frequently exceed 100 characters (e.g. Keycloak-generated secrets of 32+ random bytes encoded in various formats, or Azure AD client secrets). The limit of 100 characters for both `client_id` and `client_secret` is unnecessarily restrictive.
Additionally, when using `$secret://kubernetes/<namespace>/<secret-name>/<key>` references, the reference string itself can exceed 100 characters (e.g. `$secret://kubernetes/my-long-namespace/my-secret-name/client_secret` = 68+ chars), causing the Admin API to reject the configuration before APISIX even tries to resolve it.

### Expected Behavior

- Schema validation should run against the **pre-encryption** value, not the encrypted output
- `maxLength` for `client_id` and `client_secret` should accommodate encrypted values and `$secret://` reference strings (suggest 4096 or removing the limit entirely)

### Error Logs

_No response_

### Steps to Reproduce

1. Configure `authz-keycloak` plugin with `$secret://kubernetes/<ns>/<secret>/<key>` references for `client_id` and `client_secret`
2. Ensure the Kubernetes secret contains a `client_secret` value (any non-trivial OAuth2 secret)
3. APISIX resolves the reference at runtime, encrypts the value, and stores it in etcd
4. Restart APISIX
5. Observe `load_full_data()` error for all services using `authz-keycloak`

### Environment

The $secret://kubernetes Secret Manager implementation (apisix/secret/kubernetes.lua) appears to be absent from the official 3.16.0-debian Docker image, which means $secret://kubernetes/... references are silently treated as literal strings at the plugin level, while the Admin API validation (which has the 100-char limit) still applies. This makes the feature completely unusable in Kubernetes environments without custom image builds.
Related files:
- apisix/plugins/authz-keycloak.lua — lines 34-35 (maxLength), line 75 (encrypt_fields)
- apisix/secret/ — missing kubernetes.lua in official debian image

hiades-devops:
Implements the Kubernetes secret manager (apisix/secret/kubernetes.lua) which allows APISIX to read Kubernetes Secrets directly from the cluster using the pod's ServiceAccount token.

URI format:
  $secret://kubernetes/{manager-id}/{namespace}/{secret-name}/{data-key}

Example:
  $secret://kubernetes/my-k8s/default/api-creds/client_secret

The implementation follows the same pattern as the existing Vault, AWS, and GCP secret managers: a schema for Admin API configuration and a get(conf, key) function called by apisix.secret.fetch_secrets() at request time.

Key features:
- Uses pod ServiceAccount token (default in-cluster path) for auth
- Reads KUBERNETES_SERVICE_HOST/PORT from env if not explicitly configured
- Verifies TLS using the in-cluster CA bundle by default
- Decodes base64 Secret.data values automatically
- Clear error messages for auth failures, missing secrets, missing keys

Also fixes two bugs in authz-keycloak.lua discovered while using the Kubernetes secret manager in production (fixes #13493):

1. client_id and client_secret maxLength was 100 characters. When APISIX resolves a $secret:// reference, encrypts the value (encrypt_fields), and stores it back to etcd, the AES-encrypted result can be 128-152 chars — exceeding the schema limit and causing load_full_data() to fail on restart, dropping all authz-keycloak services with a 404. Increased maxLength to 4096 for both fields.

2. The same maxLength = 100 constraint also blocked $secret://kubernetes/ references from being stored at all via the Admin API, since the reference string itself (e.g. $secret://kubernetes/my-k8s/my-ns/ my-secret/client_secret) can exceed 100 characters.

Closes #13493

Fixes #

### Checklist

### DIFF
--- a/apisix/secret/kubernetes.lua
+++ b/apisix/secret/kubernetes.lua
@@ -1,0 +1,262 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+--- Kubernetes Secret Manager.
+--  Fetches secrets from the Kubernetes API server using the pod's ServiceAccount
+--  token. This allows APISIX to read Kubernetes Secrets directly from the cluster
+--  it is running in, without requiring an external secrets management service.
+--
+--  URI format:
+--    $secret://kubernetes/{manager-id}/{namespace}/{secret-name}/{data-key}
+--
+--  Example:
+--    $secret://kubernetes/my-k8s/default/my-secret/password
+--
+--  The manager is configured once via the Admin API:
+--    PUT /apisix/admin/secrets/kubernetes/my-k8s
+--    {
+--      "service_account_file": "/var/run/secrets/kubernetes.io/serviceaccount/token",
+--      "kubernetes_host": "kubernetes.default.svc",
+--      "kubernetes_port": "443",
+--      "ssl_verify": true
+--    }
+--
+--  All fields have sensible defaults for in-cluster usage, so the minimal
+--  valid configuration is an empty object `{}`.
+--
+--  TLS note: when ssl_verify is true (the default), the TLS handshake is
+--  validated against the nginx-level lua_ssl_trusted_certificate bundle.
+--  For in-cluster usage, set apisix.ssl.ssl_trusted_certificate in config.yaml
+--  to /var/run/secrets/kubernetes.io/serviceaccount/ca.crt so that the
+--  kube-apiserver certificate (signed by the cluster CA) is trusted.
+
+local core = require("apisix.core")
+local http = require("resty.http")
+local env  = core.env
+
+local io_open = io.open
+local find    = core.string.find
+local sub     = core.string.sub
+local ngx_decode_base64 = ngx.decode_base64
+
+local DEFAULT_SA_FILE = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+local schema = {
+    type = "object",
+    properties = {
+        service_account_file = {
+            type = "string",
+            description = "Path to the ServiceAccount token file. "
+                       .. "Defaults to the standard in-cluster path.",
+            default = DEFAULT_SA_FILE,
+        },
+        kubernetes_host = {
+            type = "string",
+            description = "Kubernetes API server hostname or IP. "
+                       .. "Defaults to the KUBERNETES_SERVICE_HOST environment variable.",
+        },
+        kubernetes_port = {
+            type = "string",
+            description = "Kubernetes API server port. "
+                       .. "Defaults to the KUBERNETES_SERVICE_PORT environment variable.",
+        },
+        endpoint = {
+            type = "string",
+            description = "Full base URL of the Kubernetes API server "
+                       .. "(e.g. https://kubernetes.default.svc:443). "
+                       .. "When set, kubernetes_host and kubernetes_port are ignored. "
+                       .. "Useful for testing with plain HTTP mock servers.",
+        },
+        ssl_verify = {
+            type = "boolean",
+            description = "Verify the Kubernetes API server TLS certificate. "
+                       .. "When true, validation uses the nginx-level "
+                       .. "lua_ssl_trusted_certificate bundle; set "
+                       .. "apisix.ssl.ssl_trusted_certificate to the cluster CA "
+                       .. "(/var/run/secrets/kubernetes.io/serviceaccount/ca.crt) "
+                       .. "in config.yaml for in-cluster usage.",
+            default = true,
+        },
+    },
+    required = {},
+}
+
+local _M = {
+    schema = schema,
+}
+
+
+local function read_file(path)
+    local f, err = io_open(path, "r")
+    if not f then
+        return nil, "failed to open file " .. path .. ": " .. err
+    end
+    local content = f:read("*a")
+    f:close()
+    if not content or content == "" then
+        return nil, "file is empty: " .. path
+    end
+    return content
+end
+
+
+local function make_request_to_k8s(conf, namespace, secret_name)
+    local sa_file = conf.service_account_file or DEFAULT_SA_FILE
+    local token, err = read_file(sa_file)
+    if not token then
+        return nil, err
+    end
+    -- strip trailing newline
+    token = token:gsub("%s+$", "")
+
+    local k8s_host = conf.kubernetes_host
+                  or env.fetch_by_uri("$ENV://KUBERNETES_SERVICE_HOST")
+                  or os.getenv("KUBERNETES_SERVICE_HOST")
+
+    local k8s_port = conf.kubernetes_port
+                  or env.fetch_by_uri("$ENV://KUBERNETES_SERVICE_PORT")
+                  or os.getenv("KUBERNETES_SERVICE_PORT")
+                  or "443"
+
+    local base_url
+    if conf.endpoint then
+        base_url = conf.endpoint
+    else
+        if not k8s_host then
+            return nil, "kubernetes_host is not set and KUBERNETES_SERVICE_HOST env var is missing"
+        end
+        base_url = "https://" .. k8s_host .. ":" .. k8s_port
+    end
+
+    local uri = base_url
+             .. "/api/v1/namespaces/" .. namespace
+             .. "/secrets/" .. secret_name
+
+    core.log.info("fetching Kubernetes secret from: ", uri)
+
+    local httpc = http.new()
+    httpc:set_timeout(5000)
+
+    local ssl_verify = conf.ssl_verify
+    if ssl_verify == nil then
+        ssl_verify = true
+    end
+
+    local request_opts = {
+        method = "GET",
+        headers = {
+            ["Authorization"] = "Bearer " .. token,
+            ["Accept"] = "application/json",
+        },
+        ssl_verify = ssl_verify,
+    }
+
+    local res, req_err = httpc:request_uri(uri, request_opts)
+    if not res then
+        return nil, "failed to request Kubernetes API: " .. req_err
+    end
+
+    if res.status == 401 or res.status == 403 then
+        return nil, "unauthorized to read Kubernetes secret "
+               .. namespace .. "/" .. secret_name
+               .. " (HTTP " .. res.status .. "): check RBAC permissions for the ServiceAccount"
+    end
+
+    if res.status == 404 then
+        return nil, "Kubernetes secret not found: " .. namespace .. "/" .. secret_name
+    end
+
+    if res.status ~= 200 then
+        return nil, "unexpected HTTP status " .. res.status
+               .. " from Kubernetes API for secret "
+               .. namespace .. "/" .. secret_name
+    end
+
+    return res.body
+end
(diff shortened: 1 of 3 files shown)


apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,242 @@
...
+        method = "GET",
+        headers = {
+            ["Authorization"] = "Bearer " .. token,
+            ["Accept"] = "application/json",
+        },
+        ssl_verify = ssl_verify,
+    }
+
+    if ssl_verify then
+        request_opts.ssl_trusted_certificate = DEFAULT_CA_FILE
```
nic-6443: `ssl_trusted_certificate` is not a supported `request_uri` option in lua-resty-http (the api7 fork APISIX uses only honors `ssl_verify`, `ssl_server_name` and `ssl_send_status_req` in `http_connect.lua`), so this line is silently ignored. With `ssl_verify = true`, the handshake verifies against the nginx-level `lua_ssl_trusted_certificate`, which defaults to the system CA bundle (`apisix.ssl.ssl_trusted_certificate: system`) — and the kube-apiserver cert is signed by the cluster CA, not a public one. So the default config will fail the TLS handshake in a real cluster, and the obvious workaround users will reach for is `ssl_verify: false`, which means sending the ServiceAccount token over an unverified connection.

I think this needs to either drop the no-op option and document that `apisix.ssl.ssl_trusted_certificate` must be set to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (the doc currently claims the in-cluster CA is used by default, which is not accurate), or find another way to actually load the cluster CA.

apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,242 @@
...
+    if not k8s_host then
+        return nil, "kubernetes_host is not set and KUBERNETES_SERVICE_HOST env var is missing"
+    end
+
+    local k8s_port = conf.kubernetes_port
+                  or env.fetch_by_uri("$ENV://KUBERNETES_SERVICE_PORT")
+                  or os.getenv("KUBERNETES_SERVICE_PORT")
+                  or "443"
+
+    local uri = "https://" .. k8s_host .. ":" .. k8s_port
```
nic-6443: Hardcoding `https://` makes the bundled tests impossible to pass: the mock locations in t/secret/kubernetes.t serve plain HTTP on port 1984, so TESTs 10-14 will fail at the TLS handshake before ever reaching the mock. gcp.lua handles this by letting the full endpoint be configured (`entries_uri`), which is exactly what its tests use to mock with `http://127.0.0.1:1984`. I would suggest the same here — e.g. an optional scheme/endpoint override defaulting to https — rather than rewriting the tests around a TLS mock.

Separately, TEST 14 will still fail after that: the manager registered in TEST 13 does not set `service_account_file`, so `get()` falls back to `/var/run/secrets/kubernetes.io/serviceaccount/token`, which does not exist in CI. Worth running the suite locally before the next push.

apisix/plugins/authz-keycloak.lua
```diff
@@ -30,8 +30,8 @@ local schema = {
         discovery = {type = "string", minLength = 1, maxLength = 4096},
         token_endpoint = {type = "string", minLength = 1, maxLength = 4096},
         resource_registration_endpoint = {type = "string", minLength = 1, maxLength = 4096},
-        client_id = {type = "string", minLength = 1, maxLength = 100},
-        client_secret = {type = "string", minLength = 1, maxLength = 100},
+        client_id = {type = "string", minLength = 1, maxLength = 4096},
+        client_secret = {type = "string", minLength = 1, maxLength = 4096},
```
nic-6443: The rationale for this bump does not hold on master. Since #13312, `core.schema.check` skips validation for `$secret://` / `$env://` references on string fields (`skip_validation = secret.is_secret_ref` in apisix/core/schema.lua), so a long reference string already passes the 100-char limit. And on reload, `plugin_checker` runs `decrypt_conf` before the schema check, so validation sees the original value, not the AES output — what #13493 describes is real on 3.16.0 but is already addressed here.

Raising the limit might still be worth doing for genuinely long client secrets, but then it should be justified on those grounds; otherwise I would drop this hunk from the PR and keep it focused on the new secret manager.

LuuOW (commented):
Technical audit: code patterns and implementation verified for alignment with modern software engineering standards.

apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,262 @@
...
+            description = "Verify the Kubernetes API server TLS certificate. "
+                       .. "When true, validation uses the nginx-level "
+                       .. "lua_ssl_trusted_certificate bundle; set "
+                       .. "apisix.ssl.ssl_trusted_certificate to the cluster CA "
+                       .. "(/var/run/secrets/kubernetes.io/serviceaccount/ca.crt) "
+                       .. "in config.yaml for in-cluster usage.",
+            default = true,
+        },
+    },
+    required = {},
```
AlinsRan, line 95: This one line makes the manager unregisterable.

An empty Lua table serializes to an empty JSON *object*, not an array, so jsonschema reads it as "input must be an empty table" and rejects anything with properties:

```
required={} + {}                 -> OK
required={} + {ssl_verify=true}  -> the input data should be an empty table
no required + {ssl_verify=true}  -> OK
```

`secret.lua:52` and `admin/secrets.lua:35` both run `core.schema.check(secret_manager.schema, conf)`, so every non-empty PUT body is rejected. That's TEST 13's `error_msg`, and TEST 14 then cascades to `no secret conf`.

Drop the line. vault/aws/gcp omit `required` when nothing is mandatory.

apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,262 @@
...
+    local token, err = read_file(sa_file)
+    if not token then
+        return nil, err
+    end
+    -- strip trailing newline
+    token = token:gsub("%s+$", "")
+
+    local k8s_host = conf.kubernetes_host
+                  or env.fetch_by_uri("$ENV://KUBERNETES_SERVICE_HOST")
+                  or os.getenv("KUBERNETES_SERVICE_HOST")
```
AlinsRan, line 128: This is the lint failure (`line 128: getting the Lua global "os"`, same on 132).

Use `core.env.get("KUBERNETES_SERVICE_HOST")`. `discovery/kubernetes/core.lua:74` already switched for this reason — `os.getenv` hits a lua-resty-core prefix-collision bug during worker init (#13055).

It's redundant too: `env.fetch_by_uri("$ENV://...")` above already falls back to `os.getenv` internally. `core.env.get` alone covers both.

apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,262 @@
...
+local function make_request_to_k8s(conf, namespace, secret_name)
+    local sa_file = conf.service_account_file or DEFAULT_SA_FILE
+    local token, err = read_file(sa_file)
+    if not token then
+        return nil, err
+    end
+    -- strip trailing newline
+    token = token:gsub("%s+$", "")
+
+    local k8s_host = conf.kubernetes_host
```
AlinsRan, line 126: The in-cluster default won't work. Workers only see env vars declared via an `env` directive, and `cli/ngx_tpl.lua` doesn't emit `KUBERNETES_SERVICE_HOST`/`_PORT`. So with the documented minimal config `{}` this is nil and you get `kubernetes_host is not set and KUBERNETES_SERVICE_HOST env var is missing`.

Either require users to declare both under `nginx_config.envs`, or drop the env fallback and require `kubernetes_host`. The doc table currently presents `$KUBERNETES_SERVICE_HOST` as a working default.

No test covers this branch — every test passes an explicit `endpoint`.

apisix/secret/kubernetes.lua
```diff
@@ -0,0 +1,262 @@
...
+
+    local idx1 = find(key, "/")
+    if not idx1 then
+        return nil, "invalid key format, expected {namespace}/{secret-name}/{data-key}, got: "
+               .. key
+    end
+
+    local namespace = sub(key, 1, idx1 - 1)
+    if namespace == "" then
+        return nil, "namespace is empty in key: " .. key
```
AlinsRan, line 205: `namespace` and `secret_name` land in the request path unescaped. resty.http's `_format_request` inserts it verbatim and `parse_uri` splits on `([^\?]*)`, so `?`, `#` and CRLF pass through to the apiserver. `$secret://kubernetes/mgr/default?labelSelector=x/mysecret/key` emits `GET /api/v1/namespaces/default?labelSelector=x/secrets/mysecret`.

Not a vulnerability — Admin API operators are trusted and already have `script`. But a DNS-1123 check on both is cheap and turns a malformed ref into a clear error instead of a silently reshaped request.

t/secret/kubernetes.t
```diff
@@ -0,0 +1,463 @@
...
+GET /t
+--- response_body_like
+failed to request Kubernetes API:.*
+--- timeout: 6
+
+
+
+=== TEST 9: get - mock Kubernetes API returns 401
+--- config
+    location /mock-k8s-401 {
```
AlinsRan, line 233: Fails in CI, and can't test what the name says. The mock sits at `/mock-k8s-401` but the code requests `/api/v1/namespaces/default/secrets/my-secret`, so it hits the catch-all, gets 404, and returns `Kubernetes secret not found: default/my-secret` — not in the assertion's alternation.

Put the mock at `location ~ "^/api/v1/namespaces/default/secrets/my-secret$"` returning 401 and assert the exact message. The three-way alternation would pass on nearly any failure anyway.

Also: the `BEGIN { $ENV{KUBERNETES_SERVICE_HOST} = ... }` block is dead code, nothing exercises the env fallback. Untested branches: non-200/401/403/404 status, malformed JSON, secret with no `data`, and the `ngx.decode_base64` nil path.

AlinsRan (changes requested):
@hiades-devops CI is red, and both failures are real bugs.

`lint`: `kubernetes.lua:128` and `:132` — `getting the Lua global "os"`.

`build (t/secret ...)`: `kubernetes.t` 43 tests, 7 failed (TEST 1, 9, 13, 14). TEST 13 gets `invalid configuration: the input data should be an empty table` — meaning you can't register a kubernetes secret manager with any config at all. The feature doesn't work end to end right now. Root cause inline on line 95.

Two things outside the diff:

The PR body claims an authz-keycloak maxLength 100→4096 fix and `Closes #13493`, but that change isn't here — `authz-keycloak.lua` still has `maxLength = 100`. It's also not needed: `$secret://` refs skip schema validation. Please drop that part and the `Closes`, otherwise it'll close an unrelated issue.

`docs/zh/latest/terminology/secret.md` needs the same section. The GCP manager commit (2fcfbd83e) did en and zh together.

The module itself reads well — TLS verify defaults on and fails closed, and errors never leak the token or the secret value.

PR #13518, opened 2026-06-11
fix(jwt-auth): reject malformed JWT signature instead of erroring

shreemaan-abhishek:
### Description

`jwt-auth`'s custom JWT parser (`apisix/plugins/jwt-auth/parser.lua`) replaced `resty.jwt` verification with per-algorithm verifiers. `verify_signature` looked up the verifier and passed the decoded signature without guarding either path:

```lua
function _M.verify_signature(self, key)
    return alg_verify[self.header.alg](self.raw_header .. "." ..
               self.raw_payload, base64_decode(self.signature), key)
end
```

The per-algorithm verifiers `assert` on signature length and key validity, and `base64_decode` returns `nil` for an invalid base64url signature. So a token with a **wrong-length signature** (`assert(#signature == 64, "Signature must be 64 bytes.")`) or a **non-base64url signature** (`#nil` -> "attempt to get length of a nil value") makes the verifier raise a Lua error. Since the caller invokes `verify_signature` without a guard, the error propagates and the request returns **500 instead of 401** for any route protected by `jwt-auth`. An unauthenticated client that knows a valid consumer key can repeatedly trigger these internal errors and error-log noise.

This makes `verify_signature` defensive: guard the algorithm lookup and the base64url decode, and run the verifier under `pcall` so a malformed token is cleanly rejected as an invalid signature instead of crashing the request.

N/A (reported via a private security scan)

### Checklist

- [ ] I have updated the documentation to reflect this change (no user-facing behavior or config change; internal robustness fix only)

### DIFF
--- a/apisix/plugins/jwt-auth/parser.lua
+++ b/apisix/plugins/jwt-auth/parser.lua
@@ -223,8 +223,26 @@ end
 
 
 function _M.verify_signature(self, key)
-    return alg_verify[self.header.alg](self.raw_header .. "." ..
-               self.raw_payload, base64_decode(self.signature), key)
+    local verifier = alg_verify[self.header.alg]
+    if not verifier then
+        return false, "unsupported algorithm: " .. tostring(self.header.alg)
+    end
+
+    local signature = base64_decode(self.signature)
+    if not signature then
+        return false, "failed to decode signature"
+    end
+
+    -- the per-algorithm verifiers assert on signature length and key validity,
+    -- so guard with pcall to turn a malformed token into a clean rejection
+    -- instead of letting the error propagate as a 500 response
+    local ok, verified = pcall(verifier, self.raw_header .. "." .. self.raw_payload,
+                               signature, key)
+    if not ok then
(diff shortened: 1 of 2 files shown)


shreemaan-abhishek:
Addressed the lint failure and review feedback in the follow-up commit:

- **Lint (`tostring` global):** localized `tostring` alongside the other standard-library locals so the lj-releng global check passes.
- **Verifier return contract:** the `pcall` wrapper now captures and forwards the verifier's own `(verified, err)` values instead of dropping the secondary error detail.
- **Test order dependency:** merged the two malformed-signature cases into a single self-contained test that sets up its own consumer/route and exercises both the wrong-length and non-base64url signatures.

Verified locally: `luacheck` + lj-releng clean, and `t/plugin/jwt-auth.t` passes (and the merged case still fails as expected when the fix is reverted).

membphis (approved):

AlinsRan (approved):

nic-6443 (approved):

PR #13524, opened 2026-06-11
fix(openid-connect): token introspection with client_secret_basic sends credentials in both header and body

### LINKED ISSUE #13085: bug: openid-connect token introspection client_secret_basic, secret sent in header and body
### Current Behavior

Hello, 
I'm experiencing an issue with the openid-connect plugin. I'm using client credentials with token introspection, but it doesn't seem to be working as expected.
When I enable Nginx logs, I notice that client_secret and client_id are being sent in the request body along with the token, even though the plugin is configured to use client_secret_basic (the basic auth is send as well). 

I have debugged the conf object passed to the openidc.introspect(conf) method, and I couldn't identify any problematic values. Do you have any idea what might be causing this ?

My Authorization Server rejects the request if the client credentials are sent simultaneously in the Authorization header and the request body.

Is it possible that the issue lies within the OpenResty layer (resty.openidc)? It seems the introspect function is duplicating the credentials in the POST body even when the authentication method is explicitly set to Basic

Thanks.

### Expected Behavior

When use "introspection_endpoint_auth_method": "client_secret_basic", only basic authentification is use ( without client info in body )

### Error Logs

nginx logs : 

[debug]  openidc.lua:456: call_token_endpoint(): client_secret_basic: authorization header 'Basic xxxxxx'
[debug]  openidc.lua:511: call_token_endpoint(): request **body** for introspection endpoint call: **client_secret=xxxxxx&token=xxxxxxx&client_id=my.client.id**
[debug]  openidc.lua:430: openidc_configure_proxy(): openidc_configure_proxy : don't use http proxy
[info]  client.lua:123: dns_parse(): dns resolve myauthserver, result: {"address":"xx.x.xxx.xxx","type":1,"class":1,"name":"myauthserver","ttl":3600,"section":1}, client: xx.x.xxx.xxx, server: _, request: "GET /apisix-openid-connect/v1/response/ok HTTP/1.1", host: "myapisix.com"
[info]  resolver.lua:84: parse_domain(): parse addr: {"address":"xx.x.xxx.xxx","type":1,"class":1,"name":"myauthserver","section":1,"ttl":3600}, client: xx.x.xxx.xxx, server: _, request: "GET /apisix-openid-connect/v1/response/ok HTTP/1.1", host: "myapisix.com"
[info]  resolver.lua:85: parse_domain(): resolver: ["xx.x.xxx.xxx","xx.x.xxx.xxx","xx.x.xxx.xxx"], client: xx.x.xxx.xxx, server: _, request: "GET /apisix-openid-connect/v1/response/ok HTTP/1.1", host: "myapisix.com"
[info]  resolver.lua:86: parse_domain(): host: myauthserver, client: xx.x.xxx.xxx, server: _, request: "GET /apisix-openid-connect/v1/response/ok HTTP/1.1", host: "myapisix.com"
[info]  resolver.lua:88: parse_domain(): dns resolver domain: myauthserver to xx.x.xxx.xxx, client: xx.x.xxx.xxx, server: _, request: "GET /apisix-openid-connect/v1/response/ok HTTP/1.1", host: "myapisix.com"
[debug]  openidc.lua:529: call_token_endpoint(): introspection endpoint response: {"error":"invalid_client"}

conf object parameter in call "openidc.introspect(conf)"
{
  "unauth_action": "auth",
  "timeout": 3000,
  "client_jwt_assertion_expires_in": 60,
  "discovery": "https://myauthserver/.well-known/openid-configuration",
  "renew_access_token_on_expiry": true,
  "bearer_only": true,
  "client_secret": "xxxxxxxx",
  "use_pkce": false,
  "introspection_interval": 0,
  "introspection_endpoint_auth_method": "client_secret_basic",
  "iat_slack": 120,
  "client_id": "my.client.id",
  "access_token_in_authorization_header": false,
  "set_access_token_header": true,
  "realm": "apisix",
  "ssl_verify": "no",
  "access_token_expires_leeway": 0,
  "set_id_token_header": true,
  "force_reauthorize": false,
  "token_endpoint_auth_method": "client_secret_basic",
  "use_nonce": false,
  "set_refresh_token_header": false,
  "revoke_tokens_on_logout": false,
  "jwt_verification_cache_ignore": false,
  "jwk_expires_in": 86400,
  "accept_none_alg": false,
  "logout_path": "/logout",
  "accept_unsupported_alg": true,
  "redirect_uri": "/apisix-openid-connect/v1/response/ok/.apisix/redirect",
  "set_userinfo_header": true,
  "scope": "openid profile email",
  "_meta": {}
}

### Steps to Reproduce

Here is my configuration:

curl http://localhost:9180/apisix/admin/routes/00000000000000021323 \
-H "X-API-KEY:$ADC_TOKEN" -X PUT -i -d '
{
  "uri": "/apisix-openid-connect/*",
  "name": "apisix-openid-connect",
  "plugins": {
    "openid-connect": {
      "bearer_only": true,
      "client_id": "my.client.id",
      "client_secret": "xxxxxxxxx",
      "discovery": "https://myauthserver/.well-known/openid-configuration",
      "scope": "openid profile email",
      "introspection_endpoint_auth_method": "client_secret_basic"
    }
  },
  "upstream": {
    "name": "server-api-test",
    "nodes": [
      {
        "host": "server-api-test",
        "port": 3003,
        "weight": 100
      }
    ],
    "timeout": {
      "connect": 60,
      "send": 60,
      "read": 60
    }
  }
}'

### Environment

- APISIX version (run `apisix version`): 3.14
- Operating system (run `uname -a`):  Debian 5.10.249-1
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):  3.5.24

nic-6443:
### Description

When `introspection_endpoint_auth_method` is `client_secret_basic` (the default), the token introspection request sends the client credentials twice: once in the `Authorization: Basic` header and once more as `client_id`/`client_secret` in the POST body. RFC 6749 Section 2.3.1 says the client must not use more than one authentication mechanism per request, so strict authorization servers reject the introspection call with `invalid_client` and every bearer-token request through the route fails with 401 — with the default plugin config.

The root cause is in lua-resty-openidc 1.8.0: `introspect()` unconditionally copies `client_id`/`client_secret` into the request body, and `call_token_endpoint()` then adds the Basic header for `client_secret_basic` without removing the body copies. There is an upstream issue (zmartzone/lua-resty-openidc#556), but it has been open for a while with no release containing a fix, so this patches it on the APISIX side without forking the library.

The fix extends the existing `http_request_decorator` hook usage in the plugin's introspection path: when the auth method is `client_secret_basic`, the decorator strips `client_id`/`client_secret` from the urlencoded request body before the request is sent. Scoping notes:

- The decorator is set right before `openidc.introspect()` and cleared right after it returns (existing behavior), so it only affects the introspection call, not the token-endpoint calls of the authorization code flow.
- Within `openidc.introspect()`, the decorator can also see a body-less discovery GET (when `introspection_endpoint` is not set explicitly), hence the `req.body` guard.
- `client_secret_post`, `private_key_jwt` and `client_secret_jwt` are untouched: stripping only happens for `client_secret_basic`, where the credentials are already carried by the Basic header.

The new test file `t/plugin/openid-connect11.t` uses a mock introspection endpoint that behaves like a strict authorization server (rejects requests carrying credentials in both the header and the body) and covers `client_secret_basic` (default and explicit, with and without `introspection_addon_headers`) as well as `client_secret_post`, which must keep sending body credentials.

Fixes #13085

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -625,17 +625,35 @@ local function introspect(ctx, conf)
     else
         -- Validate token against introspection endpoint.
         -- TODO: Same as above for public key validation.
-        if conf.introspection_addon_headers then
+        -- lua-resty-openidc puts the client credentials into the introspection
+        -- request body even when they are already sent in the Authorization
+        -- header with client_secret_basic. Sending both violates
+        -- RFC 6749 Section 2.3.1 and strict authorization servers reject such
+        -- requests, so strip the duplicated credentials from the body.
+        local strip_body_credentials =
+            conf.introspection_endpoint_auth_method == "client_secret_basic"
+        if conf.introspection_addon_headers or strip_body_credentials then
             -- http_request_decorator option provided by lua-resty-openidc
             conf.http_request_decorator = function(req)
-                local h = req.headers or {}
-                for _, name in ipairs(conf.introspection_addon_headers) do
-                    local value = core.request.header(ctx, name)
-                    if value then
-                        h[name] = value
+                if conf.introspection_addon_headers then
+                    local h = req.headers or {}
+                    for _, name in ipairs(conf.introspection_addon_headers) do
+                        local value = core.request.header(ctx, name)
+                        if value then
+                            h[name] = value
+                        end
                     end
+                    req.headers = h
                 end
-                req.headers = h
+                -- the body is already urlencoded here; the decorator is also
+                -- applied to body-less requests like discovery, so only
+                -- process requests carrying a body
+                if strip_body_credentials and req.body then
+                    local args = ngx.decode_args(req.body)
+                    args.client_id = nil
+                    args.client_secret = nil
+                    req.body = ngx.encode_args(args)
+                end
                 return req
             end
         end
--- a/t/plugin/openid-connect11.t
+++ b/t/plugin/openid-connect11.t
@@ -1,0 +1,318 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+use t::APISIX 'no_plan';
+
+log_level('debug');
+repeat_each(1);
+no_long_string();
+no_root_location();
+
+add_block_preprocessor(sub {
+    my ($block) = @_;
+
+    if ((!defined $block->error_log) && (!defined $block->no_error_log)) {
+        $block->set_value("no_error_log", "[error]");
+    }
+
+    if (!defined $block->request) {
+        $block->set_value("request", "GET /t");
+    }
+});
+
+# The introspection mock used below acts as a strict authorization server
+# which follows RFC 6749 Section 2.3.1: the client MUST NOT use more than
+# one authentication mechanism in each request. It rejects introspection
+# requests that carry client credentials in both the Authorization header
+# and the request body.
+run_tests();
+
+__DATA__
+
+=== TEST 1: Set up route with default introspection_endpoint_auth_method (client_secret_basic)
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local code, body = t('/apisix/admin/routes/1',
+                 ngx.HTTP_PUT,
+                 [[{
+                        "plugins": {
+                            "openid-connect": {
+                                "client_id": "course_management",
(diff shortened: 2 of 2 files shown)


kevinlzw:
Hi @nic-6443 , 

I have tried to fix the issue of upstream lua-resty-openidc in this [PR](https://github.com/zmartzone/lua-resty-openidc/pull/561). Can you check does it resolve your problem? Thank you.

nic-6443:
@kevinlzw Thanks! I verified your fix against APISIX: I ran this PR's test suite (a strict mock introspection endpoint that rejects requests carrying credentials in both the Authorization header and the body) on current master with lua-resty-openidc bumped to 1.9.0 and **without** the plugin-side workaround from this PR — all tests pass. With 1.8.0 the same tests fail, so the upstream fix alone resolves the problem.

One detail worth noting: your fix keeps the legacy body-credentials behavior when `introspection_endpoint_auth_method` is unset, but that doesn't affect APISIX — the plugin schema declares `default = "client_secret_basic"`, so the option is always materialized into the opts passed to `openidc.introspect()` and the new code path is taken.

Since v1.9.0 is already released and published to LuaRocks, the right fix for APISIX is now simply bumping the dependency in the rockspec instead of patching around it, so I'm closing this PR. Would you like to submit the version-bump PR? Since you authored the upstream fix, it seems fitting — and feel free to pick up `t/plugin/openid-connect11.t` from this PR as regression coverage if useful. Otherwise happy to do it myself.

kevinlzw:
@nic-6443 Thanks for pointing this out. I have included this upgrade together with the related changes in apache/apisix#13649, so it should be covered there as part of the same update. Please feel free to take a look at that PR when you have a chance.

PR #13528, opened 2026-06-11
fix(ai-aws-content-moderation): moderate decoded LLM content instead of raw body

shreemaan-abhishek:
### Description

The `ai-aws-content-moderation` plugin sent the **raw** HTTP request body to AWS Comprehend instead of parsing the LLM request and extracting the actual prompt content. As a result, moderation scored the undecoded JSON envelope, while the upstream LLM acts on the decoded content, so the two see different text:

- A body containing `"content":"toxic"` was scored by Comprehend as the literal escaped string, whereas the upstream model decodes it to `toxic`.
- The full `{"model":...,"messages":[...]}` envelope was scored as-is, adding noise to the toxicity result.

This is inconsistent with the sibling `ai-aliyun-content-moderation` plugin, which is protocol-aware.

#### What changed

Make the plugin protocol-aware, in the same style as `ai-aliyun-content-moderation`:

- Require/validate the `application/json` content type.
- Parse the JSON body, detect the client protocol via `ai-protocols`, and extract the LLM-visible content (`messages[].content`, Responses input/instructions, Anthropic message content, etc.).
- Send only the normalized, decoded content to Comprehend.

Because the plugin runs in the `rewrite` phase (before `ai-proxy`), it detects the protocol directly rather than relying on `ctx.ai_client_protocol`. Requests that are not recognized AI requests (non-JSON content type, unparseable bodies, or JSON that carries no LLM content) are handled by the existing `fail_mode` (`skip` by default), so non-AI traffic on Consumer-bound plugins keeps its current pass-through behavior.

#### Behavior change

Making the plugin JSON/LLM-only is a behavior change for anyone running it on non-JSON routes. With the default `fail_mode: skip`, such requests now pass through unchecked instead of being forwarded verbatim to Comprehend; set `fail_mode: error` to reject them.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/ai-aws-content-moderation.lua
+++ b/apisix/plugins/ai-aws-content-moderation.lua
@@ -18,6 +18,7 @@ require("resty.aws.config") -- to read env vars before initing aws module
 
 local core = require("apisix.core")
 local binding = require("apisix.plugins.ai-protocols.binding")
+local protocols = require("apisix.plugins.ai-protocols")
 local aws = require("resty.aws")
 local aws_instance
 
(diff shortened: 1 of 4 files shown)


shreemaan-abhishek:
Superseded by #13647, which refactors the plugin to run in the access phase like ai-aliyun-content-moderation and extracts the decoded LLM content via the detected protocol (fixing the decode issue this PR addressed as part of the restructure).

PR #13174, opened 2026-04-07
docs: add Gemini, Vertex AI, and Anthropic examples to ai-proxy plugin

Yilialinn:
## Summary

- Add missing provider examples to the `ai-proxy` plugin documentation for providers that were added in previous PRs (e.g. #12933) but lacked usage examples.
- New example sections added (both EN and ZH):
  - **Proxy to Gemini** — chat completion using Gemini API with API key auth
  - **Proxy to Vertex AI Chat Completion** — using GCP service account authentication
  - **Proxy to Vertex AI Embedding Models** — embedding models with GCP auth
  - **Proxy to Anthropic** — Claude chat completion with `x-api-key` header auth
  - **Convert Anthropic Requests to OpenAI-Compatible Backend** — protocol conversion via `/v1/messages` URI

PR #13539, opened 2026-06-12
docs(ai-proxy): restore max_stream_duration_ms and max_response_bytes

AlinsRan:
### What this fixes

The `max_stream_duration_ms` and `max_response_bytes` safeguards are implemented and present in the ai-proxy plugin schema, but they are **undocumented** — the attribute table in `docs/{en,zh}/latest/plugins/ai-proxy.md` does not list them.

### Source

- Added (code + docs) in **#13250** ("feat(ai-proxy): add max_stream_duration_ms and max_response_bytes safeguards").
- The doc rows were then removed in **#13174** ("docs: add Gemini, Vertex AI, and Anthropic examples to ai-proxy plugin"), which also trimmed the `timeout` description — likely an unintended drop from a stale branch. The schema fields remain (`apisix/plugins/ai-proxy/schema.lua` lines 254 / 263 and 379 / 388), so the guardrails exist but are invisible to users.

### Change (docs only)

- Restore the `max_stream_duration_ms` and `max_response_bytes` rows in both `docs/en` and `docs/zh`.
- Restore the per-socket clarification on the `timeout` description ("Applied per socket operation … does not cap the total duration of a streaming response").

### Checklist

- [x] I have added tests corresponding to this change (N/A, docs only)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Looks good to me.

PR #13541, opened 2026-06-12
fix(nacos): declare nacos-stream shared dict in the stream subsystem

AlinsRan:
### TL;DR

Enabling nacos discovery while the **stream** subsystem is on crashes the stream worker at startup: `lua_shared_dict "nacos-stream" not configured`.

### Cause

nacos selects its shared dict by subsystem (`nacos` for http, `nacos-stream` for stream), but `nacos-stream` is declared **nowhere**. `init_worker()` errors out before any network call:

```
init_worker_by_lua error: .../nacos/init.lua:284: lua_shared_dict "nacos-stream" not configured
```

Introduced by the refactor in #13201.

### Fix

Declare `nacos-stream` in the `stream` block of `ngx_tpl.lua`, mirroring the unconditional `nacos` dict in the http block.

### Test

`t/cli/test_discovery_nacos_stream.sh`: starts APISIX with `proxy_mode: stream` + nacos discovery and asserts the stream worker initializes without the error (no nacos server needed — the dict check runs before any network call). Verified locally: passes with the fix, fails (exit 1, with the exact error above) without it.

The bug was missed because it lives in the production `ngx_tpl.lua` template + the stream `init_worker`, which the Test::Nginx `.t` suite bypasses (it uses its own nginx config); only the `t/cli` tests exercise the real generated config.

### Checklist

### DIFF
--- a/apisix/cli/ngx_tpl.lua
+++ b/apisix/cli/ngx_tpl.lua
@@ -149,6 +149,7 @@ stream {
     lua_shared_dict lrucache-lock-stream {* stream.lua_shared_dict["lrucache-lock-stream"] *};
     lua_shared_dict etcd-cluster-health-check-stream {* stream.lua_shared_dict["etcd-cluster-health-check-stream"] *};
     lua_shared_dict worker-events-stream {* stream.lua_shared_dict["worker-events-stream"] *};
+    lua_shared_dict nacos-stream 10m;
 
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Looks good to me.

PR #13542, opened 2026-06-12
fix(elasticsearch-logger): guard os.date against invalid index template

AlinsRan:
### TL;DR

Dynamic Elasticsearch index names containing `{*t}` / `{!*t}` produced a garbage index like `services-table: 0x7f...` instead of being rejected.

### Cause

The `index` field supports `{time_format}` placeholders passed straight to `os.date`. The existing `if not time` guard assumes bad input returns `nil`, but on LuaJIT/OpenResty it never does:

- `{*t}` / `{!*t}` make `os.date` return a **table**, not a string. It then flows into `ngx.re.gsub`, which stringifies it into the index name.

### Fix

Require `os.date` to return a string before using it; otherwise log `failed to parse time format` and fall back to an empty replacement. Wrapped in `pcall` for defense-in-depth on non-LuaJIT runtimes (where `os.date` *can* raise on a bad escape).

```lua
local ok, time = pcall(os_date, time_format)
if not ok or type(time) ~= "string" then
    core.log.error("failed to parse time format: ", time_format)
    return ""
end
```

### Test

`t/plugin/elasticsearch-logger2.t` TEST 10: `prefix{*t}suffix` (and `{!*t}`) now resolves to `prefixsuffix` with a `failed to parse time format` log, instead of embedding a `table: 0x...` address.

### Checklist

### DIFF
--- a/apisix/plugins/elasticsearch-logger.lua
+++ b/apisix/plugins/elasticsearch-logger.lua
@@ -24,6 +24,8 @@ local ngx_re          = ngx.re
 local str_format      = core.string.format
 local math_random     = math.random
 local os_date         = os.date
+local pcall           = pcall
+local type            = type
 local pairs           = pairs
 
 local plugin_name = "elasticsearch-logger"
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Looks good to me.

PR #13543, opened 2026-06-12
docs: fix mocking config note and invalid JSON in elasticsearch-logger examples

AlinsRan:
### What this fixes

Two documentation issues found during a docs audit:

**1. mocking — inaccurate constraint description**

`docs/{en,zh}/latest/plugins/mocking.md` state that `response_example` and `response_schema` "must not be configured together". That is wrong: the schema uses `anyOf` (at least one required), so configuring both is valid, and the plugin prefers `response_example` when both are set:

```lua
-- apisix/plugins/mocking.lua
anyOf = {
    { required = { "response_example" } },
    { required = { "response_schema" } },
}
...
if conf.response_example then          -- response_example wins when both set
    response_content = conf.response_example
else
    local output = gen_object(conf.response_schema)
```

Corrected to: at least one must be configured; if both are configured, `response_example` takes precedence and `response_schema` is ignored.

**2. elasticsearch-logger — invalid JSON in examples**

The Admin API examples had a trailing comma after the only key (`"index": "gateway",`), which makes the JSON invalid on copy-paste. Removed it (en + zh, two places each).

### Checklist

- [x] I have added tests corresponding to this change (N/A, docs only)


nic-6443 (dismissed):

shreemaan-abhishek (dismissed):

membphis (commented):
I think one documentation inconsistency still needs to be fixed before this lands.

The updated descriptions now correctly say that `response_example` and `response_schema` may both be configured, with `response_example` taking precedence. However, the attribute table still uses mutually exclusive wording in the Required column:

- `docs/en/latest/plugins/mocking.md`: `One of this or response_schema` / `One of this or response_example`
- `docs/zh/latest/plugins/mocking.md`: `二选一`

That still conflicts with the actual schema and runtime behavior: `apisix/plugins/mocking.lua` uses `anyOf` to require at least one of the two fields, and the plugin prefers `response_example` when both are present.

Could you also update the Required column to say that at least one of `response_example` or `response_schema` is required, rather than implying they are mutually exclusive?

nic-6443 (approved):

membphis (approved):

kayx23 (approved):

PR #13545, opened 2026-06-12
fix(brotli): don't log expected non-standard ETag at error level

AlinsRan:
### TL;DR

`brotli.lua` logged an **expected** non-quoted upstream ETag at `error` level (printing `... : nil`), spamming the error log.

### Cause

`weak_etag_header()` downgrades a strong `ETag` to a weak one (Brotli alters the body, so a strong validator no longer holds). When the upstream `ETag` is not a quoted string the regex simply does not match — an expected case — but the code logs it at `error` level with a nil `err`:

```lua
local matched, err = ngx.re.match(etag, regex, "jo")
if not matched or err then
    core.response.set_header("Etag", nil)
    core.log.error("no standard etag or regex match failed: ", err)  -- err is nil here
    return
end
```

### Fix

- Only log at `error` level when `ngx.re.match` actually returns an error; otherwise drop the ETag silently (behavior unchanged).
- Tighten the capture group from `(.*)` to `([^"]*)` so a malformed ETag containing internal quotes is treated as non-standard (and dropped) rather than weakened into another malformed value.

### Checklist

### DIFF
--- a/apisix/plugins/brotli.lua
+++ b/apisix/plugins/brotli.lua
@@ -158,12 +158,15 @@ local function weak_etag_header()
         return
     end
 
-    local regex = [[^(W/)?"(.*)"$]]
+    local regex = [[^(W/)?"([^"]*)"$]]
     local matched, err = ngx.re.match(etag, regex, "jo")
-    if not matched or err then
-        -- not standard etag, no quote
+    if err then
+        core.log.error("failed to match etag: ", err)
+    end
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Looks good to me.

PR #13546, opened 2026-06-12
fix(ai-proxy): forward client method and query string for passthrough

AlinsRan:
### TL;DR

The ai-proxy **passthrough** protocol dropped the client's query string and forced the method to POST, breaking providers that need query params on POST — e.g. Azure OpenAI's `?api-version=...`.

### Cause

In `apisix/plugins/ai-providers/base.lua`, the upstream request builder hardcodes `method = "POST"` and merges only `auth.query` + the `override.endpoint` URL query — the client's own query string (`ctx.var.args`) is never added.

### Fix

For the passthrough protocol (`ctx.ai_target_protocol == "passthrough"`), forward the client's HTTP method (`core.request.get_method()`) and merge the client's query string (`ctx.var.args`) into the upstream request. Other protocols are unchanged (still POST with provider-specific query args).

### Test

`t/plugin/ai-proxy-passthrough.t`: a passthrough request with `?name=foo` asserts the arg reaches the upstream; a `PUT` passthrough request asserts the upstream sees `PUT`.

> Note: I could not run the ai-proxy test harness locally — please rely on CI to validate the new tests.

### Checklist

### DIFF
--- a/apisix/plugins/ai-providers/base.lua
+++ b/apisix/plugins/ai-providers/base.lua
@@ -201,8 +201,20 @@ function _M.build_request(self, ctx, conf, request_body, opts)
         ctx.ai_converter.convert_headers(headers)
     end
 
+    -- For the passthrough protocol the gateway acts as a catch-all proxy, so
+    -- forward the client's HTTP method and original query string unchanged.
+    -- Other protocols always issue a POST with provider-specific query args.
+    local method = "POST"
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
Looks good to me.

PR #13547, opened 2026-06-12
fix: use description instead of title for field-level schema doc strings

nic-6443:
A few plugin schema fields use `title` where they actually mean `description`: `validate_request_body` in hmac-auth and the sampler fields in opentelemetry. In JSON Schema `title` is a short display name, while these strings are explanatory doc text — every other field-level doc string in the codebase (including `max_req_body_size` right next to `validate_request_body`) already uses `description`.

This keeps schema consumers (docs tooling, dashboards reading `/v1/schema`) from having to look in two places for the same thing, and avoids the wrong pattern spreading through copy-paste when new fields are added.

No behavior change. Updated the schema assertion in `t/plugin/hmac-auth2.t` accordingly.


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
Reviewed for merge readiness. No blocking issues found.

PR #13550, opened 2026-06-12
fix(mcp-bridge): only accept messages for an established session

shreemaan-abhishek:
### Description

The `mcp-bridge` message endpoint (`POST <base_uri>/message?sessionId=<id>`) trusted the supplied `sessionId` unconditionally: it constructed a server for whatever value was passed and pushed the body into that session's shared-dict queue, returning `202`. This happened even when no SSE session was ever established for that id (including a missing `sessionId`), leaving orphaned entries in the `mcp-session` shared dict that nothing drains. This also removes the long-standing `TODO: check ctx.var.arg_sessionId` in the access path.

This PR makes the message endpoint only accept messages for a session that is actually established:

- When an SSE session is established, it is recorded with a liveness marker in the `mcp-session` shared dict. The marker is refreshed by the existing keepalive loop and self-expires (TTL) if the worker holding the session goes away without running teardown; it is cleared on normal teardown.
- The message endpoint looks up that marker and returns `404` when the `sessionId` is missing or does not correspond to a live session, instead of creating a new queue.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/mcp/broker/shared_dict.lua
+++ b/apisix/plugins/mcp/broker/shared_dict.lua
@@ -30,6 +30,12 @@ local mt = { __index = _M }
 
 
 local STORAGE_SUFFIX_QUEUE = ":queue"
+local STORAGE_SUFFIX_ALIVE = ":alive"
+
+-- liveness marker TTL in seconds. it is refreshed on each keepalive so it
+-- comfortably outlives the 30s ping interval, and self-expires if the worker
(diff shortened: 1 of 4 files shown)


shreemaan-abhishek:
Superseded by #13551, which takes the same approach and additionally bounds the per-session queue (length cap + TTL) and cleans it up on teardown. Closing this one in favor of #13551.

PR #13553, opened 2026-06-12
fix(limit-count): redis-sentinel keepalive pools must not be shared across database/credentials

### LINKED ISSUE #13454: bug: module apisix.utils.redis - Redis connection reuse issue
### Current Behavior

When we use multiple rate-limiting plugins that share the same Redis address but connect to different databases or use different credentials (username/password), connection reuse can lead to getting a connection with the wrong database or user/password.

For example, assuming I configure two such limit-count plugins on a routes, 

```json
"limit-count": {
    "count": 1,
    "time_window": 30,
    "rejected_code": 429,
    "key": "remote_addr",
    "policy": "redis",
    "redis_host": "redis1.local",
    "redis_port": 6379,
    "redis_database": 1
}

"limit-count": {
    "count": 1,
    "time_window": 30,
    "rejected_code": 429,
    "key": "remote_addr",
    "policy": "redis",
    "redis_host": "redis1.local",
    "redis_port": 6379,
    "redis_database": 2
}
```

I will randomly get connections from the pool that use the wrong database. The core issue here is that authentication and database selection are executed only during the initial connect, after which the connection is placed back into the pool.

To fix this problem, we could re-authenticate and re-select the database whenever we detect that the connection's current credentials or database don't match the required ones. Alternatively – which in my opinion is a better approach – we can introduce separate connection pools for such connections. We can achieve this by implementing a function that generates a pool name based on the username, password, and database number,
```lua
local function get_pool_name(conf)
    local database = conf.redis_database or 0
    local connection_string = conf.redis_host .. ":" .. conf.redis_port .. "/" .. database
    if conf.redis_username and conf.redis_username ~= "" then
        local password = conf.redis_password or ""
        local hashed_credentials = ngx.md5(conf.redis_username .. ":" .. password)
        return hashed_credentials .. "@" .. connection_string
    else
        return connection_string
    end
end
```
 and then use it in sock_opts.

```lua
local sock_opts = {
        ...
        pool = get_pool_name(conf)
    }
```

### Expected Behavior

Redis commands must be executed using a connection configured with the appropriate user, password, and database ID, in accordance with the plugin's configuration.

### Error Logs

_No response_

### Steps to Reproduce

No description needed.

### Environment

It's irrelevant.

nic-6443:
### Description

Fixes #13454

Sentinel-mode counterpart of #13516. The `redis-sentinel` policy goes through api7-lua-resty-redis-connector, which skips AUTH/SELECT on connections reused from the keepalive pool while pooling by the default `host:port` key — so two routes pointing at the same sentinel-managed master with different `redis_database` or credentials cross-contaminate exactly like the plain-redis case: limit counters land in the wrong database.

This bumps the connector to 0.13.0 (api7/lua-resty-redis-connector#9), which derives the default pool name from everything the connection gets bound to at handshake time — resolved address, database, a credentials digest, and TLS options — while an explicit `connection_options.pool` still wins. Building the name from the resolved address keeps failover correct: a new master gets a new pool instead of reusing connections to the old one.

The regression test mirrors the plain-redis one: two routes on the same sentinel master with `redis_database` 1 and 2 must keep their counters in their own databases. Without the bump it fails with both keys landing in db 1. The test helper now also flushes the sentinel master (port 6479) so counters don't survive CI re-runs.

### Checklist

### DIFF
--- a/apisix-master-0.rockspec
+++ b/apisix-master-0.rockspec
@@ -32,7 +32,7 @@ description = {
 
 dependencies = {
     "lua-resty-ctxdump = 0.1-0",
-    "api7-lua-resty-redis-connector = 0.12.0",
+    "api7-lua-resty-redis-connector = 0.13.0",
     "lyaml = 6.2.8-1",
     "api7-lua-resty-dns-client = 7.1.1-0",
     "lua-resty-template = 2.0-1",
--- a/t/lib/test_redis.lua
+++ b/t/lib/test_redis.lua
@@ -23,7 +23,9 @@ local tonumber = tonumber
 
 local _M = {}
 
-local DEFAULT_PORTS = {6379, 5000, 5001, 5002, 5003, 5004, 5005, 5006}
+-- 6479 is the sentinel master used by the redis-sentinel policies; leftover
+-- counters there survive CI re-runs if it is not flushed
+local DEFAULT_PORTS = {6379, 6479, 5000, 5001, 5002, 5003, 5004, 5005, 5006}
 local DEFAULT_HOST = "127.0.0.1"
 
 local function log_warn(...)
--- a/t/plugin/limit-count-redis-sentinel.t
+++ b/t/plugin/limit-count-redis-sentinel.t
@@ -507,3 +507,118 @@ GET /hello
 --- error_code: 500
 --- error_log
 invalid username-password pair
+
+
+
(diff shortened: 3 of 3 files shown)


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
Reviewed for merge safety; no blocking issues found.

PR #13567, opened 2026-06-17
fix(mcp-bridge): encode stderr notifications as JSON

guankai:
### What changed

This fixes how `mcp-bridge` sends stderr notifications from the child process.

Before this, stderr content was put into a JSON string directly. If the stderr line had quotes, backslashes, or newlines, the plugin could send invalid JSON to the MCP client.

This patch builds the notification with `core.json.encode` instead.

I also added docs for `mcp-bridge`, since the plugin is already in the default plugin list but did not have plugin docs yet.

### Tests

Ran locally:

- `bash ./utils/check-plugins-code.sh`
- `python3 utils/check-category.py`
- `PYTHONPATH=/private/tmp/apisix-zhon-check python3 utils/fix-zh-doc-segment.py`
- `git diff --check`

I could not run the full `t/plugin/mcp-bridge.t` locally because this checkout is missing the `test-nginx` Perl dependency.

### DIFF
--- a/apisix/plugins/mcp-bridge.lua
+++ b/apisix/plugins/mcp-bridge.lua
@@ -65,6 +65,28 @@ function _M.check_schema(conf, schema_type)
 end
 
 
+local function build_stderr_notification(content)
+    local message, err = core.json.encode({
+        jsonrpc = "2.0",
+        method = "notifications/stderr",
+        params = {
+            content = content,
+        },
+    })
+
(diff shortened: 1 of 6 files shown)


guankai:
I noticed #13549 is already covering the same mcp-bridge stderr JSON encoding issue.

Closing this one to avoid duplicate work. Thanks.

PR #13568, opened 2026-06-17
feat: add log_format_extra to enrich the default logger format

### LINKED ISSUE #11818: feat: As a user, I want to be added in default log-format the `consumer_group_id` variable
### Description

I need to include a consumer_group_id in the default log variables or have an option to add extra variables to an existing log.

I understand we have log_format, but it is going to take too long, and I need the full log information. Creating this will result in a huge template and a complex plugin configuration.

shreemaan-abhishek:
### Description

Setting `log_format` on a logger plugin (via `plugin_metadata` or plugin `conf`) **replaces** the default rich log entry with a flat custom format. The custom format only substitutes scalar `$variables`, so it silently drops every default field that has no scalar-variable equivalent: the full request/response header maps, the parsed querystring, the composed request URL, the APISIX version, and the computed `apisix_latency`.

Customers often set `log_format` just to add or rename one field, and unknowingly lose all of those.

This PR adds **`log_format_extra`**: an additive log format that keeps the default rich log entry intact and only overlays the user-defined fields on top. It is ignored when `log_format` is set (which keeps the replace semantics).

It also exposes a new core variable **`upstream_unresolved_host`** — the upstream host *before* DNS resolution (the configured domain or IP of the picked node). The default log only carries `upstream` (`$upstream_addr`, the resolved `ip:port`), so this variable lets users log the original hostname via `log_format_extra`.

### Example

```json
"plugin_metadata": {
  "log_format_extra": {
    "upstream_host": "$upstream_unresolved_host"
  }
}
```

The log keeps every default field (`request`, `response`, `server`, `route_id`, ...) and gains `upstream_host`.

### Changes

- `apisix/utils/log-util.lua`: `get_log_entry` builds the default log then overlays `log_format_extra` (http subsystem only, since the default log is http-only).
- `apisix/core/ctx.lua`: register the `upstream_unresolved_host` variable.
- `apisix/balancer.lua`: stash the pre-DNS host on `ctx` when picking a server (single-node and balancer paths).
- All 16 HTTP logger plugins (`file-logger`, `loki-logger`, `clickhouse-logger`, `elasticsearch-logger`, `google-cloud-logging`, `http-logger`, `kafka-logger`, `loggly`, `rocketmq-logger`, `skywalking-logger`, `sls-logger`, `splunk-hec-logging`, `syslog`, `tcp-logger`, `tencent-cloud-cls`, `udp-logger`): accept `log_format_extra` in conf and metadata schemas, documented in each plugin's reference. The stream `syslog` is intentionally excluded (the enrich path is http-only).
- Docs: `loki-logger.md`, `apisix-variable.md`.
- Tests: enrich scenario in `t/plugin/file-logger.t` (self-contained) and `t/plugin/loki-logger.t`.

### Checklist

- [x] Tests added
- [x] Docs updated

Fixes #11818


membphis:
[P2] `log_format_extra` is silently dropped by Google Cloud Logging and Splunk HEC in the default-format path.

This PR adds `log_format_extra` to every logger schema and documents it as an additive field set, but not every logger sends the enriched entry directly. In particular, `google-cloud-logging` and `splunk-hec-logging` call `log_util.get_log_entry(...)` and then rebuild the provider-specific default payload from a fixed list of fields.

Because `log_format_extra` does not set `customized = true`, these plugins still take their default mapping path:

- `google-cloud-logging` only writes fixed fields such as `route_id` and `service_id` into `jsonPayload` for the default entry.
- `splunk-hec-logging` rebuilds `event` from fixed request/response fields.

So a user can configure, for example, `log_format_extra = { upstream_host = "$upstream_unresolved_host" }`, pass schema validation, and still never see `upstream_host` in Google Cloud Logging or Splunk HEC output.

Please either merge extra fields into the final provider-specific payload for these plugins, or do not advertise/support `log_format_extra` for plugins where the default output path drops it. It would also be good to add coverage for at least one provider-specific remapping logger so this does not regress.

shreemaan-abhishek:
Good catch, fixed in the latest commit.

`google-cloud-logging` and `splunk-hec-logging` were the only two loggers that rebuild a fixed provider payload on the default (`not customized`) path (I swept all 18 loggers to confirm). `get_log_entry` now returns the resolved `log_format_extra` table as a third value, and both plugins merge it into their provider payload (`jsonPayload` for GCP, `event` for Splunk) without clobbering existing fixed fields. So `log_format_extra = { upstream_host = "$upstream_unresolved_host" }` now actually shows up in their output.

Added a Splunk test exercising the provider default-path so it cannot regress.

membphis (dismissed):
LGTM. The previous concern around log_format_extra being dropped by provider-specific default payloads has been addressed, and I do not see remaining blocking issues in the updated diff.

apisix/core/ctx.lua
```diff
@@ -244,6 +244,8 @@ do
         route_name = true,
         service_id = true,
         service_name = true,
+        -- the upstream host before DNS resolution (configured domain/host)
+        upstream_unresolved_host = true,
```
nic-6443, line 249: I suggest putting this variable into nginx_tpl.lua so that it can also be used in the access log.

shreemaan-abhishek: Good idea, done in the latest commit (db37b1f82).

Declared `set $upstream_unresolved_host '';` in `ngx_tpl.lua` and routed the balancer writes through `ctx.var` (added it to `ngx_var_names`, removed it from `apisix_var_names`), mirroring how `$llm_content_risk_level` is wired. So it is now a real nginx variable usable directly in the access log format, in addition to the APISIX logger plugins.

Added a `t/core/ctx.t` case that proxies to a domain upstream and asserts the pre-DNS host shows up in the nginx access log.

nic-6443 (dismissed):

nic-6443 (approved):

membphis (approved):

AlinsRan (approved):

Issue #13566, opened 2026-06-16
feat(ai-proxy): support OpenAI Chat Completions clients with native Anthropic Messages API upstreams


GeorgelPreput:
### Description

ai-proxy/ai-proxy-multi can translate Anthropic-format clients → OpenAI-compatible upstreams (via the anthropic-messages-to-openai-chat converter), but not the reverse. There is currently no way to accept OpenAI Chat Completions requests and forward them to an upstream that only speaks the native Anthropic Messages API (/v1/messages).

I'm trying to proxy an Open WebUI message to Claude on Azure AI Foundry, whose [API surface](https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/claude-models#api-surface) is exclusively the native Anthropic Messages API (https://<resource>.services.ai.azure.com/anthropic/v1/messages), exposing no OpenAI-compatible /chat/completions endpoint.

The APISIX anthropic provider's openai-chat capability assumes the upstream offers Anthropic's own OpenAI-compatible endpoint (api.anthropic.com/v1/chat/completions). When pointed at a Messages-only upstream via override.endpoint, the OpenAI body is passed through unchanged and rejected by the upstream.

### Steps to reproduce
Route with ai-proxy-multi, provider: anthropic, and override.endpoint set to an Azure Foundry Claude deployment (.../anthropic/v1/messages). Send a standard OpenAI Chat Completions request:
```
curl "$GATEWAY/v1/chat/completions" -X POST \
  -H "Content-Type: application/json" \
  -d '{
    "model": "claude-opus-4-5",
    "messages": [
      { "role": "system", "content": "You are a mathematician" },
      { "role": "user", "content": "What is 1+1?" }
    ]
  }'
```

Actual behavior
The request is forwarded to the Messages API verbatim (still OpenAI-shaped), and the upstream rejects it:

{"type":"error","error":{"type":"invalid_request_error",
 "message":"messages.0: use the top-level 'system' parameter for the initial system prompt"}}
The OpenAI convention of a system role inside messages[] is incompatible with the Messages API, which requires a top-level system parameter (among other format differences: max_tokens required, content blocks, input_tokens/output_tokens usage fields, etc.).

Expected behavior
[...]

PR #13569, opened 2026-06-17
feat(ai-proxy): support OpenAI Chat Completions clients with native Anthropic Messages API upstreams

GeorgelPreput:
### Description

`ai-proxy`/`ai-proxy-multi` can translate Anthropic-format clients to OpenAI-compatible upstreams (via the `anthropic-messages-to-openai-chat` converter), but not the reverse. There was no way to accept an OpenAI Chat Completions request and forward it to an upstream that only speaks the native Anthropic Messages API (`/v1/messages`) — e.g. **Azure AI Foundry Claude**, whose only surface is `https://<resource>.services.ai.azure.com/anthropic/v1/messages` (no OpenAI-compatible `/chat/completions`). Today an OpenAI body sent through the `anthropic` provider is passed through unchanged and rejected by the upstream (`messages.0: use the top-level 'system' parameter ...`).

This PR adds the missing direction (non-streaming).

**Changes**

- **New `anthropic-compatible` provider** exposing *only* the `anthropic-messages` capability. This is the crux: `ai-proxy/base.lua` selects native passthrough whenever the provider declares a capability matching the client protocol, *before* looking for a converter. The existing `anthropic` provider declares **both** `openai-chat` and `anthropic-messages`, so OpenAI clients always passthrough and never reach a converter. A provider exposing only the Messages capability lets the converter run — mirroring how `openai-compatible` is the vendor-neutral counterpart used as a conversion *target*.
- **New `openai-chat` → `anthropic-messages` converter**:
  - Request: `system` role → top-level `system`; `tools`/`tool_choice`; `tool` messages → user `tool_result` blocks (coalesced); image parts; `reasoning_effort` → `thinking`; `max_tokens`/`max_completion_tokens` → `max_tokens` (defaulted since Anthropic requires it).
  - Non-streaming response: content / `tool_use` / `thinking` → `choices[].message`; stop-reason and usage mapping (incl. cache tokens) → OpenAI shape.
  - Headers: adds `anthropic-version`, strips OpenAI/stainless telemetry.
- **Streaming (`stream: true`) is rejected with a clear error** for now (see follow-ups).
- Tests (`t/plugin/ai-proxy-openai-to-anthropic.t`, unit + e2e) and docs (`ai-proxy.md`, `ai-proxy-multi.md`, en + zh).

Verified end-to-end against a live Azure AI Foundry Claude deployment via `ai-proxy-multi` + `anthropic-compatible`: basic chat (with a `system` prompt), tool-call requests (`tool_use` → `tool_calls`), and a multi-turn tool-result round-trip all returned correct OpenAI-shaped responses; `stream: true` returned a clean 400.

**Open questions for maintainers (documented as potential follow-ups rather than decided here):**

1. **Streaming.** Supporting `stream: true` in this direction needs the Anthropic Messages SSE adapter (`ai-protocols/anthropic-messages.lua` `parse_sse_event`) to pass raw event data through to the converter (it currently digests/drops `tool_use` and `finish_reason` detail). Happy to do this in a follow-up — kept this PR focused to get directional feedback first.
2. **Upstream auth header handling.** The converter currently rewrites an upstream `Authorization: Bearer <key>` to `x-api-key`. This works for Azure/Anthropic **key-based** auth, but would break a **Microsoft Entra ID** bearer token. Options: make the rewrite conditional, drop it and rely solely on `auth.header`, or document it. Guidance welcome.

Partially addresses #13566 (non-streaming; streaming tracked as a follow-up).

### Checklist

- [x] I have verified that this change is backward compatible (purely additive — new provider + new converter; no existing provider or the reverse converter is modified)


GeorgelPreput:
Ummm... Nobody around to even have a look?

Issue #13291, opened 2026-04-24
feat: add ai-lakera-guard plugin for prompt injection and PII detection


nic-6443:
### Description

Add an `ai-lakera-guard` plugin that integrates with the [Lakera Guard API](https://www.lakera.ai/ai-security) for LLM request/response security scanning — covering prompt injection, jailbreak detection, PII leakage, and content moderation.

**Background:**

APISIX currently has `ai-prompt-guard` (regex-based allow/deny), `ai-aws-content-moderation` (AWS Comprehend), and `ai-aliyun-content-moderation` (Aliyun SaaS). However, there is no integration with Lakera Guard, one of the most widely adopted commercial prompt security services. Kong already ships an `ai-lakera-guard` plugin; LiteLLM also supports Lakera as a guardrail backend.

Lakera Guard provides ML-based detection that goes beyond regex patterns — it catches semantic prompt injection, jailbreak attempts, PII entities, profanity, and unknown/malicious links through a single API call.

**Proposed design (for reference, open to adjustment during implementation):**

```yaml
plugins:
  ai-lakera-guard:
    direction: both              # input / output / both
    action: block                # block / alert (log only)

    endpoint:
      url: "https://api.lakera.ai/v2/guard"
      api_key_ref: "$secret://vault/kv/lakera/..."
      timeout_ms: 1000

    categories:
      - prompt_injection
      - jailbreak
      - pii
      - profanity
      - unknown_links

    reveal_failure_categories: true   # include matched categories in error response (for debugging)

    fail_open: false             # reject requests if Lakera API is unavailable (safety-first default)

    on_block:
      status: 400
      body: "Request blocked by security guard"
```

**Key design points:**

1. Follows APISIX's existing pattern of one plugin per external service (same as `ai-aws-content-moderation`, `ai-aliyun-content-moderation`)
2. Supports bidirectional scanning (both input prompts and LLM responses)
3. Users bring their own Lakera account and API key — the plugin is just the connector
4. `fail_open: false` as the safety-first default; users can override for availability-first scenarios
5. `reveal_failure_categories` helps debugging in dev; can be disabled in production
6. Category selection lets users enable only the checks they need
7. Integrates with APISIX secret management (`$secret://`) for API key storage

[...]

PR #13570, opened 2026-06-18
feat: add ai-lakera-guard plugin

janiussyafiq:
### Description

This PR adds a new plugin, **`ai-lakera-guard`**, that integrates APISIX with the [Lakera Guard v2 `/guard` API](https://docs.lakera.ai/) to perform ML-based security scanning of LLM **requests** at the gateway — prompt injection / jailbreak, PII leakage, content-policy violations, and malicious / unknown links — so each backend LLM service no longer has to implement its own guardrails.

This is **PR-1 (input guard MVP)** of a planned, independently shippable series (input → output → streaming → observability), modeled closely on `ai-aliyun-content-moderation`.

#### How it works

- Runs in the `access` phase at priority **1028**, just below `ai-proxy` (1040) and `ai-proxy-multi` (1041), so the AI context is already populated. The plugin is meant to run behind one of those proxies; requests that did not pass through `ai-proxy`/`ai-proxy-multi` are handled per the configurable `fail_mode` (default `skip` — passed through unchecked; set `fail_mode: error` to reject them with `500`).
- Extracts the whole request conversation via `apisix.plugins.ai-protocols` (no role distinction) and sends it to Lakera `POST /v2/guard`.
- On a flagged verdict it applies the configured `action`:
  - **`block`** (default) — returns a **provider-compatible** deny response (a valid chat-completion, or SSE for streaming requests) carrying `request_failure_message`, built via `proto.build_deny_response`, so client SDKs render the refusal as a normal completion. The status is `deny_code` (default `200`; set a 4xx to surface blocks as HTTP errors).
  - **`alert`** — log-only shadow mode; traffic passes through.
- Lakera errors / timeouts are governed by `fail_open` (fail-**closed** by default).
- `api_key` is secret-managed via `encrypt_fields` + native `$secret://` / `$env://` resolution.
- `reveal_failure_categories` optionally appends the matched detectors to the deny message; every flagged verdict logs Lakera's full per-detector breakdown and `request_uuid`.

#### Configuration

`api_key` is the only required field. Others: `lakera_endpoint`, `project_id`, `direction` (`input` only in this PR), `action`, `fail_open`, `timeout`, `ssl_verify`, `reveal_failure_categories`, `deny_code`, `request_failure_message`.

#### Files

- Plugin: `apisix/plugins/ai-lakera-guard.lua`, `apisix/plugins/ai-lakera-guard/schema.lua`, `apisix/plugins/ai-lakera-guard/client.lua`
- Registration: `apisix/cli/config.lua`, `conf/config.yaml.example`
- Docs: `docs/en/latest/plugins/ai-lakera-guard.md`, `docs/en/latest/config.json`
- Tests: `t/plugin/ai-lakera-guard.t`, `t/plugin/ai-lakera-guard-secrets.t`, fixtures under `t/fixtures/lakera/`

Part of #13291

### Checklist

- [x] I have verified that this change is backward compatible (new, opt-in plugin disabled by default; additive registration only)

### DIFF
--- a/apisix/plugins/ai-lakera-guard.lua
+++ b/apisix/plugins/ai-lakera-guard.lua
@@ -18,6 +18,7 @@ local core       = require("apisix.core")
 local schema_mod = require("apisix.plugins.ai-lakera-guard.schema")
 local client     = require("apisix.plugins.ai-lakera-guard.client")
 local protocols  = require("apisix.plugins.ai-protocols")
+local binding    = require("apisix.plugins.ai-protocols.binding")
 
 local ipairs = ipairs
 local type   = type
@@ -116,8 +117,15 @@ end
 
 function _M.access(conf, ctx)
     if not ctx.picked_ai_instance then
-        return 500, "no ai instance picked, ai-lakera-guard plugin must be used with "
-                    .. "ai-proxy or ai-proxy-multi plugin"
+        local handled, code, body = binding.on_unsupported(
+            conf.fail_mode, _M.name, ctx,
+            "no ai instance picked (request did not pass through ai-proxy/ai-proxy-multi)",
+            500, "no ai instance picked, ai-lakera-guard plugin must be used with "
+                 .. "ai-proxy or ai-proxy-multi plugin")
+        if handled then
+            return code, body
+        end
+        return
     end
 
     -- ai-proxy / ai-proxy-multi runs first (higher priority) and already
@@ -127,7 +135,14 @@ function _M.access(conf, ctx)
 
     local proto = protocols.get(ctx.ai_client_protocol)
     if not proto or not proto.extract_request_content then
-        return 500, "unsupported protocol: " .. (ctx.ai_client_protocol or "unknown")
+        local handled, code, body = binding.on_unsupported(
+            conf.fail_mode, _M.name, ctx,
+            "unsupported protocol: " .. (ctx.ai_client_protocol or "unknown"),
+            500, "unsupported protocol: " .. (ctx.ai_client_protocol or "unknown"))
+        if handled then
+            return code, body
+        end
+        return
     end
 
     local contents = proto.extract_request_content(request_tab)
--- a/apisix/plugins/ai-lakera-guard.lua
+++ b/apisix/plugins/ai-lakera-guard.lua
@@ -82,12 +82,44 @@ local function deny_message(ctx, conf, message, breakdown)
 end
 
 
-local function request_content_moderation(ctx, conf, content)
-    if not content or #content == 0 then
+-- Normalize a protocol's canonical {role, content} messages into the shape
+-- Lakera /v2/guard accepts: role preserved, content coerced to a plain string.
+-- Some adapters (e.g. openai-chat) return body.messages verbatim, so a message's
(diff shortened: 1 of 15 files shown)


apisix/plugins/ai-lakera-guard.lua
```diff
@@ -0,0 +1,148 @@
...
+
+    return conf.deny_code, deny_message(ctx, conf, conf.request_failure_message, result.breakdown)
+end
+
+
+function _M.access(conf, ctx)
+    if not ctx.picked_ai_instance then
+        return 500, "no ai instance picked, ai-lakera-guard plugin must be used with "
+                    .. "ai-proxy or ai-proxy-multi plugin"
+    end
```
nic-6443, line 161: ref: https://github.com/apache/apisix/blob/6f62be4f9b99a30fdc5b23c1622eb0dcdfffc5dd/apisix/plugins/ai-aliyun-content-moderation.lua#L309-L320

janiussyafiq: fixed

nic-6443 (dismissed):

nic-6443 (dismissed):

membphis:
### P1: Preserve Lakera message roles instead of flattening the conversation into one user message

The plugin currently calls `proto.extract_request_content(request_tab)`, concatenates all extracted text, and `client.scan` sends the result as:

```lua
messages = { { role = "user", content = content } }
```

This loses the original role and turn boundaries. For OpenAI Chat, this can turn system, assistant, historical user, and current user content into one current `user` message. For Anthropic and Responses requests, the protocol adapters already have role-preserving canonical message helpers, so flattening here bypasses information the codebase can keep.

Why this blocks merge: Lakera Guard's `/v2/guard` API is message-based, and role/context semantics matter for policy behavior. Sending the system prompt, assistant output, or older history as a new user message can block valid follow-up requests because old or non-user content is rescanned as the current user input. It can also make the gateway's enforcement differ from the API contract this plugin is integrating with.

Suggested fix:

- Pass a `messages` array to `client.scan`, not a flattened string.
- Build it from the protocol-normalized message helper, preserving `system`, `user`, and `assistant` roles where available.
- Only fall back to one `user` message when the protocol has no role-preserving representation.
- Update the "whole conversation is scanned" test to verify the full message array is sent without converting history/system/assistant messages into the latest user input.

janiussyafiq:
Reply to @membphis :
- `client.scan` now receives and forwards a role-tagged `messages` array instead of a concatenated string.
- The array is built from `proto.get_messages()` — the protocol's canonical `{role, content}` helper — so `system`/`user`/`assistant` turns are preserved for openai-chat, Anthropic and Responses.
- Each message's content is coerced to text and non-text parts (e.g. multimodal `image_url`) are dropped. Lakera `/v2/guard` rejects non-text content with HTTP 400, which under the default fail-closed mode would otherwise block legitimate multimodal requests.
- Falls back to a single `user` message only when a protocol exposes no role-preserving representation.
- The "whole conversation" test now asserts the full role-tagged array reaches Lakera, with each role paired to its own content.

This also matches how other gateways integrate Lakera.

shreemaan-abhishek (approved):

membphis (approved):

nic-6443 (approved):

Issue #13290, opened 2026-04-24
feat: add ai-cache plugin for LLM semantic caching


nic-6443:
### Description

Add an `ai-cache` plugin that provides two-layer caching for LLM API responses:

- **L1 exact-match**: hash-based lookup on the prompt content, zero cost, deterministic
- **L2 semantic**: embedding-based vector similarity search when L1 misses, catching semantically equivalent queries (e.g., "how to return an item" ≈ "what is the return policy")

This is one of the highest-ROI capabilities for AI Gateway — industry reports show 30-60% cost and latency savings for common workloads (FAQ bots, document Q&A, translation).

**Background:**

All major AI gateway products already ship semantic caching: Kong (`ai-semantic-cache`), LiteLLM, Portkey, Higress (`ai-cache`), Helicone. APISIX has `proxy-cache` for generic HTTP caching but nothing that understands LLM prompt semantics.

**Proposed design (for reference, open to adjustment during implementation):**

```yaml
plugins:
  ai-cache:
    layers: [exact, semantic]

    cache_key:
      include_consumer: false    # default shared; set true for multi-tenant isolation
      include_vars: []           # additional variables for cache key scoping

    exact:
      ttl: 3600
      match_fields: [messages]

    semantic:
      vector_backend: redis      # Redis Stack with RediSearch module
      similarity_threshold: 0.95
      top_k: 1
      ttl: 86400
      embedding:
        provider: openai
        model: text-embedding-3-small
        endpoint: "https://api.openai.com/v1/embeddings"
        auth_ref: "$secret://..."

    bypass_on:
      - header: "X-Cache-Bypass"
        equals: "1"

    headers:
      cache_status: "X-AI-Cache-Status"      # HIT-L1 / HIT-L2 / MISS / BYPASS
      cache_similarity: "X-AI-Cache-Similarity"
      cache_age: "X-AI-Cache-Age"
```

**Key design points:**

1. L1 hash → L2 embedding dual-layer (industry consensus, not either/or)
2. Cache key isolation is user-configured: default is shared (maximizes hit rate for public knowledge scenarios); multi-tenant or RAG scenarios explicitly enable `include_consumer` or `include_vars`
3. L2 hit backfills L1 for next identical query
4. Cache write only on complete successful upstream response (2xx)
5. Streaming responses: accumulate chunks then write; cache replay maintains SSE contract
[...]

PR #13578, opened 2026-06-19
feat: add ai-cache plugin

janiussyafiq:
### Description

Adds a new `ai-cache` plugin that caches LLM responses and replays them for subsequent requests that resolve to the same prompt, cutting upstream token cost and latency for repetitive workloads (FAQ bots, document Q&A, translation).

This PR implements the **exact (L1)** cache layer:

- **Cache key** — a SHA-256 fingerprint of the request as received: client protocol, requested model, normalized messages, and the remaining response-determining body parameters (`temperature`, `top_p`, `max_tokens`, `tools`, …). Provider-agnostic via `ai-protocols`, so it works for every chat protocol `ai-proxy` supports (OpenAI Chat, Anthropic Messages, Bedrock Converse, OpenAI Responses). The key also segments by the **selected AI instance** — the `ai-proxy` provider, or the `ai-proxy-multi` instance picked for the request (recomputed if a `retry_on_error` fallback answers) — so identical prompts that resolve to instances backed by different models or providers never share an entry.
- **Storage** — Redis (single-node); connection fields are sourced from `apisix.utils.redis-schema` via the `policy` + `if/then` convention used by `limit-count` / `limit-req` / `limit-conn`.
- **Scope** — per-route by default (`cache_key.share_across_routes` to share one cache space across routes); opt-in per-consumer / per-variable isolation (`cache_key.include_consumer` / `include_vars`).
- **Behavior** — write-on-200 only (non-streaming); `bypass_on` opt-out (exact request-header match); `fail_mode` (`skip` / `warn` / `error`) when a request did not pass through `ai-proxy` / `ai-proxy-multi`; `max_cache_body_size` cap; `X-AI-Cache-Status` / `X-AI-Cache-Age` response headers; fails open (proxies as a normal miss) when Redis is unreachable.
- Runs below `ai-proxy` (priority `1035`) and depends on `ai-proxy` / `ai-proxy-multi`.

Semantic cache, streaming support, and observability are planned as follow-up PRs. User-facing documentation will be added in a later PR once the series is further along.

Related to #13290

### Checklist


apisix/plugins/ai-cache/key.lua
```diff
@@ -0,0 +1,72 @@
...
+end
+
+
+function _M.fingerprint(ctx, body)
+    local params = core.table.deepcopy(body)
+    params.messages = nil
+    params.model    = nil
+    params.stream   = nil
+
+    local repr = core.json.stably_encode({
```
nic-6443: ref: https://github.com/apache/apisix/blob/34941173a5e1203b979bfe1d7d157b3260b1d57e/apisix/plugins/ai-transport/http.lua#L106-L117

apisix/plugins/ai-cache/key.lua
```diff
@@ -0,0 +1,108 @@
...
+        obj[k] = to_rapidjson_value(v)
+    end
+    return obj
+end
+
+
+local function canonical_encode(value)
+    return rapidjson.encode(to_rapidjson_value(value), ENCODE_OPTS)
+end
+
```
nic-6443: Reuse the code from AI Proxy. You may consider placing all of these in `apisix.core.json` as public utility functions.

apisix/plugins/ai-cache/key.lua
```diff
@@ -0,0 +1,108 @@
...
+    })
+    return hex_digest(repr)
+end
+
+
+function _M.scope(conf, ctx)
+    local ck = conf.cache_key
+    local inc_vars = ck and ck.include_vars
+    if not (ck and ck.include_consumer) and (not inc_vars or #inc_vars == 0) then
+        return "shared"
```
nic-6443: The default scope is just the literal `shared`, with no route in it. So if two routes both enable ai-cache against the same Redis and see the same protocol + requested model + messages, they compute the same key and serve each other's responses, even when they proxy to completely different upstreams (e.g. one route -> OpenAI `gpt-4o`, another -> a self-hosted `gpt-4o`). The docs `:::note` only warns about server-side model rewrite via `options.model`, but this collision happens with no rewrite at all.

Would it make sense to fold `route_id` into the default scope, so cross-route sharing is opt-in rather than the default footgun? What do you think?

apisix/plugins/ai-cache.lua
```diff
@@ -0,0 +1,199 @@
...
+    if not red then
+        -- fail-open: never let a cache-backend outage break the request.
+        core.log.warn("ai-cache: redis unavailable, fail-open as MISS: ", err)
+        ctx.ai_cache_status = "MISS"
+        return
+    end
+
+    local res
+    res, err = red:get(ctx.ai_cache_key)
+    release(conf, red)
```
nic-6443: `release()` runs unconditionally here, before the `if err` check just below, so a connection that just errored/timed out on `get` still gets handed back to the keepalive pool. lua-resty-redis recommends not reusing a connection after a failure. The limit-count helper handles this by returning on error and only calling `set_keepalive` on the success path (see `apisix/plugins/limit-count/util.lua`).

Suggest releasing only when there's no error (and `red:close()` otherwise), so one transient timeout doesn't poison pooled connections for subsequent requests.

apisix/plugins/ai-cache.lua
```diff
@@ -0,0 +1,199 @@
...
+        return
+    end
+    local envelope = core.json.encode({ body = response_body, created_at = ngx.time() })
+    local ttl = (conf.exact and conf.exact.ttl) or DEFAULT_TTL
+    local ok
+    ok, err = red:set(cache_key, envelope, "EX", ttl)
+    if not ok then
+        core.log.warn("ai-cache: redis set failed: ", err)
+    end
+    release(conf, red)
```
nic-6443, line 195: Same keepalive-on-error issue as the `get` path above: `release()` is called regardless of whether `red:set` failed. Better to skip the keepalive (close instead) when `err` is set.

apisix/plugins/ai-cache/schema.lua
```diff
@@ -0,0 +1,95 @@
...
+
+local core         = require("apisix.core")
+local redis_schema = require("apisix.utils.redis-schema")
+
+local policy_to_additional_properties = core.table.deepcopy(redis_schema.schema)
+
+local _M = {
+    type = "object",
+    properties = {
+        layers = {
```
nic-6443: `layers` is validated and documented (defaults to `["exact"]`), but nothing in the plugin ever reads `conf.layers` — exact caching always runs unconditionally. Since `exact` is the only allowed value today, maybe drop the knob until the semantic/L2 layer actually lands, or at least note that it's currently inert? As is, it reads like a working toggle that does nothing.

janiussyafiq: dropped the knob until L2

apisix/plugins/ai-cache.lua
janiussyafiq: fixed

apisix/plugins/ai-cache.lua
janiussyafiq: fixed

apisix/plugins/ai-cache/key.lua
janiussyafiq: good catch, at first i though if users can specify route_id explicitly in include_vars can solve the issue, but it seems the approach u suggested worked better. Implemented share_across_route within cache_key

apisix/plugins/ai-cache/key.lua
janiussyafiq: fixed

nic-6443 (dismissed):
Thanks for the quick turnaround — all my comments are addressed: per-route scoping by default with `share_across_routes` opt-out, `red:close()` on Redis errors instead of pooling a broken connection, the dead `layers` knob dropped, and the canonical encoding pulled up into `core.json.canonical_encode` (nicely de-duped with ai-transport). LGTM.

shreemaan-abhishek (dismissed):

membphis:
I found two merge-blocking issues in the current `ai-cache` implementation:

### [P1] Cache key does not include the effective model or picked AI instance

`ai-cache` computes the fingerprint from `ctx.var.request_llm_model or body.model`, but it does not include `ctx.picked_ai_instance_name`, provider, or the route / instance effective `options.model`:

- `apisix/plugins/ai-cache/key.lua`: the fingerprint uses only protocol, requested model, normalized messages, and remaining body params.
- `apisix/plugins/ai-cache.lua`: the lookup happens in `access`, before the upstream request is built.
- `ai-proxy-multi` has already selected `ctx.picked_ai_instance` before lower-priority plugins run, so that selected instance is available at cache lookup time.

This can return the wrong provider/model response on an `ai-proxy-multi` route. A request can warm the cache through instance A, then a later identical request can be routed to instance B but still hit and replay instance A's response because both requests share the same cache key.

This should be fixed before merge by including the selected AI instance and/or effective model/provider in the cache key or scope, with a regression test covering `ai-proxy-multi` instances that use different models or providers.

### [P2] The plugin can cache ordinary JSON traffic when it is not behind `ai-proxy`

The docs say `ai-cache` must be used with `ai-proxy` or `ai-proxy-multi`, but the implementation does not enforce or safely bypass that condition. `ai-cache.access` reads any JSON request body, computes a key, and marks the request as `MISS`; then `log` writes any 200 response to Redis. There is no `ctx.picked_ai_instance` guard like the existing AI moderation plugins use.

If the plugin is accidentally attached at Route / Service / Consumer level without an AI proxy, ordinary JSON upstream responses can be cached and replayed. That is a surprising behavior and can leak stale or incorrect non-AI responses.

Please add a guard before key computation, either bypassing by default or using the shared `ai-protocols.binding` `fail_mode` behavior, and add coverage for the no-`ai-proxy` case.

membphis:
I rechecked the latest update. The no-`ai-proxy` guard looks addressed now: `ai-cache.access` checks `ctx.picked_ai_instance` first, uses the shared `ai-protocols.binding` `fail_mode`, and the tests cover both the default bypass behavior and `fail_mode=error`.

One merge-blocking cache-key issue still remains:

### [P1] `share_across_routes` can still reuse a response across different effective models on plain `ai-proxy` routes

The new scope includes `ctx.picked_ai_instance_name`, which fixes the `ai-proxy-multi` case because different picked instances have different names. However, for the plain `ai-proxy` plugin, `ctx.picked_ai_instance_name` is only `ai-proxy-<provider>` (for example, `ai-proxy-openai`). It does not include the route-level effective model.

The fingerprint still uses the requested model from `ctx.var.request_llm_model or body.model`, not the effective model selected by `ai-proxy` from `ai_instance.options.model or request_model`. This leaves a real collision when cache sharing is enabled across routes:

1. Route A uses `ai-proxy` with `provider=openai`, `options.model=gpt-4o`, and `ai-cache.cache_key.share_across_routes=true`.
2. Route B uses `ai-proxy` with `provider=openai`, `options.model=gpt-4o-mini`, and the same Redis/cache settings.
3. The client sends the same body to both routes, especially if the body omits `model` or carries the same requested `model`.
4. Both routes compute the same scope (`instance=ai-proxy-openai`) and the same fingerprint, so Route B can replay Route A's cached response even though the effective upstream model is different.

This also contradicts the new docs, which say that even with `share_across_routes` enabled, responses from different upstream models or providers are kept in separate cache entries.

Please include the effective model in the key/scope for the plain `ai-proxy` case, for example by deriving it from `ctx.picked_ai_instance.options.model or body.model` before lookup. It would also be good to add a regression test with two plain `ai-proxy` routes using the same provider, different `options.model`, the same Redis, and `share_across_routes=true`; the second route should be a `MISS`, not a `HIT`.

nic-6443 (dismissed):

membphis:
I think this should be addressed before merge: the cache key is currently computed before `ai-proxy` builds the final upstream request, so it does not cover all server-side mutations that can affect the LLM response.

`ai-cache` computes `ctx.ai_cache_fingerprint` in `access` from the client request body and then scopes the key mostly by instance/model/route/consumer/vars. However, `ai-proxy` later mutates the request that is actually sent upstream through `options`, `override.llm_options`, and `override.request_body`. Those fields can change response-determining parameters such as `temperature`, `top_p`, `max_tokens`, tools, provider-specific body fields, etc.

A concrete failure case:

1. Two routes enable `ai-cache` with `cache_key.share_across_routes = true`.
2. The client sends the same prompt and model to both routes.
3. Route A configures `ai-proxy.options.temperature = 0.2`.
4. Route B configures `ai-proxy.options.temperature = 0.8`, or uses `override.request_body` / `override.llm_options` to change the final upstream request.
5. Route A warms the cache first.
6. Route B can hit Route A's cached response even though the actual upstream LLM request should be different.

This breaks the exact-cache contract because the key no longer represents the final request that produced the cached response. It can return a response generated under a different route-side AI configuration.

Suggested fix: base the cache key on the final canonical upstream request body after `ai-proxy` has applied provider conversion, `options`, `override.llm_options`, and `override.request_body`; or include a canonicalized representation of all response-determining AI instance configuration in the key. Please also add regression tests covering shared-cache routes with identical client bodies but different server-side `options` / `override` values.

janiussyafiq:
@membphis 
Addressed all the previous concerns. The three cache-key reports were one root cause: the key was computed in `access` from the client request, but the response is determined by the *effective* upstream request `ai-proxy` builds later in `before_proxy` (after `ai-cache` runs), so server-side `options`/`override` never reached the key.

The fingerprint now identifies that effective request. Since `final_body = f(client_body, protocol, instance{provider, options, override})` is deterministic, it hashes those inputs:

- **client:** protocol, messages, params
- **effective:** provider, effective model (`options.model or body.model`), `options`, `override.llm_options`, `override.request_body` (+ `request_body_force_override`), `override.endpoint`

`scope()` is reduced to pure isolation (route / consumer / `include_vars`). One rule closes all three cases, the plain-`ai-proxy` effective model, the `options`/`override` params, and the earlier instance/provider collision.

Keying off the final upstream body directly isn't feasible: `before_proxy` is where that body is built *and* where the upstream call happens, so a read-through cache that short-circuits in `access` can't see it but hashing the builder's inputs is equivalent.

While reworking this I also closed a related case: on `ai-proxy-multi` failover, the fallback instance's response could be written under the originally-picked instance's key. The log-phase write is now skipped when the serving instance differs from the one the key was computed for. Let me know what your view on this.

membphis:
I studied your method, and after your explanation, I agree with the current implementation.

membphis (dismissed):

membphis (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13583, opened 2026-06-22
fix(ai-proxy): drop tool_choice without tools and stop hanging Anthropic stream

AlinsRan:
### Description

Two bugs in the Anthropic-to-OpenAI converter (`apisix/plugins/ai-protocols/converters/anthropic-messages-to-openai-chat.lua`):

1. **`tool_choice` forwarded without `tools` → upstream 400.** `tool_choice` (and `parallel_tool_calls`) were sent to the OpenAI-compatible upstream even when **no tools** ended up in the request — either the client provided none, or all were dropped as Anthropic built-ins (e.g. `web_search`) / invalid. OpenAI-compatible upstreams reject `tool_choice` without `tools`. They are now dropped when the tools array is empty.

2. **Streaming client hang.** When a stream emitted `message_start` but the upstream sent `[DONE]` before any content block opened and without a `finish_reason` chunk, `convert_sse_events` returned `nil` and never emitted `message_stop`, leaving the client hanging until timeout. `message_delta` + `message_stop` are now emitted whenever the message was already started.

### Tests

Added two unit tests to `t/plugin/ai-proxy-anthropic.t`:
- `tool_choice`/`parallel_tool_calls` dropped when no tools are forwarded (no-tools and all-built-in cases); preserved when a real tool remains.
- `[DONE]` after `message_start` without a content block still emits `message_stop`.

### Checklist

- [x] I have updated the documentation to reflect this change (N/A — bugfix, no config/behavior doc change)

### DIFF
--- a/apisix/plugins/ai-protocols/converters/anthropic-messages-to-openai-chat.lua
+++ b/apisix/plugins/ai-protocols/converters/anthropic-messages-to-openai-chat.lua
@@ -579,6 +579,15 @@ function _M.convert_request(request_table, ctx)
         end
     end
 
+    -- tool_choice and parallel_tool_calls are only valid alongside a non-empty
+    -- tools array. If no tools are forwarded to the upstream -- either none were
+    -- provided or all were dropped (Anthropic built-ins / invalid) -- drop them
+    -- to avoid the OpenAI-compatible upstream rejecting the request.
+    if openai_body.tools == nil then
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

membphis (approved):
Approved. No blocking issues found.

shreemaan-abhishek (approved):

PR #13592, opened 2026-06-22
fix(ai-proxy-multi): guard construct_upstream in healthcheck timers

shreemaan-abhishek:
### Description

`ai-proxy-multi`'s `construct_upstream` returns `nil, err` on its normal failure path (unresolvable endpoint, missing host/port, nil instance) and can also raise a Lua error. The two healthcheck-manager timers called it without a `pcall` or a nil check:

```lua
upstream = plugin.construct_upstream(upstream_constructor_config)
upstream.resource_key = resource_path
```

When `construct_upstream` returned `nil` (or threw), the following `upstream.resource_key = ...` / `upstream._nodes_ver` access raised `attempt to index a nil value` inside the timer callback that iterates **all** healthcheck resources. A single malformed `ai-proxy-multi` instance could therefore abort checker creation/refresh for every resource in that worker, not just the offending one.

This wraps both call sites in `pcall`. On failure the timer logs and skips the resource (creation path) or keeps the existing checker (refresh path) instead of crashing.

### Tests

Adds `t/plugin/ai-proxy-multi-construct-upstream-panic.t` covering both timer paths: a `construct_upstream` override that raises a Lua error during checker creation and during the working-pool check. Proxying stays unaffected and the guarded message is logged instead of the worker timer dying.

### Checklist

- [x] I have updated the documentation to reflect this change (N/A: internal robustness fix, no user-facing behavior change)

### DIFF
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -256,8 +256,11 @@ local function timer_working_pool_check()
                 local plugin = require("apisix.plugins." .. plugin_name)
                 ok, upstream, err = pcall(plugin.construct_upstream, upstream_constructor_config)
                 if not ok or not upstream then
-                    -- retain the existing checker through a transient failure
-                    need_destroy = false
+                    -- a nil constructor config means the instance was removed, so let
+                    -- the checker be destroyed; otherwise keep it through a transient failure
+                    if upstream_constructor_config ~= nil then
+                        need_destroy = false
+                    end
                     err = err or upstream or "unknown error"
                     upstream = nil
                     local err_msg = "[checking checker] unable to construct upstream for plugin: "
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -256,7 +256,9 @@ local function timer_working_pool_check()
                 local plugin = require("apisix.plugins." .. plugin_name)
                 ok, upstream, err = pcall(plugin.construct_upstream, upstream_constructor_config)
                 if not ok or not upstream then
-                    err = err or upstream
+                    -- retain the existing checker through a transient failure
+                    need_destroy = false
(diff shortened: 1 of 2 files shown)


AlinsRan (dismissed):

apisix/healthcheck_manager.lua
```diff
@@ -247,17 +254,34 @@ local function timer_working_pool_check()
                 --- callback construct_upstream to create an upstream dynamically
                 local upstream_constructor_config = jp.value(res_conf.value, json_path)
                 local plugin = require("apisix.plugins." .. plugin_name)
-                upstream = plugin.construct_upstream(upstream_constructor_config)
-                upstream.resource_key = resource_path
+                ok, upstream, err = pcall(plugin.construct_upstream, upstream_constructor_config)
```
membphis, line 257: [P2] This keeps the old checker for every construct_upstream failure. If the JSON path no longer exists because the ai-proxy-multi plugin or instance was removed, the old healthchecker can stay running. Consider treating upstream_constructor_config == nil as a real removal and allowing the checker to be destroyed.

membphis (dismissed):
Approved with one follow-up comment.

apisix/healthcheck_manager.lua
shreemaan-abhishek: Good catch, thanks. You're right that retaining on every failure would leak a checker when the instance/plugin is genuinely removed. Fixed: the retain is now gated on `upstream_constructor_config ~= nil`, so a nil constructor config (json path no longer resolves) is treated as a real removal and the checker is destroyed, while a transient `construct_upstream` failure still keeps it. Added a regression test that removes the plugin from the route and asserts the stale checker is released.

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):

PR #13597, opened 2026-06-23
feat(proxy-rewrite): support multiple same-name headers in headers.add/set

### LINKED ISSUE #11587: bug: proxy-rewrite header priority wrong

### LINKED ISSUE #13163: bug: proxy-rewrite silently drops duplicate keys in headers.add due to JSON object limitation

AlinsRan:
### Description

Allow `headers.set` and `headers.add` values in `proxy-rewrite` to be an **array**, producing multiple headers with the same name:

```json
"proxy-rewrite": {
  "headers": {
    "set": { "X-Api-Version": ["v1", "v2"] }
  }
}
```

- **`set`** with an array replaces any incoming header with the listed values.
- **`add`** with an array appends the listed values to the existing ones.
- **`remove`** is unchanged (it already takes an array of names).

### Why

`proxy-rewrite` models `headers.add/set` as a JSON object (map), which structurally cannot express same-name multi-value headers — a JSON object can't hold two values for one key. This is a real gap vs. other gateways that model header transforms as an ordered list of (name, value) pairs (Kong `request-transformer` `append.headers`, Envoy `request_headers_to_add`, HAProxy `add-header`). The array form closes it. Genuine use cases: gRPC metadata (repeated keys → repeated HTTP/2 headers) and backends/auth proxies that read repeated custom headers.

Fully backward compatible: existing string/number values keep working; the array is an additive schema option. Each element still supports NGINX variables and `regex_uri` captures.

This also gives the clean answer to the `remove + add` ordering confusion in #11587 — multi-value is a data-model feature, not an operation-order one.

Closes #11587 
Closes #13163

### Changes

- `apisix/plugins/proxy-rewrite.lua`: schema allows `array<string|number>` for `add`/`set` values; apply loop resolves each element (`set` → one `set_header` with a table; `add` → one `add_header` per element).
- `t/plugin/proxy-rewrite.t`: tests for `set` (replace with multiple) and `add` (append multiple); `t/lib/server.lua`: echo helper joining same-name headers.
- `docs/{en,zh}/.../proxy-rewrite.md`: attribute docs + example.

### Checklist

- [x] I have verified that this change is backward compatible (if not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/blob/master/CONTRIBUTING.md#submit-a-proposal) first)


apisix/plugins/proxy-rewrite.lua
```diff
@@ -106,6 +118,18 @@ local schema = {
...
+                                            -- values, e.g. ["v1", "v2"]
+                                            type = "array",
+                                            minItems = 1,
+                                            items = {
+                                                oneOf = {
+                                                    { type = "string" },
+                                                    { type = "number" },
+                                                }
+                                            }
+                                        }
```
nic-6443, line 132: The semantic meaning of SET is replace; there is no need to support arrays, right?

membphis (dismissed):
LGTM

apisix/plugins/proxy-rewrite.lua
AlinsRan: Supporting multiple values directly on set is exactly what avoids complexity here. 
SET still means replace here — set with an array replaces the header with multiple values rather than a single one, which is a capability add (append, can't drop existing) and scalar set (single value) can't express. Since the action order is add→set→remove, a remove+add combination can't atomically "replace with multiple values" either.

nic-6443 (dismissed):

shreemaan-abhishek (dismissed):

shreemaan-abhishek (commented):
I think both "multiple values for same header" and "same header with multiple values - (comma separated or not)" are the same.

So even without this PR, `set` can already `set` multiple values for a header by specifying comma separated values.

The `add` feature in this PR is nice to have.

AlinsRan:
> I think both "multiple values for same header" and "same header with multiple values - (comma separated or not)" are the same.


I don't think they're fully interchangeable. A headers.set object can't express same-name multi-value headers at all (a JSON object can't hold two values for one key), and emitting repeated lines is a first-class, intentional behavior — both HTTP and the underlying OpenResty support it (ngx.req.set_header(name, {...})). The array closes that gap.

Real headers that rely on multiple same-name fields rather than a comma list:
- Set-Cookie — non-combinable by spec.
- Cookie over HTTP/2 — carried as multiple field lines, recombined with "; " not ",".
- gRPC metadata — a multimap; in practice grpc-go/grpc-java neither fold duplicates nor split on commas, so ["a","b"] reaches the app as two entries and "a,b" as one.

membphis (approved):
LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13600, opened 2026-06-23
fix(limit-conn): use parent resource key for consumer isolation

### LINKED ISSUE #13584: bug: Is it expected that per-consumer limit-conn is isolated per-route and per-pod?
### Description

We noticed an inconsistency in how the `limit-conn` plugin handles consumer-level limits compared to the `limit-req` plugin.

When configuring the `limit-conn` plugin directly on a **Consumer** object:
1. Requests made by the same consumer to **Route A** and **Route B** do not share the same concurrency counter (cross-route leakage).
2. Requests made across **multiple APISIX instances (replica pods)** do not share the counter because different instances can end up using slightly different configuration versions internally, generating different Redis keys (cross-pod leakage).
3. Under the hood, the Redis/Valkey key is constructed using `ctx.conf_version` and `ctx.conf_type` (in `apisix/plugins/limit-conn/init.lua`):
   ```lua
   key = key .. ctx.conf_type .. ctx.conf_version
   ```
   This isolates the counter per-route and across instances (pods).

We recall that for `limit-req`, a similar issue was fixed in PR #13019 by using `parent.resource_key` to share the counter across routes and instances at the consumer level.

We would like to ask the community/maintainers:
* Is this behavior intended for `limit-conn`, or is it an inconsistency/oversight that should be aligned with how `limit-req` works to support shared counters across routes and multiple gateway instances?

### Expected Behavior

_No response_

### Error Logs

_No response_

### Steps to Reproduce

1. Define a consumer with the `limit-conn` plugin activated (concurrency limit `conn: 1`, policy `redis` / `redis-cluster`).
2. Deploy APISIX with 2 replica pods.
3. Scenario A (Cross-route): Make concurrent requests to `/route-a` and `/route-b` using the same consumer credentials.
4. Scenario B (Cross-pod): Make concurrent requests targeting different pods (Pod A and Pod B) for the same consumer credentials.
5. **Observation:** Both requests succeed simultaneously. Checking Redis shows separate keys generated because the suffix (`ctx.conf_version` / `ctx.conf_type`) differs between the routes/pods.

### Environment

- APISIX version: >= 3.16.0 (including 3.17.0)

DanielWu-star:
Previously, when limit-conn was configured at the consumer level, connection limits were applied per-route instead of globally per consumer. This occurred because the key generation concatenated ctx.conf_type and ctx.conf_version, which included route-specific information.

This commit fixes the issue by using conf._meta.parent.resource_key for key generation, consistent with the same fix already applied to limit-req (#13019) and limit-count plugins. Now consumer-level connection limits are properly shared across all routes accessed by the same consumer.

Changes:
- Added gen_limit_key() function using parent.resource_key approach
- Replaced key = key .. ctx.conf_type .. ctx.conf_version with gen_limit_key(conf, ctx, key) in run_limit_conn()
- Updated test cases to match new key format
- Added test cases (TEST 35-42) verifying consumer isolation across multiple routes

Fixes #13584

### Checklist


membphis (approved):
LGTM

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13606, opened 2026-06-25
feat(ai-lakera-guard): scan LLM responses (direction output/both, non-streaming + streaming)

janiussyafiq:
### Description

PR-2 of `ai-lakera-guard`, following the input-scanning MVP (#13570). Adds **response (output) scanning** for non-streaming and streaming (SSE) traffic. Back-compatible: `direction` still defaults to `input`.

- **Schema:** `direction` extended to `input`/`output`/`both`; adds `response_failure_message`.
- **Response path** (`lua_body_filter`, the same dispatch `ai-aliyun-content-moderation` uses):
  - **Non-streaming:** scans `ctx.var.llm_response_text`; a flagged response is replaced with a provider-compatible deny.
  - **Streaming:** buffers the SSE response, scans the assembled completion once, then releases it verbatim (clean) or replaces it with a deny SSE terminated by `[DONE]` (flagged). Because the stream's `200`/`text/event-stream` headers are already committed when buffering begins, a streamed block is delivered as the deny **body** — `deny_code` does not apply to streams.
- Docs (en + zh) and tests added.

Part of #13291.

### Checklist


apisix/plugins/ai-lakera-guard.lua
```diff
@@ -206,4 +215,60 @@ function _M.access(conf, ctx)
...
+        end
+        buffer[#buffer + 1] = body or ""
+
+        if not ctx.var.llm_request_done then
+            return nil, ":\n\n"
+        end
+
+        -- End of stream: ai-proxy has assembled the full completion text.
+        local text = ctx.var.llm_response_text
+        if not text or text == "" then
```
nic-6443: This path silently fails *open*, which is surprising for a guard that defaults to fail-closed everywhere else. The subtlety is that `ctx.var.llm_response_text` is published only on a `usage` SSE event (`ai-providers/base.lua` sets it inside `if parsed.usage then`), not on a plain `[DONE]`. ai-proxy injects `stream_options.include_usage=true`, but lots of OpenAI-compatible / self-hosted providers ignore it and stream content followed by `[DONE]` with no usage chunk. In that case `buffer` still holds the real (possibly flagged) completion, but `text` is nil, so `return nil, concat(buffer)` releases it to the client unscanned — and with no log line saying scanning was skipped. The non-streaming branch (line 230) has the same shape; lower impact there since nothing is withheld.

Every test passes because the fixtures always carry a usage chunk, so this gap is invisible in CI. Could this honor `fail_open` (block by default) when there's buffered content but no assembled text, instead of releasing — or at least emit a warn so the skip is observable?

apisix/plugins/ai-lakera-guard.lua
```diff
@@ -206,4 +215,60 @@ function _M.access(conf, ctx)
...
+        local messages = { { role = "assistant", content = text } }
+        return moderate(ctx, conf, messages, "response", conf.response_failure_message)
+    end
+
+    -- Streaming: lua_body_filter is invoked once per upstream chunk. We cannot
+    -- scan a partial completion and we must not let flagged tokens reach the
+    -- client, so we buffer every chunk (withholding it with an empty body) and
+    -- scan the assembled completion once at end-of-stream. This trades
+    -- incremental delivery for true blocking.
+    if ctx.var.request_type == "ai_stream" then
```
nic-6443, line 242: In `alert` (shadow) mode this still buffers the whole stream and withholds every chunk until end-of-stream, then releases it all at once. The docs frame `alert` as a non-intrusive, log-only pass-through, but on streaming routes the client loses token-by-token delivery and instead receives the full body in one shot once scanning finishes. So someone who sets `direction: output, action: alert` precisely to observe Lakera verdicts *without* affecting traffic does change the observable latency/streaming behavior.

Would it make sense to skip the buffering when `action == "alert"` (let chunks flow through live and scan a copy at the end), or at least document the streaming caveat for shadow mode?

apisix/plugins/ai-lakera-guard.lua
```diff
@@ -206,4 +215,60 @@ function _M.access(conf, ctx)
...
+        local buffer = ctx.lakera_response_buffer
+        if not buffer then
+            buffer = {}
+            ctx.lakera_response_buffer = buffer
+        end
+        buffer[#buffer + 1] = body or ""
+
+        if not ctx.var.llm_request_done then
```
nic-6443, line 274: The buffer is only ever released when `llm_request_done` is observed. The note in the docs covers a dropped connection, but ai-proxy's runaway safeguards (`max_stream_duration_ms` / `max_response_bytes`) also set `llm_request_done = true` and then `return` without dispatching another body_filter pass — so the buffered content is stranded and never released. That hits *clean* responses too, not just flagged ones: the client ends up with only the `:\n\n` heartbeats and no `[DONE]`, even though the response was fine and merely tripped a size/duration cap.

Might be worth flushing (and scanning) whatever is buffered on that abort path, or at least widening the doc caveat beyond "dropped connection" to include the gateway-side safeguards.

apisix/plugins/ai-lakera-guard.lua
nic-6443, line 242: Small optimization for shadow mode: when `action: alert`, this branch still buffers the entire stream and withholds every chunk (the `:\n\n` heartbeats) until end-of-stream, then releases it all at once — so shadow mode pays the full latency/TTFT cost even though it never blocks anything. `action` is known up front, so alert mode could skip buffering entirely: let each chunk pass through live (return nothing), and just scan `ctx.var.llm_response_text` and log once at `llm_request_done`. That keeps shadow mode zero-impact on the stream, which is rather the point of running it. The withhold-and-buffer path is only needed for `action: block`.

apisix/plugins/ai-lakera-guard.lua
```diff
@@ -206,4 +215,60 @@ function _M.access(conf, ctx)
...
+        if not text or text == "" then
+            return
+        end
+        local messages = { { role = "assistant", content = text } }
+        return moderate(ctx, conf, messages, "response", conf.response_failure_message)
+    end
+
+    -- Streaming: lua_body_filter is invoked once per upstream chunk. We cannot
+    -- scan a partial completion and we must not let flagged tokens reach the
+    -- client, so we buffer every chunk (withholding it with an empty body) and
```
nic-6443: Two minor things while you're here:

- This comment says the chunk is withheld "with an empty body", but it's actually replaced with a `:\n\n` SSE keep-alive (line 251). Worth fixing the wording to match — and maybe noting *why* a keep-alive rather than `""`: an empty string trips nginx's "nothing to flush", and returning `nil` would let the original chunk leak to the client.
- `local messages = { { role = "assistant", content = text } }` followed by the `moderate(..., "response", conf.response_failure_message)` call is duplicated verbatim between this streaming branch and the non-streaming `ai_chat` branch above. A small `moderate_response(ctx, conf, text)` local would collapse both into one.

apisix/plugins/ai-lakera-guard.lua
janiussyafiq: agreed, alert mode show allow chunks flow. fixed

janiussyafiq: fixed, same as previous comment

apisix/plugins/ai-lakera-guard.lua
janiussyafiq: fixed, now follows fail_open

apisix/plugins/ai-lakera-guard.lua
janiussyafiq: fixed in `apisix/plugins/ai-providers/base.lua`.

apisix/plugins/ai-lakera-guard.lua
janiussyafiq: agreed with the refactoring. fixed

nic-6443 (dismissed):

AlinsRan (dismissed):

membphis (dismissed):

apisix/plugins/ai-lakera-guard.lua
```diff
@@ -206,4 +221,76 @@ function _M.access(conf, ctx)
...
+            return ngx.OK, deny_message(ctx, conf, conf.response_failure_message)
+        end
+
+        local code, message = moderate_response(ctx, conf, text)
+        if code then
+            return ngx.OK, message
+        end
+
+        -- Clean: release the buffered stream verbatim, preserving SSE framing.
+        return nil, concat(buffer)
```
AlinsRan, line 311: **Block-mode streaming can duplicate or drop the response when a protocol converter is active** (e.g. Anthropic client → OpenAI upstream via `ai-proxy-multi`). The same-protocol path tested here is fine.

**Root cause:** release is gated only on `llm_request_done`. In `ai-providers/base.lua` that flag is set *before* the dispatch loop, and with a converter the loop calls `lua_body_filter` **once per converted chunk**. On `[DONE]`, `anthropic-messages-to-openai-chat` emits **two** chunks (`message_delta` + `message_stop`) — both dispatched with `llm_request_done == true`.

**Two symptoms:**
- *Upstream without `include_usage`:* both terminal chunks hit `return nil, concat(buffer)` → **response sent twice + Lakera scanned twice**.
- *Upstream with `include_usage`:* the converter defers the terminal events to the usage chunk (type `usage`, doesn't set `llm_request_done`), then `[DONE]` converts to nothing → **buffer never released, clean response dropped** (client gets only keep-alives).

**Fix:** gate scan+release behind a one-shot flag (e.g. `ctx.lakera_response_released`) so it runs exactly once, and ensure the buffer is flushed at EOF even when the final dispatch produced no chunk (an end-of-stream filter pass in `base.lua`, like the abort path this PR adds). A block-mode `ai-proxy-multi` regression test (Anthropic client + OpenAI streaming upstream) would lock it down.

janiussyafiq: Thanks for the review @AlinsRan. Good catch, fixed, with additional test cases that covers the symptoms you just mentioned.

AlinsRan (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13611, opened 2026-06-26
fix: reset X-Forwarded-For only for untrusted sources under trusted_addresses

shreemaan-abhishek:
### Description

Refines how `handle_x_forwarded_headers()` treats the inbound `X-Forwarded-For` header, based on `apisix.trusted_addresses`.

Previously, when the downstream connection was **not** listed in `apisix.trusted_addresses`, the incoming `X-Forwarded-For` was cleared outright. For a multi-tier proxy chain (`client -> front proxy -> APISIX -> upstream`) where the front proxy is not in `trusted_addresses`, the upstream lost the original client IP and only saw the connection IP rebuilt by `$proxy_add_x_forwarded_for`. That is both a behavior change from older releases and, when `trusted_addresses` is unset, a silent default that breaks upstreams reading the client IP from `X-Forwarded-For`.

This PR replaces the single "clear when untrusted" rule with a three-case model:

1. **`apisix.trusted_addresses` is not configured** — preserve the inbound `X-Forwarded-For` and append the connection IP via `$proxy_add_x_forwarded_for`. This is the backward-compatible default; without a declared trust boundary APISIX makes no trust decision about the chain.
2. **Configured and the source is trusted** — preserve and append, same as above.
3. **Configured and the source is untrusted** — reset `X-Forwarded-For` so the upstream sees only the APISIX-observed connection IP, dropping the spoofable inbound chain.

`X-Forwarded-Proto/Host/Port` are always overwritten with values APISIX observes directly, and the RFC 7239 `Forwarded` header is always cleared for untrusted sources, unchanged by this PR.

### Behavior

For a source sending `X-Forwarded-For: 1.2.3.4`:

| `trusted_addresses` | source | upstream receives |
|---|---|---|
| unset | any | `1.2.3.4, <connection-ip>` |
| configured | trusted | `1.2.3.4, <connection-ip>` |
| configured | untrusted | `<connection-ip>` |

### Checklist

- [x] I have verified that this change is backward compatible (if not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)


membphis (dismissed):

membphis (commented):
Approval withdrawn per follow-up review.

membphis:
Please update the X-Forwarded-For behavior before merge.

The required behavior should be:

1. When `apisix.trusted_addresses` is not configured, keep the current compatible default: preserve the incoming `X-Forwarded-For` and let Nginx append the observed connection IP.
2. When `apisix.trusted_addresses` is configured and the downstream source matches the trusted list, preserve the incoming `X-Forwarded-For` and append the observed connection IP.
3. When `apisix.trusted_addresses` is configured and the downstream source does not match the trusted list, reset `X-Forwarded-For` instead of appending the untrusted incoming chain. In practice, discard the inbound XFF before proxying so the upstream receives only the APISIX-observed connection IP from `$proxy_add_x_forwarded_for`.

Once a trust boundary is configured, forwarding an unmatched client's supplied XFF keeps a spoofable address chain visible to upstream applications. Please also update the tests and config comments to lock in this three-case behavior.

membphis (approved):
Approved from code review. Please update the PR title/description before merge so it matches the final behavior: X-Forwarded-For is preserved and appended only when apisix.trusted_addresses is unset; when trusted_addresses is configured and the peer is untrusted, X-Forwarded-For is reset so upstream only sees the APISIX-observed connection IP.

nic-6443 (approved):

AlinsRan (approved):

PR #13613, opened 2026-06-26
ci: run proxy-cache and dubbo-proxy plugin subdirectory tests

nic-6443:
### Description

The plugin test shards use globs that only match **top-level** `.t` files (`t/plugin/[a-k]*.t`, `t/plugin/[l-z]*.t`). A glob like that doesn't match a *directory* and `*` doesn't cross `/`, so plugin tests living in subdirectories were never handed to `prove` and **never ran in CI**.

This enables two of those subdirectories:

```
t/plugin/proxy-cache/{disk,memory}.t
t/plugin/dubbo-proxy/{route,upstream}.t
```

`build.yml` already starts the Dubbo backend, so dubbo-proxy was prepared to run but the glob never reached it.

**Stale test fix:** `dubbo-proxy/route.t` TEST 1 had rotted while unrun. It asserted that a route referencing a globally-disabled plugin is silently loaded with the plugin skipped (returns `hello world`). Current apisix instead **rejects** such a route — `config_yaml` reports `unknown plugin [dubbo-proxy]` and the route is not created, so the request gets `404`. The test is updated to assert the current behavior.

**Left out for now:** `t/plugin/ext-plugin/` — those tests are self-contained (Lua mock runner) but timing-flaky when run in a batch (the plugin races to connect to the mock runner's unix socket). They need separate test-isolation work before they can run reliably in CI.

### Checklist


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):

membphis:
P2: `graphql-proxy-cache` subdirectory tests are still not included in CI.

This PR fixes the plugin subdirectory coverage gap by explicitly adding `t/plugin/dubbo-proxy` and `t/plugin/proxy-cache`, and it explains why `t/plugin/ext-plugin/` is intentionally left out. However, `t/plugin/graphql-proxy-cache/` also contains runnable `.t` tests (`disk.t`, `graphql.t`, and `memory.t`) and is still not matched by the current matrix:

```yaml
- t/plugin/[a-k]*.t t/plugin/dubbo-proxy
- t/stream-plugin t/plugin/[l-z]*.t t/plugin/proxy-cache
```

That means this PR would still leave a plugin subdirectory test gap undisclosed. Please either add `t/plugin/graphql-proxy-cache` to the appropriate shard, or document why it is intentionally excluded and should be handled separately.

PR #13616, opened 2026-06-26
feat(openid-connect): add set_raw_id_token_header option

### LINKED ISSUE #13615: feat: As a user, I want the openid-connect plugin to expose the raw signed ID token to   upstream for JWKS verification
## Summary

  The `openid-connect` plugin currently exposes the ID token via the `X-ID-Token` header, which
  contains `base64(JSON(decoded_claims))`. This representation carries **no cryptographic
  signature** and cannot be verified by downstream services.

  ## Problem

  Downstream services that need to verify the identity of the caller cannot use `X-ID-Token`
  for this purpose — the token is already decoded and stripped of its signature. The only way
  to cryptographically verify the caller's identity is to have access to the original
  RS256-signed JWT from the identity provider, verifiable against the provider's JWKS endpoint.

  ## Proposed Solution

  enabled, forwards the raw signed ID token JWT to upstream via the `X-Raw-ID-Token` request
  header.
  
  Key design points:
  - The header is cleared at rewrite entry to prevent client-supplied values from bleeding
  through.
  - When `session_contents` is explicitly restricted by the user, the plugin automatically
  ensures `enc_id_token` is included so the raw JWT is persisted in the session by
  lua-resty-openidc.
  - Only available in the full OIDC session flow (not in bearer/introspection path).
  
  ## Use Case

  A backend service sits behind an APISIX route protected by the `openid-connect` plugin. The
  service reads `X-Raw-ID-Token`, verifies the RS256 signature against the identity provider's
  JWKS endpoint, and extracts claims to enforce access control (e.g. restricting access to
  users from a specific domain).

luarx:
### Description
  
  Add a new boolean plugin option `set_raw_id_token_header` (default: `false`) that, when
  enabled, forwards the raw RS256-signed ID token JWT to upstream via the `X-Raw-ID-Token`
  request header.

  Unlike the existing `X-ID-Token` header (which contains `base64(JSON(decoded_claims))` with
  no cryptographic signature), `X-Raw-ID-Token` carries the original signed token verifiable
  against the identity provider's JWKS endpoint.

  Implementation details:
  - The header is cleared at rewrite entry to prevent client injection.
  ensures `enc_id_token` is included so the raw JWT is persisted in the session by
  lua-resty-openidc. When `session_contents` is nil (the default), lua-resty-openidc stores all
  session data including `enc_id_token`, so no action is needed.
  - Only available in the full OIDC session flow (not in the bearer/introspection path, where
  no session exists).

  Fixes #13615

  ### Checklist

  - [x] I have verified that this change is backward compatible (If not, please discuss on the
  [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)


apisix/plugins/openid-connect.lua
```diff
@@ -910,6 +926,12 @@ function _M.rewrite(plugin_conf, ctx)
             if refresh_token and conf.set_refresh_token_header then
                 core.request.set_header(ctx, "X-Refresh-Token", refresh_token)
             end
+
+            -- Add X-Enc-ID-Token header, maybe.
+            local enc_id_token = session:get("enc_id_token")
+            if enc_id_token and conf.set_enc_id_token_header then
+                core.request.set_header(ctx, "X-Enc-ID-Token", enc_id_token)
```
nic-6443: `enc_id_token` is an internal naming within the lua-resty-openidc library, which is unintelligible to gateway users. I suggest naming it `X-Raw-ID-Token`, as it aligns with the description of this field in the schema. What do you think?

nic-6443: change `set_enc_id_token_header`  to `set_raw_id_token_header` too.

luarx: Good point! I've renamed those items as you suggested 👍
Could you check? 🙏

AlinsRan (approved):

shreemaan-abhishek (approved):

nic-6443 (approved):

bzp2010 (commented):
Hold on a moment.

What does `enc` mean? What is it an abbreviation for? `encrypt`?

If so, JWT does not involve encryption. Its payload is always in plaintext; JWT is merely a signature, not encryption. It seems more appropriate to call it `raw_id_token`.

This corresponds to the stripped `id_token` currently provided.

shreemaan-abhishek:
adding on to what @bzp2010 said, it might be worthwile to make the header name configurable.

nic-6443:
@bzp2010 The code already uses `raw_id_token`. The contributor forgot to update the PR title and description.

@shreemaan-abhishek I don't think we need to add configuration options for now. There are already several headers in this plugin without corresponding configuration options. We can add them all together when there is demand.

luarx:
> The code already uses raw_id_token. The contributor forgot to update the PR title and description.

That is correct, thanks for changing it 🙏 

> I don't think we need to add configuration options for now. There are already several headers in this plugin without corresponding configuration options

That's, I basically tried to follow the other header options to maintain consistency.

PR #13623, opened 2026-06-29
fix(ai-aliyun-content-moderation): return nil instead of ngx.OK in lua_body_filter

### LINKED ISSUE #13291: feat: add ai-lakera-guard plugin for prompt injection and PII detection

### LINKED ISSUE #13620: bug: ai-lakera-guard streaming output scanning skipped when chained after a higher-priority response-filter plugin
### Current Behavior

When `ai-lakera-guard` streaming output scanning runs on the same route as a response-filter plugin with **higher priority**, the other plugin can terminate the response body-filter chain before `ai-lakera-guard` runs its end-of-stream pass. Lakera's streaming output decision never executes, and the content it buffered is stranded.

Mechanism:
- `apisix/plugin.lua` `lua_response_filter()` iterates plugins in priority-descending order and, the moment any plugin's `lua_body_filter` returns a code, prints that body and calls `ngx.exit()` — later plugins in the chain do not run for that pass.
- `ai-aliyun-content-moderation` (priority **1029**) runs before `ai-lakera-guard` (priority **1028**). In its default `stream_check_mode=final_packet`, aliyun returns `ngx.OK` on the final packet — even for a clean response — which triggers that `ngx.exit()`.
- As a result, `ai-lakera-guard` never receives the end-of-stream pass that PR #13606 dispatches specifically so buffering plugins can flush/replace their content.

Effect:
- `ai-lakera-guard` output scan is skipped (silent moderation gap).
- In `action=block` it is worse: lakera withholds every chunk (emitting only SSE keep-alive comments) and buffers the real content to release at end-of-stream. Since that final pass never happens, the buffered content is never released — the client receives only keep-alives plus aliyun's `[DONE]`, i.e. an effectively empty response.

### Expected Behavior

A buffering output-filter plugin should still get to make and apply its end-of-stream decision when chained with another response-path plugin. `ai-lakera-guard`'s streaming output scan should run and either release or block its buffered content regardless of other response-filter plugins on the route.

Note on the fix: a priority swap alone is only a partial fix — `ai-lakera-guard` returns a nil code on the clean path (chain continues) but `ngx.OK` on block, while `ai-aliyun-content-moderation` returns `ngx.OK` unconditionally on the final packet. A robust fix should let buffering output filters complete instead of terminating the chain on the first returned code (and/or order lakera's output filter ahead of aliyun's `final_packet` filter). A combined-plugin streaming test should be added.

### Error Logs

No errors are logged — both plugins report success. The failure is silent: lakera's scan never runs and, in block mode, the buffered content is dropped.

### Steps to Reproduce

1. Run APISIX with PR #13606 applied.
2. Create a route with `ai-proxy` to an OpenAI-compatible streaming upstream, plus both plugins (default priorities, no `_meta.priority` override):
   - `ai-lakera-guard`: `direction=output` (or `both`), `action=block`
   - `ai-aliyun-content-moderation`: `check_response=true`, `stream_check_mode=final_packet` (default)
3. Send a streaming chat completion request (`"stream": true`).
4. Observe: lakera's output scan does not run; the client receives only SSE keep-alive comments and `[DONE]`, not the assistant content. Configure Lakera to flag the response and confirm it is neither blocked nor delivered as expected.

### Environment

- APISIX version: `master` with PR #13606 (`ai-lakera-guard` response/output scanning).
- Plugins involved: `ai-lakera-guard` (priority 1028) + `ai-aliyun-content-moderation` (priority 1029).
- Code-path defect independent of OS / OpenResty / etcd versions; reproducible on any environment running that branch.

janiussyafiq:
### Description

`ai-aliyun-content-moderation`'s `lua_body_filter` returned `ngx.OK` on the streaming `final_packet` path. In `apisix/plugin.lua`'s `lua_response_filter`, returning a non-nil code prints the body and `ngx.exit`s the whole response filter chain, so any lower-priority response body filter is skipped.

Because `ai-aliyun-content-moderation` (priority 1029) runs before `ai-lakera-guard` (priority 1028), a route chaining both terminated the chain before `ai-lakera-guard`'s output scan ran — even though `final_packet` only annotates `risk_level` and never blocks. The result was a silent moderation gap: a streamed response that `ai-lakera-guard` would block could leak through unscanned.

This returns `nil` (transform-and-continue) instead of `ngx.OK` on the `final_packet` path, so the chain keeps running lower-priority response body filters. The paths that genuinely terminate the response (`realtime` block) still return a code and are unchanged.

A combined-plugin streaming test (`t/plugin/ai-lakera-guard-chain.t`) is added: it asserts `ai-lakera-guard`'s output scan still runs and blocks a flagged stream when chained after `ai-aliyun-content-moderation`, and that a clean stream still passes through the chain.

Fixes #13620 
Fixes #13291 

### Checklist

### DIFF
--- a/apisix/plugins/ai-aliyun-content-moderation.lua
+++ b/apisix/plugins/ai-aliyun-content-moderation.lua
@@ -464,7 +464,7 @@ function _M.lua_body_filter(conf, ctx, headers, body)
         if not contains_done_event and proto then
             table.insert(raw_events, proto.build_done_event())
         end
-        return ngx_ok, table.concat(raw_events, "\n")
+        return nil, table.concat(raw_events, "\n")
     end
 
     if conf.stream_check_mode == "realtime" then
--- a/apisix/plugins/ai-aliyun-content-moderation.lua
+++ b/apisix/plugins/ai-aliyun-content-moderation.lua
@@ -437,8 +437,11 @@ function _M.lua_body_filter(conf, ctx, headers, body)
         if not ctx.var.llm_response_text then
             return
         end
-        response_content_moderation(ctx, conf, ctx.var.llm_response_text)
-        release_cm_httpc(ctx, conf)
+        if not ctx.ai_aliyun_response_moderated then
+            response_content_moderation(ctx, conf, ctx.var.llm_response_text)
+            release_cm_httpc(ctx, conf)
+            ctx.ai_aliyun_response_moderated = true
+        end
         local events = sse.decode(body)
         for _, event in ipairs(events) do
             if proto and proto.is_data_event(event) then
@@ -461,7 +464,7 @@ function _M.lua_body_filter(conf, ctx, headers, body)
             end
             table.insert(raw_events, sse.encode(event))
         end
-        if not contains_done_event and proto then
+        if not contains_done_event and proto and ctx.var.llm_request_done then
             table.insert(raw_events, proto.build_done_event())
         end
         return nil, table.concat(raw_events, "\n")
--- a/t/lib/fixture_loader.lua
+++ b/t/lib/fixture_loader.lua
@@ -22,6 +22,7 @@
 -- Supported headers:
 --   X-AI-Fixture: <path>          -- fixture file path relative to t/fixtures/
 --   X-AI-Fixture-Status: <code>   -- optional HTTP status code (default 200)
+--   X-AI-Fixture-Flush-Events     -- for .sse fixtures, flush each SSE event separately
 
 local _M = {}
 local io = io
@@ -99,6 +100,17 @@ local function apply_template(content)
 end
 
 
+-- Print each SSE event (delimited by a blank line) as its own flushed chunk,
+-- so the upstream is seen as a multi-chunk stream rather than one body.
+local function flush_sse_events(content)
+    for event in content:gmatch("(.-\n\n)") do
+        ngx.print(event)
+        ngx.flush(true)
+        ngx.sleep(0.01)
+    end
+end
+
+
 -- Serve a fixture based on the X-AI-Fixture request header.
 -- For .sse files: sets Content-Type to text/event-stream and sends content as-is.
 -- For other files: sets Content-Type to application/json.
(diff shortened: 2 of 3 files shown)


nic-6443 (dismissed):

membphis (dismissed):
Code review looks good to me. Please make sure CI is green before merging.

shreemaan-abhishek (dismissed):

membphis (approved):

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13626, opened 2026-06-29
feat(router): add match_uri_encoded_slash to keep %2F in path parameters

### LINKED ISSUE #11810: bug/feat: As a user, I want to use '%2F' in a path parameter
## User Story

As a user, I want to include URL-encoded special characters such as `%2F` in request path parameters.

Example:

- APISIX config:
    - running on localhost;
    - dataplane exposed on port 9080;
    - router `radixtree_uri_with_parameter`;
- API definition: `/api/:path_parameter/foo`;
- request URL: `http://localhost:9080/api/test%2Ftest/foo`;

## Problem

Currently, APISIX with [radixtree_uri_with_parameter](https://apisix.apache.org/docs/apisix/router-radixtree/#5-parameter-match) router **doesn't support** `%2F` in path parameters.

See:

- PR [apache/apisix#11788](https://github.com/apache/apisix/pull/11788)
- Issue [api7/lua-resty-radixtree#148](https://github.com/api7/lua-resty-radixtree/issues/148), 

### MRE (Minimal Reproducible Example)

The following MRE demonstrates the issue. For simplicity we use APISIX in standalone mode.
**NB**: verify that the `lib-resty-radixtree` version is >= **`2.9.2`**. 

File `config.yaml`:

```yaml
apisix:
  router:
    http: radixtree_uri_with_parameter

deployment:
  role: data_plane
  role_data_plane:
    config_provider: yaml
```

File `apisix.yaml`:

```yaml
routes:
  # Test path parameters
  - id: path-params
    uri: /v1/:id/products/electronics/list
    upstream:
      nodes:
        "httpbin.org": 1
      type: roundrobin
    plugins:
      proxy-rewrite:
        uri: /status/200
        method: GET
#END
```

Test the routes:

- path parameter without special characters: 🟢 

    ```bash
    curl -s -i localhost:9080/v1/test/products/electronics/list | grep HTTP
    HTTP/1.1 200 OK
    ```

- path parameter with `%20` (URL-encoded space ` `): 🟢 

    ```bash
    curl -s -i localhost:9080/v1/te%20st/products/electronics/list | grep HTTP
    HTTP/1.1 200 OK
    ```

- path parameter with `%2F` (URL-encoded slash `/`): 🔴 

    ```bash
    curl -s -i localhost:9080/v1/te%2Fst/products/electronics/list | grep HTTP
    HTTP/1.1 404 Not Found
    ```

## Additional Information

> Why is this feature important?

The **current behaviour** should ideally be **wrong**, since the route exists but the request URL is parsed wrongly — contrary to what the `404` status code says. Some frameworks (see the examples below) provide a more expressive implementation, such as allowing further configuration for encoded `/` and returning `400 Bad Request` when it's not allowed.

### Other Technologies/Frameworks

> How do other technologies handle this?

For reference, we made some test with different technologies among the most popular web servers and frameworks and almost every of them support this behaviour.

We collected some examples in a repository along with instructions for the setup and test ([xpicio/routing-with-encoded-slash](https://github.com/xpicio/routing-with-encoded-slash)) and below is a summary of what we have found:

| Technology         | Supports URL-encoded chars | Behaviour              | Docs |
| ------------------ | ---------------------------| ---------------------- | ---- |
| Django (Python)    | 🟢                         | Using `<path:id>`      | [Django - Path Converters](https://docs.djangoproject.com/en/5.1/topics/http/urls/#path-converters) |
| Fastapi (Python)   | 🟢                         | Using `{id:path}`      | [Fastapi - Path Convertor](https://fastapi.tiangolo.com/tutorial/path-params/?h=path+para#path-convertor) |
| Flask (Python)     | 🟢                         | Using `<path:id>`      | [Flask - Variable Rules](https://flask.palletsprojects.com/en/stable/quickstart/#variable-rules) |
| Express (JS)       | 🟢                         | Using `:id` (default)  | - |
| Fastify (JS)       | 🟢                         | Using `:id` (default)  | - |
| Actix Web (Rust)   | 🟢                         | Using `{id}` (default) | - |
| ASP.NET Core (C#)  | 🟢                         | Using `{id}` (default) | - |
| Spring Boot (Java) | 🟢                         | By default requests containing `%2F` in path parameters return `400: Bad Request`, but Tomcat can be configured to allow them | - |

### Our Scenario

> Why is this feature important to us?

Our company has some legacy APIs that expect **serial numbers** containing **`/`** (encoded as `%2F`) in path parameters. Since we maintain serial numbers _virtually forever_, we cannot change them.
To bypass the problem we defined the routes with `*` instead of named path parameters, but that makes our APIs definition **not homogeneous** and introduces the risk of **unexpected collisions**.

AlinsRan:
### Description

Nginx decodes an URL-encoded slash (`%2F`) into a real `/` in `$uri` *before*
route matching happens. As a result a request like
`/v1/te%2Fst/products/electronics/list` is seen by the router as
`/v1/te/st/products/electronics/list`, which has a different number of path
segments and therefore fails to match a route defined with path parameters such
as `/v1/:id/products/:type/list` (returns `404 Route Not Found`). This is the
problem reported in #11810, and it cannot be fixed in `lua-resty-radixtree`
(the parameter pattern `[^/]+` must stop at a real slash to keep segmentation
correct) because the information is already lost by the time the router runs.

This PR adds an opt-in `apisix.match_uri_encoded_slash` option (default `false`).
When enabled, the route matching uri keeps `%2F`/`%2f` encoded so it is treated
as part of a path parameter instead of a separator.

This only affects **route matching**: the request forwarded to the upstream
still uses APISIX's normal (decoded) URI, so upstream behaviour is unchanged.

#### How it stays safe

Rather than re-implementing Nginx's URI normalization in Lua (dot-segment
resolution, slash merging, NUL rejection, …), the option uses the
already-normalized `$uri` as an **oracle**:

- It fully percent-decodes the raw request path and compares it to `$uri`.
- If they are **equal**, Nginx applied nothing beyond decoding (no `../`, no
  merged `//`, no fragment, not an absolute-form request line). Only then does
  it build the matching uri by decoding everything **except** `%2F`. The result
  provably differs from `$uri` only by showing some `/` as `%2F`.
- Otherwise it returns `nil` and matching falls back to the normalized `$uri`.

So traversal (`..%2F..%2F`, `%2e%2e`), consecutive slashes, `%00` and exotic
request lines all fail the equivalence check and fall back — no bypass is
possible even if the decode itself is wrong. The kept slash is normalized to
upper-case `%2F`. Requests without `%2F` in the path skip the check entirely and
keep using `$uri` with no overhead.

Combining `%2F` with dot segments or consecutive slashes therefore falls back to
standard `$uri` matching (such ambiguous requests are handled conservatively,
similar to Envoy's default for escaped-slash normalization).

Fixes #11810

### Checklist

- [x] I have verified that this change is backward compatible (the option defaults to `false`)


shreemaan-abhishek (dismissed):

membphis:
[P1] The encoded-slash match URI currently leaks beyond route matching.

When `match_uri_encoded_slash` is enabled and `build_match_uri_keep_encoded_slash()` succeeds, the implementation writes the encoded-slash-preserved value back into `api_ctx.var.uri` and then rebuilds `api_ctx.var.request_uri` from it. That value is not restored after router dispatch, so route/global plugins in the later rewrite/access phases also observe the encoded match URI.

Why this blocks merge: the PR says this only affects route matching and that upstream behavior is unchanged, but many existing plugins use `ctx.var.uri` for security or rewrite decisions, for example `authz-casbin`, OPA, OAS validation, and `proxy-rewrite`. A request such as `/v1/te%2Fst/products/electronics/list` can therefore make those plugins see `/v1/te%2Fst/...` instead of the normalized `/v1/te/st/...`, which changes the behavior of opt-in deployments beyond routing.

Suggested fix: keep the encoded URI as a router-match-only value, restore the normalized `ctx.var.uri` / `ctx.var.request_uri` immediately after matching, and retain only the matched route/params from the encoded match. Please add a regression test showing that with `match_uri_encoded_slash=true`, a plugin or serverless hook still sees the normalized URI while route params can still capture the encoded slash.

apisix/init.lua
```diff
@@ -487,6 +490,42 @@ local function normalize_uri_like_servlet(uri)
...
+-- slash (%2F/%2f) is left as the literal text "%2F" instead of being turned
+-- into a real path separator -- Nginx decodes it into '/' in $uri, which makes
+-- it indistinguishable from a real separator and breaks path parameter
+-- matching (see issue #11810). The kept slash is always emitted upper-case so
+-- an exact route written with %2F matches regardless of the client's casing.
+local function percent_decode(path, keep_slash)
+    local decoded = re_gsub(path, [[%([0-9a-fA-F][0-9a-fA-F])]], function(m)
+        local hex = m[1]
+        if keep_slash and (hex == "2f" or hex == "2F") then
+            return "%2F"
```
nic-6443, line 502: Since the kept slash is always re-emitted upper-case, an exact route authored with lowercase `%2f` (e.g. `/blog/a%2fb`) can never match — radixtree comparison is byte-exact. The code comment covers the rationale but the user docs don't mention the casing at all; worth one sentence in router-radixtree.md telling users to write `%2F` upper-case in route uris.

apisix/init.lua
```diff
@@ -784,8 +823,37 @@ function _M.http_access_phase()
...
+    -- When match_uri_encoded_slash is on, match the route against a uri that
+    -- keeps the encoded slash (%2F) so it is treated as part of a path
+    -- parameter. This is a router-match-only value: it is swapped in just for
+    -- dispatch and restored right after, so the rewrite/access phases, plugins
+    -- and the upstream keep seeing the normalized ctx.var.uri. Only the matched
+    -- route and its captured params (uri_param_*) retain the encoded slash.
+    local match_uri
+    if local_conf.apisix and local_conf.apisix.match_uri_encoded_slash then
+        local path = api_ctx.var.real_request_uri
+        if path then
```
nic-6443: `$request_uri` is always set for an HTTP request, and `real_request_uri` was just assigned unconditionally a few lines above, so this guard can't be false — dropping it saves one nesting level.

apisix/init.lua
```diff
@@ -784,8 +823,37 @@ function _M.http_access_phase()
...
+            local args_pos = core.string.find(path, "?")
+            if args_pos then
+                path = str_sub(path, 1, args_pos - 1)
+            end
+            if str_find(path, "%2f", 1, true)
```
nic-6443: This is exactly `core.string.find` (it's `string.find(haystack, needle, from or 1, true)`), and this block already uses it for the `?` split above. Using it here too lets you drop the new `str_find` import at the top of the file.

apisix/init.lua
```diff
@@ -784,8 +823,37 @@ function _M.http_access_phase()
...
+               or str_find(path, "%2F", 1, true) then
+                match_uri = build_match_uri_keep_encoded_slash(path, api_ctx.var.uri)
```
nic-6443: One subtlety here: at this point `api_ctx.var.uri` may already have been rewritten by `delete_uri_tail_slash` or `normalize_uri_like_servlet`. E.g. with `delete_uri_tail_slash` on, a request `/a%2Fb/` fully decodes to `/a/b/` while the oracle is the stripped `/a/b`, so the equality check fails and the request silently falls back to normal matching — the feature quietly turns off exactly when another uri option actually changed the uri. The fallback direction is safe, but the interaction deserves either a line in the docs or proper composition (e.g. applying the same tail-slash strip to `match_uri` and comparing against the pre-option uri). Right now there's no test covering these combos either.

t/router/radixtree-uri-with-parameter-encoded-slash.t
```diff
@@ -0,0 +1,206 @@
...
+    }
+--- request
+GET /t
+--- response_body
+passed
+
+
+
+=== TEST 12: the encoded slash is match-only: plugins see the normalized uri while params keep %2F
+--- request
```
nic-6443, line 214: TEST 12 pins the two invariants that matter (plugin-visible uri is normalized, param keeps `%2F`) — nice. One more assertion worth adding: what the upstream actually receives for this request. With the match-only design the wire uri should be the original encoded form (empty `$upstream_uri` makes proxy_pass forward `$request_uri` as-is), and you can observe it deterministically by having the upstream echo its own `$request_uri` (t/lib/server.lua). That would turn the doc claim "the request forwarded to the upstream still uses APISIX's normal URI" into a tested contract instead of a promise — especially valuable since an earlier revision of this PR did leak the rebuilt uri into the forwarding path.

apisix/init.lua
AlinsRan: Added a line to router-radixtree.md (en + zh): the kept slash is normalized to upper-case `%2F` and radixtree compares byte-for-byte, so routes must be authored with upper-case `%2F`. f0a07aa66

apisix/init.lua
AlinsRan: Dropped the guard — `real_request_uri` is assigned unconditionally a few lines above, so it is always set. f0a07aa66

apisix/init.lua
AlinsRan: Switched both checks to `core.string.find` and removed the now-unused `str_find` import. f0a07aa66

apisix/init.lua
AlinsRan: Documented the interaction in router-radixtree.md. I kept the fail-closed fallback rather than composing the tail-slash strip into `match_uri`: the combo is rare and the safe direction (fall back to normal matching) is the one we want, so quietly not keeping `%2F` in that case is acceptable. f0a07aa66

nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13627, opened 2026-06-29
fix(healthcheck): purge stale targets and release periodic lock (bump lua-resty-healthcheck-api7 to 3.2.2)

AlinsRan:
### Description

Fixes a cluster of active-health-check bugs whose root cause is **in the health-check library**, by bumping `lua-resty-healthcheck-api7` from `3.2.1-0` to `3.2.2-0` (which carries api7/lua-resty-healthcheck#55) and adding a regression test.

apisix already marks removed targets via `delayed_clear()` and re-creates checkers on enable; the library was the layer failing to act on that:

- **#13385** — a deleted node was still returned by `/v1/healthcheck`.
- **#13141** — a deleted node was still actively probed.
  Both share one cause: the library advanced a **module-level** cleanup timestamp *inside* its per-checker loop, so only the **first** checker was purged each window. With multiple health-checked upstreams the rest kept their `delayed_clear()`-marked nodes forever. (#55 elects one cleanup worker per window and purges every checker; it also decouples cleanup from the periodic lock so passive-only deployments are cleaned too.)
- **#13235** — active health checks stayed dead after a disable→re-enable cycle, because the periodic-lock holder never released the lock when it went idle, starving the worker that still owned active checkers. (#55 releases the lock once no active checker remains.)

This is a dependency upgrade only — **no apisix production code changes** are required, because the fix belongs in the library.

### Tests

`t/node/healthcheck-stale-cleanup.t` sets up **two** health-checked upstreams (each with two nodes), lets both checkers register their targets, then drops one node and changes the `checks` config on each upstream so the manager rebuilds the checker and `delayed_clear()`s the old target. It asserts, via `/v1/healthcheck`:

- `probed_before: 2` — before removal both soon-to-be-dropped nodes are in the checkers' target lists, i.e. **actively probed** (covers #13141).
- `stale_after: 0` — after the cleanup window the dropped nodes are gone from **every** checker, so they can be neither queried via the control API (#13385) nor probed (#13141).

A node present in a checker's target list *is* being actively probed, so the single `/v1/healthcheck` observation covers #13385 and #13141 together. The bug reproduces only with **multiple** upstreams (a single-upstream setup is always the "first" checker and cleans), which is why earlier single-upstream repros failed.

Verified locally against both library versions:

| lua-resty-healthcheck-api7 | result |
|---|---|
| `3.2.1-0` (before) | **FAIL** — `probed_before: 2` / `stale_after: 1` (the second checker keeps its removed node) |
[...]

Issue #13385, opened 2026-05-18
bug: After a node is deleted from the upstream, its information can still be queried from the health check interface.


wangwei0326:
### Current Behavior

After a node is deleted from the upstream, its information can still be queried from the health check interface.

### Steps to Reproduce

 node 10.205.102.167:8900 before delete:
curl http://127.0.0.1:9180/apisix/admin/upstreams/619149297325179581 -H "X-API-KEY: xxxxxxxxxxxxxxxxxxxx"
{
    "createdIndex": 13033,
    "key": "/apisix/upstreams/619149297325179581",
    "modifiedIndex": 13117,
    "value": {
        "checks": {
            "active": {
                "concurrency": 10,
                "healthy": {
                    "http_statuses": [
                        200,
                        302
                    ],
                    "interval": 1,
                    "successes": 2
                },
                "http_path": "/",
                "timeout": 1,
                "type": "tcp",
                "unhealthy": {
                    "http_failures": 5,
                    "http_statuses": [
                        429,
                        404,
                        500,
                        501,
                        502,
                        503,
                        504,
                        505
                    ],
                    "interval": 1,
                    "tcp_failures": 2,
                    "timeouts": 3
                }
            }
        },
        "create_time": 1778571339,
        "id": "619149297325179581",
        "keepalive_pool": {
            "idle_timeout": 60,
            "requests": 1000,
            "size": 320
        },
        "name": "app-robo-advisor-api",
        "nodes": {
            "10.195.26.32:8600": 1,
            "10.195.26.33:8600": 1,
            "10.205.102.167:8900": 1
        },
        "pass_host": "pass",
        "scheme": "http",
        "timeout": {
            "connect": 6,
            "read": 6,
            "send": 6
        },
        "type": "roundrobin",
        "update_time": 1778580748
    }
} 

curl http://127.0.0.1:9090/v1/healthcheck/upstreams/619149297325179581
{
    "name": "/apisix/upstreams/619149297325179581",
    "nodes": [
        {
            "counter": {
                "http_failure": 0,
                "success": 2,
                "tcp_failure": 0,
                "timeout_failure": 0
            },
            "hostname": "10.205.102.167",
            "ip": "10.205.102.167",
            "port": 8600,
            "purge_time": 1778580075.06,
            "status": "healthy"
        },
        {
            "counter": {
                "http_failure": 0,
                "success": 2,
                "tcp_failure": 0,
                "timeout_failure": 0
            },
[...]

Issue #13141, opened 2026-04-01
bug: health check keeps probing stale upstream nodes after nodes update (checker cache not invalidated)


xinlifoobar:
### Current Behavior

When updating an upstream's nodes (static upstream, no service discovery), APISIX health check continues probing nodes that no longer exist in the upstream configuration.

Specifically:

After removing a node from upstream.nodes,
APISIX still performs active health checks on the removed node
This results in repeated error logs like:
2026/03/31 11:37:15 [error] 60#60: *156956195 [lua] healthcheck.lua:1383: log(): [healthcheck] (upstream#/apisix/upstreams/00000000000000085365) failed to receive status line from '10.xx.xx.xx(10.xx.xx.xx:9358)': timeout, context: ngx.timer

The stale node is not present in the current upstream config, but still exists in the health checker.

### Expected Behavior

When upstream nodes are updated:

Removed nodes should also be removed from the health checker
Health check should only probe current upstream nodes
No stale endpoints should remain in the checker

### Error Logs

[lua] healthcheck.lua:1383: log(): [healthcheck] 
(upstream#/apisix/upstreams/xxxx)
failed to receive status line from '10.xx.xx.xx:9358': timeout, context: ngx.timer

### Steps to Reproduce

1. Run APISIX (e.g., via Docker)

2. Create an upstream with multiple nodes and enable active health check:
```
{
  "nodes": {
    "10.0.0.1:8080": 1,
    "10.0.0.2:8080": 1
  },
  "checks": {
    "active": {
      "type": "http",
      "http_path": "/health",
      "interval": 1
    }
  }
}
```
3. Wait for health check to start
4. Update upstream and remove one node, e.g.:

{
  "nodes": {
    "10.0.0.1:8080": 1
  }
}
Observe:
- APISIX still performs health check on 10.0.0.2:8080
- Error logs continue for the removed node

5. Reload APISIX:

`apisix reload
`
→ Issue disappears

### Environment

- APISIX version (run `apisix version`): 3.14.1-ubuntu
- Operating system (run `uname -a`): Linux core-gateway-sh-hd-volc-apisix-6576dd94db-v5vqk 5.4.0-200-generic #220-Ubuntu SMP Fri Sep 27 13:19:16 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):

nginx version: openresty/1.27.1.2
[...]

PR #13629, opened 2026-06-30
fix(healthcheck): reuse the checker on node-only upstream changes instead of destroy-and-rebuild

### LINKED ISSUE #13282: bug: Health check state lost and checker not working after upstream node changes
### Current Behavior

When upstream nodes change (e.g., Kubernetes pod scaling, service discovery update, or DNS resolution change), the health checker has two critical issues:

1. **Health check status lost**: Previously detected unhealthy nodes reset to healthy after node changes
2. **Health check not running**: There is a probability that the health checker stops checking after node changes

### Root Cause

APISIX uses a **full destroy-and-rebuild** strategy for health checkers when upstream nodes change. The core flow is:

1. Node change → `_nodes_ver` increments → `resource_version` changes
2. `fetch_checker()` detects version mismatch → adds to `waiting_pool`, returns `nil` (no checker during this period)
3. Timer (1s interval) destroys old checker → calls `delayed_clear()` to **clear all health status from shared dict**
4. Creates brand new checker → **all nodes start as healthy**

### Impact

- Traffic routed to unhealthy nodes during the window between checker rebuild and next active check cycle
- Removed nodes remain in the health checker's target list, consuming resources and potentially affecting health check results
- In high-frequency node change scenarios, the checker may never be successfully created due to version race conditions

### Suggested Fix

Implement **incremental target update** instead of full rebuild:

1. When nodes change but `checks` config remains the same, only add/remove targets on the existing checker
2. Use `target.hostname` from `get_target_list()` when calling `remove_target()` to ensure the correct target is matched
3. Only do full rebuild when `checks` configuration changes

Key changes in `healthcheck_manager.lua`:

- Add `update_checker_targets()`: incrementally adds new targets and removes stale ones
- Add `checks_config_equal()`: compares checks config to decide incremental vs full rebuild
- Fix `remove_target()` hostname: use stored `target.hostname` instead of `checks.active.host`
- Save `checks` config in working pool for later comparison

### Expected Behavior

1. Existing nodes should retain their health status (healthy/unhealthy) when nodes are added/removed
2. Removed nodes should be properly cleaned up from the health checker
3. Health checking should not have gaps during node changes

### Error Logs

_No response_

### Steps to Reproduce

1. Create a route with health check enabled and multiple upstream nodes
2. Wait for one node to be detected as unhealthy
3. Add a new node to the upstream (or trigger service discovery update)
4. Observe that the previously unhealthy node resets to healthy
5. Check health checker target list — removed nodes may still be present

### Environment

Environment
APISIX version: 3.16.0
lua-resty-healthcheck-api7: 3.2.1-0

AlinsRan:
### Description

Fixes **#13282**: the health-check manager **destroys and rebuilds** the checker on every upstream change, even when only the upstream nodes changed and the `checks` config is identical. The rebuild opens a window where health filtering is bypassed, and — because a checker's target list lives in a **shm shared by name across all workers** — the per-worker destroy/rebuild also races and corrupts state across NGINX workers.

#### Root cause

`fetch_checker()` keys the working checker by a version derived from **both** `modifiedIndex` and the nodes version. A node-only change bumps that version, so the checker is queued for an asynchronous rebuild. During that window:

1. `fetch_checker()` returned `nil`, so `api_ctx.up_checker` was `nil` and the balancer's health filtering was bypassed — traffic flowed to nodes already known **unhealthy** for ~1-2s.
2. The rebuild discarded the checker's accumulated health state and re-probed every node from scratch.

`timer_working_pool_check` also destroyed the checker on any version mismatch, racing `timer_create_checker` and widening the gap.

#### Fix — incremental reconcile

- `sync_checker_targets()`: when the `checks` config is **unchanged** (`core.table.deep_eq`) and only the nodes changed, reconcile the existing checker's targets in place via `remove_target`/`add_target` against the authoritative shm target list — keeping the checker and its accumulated health state alive instead of rebuilding.
- `timer_working_pool_check` no longer destroys a checker for a node-only version change (it destroys only when the resource is gone or the node count drops to 0).
- When a rebuild is genuinely required (the `checks` config changed), the new checker is created and published **before** the old one is released, eliminating the `up_checker == nil` gap.
- The working-pool entry now stores the `checks` config so node-only vs checks changes can be told apart.

#### Fix — multi-worker correctness

A checker's target membership lives in a shm keyed only by checker **name** (`upstream#<resource_key>`), shared by every worker, while `working_pool`/`waiting_pool` are per-worker and populated **request-driven** — so workers hold asymmetric views and a per-worker lifecycle op mutates shared state. Several fixes close the gaps this opens (surfaced in review):

- **`fetch_checker()` returns the live checker during a version transition** instead of `nil`, so requests keep filtering unhealthy nodes across the ~1s reconcile window (closes root-cause #1 on the request path).
- **`create_checker()` reconciles the shm** — a worker that never held the checker removes stale targets (not just adds), so a node removed after a peer worker populated the shm stops being probed / reported by `/v1/healthcheck`. Removal runs **before** the adds: resty.healthcheck keys a target by `ip+port+hostname` (not the Host header), so a Host-header-only change must free the old identity first — otherwise `add_target` is a no-op on the existing identity and the following `remove_target` wipes the still-desired target.
- **`timer_working_pool_check` clears the shared shm only when no same-name replacement will own it** (resource deleted, or new config has no checks/nodes). When the config still has checks and nodes, a replacement checker built by whichever worker serves traffic owns the shm, so the destroying worker drops only its **local** handle (`stop`) — it must not `delayed_clear`, which would purge the peer's live targets on every worker.

Trade-off: after a `checks` change, a worker that never serves that resource again leaves the (unmarked) old targets in the shm until traffic resumes and `create_checker` reconciles them — they are not probed in the meantime, and this is strictly safer than the eager clear that caused the cross-worker purge.

#### Checker lifecycle after this change

| Resource event | Path | Operation |
|---|---|---|
| **Create** | request → `fetch_checker` enqueues to `waiting_pool` → `timer_create_checker` (no existing checker) | `healthcheck.new()` → remove-stale → `add_target` per node |
| **Update — nodes only** (DNS / k8s scale) | `timer_create_checker` reuse branch | `sync_checker_targets`: `remove_target`/`add_target` on the delta only, **no `delayed_clear`** |
| **Update — `checks` changed** | `timer_create_checker` rebuild branch | `delayed_clear(old)` → `new()` + `add_target(new)` → publish → `stop(old)` |
| **Delete** | `timer_working_pool_check` (`fetch_latest_conf` = `nil`) | `delayed_clear(old)` + `stop` + `working_pool[path]=nil`; periodic cleanup removes the targets |

On the `checks`-changed rebuild, `delayed_clear(old)` runs **before** `create_checker(new)` on purpose: the two checkers share the same shm target list, and `add_target()` only clears a target's `purge_time` when it is re-added *after* being marked. Clearing first lets surviving targets get un-marked on re-add, while genuinely dropped targets keep their `purge_time` and are cleaned up — so a `checks`-only update cannot make the live checker's targets disappear after `DELAYED_CLEAR_TIMEOUT`.

This is **apisix-side only** — it uses `add_target`/`remove_target`, which already exist in the current `lua-resty-healthcheck-api7`. **No dependency bump is required.**

#### Tests

`t/node/healthcheck-incremental-update.t`:
- TEST 1 — a node-only change must **not** destroy/rebuild the checker (`create new checker` once, no `clear checker`).
- TEST 2 — a `checks`-config change **must** still rebuild (`clear checker`).
- TEST 3 — surviving targets are **not** purged after a `checks`-config rebuild.
- TEST 4 — a node-only update keeps filtering an already-unhealthy node during the transition (the request-path nil window).
- TEST 5 — `create_checker` removes targets left stale in the shm by another worker.
- TEST 6 — destroying a stale local checker does not purge a peer worker's live shm targets.
- TEST 7 — recreating the same upstream id within the delayed-clear window keeps its targets.
- TEST 8 — deleting an upstream cleans its shm target list.
- TEST 9 — `create_checker` applies a Host-header-only change instead of wiping the target.

`t/node/healthcheck-multiworker-reconcile.t` (2 workers):
- TEST 1 — the shared shm target set converges to the desired node set under node churn (incremental) and a `checks` rebuild, with no orphan and no purge of survivors.
- TEST 2 — an unhealthy node is detected and filtered consistently across workers (health status, not just membership).

Each fix is verified fail-before / pass-after locally (OpenResty 1.29.2.4).

#### Scope

This PR is **independent of the library**. The related #13385 / #13141 / #13235 are library-side bugs (stale-target cleanup and periodic-lock release) fixed in api7/lua-resty-healthcheck#55 and delivered to apisix by the dependency bump in #13627 — they are **not** part of this PR. (Supersedes the #13282 portion of the now-closed #13582.)

Fixes #13282.

### Checklist

### DIFF
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -198,7 +198,11 @@ local function sync_checker_targets(checker, up_conf)
         end
     end
 
-    -- add targets that are desired but not present, in node order
+    -- add targets that are desired but not present, in node order. Unlike
+    -- create_checker this does not re-add already-present targets, so it cannot
+    -- un-mark a pending purge_time. That is safe here only because sync runs on a
+    -- reuse-eligible checker (checks unchanged), which is never delayed_clear'd --
+    -- delayed_clear only happens on a checks-change rebuild or a destroy.
     for _, target in ipairs(targets) do
         if not current[target.key] then
             local ok, err = checker:add_target(target.host, target.port, target.check_host,
@@ -387,8 +391,10 @@ local function timer_create_checker()
             -- un-marked on re-add, while genuinely dropped targets keep their
             -- purge_time and are cleaned up; clearing after create (the reverse)
             -- would leave the live checker's targets marked and purge them later.
-            -- The old checker is only stopped after the new one is published, so
-            -- fetch_checker never observes a nil/stopped checker (no rebuild gap).
+            -- In this rebuild path the old checker is only stopped after the new
+            -- one is published, so this path never leaves a nil/stopped checker for
+            -- fetch_checker (the checks-change destroy in timer_working_pool_check
+            -- is a separate path that enqueues a rebuild instead).
             if existing_checker then
                 existing_checker.checker:delayed_clear(DELAYED_CLEAR_TIMEOUT)
             end
@@ -436,6 +442,7 @@ local function timer_working_pool_check()
         local res_conf = resource.fetch_latest_conf(resource_path)
         local need_destroy = true
         local has_live_replacement = false
+        local replacement_ver
         if res_conf and res_conf.value then
             local ok, upstream, err
             local plugin_name = get_plugin_name(resource_path)
@@ -504,6 +511,7 @@ local function timer_working_pool_check()
                 -- name. This worker must then NOT clear that shm on destroy.
                 if upstream.checks and upstream.nodes and #upstream.nodes > 0 then
                     has_live_replacement = true
+                    replacement_ver = current_ver
                 end
             end
         end
@@ -513,11 +521,17 @@ local function timer_working_pool_check()
             item.checker.dead = true
             -- Only tear down the shared shm target list when no same-name checker
             -- will own it (resource deleted, or new config has no checks/nodes).
-            -- If the config still has checks and nodes, a replacement checker built
-            -- by whichever worker serves traffic owns the shm; clearing it here
-            -- would purge that live checker's targets on every worker
-            -- (apache/apisix#13282, multi-worker).
-            if not has_live_replacement then
+            -- If the config still has checks and nodes, enqueue a rebuild so the
+            -- replacement checker is built on this worker's timer (not gated on
+            -- traffic); otherwise, if every worker went cold, the shm targets would
+            -- be left neither probed nor purged. The rebuild's create_checker
+            -- reconciles the shm, so we must NOT clear it here -- clearing would
+            -- purge a peer worker's live targets (apache/apisix#13282, multi-worker).
+            if has_live_replacement then
+                if waiting_pool[resource_path] ~= replacement_ver then
+                    waiting_pool[resource_path] = replacement_ver
+                end
+            else
                 item.checker:delayed_clear(DELAYED_CLEAR_TIMEOUT)
             end
             item.checker:stop()
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -182,13 +182,23 @@ function _M.fetch_checker(resource_path, resource_ver)
         return working_item.checker
     end
 
-    if waiting_pool[resource_path] == resource_ver then
-        return nil
+    -- The requested version differs from the working checker -- e.g. a
+    -- discovery/DNS change bumped _nodes_ver. Enqueue the new version so
+    -- timer_create_checker reconciles (or rebuilds) it, but keep returning the
+    -- existing live checker in the meantime: its accumulated health state is
+    -- still valid, so requests during the ~1s transition keep filtering
+    -- unhealthy nodes instead of falling back to "all nodes available", which
+    -- would let a node already known to be unhealthy receive traffic
+    -- (apache/apisix#13282).
+    if waiting_pool[resource_path] ~= resource_ver then
+        core.log.info("adding ", resource_path, " to waiting pool with version: ", resource_ver)
+        waiting_pool[resource_path] = resource_ver
     end
 
-    -- Add to waiting pool with version
-    core.log.info("adding ", resource_path, " to waiting pool with version: ", resource_ver)
-    waiting_pool[resource_path] = resource_ver
+    if working_item and working_item.checker and not working_item.checker.dead then
+        return working_item.checker
+    end
+
     return nil
 end
 
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -46,8 +46,10 @@ _M.get_healthchecker_name = get_healthchecker_name
 
 
 -- Compute the desired set of health-check targets for an upstream config.
--- Returns a map keyed by "host:port:hostheader" so the working set can be
--- diffed cheaply against a checker's current targets.
+-- Returns an ordered array preserving up_conf.nodes order so that targets are
+-- always added to a checker deterministically; each entry also carries a
+-- "host:port:hostheader" key so the working set can be diffed cheaply against
+-- a checker's current targets.
 local function compute_targets(up_conf)
     local host = up_conf.checks and up_conf.checks.active and up_conf.checks.active.host
     local port = up_conf.checks and up_conf.checks.active and up_conf.checks.active.port
@@ -57,14 +59,14 @@ local function compute_targets(up_conf)
     local targets = {}
     for _, node in ipairs(up_conf.nodes) do
         local host_hdr = up_hdr or (use_node_hdr and node.domain) or nil
-        local target = {
+        local target_port = port or node.port
+        targets[#targets + 1] = {
             host = node.host,
-            port = port or node.port,
+            port = target_port,
             check_host = host,
             host_hdr = host_hdr,
+            key = node.host .. ":" .. tostring(target_port) .. ":" .. tostring(host_hdr or ""),
         }
-        local key = target.host .. ":" .. tostring(target.port) .. ":" .. tostring(host_hdr or "")
-        targets[key] = target
     end
     return targets
 end
@@ -97,7 +99,7 @@ local function create_checker(up_conf)
     end
 
     -- Add target nodes
-    for _, target in pairs(compute_targets(up_conf)) do
+    for _, target in ipairs(compute_targets(up_conf)) do
         local ok, err = checker:add_target(target.host, target.port, target.check_host,
                                         true, target.host_hdr)
         if not ok then
@@ -115,7 +117,11 @@ end
 -- so the checker can keep running (and keep its accumulated health state)
 -- instead of being destroyed and rebuilt.
 local function sync_checker_targets(checker, up_conf)
-    local desired = compute_targets(up_conf)
+    -- index the desired targets by key so they can be diffed against current
+    local desired = {}
+    for _, target in ipairs(compute_targets(up_conf)) do
+        desired[target.key] = target
+    end
 
     -- index current targets the same way as desired. Read the authoritative
     -- shm target list (the per-worker checker.targets array can lag behind a
@@ -287,6 +293,7 @@ local function timer_create_checker()
             if existing_checker and existing_checker.checker
                and not existing_checker.checker.dead
                and upstream.checks
+               and upstream.nodes and #upstream.nodes > 0
                and core.table.deep_eq(existing_checker.checks, upstream.checks) then
                 sync_checker_targets(existing_checker.checker, upstream)
                 add_working_pool(resource_path, resource_ver, existing_checker.checker,
@@ -376,11 +383,14 @@ local function timer_working_pool_check()
                             " current version: ", current_ver, " item version: ", item.version)
                 if item.version == current_ver then
                     need_destroy = false
-                elseif upstream.checks and core.table.deep_eq(item.checks, upstream.checks) then
-                    -- Version changed but only because of the upstream nodes; the
-                    -- `checks` config is identical. Keep the checker alive so
-                    -- timer_create_checker can reconcile its targets incrementally
-                    -- (avoids a destroy-and-rebuild nil window for the checker).
+                elseif upstream.checks and upstream.nodes and #upstream.nodes > 0
+                       and core.table.deep_eq(item.checks, upstream.checks) then
+                    -- Version changed but only because of the upstream nodes (and at
+                    -- least one node remains); the `checks` config is identical. Keep
+                    -- the checker alive so timer_create_checker can reconcile its
+                    -- targets incrementally (avoids a destroy-and-rebuild nil window).
+                    -- When the node count drops to 0 we deliberately fall through to
+                    -- destroy the checker, matching the original behaviour.
                     need_destroy = false
                 end
             end
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -322,11 +322,19 @@ local function timer_create_checker()
                 goto continue
             end
 
-            -- The checks config changed (or no checker exists): build a fresh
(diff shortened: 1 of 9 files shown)


membphis (changes requested):
This rebuild order is unsafe when the checks config changes while the target identity stays the same. The new checker is created and published first, so its `add_target()` calls reuse the existing shared shm target list entries for the same `ip:port:hostname`. Then the old checker calls `delayed_clear()`, which marks that same shared target list for purge. After the cleanup window, the targets that the newly published checker still depends on can be removed.

This blocks merge because checks-only updates can make active health-check targets disappear after `DELAYED_CLEAR_TIMEOUT`, especially once the cleanup fix in #13627 makes stale-target purging reliable. Please reorder the swap so `delayed_clear()` happens before the new checker re-adds targets, or otherwise ensure the new checker clears any inherited `purge_time` before it is considered published. It would also be good to add a regression test that waits past the cleanup window after a checks-config rebuild and verifies `/v1/healthcheck` still contains the live targets.

membphis:
This still leaves the health-filter bypass window on node-only updates.

The PR keeps the old checker alive and reconciles targets in `timer_create_checker`, but `fetch_checker()` still returns a checker only when `working_item.version == resource_ver`. When a discovery/DNS change bumps `_nodes_ver`, `upstream.lua` asks for the new version, `fetch_checker()` enqueues that version and returns `nil` until the 1s timer runs. During that window `api_ctx.up_checker` is `nil`; the balancer treats a nil checker as all upstream nodes being available. So a previously unhealthy node can still receive traffic immediately after a node-only update, even though the checker state was not destroyed.

Trigger:
1. Existing checker marks node A unhealthy.
2. Service discovery/DNS adds or removes another node without changing `checks`.
3. The first request after `_nodes_ver` changes calls `fetch_checker()` with the new version.
4. `fetch_checker()` returns `nil` before `timer_create_checker` reconciles and publishes the old checker under the new version.
5. The balancer builds the picker from all upstream nodes, including node A.

This blocks the fix for #13282 because the health state is preserved internally but is not used by requests during the version transition. Please let `fetch_checker()` return the existing live checker for the same resource when the latest config is reuse-eligible (`checks` unchanged and nodes non-empty), while still enqueueing the new version for target reconciliation. A regression test should make one node unhealthy, apply a node-only update, and assert that requests before the timer reconcile do not hit the unhealthy node.

AlinsRan:
Good catch — confirmed. `fetch_checker()` gated on an exact `version` match, so on a node-only `_nodes_ver` bump it returned `nil` (only enqueuing the new version) until the 1s timer ran; `upstream.lua` then set `api_ctx.up_checker = nil` and `balancer.lua` built the picker from all nodes. The health state was preserved internally but not used during the window — exactly as you described.

Fixed in 3688210f3: `fetch_checker()` now returns the existing **live** (non-dead) checker for the resource while it enqueues the new version for reconciliation, so requests keep filtering unhealthy nodes throughout the transition.

One deviation from your suggestion worth flagging: I return the live checker whenever one exists (and isn't dead), rather than gating on reuse-eligibility (`checks` unchanged + nodes non-empty). That guard would need a `fetch_latest_conf()` + `deep_eq` of the checks on the **per-request hot path**. Returning the live checker is safe even in the rarer checks-change window — its health status is real and current for the pre-change checks, which is strictly better than `nil`/no-filtering — and once the timer publishes the rebuilt checker the exact-version branch takes over. Happy to add the eligibility guard if you'd prefer it despite the hot-path cost.

Regression test (`healthcheck-incremental-update.t` TEST 4): mark a node unhealthy via active checks, apply a node-only update, then burst requests in the transition window and assert none hit the unhealthy node. Verified locally — it **fails before** the fix (`errors: 3`, the dead node is picked during the window) and **passes after** (`errors: 0`).

arunmat27:
I hit an issue while testing these changes.

**Setup:** APISIX container (with 2 NGINX worker processes) and 2 upstream nodes.

**Steps:**

1. Start APISIX and make a request. The worker that handles the request creates a new checker and adds both upstream nodes to shared memory.
2. Remove one of the upstream nodes.
3. Send another request and if that is handled by the **other** NGINX worker (i.e., not the worker from step 1) - it won't have the checker in its `working_pool`, it creates a new one. However, during the `create_checker()` flow, it only adds targets and never removes. As a result, the removed upstream node remains in shared memory and continues to appear in the `/v1/healthcheck` response.

**Potential fix:** Update `create_checker()` to call `sync_checker_targets()` instead of `checker:add_targets()`.

AlinsRan:
> I hit an issue while testing these changes.

Good catch, and thanks for the clear repro — confirmed. A worker that never held the checker reaches `create_checker()`, which only *added* targets, so a node removed after another worker populated the shared shm lingered there and kept showing up in `/v1/healthcheck`.

On the suggested `sync_checker_targets()` swap — I went a slightly different route, worth flagging why:

- `sync_checker_targets()` only *adds missing* targets (`if not current[key]`); it never re-adds an already-present one.
- But re-adding is exactly what clears a `purge_time` that a concurrent `delayed_clear()` sets during a checks-config rebuild.
- So swapping to it would bring back the "surviving targets get purged after a rebuild" bug.

Instead `create_checker()` keeps its add-all pass (which un-marks purge_time) and adds a **remove-stale** pass — same net effect as your fix, without regressing the rebuild path.

membphis (dismissed):
LGTM

apisix/healthcheck_manager.lua
```diff
@@ -284,14 +484,56 @@ local function timer_working_pool_check()
                             " current version: ", current_ver, " item version: ", item.version)
                 if item.version == current_ver then
                     need_destroy = false
+                elseif upstream.checks and upstream.nodes and #upstream.nodes > 0
```
nic-6443: There's still a nil window left on a checks-config change: this timer and `timer_create_checker` both run every 1s and their relative order isn't guaranteed. When this timer observes the checks change first, `deep_eq` fails and we fall into the destroy path below (`dead` + `stop` + `working_pool[resource_path] = nil`), only enqueueing a rebuild — so until the next `timer_create_checker` tick, `fetch_checker` returns nil and health filtering is bypassed. That's the same failure mode this PR fixes for node-only changes, just narrowed to checks changes (the PR description's "it destroys only when the resource is gone or the node count drops to 0" doesn't match this branch either).

I think this branch can simply drop the `deep_eq` condition: keep the checker alive whenever `upstream.checks` exists and `#upstream.nodes > 0`, and let `timer_create_checker` decide reuse vs rebuild — its rebuild path already does the atomic publish-new-then-stop-old swap. That closes the window and lets you delete the `has_live_replacement`/`replacement_ver` bookkeeping entirely, since the destroy path would then only run when no replacement can exist (where `delayed_clear` is always correct).

It also fixes a smaller edge on a zero-traffic worker: destroy + enqueue v1 here, then a second checks change to v2 lands before `timer_create_checker` drains → the waiting entry is dropped on the version mismatch, `working_pool` is already empty so nothing re-enqueues, and the shm targets are left neither probed nor purged until traffic resumes. With the checker kept in `working_pool`, this self-heals on the next tick of this timer.

What do you think?

apisix/healthcheck_manager.lua
```diff
@@ -209,22 +349,80 @@ local function timer_create_checker()
...
+            -- sync_checker_targets is the last condition so it only runs when the
+            -- checker is reuse-eligible; if it reports a partial failure the whole
+            -- guard is false and we fall through to a full rebuild below, which
+            -- converges the upstream to the desired targets instead of committing
+            -- the new version against a half-reconciled checker.
             local existing_checker = working_pool[resource_path]
+            if existing_checker and existing_checker.checker
+               and not existing_checker.checker.dead
+               and upstream.checks
+               and upstream.nodes and #upstream.nodes > 0
```
nic-6443: nit: `upstream.nodes and #upstream.nodes > 0` is already guaranteed by the `goto continue` guard added above, so this condition can't be false here.

apisix/healthcheck_manager.lua
```diff
@@ -209,22 +349,80 @@ local function timer_create_checker()
...
             local checker = create_checker(upstream)
             if not checker then
+                -- create_checker failed (upstream healthcheck disabled or
+                -- healthcheck.new errored). The old checker's shm targets were
+                -- already delayed_clear'd above, so it can no longer health-check
+                -- reliably; tear it down and drop it from the working pool instead
+                -- of leaving a stopped/cleared checker that fetch_checker would
+                -- still hand out (it only checks .dead).
+                if existing_checker then
+                    existing_checker.checker.dead = true
```
nic-6443, line 405: nit: worth a log line here — the old code logged "releasing existing checker" on every teardown. When `create_checker` fails (e.g. `disable_upstream_healthcheck` or a `healthcheck.new` error), the checker now silently disappears from the working pool, which will be hard to trace from logs.

apisix/healthcheck_manager.lua
```diff
@@ -209,22 +349,80 @@ local function timer_create_checker()
...
+               and sync_checker_targets(existing_checker.checker, upstream) then
+                add_working_pool(resource_path, resource_ver, existing_checker.checker,
+                                 upstream.checks)
+                core.log.info("reused checker with incremental targets: ",
+                              tostring(existing_checker.checker), " for resource: ",
+                              resource_path, " and version: ", resource_ver)
+                goto continue
+            end
+
+            -- The checks config changed (or no checker exists): rebuild the
```
nic-6443: nit (here and in a few other spots): some of the new comment blocks narrate the issue background / review history rather than constraints the code can't express. The ordering constraints themselves (`delayed_clear` before `create_checker`, remove-before-add for the identity change) are definitely worth keeping — I'd just trim the surrounding narrative so they stay readable.

t/node/healthcheck-leak-bugfix.t
```diff
@@ -17,15 +17,16 @@
...
+log_level('info');
 no_root_location();
 no_shuffle();
 
 run_tests();
 
 __DATA__
 
-=== TEST 1: ensure the old check is cleared after configuration updated
+=== TEST 1: reuse the checker without clearing it when only the nodes change
```
nic-6443: nit: the title says "when only the nodes change", but the test PUTs an identical config — only `modifiedIndex` changes. Maybe "on a version-only change" or "when the config content is unchanged".

apisix/healthcheck_manager.lua
AlinsRan: Fixed in 5c75c7b. timer_working_pool_check no longer destroys on a checks change: whenever the upstream still defines checks and keeps at least one node, it keeps the checker alive and enqueues the rebuild, so timer_create_checker rebuilds it (building the new checker before stopping the old one) and fetch_checker never returns nil. This also made the has_live_replacement gate redundant, so I dropped it.

apisix/healthcheck_manager.lua
AlinsRan: Removed.

apisix/healthcheck_manager.lua
AlinsRan: Added a core.log.warn before dropping the checker from the working pool.

apisix/healthcheck_manager.lua
AlinsRan: Trimmed the narrative on the reuse/rebuild comment blocks; kept the ordering constraints (delayed_clear before create_checker, remove-before-add).

t/node/healthcheck-leak-bugfix.t
AlinsRan: Renamed to "on a version-only change".

membphis (approved):
LGTM

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13632, opened 2026-06-30
feat(ai-cache): add semantic (L2) cache layer

janiussyafiq:
### Description

Adds the optional **semantic (L2)** cache layer to `ai-cache`. On an exact (L1) miss, the request prompt is embedded and matched against a RediSearch cosine vector index; a hit above `similarity_threshold` is served and backfilled into L1. The layer is strictly opt-in via `layers: ["exact", "semantic"]`, so the default exact-only path is unchanged.

- Embedding providers: `openai` and `azure_openai`.
- Per effective-model and per-tenant (`cache_key`) partition isolation of the vector index.
- Fail-open: any embedding or vector-store error degrades to a MISS, never a 5xx.
- Response status reported as `HIT-L1` / `HIT-L2` (with `X-AI-Cache-Similarity`).

Part of #13290 

### Checklist


apisix/plugins/ai-cache/semantic.lua
```diff
@@ -0,0 +1,281 @@
...
+-- from the same model, so this is folded into the index name and key prefix:
+-- changing the embedding model (even to one of the same dimensionality) lands
+-- in a fresh, isolated index instead of being compared against stale vectors
+-- from the previous model.
+local function emb_identity(conf)
+    local emb      = conf.semantic.embedding
+    local provider = emb.openai and "openai" or "azure_openai"
+    local c        = emb[provider]
+    local repr = provider .. "|" .. (c.model or "") .. "|" .. (c.endpoint or "")
+                 .. "|" .. tostring(c.dimensions or "")
```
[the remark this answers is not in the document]
janiussyafiq: fixed

apisix/plugins/ai-cache/semantic.lua
```diff
@@ -0,0 +1,281 @@
...
+    local ctxmsgs  = _M.context_messages(messages, sem.match)
+    local part_ctx = #ctxmsgs > 0 and ctxmsgs or nil
+    ctx.ai_cache_embedding  = vec
+    ctx.ai_cache_dim        = #vec
+    ctx.ai_cache_partition  = key_mod.partition(conf, ctx, body, part_ctx)
+
+    local index = _M.index_name(conf, #vec)
+    local ok
+    ok, err = vs.ensure_index(red, index, l2_prefix(conf, #vec), #vec)
+    if not ok then
```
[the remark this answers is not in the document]
janiussyafiq: fixed

apisix/plugins/ai-cache.lua
```diff
@@ -126,16 +159,35 @@ function _M.access(conf, ctx)
...
         end
         core.log.warn("ai-cache: discarding malformed cache entry for ", ctx.ai_cache_key)
     end
 
+    -- L1 miss: attempt an L2 semantic lookup when the layer is configured.
+    -- `red` is still open and reused inside lookup() for the L1 backfill SET.
+    -- pcall ensures ANY throw (e.g. FFI pack failure on malformed embedding) stays
+    -- fail-open and never leaks the connection or 500s the request.
+    if has_layer(conf, "semantic") and conf.semantic then
+        local ok, hit = pcall(semantic.lookup, red, conf, ctx, body)
```
nic-6443: On an L1 miss `red` is still checked out from the keepalive pool here, but `semantic.lookup` calls `embed()` first — an outbound HTTP request to the embedding provider, bounded by `timeout` (default 5s) — and only touches `red` afterwards for `ensure_index`/`knn_search`. So a pooled Redis connection stays pinned for the whole embedding round-trip; under load that drains `redis_keepalive_pool` and forces a fresh TCP connection to Redis on every miss.

`embed()` doesn't need `red`, so computing the embedding before reusing the connection (or releasing/re-acquiring around it) would avoid holding the pool slot across the network call. What do you think?

apisix/plugins/ai-cache/vector-search/redis.lua
```diff
@@ -0,0 +1,123 @@
...
+    local buf = ffi_new("float[?]", n)
+    for i = 1, n do
+        buf[i - 1] = vec[i]
+    end
+    return ffi_str(buf, n * 4)
+end
+
+
+function _M.ensure_index(red, index, prefix, dim)
+    if ensured[index] then
```
nic-6443: `ensured` is keyed only by the index name, not by the Redis target. If two ai-cache instances both keep `index` at the default `ai-cache` with the same embedding model/dim but different `redis_host`, they collapse to the same key here — so once `FT.CREATE` has run against one host, a request routed to the other skips creation. The read path self-heals (an `FT.SEARCH` error resets `ensured[index]`), but a write landing in that window persists an unindexed HASH until the reset happens.

It does converge, so this is minor, but keying `ensured` by host+port+db+index would make it correct regardless of shared index names. WDYT?

apisix/plugins/ai-cache.lua
janiussyafiq: Agreed, after the fix I ran some QA test and found that peak conns dropped significantly. Good catch!
image

apisix/plugins/ai-cache/vector-search/redis.lua
janiussyafiq: Agreed and fixed.

membphis:
One compatibility concern: this PR changes the existing exact-cache hit header from `X-AI-Cache-Status: HIT` to `X-AI-Cache-Status: HIT-L1`, even when the semantic layer is not enabled.

The `-L1` suffix also feels a bit odd as a user-facing status, and this may break existing clients, tests, dashboards, or log parsing that already check for `HIT`. Could we keep the exact-cache status as `HIT` for backward compatibility, and only use a distinct value such as `HIT-L2` when the semantic layer serves the response? Another option would be to add a separate cache-layer header if we need to expose the serving layer explicitly.

nic-6443 (dismissed):

janiussyafiq:
> One compatibility concern: this PR changes the existing exact-cache hit header from `X-AI-Cache-Status: HIT` to `X-AI-Cache-Status: HIT-L1`, even when the semantic layer is not enabled.
> 
> The `-L1` suffix also feels a bit odd as a user-facing status, and this may break existing clients, tests, dashboards, or log parsing that already check for `HIT`. Could we keep the exact-cache status as `HIT` for backward compatibility, and only use a distinct value such as `HIT-L2` when the semantic layer serves the response? Another option would be to add a separate cache-layer header if we need to expose the serving layer explicitly.

@membphis The initial plan was to keep using `HIT` for both cases (L1 and L2). However, I realized that it is easier for debugging (for testing) and improve readability so we can exactly know which layer was getting hit. One thing to note is that special case for L2, a unique header will appear upon hit, `X-AI-Cache-Similarity` in comparison to L1. To answer your concern about breaking existing clients, tests, etc, since the current ai-cache version hasn't been released hence I think it is safe for the current change. WDYT?

membphis:
Thanks, that makes sense if `ai-cache` has not been released yet.

One extra question: what is the expected user-facing distinction between L1 and L2 here? My understanding is that the two-layer design is mainly an internal optimization/correctness model: L1 is deterministic and cheap, while L2 is approximate and more expensive because it requires embedding plus vector search. Since an L2 hit already has `X-AI-Cache-Similarity`, do we need to expose `HIT-L1` in `X-AI-Cache-Status`, or could we keep the simpler `HIT` for exact hits and use either `HIT-L2` or a separate layer/similarity header only for semantic hits?

janiussyafiq:
> Thanks, that makes sense if `ai-cache` has not been released yet.
> 
> One extra question: what is the expected user-facing distinction between L1 and L2 here? My understanding is that the two-layer design is mainly an internal optimization/correctness model: L1 is deterministic and cheap, while L2 is approximate and more expensive because it requires embedding plus vector search. Since an L2 hit already has `X-AI-Cache-Similarity`, do we need to expose `HIT-L1` in `X-AI-Cache-Status`, or could we keep the simpler `HIT` for exact hits and use either `HIT-L2` or a separate layer/similarity header only for semantic hits?

I believe it is a matter of design, both of your points are legit. I simply reverted back to design choice here in the pic (ref https://github.com/apache/apisix/issues/13290):
image
I guess we can keep `HIT` for both to remain consistent with previous test cases.

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):
Reviewed, no blocking issues found.

PR #13644, opened 2026-07-02
feat(ai-cache): add streaming support with format tagging

janiussyafiq:
### Description

Extends the `ai-cache` plugin (#13632) to cache streaming responses: SSE streams are captured in `body_filter` behind a completeness gate and replayed on cache hits with the correct `Content-Type`, with cache entries tagged by body format (`json`/`sse`) sourced from the ai-providers streaming framing. Also refactors Redis connection management into a shared `with_redis` helper with fail-open handling and improves error handling in the L1 backfill path.

Part of #13290 

### Checklist

### DIFF
--- a/apisix/plugins/ai-cache.lua
+++ b/apisix/plugins/ai-cache.lua
@@ -82,10 +82,8 @@ local function release(conf, red)
 end
 
 
--- Run fn(red) on a pooled connection: released to the keepalive pool on
--- success, closed (discarded) when fn returns an error or throws. Returns
--- fn's result, or (nil, err) on any failure -- acquisition, thrown, or
--- returned -- so callers translate all of them at one place.
+-- Run fn(red) on a pooled connection: released on success, closed when fn
+-- returns an error or throws. Returns fn's result, or (nil, err) on any failure.
 local function with_redis(conf, fn)
     local red, err = redis_util.new(conf)
     if not red then
@@ -109,17 +107,14 @@ local function fail_open(ctx, what, err)
 end
 
 
--- The L1 entry envelope -- the exact layer's stored value. Encoded only here
--- (write-back and backfill) so the shape has one home.
+-- The L1 stored value; encoded only here so the shape has one home.
 local function encode_entry(body, created_at, format)
     return core.json.encode({ body = body, created_at = created_at, format = format })
 end
 
 
--- L2 -> L1 backfill: copy a semantic hit into the exact layer under this
--- request's L1 key, carrying created_at (consistent Age) and format (same
--- replay content-type) whichever layer serves the next hit. Best-effort: the
--- hit is served regardless.
+-- Best-effort L2 -> L1 backfill under this request's L1 key, carrying
+-- created_at and format so either layer replays the hit identically.
 local function backfill_l1(conf, ctx, red, hit)
     local envelope = encode_entry(hit.body, hit.created_at, hit.format)
     if not envelope then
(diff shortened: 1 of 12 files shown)


apisix/plugins/ai-cache/stream.lua
```diff
@@ -0,0 +1,113 @@
...
+        local b = body:byte(content_end)
+        if b ~= 10 and b ~= 13 and b ~= 32 and b ~= 9 then
+            break
+        end
+        content_end = content_end - 1
+    end
+    local lf   = find_last(body, "\n\n", content_end) or 0
+    local crlf = find_last(body, "\r\n\r\n", content_end) or 0
+    local frame_start = (lf > crlf and lf or crlf) + 1
+    local events = sse.decode(body:sub(frame_start))
```
nic-6443: Decoding only the last frame breaks on a spec-legal trailing comment/heartbeat frame after the terminal event. I traced this locally:

```lua
stream_completed(octx, 'data: hi\n\ndata: [DONE]\n\n: keepalive\n\n')  -- false
```

`find_last` picks the `: keepalive` frame as the "terminal" one, `parse_raw_event` skips comment-only frames, so `events` is empty and `capture_format` returns nil — every response from an upstream that pings after `[DONE]` is silently never cached (no log line either, since `log()` just returns).

Also the comment above says full-buffer `sse.decode` is quadratic, but it's a single forward pass over the buffer — O(n). So you could keep the frame-boundary tail check as the truncation guard and then just `sse.decode(body)` and test `is_done_event(events[#events])`: comments are dropped by the decoder, so the last event is the real terminal one, which fixes the case above and deletes the `find_last`/`frame_start` boundary math. WDYT?

apisix/plugins/ai-providers/base.lua
```diff
@@ -549,6 +554,7 @@ function _M.parse_streaming_response(self, ctx, res, target_proto, converter, co
 
         local chunk, err = body_reader()
         if err then
+            ctx.ai_stream_aborted = "read_error"
```
nic-6443, line 561: `ctx.ai_stream_aborted` is never reset, and it survives ai-proxy-multi's retry. If attempt 1's `body_reader()` errors before the first chunk is dispatched (e.g. read timeout waiting for the first token), `retry_on_error` can re-pick the same instance and attempt 2 can stream a complete `[DONE]`-terminated response — but ai-cache's `log()` still bails on the stale flag and skips a legitimately cacheable write. The capture buffer only contains attempt 2's frames at that point (attempt 1 emitted nothing), so the veto isn't protecting anything there.

Resetting the flag at `parse_streaming_response` entry (next to the `ai_stream_framing` stamp) looks safe: a retry after bytes were already flushed can't re-enter the parser anyway — attempt 2's `do_request` dies earlier on the "headers have already been sent" error in `core.response.set_header`.

apisix/plugins/ai-cache.lua
```diff
@@ -222,7 +253,8 @@ end
 
 function _M.body_filter(conf, ctx)
     -- only a MISS gets written back; HIT exited in access, BYPASS opts out.
-    if ctx.ai_cache_status ~= "MISS" or ctx.ai_cache_oversized then
+    if ctx.ai_cache_status ~= "MISS" or ctx.ai_cache_oversized
+       or not stream.capturable(ctx) then
```
nic-6443, line 266: Now that the access-phase `ai_stream` bypass is gone, a non-SSE-framed stream (bedrock's aws-eventstream) pays the L1 GET on every request but can never be written back — `capturable()` blocks buffering here and `capture_format()` blocks the log-phase write, so those keys are a guaranteed 0% hit rate forever. The embedding path is already blocked by the semantic protocol allowlist, so the cost is just one wasted redis round-trip per request.

`ctx.picked_ai_instance` is available in ai-cache's access phase (you already read it there), so an access-time bypass keyed off the provider's `streaming_framing` would restore the cheap-out for these routes. Might not be worth the coupling to the provider module — up to you, just flagging that the old bypass's behavior for bedrock streams quietly changed from "free" to "always pay, never win".

apisix/plugins/ai-cache/stream.lua
janiussyafiq: fixed according to your suggestion

apisix/plugins/ai-providers/base.lua
janiussyafiq: fixed

apisix/plugins/ai-cache.lua
janiussyafiq: fixed

membphis (approved):

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13646, opened 2026-07-02
feat(ai-aliyun-content-moderation): moderate system and tool role content

AlinsRan:
### Description

Extends the request-side moderation of `ai-aliyun-content-moderation` beyond the `user` role, to cover the Agent + tool-calling (MCP) threat model where tool results and a poisoned system prompt can carry indirect prompt injection into the next LLM turn.

New `request_check_roles` option (array, default `["user"]`, **fully backward compatible**):

- **user / tool** follow the existing `request_check_mode`. `last` now walks the trailing consecutive block of selected-role messages, so a fresh user turn or the current round's tool results are moderated without re-checking conversation history.
- **system** ignores `request_check_mode` and is moderated on **every request** (all system messages), because it can be poisoned by malicious ToolCall arguments overwriting the system prompt.

All selected roles' text is concatenated and moderated in a **single** request (not a separate call per role): the Aliyun service takes a flat `content` string with no role field, so separate per-role calls give no benefit.

Protocol layer gains `extract_turn_content(body, mode, roles)` and `extract_system_content(body)` across `openai-chat` / `anthropic-messages` / `openai-responses` / `bedrock-converse` / `openai-embeddings`. When a configured role has no extractor on the current protocol, the request is routed through `binding.on_unsupported(...)` so `fail_mode` decides, instead of silently passing unmoderated.

With the default `["user"]`, `extract_turn_content(body, mode, {user=true})` is equivalent to the previous `extract_user_content(body, mode)`, so existing behavior is unchanged.

Note: tool-result moderation applies to OpenAI-compatible formats where the tool output is a distinct `tool` role/item; Anthropic/Bedrock nest tool results inside user messages and are not extracted (documented in the option table).

### Checklist

- [x] I have added tests corresponding to this change (unit tests for the extractors + end-to-end role moderation, `t/plugin/ai-aliyun-content-moderation.t` TEST 57–72)
- [x] I have updated the documentation (en/zh `ai-aliyun-content-moderation.md`)
- [x] I have verified backward compatibility (default `["user"]` preserves prior behavior)


shreemaan-abhishek (dismissed):

membphis (dismissed):
Reviewed the implementation, documentation, and tests. The default behavior remains backward compatible, and I did not find merge-blocking issues.

nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13647, opened 2026-07-02
refactor(ai-aws-content-moderation): moderate decoded LLM content in access phase

shreemaan-abhishek:
### Description

Refactor `ai-aws-content-moderation` to be structured like its sibling `ai-aliyun-content-moderation`.

Previously the plugin ran in the `rewrite` phase (before `ai-proxy`) and sent the **raw** HTTP request body to AWS Comprehend. As a result Comprehend scored the undecoded JSON envelope (e.g. the literal escaped string `"content":"toxic"` and `{"model":...,"messages":[...]}`), while the upstream LLM acts on the decoded prompt, so the two saw different text.

#### What changed

- Move the plugin from the `rewrite` phase to the `access` phase (priority `1050` -> `1031`, below `ai-proxy`'s `1040`), so it runs after `ai-proxy` and can reuse `ctx.ai_client_protocol` and `ctx.picked_ai_instance`.
- Make it protocol-aware: parse the JSON body and extract the LLM-visible prompt content via the detected protocol (`ai-protocols`), sending only the normalized, decoded content to Comprehend. This is what the sibling `ai-aliyun-content-moderation` plugin does.
- Guard on `ctx.picked_ai_instance` via `fail_mode`, so the plugin reports clearly when it is used without `ai-proxy`/`ai-proxy-multi`.
- Return a provider-compatible deny response (via `ai-protocols`) instead of a raw text body, so AI clients are not broken. Adds `check_request`, `deny_code` (default `200`) and `deny_message` options.

The AWS Comprehend decision model (`moderation_categories` per-category thresholds and the overall `moderation_threshold`) is unchanged.

#### Behavior change

The plugin now must be used together with `ai-proxy`/`ai-proxy-multi`, consistent with `ai-aliyun-content-moderation` and with how the plugin is documented. A request that does not pass through `ai-proxy` is governed by `fail_mode` (default `skip`, i.e. passes through unchecked; set `fail_mode: error` to reject). This supersedes #13528.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/ai-aws-content-moderation.lua
+++ b/apisix/plugins/ai-aws-content-moderation.lua
@@ -16,17 +16,20 @@
 --
 require("resty.aws.config") -- to read env vars before initing aws module
 
-local core = require("apisix.core")
-local binding = require("apisix.plugins.ai-protocols.binding")
-local aws = require("resty.aws")
+local core      = require("apisix.core")
+local protocols = require("apisix.plugins.ai-protocols")
+local binding   = require("apisix.plugins.ai-protocols.binding")
+local aws       = require("resty.aws")
 local aws_instance
 
 local http = require("resty.http")
 
-local pairs = pairs
-local unpack = unpack
-local type = type
-local ipairs = ipairs
+local ngx     = ngx
+local pairs   = pairs
+local unpack  = unpack
+local type    = type
+local ipairs  = ipairs
+local table   = table
 local HTTP_INTERNAL_SERVER_ERROR = ngx.HTTP_INTERNAL_SERVER_ERROR
 local HTTP_BAD_REQUEST = ngx.HTTP_BAD_REQUEST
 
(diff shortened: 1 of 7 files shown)


membphis (changes requested):
[P1] Legacy configurations without ai-proxy now fail open

Before this PR, ai-aws-content-moderation ran in the rewrite phase and moderated the raw request body for routes that used only this plugin. After the refactor, access() returns through binding.on_unsupported when ctx.picked_ai_instance is absent; with the existing default fail_mode=skip, the request is passed through unchecked. The new test also asserts that a toxic request without ai-proxy returns 200.

This is a content-moderation policy regression for existing deployments that already use this plugin without ai-proxy or ai-proxy-multi. Please keep a legacy raw-body moderation fallback, or make this migration fail closed / explicit opt-in with a documented compatibility path.

membphis (approved):
Approving this PR. For the legacy configurations without ai-proxy case, I recommend handling the compatibility and fail-open behavior in a separate follow-up PR, with an explicit migration path for standalone ai-aws-content-moderation usage.

nic-6443 (approved):

AlinsRan (approved):

PR #13648, opened 2026-07-02
perf(loggly): hoist per-request closure out of log phase

shreemaan-abhishek:
### Description

The `loggly` logger reassigned the `handle_http_payload` closure on **every request** inside `_M.log()`, solely to capture `conf`. Each request therefore allocated a fresh function object, adding avoidable GC pressure in the log phase. Benchmarking loggers showed `loggly` carrying a disproportionately high per-request overhead compared to peers in the same class, and this per-request allocation was the cause (P50 is unaffected since the actual send happens in the async batch-processor timer).

This PR hoists the handler to module scope: `conf` is threaded into `handle_log(entries, conf)` directly, and the small binding closure is created **once per batch processor** (i.e. once per config version) rather than once per request. The separate `handle_http_payload` indirection is removed.

As a side effect this also removes a latent correctness bug: the shared module-level `handle_http_payload` was overwritten by whichever request ran `_M.log()` last, so with two routes using different loggly configs the async handler could send a batch with the wrong `conf`. Binding `conf` per processor fixes that.

Behavior is otherwise unchanged; existing `t/plugin/loggly.t` cases (including TEST 15, the http bulk path that asserts `conf.tags`) continue to cover both the syslog and http paths.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/loggly.lua
+++ b/apisix/plugins/loggly.lua
@@ -309,9 +309,7 @@ local function send_bulk_over_http(message, metadata, conf)
 end
 
 
-local handle_http_payload
-
-local function handle_log(entries)
+local function handle_log(entries, conf)
     local metadata = plugin.plugin_metadata(plugin_name)
     core.log.info("metadata: ", core.json.delay_encode(metadata))
 
@@ -330,11 +328,12 @@ local function handle_log(entries)
                 return false, err, i
             end
         end
-    else
-        return handle_http_payload(entries, metadata)
+        return true
     end
 
-    return true
+    -- loggly bulk endpoint expects entries concatenated in newline("\n")
+    local message = tab_concat(entries, "\n")
+    return send_bulk_over_http(message, metadata, conf)
 end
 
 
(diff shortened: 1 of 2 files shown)


membphis (changes requested):
[P2] Missing regression coverage for the cross-config wrong-conf bug

The code change looks right, but the PR claims to fix a bug where the module-level handle_http_payload could be overwritten by a later request from a different loggly configuration. The patch does not add a test that would fail on the old behavior: two routes/configs with different customer_token/tags, delayed batch flush, and interleaved requests.

Please add a Test::Nginx regression case that creates two distinct HTTP bulk loggly configs, sends interleaved requests before the first batch flushes, and asserts each batch is sent with its own token/tags. Without that coverage, this correctness fix can regress unnoticed.

shreemaan-abhishek:
Added a regression test (`TEST 22` in `t/plugin/loggly.t`) covering exactly this case:

- Two routes with distinct HTTP-bulk loggly configs (`customer_token` `token-a`/`token-b`, tags `aaa`/`bbb`).
- Both send one entry each, which buffer before either batch flushes (`inactive_timeout: 1`, default batch size), so the two batch processors flush after both `log()` calls have run.
- The mock bulk endpoint captures the token from the request path and the `X-LOGGLY-TAG` header, and the test asserts each batch arrives with its own token/tags.

This fails on the previous behavior: the shared module-level `handle_http_payload` was left holding whichever request ran `log()` last, so both batches were sent with `token-b` and the `token-a` assertion never matches. With this PR each batch processor binds its own `conf`, so both assertions pass.

membphis (approved):
The added regression coverage now addresses the previous cross-config concern. No merge-blocking issue remains from my side.

nic-6443 (approved):

AlinsRan (approved):

PR #13652, opened 2026-07-03
fix(ai-cache): bypass caching for prompts carrying non-text content

AlinsRan:
### Description

https://github.com/apache/apisix/actions/runs/28582602012/job/84746279063

Since #13634 the protocol layer's `get_messages()` flattens structured message content to plain text, dropping image / audio / other non-text typed parts. `ai-cache` consumed that flattened view, so a `text + image` prompt became indistinguishable from a text-only one:

- **L1 (exact):** the fingerprint is built from the flattened messages, so `text+image` collapses to the same key as the text-only prompt → cross-modal L1 collision.
- **L2 (semantic):** `window_has_nontext()` inspected the flattened messages (`content` is already a plain string), so it could never detect the image → the multimodal request was embedded and matched the text-only vector.

Either way a multimodal request was wrongly served a text-only cached response (`X-AI-Cache-Status: HIT`).

`#13634` did not touch `ai-cache`, so this regressed silently on master. `t/plugin/ai-cache-semantic.t` TEST 64 fails: `got 'HIT', expected 'MISS'`.

### Fix

Detect non-text content from the **raw** request body (before `get_messages()` flattens it) and bypass the cache entirely for such a request: read nothing, write nothing (fingerprint left unset so `log()` skips the write-back), and report a `MISS` so it always proceeds to the upstream. This restores the multimodal-bypass invariant guarded by TEST 62-64. The semantic layer's own check is switched to the same raw-body detector as a second line of defence.

A prompt that carries non-text content cannot be faithfully keyed while the canonical form is text-only, so not caching it is the safe behaviour.

### Verification

- `t/plugin/ai-cache-semantic.t`: **212/212 pass** (was 1 failing — TEST 64).
- Reverting the fix reproduces exactly `got 'HIT', expected 'MISS'`.
- `luacheck` clean on both changed files.

### Checklist

- [x] I have added tests corresponding to this change (existing TEST 62-64 now pass)
- [x] I have updated the documentation to reflect this change (N/A — behaviour fix)

### DIFF
--- a/apisix/plugins/ai-cache.lua
+++ b/apisix/plugins/ai-cache.lua
@@ -135,6 +135,17 @@ function _M.access(conf, ctx)
         return
     end
 
+    -- A prompt carrying non-text content (images, audio, other typed parts)
+    -- cannot be faithfully keyed: the protocol get_messages() flattens content
+    -- to plain text, so text+image collides with a text-only entry at L1 and
+    -- across different images at L2. Never serve or store such a request -- leave
+    -- the fingerprint unset so log() writes nothing back -- and report a MISS so
+    -- it proceeds to the upstream uncached.
+    if semantic.body_has_nontext(body) then
+        ctx.ai_cache_status = "MISS"
+        return
+    end
+
     ctx.ai_cache_fingerprint = key_mod.fingerprint(ctx, body)
     ctx.ai_cache_key = key_mod.build(conf, ctx, ctx.ai_cache_fingerprint)
     -- Remember which instance the fingerprint was computed for. ai-proxy-multi
--- a/apisix/plugins/ai-cache/semantic.lua
+++ b/apisix/plugins/ai-cache/semantic.lua
@@ -127,14 +127,31 @@ function _M.context_messages(messages, match)
 end
 
 
+-- A content block is text-only iff it carries a string `text` and nothing that
+-- marks it as another modality. Two block shapes exist across supported client
+-- protocols, so the test must cover both:
+--   * OpenAI Chat tags every part with `type` ({type="text",text=...} vs
+--     {type="image_url",...}), so a `type` other than "text" is non-text.
+--   * Bedrock Converse uses a distinct key per modality with no `type`
+--     ({text=...} vs {image=...}/{document=...}/{toolUse=...}), so a block
+--     without a string `text` is non-text there.
+-- Anything that is not a plain text block (including a non-table element) is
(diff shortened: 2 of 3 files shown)


nic-6443 (dismissed):

membphis:
I do not think this is ready to merge yet.

The OpenAI Chat regression fix is headed in the right direction, but the new raw-body detector still misses non-text content for an already supported protocol: Bedrock Converse.

`body_has_nontext()` currently only returns true when a content block has a `type` field and that value is not `text`:

https://github.com/apache/apisix/blob/0339f43153636823e146db5749ca85195a8d8f14/apisix/plugins/ai-cache/semantic.lua#L138-L148

However, Bedrock Converse content blocks are flattened by reading only `block.text`, and non-text blocks such as image/document blocks do not need to use this OpenAI-style `type` field:

https://github.com/apache/apisix/blob/0339f43153636823e146db5749ca85195a8d8f14/apisix/plugins/ai-protocols/bedrock-converse.lua#L152-L160
https://github.com/apache/apisix/blob/0339f43153636823e146db5749ca85195a8d8f14/apisix/plugins/ai-protocols/bedrock-converse.lua#L232-L258

At the same time, the L1 fingerprint excludes raw `messages` from `params`, so a Bedrock text-only request and a text+image/text+document request can still canonicalize to the same cache key:

https://github.com/apache/apisix/blob/0339f43153636823e146db5749ca85195a8d8f14/apisix/plugins/ai-cache/key.lua#L63-L67

Suggested fix: make the raw-body non-text detector protocol-aware, or at least treat Bedrock content blocks that are not pure `{ text = ... }` blocks as non-text, then add a Bedrock exact-cache regression test for text-only versus same text plus image/document.

AlinsRan:
Good catch. Bedrock Converse blocks use `{image=...}`/`{document=...}` with no `type`, so the OpenAI-only `type ~= "text"` check missed them and a Bedrock text+image request still collided with text-only at L1.

Reworked the detector to be shape-agnostic: a block counts as text only if it has a string `text` and no non-text discriminator, covering both the OpenAI `type` tag and Bedrock's key-per-modality blocks. Added a Bedrock exact-cache regression (TEST 66-68) — text-only warms L1, same text + image is a MISS; reverting the detector reproduces the HIT, so the test bites.

The access-phase guard bypasses before the fingerprint is computed, so this raw-body detection also covers the L1 canonicalization gap you pointed at — no key.lua change needed.

PR #13653, opened 2026-07-03
perf(datadog): reuse one UDP socket per batch and coalesce metrics

shreemaan-abhishek:
### Description

The `datadog` plugin opened a fresh UDP socket for **every entry** in a flushed batch and sent **6 separate datagrams per entry** (`request.counter`, `request.latency`, `upstream.latency`, `apisix.latency`, `ingress.size`, `egress.size`). For a batch of N entries this cost N socket setups and 6N `sendto` syscalls in the batch-processor flush.

This PR:
- reuses a **single** UDP socket across the whole batch instead of one per entry, and
- coalesces each entry's metrics into **one newline-delimited DogStatsD datagram**.

Per-batch work drops from `N sockets + 6N sends` to `1 socket + N sends`. Newline-delimited multi-metric packets are standard DogStatsD and are parsed by the Datadog agent, so this is wire-compatible with existing receivers. The plugin schema and the metrics emitted are unchanged.

The test mock (`t/lib/mock_layer4.lua` `dogstatsd()`) now splits received datagrams on newlines, as a real DogStatsD server does, so the existing per-metric assertions in `t/plugin/datadog.t` continue to hold without changes.

### Checklist

- [x] I have added tests corresponding to this change (existing `t/plugin/datadog.t` assertions cover the emitted metrics; mock updated to reflect coalesced framing)
- [x] I have updated the documentation accordingly (no user-facing behavior/config change; docs already describe batch behavior)
- [x] I have verified linting passes

### DIFF
--- a/apisix/plugins/datadog.lua
+++ b/apisix/plugins/datadog.lua
@@ -142,17 +142,8 @@ local function generate_tag(entry, const_tags)
 end
 
 
-local function send_metric_over_udp(entry, metadata)
-    local err_msg
-    local sock = udp()
-    local host, port = metadata.value.host, metadata.value.port
-
-    local ok, err = sock:setpeername(host, port)
-    if not ok then
-        return false, "failed to connect to UDP server: host[" .. host
-                      .. "] port[" .. tostring(port) .. "] err: " .. err
-    end
-
+-- Build all DogStatsD metrics for one entry as a single newline-delimited packet.
+local function build_metrics(entry, metadata)
     -- Generate prefix & suffix according dogstatsd udp data format.
     local suffix = generate_tag(entry, metadata.value.constant_tags)
     local prefix = metadata.value.namespace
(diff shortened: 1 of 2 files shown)


membphis (approved):
[P2] Coalesced DogStatsD datagram has no size guard

This PR changes the Datadog plugin from sending each metric in its own UDP datagram to sending all metrics for one entry as a newline-delimited DogStatsD datagram. DogStatsD supports newline-delimited packets, but the implementation does not enforce any packet size limit or split oversized payloads.

APISIX allows long tag sources: `constant_tags` has no maxItems limit, route/service/consumer names can be long, and `include_path` can add a long route path. With those inputs, a single metric line may still be acceptable while the coalesced 5/6-line datagram exceeds the DogStatsD Agent buffer or practical UDP packet size limits, causing silent metric loss.

Please add packet-size-aware splitting or fall back to smaller datagrams when the coalesced payload is too large, and add a test covering long tags/path.

nic-6443 (approved):

AlinsRan (approved):

PR #13656, opened 2026-07-03
perf(zipkin): skip span tag construction for unsampled requests

shreemaan-abhishek:
### Description

The zipkin plugin builds the request span with its full tags table (`http.method`, `http.url`, and the `get_remote_client_ip`/`get_remote_client_port` lookups) **before** deciding whether the request is sampled. But `apisix/plugins/zipkin/reporter.lua` drops any span whose `x-b3-sampled` baggage is `0`, so that per-request work is pure waste on unsampled requests. On high-throughput routes with a low `sample_ratio` this overhead is paid on every request.

This PR:

- Moves the sampling decision ahead of span construction. On unsampled requests the tags table and the two remote-addr lookups are skipped (`tags = nil`, which `opentracing` `start_span` handles cleanly). The bare span is still created because header propagation (`inject_header`) needs its trace context, and `set_ngx_var` output is unchanged.
- Guards the span-version-1 child spans in the `access` phase behind `ctx.opentracing_sample` so unsampled requests skip those allocations too (also removes a latent nil-timestamp path).
- Documents sampling as the primary performance lever for the zipkin and skywalking plugins.

No behavior change for sampled requests: identical tags, span order, and `x-b3-sampled` propagation.

### Checklist

- [x] I have added tests corresponding to this change (`t/plugin/zipkin3.t` TEST 6 asserts an unsampled request still builds span context for `set_ngx_var`; sampled propagation covered by existing cases)

### DIFF
--- a/apisix/plugins/zipkin.lua
+++ b/apisix/plugins/zipkin.lua
@@ -182,19 +182,34 @@ function _M.rewrite(plugin_conf, ctx)
 
     local wire_context = tracer:extract("http_headers", ctx)
 
-    local start_timestamp = ngx.req.start_time()
-    local request_span = tracer:start_span("apisix.request", {
-        child_of = wire_context,
-        start_timestamp = start_timestamp,
+    -- Decide sampling before building the span. The reporter drops unsampled
+    -- spans, so skip tag construction and remote-addr lookups for them.
+    if sampled == "1" or sampled == "true" then
+        per_req_sample_ratio = 1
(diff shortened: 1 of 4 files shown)


membphis (approved):
The sampling change keeps trace context propagation intact while avoiding the unnecessary work on unsampled requests. No merge-blocking issue found from my side.

nic-6443 (approved):

AlinsRan (approved):

PR #13663, opened 2026-07-06
fix(resource): support all plugin-bearing resource types in fetch_latest_conf

### LINKED ISSUE #13456: bug: ai-proxy-multi plugin fails with plugin_config_id when multiple instances are configured

AlinsRan:
### Description

Follow-up to #13458 (and issue #13456).

`resource.fetch_latest_conf` only whitelisted `upstreams`/`routes`/`services`/`stream_routes`. Any resource that attaches plugins via `_meta.parent` (set by `plugin.set_plugins_meta_parent`) but isn't in that list — `plugin_configs`, `global_rules`, `consumers`, `consumer_groups` — hit `unsupported resource type` and returned `nil`.

Concretely, `ai-proxy-multi` with multiple instances referenced through a `plugin_config_id` calls `fetch_latest_conf(conf._meta.parent.resource_key)` in `pick_target`, gets `nil`, and the request fails with `failed to fetch the parent config` (5xx). The same 5xx happens for ai-proxy-multi carried by a global_rule / consumer / consumer_group.

#13458 fixed only `plugin_configs` by adding one more branch. As @nic-6443 pointed out there, every supported prefix just maps to itself, so this replaces the if/else chain with a whitelist lookup table covering all plugin-bearing resource types; each maps to its `/<type>` top-level config key.

### Fixes

closes #13456.

### Test

Added a regression test in `t/plugin/ai-proxy-multi.balancer.t`: an `ai-proxy-multi` plugin_config with two instances, referenced by a route via `plugin_config_id`. Without the fix the request returns 5xx (`failed to fetch the parent config`); with it, it proxies successfully. Verified locally — the whole file passes, and reverting `resource.lua` makes only the new test fail.

### DIFF
--- a/apisix/resource.lua
+++ b/apisix/resource.lua
@@ -22,6 +22,23 @@ local config_util = require("apisix.core.config_util")
 local require = require
 
 local _M = {}
+
+
+-- resource types whose config may attach plugins via `_meta.parent`
+-- (set by plugin.set_plugins_meta_parent) and thus reach fetch_latest_conf.
+-- Every type maps to its own "/<type>" top-level config key.
+local SUPPORTED_RESOURCE_TYPES = {
+    upstreams = true,
+    routes = true,
+    services = true,
+    stream_routes = true,
+    plugin_configs = true,
+    global_rules = true,
+    consumers = true,
+    consumer_groups = true,
+}
+
+
 local function remove_etcd_prefix(key)
     local prefix = ""
     local local_conf =  require("apisix.core.config_local").local_conf()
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13666, opened 2026-07-07
fix(balancer): keep least_conn load state across upstream scaling

### LINKED ISSUE #12217: help request: WebSocket Load Balancing Imbalance Issue After Upstream Node Scaling
### Description

Issue Description
When using APISIX to proxy WebSocket requests, we've observed that when upstream nodes are scaled out, the load distribution of WebSocket connections becomes unbalanced.

Steps to Reproduce
Configure APISIX to proxy WebSocket requests to backend services
Start with 2 upstream nodes providing service
Establish a large number of WebSocket long connections
Scale out the upstream (e.g., from 2 nodes to 3 or more)
Observe the connection distribution across nodes
Current Behavior
After scaling, new connections are evenly distributed across all nodes, but previously established WebSocket long connections remain concentrated on the original two nodes, resulting in an unbalanced load distribution.

Expected Behavior
After scaling, the system should take into account existing long connections, resulting in a more balanced load distribution across all nodes (including newly added ones).

Root Cause Analysis
Based on observation, the issue appears to be caused by:

APISIX maintains a counter mechanism for load balancing. For example, when there are 2 nodes, each node's counter is initialized to 10000. When downstream nodes are scaled out, APISIX resets all counters, but the previously established WebSocket long connections are not recorded in the new count, causing inaccurate load calculations.

Specifically:

A large number of WebSocket long connections have already been established on the original two nodes
After scaling, counters are reset, and these existing connections are "forgotten" in load balancing decisions
New connections will be evenly distributed, but when combined with existing connections, the overall load distribution is unbalanced
Environment Information
APISIX version: latest
Operating system: liunx
Deployment method: Kubernetes
Additional Information
This issue is particularly noticeable in high-concurrency WebSocket application scenarios, especially when long connections persist for extended periods. We hope the load balancing algorithm can be improved to consider existing long connections when nodes are scaled out.

### Environment

- APISIX version (run `apisix version`): latest
- Operating system (run `uname -a`): liunx
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): latest
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): NA
- APISIX Dashboard version, if relevant: NA
- Plugin runner version, for issues related to plugin runners: NA
- LuaRocks version, for installation issues (run `luarocks --version`): NA

AlinsRan:
Fixes #12217

## The problem

Scale a `least_conn` upstream out and the new node is not preferred — the load stays on the original nodes. With WebSocket it never recovers, because those connections outlive the scale out.

Reproduced by `t/node/least_conn3.t` TEST 9 (real WebSocket sessions through APISIX). Two nodes, four live sessions, add a third node, open two more:

```
sessions before the scale out    0.0.0.0    127.0.0.1  127.0.0.1  0.0.0.0     <- two each
the two sessions after it        0.0.0.0    127.0.0.2
                                 ^^^^^^^                ^^^^^^^^^
                    already carrying two sessions       the new, empty node
```

Round-robin. The empty node gets half, an already-loaded node gets the other half.

## Why

The per-server score heap lives **inside the picker**, and the picker is cached by upstream version. Scaling bumps the version, the heap is rebuilt, every score resets to `1/weight` — the four live sessions are forgotten.

Seeding the new heap from a side count doesn't fix it either. In-flight requests keep the picker they were routed with and release their server on **that** picker in the log phase, so those releases land on a heap nobody reads. A heap that is only correct at rebuild time drifts: once the old sessions drain, the surviving nodes stay penalized for load they no longer carry — the same imbalance, mirrored.

## The fix

**Move the heap and the in-flight counts out of the picker**, into per-worker state keyed by `(resource_key, priority)` — stable across scaling and health flips, unlike the picker version. The picker now *reconciles* that heap with the current node set instead of rebuilding it:

- surviving nodes keep their load
- a new node starts empty and is preferred right away
- a node that leaves keeps its count, so its score is restored if it comes back
- every picker generation shares one view of the load, so a release always lands on the heap in use

**Single-node upstreams now go through the balancer.** `pick_server` used to skip it when `#nodes == 1`, so anything routed while the upstream was alone was never counted — and a 1→N scale out then saw the surviving node as empty. One node is where a k8s deployment or a discovery service *starts*, so this was most of the bug, not an edge case.

**The state is held weakly** (`__mode = "v"`), which is exactly the lifetime it needs: a picker holds its state, an in-flight request holds its picker, the picker cache holds the current one. So a state lives as long as anything can still release into it — and when nothing can, there is no connection left to count. Nothing to size, nothing to evict. (An LRU would rank states by how recently they were *rebuilt*; a stable upstream full of long-lived connections is precisely the one that is never rebuilt.)

Scores are derived from the count, `(count + 1) / weight`, instead of accumulated with `± effect_weight`. Same semantics, exact over time.

## Release discipline

Because the state now outlives the picker, releasing a server the request no longer holds is no longer self-healing — it used to be washed away by the next rebuild, now it is written into shared state for good.

So `after_balance` owns the invariant: **`ctx.balancer_server` names only the server the request currently holds**, and releasing it clears the field. Three release sites relied on the caller to remember, and didn't:

| where | what happened |
|---|---|
| `pick_server`, on entering a retry | ran out of servers → log phase released the same one again |
| `pick_server`, skipping an unhealthy node | same |
| `ai-proxy-multi.retry_on_error` | releases before it knows whether it will pick again |

The first needs nothing unusual to hit: `retries` is derived from *all* nodes while only the *healthy* ones are picked from, so one node failing its health check is enough. `ai-proxy-multi` is fixed by construction — the plugin isn't touched.

Two more, in the same vein:
- `ctx.server_picker` is published as soon as a server is held, not at the end of `pick_server`, so a request that bails out after picking one still releases it.
- The retry entry only marks a server tried when there is **another one to move to**. Otherwise a single-node upstream would mark its only node tried, `get()` would find nothing left, and a configured `retries` would turn a retryable blip into a hard 502.

## Stream

`stream_log_phase()` never called `after_balance`. On L4 the count only ever went up, so `least_conn` could not balance TCP long connections at all, and `ctx.balancer_tried_servers` leaked.

## Scope and known limits

State is shared per `(resource_key, priority)` per worker. An upstream with **no stable resource key** falls back to state private to its picker — today's behaviour, so no regression, but no fix either:

- **traffic-split inline upstreams** — `up_conf` is synthesised in the plugin, and its own picker cache key already embeds the nodes table address, so there is nothing stable to key on
- **`ai-proxy-multi` instances** (`ups_tab = {}`)
- **`conf_server`** in the EE port (`picker.new(nodes, {})`)

Two things routing single-node `least_conn` through the balancer changes:

1. **Cost.** Such upstreams now pay the picker path (`lrucache_server_picker` + `lrucache_addr` + a size-1 heap peek/update per request) — the same cost every multi-node upstream already pays, but a real one on the hottest path for the most common topology.
2. **Unbracketed IPv6 node hosts.** A single node used to bypass `parse_addr`; it no longer does, so `"::1:1980"` now fails the way it already fails everywhere else. **Pre-existing bug**, not a new one — on `origin/master`, a *two*-node upstream with unbracketed IPv6 hosts crashes identically (`balancer.lua:238: attempt to concatenate field 'port' (a nil value)`) for every balancer type. The Admin API rejects such nodes; only the DP-side check skips that validation. Worth fixing, but on its own: the addr string is built in `transform_node` and shared with chash and the health checker.

No new config, no shared dict, on by default. Counting stays per-worker, matching the existing per-worker heap. Short requests return their count on completion, so steady-state behaviour is unchanged.

## Tests

**These three fail on `master`:**

| test | what it covers |
|---|---|
| `least_conn3.t` TEST 9 | the issue as reported — real WebSocket sessions, scaled out underneath them |
| `least_conn3.t` TEST 10 | the 1→N case |
| `least_conn3.t` TEST 7 | the same scale-out with long-running HTTP requests |

The rest pin the internals: count kept across a rebuild, drained node returns to the pool, scale-down, releases on the old picker seen by the new one, priority isolation, weight change, keyless-upstream fallback, and no double release (TEST 6 drives the real `balancer.pick_server`).

Also `least_conn.t` TEST 5 (more retries than nodes — that path had no coverage) and TEST 6 (a single node is retried, because it is the only one there is), and `t/stream-node/least_conn.t` for the L4 release.

## Checklist

### DIFF
--- a/apisix/balancer.lua
+++ b/apisix/balancer.lua
@@ -174,9 +174,9 @@ local function create_server_picker(upstream, checker)
             return server_picker
         end
 
-        core.log.info("upstream nodes: ",
-                      core.json.delay_encode(up_nodes[up_nodes._priority_index[1]]))
-        local server_picker = picker.new(up_nodes[up_nodes._priority_index[1]], upstream)
+        local priority = up_nodes._priority_index[1]
+        core.log.info("upstream nodes: ", core.json.delay_encode(up_nodes[priority]))
+        local server_picker = picker.new(up_nodes[priority], upstream, priority)
         server_picker.addr_to_domain = addr_to_domain
         return server_picker
     end
@@ -248,7 +248,13 @@ local function pick_server(route, ctx)
     local up_conf = ctx.upstream_conf
 
     local nodes_count = #up_conf.nodes
-    if nodes_count == 1 then
+    -- least_conn counts the in-flight connections of every request it routes, so it
+    -- has to see them even while the upstream has a single node: those connections
+    -- are what tells a later scale out that the node is not empty. Skipping the
+    -- balancer here would leave it blind to everything routed before the second
+    -- node showed up, which is the state a k8s deployment or a discovery service
+    -- starts from. See #12217
+    if nodes_count == 1 and up_conf.type ~= "least_conn" then
         local node = up_conf.nodes[1]
         ctx.balancer_ip = node.host
         ctx.balancer_port = node.port
@@ -268,7 +274,10 @@ local function pick_server(route, ctx)
     ctx.balancer_try_count = (ctx.balancer_try_count or 0) + 1
     if ctx.balancer_try_count > 1 then
         if ctx.server_picker and ctx.server_picker.after_balance then
-            ctx.server_picker.after_balance(ctx, true)
+            -- remembering the server as tried is what keeps the next pick off it, so
+            -- only do it when there is another one to move to. With a single node the
+            -- retry has to land on it again, the way the fast path below always did
+            ctx.server_picker.after_balance(ctx, nodes_count > 1)
         end
 
         if checker then
@@ -331,6 +340,9 @@ local function pick_server(route, ctx)
         return nil, "failed to find valid upstream server, all upstream servers tried"
     end
     ctx.balancer_server = server
+    -- from here on the request holds a server, so the log phase must be able to
+    -- release it even if we bail out below
+    ctx.server_picker = server_picker
 
     local domain = server_picker.addr_to_domain[server]
     local res, err = lrucache_addr(server, nil, parse_addr, server)
@@ -347,7 +359,6 @@ local function pick_server(route, ctx)
     if is_http and ctx.var then
         ctx.var.upstream_unresolved_host = ctx.upstream_unresolved_host
     end
-    ctx.server_picker = server_picker
     res.upstream_host = parse_server_for_upstream_host(res, ctx.upstream_scheme)
 
     return res
--- a/apisix/balancer/chash.lua
+++ b/apisix/balancer/chash.lua
@@ -123,6 +123,11 @@ function _M.new(up_nodes, upstream)
             return servers[id]
         end,
         after_balance = function (ctx, before_retry)
+            -- the release is what makes the request stop holding the server, so drop
+            -- the reference here instead of leaving every caller to remember
+            local server = ctx.balancer_server
+            ctx.balancer_server = nil
+
             if not before_retry then
                 if ctx.balancer_tried_servers then
                     core.tablepool.release("balancer_tried_servers", ctx.balancer_tried_servers)
@@ -132,11 +137,15 @@ function _M.new(up_nodes, upstream)
                 return nil
             end
 
+            if not server then
+                return nil
+            end
+
             if not ctx.balancer_tried_servers then
                 ctx.balancer_tried_servers = core.tablepool.fetch("balancer_tried_servers", 0, 2)
             end
 
-            ctx.balancer_tried_servers[ctx.balancer_server] = true
+            ctx.balancer_tried_servers[server] = true
             ctx.balancer_tried_servers_count = (ctx.balancer_tried_servers_count or 0) + 1
         end,
         before_retry_next_priority = function (ctx)
--- a/apisix/balancer/ewma.lua
+++ b/apisix/balancer/ewma.lua
@@ -187,12 +187,21 @@ local function _ewma_find(ctx, up_nodes)
 end
 
 local function _ewma_after_balance(ctx, before_retry)
+    -- the release is what makes the request stop holding the server, so drop the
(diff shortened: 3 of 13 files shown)


membphis:
[P2] This change regresses single-node `least_conn` upstreams that use an unbracketed IPv6 host. The old single-node fast path used the structured `host` and `port` fields directly, but this PR now sends `least_conn` through the picker, where `transform_node()` builds `::1:1980`. `core.utils.parse_addr()` treats that as an IPv6 address without a port, so the request fails. The fact that multi-node upstreams already have the same bug does not make expanding it to the common single-node path backward compatible. Please make the balancer server key IPv6-safe and add single- and multi-node IPv6 regression coverage.

shreemaan-abhishek (dismissed):

AlinsRan:
Fixed — you're right that pre-existing multi-node breakage doesn't make the single-node expansion acceptable.

**Root cause:** the unbracketed host only reaches the balancer via **service discovery** — it skips both the Admin API's `check_schema` and `check_upstream_conf`'s bracketing, so `::1` arrives raw and `transform_node` builds the ambiguous `::1:1980`.

**Fix:** bracket once in `fill_node_info`, the single point every node passes before *both* the balancer and the health checker consume it (a key-only fix would leave the health checker seeing bare `::1`). `original_nodes` keeps the bare host, so a discovery re-fetch isn't seen as a node change.

**Narrowing:** only a node with an explicit port is bracketed — a portless `"::1:1980"` is a valid IPv6 literal as much as `host ::1 port 1980`, so it's left to existing malformed-config handling (keeps `upstream-ipv6.t` TEST 10). Discovery endpoints always carry a port.

**Tests** (`t/node/upstream-discovery-ipv6.t`): single- and multi-node bare-IPv6 discovery (both crash on master), plus a re-fetch case asserting the picker builds only once.

nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13668, opened 2026-07-07
fix: invalidate secret LRU cache on /secrets changes

shreemaan-abhishek:
### Description

After a `secret` object is deleted via the Admin API, `$secret://` references keep resolving to the previously cached value indefinitely. Updates to a secret are only picked up after the cache TTL window.

**Root cause:** `apisix/secret.lua` resolved secret URIs through the secret LRU cache with a constant version string:

```lua
return secrets_cache(uri, "", fetch_by_uri, uri)
```

`core.lrucache` uses the `version` argument to detect staleness (an entry is re-resolved only when `obj.ver ~= version`). Every other cached resource passes its real `conf_version`; the secret cache passed `""`, so no `/secrets` change ever invalidated an entry. With `refresh_stale = true` an expired entry keeps being served while a background timer retries, so once the secret is deleted the refresh fails forever and the stale value is served forever.

**Fix:** Pass the real secrets `conf_version` (already exposed via `_M.secrets()`). A change to `/secrets` bumps the version, so the cache re-resolves on the next access, matching how route/service caches behave.

Notes:
- `/secrets` has one shared `conf_version`, so any secret change invalidates all cached secret values (consistent with route/service cache behavior).
- `$env://` shares this cache; bumping on `/secrets` changes also drops env entries, which is harmless since env values are static.

### Checklist

- [x] I have updated the documentation to reflect this change (no user-facing config/behavior change; restores expected behavior)

### Test

`t/secret/secret_lru.t` adds a case that resolves a secret through the cache, deletes it, and asserts the reference stops resolving (falls back to the literal string) instead of serving the stale value. Verified the test fails before the fix and passes after.

### DIFF
--- a/apisix/secret.lua
+++ b/apisix/secret.lua
@@ -233,7 +233,9 @@ local function fetch(uri, use_cache)
         return val
     end
 
-    return secrets_cache(uri, "", fetch_by_uri, uri)
+    -- pass the secrets conf_version so the cache re-resolves when /secrets changes
+    local version = secrets and secrets.conf_version or ""
+    return secrets_cache(uri, version, fetch_by_uri, uri)
 end
 
 local function retrieve_refs(refs, use_cache)
--- a/t/secret/secret_lru.t
+++ b/t/secret/secret_lru.t
@@ -96,3 +96,75 @@ GET /t
     }
 --- response_body
 nil
+
+
+
+=== TEST 2: store secret into vault
+--- exec
+VAULT_TOKEN='root' VAULT_ADDR='http://0.0.0.0:8200' vault kv put kv/apisix/lru-key/jack key=value
+--- response_body
+Success! Data written to: kv/apisix/lru-key/jack
+
+
+
+=== TEST 3: deleted secret is evicted from the LRU cache
+--- config
+    location /t {
+        content_by_lua_block {
+            local t = require("lib.test_admin").test
+            local secret = require("apisix.secret")
+
+            -- configure the vault secret manager
+            local code, body = t('/apisix/admin/secrets/vault/lru',
+                ngx.HTTP_PUT,
+                [[{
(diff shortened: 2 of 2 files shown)


apisix/secret.lua
```diff
@@ -233,7 +233,9 @@ local function fetch(uri, use_cache)
         return val
     end
 
-    return secrets_cache(uri, "", fetch_by_uri, uri)
+    -- pass the secrets conf_version so the cache re-resolves when /secrets changes
```
nic-6443, line 236: I disagree with this fix. The design of this LRU Cache dictates that all changes take effect only after the TTL expires. Any modification, whether to the values stored in Secret Manager or to the Secret Manager's own configurations, must wait for the TTL to apply. This design rule is consistent.

A new issue introduced by the current fix is that any single configuration change in Secret Manager will invalidate all cached entries, triggering a cache storm.

shreemaan-abhishek: > The design of this LRU Cache dictates that all changes take effect only after the TTL expires.

while this is true, the version arg is also a invalidation mechanism here. Every other cache (routes, services, consumers) passes its real conf_version and picks up changes immediately. The secret cache passing "" was the odd one out, which is why deleted/updated secrets are served stale until it expires. This just makes secrets behave like everything else.

and about the the "storm", re-resolution is lazy (each entry only re-fetches on its next access, not all at once), bounded by the number of distinct secret refs, and only triggered by admin-plane changes, not request traffic. Same as how one route edit rebuilds the router cache.

this is indeed a tradeoff but is serving a deleted secret forever doesn't sound like a good tradeoff.

nic-6443 (approved):

membphis (approved):
LGTM

AlinsRan (approved):

PR #13670, opened 2026-07-07
feat(ai-rate-limiting): support redis policy for shared counters

### LINKED ISSUE #12482: help request: How to config ai-rate-limiting plugin's  counter to redis, as  limit-count plugin

shreemaan-abhishek:
### Description

The `ai-rate-limiting` Plugin currently keeps token counters only in per-worker shared memory (`policy = "local"`). In a multi-node deployment the counters are not shared across nodes, so the effective quota scales with the number of gateway nodes and over-allows.

This PR adds a Redis-based counter policy to the Plugin, reusing the existing `limit-count` Redis backends, so counters can be centralized and a single quota is enforced across all nodes.

Changes:

- Export `policy_to_additional_properties` from `apisix/plugins/limit-count/init.lua` so the Plugin can reuse the shared Redis schema branches.
- Add `policy` (`local` / `redis` / `redis-cluster` / `redis-sentinel`) and `allow_degradation` schema fields, plus the Redis attribute wiring in `transform_limit_conf`.
- Document the new attributes and add a Redis example to `docs/en/latest/plugins/ai-rate-limiting.md`.
- Add schema-validation and Redis rate-limiting test cases.

The default remains `local`, so existing configurations are unaffected.

Fixes #

### Checklist

Fixes #12482

### DIFF
--- a/apisix/plugins/ai-rate-limiting.lua
+++ b/apisix/plugins/ai-rate-limiting.lua
@@ -259,34 +259,12 @@ local function transform_limit_conf(plugin_conf, instance_conf, instance_name)
         limit_conf.reset_header = "X-AI-RateLimit-Reset-" .. name
     end
 
-    if plugin_conf.policy == "redis" then
-        limit_conf.redis_host = plugin_conf.redis_host
-        limit_conf.redis_port = plugin_conf.redis_port
-        limit_conf.redis_username = plugin_conf.redis_username
-        limit_conf.redis_password = plugin_conf.redis_password
-        limit_conf.redis_database = plugin_conf.redis_database
-        limit_conf.redis_timeout = plugin_conf.redis_timeout
-        limit_conf.redis_ssl = plugin_conf.redis_ssl
-        limit_conf.redis_ssl_verify = plugin_conf.redis_ssl_verify
-    elseif plugin_conf.policy == "redis-cluster" then
-        limit_conf.redis_cluster_nodes = plugin_conf.redis_cluster_nodes
-        limit_conf.redis_cluster_name = plugin_conf.redis_cluster_name
-        limit_conf.redis_password = plugin_conf.redis_password
-        limit_conf.redis_timeout = plugin_conf.redis_timeout
-        limit_conf.redis_cluster_ssl = plugin_conf.redis_cluster_ssl
-        limit_conf.redis_cluster_ssl_verify = plugin_conf.redis_cluster_ssl_verify
-    elseif plugin_conf.policy == "redis-sentinel" then
-        limit_conf.redis_sentinels = plugin_conf.redis_sentinels
-        limit_conf.redis_master_name = plugin_conf.redis_master_name
-        limit_conf.redis_username = plugin_conf.redis_username
-        limit_conf.redis_password = plugin_conf.redis_password
-        limit_conf.sentinel_username = plugin_conf.sentinel_username
-        limit_conf.sentinel_password = plugin_conf.sentinel_password
-        limit_conf.redis_role = plugin_conf.redis_role
-        limit_conf.redis_connect_timeout = plugin_conf.redis_connect_timeout
-        limit_conf.redis_read_timeout = plugin_conf.redis_read_timeout
-        limit_conf.redis_keepalive_timeout = plugin_conf.redis_keepalive_timeout
-        limit_conf.redis_database = plugin_conf.redis_database
+    -- copy the redis fields straight from the policy's schema so no field is missed
+    local extra = policy_to_additional_properties[plugin_conf.policy]
+    if extra then
+        for k in pairs(extra.properties) do
+            limit_conf[k] = plugin_conf[k]
+        end
     end
     return limit_conf
 end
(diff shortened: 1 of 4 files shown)


t/plugin/ai-rate-limiting.t
```diff
@@ -1538,3 +1538,133 @@ X-AI-Fixture: openai/chat-model-echo.json
...
+        content_by_lua_block {
+            local redis = require("resty.redis")
+            local red = redis:new()
+            red:set_timeout(1000)
+            local ok, err = red:connect("127.0.0.1", 6379)
+            if not ok then
+                ngx.say("failed to connect: ", err)
+                return
+            end
+            red:flushall()
```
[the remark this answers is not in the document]
shreemaan-abhishek: there is no parallel usage in CI so ignore this.

membphis:
I think this should be fixed before merge.

`ai-rate-limiting` now exposes Redis credential fields by reusing the `limit-count` Redis schemas, but its own plugin schema does not declare those fields in `encrypt_fields`. APISIX config encryption is driven by each plugin schema's `encrypt_fields`, so `redis_password` / `sentinel_password` would be stored as plain plugin config here, while comparable plugins such as `limit-count`, `limit-req`, and `ai-cache` encrypt Redis passwords.

There is also a sentinel wiring gap: the schema accepts Redis master credentials for `redis-sentinel`, and the backend uses `redis_username` / `redis_password` for the Redis master connection, but `transform_limit_conf` only forwards the sentinel credentials and does not pass `redis_username` / `redis_password` in the sentinel branch. A valid sentinel config with Redis master auth can pass validation but fail at runtime.

Suggested fix: add `encrypt_fields = {"redis_password", "sentinel_password"}` to the `ai-rate-limiting` schema, forward `redis_username` / `redis_password` for `redis-sentinel`, and add coverage for encrypted Redis credentials plus sentinel master auth.

shreemaan-abhishek:
Thanks @membphis, good catches. Fixed in the latest commit:

- Added \`encrypt_fields = {"redis_password", "sentinel_password"}\` to the plugin schema so the Redis credentials are encrypted at rest, matching \`limit-count\` / \`limit-req\`.
- Forwarded \`redis_username\` / \`redis_password\` in the \`redis-sentinel\` branch of \`transform_limit_conf\` so a Redis master with ACL/requirepass can be authenticated.
- Added a schema test for a sentinel config with Redis master auth, and a test asserting \`redis_password\` is encrypted at rest.

membphis (dismissed):
LGTM

apisix/plugins/ai-rate-limiting.lua
```diff
@@ -191,43 +223,71 @@ end
...
+        limit_conf.redis_username = plugin_conf.redis_username
+        limit_conf.redis_password = plugin_conf.redis_password
+        limit_conf.sentinel_username = plugin_conf.sentinel_username
+        limit_conf.sentinel_password = plugin_conf.sentinel_password
+        limit_conf.redis_role = plugin_conf.redis_role
+        limit_conf.redis_connect_timeout = plugin_conf.redis_connect_timeout
+        limit_conf.redis_read_timeout = plugin_conf.redis_read_timeout
+        limit_conf.redis_keepalive_timeout = plugin_conf.redis_keepalive_timeout
+        limit_conf.redis_database = plugin_conf.redis_database
     end
```
AlinsRan: Hand-copying the redis fields per policy is the root cause of the sentinel, and it's still leaking: neither the `redis` nor the `redis-cluster` branch forwards `redis_keepalive_timeout` / `redis_keepalive_pool`, even though both are accepted (with defaults) by the schema branches you're reusing — see `apisix/utils/redis-schema.lua:47-52` and `:80-85`.

So today a user who sets them gets them silently dropped: `redis` falls back to the hardcoded `or 10000` / `or 100` in `limit-count/util.lua:177-178`, and `redis-cluster` passes `nil` into `utils/rediscluster.lua:27-28`.

Rather than patching two more lines, drive the copy from the schema you already have in hand — same trick `limit-count/init.lua:274-278` uses for the `group` whitelist:

```lua
local extra = policy_to_additional_properties[plugin_conf.policy]
if extra then
    for k in pairs(extra.properties) do
        limit_conf[k] = plugin_conf[k]
    end
end
```

That drops all three branches (46 lines) and makes future schema additions impossible to miss.

shreemaan-abhishek: nice, I will add this.

shreemaan-abhishek: Good catch, adopted. Replaced the three hand-copied branches with the schema-driven copy so every field the policy schema accepts is forwarded (this fixes the dropped `redis_keepalive_timeout` / `redis_keepalive_pool` on the `redis` and `redis-cluster` policies too). Ported the same change to keep it DRY.

membphis:
The attributes table in `docs/en/latest/plugins/ai-rate-limiting.md` is missing `redis_keepalive_pool`, and it documents `redis_keepalive_timeout` as sentinel-only (default 60000).

However, the `redis` and `redis-cluster` schema branches also define `redis_keepalive_timeout` (default 10000) and `redis_keepalive_pool` (default 100), and both fields are forwarded to limit-count by the schema-driven copy in `transform_limit_conf`, so they take effect for all three redis policies.

Please complete the doc table to cover these two fields for `redis` / `redis-cluster` as well, with the correct defaults.

membphis (approved):

nic-6443 (approved):

AlinsRan (approved):

PR #13671, opened 2026-07-08
fix(openid-connect): scope session cookie per route to avoid state collision

shreemaan-abhishek:
### Description

When two `openid-connect` routes on the same host share a `session.secret`, they both fall back to [lua-resty-session](https://github.com/bungle/lua-resty-session)'s default cookie name `"session"`. An unauthenticated visit to route A stores the pre-login `state`/`nonce` in that cookie; a subsequent unauthenticated visit to route B (carrying the same cookie) overwrites it. When route A's callback then arrives, the state no longer matches and `lua-resty-openidc` returns:

```
state from argument: <A> does not match state restored from session: <B>
```

surfacing as an HTTP 500 to the user. This is purely a cookie-naming collision, not an auth bypass.

This PR defaults an unset `session.cookie_name` to a stable per-route name derived from the route id (falling back to `client_id`), so sibling routes each get their own session cookie and no longer clobber each other's pre-login state.

**Behavior change / opt-in sharing:** cross-route session sharing (e.g. SSO across routes on the same host with redis storage) previously worked implicitly via the shared default cookie name. With per-route defaults, sharing is now explicit: set the same `session.cookie_name` on each route that should share a session. An operator-set `session.cookie_name` is always honored. The redis session-sharing test is updated to set a shared `cookie_name` accordingly.

On upgrade, in-flight sessions using the old default `"session"` cookie are not recognized and users re-authenticate once. Config API and behavior contracts are otherwise unchanged.

### Checklist

- [x] I have verified that this change is backward compatible (config and behavior contracts preserved; see the sharing/re-login notes above)

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -33,11 +33,26 @@ local ngx_encode_base64 = ngx.encode_base64
 local plugin_name       = "openid-connect"
 
 
--- Session config is passed as-is to resty.session.start(); the only
--- translation is the legacy session.cookie.lifetime alias from the
--- lua-resty-session 3.x schema, which is mapped to absolute_timeout
--- when the latter is unset.
-local function build_session_opts(session_conf)
+-- Derive a stable, per-route session cookie name. resty.session names every
+-- cookie "session" by default, so sibling openid-connect routes on the same
+-- host that share a session.secret overwrite each other's pre-login state.
+-- Keying the cookie name on the route id isolates them.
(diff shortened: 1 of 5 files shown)


membphis (approved):
LGTM

nic-6443 (commented):
If users want to distinguish login status across multiple routes, they can manually configure different Cookie names. Configuring the same Cookie name among multiple routes is also a reasonable scenario in itself.

PR #13678, opened 2026-07-09
fix(grpc-transcode): encode empty repeated fields as JSON arrays

### LINKED ISSUE #11440: bug:  grpc-transcode cant transcode empty array list to [] , but it did to {}

AlinsRan:
### Description

Fixes #11440

An empty `repeated` field is decoded by lua-protobuf into an empty Lua table. Lua cannot tell an empty array from an empty object, so `cjson.encode` emits `{}` where the client expects `[]`.

lua-protobuf 0.5.3 (the version pinned in the rockspec) supports a `*array` [default metatable](https://github.com/starwing/lua-protobuf#default-values). Setting it to `core.json.array_mt` inside `compile_proto()` makes `pb.decode` tag every repeated field as an array by itself:

```lua
pb.defaults("*array", core.json.array_mt)
```

Behavior with the plugin's default `pb_option`:

- empty repeated fields encode as `[]`
- map fields keep object semantics (`{}`) — lua-protobuf distinguishes maps from arrays natively
- nested repeated fields inside messages, repeated messages and map values are all covered
- it works for both the text proto path and the binary descriptor set path

`pb.defaults` is bound to the active pb state, so it is set per compiled proto rather than at module load time. The same is done for the gRPC status pb state, so that an empty `ErrorStatus.details` is also rendered as `[]`.

This is a pure decode-side change with no per-request cost.

An alternative implementation exists in #12649, which builds a message descriptor index from the proto and walks the decoded table on every response. Besides the per-request traversal, the numeric `label`/`type` comparisons it relies on do not match on the binary descriptor path (`pb.decode` of a `FileDescriptorSet` yields `"LABEL_REPEATED"` / `"TYPE_MESSAGE"` under the default `enum_as_name`), so that path is left unfixed there.

### Checklist

### DIFF
--- a/apisix/plugins/grpc-transcode/proto.lua
+++ b/apisix/plugins/grpc-transcode/proto.lua
@@ -109,6 +109,11 @@ local function compile_proto(content)
         end
     end
 
+    -- an empty repeated field is decoded into an empty Lua table, which cjson would
+    -- encode as `{}`. Tag it as an array so that it is encoded as `[]` instead.
+    -- The default is bound to the current pb state, so it has to be set per proto.
+    pb.defaults("*array", core.json.array_mt)
+
     -- fetch pb state
     compiled.pb_state = pb.state(old_pb_state)
     return compiled
@@ -237,6 +242,8 @@ local function init_status_pb_state()
             pb.state(old_pb_state)
             return "failed to load grpc status protocol: " .. err
         end
+
+        pb.defaults("*array", core.json.array_mt)
 
         status_pb_state = pb.state(old_pb_state)
     end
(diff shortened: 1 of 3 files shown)


membphis (approved):
LGTM

shreemaan-abhishek (approved):

nic-6443 (approved):

PR #13686, opened 2026-07-09
fix(cli): skip commented-out lines when resolving env vars in apisix.yaml

### LINKED ISSUE #13685: bug: commented out env variables cause app to fail to boot
### Current Behavior

There is a bug that started happening in v3.17.0.

I have a few commented out environment variables using the syntax `${{MY_ENV_VAR}}`. It appears something is attempting to resolve those, even though commented out.

It should not be resolving env vars from commented out lines.
```ansii
/usr/local/openrestv//luajit/bin/luajit ./apisix/cli/apisix.lua init
failed to read local vaml config of apisix: failed to handle configuration: can't find environment variable MY_ENV_VAR
```

The routes files contains:

```yaml
# ${{MY_ENV_VAR}}
```

### Expected Behavior

The app should start without weeping due to commented out environment variables.

### Error Logs

_No response_

### Steps to Reproduce

See above.

### Environment

From helm chart v2.16.0 = APISix 3.17.0.

asdf2014:
### Description

Standalone mode fails to boot when `apisix.yaml` contains a commented-out env var reference:

```yaml
# ${{MY_ENV_VAR}}
```

Since #13078 (shipped in 3.17.0), `${{VAR}}` is substituted on the raw text *before* YAML parsing so the parser can infer scalar types — which means comment lines are substituted too, and an unset var inside a comment aborts startup with `can't find environment variable MY_ENV_VAR`. (`config.yaml` is unaffected: it substitutes on the parsed table where comments are already gone; same for JSON standalone configs, which also resolve post-parse.)

This patch makes the raw-text substitution run line by line, keeping full-line comments (and blank / no-reference lines) verbatim, reusing the existing `is_empty_yaml_line` predicate. A leading UTF-8 BOM is detached first so a first-line comment is still recognized. Unset vars on data lines still fail with the same error message; output is byte-identical to the previous implementation for comment-free input (line endings, missing trailing newline, and the `#END` marker are all preserved). Both callers — CLI init (`cli/file.lua`) and the runtime hot-reload (`core/config_yaml.lua`) — share the fixed function.

Behavior notes (accepted trade-offs of the text-level heuristic, documented in the code comment):

- inline trailing comments (`key: val # ${{VAR}}`) are still substituted — `#` may legally appear inside YAML strings, and telling those apart requires a full parse, which can only happen after substitution;
- a block-scalar content line that itself starts with `#` is now skipped like a comment (no in-tree config or doc example hits this; the failure mode is a visible literal `${{VAR}}`, not a crash);
- a `${{...}}` reference spanning a physical line no longer matches (all in-tree references are single-line);
- a var referenced *only* in a comment no longer emits an `env VAR;` nginx directive (comments having side effects was part of the bug).

The new test in `t/cli/test_standalone.sh` covers both paths: `make init` (CLI) and `make run` + request (runtime hot-reload) against a config whose commented-out lines reference unset vars.

Fixes #13685

### Checklist

### DIFF
--- a/apisix/cli/file.lua
+++ b/apisix/cli/file.lua
@@ -31,6 +31,7 @@ local getenv = os.getenv
 local str_gmatch = string.gmatch
 local str_find = string.find
 local str_sub = string.sub
+local table_concat = table.concat
 local print = print
 
 local _M = {}
@@ -92,12 +93,32 @@ end
 
 -- Substitute env vars in raw text before parsing. YAML parser then infers
 -- types naturally: `"${{VAR}}"` stays string, `${{VAR}}` infers from value.
+-- Substitution runs line by line so that full-line comments stay untouched:
+-- a commented-out `# ${{VAR}}` must not fail startup when VAR is unset.
+-- Known limits of this text-level heuristic: an inline trailing comment is
+-- still substituted (`#` may legally appear inside YAML strings), and a
+-- block-scalar content line starting with `#` is skipped like a comment.
 local function resolve_conf_var_in_text(text)
-    local new_text, _, err = var_sub(text)
-    if err then
-        return nil, err
+    local parts = {}
+    -- a UTF-8 BOM would hide the `#` of a first-line comment, so detach it
+    if str_sub(text, 1, 3) == "\239\187\191" then
+        parts[1] = str_sub(text, 1, 3)
+        text = str_sub(text, 4)
     end
-    return new_text
+    for line, eol in str_gmatch(text, "([^\n]*)(\n?)") do
+        if is_empty_yaml_line(line) or not str_find(line, "${{", 1, true) then
+            -- comment / blank / no reference: keep the line as-is
+            parts[#parts + 1] = line
+        else
+            local new_line, _, err = var_sub(line)
+            if err then
+                return nil, err
+            end
+            parts[#parts + 1] = new_line
+        end
+        parts[#parts + 1] = eol
+    end
+    return table_concat(parts)
 end
 
 
--- a/t/cli/test_standalone.sh
+++ b/t/cli/test_standalone.sh
@@ -660,5 +660,55 @@ fi
 
 echo "passed: missing env var produces clear startup error"
 
+# test: unset env vars referenced only in commented-out lines must be ignored
+echo '
+apisix:
+  enable_admin: false
+deployment:
+  role: data_plane
+  role_data_plane:
+    config_provider: yaml
+' > conf/config.yaml
+
+echo '
+# ${{COMMENTED_UNSET_VAR}}
+routes:
+  -
+    uri: /test-commented-var
+    # indented comment: ${{ANOTHER_COMMENTED_UNSET_VAR}}
+    plugins:
+      response-rewrite:
+        body: "commented-ok"
+        status_code: 200
+    upstream:
+      nodes:
+        "127.0.0.1:9091": 1
+      type: roundrobin
+#END
+' > conf/apisix.yaml
+
+unset COMMENTED_UNSET_VAR
+unset ANOTHER_COMMENTED_UNSET_VAR
+make init
+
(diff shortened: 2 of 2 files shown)


AlinsRan:
The full-line skip does fix the reported case, but I'd suggest deciding whether we want to keep patching at the text level at all before merging this.

Three things I verified by calling master's `resolve_conf_var_in_text()` directly (`82fcf548f`):

```
key: 1  # ${{UNSET}}           -> ERR: can't find environment variable UNSET   # inline comment, not covered here
key: ${{V}}   V='secret # x'   -> {"key":"secret"}                             # value silently truncated
key: ${{V}}   V=$'a\nb: 2'     -> {"b":2,"key":"a"}                            # env value injects a top-level key
```

And one regression this PR would add: a block-scalar content line starting with `#` is now skipped, so

```yaml
response_example: |
  # ${{TITLE}}
```

silently stops being substituted.

The root cause is that pre-parse text substitution (#13078) re-implements YAML lexing outside lyaml. An alternative that avoids all four: replace each `${{...}}` with a unique plain-scalar token before parsing (recording var / default / whether it was directly wrapped in quotes), let lyaml parse — comments, block scalars and quoting are then handled by the parser — then resolve the tokens that survived into the parsed tree. "Env var not found" only fires when a reference actually reaches the config, so commented-out references never fail by construction, in either comment style.

I prototyped that and ran it against the cases #13078 added plus the new ones:

```
full-line comment   {"key":1}              inline comment     {"key":1}
missing var         ERR (unchanged)        unquoted number    {"weight":3}
quoted number       {"key":"3"}            unquoted bool      {"flag":true}
default fallback    {"key":"fallback"}     var in key         {"mykey":1}
embedded in string  {"key":"pre-3-post"}   big int quoted     {"id":"356002209726529540"}
value containing #  {"key":"secret # not-a-comment"}
multiline value     {"key":"a\nb: 2"}      block scalar       {"body":"# 3 stays\n"}
```

One intentional behavior change there: env values are always injected as scalars, so `nodes: ${{NODES_JSON}}` no longer expands into a YAML structure. That is the same path as the injection case above, so I'd argue for closing it, but it needs a changelog note.

If the preference is to keep this PR minimal, it should at least become comment-aware instead of line-aware: find the first `#` that is outside quotes and preceded by start-of-line or whitespace, substitute only what precedes it, and track `|`/`>` block-scalar ranges so their content is still substituted. That covers the inline-comment case and removes the regression, though truncation and injection would remain.

Ref #13685.

asdf2014:
Agreed on the root cause: pre-parse text substitution re-implements YAML lexing outside lyaml, and the tokenize-then-parse-then-resolve approach is the right end state. It fixes all four cases by construction, and your prototype results (especially `value containing #` and the multiline injection) make a strong case. Closing the structural-expansion path also seems right to me, since it is the same injection surface, and a changelog entry covers it.

On the middle option (comment-aware substitution): I would rather not go there. Tracking quotes plus `|`/`>` ranges at the text level is halfway to a YAML lexer, with its own corner cases (multi-line quoted strings, `#` inside flow collections, nested block-scalar indentation), while truncation and injection still remain. If we want more than the minimal fix, the token approach is the better place to spend that complexity.

So my suggestion for sequencing: keep this PR as the minimal stopgap for the 3.17 regression (#13685 blocks startup today, and this change is byte-transparent for comment-free input), and land your tokenization as the follow-up that replaces this code path entirely. The block-scalar direction change here is documented in the code comment and hits nothing in-tree, and it disappears together with the rest of the text-level heuristic once tokenization lands. If the maintainers would rather go straight to the token approach in one step, I am also happy to close this in favor of your prototype, whichever you prefer.

cfitzw:
Any further updates/progress on this?

Other than having to remove all comments, this is a blocker to move to 3.17.

cfitzw:
@AlinsRan - can you review this further and provide directive?

PR #13691, opened 2026-07-13
fix(request-validation): guard non-string Content-Type before lower()

shreemaan-abhishek:
### Description

The `request-validation` plugin decides whether a request body is form-urlencoded by lowercasing the `Content-Type` header and prefix-matching it:

```lua
if headers["content-type"]
    and core.string.has_prefix(headers["content-type"]:lower(), "application/x-www-form-urlencoded")
then
```

`headers` comes from `ngx.req.get_headers()`, which returns a **table** (not a string) when a header appears more than once. So a request that carries a duplicated `Content-Type` header makes `headers["content-type"]` a table, and `table:lower()` raises `attempt to call method 'lower' (a nil value)` in the `rewrite` phase. The request then fails with a 500 and an error log, instead of being validated (or cleanly rejected).

This PR guards the value with a string type check before calling `:lower()`. A non-string `Content-Type` now falls through to the JSON default path, matching the plugin's intent for anything that is not form-urlencoded.

While here, the existing charset test used an all-lowercase media type, so it never actually exercised the `:lower()` call; it now uses a mixed-case media type. A new test covers the duplicated-header path.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/request-validation.lua
+++ b/apisix/plugins/request-validation.lua
@@ -91,6 +91,13 @@ function _M.rewrite(conf, ctx)
 
         local body_is_json = true
         local content_type = headers["content-type"]
+        -- a duplicated Content-Type is ambiguous: APISIX and the upstream may
+        -- parse the body differently, bypassing validation. reject it.
+        if type(content_type) == "table" then
+            core.log.error("duplicated Content-Type header")
+            return conf.rejected_code, conf.rejected_msg
+        end
+
         if type(content_type) == "string"
             and core.string.has_prefix(
                 content_type:lower(),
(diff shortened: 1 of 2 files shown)


nic-6443 (dismissed):

membphis:
[P2] Conflicting duplicate Content-Type values can bypass consistent body representation validation

When `Content-Type` appears more than once, `ngx.req.get_headers()` returns a table. This change falls through to JSON validation, while the original duplicate headers are still forwarded upstream. Different upstreams may select the first value, the last value, or reject the request, so APISIX can validate one representation while the upstream parses another.

Please reject conflicting duplicate `Content-Type` values (or only accept duplicates after confirming all normalized values are identical), and add tests for both header orders. RFC 9110 Section 8.3 explicitly warns that repeated `Content-Type` values can create interoperability and security issues when recipients handle them differently.

membphis (approved):
LGTM

nic-6443 (approved):

AlinsRan (approved):

PR #13692, opened 2026-07-13
fix(ai-content-moderation): validate deny_code is a valid HTTP status

shreemaan-abhishek:
### Description

`deny_code` in `ai-aws-content-moderation` and `ai-aliyun-content-moderation` was an unbounded `number`, so any value (a negative number, `0`, `700`, a fractional code) could be configured even though the plugins use it as the HTTP status of the deny response.

This bounds `deny_code` to a valid HTTP status range and documents it:

```lua
deny_code = {
    type = "integer",
    minimum = 200,
    maximum = 599,
    default = 200,
    description = "...",
},
```

The default (`200`) and runtime behavior are unchanged. Only previously-invalid configs are now rejected at `check_schema` time.

Fixes #

### Checklist

- [x] I have verified that this change is backward compatible (default and behavior unchanged; only out-of-range values, which never produced a valid HTTP status, are now rejected)


nic-6443 (approved):

membphis (approved):
LGTM

AlinsRan (approved):

PR #13693, opened 2026-07-13
fix(key-auth): propagate real auth error to multi-auth orchestrator

shreemaan-abhishek:
### Description

Under the `multi-auth` plugin, `key-auth` returned the generic `Invalid API key in request` instead of the real failure reason, so the orchestrator could not surface *why* key-auth rejected the request. Its sibling auth plugins (`basic-auth`, `hmac-auth`, `jwt-auth`) already propagate the real reason via `auth_utils.is_running_under_multi_auth(ctx)`; key-auth was the one plugin skipped when this behavior was introduced (#11775 touched `utils/auth.lua` + basic/hmac/jwt but not key-auth).

This applies APISIX's own established pattern to key-auth:

- `find_consumer`: on the consumer-not-found path, return the real `failed to find consumer: <reason>` to the orchestrator when running under multi-auth (mirrors `basic-auth.lua`).
- `rewrite`: on the anonymous-fallback failure, return the real error to the orchestrator when running under multi-auth (mirrors `hmac-auth.lua`).

Standalone key-auth behavior is unchanged: clients still receive the generic message and the detailed reason is still logged.

### Tests

- Updated `t/plugin/multi-auth2.t` TEST 3 to assert the detailed key-auth reason now reaches the multi-auth error log, plus a `no_error_log` guard that the old generic message is gone.
- `t/plugin/multi-auth.t`, `t/plugin/multi-auth2.t`, `t/plugin/key-auth.t` all pass locally.
- Added a note to `docs/en/latest/plugins/multi-auth.md`.

### Checklist

### DIFF
--- a/apisix/plugins/key-auth.lua
+++ b/apisix/plugins/key-auth.lua
@@ -18,6 +18,7 @@ local core     = require("apisix.core")
 local consumer_mod = require("apisix.consumer")
 local plugin_name = "key-auth"
 local schema_def = require("apisix.schema_def")
+local auth_utils = require("apisix.utils.auth")
 
 local schema = {
     type = "object",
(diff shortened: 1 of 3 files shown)


membphis (approved):
LGTM

nic-6443 (approved):

AlinsRan (approved):

PR #13696, opened 2026-07-14
fix(wolf-rbac): clear upstream identity headers when auth response omits userInfo

shreemaan-abhishek:
### Description

The `wolf-rbac` plugin injects `X-UserId` / `X-Username` / `X-Nickname` upstream headers from the auth server's `userInfo`, but only inside the `if type(res.userInfo) == 'table'` guard. When the access-check backend authorizes a request (HTTP 200) without returning `userInfo`, that block is skipped and any client-supplied `X-UserId` / `X-Username` / `X-Nickname` headers reach the upstream unchanged.

This aligns `wolf-rbac` with the other auth plugins (`forward-auth`, `opa`, `openid-connect`, `dingtalk-auth`, `feishu-auth`), which already clear these client-supplied identity headers unconditionally before applying the auth response. The plugin now drops the three headers before the `userInfo` gate, so the happy path still overwrites them from `userInfo` and the no-`userInfo` path no longer forwards client values.

Fixes #

### Checklist

- [ ] I have updated the documentation to reflect this change (no user-facing config or documented behavior changes)

### DIFF
--- a/apisix/plugins/wolf-rbac.lua
+++ b/apisix/plugins/wolf-rbac.lua
@@ -287,13 +287,17 @@ function _M.rewrite(conf, ctx)
 
     local username = nil
     local nickname = nil
+    local prefix = cur_consumer.auth_conf.header_prefix or ''
+    -- drop client-supplied identity headers before trusting the auth response
+    core.request.set_header(ctx, prefix .. "UserId", nil)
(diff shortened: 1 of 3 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

AlinsRan (approved):

PR #13697, opened 2026-07-16
chore: bump apisix-runtime to 1.3.11

nic-6443:
### Description

Bump apisix-runtime from 1.3.9 to 1.3.11.

**1.3.10** picks up [apisix-nginx-module 1.19.8](https://github.com/api7/apisix-nginx-module/releases/tag/1.19.8), which fixes several long-standing defects in the runtime patches:

- balancer keepalive pools are now keyed by the full pool name instead of a crc32 digest (colliding pool names could silently share connections across unrelated upstream pools), the default pool identity covers connect-time properties (local bind address, TLS server name), and `set_current_peer(addr, port, host)` accepts the documented host argument again;
- `ngx.socket.stream():sslhandshake()` no longer fails with an undefined FFI symbol, and `tlshandshake()` no longer mutates the caller's options table (reusing the same file-based mTLS options across connections works now);
- xRPC `move()`/`reset_read_buf()` reject a source socket with a pending read instead of corrupting the in-flight data, and the reset recycles the whole receive chain instead of leaking middle nodes into the request pool;
- privileged agent listeners (`enable_process=privileged_agent`, used by the status API server) are now opened and owned by the master process: a bind conflict fails startup loudly instead of leaving a healthy-looking nginx without the listener, reloads no longer have a no-listener window, socket options on those listeners actually take effect, and the address-conflict error reports the actual owner role.

**1.3.11** adds `--with-stream_realip_module`. nginx does not build this module by default, so `set_real_ip_from` is currently unusable in the `stream` block. The module is inert unless `set_real_ip_from` is configured — its post-accept handler returns `NGX_DECLINED` when no trusted address is set — so it changes nothing for existing deployments.

The runtime-debug checksums in `ci/linux-install-openresty.sh` are updated from the published release artifacts.

### Checklist

---

**Note on the test change:** the runtime bump changes the keepalive pool debug log from `crc32: <hash>` to `name: <pool-name>` (the pool is now keyed by its full name string, which also fixes a crc32-collision bug where distinct pools could be merged). The pool lifecycle is unchanged — only the log label differs — so `t/node/upstream-keepalive-pool.t`'s `grep_error_log` assertions are updated from `crc32: \S+` to `name: \S+`.


AlinsRan (dismissed):

membphis (dismissed):
LGTM

shreemaan-abhishek (approved):

AlinsRan (approved):

membphis (approved):
LGTM

Issue #1597, opened 2020-05-19
request help: Better documentation for standalone mode configuration


wangyjx:
### Issue description
The documentation for standalone mode is very few, it doesn't cover many scenarios, like host match, plugins, people can only make trials with lots of failures. It is time-consuming. Please complete the document for this.

### Environment

* apisix version (cmd: `apisix version`):
* OS:

membphis:
@apache/apisix-committers Anyone wants to do this?

juzhiyuan:
Just waiting for this

PR #13699, opened 2026-07-16
refactor(ai-providers): make the LLM request client ctx-free

AlinsRan:
### Description

`ai-providers/base.lua`'s `build_request` / `request` took the request `ctx` and reached into the downstream request through it. That coupling was wrong in two ways:

1. **It leaked client headers.** `construct_forward_headers` pulled `core.request.headers(ctx)` and always forwarded them, so `ai-request-rewrite` — which calls an LLM to rewrite the body — sent the client's `Authorization`/`Cookie` to that third-party endpoint.
2. **It made a self-contained LLM call impossible without faking a ctx.** None of what `ctx` carried is meaningful for an internal call, so callers had to pass a throwaway ctx; an `auth.gcp` config then blew up on `plugin_ctx_id(fake ctx)`.

So: **drop the `ctx` parameter**, and while the seams are open, cut them where the concepts actually are.

### The client is now three layers, split by what each produces

```lua
provider:build_body(request_body, opts)  -> body, changed   -- shape the body for the target protocol
provider:build_request(conf, body, opts) -> params          -- assemble the HTTP request, sign last
provider:request(conf, request_table, opts) -> status, raw_body   -- = build_body + build_request + send
```

`build_request` used to do all three jobs at once. Its seam followed an implementation detail rather than a concept: it knew about `target_protocol` and mutated the body in four places, yet protocol *conversion* sat outside — purely because that one function takes `ctx`. A reader could not state what the function's job was.

Every protocol-driven decision in there turned out to be a **body** decision: nothing protocol-shaped touched the URL, headers, method or signing. So `build_body` owns the protocol, and `build_request` no longer knows protocols exist.

### What the caller now supplies

Everything taken from the downstream request is grouped under one key — its **presence** is what marks a call as proxying an inbound request:

```lua
opts.client = { headers, args, method }   -- proxy path only; internal calls omit it entirely
```

| opts | was |
| --- | --- |
| `client.headers` | `core.request.headers(ctx)` inside the transport |
| `client.method` / `client.args` | `core.request.get_method()` / `ctx.var.args` (passthrough) |
| `target_protocol` | `ctx.ai_target_protocol` |
| `header_transform` | `ctx.ai_converter.convert_headers` |
| `access_token` | `fetch_gcp_access_token(ctx, ...)` (ctx only keyed its cache) |

Protocol **conversion** stays with the caller: converters stash state on `ctx` for the response side to read back, so `ai-proxy` converts and hands over the converted body.

The raw-body reuse optimisation (#1597) is preserved but no longer implicit. The transport already sends a string body verbatim and encodes a table, so "reuse the client's bytes" is just the caller passing the string — one visible line at the call site instead of a contract (`raw_request_body` plus an internal `body_changed`) that the caller had to know `build_request`'s internals to use correctly.

### Call flow

**Proxy path**

```
client request
└─ nginx access phase  (apisix/init.lua)
   ├─ common_phase("access") → ai-proxy-multi._M.access
   │  ├─ base.detect_request_type(ctx)        → ctx.ai_client_protocol
   │  └─ pick_ai_instance()                   → ctx.picked_ai_instance
   │
   └─ common_phase("before_proxy") → ai-proxy/base.lua before_proxy
      ├─ extra_opts{ endpoint, auth, target_*, model_options, override_* }
      ├─ extra_opts.client = { headers, args, method }     ← downstream facts
      ├─ gcp                → extra_opts.access_token
      ├─ converter.convert_request(body, ctx)              ← conversion (stateful → caller)
      └─ do_request()
         ├─ provider:build_body(body, opts)  -> body, shaped     ← protocol shaping (pure)
         ├─ if not shaped and not converted and not ctx.ai_request_body_changed
         │     then body = core.request.get_body() end           ← reuse client bytes (explicit)
         ├─ provider:build_request(conf, body, opts) -> params   ← HTTP + SigV4 (pure)
         ├─ transport_http.request(params, conf.timeout)
         └─ parse_response / parse_streaming_response(ctx, …)    ← response side keeps ctx
```

**Internal call** (`ai-request-rewrite`)

```
ai-request-rewrite._M.access(conf, ctx)
├─ proto.build_simple_request(prompt, client_body)     → its own { messages }
├─ gcp → extra_opts.access_token
├─ provider:request(conf, request_table, extra_opts)   ← no ctx, no opts.client
│  ├─ build_body(request_table, opts)
│  ├─ build_request(conf, body, opts)
│  └─ transport_http.request() → (status, raw_body)    ← never touches parse_response
├─ proto.extract_response_text()
└─ ngx.req.set_body_data(content); ctx.ai_request_body_changed = true
      └─► later ai-proxy sees the flag → does not reuse the client's raw bytes
```

`ctx.ai_request_body_changed` stays: it is the cross-plugin "an earlier plugin rewrote the body" signal that the raw-body reuse optimisation depends on. It is now read by `ai-proxy` — which legitimately holds `ctx` — rather than by the transport.

### Logging

`redact_extra_opts()` drops the whole `client` key, so the downstream headers (`Authorization`/`Cookie`), query args and verbatim body (the user's prompt) never reach the info-level `request extra_opts` log line. Dropping one key drops all of it, instead of enumerating fields.

### Scope

Request side only. `parse_response` / `parse_streaming_response` keep `ctx` by design — there it is an output sink (`ngx.status`, `ctx.var.llm_*`) and the downstream plugin chain (`lua_response_filter` runs each plugin's `lua_body_filter` off `api_ctx.plugins`). Internal callers never reach that path; they take `(status, raw_body)` and parse it themselves.

Still deliberately left in place: `opts.header_transform`. `convert_headers` must run before SigV4 signs, and signing is last inside `build_request` — exposing `sign()` to callers so the hook could move out would trade a small wart for "someone forgets to sign".

### Tests

New `t/plugin/ai-transport-header-forwarding.t`: the internal rewrite call reaches the LLM with its **own** credentials and no client `Cookie`; the transparent proxy path still forwards it; the client's `Authorization` and prompt never appear in any log. Each assertion fails against the pre-fix code.

`build_request` unit tests retargeted at the new contract (body passed through untouched, string sent verbatim), plus a new one for `build_body`'s `changed` flag — the flag the reuse decision hangs on.

Existing suites covering the moved logic all pass: `ai-proxy-request-body-override` (raw-body reuse), `ai-proxy-protocol-conversion` (converter), `ai-proxy-passthrough` (method/query), `ai-proxy-bedrock` (SigV4 signs the finalized body last), `ai-proxy-vertex-ai` (GCP token), `ai-proxy-multi.balancer`.

### DIFF
--- a/apisix/plugins/ai-providers/base.lua
+++ b/apisix/plugins/ai-providers/base.lua
@@ -188,7 +188,12 @@ function _M.build_request(self, ctx, conf, request_body, opts)
             .. "includes a path", 400
     end
 
-    local headers = transport_http.construct_forward_headers(auth.header or {}, ctx)
+    -- Only the proxy path (ai-proxy / ai-proxy-multi) forwards the downstream
+    -- request's headers, and it passes them in via opts.client_headers. A
+    -- self-contained internal request (ai-request-rewrite, embeddings, ...)
+    -- passes none, so its own credentials — not the client's — reach the LLM.
+    local headers = transport_http.construct_forward_headers(auth.header or {},
+                                                             opts.client_headers)
     if opts.host_header then
         headers["Host"] = opts.host_header
     end
--- a/apisix/plugins/ai-providers/base.lua
+++ b/apisix/plugins/ai-providers/base.lua
@@ -98,37 +98,84 @@ local function merge_usage(ctx, parsed)
 end
 
 
---- Build HTTP request parameters from driver config and extra_opts.
+--- Shape the request body for the target protocol and this provider.
 --
--- This is a pure client: it never reads the downstream request. Everything taken
--- from that request is grouped under `opts.client` and supplied by the caller —
--- its presence is what marks this as proxying an inbound request. A
--- self-contained internal call (ai-request-rewrite, embeddings, ...) omits
--- `opts.client` entirely, so nothing of the client's can reach the LLM or the
--- logs. Request state the caller resolves for us arrives as opts.access_token /
--- opts.target_protocol / opts.header_transform.
---
--- Protocol conversion is deliberately NOT done here: converters carry state on
+-- Pure: no ctx, no downstream request, no network -- body in, body out. This is
+-- where the target protocol matters; build_request below does not know protocols
+-- exist. Protocol *conversion* is not done here either: converters carry state on
 -- the request ctx for the response side to read back, so the caller converts and
--- passes the already-converted body.
--- @return table params HTTP parameters ready for transport_http.request()
--- @return string|nil err Error message
-function _M.build_request(self, conf, request_body, opts)
-    local body_changed = false
+-- hands us the converted body.
+-- @return table body The body to send
+-- @return boolean changed Whether anything here modified it. The caller uses this
+--   to decide whether the client's verbatim bytes may still be reused.
+function _M.build_body(self, request_body, opts)
+    local changed = false
+    local target_protocol = opts.target_protocol
 
     -- Inject target-protocol-specific parameters (e.g. stream_options for OpenAI).
-    -- This runs after the caller's conversion so it covers both passthrough and
-    -- convert scenarios.
-    local target_protocol = opts.target_protocol
     if target_protocol then
         local target_proto = protocols.get(target_protocol)
         if target_proto and target_proto.prepare_outgoing_request then
             if target_proto.prepare_outgoing_request(request_body) then
-                body_changed = true
+                changed = true
             end
         end
     end
 
+    -- Inject model options (flat overwrite)
+    if opts.model_options then
+        for opt, val in pairs(opts.model_options) do
+            if request_body[opt] ~= nil then
+                core.log.info("model_options overwriting request field '", opt, "'")
+            end
+            request_body[opt] = val
+            changed = true
+        end
+    end
+
+    -- Apply llm_options via provider capability hook (always force-overwrites)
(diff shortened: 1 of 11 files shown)


membphis:
[P2] GCP token cache key can alias different credentials and TTL policies

The new cache key uses `ngx.crc32_long(service_account_json) .. "#" .. name`. Because it relies on a 32-bit CRC and omits `expire_early_secs` and `max_ttl`, separate configurations can alias: different credentials may reuse the wrong cached access token after a collision, while identical credentials and names with different TTL policies can bypass the latter policy.

Please use a collision-resistant digest of the resolved credentials and all fields that affect cache reuse, without logging plaintext secrets, and add coverage for same-name configurations with different TTL policies.

AlinsRan:
Fixed here + upstream apache/apisix#13699. Key is now `md5(sa \0 expire_early_secs \0 max_ttl) .. "#" .. name`: 128-bit so distinct creds cannot collide, TTL fields in the key so different policies do not alias, raw JSON stays out of the key/log.

Regression test (vertex-ai TEST 13): same name+creds with different `max_ttl` must not share a token — old key gave `calls=1 distinct=false`, fixed gives `calls=2 distinct=true`.

md5 follows the existing cache-key convention here; can move to SHA-256 if you prefer.

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13700, opened 2026-07-16
feat(stream): support set_real_ip_from to trust an inbound PROXY protocol header

nic-6443:
### Description

A stream port can accept the PROXY protocol (`proxy_protocol`) and send it on to the upstream (`proxy_protocol_to_upstream`), but the two don't compose into "pass the client address through". `ngx_proxy_protocol_write()` builds the outbound header from `c->sockaddr` — the directly connected peer — and never looks at `c->proxy_protocol`. So with a load balancer in front, the upstream is told the load balancer's address and the client's is lost. The stream `$remote_addr`, the access log and address matching such as `ip-restriction` see the load balancer for the same reason.

nginx's answer is `ngx_stream_realip_module`, but there was no way to configure it: `set_real_ip_from` is only rendered into the `http` block.

This adds `nginx_config.stream.real_ip_from`, mirroring the existing `nginx_config.http.real_ip_from`:

```yaml
apisix:
  proxy_mode: http&stream
  stream_proxy:
    tcp:
      - addr: 9100
        proxy_protocol: true
        proxy_protocol_to_upstream: true
nginx_config:
  stream:
    real_ip_from:
      - 192.168.1.0/24
```

On a connection from a trusted address that carries an inbound PROXY protocol header, the module writes the header's address into `c->sockaddr`, so the rebuilt outbound header — and every stream-side consumer of the client address — carries the client. The directly connected address stays available as `$realip_remote_addr`.

The option is empty by default, so nothing changes for existing configurations. It only takes effect on ports that accept the PROXY protocol, and only for peers it matches.

### Runtime requirement

`ngx_stream_realip_module` is not built by default, so `set_real_ip_from` is an unknown directive on runtimes older than 1.3.11. #13697 bumped `APISIX_RUNTIME` to 1.3.11, which adds `--with-stream_realip_module`; it has merged and master is merged into this branch, so CI here runs against 1.3.11.

### Docs

`docs/en/latest/stream-proxy.md` and its zh counterpart gain a section on preserving the client address behind a load balancer, plus a table of the four combinations of accepting and sending the PROXY protocol — including the one that silently loses the client address today.

### Tests

`t/cli/test_stream_proxy_protocol.sh` grows four cases. Two cover rendering: nothing is trusted by default, and each `real_ip_from` entry renders one `set_real_ip_from`. The assertions are scoped to the stream block, since the http block renders the same directive by default.

Two are end to end against a PROXY-protocol-aware stream upstream that echoes the address it was handed, so they assert on the header APISIX actually rebuilds rather than on `nginx.conf`:

- with the peer trusted, a client announcing `198.51.100.7` reaches the upstream as `198.51.100.7`;
- with nothing trusted, the same connection reaches the upstream as the direct peer, so an untrusted peer's claim is ignored.

All eleven cases in the file pass locally against apisix-runtime 1.3.11.

### Checklist


shreemaan-abhishek (approved):

membphis:
[P2] Please document the runtime prerequisite for nginx_config.stream.real_ip_from. The rendered set_real_ip_from directive requires ngx_stream_realip_module. APISIX Runtime 1.3.11 includes it, but stock/custom OpenResty builds without --with-stream_realip_module will fail with an unknown directive when this option is enabled. Please add this requirement to both stream-proxy documents and, ideally, the config example.

membphis (approved):
LGTM

AlinsRan (approved):

PR #13704, opened 2026-07-17
fix(limit-count): commit() must report the window-weighted remaining

nic-6443:
### Description

`commit()` in the sliding-window limiter reports `limit - current_count` as the remaining quota, while `incoming()` reports the window-weighted value (`limit - current_count - last_count * remaining_time / window`). Delayed sync (`sync_interval`) flushes locally-permitted deltas through `commit()` and caches the reported remaining as the global quota for the next interval, so at every window boundary each gateway instance starts again from a nearly full budget: the sliding window effectively degrades into a per-window fixed window.

The effect is easy to measure: 2 instances sharing one Redis, `count=500`, `time_window=1`, `window_type=sliding`, `sync_interval=0.2`, sustained 800 rps for 30s. Ideal admission is 500/s. Before this change the gateways admit ~590/s (+18% over the limit, bursty at each window start); with this change they admit ~438/s and converge, matching what the pre-#13443 flush path (which used `incoming()` for flushing) produced. The remaining gap below the limit is the inherent cost of the delayed-sync snapshot, not new behavior.

The fix reads the previous window counter after the increment and reports the same weighted remaining as `incoming()`. A failed read must not fail the commit — the delta is already recorded, and the delayed syncer would retry and double-count it — so it degrades to the unweighted remaining and logs instead.

The unconditional increment behavior from #13443 is unchanged: only the reported remaining is corrected.

Introduced in #13443.

### Checklist

- [ ] I have updated the documentation to reflect this change (no documented behavior changes)

### DIFF
--- a/apisix/plugins/limit-count/sliding-window/sliding-window.lua
+++ b/apisix/plugins/limit-count/sliding-window/sliding-window.lua
@@ -169,6 +169,7 @@ end
 function _M.commit(self, key, cost)
     local now = ngx_now()
     local counter_key = get_counter_key(self, key, now)
+    local last_counter_key = get_counter_key(self, key, now - self.window_size)
     local remaining_time = self.window_size - now % self.window_size
 
     local red_cli, err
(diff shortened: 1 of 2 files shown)


membphis (approved):
LGTM

membphis:
For Redis-backed sliding-window limits, we strongly recommend enabling delayed sync (`sync_interval`). It batches counter updates instead of accessing Redis on every request, greatly reducing Redis pressure and generally improving overall stability and performance. With delayed sync enabled, the additional previous-window lookup occurs only during synchronization, not on every request.

AlinsRan (approved):

shreemaan-abhishek (approved):

PR #13708, opened 2026-07-20
fix(prometheus): bump nginx-lua-prometheus-api7 to 0.20260716 to fix 100% CPU loop on full shared dict

### LINKED ISSUE #12275: help request: My APISIX is experiencing 100% CPU utilization and has become unresponsive
### Description

When the free_size of Prometheus in APISIX's share_dict becomes 0, will it trigger a 100% CPU exception? This issue occurred in 2 out of 38 instances. Even after removing the traffic, the CPU remains fully occupied. This is what was observed using perf tools and flamegraph.    
ngx_shmtx_lock  ngx_shmtx_unlock

Image
strace信息：
Image
perf top信息：

Image

We have mounted too many routes, so the metrics are very large. The currently allocated memory is 100M.
what should i do ? scale prometheus share dict?

### Environment

apisix :2.15
Linux 4 3.1-1160.92.1.el7.x86_64 #1 SMP Tue Jun 20 11:48:01 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
openresty/1.21.4.2

nic-6443:
### Description

Bumps `nginx-lua-prometheus-api7` to `0.20260716`, which fixes the infinite busy loop behind #12275 — workers pegged at 100% CPU on a full `prometheus-metrics` shared dict, unresponsive even after traffic is removed.

The root cause is in the library's `KeyIndex:add()`: once the shared dict is full, `__ngx_prom__key_count` can be LRU-evicted (it is an ordinary node, only refreshed when a *new* series is registered, so it goes cold under steady traffic). `sync()` then reads it as `nil → 0`, and every new series does `dict:add()` against an already-occupied slot → `"exists"` → retry the same slot → forever, without ever yielding. Full analysis, profiles and a standalone reproduction are in https://github.com/apache/apisix/issues/12275#issuecomment-4988690889.

The fix (api7/nginx-lua-prometheus#16, released as `0.20260716`) advances the counter past the occupied slot on a repeated collision, so the next `sync()` adopts that slot's occupant and progress resumes; a hard cap makes it degrade to the existing dict-full error instead of spinning. The regression test lives with the fix in the library repo.

I ran `t/plugin/prometheus*.t` locally against the bumped dependency and the prometheus suites pass. The only failures are the pre-existing `worker-events ... failed to receive the header bytes: closed` shutdown-race flake, which reproduces identically on the current `0.20260623` pin and is unrelated to this change.

Fixes #12275

### Checklist

- [x] I have added tests corresponding to this change (the regression test ships with the library fix in api7/nginx-lua-prometheus#16; this PR is the version bump that pulls it in)
- [ ] I have updated the documentation to reflect this change (dependency bump only, no doc change needed)


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13710, opened 2026-07-20
change(debug): report executed plugins with phase in execution order

nic-6443:
### Description

Currently the `Apisix-Plugins` debug header is built from a hash set filled at plugin *filtering* time, so the reported list is unordered (`pairs()` iteration) and carries no phase information — it tells you which plugins matched the route, not what actually ran, and the order doesn't match the execution order at all.

This PR changes each entry to `name#phase` and makes the header report the entries in the runtime execution order, e.g.:

```
Apisix-Plugins: limit-conn#access, limit-count#access, response-rewrite#header_filter, limit-conn#log
```

How the entries are collected:

- The phase functions executed before the response header is generated (`rewrite`, `access`, `header_filter`, ...) are traced at execution time in `run_plugin` (and `lua_response_filter`). A function that runs more than once (e.g. per-chunk, or re-runs on balancer retries) is recorded once, at its first execution. `rewrite_in_consumer` executions are reported as `rewrite`, since that's the actual function being run.
- The `body_filter`/`delayed_body_filter`/`log` phase functions run after the response header is sent, so they can not be traced into the header. They are **inferred** instead, right before the response header is generated: a matched plugin carrying such a phase function is reported as if it would execute it. The inferred entries may not fully match the real execution (e.g. a plugin skipped at runtime by its `_meta.filter` is still reported) — the doc covers this.
- The phases which can be neither traced nor inferred (e.g. the post-header phases of global rule plugins, whose filtered plugin list is not kept in the request ctx) are logged as an error log at a `warn` level, e.g. `Apisix-Plugins: response-rewrite#body_filter`. An entry already reported in the header is not logged again.

I also removed the doc claim that stream subsystem plugins are logged this way: the stream Lua VM never loads `debug.yaml` (`stream_init_worker` doesn't call `debug.init_worker()`), so that path has always been dead — making basic debug mode actually work for stream can be a separate discussion.

Since the order is now deterministic, the tests that had to parse the header into a hash to tolerate the random order can just assert the literal header value.

None

### Checklist


shreemaan-abhishek (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13714, opened 2026-07-20
fix(admin): reconcile plugins reload instead of trusting the broadcast

### LINKED ISSUE #13537: bug: plugins reload event may never reach the privileged agent, leaving timers of removed plugins running
### Current Behavior

While deflaking `t/plugin/log-rotate.t` TEST 4 (#13536) I found the flake is masking a real problem: after disabling log-rotate via `PUT /apisix/admin/plugins/reload`, the rotate timer in the privileged agent sometimes keeps running essentially forever — I observed rotation continuing for 55+ seconds after the reload returned, on unmodified master, roughly 1 in 6 runs on a loaded machine.

The original test never caught this because it counts `*__error.log` files and the count plateaus at `max_kept: 3` while rotation keeps cycling (rotate one in, delete the oldest), so the assertion passes even when the disable never took effect.

Two details point at event delivery rather than the unload path:

- comparing the timestamped rotated file-name *signature* (which changes on every rotation) shows rotation continuing across a 50+ second observation window after a successful-looking reload;
- re-sending the same reload request reliably stops the rotation, so `_M.destroy()` → `timers.unregister_timer()` works fine once the event actually arrives.

This looks like the same family as the `plugins-reload.t` TEST 1 timing flake (hardened in #13332): the reload event is broadcast asynchronously, and the privileged agent can apparently miss it entirely under load — not just receive it late. Any plugin that registers privileged-agent timers (log-rotate, server-info, etc.) would keep its timer alive after being removed from the config.

### Expected Behavior

A plugins reload should deterministically (eventually) reach every worker including the privileged agent, so timers of removed plugins stop.

### Steps to Reproduce

1. Patch `t/plugin/log-rotate.t` TEST 4 to compare the sorted rotated file-name set instead of the count, with a 50-iteration (~55s) observation window after the reload (no retry).
2. Run the file repeatedly on a busy machine: `prove -Itest-nginx/lib -I./ t/plugin/log-rotate.t` — every few runs the signature keeps changing for the whole window.

### Environment

- APISIX version: master (cb72d0eb6 / eb1af76f8)
- OS: Linux x86_64
- OpenResty: openresty/1.29.2.4

AlinsRan:
### Description

Fixes #13537

`PUT /apisix/admin/plugins/reload` returns `done` as soon as `events:post()` accepts the event, but that post carries no delivery guarantee. A process that is connecting or reconnecting to the events broker at that moment loses the event permanently, and nothing ever brings it back in line — `plugin.load()` has exactly two call sites and neither reconciles.

The privileged agent is where this hurts. It hosts the timers registered by plugins, so a lost reload leaves it running the timers of plugins that were just removed. In #13537 that shows up as log-rotate continuing to rotate long after it was disabled, with no way out except sending the reload again.

This adds a version counter in the `internal-status` shared dict:

- `post_reload_plugins()` increments it **before** broadcasting, so even a completely lost event still leaves a record that a reload was requested
- every process that registers the reload handler compares the shared version against the one it applied, once a second, and calls `reload_plugins()` when they differ

The broadcast stays as the sub-second fast path; the counter is the slow path that actually guarantees convergence (≤1s). This is the same shape `admin/standalone.lua` already uses, for the same underlying reason — its comment spells it out.

On the details worth checking:

- **`internal-status` is a safe choice.** It is declared unconditionally for the http subsystem in `ngx_tpl.lua` and is already a hard dependency of the server-info plugin. The privileged agent is an http process and sees it. The key is its own, so there is no collision with server-info.
- **`ngx.timer.every` rather than `timers.lua`.** The background timer runs all registered entries through `thread_spawn`/`thread_wait`, so one slow entry would drag reconciliation with it; and `admin.init_worker()` runs before `timers.init_worker()`, so a self-contained timer is simpler.
- **Idempotence.** `plugin.load()` unloads everything and re-inits, and log-rotate's `destroy()`/`init()` register and unregister its timer symmetrically, so a redundant round costs nothing. Steady-state cost is one shdict read per process per second.
- **No spurious reloads.** Cold start: the dict is empty, both sides are 0. HUP reload: the dict survives, and a fresh worker adopts the current version at `init_worker`, so it does not reload just for being new.
- **Concurrent reloads.** The version is sampled *before* `plugin.load()`, so a reload accepted while load is running leaves the versions unequal and gets applied on the next tick rather than being swallowed.
- **`plugin.load()` throwing** (say a broken `config.yaml`) aborts the timer callback without recording the version, so it retries next tick and keeps logging. Deliberately not swallowed — a transient failure should not permanently skip a reload.

One API nuance to note: a failed `events:post` still returns 503, which is left as is. Since the increment already happened, that 503 now means "broadcast failed but the reload will still take effect within a second" rather than "nothing happened". Retrying the PUT is idempotent either way.

Two gaps are deliberately left open, both pre-existing: the stream subsystem shares neither the shared dict nor the events socket with http (`admin/standalone.lua` acknowledges the same limitation), and propagation between instances still goes through etcd `/plugins`.

### Tests

New `t/admin/plugins-reload-reconcile.t`:

- TEST 1 — a reload request bumps the shared version
- TEST 2 — a process whose applied version is behind reloads within the reconciliation interval, reproducing the "broadcast never arrived" state directly rather than through the events layer
- TEST 3 — an unchanged version never triggers a reload

Verified discriminating: without the change TEST 1 reports `bumped=false` and TEST 2 sees no reload; TEST 3 passes either way. Ran the file five times to confirm it is not timing-flaky.

### DIFF
--- a/apisix/admin/init.lua
+++ b/apisix/admin/init.lua
@@ -27,11 +27,20 @@ local ngx = ngx
 local get_method = ngx.req.get_method
 local ngx_time = ngx.time
 local ngx_timer_at = ngx.timer.at
+local ngx_timer_every = ngx.timer.every
 local ngx_worker_id = ngx.worker.id
 local tonumber = tonumber
 local tostring = tostring
 local str_lower = string.lower
 local reload_event = "/apisix/admin/plugins/reload"
+local is_http = ngx.config.subsystem == "http"
+-- declared unconditionally for the http subsystem in ngx_tpl.lua, and already a
+-- hard dependency of the server-info plugin; absent in the stream subsystem
+local plugins_conf_ver_dict = is_http and ngx.shared["internal-status"]
+local PLUGINS_CONF_VERSION_KEY = "plugins_conf_version"
+-- plugins conf version this process has applied, compared against the shared
+-- dict by the reconciliation timer registered in init_worker()
+local applied_plugins_conf_version = 0
 local ipairs = ipairs
 local error = error
 local type = type
@@ -288,6 +297,17 @@ end
 local function post_reload_plugins()
     set_ctx_and_check_token()
 
+    if plugins_conf_ver_dict then
+        -- bump the version before broadcasting, so that a process which never
+        -- receives the event (e.g. the privileged agent while it is
+        -- reconnecting to the events broker) still converges through the
+        -- periodic reconciliation below
+        local _, err = plugins_conf_ver_dict:incr(PLUGINS_CONF_VERSION_KEY, 1, 0)
+        if err then
+            core.log.error("failed to increase plugins conf version: ", err)
+        end
+    end
+
     local success, err = events:post(reload_event, get_method(), ngx_time())
     if not success then
         core.response.exit(503, err)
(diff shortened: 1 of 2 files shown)


membphis (approved):
LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13717, opened 2026-07-20
fix(config_etcd): keep the previous value when a full reload gets invalid data

### LINKED ISSUE #12834: bug: When etcd is compacted, apisix will use load_full_data(). If there is a configuration error in an individual plugin in the global plugin, it will cause all global plugins to become invalid
### Current Behavior

When etcd is compacted, apisix will use load_full_data(). If there is a configuration error in an individual plugin(For example, proxy-cache) in the global plugin, it will cause all global plugins to become invalid

It was found from etcd_config.lua that:
`
function sync_data(self)
    ……
    if self.need_reload then
        ……
        self.values = nil
        self.values_hash = nil

`

### Expected Behavior

1. Among the global plugins, the error reports of individual plugins should not affect the configuration and update of other plugins worldwide
2. If the plugin is "disable": true, the incorrect configuration of this plugin should not affect the configuration and update of other plugins

### Error Logs

1. The global plugin proxy-cache is configured incorrectly
`2025/12/18 17:06:27 [error] 53#53: *142160819 [lua] config_etcd.lua:679: sync_data(): failed to check item data of [/apisix/global_rules] err:failed to check the configuration of plugin proxy-cache err: cache_zone disk_cache_one not found ,val: {"update_time":1766048710,"create_time":1756868235,"plugins":{"proxy-cache":{"hide_cache_headers":false,"cache_control":false,"cache_ttl":300,"cache_strategy":"disk","cache_method":["GET","HEAD"],"cache_http_status":[200,301,404],"cache_zone":"disk_cache_one","cache_key":["$host","$request_uri"],"_meta":{"disable":true}},……}`

2. etcd is compacted，apisix watch canceled by etcd
`2025/12/19 13:14:13 [warn] 57#57: *156949651 [lua] config_etcd.lua:227: watch canceled by etcd, res: {
  result = {
    canceled = true,
    compact_revision = "8023",
    header = {
      cluster_id = "10063310312338577470",
      member_id = "8355680492992881584",
      raft_term = "4"
    }
  }
}, context: ngx.timer
2025/12/19 13:14:13 [error] 57#57: *156949651 [lua] config_etcd.lua:230: etcd compacted, compact_revision=8023, context: ngx.timer`

3. apisix load_full_data() ，Meanwhile reports a format error of the global plugin proxy-cache
`2025/12/19 13:14:13 [error] 57#57: *156823782 [lua] config_etcd.lua:543: load_full_data(): failed to check item data of [/apisix/global_rules] err:failed to check the configuration of plugin proxy-cache err: cache_zone disk_cache_one not found ,val: {"id":"1","plugins":{"proxy-cache":{"_meta":{"disable":true},"cache_http_status":[200,301,404],"cache_method":["GET","HEAD"],"cache_key":["$host","$request_uri"],"cache_control":false,"hide_cache_headers":false,"cache_zone":"disk_cache_one","cache_strategy":"disk","cache_ttl":300},……}`

4. All global plugins are invalid

### Steps to Reproduce

1. The proxy-cache plugin was configured incorrectly. You can refer to the configuration in the error log
2. Wait for etcd compact and apisix to handle the compact event, and execute load_full_data()

### Environment

- APISIX version (run `apisix version`): 3.10
- Operating system (run `uname -a`): debian 11.10, 4.19.0-10-amd64
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): openresty/1.25.3.2
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 3.5.16
- APISIX Dashboard version, if relevant: 3.0.1

AlinsRan:
### Description

Fixes #12834

The incremental watch path and the full reload path in `core/config_etcd.lua` disagree on invalid data:

- **watch** (`sync_data`): `goto CONTINUE` — cancels only that item's update, the value in memory keeps serving (deliberate, #11268).
- **full reload** (`load_full_data`): the old table is dropped up front and the invalid item is simply not re-inserted, so it disappears.

A full reload is triggered asynchronously when etcd cancels the watch with `compacted`. So an item that validation rejects can serve for days, then vanish the moment an unrelated compaction happens. Because global plugins usually share one global rule, one invalid plugin conf takes down every global plugin — with no visible cause.

Reporter's case: a `proxy-cache` conf with the default `cache_zone: disk_cache_one`, on a node whose `config.yaml` declares different zones. `check_schema` validates `cache_zone` against the local node config, so the item is permanently invalid on that node.

### Changes

**Commit 1 — make full reload behave like the watch path.** On validation failure `load_full_data` now carries the previous value over as is (no clean handlers, no `filter` re-run, no `conf_version` bump), instead of dropping it. It still skips an invalid item with no previous value (first load), still drops keys absent from etcd (no resurrection), and fires the clean handlers of replaced/deleted items after the new table is built. This is the same semantics the watch path has had since 3.10.0 — no new behavior, only removing "did a compaction happen to occur" as a factor.

**Commit 2 — remove the amplifier.** A plugin disabled via `_meta.disable: true` is never executed, so an environment-dependent validation failure should not invalidate the item that carries it. `check_schema` still runs (its default injection is wanted) but the failure is fatal only when the plugin is enabled; otherwise it is logged and accepted. Re-enabling goes through a normal fully-validated write, so an invalid conf can't silently become active. `stream_check_schema` gets the same treatment. This also relaxes the Admin API (shared `check_single_plugin_schema`), which is intended given heterogeneous `config.yaml` across data planes.

### Tests

- `t/core/config_etcd.t`: full reload keeps the previous value on invalid data (TEST 16), does not resurrect a deleted item (TEST 17), still skips an invalid item with no previous value (TEST 18). All drive the real sync loop against etcd.
- `t/plugin/proxy-cache/disk.t`: Admin `PUT` of a disabled `proxy-cache` with a non-existent zone returns 200; the same conf enabled still returns 400.

### Checklist

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -546,9 +546,24 @@ local function sync_status_to_shdict(status)
 end
 
 
-local function load_full_data(self, dir_res, headers)
+local function get_prev_item(prev_values, prev_values_hash, key)
+    if not prev_values or not prev_values_hash then
+        return nil
+    end
+
+    -- deleted items are tombstoned as `false` in the values array and removed
+    -- from the hash by the watch path, so a hash hit is always a live item
+    local idx = prev_values_hash[key]
+    return idx and prev_values[idx] or nil
+end
+
+
+local function load_full_data(self, dir_res, headers, prev_values, prev_values_hash)
     local err
     local changed = false
+    -- previous items carried over because their new data failed the check;
+    -- their clean handlers must NOT be fired
+    local carried = {}
 
     if self.single_item then
         self.values = new_tab(1, 0)
@@ -583,6 +598,19 @@ local function load_full_data(self, dir_res, headers)
             if self.filter then
                 self.filter(item)
             end
+
+        elseif item.value ~= nil then
+            -- new data exists but is invalid: keep the previous value like the
+            -- incremental watch path does. An absent value (deleted key) must
+            -- not be resurrected, hence the `item.value ~= nil` guard.
+            local prev_item = get_prev_item(prev_values, prev_values_hash, self.key)
+            if prev_item then
+                log.warn("failed to check item data of [", self.key,
+                         "], keep the previous configuration, err: ", err)
+                insert_tab(self.values, prev_item)
+                self.values_hash[self.key] = #self.values
+                carried[prev_item] = true
+            end
         end
 
         self:upgrade_version(item.modifiedIndex)
@@ -601,8 +629,10 @@ local function load_full_data(self, dir_res, headers)
         for _, item in ipairs(values) do
             local key = short_key(self, item.key)
             local data_valid = true
+            err = nil
             if type(item.value) ~= "table" then
                 data_valid = false
+                err = "invalid item data, it should be an object"
                 log.error("invalid item data of [", self.key .. "/" .. key,
                           "], val: ", item.value,
                           ", it should be an object")
@@ -637,9 +667,32 @@ local function load_full_data(self, dir_res, headers)
                 if self.filter then
                     self.filter(item)
                 end
+
+            else
+                local prev_item = get_prev_item(prev_values, prev_values_hash, key)
+                if prev_item then
+                    -- keep serving with the last valid configuration instead of
+                    -- silently dropping the whole item on a full reload, see the
+                    -- incremental path in sync_data for the same semantics
+                    log.warn("failed to check item data of [", self.key, "/", key,
+                             "], keep the previous configuration, err: ", err)
+                    insert_tab(self.values, prev_item)
+                    self.values_hash[key] = #self.values
+                    carried[prev_item] = true
+                end
             end
 
             self:upgrade_version(item.modifiedIndex)
+        end
+    end
+
+    -- fire the clean handlers of the previous items that were not carried
+    -- over: they were either replaced by a new value or deleted from etcd
+    if prev_values then
+        for _, item in ipairs(prev_values) do
+            if item and not carried[item] then
+                config_util.fire_all_clean_handlers(item)
+            end
         end
     end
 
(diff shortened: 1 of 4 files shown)


membphis (approved):
LGTM

apisix/core/config_etcd.lua
```diff
@@ -637,12 +667,35 @@ local function load_full_data(self, dir_res, headers)
...
 
+    -- fire the clean handlers of the previous items that were not carried
+    -- over: they were either replaced by a new value or deleted from etcd
+    if prev_values then
+        for _, item in ipairs(prev_values) do
+            if item and not carried[item] then
+                config_util.fire_all_clean_handlers(item)
+            end
+        end
+    end
```
nic-6443, line 697: The entire clean handler should be removable after https://github.com/apache/apisix/pull/12426

AlinsRan: You're right that clean handlers have no registrant after #12426 (health checks moved to `healthcheck_manager`), so this is effectively a no-op.

I've kept it here for consistency: `config_yaml`/`config_xds` full reloads fire on replaced/deleted items the same way, and dropping it only in the etcd full-reload path would leave a confusing asymmetry. Removing the whole `clean_handlers` mechanism uniformly (the `config_util` API plus all fire sites across the config backends) is broader than this bugfix, so I'll do it in a dedicated follow-up PR rather than entangle it with a backportable fix.

nic-6443 (approved):

shreemaan-abhishek (approved):

apisix/core/config_etcd.lua
AlinsRan: Done in #13761 — removes `add_clean_handler` / `cancel_clean_handler` / `fire_all_clean_handlers` and all fire sites and `clean_handlers` initializations across the config backends.

PR #13726, opened 2026-07-22
feat: support http_method and http_req_body in active health checks

AlinsRan:
### Description

The active health-check probe is currently hard-coded to a bodyless `GET`. This PR bumps `lua-resty-healthcheck-api7` to `3.2.3` and exposes two new fields on `upstream.checks.active` so a probe can be sent as any HTTP method with a request body:

- `http_method` — default `GET`
- `http_req_body` — default `""`; a non-empty body adds a matching `Content-Length` header

This lets an upstream be probed with a real request instead of a bare GET — for example, health-checking an LLM provider through `ai-proxy-multi` with an actual chat-completion `POST` body.

Defaults keep the previous `GET`-without-body behavior, so it is backward compatible.

### Changes

- `apisix-master-0.rockspec`: `lua-resty-healthcheck-api7` `3.2.2-0` → `3.2.3-0`
- `apisix/schema_def.lua`: add `http_method` / `http_req_body` to `health_checker_active` (moved `method_schema` up so the enum can be reused)
- `docs/.../health-check.md`: document the two fields
- `t/plugin/ai-proxy-multi3.t`: add a test that configures a POST-body probe and asserts the target receives the expected method / headers / body
- `t/plugin/ai-proxy-vertex-ai.t`: 3.2.3 changed the probe debug log label from `request head:` to `request:`; update the assertion

### Notes

`3.2.3` is a backward-compatible superset of `3.2.2` (the probe request is byte-identical for the default GET-without-body case; the only behavioral change for existing users is the debug log label).


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13729, opened 2026-07-23
fix(admin): encrypt only after successful validation

shreemaan-abhishek:
### Description

`resource.check_conf` calls the per-resource `encrypt_conf` hook **before** checking whether `self.checker(...)` succeeded. For invalid input, the encryption hooks therefore run against unvalidated structures: for example, a non-object `plugins` value reaches `pairs(plugins_conf)` in `admin/plugins.lua` and raises a Lua error, turning a clean **400** schema-validation error into a **500**.

This moves the `encrypt_conf` call below the `if not ok then return` so encryption runs only on validated config (invalid configs are never persisted anyway).

Also drops the `plugin_object.check_schema` gate in `plugin_metadata.encrypt_conf`: `plugin.encrypt_conf` derives fields from `metadata_schema.encrypt_fields` and does not need a custom validator, so a plugin that validates via `core.schema.check` would otherwise skip encryption and persist its metadata secret in plaintext.

Both were introduced in #12603.

### Tests

New `t/admin/encrypt-after-validation.t`:
- a non-object `plugins` value now returns **400** (previously a 500 from `bad argument #1 to 'pairs' (table expected, got string)`);
- a valid consumer with a `key-auth` secret is still stored as ciphertext at rest when `data_encryption` is enabled, confirming encryption still runs on the success path.

### Checklist

- [x] I have updated the documentation to reflect this change (N/A - no user-facing behavior/doc change)
- [x] I have verified that the code passes the existing lint (`luacheck`)

### DIFF
--- a/apisix/admin/plugin_metadata.lua
+++ b/apisix/admin/plugin_metadata.lua
@@ -94,7 +94,10 @@ local function encrypt_conf(plugin_name, conf)
     inject_metadata_schema(plugin_object)
 
     local schema = plugin_object.metadata_schema
-    if schema['$comment'] ~= injected_mark and plugin_object.check_schema then
+    -- encrypt whenever the plugin has a real metadata schema; do not gate on
+    -- check_schema, or a plugin that validates via core.schema.check would leak
+    -- its metadata encrypt_fields in plaintext
+    if schema['$comment'] ~= injected_mark then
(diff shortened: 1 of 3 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

AlinsRan (approved):

PR #13735, opened 2026-07-23
feat(ai-aws-content-moderation): moderate LLM responses, including streams

shreemaan-abhishek:
### Description

`ai-aws-content-moderation` only moderated requests, while its sibling `ai-aliyun-content-moderation` moderates both directions. AWS Comprehend `DetectToxicContent` scores response text just as well, so this was a plugin gap rather than an API limitation. It builds on #13647, which moved the plugin to the `access` phase with protocol-aware content extraction.

**What changed** — a new `check_response` option plus a `lua_body_filter`:

| Case | Behavior |
|---|---|
| non-streaming (`ai_chat`) | the assembled completion is moderated in one call; a hit returns `deny_code` with a provider-compatible refusal body |
| streaming, `stream_check_mode: realtime` | batches are moderated as they arrive, bounded by `stream_check_cache_size` (chars) and `stream_check_interval` (seconds); a hit replaces the rest of the stream with the refusal |
| streaming, `stream_check_mode: final_packet` (default) | the assembled response is moderated once and the last chunk's SSE data events are annotated with `risk_level` |

New schema fields: `check_response` (default `false`), `stream_check_mode` (default `final_packet`), `stream_check_cache_size` (default `128`), `stream_check_interval` (default `3`). All default to the previous behavior, so existing configurations are unaffected.

The verdict is published on `ctx.var.llm_content_risk_level` as `high` or `none`. Comprehend grades nothing, it only scores against the configured thresholds, so the verdict is binary; reusing the same ctx var the aliyun plugin sets keeps logging and downstream consumers uniform.

Deny reasons now name the side that tripped (`request body exceeds ...` / `response body exceeds ...`).

A Comprehend failure on the **response** side is logged and the content passes through: once bytes are on the wire there is no fail-closed option, and buffered responses behave the same way for consistency. This matches `ai-aliyun-content-moderation`. The request side is unchanged and still fails closed with a 500.

**Tests** — `t/plugin/ai-aws-content-moderation.t` TEST 18-30 cover non-streaming deny and pass-through, the provider-compatible deny body with a custom `deny_message` at the default `deny_code`, `check_response` off leaving a toxic response untouched, `final_packet` annotation, `realtime` cutting the stream off mid-flight, the upstream-error skip, and schema validation of the new fields. New fixtures live under `t/fixtures/aws/`. The Comprehend mock gained a fallback for free-form response text, since response content is not a fixture key; exact-key lookup still runs first, so every existing test is unchanged.

**Docs** — `docs/en/latest/plugins/ai-aws-content-moderation.md` updated with the new attributes and the response-moderation description.

# Checklist

### DIFF
--- a/apisix/plugins/ai-aws-content-moderation.lua
+++ b/apisix/plugins/ai-aws-content-moderation.lua
@@ -19,12 +19,14 @@ require("resty.aws.config") -- to read env vars before initing aws module
 local core      = require("apisix.core")
 local protocols = require("apisix.plugins.ai-protocols")
 local binding   = require("apisix.plugins.ai-protocols.binding")
+local sse       = require("apisix.plugins.ai-transport.sse")
 local aws       = require("resty.aws")
 local aws_instance
 
 local http = require("resty.http")
 
 local ngx     = ngx
+local ngx_ok  = ngx.OK
 local pairs   = pairs
 local unpack  = unpack
 local type    = type
@@ -73,6 +75,27 @@ local schema = {
             default = 0.5
         },
         check_request = { type = "boolean", default = true },
+        check_response = { type = "boolean", default = false },
+        stream_check_mode = {
+            type = "string",
+            enum = { "realtime", "final_packet" },
+            default = "final_packet",
+            description = "realtime: moderate batches while the response streams, replacing " ..
+                          "the rest of the stream on a hit | final_packet: moderate the " ..
+                          "assembled response and annotate the last chunk with risk_level.",
+        },
+        stream_check_cache_size = {
+            type = "integer",
+            minimum = 1,
+            default = 128,
+            description = "max characters per moderation batch in realtime mode",
+        },
+        stream_check_interval = {
+            type = "number",
+            minimum = 0.1,
+            default = 3,
+            description = "seconds between batch checks in realtime mode",
+        },
         deny_code = {
             type = "integer",
             minimum = 200,
@@ -103,10 +126,23 @@ function _M.check_schema(conf)
 end
 
 
+-- Comprehend scores content, it doesn't grade it, so the verdict is binary
+-- against the configured thresholds. Report it through the same ctx var the
+-- aliyun plugin uses so logging and downstream consumers stay uniform.
+local function set_risk_level(ctx, flagged)
+    if flagged then
+        ctx.var.llm_content_risk_level = "high"
+    elseif ctx.var.llm_content_risk_level ~= "high" then
+        ctx.var.llm_content_risk_level = "none"
+    end
+end
+
+
 -- Score content with AWS Comprehend detectToxicContent.
+-- `subject` names the moderated text in the deny reason ("request"/"response" body).
 -- Returns (reason, nil) when a category/toxicity threshold is exceeded,
 -- (nil, err) on a service error, and (nil, nil) when the content is clean.
-local function detect_toxic(conf, content)
+local function detect_toxic(conf, ctx, content, subject)
     local comprehend = conf.comprehend
 
     if not aws_instance then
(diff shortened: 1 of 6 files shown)


apisix/plugins/ai-aws-content-moderation.lua
```diff
@@ -257,4 +357,62 @@ function _M.access(conf, ctx)
...
+            ctx.aws_cm_response_moderated = true
+            moderate_response(ctx, conf, ctx.var.llm_response_text)
+        end
+        return nil, annotate_stream(ctx, body)
+    end
+
+    -- realtime: moderate batches as they arrive so a hit can cut the stream off
+    ctx.aws_cm_cache = ctx.aws_cm_cache or ""
+    ctx.aws_cm_cache = ctx.aws_cm_cache
+                       .. table.concat(ctx.llm_response_contents_in_chunk or {}, "")
```
AlinsRan, line 405: `ctx.llm_response_contents_in_chunk` is reset once per *upstream* chunk in `ai-providers/base.lua`, but when a converter is active that same loop calls `lua_response_filter` once per *converted* chunk. For an upstream chunk that yields N converted chunks, this line appends the identical text to `aws_cm_cache` N times: the batch sent to Comprehend contains duplicated content, and `stream_check_cache_size` trips N times sooner.

Nothing is missed, but the request volume scales with the converter's fan-out, which shows up as latency and billing on any cross-protocol route (e.g. Anthropic client over an OpenAI upstream).

`ai-aliyun-content-moderation` has the same shape, so this can be a separate fix — but it should be a conscious one rather than inherited.

apisix/plugins/ai-aws-content-moderation.lua
```diff
@@ -257,4 +357,62 @@ function _M.access(conf, ctx)
...
+    if ngx.status >= 400 then
+        core.log.info("skip response check because upstream returned error status: ", ngx.status)
+        return
+    end
+
+    local request_type = ctx.var.request_type
+
+    -- ai-proxy hands us the fully assembled completion, so one check covers it.
+    if request_type == "ai_chat" then
+        return moderate_response(ctx, conf, ctx.var.llm_response_text)
```
AlinsRan: On this path `moderate_response` swallows a Comprehend failure and lets the completion through. For streaming that is forced — bytes are already on the wire — but here the body has not been sent yet, so fail-closed is still reachable.

As written, `fail_mode: deny` applies to the request direction only, while the schema presents it as a plugin-wide setting. Either honour `fail_mode` for buffered responses, or document that it is request-side only; right now the option reads stricter than it behaves.

apisix/plugins/ai-aws-content-moderation.lua
shreemaan-abhishek: Good catch, confirmed. When a converter fans one upstream chunk into N downstream chunks, `lua_body_filter` runs N times and re-reads the same `llm_response_contents_in_chunk`, so the realtime cache double-counts and `stream_check_cache_size` trips early. Moderation still catches everything, but Comprehend request volume scales with fan-out. `ai-aliyun-content-moderation` has the same shape. Tracked in api7/api7-ee-3-gateway#2053 and will be fixed in a follow-up PR across both plugins and both repos.

apisix/plugins/ai-aws-content-moderation.lua
shreemaan-abhishek: Fixed in 1ca4d79bc. The buffered (non-streaming) response path now fails closed with a 500 on a Comprehend error, mirroring the request side, since the body has not been sent yet. Streaming stays best-effort by necessity (once bytes are on the wire the response cannot be blocked), and the docs now spell out both. I kept `fail_mode` as-is: it governs the unsupported-protocol / no-ai-proxy case, not moderation-service outages, so this is really a request/response asymmetry rather than `fail_mode` scope.

One note on syncing this to `ai-aliyun-content-moderation`: unlike AWS, aliyun already fails open on *both* directions on a service error, so it is internally symmetric. Making only its response side fail closed would introduce the reverse asymmetry, so I left it out of this change. Happy to align both plugins to fail-closed as a separate, deliberate decision if you prefer that.

AlinsRan (approved):

membphis (approved):
LGTM

guoqqqi (approved):

PR #13743, opened 2026-07-24
fix(healthcheck): probe domain nodes by node domain, not resolved ip

AlinsRan:
### Description

With the default `pass_host`, an active health check on a **domain-based HTTPS upstream always marks the node unhealthy**, so the whole upstream is taken out of service and never recovers. The active probe sends the TLS **SNI as the resolved ip**, so the handshake fails against a backend whose certificate is issued for the domain. The same root cause affects plain HTTP: the probe sends `Host: <resolved-ip>`, hitting the wrong virtual host and getting a misleading result.

#### Root cause

An active health check runs independently of any request — there is no downstream client. **The gateway itself is the client** probing the node, so the probe must address the node by the node's own identity: its configured **domain**. `pass_host` is a *request-forwarding* rule (how a client request's Host is passed to the upstream); it has no meaning for a client-less probe.

`compute_targets()` in `apisix/healthcheck_manager.lua` nonetheless derives the probe Host header by reusing that request-path logic:

```lua
local use_node_hdr = up_conf.pass_host == "node" or nil
local host_hdr = up_hdr or (use_node_hdr and node.domain) or nil
```

For a domain node with the default `pass_host` (`pass`) this yields a `nil` Host header, and the health-check library then falls back to `hostname`, which is the **resolved ip** for a domain node. Both the probe `Host` header and the HTTPS `SNI` derive from this value, so both become the ip.

Real traffic is unaffected: for `pass_host = node`/`rewrite` the request Host/SNI is `node.domain`/`upstream_host`, and for `pass` the real client's Host is forwarded. Only the client-less health checker degrades to the ip.

### Fix

```lua
local host_hdr = up_hdr or node.domain
```

Always carry `node.domain` as the probe Host header, except when `pass_host` is `rewrite` (which still pins `upstream_host` via `up_hdr`). The probe then addresses the node by its own identity, independent of `pass_host`. `node.domain` is `nil` for ip-based nodes, so ip upstreams are unaffected.

Per `pass_host` mode:

| `pass_host` | before | after |
|---|---|---|
| `node` | `node.domain` | `node.domain` (unchanged) |
| `rewrite` | `upstream_host` | `upstream_host` (unchanged) |
| `pass` (default), domain node | `nil` → probe uses resolved **ip** | `node.domain` |
| any, ip node | `nil` | `nil` (unchanged) |

### Tests

`t/node/healthcheck-node-domain-host.t`, both with a domain node (`test.com`, resolved locally to `127.0.0.1`) and the default `pass_host`:
- **TEST 2 (HTTPS SNI)** — the failing case above: an `https` upstream + `type: https` active check; the TLS probe target logs `$ssl_server_name`. Asserts the probe's SNI is `test.com`, not `127.0.0.1`.
- **TEST 1 (HTTP Host)**: the probe target logs the received `Host` header. Asserts the probe sends `Host: test.com`, not `Host: 127.0.0.1`.

### Checklist

- [ ] I have updated the documentation to reflect this change (no user-facing config or documented behavior change; this is an internal probe-addressing fix)
- [x] I have verified that the change follows the existing code style

### DIFF
--- a/apisix/healthcheck_manager.lua
+++ b/apisix/healthcheck_manager.lua
@@ -57,11 +57,15 @@ local function compute_targets(up_conf)
     local host = up_conf.checks and up_conf.checks.active and up_conf.checks.active.host
     local port = up_conf.checks and up_conf.checks.active and up_conf.checks.active.port
     local up_hdr = up_conf.pass_host == "rewrite" and up_conf.upstream_host
-    local use_node_hdr = up_conf.pass_host == "node" or nil
 
     local targets = {}
     for _, node in ipairs(up_conf.nodes) do
-        local host_hdr = up_hdr or (use_node_hdr and node.domain) or nil
+        -- A health check has no client; the gateway probes the node by the node's own
+        -- identity (its domain), so use node.domain as the Host header regardless of
+        -- pass_host (except "rewrite", which pins an explicit Host). Otherwise a domain
+        -- node probes with the resolved ip as Host/SNI, breaking HTTPS health checks.
+        -- node.domain is nil for ip nodes.
+        local host_hdr = up_hdr or node.domain
         local target_port = port or node.port
         -- add_target defaults the hostname to the ip when checks.active.host is
         -- unset, so mirror that here to match the shm entry's stored hostname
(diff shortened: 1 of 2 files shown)


shreemaan-abhishek (approved):

nic-6443 (approved):

membphis (approved):
LGTM

PR #13745, opened 2026-07-24
fix(control): reconcile /v1/plugins/reload like the admin reload path

### LINKED ISSUE #13537: bug: plugins reload event may never reach the privileged agent, leaving timers of removed plugins running

AlinsRan:
### Description

Follow-up to #13714 (which fixed #13537 for the admin reload path).

#13714 made the **admin** plugins reload (`PUT /apisix/admin/plugins/reload`) resilient to the `resty.events` broker's lack of a delivery guarantee: it bumps a shared `plugins_conf_version` in `ngx.shared["internal-status"]` and a per-worker reconciliation timer reloads any worker whose applied version is behind. That way a worker that misses the broadcast (e.g. while it is reconnecting to the broker) still converges.

The **control** reload path (`PUT /v1/plugins/reload`) has the identical exposure but was left out of #13714 — `post_reload_plugins()` only calls `events:post()`. So a worker can miss the control broadcast and keep running the plugins/timers of a removed plugin (exactly the #13537 symptom) even though the API returned `200`.

Note this is not a `lua-resty-events` bug to fix here — the broker has no replay for a reconnecting client by design; the reconciliation is the intended application-level compensation. This PR simply applies the existing compensation to the control path.

### Fix

`apisix/control/v1.lua` `post_reload_plugins()` now bumps the same `plugins_conf_version` before broadcasting, so the reconciliation timer registered by the admin path converges any worker that missed the event. When the admin is disabled the timer is absent and the bump is a no-op, leaving the broadcast as the only path (unchanged).

### Tests

`t/control/plugins-reload-reconcile.t` (mirrors `t/admin/plugins-reload-reconcile.t`):
- **TEST 1**: `PUT /v1/plugins/reload` bumps `plugins_conf_version`.
- **TEST 2**: reproduces a *missed* broadcast by bumping the shared version directly (without going through the events layer) and asserts the reconciliation timer reloads the worker — i.e. a worker that missed the event converges.

### Checklist

- [ ] I have updated the documentation to reflect this change (no user-facing config/behavior change)
- [x] I have verified that the change follows the existing code style

### DIFF
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -30,6 +30,11 @@ local str_format = string.format
 local ngx = ngx
 local ngx_var = ngx.var
 local events = require("apisix.events")
+-- Shared with apisix/admin/init.lua: the admin reload path bumps this version and
+-- runs a periodic reconciliation timer that reloads any worker whose applied
+-- version is behind. Keep the dict name and key in sync with admin/init.lua.
+local plugins_conf_ver_dict = ngx.shared["internal-status"]
+local PLUGINS_CONF_VERSION_KEY = "plugins_conf_version"
 
 
 local _M = {}
@@ -409,6 +414,19 @@ function _M.dump_plugin_metadata()
 end
 
 function _M.post_reload_plugins()
+    -- Bump the shared version before broadcasting so that a worker which misses the
+    -- event (the resty.events broker gives no delivery guarantee while a worker is
+    -- reconnecting) still converges through the admin reconciliation timer. This is
+    -- the same guard the admin reload path added in #13714; the control path was
+    -- left out. When the admin is disabled the timer is absent and this is a no-op.
+    if plugins_conf_ver_dict then
+        local _, incr_err = plugins_conf_ver_dict:incr(PLUGINS_CONF_VERSION_KEY, 1, 0)
+        if incr_err then
+            core.log.error("failed to increase plugins conf version: ", incr_err)
+            core.response.exit(503, {error_msg = "failed to record plugins reload"})
+        end
+    end
+
     local success, err = events:post(_M.RELOAD_EVENT, ngx.req.get_method(), ngx.time())
     if not success then
         core.response.exit(503, err)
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13756, opened 2026-07-29
feat(data_encryption): support AES-256 keys in the encryption keyring

AlinsRan:
### Description

`core.data_encryption.init_iv_tbl` only ever built AES-128 ciphers:

```lua
tbl.insert(iv_tbl, assert(aes:new(iv, nil, aes.cipher(128, "cbc"), {iv = iv})))
```

and the CLI config schema pinned every `apisix.data_encryption.keyring` entry to exactly 16 characters, so a 32-byte key could not even get past `apisix start`.

This selects the cipher by key length: a 16-byte key → AES-128, a 32-byte key → AES-256. The schema accepts those two lengths and keeps rejecting everything else, because an unsupported key is dropped by `init_iv_tbl` and an empty keyring makes `encrypt()` store the value in clear text. A dropped key is also logged as an error, for the paths that do not go through the CLI schema.

Both lengths can be mixed in one keyring, so an existing AES-128 keyring can be rotated to AES-256 without losing access to the already encrypted data.

Lets operators use a stronger AES-256 keyring for `encrypt_fields` / SSL key encryption.

### Checklist

- [x] I have added tests corresponding to this change (`t/core/data_encryption.t` for the cipher selection, the keyring read from `config.yaml` and AES-128 → AES-256 rotation; `t/cli/test_main.sh` for the schema validation)
- [x] I have updated the documentation accordingly (`conf/config.yaml.example`, en/zh `plugin-develop.md`)
- [x] I have verified locally (`t/core/data_encryption.t` passes; the `make init` schema cases were checked by hand and confirmed to fail without the schema change)

### DIFF
--- a/apisix/cli/schema.lua
+++ b/apisix/cli/schema.lua
@@ -67,6 +67,18 @@ local etcd_schema = {
     required = {"prefix", "host"}
 }
 
+-- a keyring entry is used both as the AES key and as its IV, so only the two
+-- key lengths AES-CBC accepts here are allowed: 16 bytes (AES-128) and 32
+-- bytes (AES-256). Any other length would be dropped by
+-- core.data_encryption.init_iv_tbl and silently leave the data unencrypted.
+local keyring_key_schema = {
+    type = "string",
+    anyOf = {
+        {minLength = 16, maxLength = 16},
+        {minLength = 32, maxLength = 32},
+    }
+}
+
 local config_schema = {
     type = "object",
     properties = {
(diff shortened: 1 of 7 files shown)


shreemaan-abhishek (dismissed):

nic-6443 (dismissed):

membphis (commented):
[P1] Invalid key lengths can silently disable data encryption

`init_iv_tbl` now skips every key whose length is not 16 or 32 bytes. If all configured keys are invalid, `iv_tbl` stays empty, and the existing empty-keyring branches return plaintext unchanged. This turns a configuration error into a fail-open path for encrypted plugin fields and SSL private keys. Please reject unsupported key lengths during schema or initialization validation and fail startup or configuration loading instead of treating this as an absent keyring.

[P2] The mixed-keyring test does not exercise legacy ciphertext fallback

The test encrypts with the first AES-256 entry and then decrypts with the same entry; the trailing AES-128 key is never used. Please create ciphertext with the legacy 16-byte key only, then decrypt it using `{new_32_byte_key, old_16_byte_key}` to prove existing AES-128 data remains readable during key rotation.

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

PR #13761, opened 2026-07-30
chore(core): remove the dead clean_handlers mechanism

AlinsRan:
### Description

Follow-up to https://github.com/apache/apisix/pull/13717#discussion_r3662535767.

After #12426 moved health checks to `healthcheck_manager`, nothing registers a clean handler any more: `config_util.add_clean_handler()` has no caller outside its own unit test. That makes every `fire_all_clean_handlers()` call a no-op and every `item.clean_handlers = {}` initialization dead weight on each config item.

This removes the mechanism uniformly instead of dropping it from one config backend only:

- `core.config_util`: drop `add_clean_handler`, `cancel_clean_handler`, `fire_all_clean_handlers`
- `config_etcd` / `config_yaml` / `config_xds`: drop the fire sites and the `clean_handlers` initialization
- `plugin.lua`: drop the field from the dummy global rule
- `control/v1.lua`: drop the field from the routes/services dump
- docs: drop the stale `clean_handlers` entries from the control API response samples
- `t/core/config_util.t`: drop the test that only covered the removed API

No behavior change — the removed code paths were unreachable.

### Checklist

- [x] I have verified that this change is backward compatible (if not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

Issue #8958, opened 2023-02-28
Proposal: Advanced LDAP authentication plugin


bzp2010:
# Background

Currently APISIX has provided a simple LDAP authentication plugin that supports simple bind functions, but does not support, for example, user groups, for which users have a need. We can do this by introducing a new plugin to provide a platform to host these newly added and future needed features.

# Benefits

Users can use the new plug-in to bind identity groups to consumers, and colleagues can also use consumerless mode (i.e., authentication only). In the future, the plug-in can be extended to support more complex authentication models.

# Goals

Create a new plugin, called `ldap-auth-advanced`.

### Route configuration

Name | Type | Optional | Default | Description
-- | -- | -- | -- | --
ldap_uri | string | no |   |  
use_starttls | boolean | yes | false |  
use_ldaps | boolean | yes | false |  
ssl_verify | boolean | yes | true |  
timeout | integer | yes | 10000 | socket timeout time
keepalive | boolean | yes | yes | Whether to enable connection pool
keepalive_timeout | integer | yes | 60000 | Connection pool connection timeout time
keepalive_pool_size | integer | yes | 5 | Connection pool size
keepalive_pool_name | string | yes | <host>:<port> | Connection pool name
ldap_debug | boolean | yes | false | Record search logs
consumer_required | boolean | yes | true | Whether the consumer must attach it or not, the certification will fail.
user_dn | string | no |   | DN template for identifying users (cn=%s,ou=users,dc=example,dc=org)
user_membership_attribute | string | no | memberOf | Add: Used to indicate which attribute in a user object indicates the group to which it belongs. For example memberOf. 

### Consumer configuration

Name | Type | Optional | Default | Description
-- | -- | -- | -- | --
user_dn | string | yes |   | User DN, mutually exclusive
group_dn | string | yes |   | Group DN, mutually exclusive

## Plugin logic

1. User access triggered route's access phase
2. The plugin uses `lua-resty-ldap` to connect to the LDAP server. (Connection reuseable)
3. Execute the binding according to the configuration.
4. Performs a search based on the user dn to fetch `user_membership_attribute`
5. Try to attach user dn or group dn to consumer (user dn > group dn)

### Tips

- The plugin allows the use of route + consumer and pure route modes.
- Consumer support for mapping individual users or individual groups.
- The plugin runs in the access phase.
- The plugin maintains its own connection pool (specified configuration).

# Other

I would like to hear your opinions and welcome comments.

Fabriceli:
[...]

PR #13762, opened 2026-07-30
feat: add ldap-auth-advanced plugin (core authentication)

janiussyafiq:
### Description

Adds the `ldap-auth-advanced` plugin (priority 2541), bringing Kong `ldap-auth-advanced` parity for core authentication: search-then-bind (AD `sAMAccountName` shape), service-account or anonymous search bind, `user_dn` Consumer association, LDAPS/StartTLS, LDAP filter-injection defense, and a strict 401-vs-500 auth/transport error split. This is part 1 of 3 (core authentication); group collection/authorization and credential caching/anonymous-consumer follow in separate PRs. Prior discussion in #8958.

This PR also bumps `lua-resty-ldap` 0.1.0 -> 0.3.0. Under 0.1.0, the existing `ldap-auth` plugin's `tls_verify` option was a silent no-op (the library read a different field name); 0.3.0 fixes this, so `tls_verify: true` configs now perform real certificate verification. This also required regenerating `t/certs/localhost_slapd_cert.pem` (same key, same test CA) to add a `subjectAltName` (`DNS:test.com, DNS:localhost, IP:127.0.0.1`) — the old cert had `CN=test.com` but no SAN, and nginx's hostname verification does not fall back to CN, so the newly-real `ldap-auth` TLS-verify test started failing a certificate host-mismatch check that the 0.1.0 bug had been silently masking.

Related: #8958

### Checklist

- [ ] I have updated the documentation to reflect this change (docs in a follow-up PR)
- [x] I have verified that this change is backward compatible (the `lua-resty-ldap` bump activates real `tls_verify` certificate verification, previously a silent no-op under 0.1.0 -- see description)


apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+        client:set_keepalive()
+    end
+
+    return nil, nil, user_dn
+end
+
+
+function _M.rewrite(conf, ctx)
+    -- Strip any client-supplied X-Authenticated-Groups before any auth work.
+    core.request.set_header(ctx, "X-Authenticated-Groups", nil)
```
nic-6443, line 365: P1: Could we also clear `X-Consumer-Username`, `X-Consumer-Custom-ID`, and `X-Credential-Identifier` here? With `consumer_required=false`, `attach_consumer()` is skipped, so client-supplied values for these standard identity headers pass through unchanged. An authenticated client can therefore impersonate another Consumer to an upstream that trusts APISIX identity headers. All APISIX-authenticated identity headers should be stripped before authentication, then only values derived from the matched Consumer should be added.

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+local type = type
+local ngx_decode_base64 = ngx.decode_base64
+local ngx_re_match = ngx.re.match
+local str_find = string.find
+local str_sub = string.sub
+local parse_addr = core.utils.parse_addr
+
+-- RFC 4512 attribute-description shape: a leading letter, then letters,
+-- digits, semicolons (option separators) or hyphens.
+local ATTR_PATTERN = "^[A-Za-z][A-Za-z0-9;-]*$"
```
nic-6443: P2: RFC 4512 allows an `AttributeType` to be either a descriptor or a numeric OID. This pattern rejects valid values such as `1.2.840.113556.1.4.656`, so directories configured with OID attribute names cannot enable the plugin. Could we accept the `numericoid` form as well while still validating any `;option` suffixes?

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+
+
+-- Extract username/password from the credential header: the scheme word is
+-- conf.header_type ("ldap" or "basic", case-insensitive), the payload is
+-- base64("username:password").
+local function extract_credentials(conf, ctx)
+    -- Proxy-Authorization is checked before Authorization
+    local header_name = "Proxy-Authorization"
+    local auth_header = core.request.header(ctx, header_name)
+    if not auth_header then
```
nic-6443: P2: Please fall back to `Authorization` when `Proxy-Authorization` is present but cannot produce credentials for `conf.header_type`. A forward proxy can send its own `Proxy-Authorization: Basic ...` while forwarding a valid end-user `Authorization: ldap ...`; currently the mere presence of the proxy header makes the valid Authorization unreachable and returns 401. Proxy credentials should take priority only after scheme, base64, and separator parsing succeeds.

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+
+
+-- Tell a directory result-code failure (an auth failure -> 401) apart from a
+-- socket/TLS/timeout error (an outage -> 500) by the library's error prefixes.
+local function is_result_code_failure(err)
+    if type(err) ~= "string" then
+        return false
+    end
+    return str_sub(err, 1, 18) == "simple bind failed"
+        or str_sub(err, 1, 13) == "search failed"
```
nic-6443: P2: This classifies every LDAP operation result as a client authentication failure. A service-account bind rejected after password rotation and search results such as `busy` or `unavailable` all have these prefixes, so the route returns 401 even though the client credentials have not been evaluated. Could we classify by operation and result code—for example, return 401 only for `invalidCredentials` from the final user bind (plus the not-found/ambiguous cases handled below), and treat search-bind/search operational failures as 5xx?

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+
+
+-- Build the user_dn lookup map in one pass over the Consumers, cached
+-- against the Consumer config version.
+local consumer_lrucache = core.lrucache.new({ ttl = 300, count = 512 })
+
+local function build_consumer_maps(consumer_conf)
+    local by_user_dn = {}
+    for _, node in ipairs(consumer_conf.nodes) do
+        local auth_conf = node.auth_conf
```
nic-6443: P2: `node.auth_conf` still contains literal `$secret://...` or `$ENV://...` references here. This custom map bypasses `consumer_mod.consumers_kv()`, which clones the auth config, resolves secrets, and skips unresolved references fail-closed. As a result, a valid Consumer whose `user_dn` is stored by reference can never match the DN returned by LDAP. Please build the map from resolved Consumer config or reuse `consumers_kv()`.

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,382 @@
...
+        if auth_conf and auth_conf.user_dn then
+            by_user_dn[auth_conf.user_dn] = node
```
nic-6443: P2: Please also register `ldap-auth-advanced` → `user_dn` in `plugin_unique_key_attrs` in `apisix/consumer.lua`. Otherwise the Admin API accepts duplicate direct `user_dn` values and this assignment silently lets the last-loaded Consumer win, so Consumer-level ACL or rate-limit policy can run under the wrong identity.

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: I've some reserve on this but already addressed. Can u take a look at Test 62 and forward and see if this is what you expected?
https://github.com/apache/apisix/pull/13762/changes#diff-7fddff66e1de9a3b6a75a346d6bf0579d9bcad7e764425094fd3983160e56fa0R1417

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,390 @@
...
+local ATTR_PATTERN = "^(?:[A-Za-z][A-Za-z0-9-]*"
+                     .. "|(?:0|[1-9][0-9]*)(?:\\.(?:0|[1-9][0-9]*))+)"
+                     .. "(?:;[A-Za-z0-9-]+)*$"
+
+local schema = {
+    type = "object",
+    title = "work with route or service object",
+    properties = {
+        -- connection
+        ldap_uri     = { type = "string" },                        -- "host[:port]"
```
shreemaan-abhishek: can we add min and max lengths here? (and other relevant fields)?

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,390 @@
...
+    -- under multi-auth, decline quietly and let the wrapper render the 401
+    if auth_utils.is_running_under_multi_auth(ctx) then
+        return 401
+    end
+
+    if reason then
+        core.log.warn(plugin_name, ": ", reason)
+    end
+    core.response.set_header("WWW-Authenticate",
+                             conf.header_type .. " realm=\"" .. conf.realm .. "\"")
```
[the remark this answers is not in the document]
janiussyafiq: fixed

ci/pod/openldap/enable-anon-bind.sh
```diff
@@ -0,0 +1,67 @@
...
+olcAllows: bind_anon_dn
+EOF
+
+info "Installing the bind-identity olcAccess rules on the data database"
+
+# Resolve the data database DN by suffix so we do not hard-code the {N} index.
+DATA_DB_DN="$(ldapsearch -Y EXTERNAL -H "ldapi:///" -b cn=config \
+    "(olcSuffix=dc=example,dc=org)" dn 2>/dev/null \
+    | awk '/^dn:/ { $1=""; sub(/^ /,""); print; exit }')"
+
```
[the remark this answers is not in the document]
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,390 @@
...
+    type = "object",
+    title = "work with route or service object",
+    properties = {
+        -- connection
+        ldap_uri     = { type = "string" },                        -- "host[:port]"
+        use_ldaps    = { type = "boolean", default = false },
+        use_starttls = { type = "boolean", default = false },
+        ssl_verify   = { type = "boolean", default = true },
+        timeout      = { type = "integer", minimum = 1, maximum = 60000,
+                         default = 3000 },                          -- milliseconds
```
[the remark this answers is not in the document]
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

apisix/plugins/ldap-auth-advanced.lua
nic-6443: Thanks, scheme mismatches and decode failures now fall back correctly. One unusable case still bypasses the fallback: Lua treats `""` as truthy, so `Proxy-Authorization: ldap OnBhc3M=` (`:pass`) or `ldap dXNlcjo=` (`user:`) enters this branch; the later empty-field check returns 401 without trying a valid `Authorization`. Please reject empty usernames/passwords before granting Proxy priority and cover both cases.

apisix/plugins/ldap-auth-advanced.lua
```diff
@@ -0,0 +1,399 @@
...
+-- released to the pool only on success, so a poisoned socket is never pooled.
+local function ldap_resolve(conf, ctx, username, password)
+    -- The only client-controlled part of the search filter is the escaped
+    -- username. filter.escape leaves bytes the filter grammar rejects (e.g.
+    -- invalid UTF-8), and a grammar reject at search time would surface as a
+    -- 500 -- misclassifying a bad credential as a server fault. Pre-compile
+    -- the filter so any reject is a clean 401.
+    local search_filter = "(" .. conf.attribute .. "="
+                          .. ldap_filter.escape(username) .. ")"
+    if not ldap_filter.compile(search_filter) then
```
nic-6443, line 250: P2: This preflight rejects a valid RFC 4515 assertion value. `~` (0x7e) belongs to `UTF1SUBSET` (`%x5D-7F`), so `admin~` is a legal username, but the bundled filter grammar rejects it at the end and TEST 46 now codifies a 401. Please fix/upgrade the parser or escape `~` as `\\7e` before compiling so valid directory users remain authenticatable.

apisix-master-0.rockspec
```diff
@@ -79,7 +79,7 @@ dependencies = {
     "net-url = 1.2-1",
     "xml2lua = 1.6-2",
     "lua-resty-mediador = 0.1.2-1",
-    "lua-resty-ldap = 0.1.0-0",
+    "lua-resty-ldap = 0.3.0-0",
```
nic-6443: P2: `lua-resty-ldap 0.3.0-0` pins `lpeg = 1.0.2-1`. In a clean APISIX dependency tree, applying this bump removes the currently resolved `lpeg 1.1.0-2` and installs 1.0.2; `graphql` and `jsonpath` then also run against the downgraded global module. Please publish/use an LDAP rock with a compatible range such as `lpeg >= 1.0.2`, so APISIX retains its current LPeg version.

apisix/plugins/ldap-auth-advanced.lua
nic-6443: Yes, Tests 62-67 and the operation/result-code split now match what I expected: service-bind failures and non-auth LDAP results return 500, while only final-user `invalidCredentials` and ambiguous/not-found users return 401. Thanks.

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed in https://github.com/api7/lua-resty-ldap/pull/38

apisix-master-0.rockspec
janiussyafiq: released https://github.com/api7/lua-resty-ldap/pull/39

apisix/plugins/ldap-auth-advanced.lua
janiussyafiq: fixed

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (commented):
Two issues need to be addressed before this can merge:

1. **[P2] Avoid three LDAP operations on every request**

The successful `rewrite` path currently performs a service-account/anonymous bind, a user search, and a user bind (`0 -> 3` LDAP protocol operations per request). Connection pooling does not remove these round trips.

Please add a small local TTL cache for successful authentication results. Cache hits should reuse the resolved canonical `user_dn` and skip LDAP entirely; only cache misses should run the full bind -> search -> bind flow. The cache key must include the relevant plugin configuration and a non-reversible credential discriminator—do not key only by username or retain plaintext passwords—and the TTL must be short and bounded.

2. **[P1] Add a gateway lifecycle shell regression**

This PR adds `ldap-auth-advanced` to the default plugin registry, but the current coverage is limited to Perl `.t` tests. Please add a required-CI `t/cli/test_*.sh` regression that starts the real gateway, exercises plugin loading/configuration with bounded readiness polling, verifies externally observable behavior and plugin state, covers the reverse transition where applicable, and cleans up all processes and temporary configuration.

membphis (approved):

janiussyafiq:
> Two issues need to be addressed before this can merge:
> 
> 1. **[P2] Avoid three LDAP operations on every request**
> 
> The successful `rewrite` path currently performs a service-account/anonymous bind, a user search, and a user bind (`0 -> 3` LDAP protocol operations per request). Connection pooling does not remove these round trips.
> 
> Please add a small local TTL cache for successful authentication results. Cache hits should reuse the resolved canonical `user_dn` and skip LDAP entirely; only cache misses should run the full bind -> search -> bind flow. The cache key must include the relevant plugin configuration and a non-reversible credential discriminator—do not key only by username or retain plaintext passwords—and the TTL must be short and bounded.
> 

Will be addressed in upcoming PR slice.

> 2. **[P1] Add a gateway lifecycle shell regression**
> 
> This PR adds `ldap-auth-advanced` to the default plugin registry, but the current coverage is limited to Perl `.t` tests. Please add a required-CI `t/cli/test_*.sh` regression that starts the real gateway, exercises plugin loading/configuration with bounded readiness polling, verifies externally observable behavior and plugin state, covers the reverse transition where applicable, and cleans up all processes and temporary configuration.

Pushbacked, no existing auth cli tests being exercised.

PR #12765, opened 2025-11-21
feat: Add error ratio-based circuit breaking policy to api-breaker plugin

HaoTien:
# feat: Add error ratio-based circuit breaking policy to api-breaker plugin

## What this PR does / why we need it

This PR implements error ratio-based circuit breaking (`unhealthy-ratio` policy) for the `api-breaker` plugin, providing more intelligent and adaptive circuit breaking behavior based on error rates within a sliding time window, rather than just consecutive failure counts.

**Closes #12763** 

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

### Current Limitations
- The existing failure count-based approach only considers consecutive failures
- It doesn't account for the overall error rate in relation to total requests
- May be too sensitive during low traffic periods or not sensitive enough during high traffic periods

### New Features Added
- **Error ratio-based circuit breaking**: New `unhealthy-ratio` policy that triggers circuit breaker based on error rate within a sliding time window
- **Configurable parameters**: Support for error ratio threshold, minimum request threshold, sliding window size, etc.
- **Circuit breaker states**: Proper implementation of CLOSED, OPEN, and HALF_OPEN states
- **Backward compatibility**: Existing configurations continue to work without changes

### New Configuration Parameters

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `policy` | string | `"unhealthy-count"` | Circuit breaker policy |
| `unhealthy.error_ratio` | number | `0.5` | Error rate threshold (0-1) to trigger circuit breaker |
| `unhealthy.min_request_threshold` | integer | `10` | Minimum requests needed before evaluating error rate |
| `unhealthy.sliding_window_size` | integer | `300` | Sliding window size in seconds for error rate calculation |
| `unhealthy.permitted_number_of_calls_in_half_open_state` | integer | `3` | Number of permitted calls in half-open state |
| `healthy.success_ratio` | number | `0.6` | Success rate threshold to close circuit breaker from half-open state |

### Example Configuration

```json
{
  "plugins": {
    "api-breaker": {
      "break_response_code": 503,
      "policy": "unhealthy-ratio",
      "max_breaker_sec": 60,
      "unhealthy": {
        "http_statuses": [500, 502, 503, 504],
        "error_ratio": 0.5,
        "min_request_threshold": 10,
        "sliding_window_size": 300,
        "permitted_number_of_calls_in_half_open_state": 3
      },
      "healthy": {
        "http_statuses": [200, 201, 202],
[...]

PR #13764, opened 2026-07-30
feat: add error ratio-based circuit breaking policy to api-breaker plugin

### LINKED ISSUE #12763: feat: As a user, I want error ratio-based circuit breaking in api-breaker plugin, so that I can have more intelligent circuit breaking based on error rates instead of just failure counts
### Description

# feat: As a user, I want error ratio-based circuit breaking in api-breaker plugin, so that I can have more intelligent circuit breaking based on error rates instead of just failure counts

## Description

Currently, the `api-breaker` plugin only supports failure count-based circuit breaking (`unhealthy-count` policy), which triggers circuit breaker when consecutive failure count reaches a threshold. This approach may not be suitable for all scenarios, especially when dealing with varying traffic patterns.

I would like to propose adding an error ratio-based circuit breaking policy (`unhealthy-ratio`) that triggers circuit breaker based on error rate within a sliding time window, providing more intelligent and adaptive circuit breaking behavior.

### Current Limitations
- The existing failure count-based approach only considers consecutive failures
- It doesn't account for the overall error rate in relation to total requests
- May be too sensitive during low traffic periods or not sensitive enough during high traffic periods

### Benefits of Error Ratio-based Circuit Breaking
- More accurate representation of service health by considering error rate rather than just failure count
- Better handling of varying traffic patterns
- Configurable sliding time window for flexible error rate calculation
- Support for circuit breaker states: CLOSED, OPEN, and HALF_OPEN

## Proposed Solution

Add a new `policy` parameter to the `api-breaker` plugin with two options:
- `unhealthy-count` (default, existing behavior)
- `unhealthy-ratio` (new error ratio-based policy)

### New Configuration Parameters for `unhealthy-ratio` Policy

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `policy` | string | `"unhealthy-count"` | Circuit breaker policy |
| `unhealthy.error_ratio` | number | `0.5` | Error rate threshold (0-1) to trigger circuit breaker |
| `unhealthy.min_request_threshold` | integer | `10` | Minimum requests needed before evaluating error rate |
| `unhealthy.sliding_window_size` | integer | `300` | Sliding window size in seconds for error rate calculation |
| `unhealthy.permitted_number_of_calls_in_half_open_state` | integer | `3` | Number of permitted calls in half-open state |
| `healthy.success_ratio` | number | `0.6` | Success rate threshold to close circuit breaker from half-open state |

### Example Configuration

```json
{
  "plugins": {
    "api-breaker": {
      "break_response_code": 503,
      "policy": "unhealthy-ratio",
      "max_breaker_sec": 60,
      "unhealthy": {
        "http_statuses": [500, 502, 503, 504],
        "error_ratio": 0.5,
        "min_request_threshold": 10,
        "sliding_window_size": 300,
        "permitted_number_of_calls_in_half_open_state": 3
      },
      "healthy": {
        "http_statuses": [200, 201, 202],
        "success_ratio": 0.6
      }
    }
  }
}
```

### Circuit Breaker States
- **CLOSED**: Normal request forwarding
- **OPEN**: Direct circuit breaker response without forwarding requests
- **HALF_OPEN**: Limited requests allowed to test service recovery

### Algorithm
1. Track requests and errors within a sliding time window
2. When request count ≥ `min_request_threshold` and error rate ≥ `error_ratio`, open circuit breaker
3. After `max_breaker_sec`, transition to half-open state
4. In half-open state, allow up to `permitted_number_of_calls_in_half_open_state` requests
5. If sufficient successful requests, close circuit breaker; otherwise, reopen

## Backward Compatibility

This enhancement is fully backward compatible:
- Existing configurations continue to work without changes
- Default `policy` is `"unhealthy-count"` (existing behavior)
- No breaking changes to existing APIs

## Testing

Comprehensive test coverage will be provided including:
- Schema validation tests for new parameters
- Functional tests for error ratio calculation
- Circuit breaker state transition tests
- Integration tests with various traffic patterns
- Backward compatibility tests

## Use Cases

1. **High-traffic services**: Better handling of error spikes in high-volume scenarios
2. **Variable traffic patterns**: Adaptive behavior for services with fluctuating request rates
3. **Microservices architectures**: More precise circuit breaking for service mesh environments
4. **SLA-based circuit breaking**: Configure circuit breaker based on acceptable error rates

## Files to be Modified

- `apisix/plugins/api-breaker.lua` - Core plugin logic
- `t/plugin/api-breaker.t` - Test cases (new test file for ratio-based tests)
- `docs/en/latest/plugins/api-breaker.md` - English documentation
- `docs/zh/latest/plugins/api-breaker.md` - Chinese documentation

## Additional Information

This feature has been implemented and tested locally. I'm ready to submit a PR with:
- Complete implementation of the error ratio-based circuit breaking
- Comprehensive test suite following APISIX testing standards
- Updated documentation in both English and Chinese
- Backward compatibility preservation

Would appreciate feedback on this proposal and guidance on the contribution process.

yeganeahmadnejad:
## What this PR does / why we need it

This PR implements error ratio-based circuit breaking (`unhealthy-ratio` policy) for the `api-breaker` plugin, providing more intelligent and adaptive circuit breaking behavior based on error rates within a sliding time window, rather than just consecutive failure counts.

Closes #12763

## Context: reviving #12765

This is a revival of #12765 by @HaoTien, which was closed by the stale-bot for inactivity on 2026-07-24 despite:

- Passing all CI checks (build, lint, docs, kubernetes-discovery, tars, etc.)
- ~20 rounds of review with @Baoyuantop, addressing sliding-window-reset, half-open-failure-fallback, and over-limit half-open call test coverage
- A `CHANGES_REQUESTED` review from @moonming on 2026-03-16 asking for a summary of the design decisions made during review, which @HaoTien answered in full on 2026-03-17, followed by two more pings for re-review that went unanswered

The commit here carries over the final reviewed tree from #12765 unchanged (rebased onto current master; no functional changes), with commit authorship preserved for @HaoTien. Opening a fresh PR since the original was auto-closed and I don't have push access to reopen it or its branch.

### Current Limitations

- The existing failure count-based approach only considers consecutive failures
- It doesn't account for the overall error rate in relation to total requests
- May be too sensitive during low traffic periods or not sensitive enough during high traffic periods

### New Features Added

- **Error ratio-based circuit breaking**: New `unhealthy-ratio` policy that triggers circuit breaker based on error rate within a sliding time window
- **Configurable parameters**: Support for error ratio threshold, minimum request threshold, sliding window size, etc.
- **Circuit breaker states**: Explicit CLOSED, OPEN, and HALF_OPEN states
- **Backward compatibility**: Existing configurations continue to work unchanged (`unhealthy-count` remains the default policy)

### New Configuration Parameters

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `policy` | string | `"unhealthy-count"` | Circuit breaker policy |
| `unhealthy.error_ratio` | number | `0.5` | Error rate threshold (0-1) to trigger circuit breaker |
| `unhealthy.min_request_threshold` | integer | `10` | Minimum requests needed before evaluating error rate |
| `unhealthy.sliding_window_size` | integer | `300` | Sliding window size in seconds for error rate calculation |
| `unhealthy.half_open_max_calls` | integer | `3` | Number of permitted calls in half-open state |
| `healthy.success_ratio` | number | `0.6` | Success rate threshold to close circuit breaker from half-open state |

### Example Configuration

```json
{
  "plugins": {
    "api-breaker": {
      "break_response_code": 503,
      "policy": "unhealthy-ratio",
      "max_breaker_sec": 60,
      "unhealthy": {
        "http_statuses": [500, 502, 503, 504],
        "error_ratio": 0.5,
        "min_request_threshold": 10,
        "sliding_window_size": 300,
        "half_open_max_calls": 3
      },
      "healthy": {
        "http_statuses": [200, 201, 202],
        "success_ratio": 0.6
      }
    }
  }
}
```

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## How Has This Been Tested?

- [x] Schema validation tests for new parameters
- [x] Functional tests for error ratio calculation
- [x] Circuit breaker state transition tests (CLOSED → OPEN → HALF_OPEN → CLOSED/OPEN)
- [x] Sliding window expiration/reset tests
- [x] Half-open failure fallback tests
- [x] Over-limit half-open call rejection tests
- [x] Backward-compatibility tests for the existing `unhealthy-count` policy

This is the same test coverage that was previously reviewed and passed full CI on #12765.


apisix/plugins/api-breaker.lua
```diff
@@ -195,8 +383,145 @@ function _M.access(conf, ctx)
...
+            if transition_success then
+                -- Transition to HALF_OPEN
+                set_circuit_breaker_state(ctx, HALF_OPEN)
+                -- Reset half-open counters
+                shared_buffer:set(gen_half_open_calls_key(ctx), 0)
+                shared_buffer:set(gen_half_open_success_key(ctx), 0)
+                core.log.info("Circuit breaker transitioned from OPEN to HALF_OPEN")
+
+                -- Clean up transition lock
+                shared_buffer:delete(transition_key)
```
membphis: P1 — `half_open_max_calls` is not enforced under concurrency. The request that transitions `OPEN -> HALF_OPEN` returns here without incrementing `half_open_calls`, and deleting the lock immediately lets requests that already observed `OPEN` acquire it again. The regression test with a limit of 2 currently expects three 200 responses. Please make the transition atomic, count the transition request, and decide recovery only after all admitted probes complete.

apisix/plugins/api-breaker.lua
```diff
@@ -133,13 +267,67 @@ local _M = {
...
+end
+
+-- Sliding window management
+local function reset_sliding_window(ctx, current_time, window_size)
+    local window_start_key = gen_window_start_time_key(ctx)
+    local total_requests_key = gen_total_requests_key(ctx)
+    local unhealthy_key = gen_unhealthy_key(ctx)
+
+    shared_buffer:set(window_start_key, current_time)
+    shared_buffer:set(total_requests_key, 0)
```
membphis: P1 — This is a tumbling-window reset, not a sliding window, and it is racy across workers. Concurrent requests that observe an expired or missing start time can each zero the counters, losing increments from completed requests; recent failures just before the boundary are also discarded wholesale. Please use bounded time buckets with atomic updates and add boundary-concurrency coverage.

apisix/plugins/api-breaker.lua
```diff
@@ -61,70 +65,200 @@ local schema = {
...
     return "unhealthy-lasttime" .. core.request.get_host(ctx) .. ctx.var.uri
 end
 
+-- New key generation functions for ratio policy
+local function gen_state_key(ctx)
+    return "cb-state-" .. core.request.get_host(ctx) .. ctx.var.uri
+end
+
+local function gen_total_requests_key(ctx)
+    return "cb-total-" .. core.request.get_host(ctx) .. ctx.var.uri
```
membphis: P1 — These statistics are keyed by the raw URI and have no normal TTL. On a wildcard or high-cardinality route, every distinct path creates persistent `cb-total`, `cb-window`, and failure entries, which can fill `plugin-api-breaker` and evict unrelated breaker state. Please key by a stable route/service configuration identity and expire or otherwise bound the buckets.

membphis (commented):
Blocking findings are noted inline.

apisix/plugins/api-breaker.lua
```diff
@@ -264,4 +589,107 @@ function _M.log(conf, ctx)
...
+
+        -- If in HALF_OPEN state and got a failure, immediately go back to OPEN
+        if current_state == HALF_OPEN then
+            set_circuit_breaker_state(ctx, OPEN)
+            core.log.warn("Circuit breaker returned to OPEN state due to failure in HALF_OPEN")
+            -- Clean up half-open counters
+            shared_buffer:delete(gen_half_open_calls_key(ctx))
+            shared_buffer:delete(gen_half_open_success_key(ctx))
+        end
+    elseif is_success then
```
bzp2010: https://github.com/apache/apisix/pull/13764/changes#diff-ae3e886348fc6297466ddb8cf3bd153451146d7bbb5ff475d0fc1f684a235ea4R451-R457

If the recorded response status code is neither a success nor a failure, the number of half-open requests may be exhausted. This can result in a permanent API outage.

In any case, https://github.com/apache/apisix/pull/13764/changes#diff-ae3e886348fc6297466ddb8cf3bd153451146d7bbb5ff475d0fc1f684a235ea4R451
The `access` phase has already incremented the half-open counter, but the `log` phase does not reset the increment value for exceptions.

As a result, the state may get stuck in a "half-open" state, and new requests will always be rejected due to the "half-open" request limit. However, the log phase will never update the state machine based on unmarked status codes.

apisix/plugins/api-breaker.lua
```diff
@@ -133,13 +267,67 @@ local _M = {
...
+end
+
+local function set_circuit_breaker_state(ctx, state)
+    local state_key = gen_state_key(ctx)
+    local last_change_key = gen_last_state_change_key(ctx)
+    local current_time = ngx.time()
 
-function _M.access(conf, ctx)
+    shared_buffer:set(state_key, state)
+    shared_buffer:set(last_change_key, current_time)
```
bzp2010: The shdict setting in the new code never specifies a TTL, so shdict may gradually fill up and stop accepting new entries, which could lead to undefined behavior or, at the very least, cause the functionality to fail.

apisix/plugins/api-breaker.lua
```diff
@@ -195,8 +383,145 @@ function _M.access(conf, ctx)
...
+            local transition_success
+            transition_success, err = shared_buffer:add(transition_key, 1, 1)
+
+            if err then
+                core.log.warn("failed to add transition lock: ", err)
+            end
+
+            if transition_success then
+                -- Transition to HALF_OPEN
+                set_circuit_breaker_state(ctx, HALF_OPEN)
```
bzp2010: The request that sets the half-open state does not increment the half-open counter itself; theoretically, that request is already in and is to the half-open state, so `half_open_max_calls` actually allows for `n + 1` half-open requests.

The `return` statement below skips the increment code in `https://github.com/apache/apisix/pull/13764/changes#diff-ae3e886348fc6297466ddb8cf3bd153451146d7bbb5ff475d0fc1f684a235ea4R451`.

docs/zh/latest/plugins/api-breaker.md
```diff
@@ -30,34 +29,70 @@ description: 本文介绍了 Apache APISIX api-breaker 插件的相关操作，
...
+| unhealthy.error_ratio                                  | number         | 否     | 0.5    | [0, 1]          | 触发熔断的错误率阈值。例如 0.5 表示错误率达到 50% 时触发熔断。                           |
+| unhealthy.min_request_threshold                        | integer        | 否     | 10     | >=1             | 在滑动时间窗口内触发熔断所需的最小请求数。只有请求数达到此阈值时才会评估错误率。         |
+| unhealthy.sliding_window_size                          | integer        | 否     | 300    | [10, 3600]      | 滑动时间窗口大小，以秒为单位。用于统计错误率的时间范围。                                 |
+| unhealthy.half_open_max_calls | integer        | 否     | 3      | [1, 20]         | 在半开启状态下允许通过的请求数量。用于测试服务是否恢复正常。                             |
+| healthy.http_statuses                                  | array[integer] | 否     | [200]  | [200, ..., 499] | 上游服务处于健康状态时的 HTTP 状态码。                                                   |
+| healthy.successes                                      | integer        | 否     | 3      | >=1             | 上游服务触发健康状态的连续正常请求次数。                                                 |
```
bzp2010: The "healthy.successes" does not exist.

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1. The winning request now falls through into the same admission code path used by later probes instead of returning early, so it's counted as probe #1. The transition lock is left to expire on its own 1s TTL instead of being deleted immediately, so a request racing in during that window is conservatively rejected as still-open rather than risking another uncounted admission.

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1. Replaced the single reset-on-expiry counter pair with fixed time buckets named by their own epoch (cb-breq-<id>-<bucket_size>-<epoch> / cb-bfail-<id>-<bucket_size>-<epoch>). Aging out old data is now implicit — an epoch outside the last N buckets just isn't summed — so there's no shared 'is it time to reset' decision for concurrent requests to race on, and each increment is a single atomic shared_buffer:incr. Added a harness-based regression covering window aging (Scenario 6 in the linked test run) alongside the existing sliding-window .t test.

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1. Ratio-policy keys are now scoped by conf_type/conf_id/conf_version (gen_breaker_id) instead of host+URI — same approach limit-count/limit-conn use to bound cardinality and to start fresh on a config change. Every key now also carries a bounded TTL (get_state_ttl / get_bucket_ttl).

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1. Introduced a dedicated half-open 'completed' counter, incremented for every probe that reaches the log phase regardless of how its status is classified (success/failure/neither). The close-vs-reopen decision now fires once completed >= half_open_max_calls, so a run of unclassified status codes can no longer strand the breaker in HALF_OPEN forever. Verified in harness Scenario 5 (two unclassified 404 probes still resolve the half-open evaluation).

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1. Every shared_buffer:set/incr call in the ratio path now passes an exptime (get_state_ttl for state/half-open bookkeeping, get_bucket_ttl for window buckets).

apisix/plugins/api-breaker.lua
yeganeahmadnejad: Fixed in 6f9af8c1, same root cause as your comment above on the half_open_max_calls enforcement — the transitioning request now falls through into the shared admission/increment path instead of returning early, so it counts as probe #1 rather than granting an extra n+1'th slot.

docs/zh/latest/plugins/api-breaker.md
yeganeahmadnejad: Fixed in 6f9af8c1 — removed the stray healthy.successes row from the unhealthy-ratio table (that field only applies to the unhealthy-count policy).

yeganeahmadnejad:
Thanks @membphis and @bzp2010 for the thorough review — all 7 findings addressed in 6f9af8c1 (replied inline on each thread with specifics):

- **half_open_max_calls not enforced under concurrency / n+1 admission**: the OPEN→HALF_OPEN transitioning request now counts itself as the first probe instead of returning early; the transition lock expires on its own short TTL instead of being deleted immediately.
- **Racy tumbling-window reset**: replaced with fixed time buckets named by their own epoch, so aging out old data doesn't require any request to "reset" shared counters.
- **Unbounded per-URI key cardinality + no TTL**: ratio-policy state is now scoped by conf_type/conf_id/conf_version (matching limit-count/limit-conn) with bounded TTLs throughout.
- **Permanent half-open lockup on unclassified status codes**: added a dedicated "completed" counter that advances regardless of classification, so the close/reopen decision can't stall waiting for a status that will never come.
- **docs/zh stray `healthy.successes` row**: removed from the unhealthy-ratio table.

Also updated the half-open concurrency test (`t/plugin/api-breaker2.t`), which had pinned the buggy behavior (3 admissions against a limit of 2) as expected output.

Verified with a standalone harness that mocks `ngx.shared`/`core` and replays the state machine — including genuinely concurrent `access()` calls arriving before any `log()` resolves — across open/half-open/close/reopen and window-aging scenarios. Happy to take another look at anything.

juzhiyuan:
Just approved the CI to run.

yeganeahmadnejad:
The failing \`build (ubuntu-latest, linux_openresty, t/plugin/[a-k]*.t ...)\` job is unrelated to this PR's changes.

Both plugin test files this PR touches pass cleanly in that run:
- \`t/plugin/api-breaker.t\` — ok (44.3s)
- \`t/plugin/api-breaker2.t\` — ok (41.9s)

The actual failure is in \`t/plugin/basic-auth.t\` TEST 22, a Vault-backed secret-reference test:

\`\`\`
Failed test 't/plugin/basic-auth.t TEST 22: set basic-auth conf: password uses secret ref
- pattern "[error]" should not match any line in error.log but matches line
"... failed to resolve secret reference: \$secret://vault/test1/foo/passwd, field: password, err: no value ..."
\`\`\`

This PR doesn't touch Vault, \`secret.lua\`, \`consumer.lua\`, or basic-auth — the error (\`err: no value\`) looks like the test Vault container wasn't ready in time, i.e. CI flakiness rather than a real regression. Could someone retrigger the job? Link: https://github.com/apache/apisix/actions/runs/30573824296/job/91308878901

PR #13778, opened 2026-08-05
feat(ai-proxy): send LLM requests through ngx_http_ffi_client

shreemaan-abhishek:
### Description

**Why.** `ai-proxy`, `ai-proxy-multi` and `ai-request-rewrite` all send their
outbound LLM requests through `apisix/plugins/ai-transport/http.lua`. That path
runs on every request the AI gateway proxies, and `lua-resty-http` spends
roughly three times the outbound CPU time of `ngx_http_ffi_client`, the HTTP
client implemented as an nginx C module.

**How.** `plugin_attr.ai-proxy.http_client` names the client:
`ngx_http_ffi_client`, the default, or `lua-resty-http`. The name is validated
against an enum schema and mapped to a module, the module is loaded on the
first request, and the result is cached only once a client has actually loaded.

The two clients share an object API (`new`, `set_timeout`, `connect`,
`request`, `res.body_reader`, `res:read_body`, `set_keepalive`, `close`), so
the change is confined to which client the transport constructs; the providers
and the streaming path are untouched.

A client that is missing or cannot be created fails the request and the error
names the cause. The transport never falls back to the other client, so what
runs is always what the config asked for.

**Name resolution.** Cosockets have their hostnames resolved by
`apisix/patch.lua`, which routes them through `core.resolver` and so honours
`dns_resolver`, `/etc/hosts` and the search domains. The C client dials from C
and never touches a cosocket, so left alone it would see only nginx's
`resolver` and resolve names by different rules than the rest of the gateway.
It takes a resolver as of v0.1.3 (api7/ngx_http_ffi_client#45), so the client
is handed `core.resolver.parse_domain` once, when the module loads. The client
resolves before it derives its connection-pool key and keeps the name for the
`Host` header and the SNI. A client too old to take a resolver fails the
request and names that as the cause, rather than silently resolving by its own
rules.

### Runtime

`APISIX_RUNTIME` moves to **1.3.16**, built with `ngx_http_ffi_client`
**v0.1.3** (api7/apisix-build-tools#488, released as `apisix-runtime/1.3.16`;
the tag is that PR's merge commit). `ci/linux-install-openresty.sh` picks up the
new debug-deb checksums for both architectures.

Why this client revision. 1.3.13 and 1.3.14 pin v0.1.1, which is v0.1.0 plus a
LICENSE, a README section and a `config` tweak, with byte-identical C sources.
The releases that matter start at v0.1.2, and v0.1.3 is the first to carry all
four of the defects and gaps this cutover surfaced:

| fix | issue | without it |
|---|---|---|
| request bodies containing CR or LF are no longer rejected | api7/ngx_http_ffi_client#38 | any caller posting pretty-printed JSON gets a 500 |
| connection errors respect `lua_socket_log_errors` | api7/ngx_http_ffi_client#39 | every refused upstream writes an `[error]` line |
| trust store falls back to `lua_ssl_trusted_certificate` | api7/ngx_http_ffi_client#42 | every verified TLS connection fails, so no HTTPS provider is reachable |
| names resolve through a resolver the host application installs | api7/ngx_http_ffi_client#45 | the client sees only nginx's `resolver`, so the gateway's own resolution rules never apply to it |

To check what a given build actually contains:

```
nginx -V 2>&1 | tr ' ' '\n' | grep ffi_client
```

which prints the `--add-module=.../ngx_http_ffi_client-<tag>` path. On the
published 1.3.16 packages that reads `ngx_http_ffi_client-v0.1.3`.

v0.1.3 is v0.1.2 plus the resolver hook, and that hook is a change to the Lua
binding alone: the C sources are byte-identical between the two tags.

Performance record for the client:
https://github.com/api7/ngx_http_ffi_client/blob/63771541c5a229da8a840ab6008ba3771a22fb71/benchmark/results-full.md

A hand-built runtime without the module has to set
`http_client: lua-resty-http`, which runs exactly the code that ran before.

N/A

### Checklist

### Test plan

`t/plugin/ai-transport-http.t` covers client selection with stubs, and then
drives the **real C client** against local upstreams so module loading, the FFI
boundary and the connection lifecycle are exercised rather than mocked:

| case | what it proves |
|---|---|
| TEST 8 | the C client is the default |
| TEST 9 | a runtime whose C module is not built in fails the request, twice in a row, instead of switching clients |
| TEST 12 | a module that loads but is not a table fails the request |
| TEST 15 | a module whose loader raises fails the request, covering the `pcall(require, ...)` failure path |
| TEST 13 | an unrecognised `http_client` fails schema validation |
| TEST 10, 11 | `plugin_attr` selects `lua-resty-http`, and that path runs against a real upstream with the C client stubbed to a sentinel so a selection regression cannot pass silently |
| TEST 14 | the client is given `core.resolver.parse_domain`, the gateway's own resolver |
| TEST 21 | a client too old to take a resolver fails the request, twice in a row |
| **TEST 16** | real C client, buffered request and response body against a local upstream |
| **TEST 17** | real C client, an SSE response read to completion through `body_reader` |
| **TEST 18** | real C client, three requests over one pooled connection, asserted via the upstream's `connection_requests` counter |
| **TEST 19** | real C client to an upstream named `localhost`, which only resolves through `core.resolver`, with the name kept in the `Host` header |
| **TEST 20** | real C client over TLS with `ssl_verify` on and no per-call CA, verifying against `lua_ssl_trusted_certificate` |

TESTs 16-20 carry
`--- skip_eval: 2: !-f "/usr/local/openresty/lualib/resty/ngx_http_ffi_client.lua"`
so they run wherever the module is installed, which includes CI on 1.3.16, and
skip on a runtime without it rather than failing for the wrong reason.

Beyond the suite, this exact code was run on a gateway carrying the pinned
client against **real Azure OpenAI**, comparing both clients on the same
routes:

| scenario | ngx_http_ffi_client | lua-resty-http |
|---|---|---|
| buffered chat (gpt-4o) | 200 | 200 |
| streaming SSE | 15 chunks, `[DONE]` | 15 chunks, `[DONE]` |
| multiline JSON body | 200 | 200 |
| failover 5xx to a healthy instance | 200 | 200 |
| keepalive, 5 sequential | 5/5 | 5/5 |
| responses API (gpt-5-codex) | 200 | 200 |
| anthropic messages (claude-sonnet-4-5) | 200 | 200 |
| 36 KB request body | 200 | 200 |

That covers three protocol adapters (`openai-chat`, `openai-responses`,
`anthropic-messages`), TLS with SNI, streaming, retry and connection reuse.

The resolver hook was re-run against the same three providers. Two extra routes
went to upstreams named only in `/etc/hosts`, which nginx's `resolver` cannot
answer for: both returned 200 with the name intact in the `Host` header the
upstream received, and both failed with `<name> could not be resolved` once the
`set_resolver` call was removed, which is what pins the behaviour on the hook.
The three Azure hostnames happen to share one address, so the interleaved runs
also cover pooling across SNIs on a single peer: `ssl_server_name` is part of
the client's pool key, so they never share a connection.
Reassembled streaming text is identical on both clients; only the raw byte count
differs, and it varies run-to-run within each client, so that is provider
variance rather than client behaviour.


apisix/plugins/ai-transport/http.lua
```diff
@@ -19,17 +19,130 @@
...
+    return http_client
+end
+
+
+--- Resolve the upstream name the way every other socket in the gateway does.
+-- Cosockets are patched (apisix/patch.lua) to run names through
+-- core.resolver, which honours dns_resolver, /etc/hosts and the search
+-- domains. The C client dials on its own and only sees nginx's `resolver`,
+-- so the name is resolved here and kept for the Host header and the SNI.
+local function resolve_upstream_host(params)
```
shreemaan-abhishek, line 101: ngx_http_ffi_client dials from C, so it never touches the cosocket that patch.lua wraps — it only sees nginx's `resolver`, which does not read /etc/hosts or our dns_resolver config. That is why every `localhost` upstream 500'd. Resolving here puts it back on the same path as every other socket in the gateway, and we keep the name for Host and SNI.

membphis (commented):
[P1] Please address these blockers before merge:

1. The current t/plugin/[a-k]*.t CI shard is failing on this head, and it contains the changed t/plugin/ai-transport-http.t coverage. Please identify the cause and make the relevant checks pass on this exact head.

2. Please make the ngx_http_ffi_client revision bundled in apisix-runtime 1.3.12 traceable, and add APISIX-level coverage that runs the real C client through buffered responses, SSE streaming, TLS/DNS, and keepalive. The new tests mostly stub the module, so they do not prove that the new default client includes and preserves the required framing, connection-lifecycle, and error-semantics fixes.

3. The linux_apisix_current_luarocks_in_customed_nginx job is also failing. Because custom runtimes may not include the C module, please verify the supported compatibility path on this head: either include the module or explicitly configure and test lua-resty-http as the fallback for that build.

apisix/plugins/ai-transport/http.lua
```diff
@@ -19,17 +19,131 @@
...
+    if not ok or type(mod) ~= "table" then
+        core.log.error(module_name, " is not available: ", mod)
+        return nil, module_name .. " is not available: " .. tostring(mod)
+    end
+
+    http_client = mod
+    http_client_is_ffi = name == FFI_CLIENT
+
+    return http_client
+end
```
nic-6443: These codes should not be placed directly under the `apisix/plugins/ai-transport` module. Our plan is to gradually replace the dependencies on lua‑resty‑http across all plugins, such as forward‑auth and http‑logger. Placing the codes here will prevent them from being reused.

apisix/plugins/ai-transport/http.lua
nic-6443: Could ngx_http_ffi_client provide a hook for DNS resolution? Otherwise, the DNS‑resolving code would have to be duplicated across many plugins.

apisix/plugins/ai-transport/http.lua
shreemaan-abhishek: Done. The selection now lives in `apisix/utils/http.lua`, next to the other shared client factories (`utils/redis.lua` is the closest precedent), and `ai-transport/http.lua` is just a consumer.

- `utils/http.lua` owns the client names, a `client_schema` fragment callers embed in their own attribute schema, loading and per-name caching, and `resolve_upstream_host()`.
- the caller owns **where the preference comes from** and passes the name to `new(name)`. So `ai-transport` reads `plugin_attr.ai-proxy.http_client`, and forward-auth or http-logger can read their own key without inheriting an AI-specific one.

`resolve_upstream_host()` is the part most worth sharing. `apisix/patch.lua` wraps cosocket `connect` so hostnames go through `core.resolver`, which honours `dns_resolver`, `/etc/hosts` and the search domains. A client that dials from C never touches a cosocket and sees only nginx's `resolver`, so without this every `localhost`-style upstream fails. Any plugin moving off lua-resty-http will hit that, so it should not be reimplemented per plugin.

One option I skipped: having `new()` return a wrapper whose `connect()` resolves transparently, so callers would not need the `needs_resolve()` check. That removes a step for every future caller but adds a metatable indirection on a per-request path, so I left it explicit. Happy to switch if you prefer it.

membphis (dismissed):
LGTM

AlinsRan (dismissed):

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13779, opened 2026-08-05
perf(plugin): reuse the filtered global-rule plugin set across phases

AlinsRan:
ref: https://github.com/api7/ticket-analysis/issues/541

### Description

`run_global_rules()` rebuilds the same plugin set on every phase. Route plugins are computed once and reused through `api_ctx.plugins`; global rules had no equivalent, so a single request re-ran `_M.filter()` for the same merged rule again and again — 9 times for a plain request, and more for a streamed response, since `body_filter` and `delayed_body_filter` run once per response buffer.

This caches the filtered set on `api_ctx`, keyed by the merged rule's `modifiedIndex` and by the matched route.

The route belongs in the key: plugins declaring `run_policy = "prefer_route"` are skipped when the route configures the same plugin, and `api_ctx.matched_route` is replaced when a consumer's configuration is merged in — which happens after the `rewrite` phase has run but before `access`.

Since the table now outlives a single phase, it is taken from the `global_plugins` pool and released in `http_log_phase` rather than at the end of every phase.

### Measurements

Measured on an APISIX 3.2-based fork carrying the same change. Single worker, CPU-pinned (worker and load generator on separate cores), wrk2, 5 runs each, median, with a `global_rule` enabling prometheus:

| build | rps |
|---|---|
| baseline | 21,453 |
| cross-phase reuse | 24,062 (**+12.2%**) |

Not measured on master itself.

### Tests

`t/plugin/global-rule-phase-cache.t`:

- every phase still runs off the cached set (a global rule with plugins in `rewrite` and `body_filter`)
- `body_filter` still runs per response buffer on a chunked response — the case the cache affects most
- the set is rebuilt when the global rule changes mid-flight

**Known gap:** the fork also carries a case proving the set is rebuilt when a consumer merge replaces `matched_route` — verified there by deleting the route key and watching it fail. It observes `prefer_route` through prometheus metrics, and I could not port it: `Test::Nginx` restarts nginx between blocks whose `--- config` differs, and the metrics live in a per-process shared dict, so cross-block observation resets. As `prometheus` and `skywalking` are the only `prefer_route` plugins, I did not find another observation point. The route cache key is therefore untested here — suggestions welcome.

### Checklist

- [x] I have explained why we need this PR and what problem it solves
- [x] I have explained the changes in this PR
- [x] I have added tests
- [ ] I have updated the documentation — no user-facing behaviour change

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -1244,6 +1244,10 @@ function _M.http_log_phase()
         core.tablepool.release("plugins", api_ctx.plugins)
     end
 
+    if api_ctx.global_plugins then
+        core.tablepool.release("global_plugins", api_ctx.global_plugins)
+    end
+
     if api_ctx.curr_req_matched then
         core.tablepool.release("matched_route_record", api_ctx.curr_req_matched)
     end
--- a/apisix/plugin.lua
+++ b/apisix/plugin.lua
@@ -1525,17 +1525,37 @@ function _M.run_global_rules(api_ctx, global_rules, conf_version, phase_name)
                                                              global_rules,
                                                              conf_version)
 
-        local plugins = core.tablepool.fetch("plugins", 32, 0)
         local route = api_ctx.matched_route
         api_ctx.conf_type = "global_rule"
         api_ctx.conf_version = dummy_global_rule.modifiedIndex
         api_ctx.conf_id = dummy_global_rule.value.id
 
-        core.table.clear(plugins)
-        plugins = _M.filter(api_ctx, dummy_global_rule, plugins, route)
+        -- The filtered set depends only on the merged global rule (itself cached
+        -- per conf_version) and on the matched route, so it does not need
+        -- recomputing in every phase. This matters most for body_filter, which
+        -- runs once per response buffer. Route plugins are already reused this
+        -- way through api_ctx.plugins; global rules were not.
+        -- The route is part of the key because plugins with
+        -- run_policy == "prefer_route" are skipped when the route configures the
+        -- same plugin, and api_ctx.matched_route is replaced when a consumer's
+        -- configuration is merged in -- which happens after the rewrite phase
+        -- has run but before access.
+        -- The table is returned to the pool in http_log_phase.
+        local plugins = api_ctx.global_plugins
+        if not (plugins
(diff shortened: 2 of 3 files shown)


AlinsRan:
Force-pushed a reset of this branch. It now contains a single commit — the cross-phase reuse of the filtered global-rule set — and no longer touches `_M.filter()` at all.

The earlier revisions also added a fast path inside `_M.filter()` that iterated the resource's configured plugins instead of scanning every loaded one. I dropped it for two reasons:

1. **It broke auth plugins.** The fast path bypassed the slow path's loop, and with it this bookkeeping:

   ```lua
   if phase == "rewrite_in_consumer"
           and (not plugin_conf._from_consumer or plugin_obj.type == "auth") then
       plugin_conf._skip_rewrite_in_consumer = true
   end
   ```

   Without that marking, auth plugins ran a second time after a consumer merge, by which point the route's plugin config has been overwritten by the consumer's — which is validated against `consumer_schema` and carries none of the request-parsing fields. For jwt-auth that means `conf.header` is nil and `core.request.header(ctx, nil)` aborts the request with a 500. That was the 108 failures in `t/plugin/jwt-auth.t` on the previous run. I had a fix, but the underlying problem is that two copies of the same loop have to stay in sync forever.

2. **Its remaining value is unmeasured.** The caching in this PR already cuts `_M.filter()` calls per request from 9 to 1–2. How much a faster `filter()` is still worth on top of that was never measured — the two numbers I had came from different baselines and are not additive. Carrying the ordering-equivalence burden for an unknown gain is not a good trade.

If it turns out `filter()` is still hot after this, I would rather submit it separately, with the 2x2 measurement, and using a single index-driven path instead of two parallel ones: collect the configured plugins' positions in `local_plugins`, sort those integers, and walk one loop body. Since `local_plugins` is already sorted, ascending positions reproduce its order exactly — no comparator, and no special case for plugins that share a priority.

CI failures on the previous revision also included `nodejs.org` returning 502 during setup, unrelated to the code.

nic-6443 (approved):

membphis (approved):

shreemaan-abhishek (approved):

PR #13782, opened 2026-08-06
feat(ldap-auth-advanced): add group-based authorization

janiussyafiq:
### Description

Group-based authorization for the `ldap-auth-advanced` plugin:

- collect the authenticated user's groups, either by group search under `group_base_dn` or from the user entry's `memberOf` attribute (no extra round trip)
- `groups_required` (outer array ORs, inner array ANDs) is enforced after authentication; a denied user gets a 403, distinct from the 401 paths
- identity is exported to the upstream via `X-Authenticated-Username`, `X-Authenticated-User-Dn`, and `X-Authenticated-Groups`, each behind a `set_*_header` toggle (same idiom as `openid-connect`)

This PR also removes the plugin's Consumer support:

- with groups in play, mapping an authenticated user to a Consumer is unstable: a user who belongs to several groups cannot deterministically resolve to one Consumer
- external-directory authentication does not need Consumers, matching the existing external auth plugins (`openid-connect`, `forward-auth`, `authz-keycloak`) which attach none -- the directory is the identity store; users who want Consumer-backed LDAP keep `ldap-auth`
- attaching `ldap-auth-advanced` to a Consumer or Credential now fails with a clear error pointing at `ldap-auth`

The plugin is not part of any released version, so this is not a breaking change.

Part of #8958

### Checklist

### DIFF
--- a/apisix/consumer.lua
+++ b/apisix/consumer.lua
@@ -303,7 +303,6 @@ local plugin_unique_key_attrs = {
     ["jwt-auth"]   = "key",
     ["hmac-auth"]  = "key_id",
     ["ldap-auth"]  = "user_dn",
-    ["ldap-auth-advanced"] = "user_dn",
 }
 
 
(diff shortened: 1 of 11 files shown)


janiussyafiq:
Closing: per review feedback we're re-adding Consumer support with a deterministic matching strategy (user_dn / group_dn). The group-authorization work will be resubmitted combined with Consumer matching as a single PR.

Issue #13753, opened 2026-07-27
help request: Unable to remove or override `X-Forwarded-Host` header per-route using any plugin


mayankmurari:
### Description

I need to remove (or clear) the `X-Forwarded-Host` header for a specific API route before it reaches the upstream service. Despite trying every available approach, APISIX always sends the original `Host` header value as `X-Forwarded-Host` to the upstream.

### 1. `proxy-rewrite` with `headers.remove`

```json
{
  "plugins": {
    "proxy-rewrite": {
      "headers": {
        "remove": ["X-Forwarded-Host"]
      }
    }
  }
}
```

**Result:** `X-Forwarded-Host` still sent to upstream with original value.

### 2. `proxy-rewrite` with `headers.set` (empty string)

```json
{
  "plugins": {
    "proxy-rewrite": {
      "headers": {
        "set": {
          "X-Forwarded-Host": ""
        }
      }
    }
  }
}
```

**Result:** `X-Forwarded-Host` still sent to upstream with original value. Not overridden to empty string.

### 3. `proxy-rewrite` with `host` field

```json
{
  "plugins": {
    "proxy-rewrite": {
      "host": "my-upstream.example.com"
    }
  }
}
```

**Result:** `X-Forwarded-Host` unchanged.

### 4. Upstream `pass_host: rewrite` with `upstream_host`

```json
{
  "upstream": {
    "pass_host": "rewrite",
    "upstream_host": "my-upstream.example.com",
    "nodes": { "my-upstream.example.com:443": 1 },
    "scheme": "https"
  }
}
```

**Result:** `X-Forwarded-Host` still sent with the APISIX ingress hostname.

### 5. `serverless-pre-function` setting `ngx.var.var_x_forwarded_host`

```json
{
  "plugins": {
    "serverless-pre-function": {
      "phase": "before_proxy",
      "functions": [
        "return function(conf, ctx) ngx.var.var_x_forwarded_host = '' end"
      ]
    }
  }
}
```

**Result:** Plugin accepted (no errors), but `X-Forwarded-Host` still sent with original value.

### 6. Custom plugin in `before_proxy` phase

Created a custom Lua plugin that runs in the `before_proxy` phase and sets `ngx.var.var_x_forwarded_host = ''`.

**Result:** Plugin accepted, but `X-Forwarded-Host` still sent with original value.

## Expected Behavior

At least one of the following should work:
[...]

PR #13783, opened 2026-08-06
perf(core): skip the X-Forwarded-* rewrite when there is nothing to neutralize

AlinsRan:
## Problem

`is_trusted()` returns `false` whenever `apisix.trusted_addresses` is unset, and unset is the default — so on a default install **every** request runs the "untrusted peer" branch that #12551 added to `handle_x_forwarded_headers`: four `ngx.req.set_header()` calls plus a string parse. Configuring a trust boundary was the only way out, i.e. you had to opt *into* the feature to stop paying for it.

A second cost hides behind that branch. `set_upstream_x_forwarded_headers` reads three `http_x_forwarded_*` variables through `api_ctx.var`, and `core/ctx.lua`'s `__index` runs a `key:lower()` **and an `ngx.re.gsub()`** for every `http_*` key, then does not cache a nil result. A request carrying no `X-Forwarded-*` header — the common case — pays three PCRE calls per request, on the trusted path too.

## Fix

The expensive part is not writing the headers, it is *replacing* them. `set_header` walks the header list to find the old entry and remove it before inserting. On a request that carries none of `X-Forwarded-Proto/Host/Port/For` and `Forwarded`, the entry is known absent — that is the condition for taking the path — so appending is enough.

- `handle_x_forwarded_headers` takes a fast path when the request carries no forgeable value. It still injects `X-Forwarded-Proto/Host/Port`, so everything that reads them keeps seeing them, but with `add_header` rather than `set_header`. The rest of the slow path is a no-op on this input anyway — every `original_x_forwarded_*` value is nil, and clearing `Forwarded` or `X-Forwarded-For` has nothing to clear — so that stays skipped.
- The `$realip_remote_addr` lookup and the IP match are skipped when no trust boundary is configured; the header reads sit below the trusted early-return; `Forwarded` and `X-Forwarded-For` are only cleared when actually present.
- Request headers are read through `ngx.var` instead of `api_ctx.var`, avoiding the wrapper's per-key regex.

## Result

Eight rounds, alternating, single worker, wrk2 at saturation, upstream on localhost, worker on one P-core and the load generator on two others:

| variant | mean | range |
|---|---|---|
| master | 79,345 | 77,772–80,806 |
| **this PR** | **86,145** | 84,368–87,417 |
| an earlier revision that skipped the injection | 89,726 | 87,945–91,423 |

**+8.6%**, ranges disjoint across all eight rounds. The third row is what this branch looked like before review: +13.1%, but it changed what route `vars`, header-table plugins and the `r->headers_in` exits observe. Preserving that visibility costs 4.5 points, and this PR pays it.

(The first version of this description quoted +15%. That was measured with the worker and the load generator pinned to hyperthread siblings on a machine whose CPUs 0–11 are SMT pairs and 12–19 are E-cores. It was noise; every number above is from the corrected setup.)

## Behaviour

**Trust semantics are unchanged.** A forged `X-Forwarded-*` or `Forwarded` from an untrusted peer is still overridden or cleared exactly as before. An empty-valued header (`X-Forwarded-Proto:` with no value) reads back as `""`, truthy in Lua, so it takes the slow path — the fast path cannot be entered while carrying a forgeable value.

**Header visibility is unchanged.** Verified for each consumer that has one:

| observer | master | this PR |
|---|---|---|
| upstream through `location /` | `xfp=http xfh=localhost xfport=1984` | same |
| plugin reading `ngx.req.get_headers()` | same three values | same |
| mirror subrequest (`r->headers_in` copy) | same three values | same |
| route `vars: [["http_x_forwarded_proto","==","http"]]` | matches | matches |
| `Host` carrying an explicit port | `xfh=localhost:8443 xfport=8443` | same |

`@grpc_pass` and `@dubbo_pass` share the mirror's mechanism — `ngx_http_grpc_module` always passes request headers and `dubbo_pass_all_headers` is on — and the mirror is the one that can be driven over plain HTTP, so it stands in for them in the tests.

**One behaviour does change, and it is a fix.** A plugin-set `X-Forwarded-*` now reaches the upstream. `t/plugin/proxy-rewrite2.t` TEST 10 **fails on master** (`got: '1984', expected: '10080'`): `core.request.set_header` invalidates `ctx.var` under `http_x-forwarded-port` (hyphens, `core/request.lua`) while `core/ctx.lua`'s `__index` caches `http_x_forwarded_port` (underscores), so the value `handle_x_forwarded_headers` had cached won over the plugin's. Reading `ngx.var` bypasses the stale cache. Not a trust weakening — the value comes from plugin configuration, not client input. The underlying key mismatch is untouched here and deserves its own fix; it is also the mechanism behind reports like #13753.

Also, `set_upstream_x_forwarded_headers` now reads `ngx.var`, so a plugin that assigned `ctx.var.http_x_forwarded_proto` directly — cache only, without writing the real header — would no longer be propagated. No in-tree plugin does this.

## Also

`conf/config.yaml.example` puts the `trusted_addresses` example at the top level, between two `apisix:` entries. `apisix/cli/schema.lua` declares the option inside the `apisix` properties and `trusted-addresses.lua` reads `local_conf.apisix.trusted_addresses`, so a user who uncommented it verbatim would configure a key nothing reads, with no error to say so. Re-indented.

Follow-up to #12551. The plugin-override fix above is the mechanism behind #13753, though this PR only bypasses it rather than fixing the underlying `core/request.lua` cache-key mismatch.

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -713,11 +713,18 @@ local function handle_x_forwarded_headers(api_ctx)
     -- skipping the write would change what those see.
     --
     -- What can be skipped is the cost of *replacing* a header. `set_header` walks
-    -- the header list to find the old entry and removes it; here the entry is
-    -- known absent -- that is the condition for being on this path -- so appending
-    -- is enough. The rest of the slow path is a no-op on this input anyway: the
-    -- `original_x_forwarded_*` values are all nil, and clearing `Forwarded` or
-    -- `X-Forwarded-For` has nothing to clear.
+    -- the header list to find the old entry and removes it before inserting; here
+    -- there is no old entry, so appending is enough. The rest of the slow path is
+    -- a no-op on this input anyway: the `original_x_forwarded_*` values are all
+    -- nil, and clearing `Forwarded` or `X-Forwarded-For` has nothing to clear.
+    --
+    -- `add_header` appends unconditionally, so it is only correct while the
+    -- absence it relies on still holds. That is what the branch condition below
+    -- asserts, for all three headers at once and from `ngx.var`, which reflects
+    -- any earlier write to `r->headers_in` no matter who made it -- a second entry
+    -- would make `$http_x_forwarded_proto` read back as "http, http" rather than
+    -- nil, and this branch would not be taken. Keep the three `add_header` calls
+    -- directly under that check: nothing between them may touch a request header.
     if not (xf_proto or xf_host or xf_port or xf_for or forwarded) then
         local proto = api_ctx.var.scheme
         local http_host = ngx_var.http_host or api_ctx.var.host
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -712,19 +712,17 @@ local function handle_x_forwarded_headers(api_ctx)
     -- (`@grpc_pass`, `@dubbo_pass`, the mirror subrequest) all observe them, so
     -- skipping the write would change what those see.
     --
-    -- What can be skipped is the cost of *replacing* a header. `set_header` walks
-    -- the header list to find the old entry and removes it before inserting; here
-    -- there is no old entry, so appending is enough. The rest of the slow path is
-    -- a no-op on this input anyway: the `original_x_forwarded_*` values are all
-    -- nil, and clearing `Forwarded` or `X-Forwarded-For` has nothing to clear.
+    -- What is worth skipping is the rest of the slow path, which is a no-op on
+    -- this input: every `original_x_forwarded_*` value is nil, and clearing
+    -- `Forwarded` or `X-Forwarded-For` has nothing to clear.
     --
-    -- `add_header` appends unconditionally, so it is only correct while the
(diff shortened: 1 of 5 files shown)


shreemaan-abhishek (dismissed):

membphis (commented):
[P1] Preserve the established X-Forwarded-* visibility

This fast path changes observable behavior when an untrusted request carries none of these headers. Route variables, plugins that forward the request header table, and gRPC, Dubbo, or mirror paths that read `r->headers_in` can now see a different header set. Existing routes or policies may stop matching, and some upstreams may lose headers. Please preserve the existing request/header visibility while optimizing the expensive lookups, or treat this as an explicit breaking change with a migration path and compatibility coverage for every affected exit.

nic-6443 (dismissed):

AlinsRan (commented):
> [P1] Preserve the established X-Forwarded-* visibility

fixed.

PR #13784, opened 2026-08-06
fix(loki-logger): encrypt the headers field

nic-6443:
### Description

The `loki-logger` `headers` field is the documented place for the Loki / Grafana Cloud `Authorization` token, and it is copied verbatim into the outbound request. But `headers` is not listed in the plugin's `encrypt_fields`, so the token is stored in plaintext in etcd and returned unmasked from the Admin API — unlike the identically-shaped `elasticsearch-logger` `headers`, which is encrypted.

This adds `headers` to `loki-logger`'s `encrypt_fields`, matching `elasticsearch-logger`.

### Checklist

- [x] I have updated the documentation to reflect this change (no doc change needed — behavior is at-rest encryption of an existing field)

### DIFF
--- a/apisix/plugins/loki-logger.lua
+++ b/apisix/plugins/loki-logger.lua
@@ -111,6 +111,7 @@ local schema = {
         max_req_body_bytes = {type = "integer", minimum = 1, default = 524288},
         max_resp_body_bytes = {type = "integer", minimum = 1, default = 524288},
     },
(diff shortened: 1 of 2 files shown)


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13785, opened 2026-08-06
fix(sls-logger): verify the TLS certificate of the log server

nic-6443:
### Description

`sls-logger` performed `sock:sslhandshake(true, nil, false)` — no SNI and TLS certificate verification disabled — with no schema option to enable it. Meanwhile every RFC5424 record carries the plaintext Aliyun `access_key_id` / `access_key_secret` in its STRUCTURED-DATA. An on-path attacker between the gateway and the Aliyun SLS endpoint could present any certificate, complete the handshake, and capture the long-lived AccessKey plus all logged data.

This adds an `ssl_verify` boolean option (default `true`) and passes the configured `host` as the SNI, so the log server's certificate is verified by default. Users terminating TLS at an untrusted/self-signed endpoint can opt out with `ssl_verify: false`.

### Checklist

- [x] I have verified that this change is backward compatible (Aliyun SLS uses a publicly-trusted certificate; only untrusted/self-signed endpoints need `ssl_verify: false`)


shreemaan-abhishek (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13786, opened 2026-08-06
fix(error-log-logger): do not log the kafka sasl password

nic-6443:
### Description

When a kafka send fails, `error-log-logger` built its error string with `core.json.encode(config.kafka.brokers)`, which includes the decrypted `sasl_config.password`. That string is written at ERROR level by the batch processor, which is emitted at the default `warn` log level, so the kafka SASL password leaks into the error log — and this plugin's own purpose is to ship the error log onward. The same broker structure was also logged at INFO on each flush.

This redacts the broker list to `host`/`port` only before logging (both the failure error string and the INFO flush line).

### Checklist

- [x] I have updated the documentation to reflect this change (no user-facing config/behavior change)

### DIFF
--- a/apisix/plugins/error-log-logger.lua
+++ b/apisix/plugins/error-log-logger.lua
@@ -378,6 +378,17 @@ local function create_producer(broker_list, broker_config, cluster_name)
 end
 
 
+-- redact_brokers returns a copy of the kafka broker list carrying only host and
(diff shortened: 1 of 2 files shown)


shreemaan-abhishek (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13791, opened 2026-08-07
feat(ldap-auth-advanced): add group collection, group authorization and group-based Consumers

janiussyafiq:
### Description

Part 2 of 3 for the `ldap-auth-advanced` plugin (part 1: #13762). Adds group support:

- Collect the authenticated user's groups, either from the `memberOf` attribute on the user entry (default, no extra round trip) or by searching `group_base_dn` as the service identity (never as the end user; `sizeLimit 0` so groups are never truncated).
- `groups_required` authorization: outer OR of inner ANDs over group names, matched verbatim; an authenticated user failing it gets `403`, kept distinct from every `401` path.
- Export collected group names to the upstream in `X-Authenticated-Groups` (toggle `set_groups_header`, default `true`; the inbound header is always stripped).
- Consumers can now bind to a group instead of a user: `group_dn` (string, or array that must ALL contain the user), mutually exclusive with `user_dn`. An exact `user_dn` match always wins; among group matches the plugin picks deterministically (alphabetical group-DN order, more-specific array first) and logs a warning naming all candidates when more than one matches.

Two implementation notes for reviewers: the plugin resolves Consumers through a plugin-local index cached per consumer config version instead of `consumer_mod.find_consumer`, because the shared cache error-logs every consumer missing the lookup key, which would spam logs for legitimate group-only consumers (observable behavior, including the 401 messages, is unchanged). Separately, writing a `user_dn` consumer while `group_dn` consumers exist triggers the core write-time duplicate check to error-log "missing consumer auth credential" for the valid group consumers; cosmetic only, and fixing it means touching core beyond this PR's scope.

One existing test expectation was updated (`t/plugin/ldap-auth-advanced.t` TEST 52, one line): with `set_groups_header` defaulting `true`, the echoed headers now legitimately include the collected groups; the spoofed-inbound-value assertion is unchanged.

Related: #8958

### Checklist

- [x] I have verified that this change is backward compatible (new success-path `X-Authenticated-Groups` header is additive; existing consumer `user_dn` configs are unchanged)

### DIFF
--- a/apisix/plugins/ldap-auth-advanced.lua
+++ b/apisix/plugins/ldap-auth-advanced.lua
@@ -18,16 +18,26 @@ local core = require("apisix.core")
 local schema_def = require("apisix.schema_def")
 local auth_utils = require("apisix.utils.auth")
 local consumer_mod = require("apisix.consumer")
+local secret = require("apisix.secret")
 local ldap_client = require("resty.ldap.client")
 local ldap_protocol = require("resty.ldap.protocol")
 local ldap_filter = require("resty.ldap.filter")
 local ngx = ngx
 local ipairs = ipairs
+local pairs = pairs
 local type = type
+local tonumber = tonumber
 local ngx_decode_base64 = ngx.decode_base64
 local ngx_re_match = ngx.re.match
+local ngx_re_gsub = ngx.re.gsub
+local str_byte = string.byte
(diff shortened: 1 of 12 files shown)


janiussyafiq:
Status of this feature is to be put on hold since no demand for this feature request atm.

PR #13794, opened 2026-08-07
docs: add ldap-auth-advanced plugin documentation

janiussyafiq:
### Description

Adds the documentation pages (en + zh) for the `ldap-auth-advanced` plugin merged in #13762, which deferred docs to a follow-up PR. Covers the plugin's current feature set: search-then-bind authentication, service-account or anonymous search bind, `user_dn` Consumer association, `consumer_required`, `header_type`/`realm`, connection pooling, LDAPS/StartTLS, and the 401-vs-500 error split.

Related: #8958, #13762

### Checklist

- [x] I have added tests corresponding to this change (docs-only change; no tests apply)
- [x] I have verified that this change is backward compatible (docs-only)

### DIFF
--- a/docs/en/latest/plugins/ldap-auth-advanced.md
+++ b/docs/en/latest/plugins/ldap-auth-advanced.md
@@ -6,7 +6,7 @@ keywords:
   - Plugin
   - LDAP Authentication
   - ldap-auth-advanced
-description: The ldap-auth-advanced Plugin authenticates users against an LDAP directory using search-then-bind, so users do not need to be enumerated in APISIX.
+description: The ldap-auth-advanced Plugin authenticates users against an LDAP directory using search-then-bind; with consumer_required disabled, users do not need to be enumerated in APISIX.
 ---
 
 <!--
@@ -34,7 +34,7 @@ description: The ldap-auth-advanced Plugin authenticates users against an LDAP d
 
 ## Description
 
-The `ldap-auth-advanced` Plugin adds LDAP authentication to a Route or a Service. Unlike [`ldap-auth`](./ldap-auth.md), which binds with a DN assembled from the Consumer configuration, this Plugin *searches* the directory for the user first, then binds as the entry it found. Users therefore do not need to be enumerated in APISIX.
+The `ldap-auth-advanced` Plugin adds LDAP authentication to a Route or a Service. Unlike [`ldap-auth`](./ldap-auth.md), which binds with a DN assembled from the Consumer configuration, this Plugin *searches* the directory for the user first, then binds as the entry it found. When `consumer_required` is disabled, users do not need to be enumerated in APISIX; with the default `true`, each authenticated user must still map to a Consumer by its `user_dn`.
 
 On each request the Plugin:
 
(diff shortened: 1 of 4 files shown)


docs/en/latest/plugins/ldap-auth-advanced.md
```diff
@@ -0,0 +1,283 @@
...
+#
+-->
+
+<head>
+  <link rel="canonical" href="https://docs.api7.ai/hub/ldap-auth-advanced" />
+</head>
+
+## Description
+
+The `ldap-auth-advanced` Plugin adds LDAP authentication to a Route or a Service. Unlike [`ldap-auth`](./ldap-auth.md), which binds with a DN assembled from the Consumer configuration, this Plugin *searches* the directory for the user first, then binds as the entry it found. Users therefore do not need to be enumerated in APISIX.
```
kayx23: This is only true when `consumer_required` is `false`. With the default (`true`), `rewrite()` calls `find_consumer()` and rejects an authenticated LDAP user unless a Consumer with the same `user_dn` exists. Could we qualify this sentence (and the corresponding Chinese text), for example: “When `consumer_required` is disabled, users do not need to be enumerated in APISIX”?

docs/en/latest/plugins/ldap-auth-advanced.md
```diff
@@ -0,0 +1,283 @@
...
+3. Attaches a matching [Consumer](../terminology/consumer.md), unless `consumer_required` is `false`.
+
+The credential header uses the scheme word given by `header_type`, which defaults to `ldap` rather than `basic`, so the default expects `Authorization: ldap <base64(username:password)>`. Set `header_type` to `basic` to accept ordinary [basic access authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) instead.
+
+The Plugin distinguishes two failure modes, so an outage is never reported as a rejected credential:
+
+| Status | Cause |
+|--------|-------|
+| `401` | Missing, malformed, or rejected credentials; a username matching more than one entry; or `consumer_required` is `true` and no Consumer matches. Returned with a `WWW-Authenticate` header. |
+| `500` | The directory is unreachable or the `bind_dn` credentials were rejected. |
```
kayx23: This `500` row is not exhaustive. The implementation also returns `500` for LDAP search failures and for user-bind failures other than `invalidCredentials` (for example `busy` or `unavailable`), as well as transport/TLS/protocol errors. Could we broaden this to something like “LDAP transport, TLS, protocol, or server-side failures; or rejected `bind_dn` credentials” and mirror the change in Chinese?

docs/en/latest/plugins/ldap-auth-advanced.md
janiussyafiq: fixed

docs/en/latest/plugins/ldap-auth-advanced.md
janiussyafiq: fixed

docs/en/latest/plugins/ldap-auth-advanced.md
kayx23: Verified fixed in 843a631e: both locales now qualify the enumeration claim with `consumer_required=false` and explain the default Consumer mapping requirement.

docs/en/latest/plugins/ldap-auth-advanced.md
kayx23: Verified fixed in 843a631e: both locales now cover LDAP transport, TLS, protocol, search, server-result, and `bind_dn` credential failures.

kayx23 (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13796, opened 2026-08-10
feat(prometheus): stream active connections, termination status and bandwidth

AlinsRan:
### Description

A TCP or UDP proxy cannot be monitored from the Stream metrics APISIX has today. `apisix_stream_connection_total` is a per-Route connection counter, and three things are missing behind it:

- **No live concurrency.** Connection counts are only known once a session ends.
- **No way to tell a failure from a clean close.** NGINX reports a Stream `$status` of 200 for *every* failure that happens after the upstream connection is established — every post-connect path ends in `ngx_stream_proxy_finalize(s, NGX_STREAM_OK)` — so an idle timeout or a reset is indistinguishable from a normal close.
- **No throughput.** `$bytes_sent` and friends only exist at log time, and Lua has no handle on a live session, so a long-lived connection reports nothing until it ends.

This adds three metrics, all keyed by `listen_addr` rather than by Route: a session can end before any Stream Route is matched, and the byte counters come from NGINX, which only knows the listening address.

| Metric | Type | Labels |
|---|---|---|
| `apisix_stream_active_connections` | gauge | `listen_addr` |
| `apisix_stream_status` | counter | `code`, `listen_addr`, `node` |
| `apisix_stream_bandwidth` | counter | `listen_addr`, `type`, `side` |

`code` only ever takes values NGINX itself uses for Stream sessions — no synthetic codes. `$stream_session_reason` narrows 200 down to genuine closes and folds timeouts and resets onto 400/502. A plugin rejection is carried on `ctx` instead, because a Stream plugin rejects by closing the session (`plugin.lua`'s `run_plugin` calls `ngx_exit(1)`), so the code it returned never reaches `$status`; it is mapped through the same whitelist so an operator-supplied `rejected_code` cannot put an arbitrary value on the metric.

`apisix_stream_bandwidth` and `apisix_stream_active_connections` are read once a second from an NGINX shared memory zone, so they keep moving while a connection is open. Zone size is `nginx_config.stream.metrics_zone_size`, default `1m`.

### Runtime dependency

`apisix_stream_metrics_zone`, `$stream_session_reason` and `$stream_listen_addr` come from apisix-nginx-module 1.19.9, so `.requirements` moves to `1.3.14`, the first APISIX-Runtime carrying it.

Everything degrades on an older runtime rather than breaking: the collector logs once and stays off, and the status metric falls back to `$status`.

N/A

### Checklist


nic-6443 (approved):

membphis (commented):
**[P1] Older APISIX runtimes fail during configuration parsing instead of degrading gracefully**

`apisix_stream_metrics_zone` is provided only by apisix-nginx-module 1.19.9, but the template gates it only on `use_apisix_base`, which checks whether the module is present and does not check its version. apisix-runtime 1.3.11–1.3.13 still contain apisix-nginx-module 1.19.8, so enabling the Stream Prometheus plugin with the new default `metrics_zone_size` renders an unknown directive and makes startup or reload fail before the Lua `pcall(require, ...)` and variable fallbacks can run.

Please gate the directive on the actual runtime/module capability, omit it for versions before 1.19.9, and add a configuration/startup regression for an older runtime.

membphis (approved):

membphis:
> **[P1] Older APISIX runtimes fail during configuration parsing instead of degrading gracefully**
> 
> `apisix_stream_metrics_zone` is provided only by apisix-nginx-module 1.19.9, but the template gates it only on `use_apisix_base`, which checks whether the module is present and does not check its version. apisix-runtime 1.3.11–1.3.13 still contain apisix-nginx-module 1.19.8, so enabling the Stream Prometheus plugin with the new default `metrics_zone_size` renders an unknown directive and makes startup or reload fail before the Lua `pcall(require, ...)` and variable fallbacks can run.
> 
> Please gate the directive on the actual runtime/module capability, omit it for versions before 1.19.9, and add a configuration/startup regression for an older runtime.

ignore my message, this issue will not occur in the production environment.

shreemaan-abhishek (approved):

PR #13798, opened 2026-08-10
fix: reframe request body forwarded upstream in serverless plugins

shreemaan-abhishek:
### Description

The `openfunction`, `aws-lambda` and `azure-functions` plugins share `apisix/plugins/serverless/generic-upstream.lua`. It forwarded every inbound client header verbatim, including `Transfer-Encoding`, while supplying the request body already de-chunked by nginx (`ngx.req.get_body_data()`).

`resty.http` then keeps `Transfer-Encoding: chunked`, drops `Content-Length`, and writes the now-unframed body raw without re-chunking. The forwarded upstream request advertised chunked framing over a body that had none, so a keep-alive upstream could misread where the message ends.

Fix: strip the client's `Transfer-Encoding` and `Content-Length` before forwarding, so the HTTP client recomputes a correct `Content-Length` from the actual body and never emits chunked framing over an unframed body. The strip runs before each plugin's request processor, so `aws-lambda` SigV4 signs a clean header set.

Added an end-to-end case in `t/plugin/aws-lambda.t`: a chunked client request through the plugin, asserting the upstream receives a reframed body with a correct `Content-Length` and no `Transfer-Encoding`.

### Checklist


AlinsRan (dismissed):

nic-6443 (dismissed):

AlinsRan (approved):

nic-6443 (approved):

membphis (approved):
LGTM

PR #13800, opened 2026-08-10
feat(proxy-rewrite): support NGINX variables in regex_uri

### LINKED ISSUE #13140: feat: As a user, I want to use Nginx variables in proxy-rewrite.regex_uri
### Description

Currently, `proxy-rewrite` supports NGINX variable resolution in `uri` config parameter, **but not in `regex_uri`**. See [APISIX Docs | Plugins - proxy-rewrite](https://apisix.apache.org/docs/apisix/plugins/proxy-rewrite/#attributes) for reference.

Examples:

- 🟢 Nginx variable resolution in `uri`:

    ```yaml
    proxy-rewrite:
      uri: /api/$arg_name
    ```

- 🟢 Pattern variable resolution in `regex_uri`:

    ```yaml
    proxy-rewrite:
      regex_uri:
        - ^/api/(.*)/users/(.*)$
        - /v2/$1/external-users/$2
    ```

- 🔴 Nginx variable resolution in `regex_uri`:

    ```yaml
    proxy-rewrite:
      regex_uri:
        - ^/api/(.*)/users/(.*)$
        - /v2/$1/external-users/$arg_name
    ```
    
    This is **not** allowed and will throw an error.

## Question

Since `regex_uri` already supports variables resolution for regex patterns (e.g. patterns matched by `(.*)/(.*)` can be referenced with `$1` and `$2`), I was wondering: why does `uri` allow Nginx variables resolution, but `regex_uri` does not? Is this a design choice? Are there any contraindications in implementing this feature?

ruanweihong-aaa:
### Description

Currently, `proxy-rewrite.uri` supports NGINX variables, but the replacement part of `proxy-rewrite.regex_uri` only supports regular expression captures such as `$1` and `$2`.

This PR adds NGINX variable resolution support to `regex_uri` replacements.

For example:

```json
{
  "regex_uri": [
    "^/api/(.*)/users/(.*)$",
    "/v2/$1/external-users/$arg_name"
  ]
}
```

The replacement is processed in explicit stages:

1. Preserve literal dollar expressions such as `$$x`.
2. Resolve regular expression captures such as `$1` and `$2`.
3. Resolve NGINX variables such as `$arg_name` and variables stored in `ctx.var`.
4. Restore literal dollar expressions.

Schema validation also distinguishes NGINX variables from regular expression captures. This allows a route containing both forms to be saved successfully.

Existing `$1` and `$2` capture behavior and `$$` literal-dollar behavior remain compatible.

The English and Chinese `proxy-rewrite` documentation have also been updated to describe the new behavior.

Fixes #13140

### Tests

- Added `t/plugin/proxy-rewrite4.t`.
- Verified regular expression captures and NGINX variables in the same replacement.
- Verified existing `$1` and `$2` capture behavior.
- Verified literal `$$x` behavior.
- `t/plugin/proxy-rewrite4.t`: 6/6 passed.
- `luacheck`: 0 warnings and 0 errors for the modified Lua file.
- `lj-releng`: passed for the modified Lua file.
- Source-matched Docker BEFORE/AFTER verification: all six checks passed.

### Checklist


apisix/plugins/proxy-rewrite.lua
```diff
@@ -225,6 +247,14 @@ function _M.check_schema(conf)
...
                 local test_replacement = secret.is_secret_ref(replacement)
                                          and "" or replacement
+                if test_replacement ~= "" then
+                    local err
+                    test_replacement, err = escape_nginx_vars(test_replacement)
+                    if err then
+                        return false, "invalid regex_uri replacement(" ..
+                            replacement .. "): " .. err
+                    end
+                end
```
nic-6443, line 259: Why is this needed?

apisix/plugins/proxy-rewrite.lua
```diff
@@ -355,8 +385,14 @@ function _M.rewrite(conf, ctx)
...
 
+                local replacement = preserve_literal_dollars(conf.regex_uri[i + 1])
+                replacement = core.utils.resolve_var_with_captures(replacement, captures)
+                replacement = core.utils.resolve_var(replacement, ctx.var, escape_separator)
+                replacement = restore_literal_dollars(replacement)
                 local uri, _, err = re_sub(upstream_uri,
-                    conf.regex_uri[i], conf.regex_uri[i + 1], "jo")
+                    conf.regex_uri[i], function()
+                        return replacement
+                    end, "jo")
```
nic-6443, line 397: It's too complicated. Is there a simpler way?

apisix/plugins/proxy-rewrite.lua
ruanweihong-aaa: This is needed because `check_schema` compiles the replacement with `ngx.re.sub` to validate both the regex pattern and replacement syntax.

Without escaping NGINX variables, `$arg_name` is interpreted as a named PCRE capture and route creation fails with "failed to compile the replacement template", which is the original issue.

Escaping only NGINX variable markers as `$$` allows PCRE to keep validating regular captures such as `$1` and invalid replacement syntax, while treating `$arg_name` as a literal during schema validation. The existing invalid replacement test also depends on this validation.

I can add a comment here to explain this.

apisix/plugins/proxy-rewrite.lua
ruanweihong-aaa: The additional steps preserve three different replacement semantics:

1. `$1` and `$2` are regex captures.
2. `$arg_name` and values from `ctx.var` are NGINX variables.
3. `$$x` must remain the literal string `$x`.

Resolving NGINX variables directly would also consume `$1`, while letting `ngx.re.sub` process the replacement after variable resolution could interpret dollar signs from resolved values again.

I also tested resolving variables after `ngx.re.sub`, but at that point an escaped NGINX variable and an existing literal `$$x` both become `$...`, so they can no longer be distinguished.

I can move these stages into a focused helper to make the rewrite path easier to read. Would that address the concern, or would you prefer a different replacement behavior for literal `$$`?

apisix/plugins/proxy-rewrite.lua
ruanweihong-aaa: Added an inline comment in `d1b362b` explaining that schema validation must keep validating PCRE captures without treating NGINX variables as named captures.

PR #13805, opened 2026-08-11
fix(ldap-auth): key the consumer lookup on the escaped bind DN

shreemaan-abhishek:
### Description

`ldap_authenticate()` returns the bind DN it assembled as its third value, with the username escaped per [RFC 4514 §2.4](https://datatracker.ietf.org/doc/html/rfc4514#section-2.4). The plugin discarded that value and rebuilt the same string by raw concatenation:

```lua
local user_dn = conf.uid .. "=" .. user.username .. "," .. conf.base_dn
```

The two forms diverge as soon as the username carries a character that has structural meaning in a distinguished name (`,` `+` `=` `<` `>` `;` `"` `\`). An entry whose RDN value holds a comma binds against `cn=comma\,user,ou=users,dc=example,dc=org`, a single RDN, and is then looked up under `cn=comma,user,ou=users,dc=example,dc=org`, a two-component path. So the request either matches no consumer or matches one that belongs to a different directory entry.

This PR uses the DN the client already produced instead of rebuilding it, which keeps the bind target and the consumer key the same string by construction.

**Backward compatibility:** a consumer whose `user_dn` was written in the unescaped form to match the previous behaviour stops matching. This only affects directory entries whose RDN value contains one of the characters above, and the doc note added here spells out the escaped form to use. Every DN without those characters is unchanged.

**Tests:** the CI directory fixture gains an entry whose `cn` value is `comma,user`. Two cases assert that a consumer keyed on the escaped DN matches and one keyed on the raw expansion does not.

Fixes #

### Checklist

### DIFF
--- a/docs/en/latest/plugins/ldap-auth.md
+++ b/docs/en/latest/plugins/ldap-auth.md
@@ -44,6 +44,8 @@ For Consumer:
 | ------- | ------ | -------- | -------------------------------------------------------------------------------- |
 | user_dn | string | True     | User dn of the LDAP client. For example, `cn=user01,ou=users,dc=example,dc=org`. This field supports saving the value in Secret Manager using the [APISIX Secret](../terminology/secret.md) resource. |
 
+The Plugin builds the user dn as `<uid>=<username>,<base_dn>`, taking the username from the `Authorization` header. Characters that carry structural meaning in a distinguished name (`,` `+` `=` `<` `>` `;` `"` `\`) are escaped per [RFC 4514](https://datatracker.ietf.org/doc/html/rfc4514#section-2.4), so `user_dn` must use the escaped form to match. A user named `comma,user` under `ou=users,dc=example,dc=org` is configured as `cn=comma\,user,ou=users,dc=example,dc=org`.
+
 For Route:
 
 | Name     | Type    | Required | Default | Description                                                            |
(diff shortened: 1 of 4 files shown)


docs/en/latest/plugins/ldap-auth.md
```diff
@@ -44,6 +44,8 @@ For Consumer:
 | ------- | ------ | -------- | -------------------------------------------------------------------------------- |
 | user_dn | string | True     | User dn of the LDAP client. For example, `cn=user01,ou=users,dc=example,dc=org`. This field supports saving the value in Secret Manager using the [APISIX Secret](../terminology/secret.md) resource. |
 
+The Plugin builds the user dn as `<uid>=<username>,<base_dn>`, taking the username from the `Authorization` header. Characters that carry structural meaning in a distinguished name (`,` `+` `=` `<` `>` `;` `"` `\`) are escaped per [RFC 4514](https://datatracker.ietf.org/doc/html/rfc4514#section-2.4), so `user_dn` must use the escaped form to match. A user named `comma,user` under `ou=users,dc=example,dc=org` is configured as `cn=comma\,user,ou=users,dc=example,dc=org`.
```
nic-6443, line 47: Also add it to the Chinese documentation.

nic-6443 (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13806, opened 2026-08-11
fix(feishu-auth, dingtalk-auth): bind the authorization code to the session that started the login

shreemaan-abhishek:
### Description

`feishu-auth` and `dingtalk-auth` implement the OAuth 2.0 authorization code flow, but neither carried a `state` parameter across it. The plugin emitted the login redirect with no `state`, and on the way back it took the code straight out of the query string and exchanged it, with nothing tying that code to the browser that started the flow. A code obtained elsewhere was therefore accepted on any session.

This PR closes that gap:

- Before redirecting to `redirect_uri`, generate a random 16-byte `state`, store it in the session cookie, and append it to the redirect URL (`?state=` or `&state=` depending on whether the configured URL already has a query string).
- When the code arrives in the query string, require the `state` to match the one bound to the session, and respond `401 {"message":"Invalid state"}` when it does not. On successful authentication, the state is cleared when the updated session is saved.
- When the code arrives in the configured header (`X-Feishu-Code` / `X-DingTalk-Code`), no `state` is required. Those requests come from non-browser clients, which cannot be driven cross-site.

This mirrors what `authz-casdoor` already does in this repository (`apisix/plugins/authz-casdoor.lua:187-196` generates and stores, `:131-144` validates).

Also in this PR: the unsupported `sess:delete(...)` calls in `feishu-auth` were replaced with `sess:set(key, nil)`, the supported way to remove session keys.

#### Behaviour change

Deployments that use the query-string code path now need `state` to survive the round trip: `redirect_uri` receives it as a query parameter, and the login page is expected to pass it through to the IdP so that it comes back on the callback. This is the standard OAuth handling of `state`. Two consequences worth calling out:

- A user who was mid-flow across an upgrade (redirected before, returning after) has no `state` in their session and gets a `401`. Retrying the flow succeeds.
- Deployments that drive the plugin entirely through the code header are unaffected.

There is no new configuration option: an opt-out would just be a switch for turning the check back off.

No plugin schema fields were added or changed.

Fixes #

### Checklist


AlinsRan (approved):

nic-6443 (approved):

membphis (approved):
LGTM

PR #13808, opened 2026-08-12
feat(openid-connect): support OIDC Back-Channel Logout 1.0

### LINKED ISSUE #13750: bug: openid-connect plugin does not refresh access token when end server responds with 401

janiussyafiq:
### Description

Add an OIDC Back-Channel Logout 1.0 receiver to the `openid-connect` plugin, following the direction from the prior-art discussion in #13792: when the identity provider notifies the gateway that a user logged out elsewhere or an administrator revoked a session, the corresponding session cookie stops being accepted immediately, instead of remaining valid until its stored token expiry.

- New optional `backchannel_logout` config block: the provider POSTs a `logout_token` to `backchannel_logout.path`; after validation (JWKS signature, `iss`/`aud`/`iat`/`events`/`jti` claim checks per spec section 2.6, jti replay guard) the revoked `sid`/`sub` is written to a denylist, stored in `shm` (per-instance) or `redis` (shared across nodes, reusing `session.redis` when configured).
- Every session request is checked against the denylist: a revoked session is destroyed and the request then follows `unauth_action`; if the store is unreachable, the request fails with 503 and the session is kept.
- Covered by Keycloak E2E tests (sid and sub-only logout tokens, `unauth_action` variants, store-down behavior) plus a stub-IdP negative validation matrix; docs updated (en/zh).

Fixes #13750

### Checklist

- [x] I have verified that this change is backward compatible (the feature is off unless `backchannel_logout` is configured)

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -1420,6 +1420,49 @@ local function handle_backchannel_logout(conf)
 end
 
 
+-- Checks the session's identity against the revocation store. Returns true
+-- (revoked), false (clean), or nil plus an error when the store cannot be
+-- reached - no verdict is not a verdict.
+local function session_revoked_by_backchannel_logout(conf, response, session)
+    local id_token = response.id_token
+    if not id_token then
+        return false
+    end
+
+    if id_token.sid then
+        local revoked_at, err = bcl_store_get(conf,
+            bcl_denylist_key("sid", conf, id_token.iss, id_token.sid))
+        if err then
+            return nil, err
+        end
+        if revoked_at then
+            return true
+        end
+    end
+
+    if id_token.sub then
+        local revoked_at, err = bcl_store_get(conf,
+            bcl_denylist_key("sub", conf, id_token.iss, id_token.sub))
+        if err then
+            return nil, err
+        end
+        if revoked_at then
+            -- A sub entry means "log out every session this user had when
+            -- the logout was received": a session authenticated after it
+            -- stays valid. A missing timestamp kills the session, erring
+            -- toward revocation.
+            local auth_at = session:get("last_authenticated")
+                            or id_token.auth_time or id_token.iat
+            if not auth_at or auth_at <= revoked_at then
+                return true
+            end
+        end
+    end
+
+    return false
+end
+
+
 function _M.rewrite(plugin_conf, ctx)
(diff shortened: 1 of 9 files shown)


juzhiyuan:
Just approved the CI to run, thanks

apisix/plugins/openid-connect.lua
```diff
@@ -1117,6 +1192,335 @@ local function validate_claims_in_oidcauth_response(resp, conf)
...
+        return false, "shared dict \"bcl\" is missing"
+    end
+    -- safe_set: evicting an unexpired revocation to make room would silently
+    -- re-admit a revoked session, so a full dict must fail the write instead.
+    local ok, err = dict:safe_set(key, value, ttl)
+    if not ok then
+        return false, "failed to write to the shared dict: " .. err
+    end
+    return true
+end
```
nic-6443, line 1250: These codes that directly manipulate Redis data break the encapsulation of the OIDC and session libraries. We should not do this and need to find or add relevant features in the upstream libraries.

janiussyafiq: tracking: 
- https://github.com/bungle/lua-resty-session/issues/211
- https://github.com/zmartzone/lua-resty-openidc/issues/572
will move this PR to draft for now, will continue after upstreams support backchannel logout

PR #13809, opened 2026-08-12
fix(graphql-limit-count): measure each fragment once when computing query depth

shreemaan-abhishek:
### Description

`graphql-limit-count` computes a query's selection-nesting depth (used as the rate-limit cost) by expanding fragment spreads in place. The cycle guard (`visited`) was cleared after each descent, so a fragment referenced from more than one place was fully re-measured on every occurrence. When fragments reference other fragments repeatedly, the depth computation grows exponentially in the number of fragments, so a small document can become expensive to process in the access phase.

This memoizes each fragment's resolved depth so a fragment is measured once regardless of how many times it is spread. `visited` still guards fragment cycles; a new `memo` table caches the per-fragment result, keeping the traversal linear in the document size. The computed depth is unchanged for existing queries.

Added tests: a route whose query reuses fragments many times now returns normally instead of stalling the worker.

### Checklist

### DIFF
--- a/apisix/plugins/graphql-limit-count.lua
+++ b/apisix/plugins/graphql-limit-count.lua
@@ -90,17 +90,23 @@ local check_graphql_request = {
 -- Returns the maximum selection nesting depth of the GraphQL query AST.
 -- Fragment spreads are expanded in place using the provided fragment map;
 -- inline fragments are treated as transparent wrappers over their selections.
--- visited guards against fragment definition cycles; memo caches each
--- fragment's resolved depth so a fragment is measured once regardless of how
--- many times it is spread, keeping the traversal linear in the document size.
-local function node_depth(node, fragments, visited, memo)
+-- visited tracks the current recursion path: a spread back onto it is a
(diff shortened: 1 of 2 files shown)


AlinsRan (dismissed):

nic-6443 (dismissed):

membphis (commented):
**[P2] Fragment memoization caches a cycle-context-dependent depth**

`memo` is keyed only by fragment name, but the cached depth also depends on the current `visited` recursion stack. With mutually recursive fragments, the first traversal can cache an incomplete depth and reuse it from another path, making the computed cost traversal-order-dependent and potentially undercounting quota usage.

Please avoid memoizing cycle-dependent results, for example by excluding cyclic fragment components from memoization or rejecting cyclic fragment graphs, and add regression tests for mutually recursive fragments reached at different nesting depths and in different traversal orders.

AlinsRan (approved):

nic-6443 (approved):

membphis (approved):
LGTM

PR #13810, opened 2026-08-12
feat: add query-gateway plugin for safe caching of read-only body queries

### LINKED ISSUE #13807: feat: add query-gateway plugin for safe caching of read-only body queries
## Background

Many search, analytics, reporting, and query APIs accept a request body but are logically read-only. Today, clients and upstreams commonly use `POST` for those requests. While compatible, `POST` alone does not communicate that the operation is safe and idempotent, and it makes result caching an application-specific and potentially unsafe decision.

[RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html) defines the HTTP `QUERY` method for safe, idempotent requests that carry request content. It also permits caching QUERY responses, with a cache key that incorporates request content and relevant metadata.

The main need is therefore not only QUERY-to-POST compatibility. APISIX should provide a safe cache boundary for explicitly read-only, body-carrying queries even when both the client and the upstream use `POST` today. QUERY support is an optional compatibility and migration capability on top of that.

## Proposal

Add a `query-gateway` plugin for safe, cache-aware handling of read-only HTTP queries with request bodies.

The plugin would support:

1. Existing client `POST` requests, forwarded unchanged. Caching is opt-in and available only when the route explicitly marks those POST requests as read-only.
2. Client `QUERY` requests, forwarded as `POST` by default for compatibility with existing upstreams.
3. Native QUERY upstreams, enabled per route or upstream when support becomes available.
4. A local-memory cache by default, with optional Redis and Redis Cluster backends.
5. A short, configurable local-memory fallback when a remote cache backend is unavailable.

Example configuration:

```yaml
plugins:
  query-gateway:
    preserve_original_method_header: true
    original_method_header: X-Original-Method

    query:
      # Default: client QUERY -> upstream POST
      upstream_method: post

      # For a native QUERY upstream:
      # upstream_method: query

    post:
      # Existing POST requests remain POST.
      # Cache is enabled only when both options are explicit.
      cache_enabled: true
      read_only: true

    cache:
      enabled: true
      backend: local
      ttl: 30
      fallback_ttl: 5
```

## Cache safety model

Caching must be conservative. The cache key should include the route, target URI, client method, request content and content type, relevant content-negotiation metadata, approved cookie values, and applicable identity context.

The plugin should bypass cache for requests that include:

- `Authorization` or `Range`;
- `Cache-Control: no-store` or `no-cache`;
- `Pragma: no-cache`;
- cookies outside an explicit allow-list;
- a missing `Content-Type`;
- request or response bodies beyond configured limits.

It should not store responses that include:

- `Set-Cookie`, `WWW-Authenticate`, `Proxy-Authenticate`, or `Content-Range`;
- `Cache-Control: private`, `no-store`, `no-cache`, `max-age=0`, or `s-maxage=0`;
- `Vary` values outside a conservative allow-list.

## Use cases

- Elasticsearch and OpenSearch search endpoints;
- ClickHouse analytical queries;
- reporting, filtering, and dashboard APIs;
- GraphQL or JSON-based read-only query APIs;
- gradual migration from POST-based queries to RFC 10008 QUERY without changing the external API contract;
- sharing query results across APISIX instances with Redis or Redis Cluster.

## Proposed validation

The feature should include unit tests and a separate integration profile that builds APISIX from source and exercises a real Elasticsearch upstream. The profile should verify:

- QUERY and explicitly read-only POST cache `MISS -> HIT` flows;
- cache bypass for credential-bearing requests;
- negative `Content-Type` cases;
- that cache hits do not reach the upstream;
- correlation of client, APISIX, and upstream logs with `X-Opaque-ID`.

## Scope and non-goals

This proposal does not attempt to:

- add native QUERY support to every upstream or to the Nginx/OpenResty core;
- cache arbitrary POST traffic or write operations;
- cache user-specific or session-bound responses by default;
- define a global distributed-cache invalidation protocol.

Feedback is especially welcome on the cache-safety rules, the read-only POST opt-in model, and Redis/Redis Cluster fallback behavior.

fatihbm:
### Description

Adds the `query-gateway` plugin, an opt-in cache boundary for read-only HTTP requests that carry query content in the request body.

The plugin supports both client `POST` and RFC 10008 `QUERY` requests:

- Client `POST` requests are preserved when sent upstream.
- Client `QUERY` requests are forwarded as `POST` by default for upstream compatibility.
- Routes with upstream `QUERY` support can preserve the `QUERY` method.
- Equivalent `POST` and `QUERY` query payloads produce the same canonical cache key.
- Cache admission is conservative and explicit; arbitrary POST traffic is never cached.
- Local-memory caching is supported by default, with optional Redis and Redis Cluster backends.
- Redis failures use a bounded local-memory fallback.
- Cache safety rules exclude unsafe responses and requests with relevant cache-control, authorization, cookie, or set-cookie conditions.

This is useful for search, analytics, reporting, Elasticsearch, and ClickHouse-style read-only requests. These workloads often use `POST` because they require a request body, even when their semantics are safe and idempotent. The plugin enables safe caching today without requiring immediate native `QUERY` support from every upstream.

Fixes #13807

### Validation

- Added unit and integration coverage for request method handling, canonical cache keys, cache admission, cache hit/miss behavior, bypass behavior, and fallback behavior.

- Added Test::Nginx mock-upstream coverage for request method handling, canonical cache keys, cache admission, cache hit/miss behavior, bypass behavior, and fallback behavior.

- Verified client `POST` and `QUERY` scenarios, upstream `POST` and native `QUERY` forwarding modes, and cache bypass cases.

### Checklist


apisix/plugins/query-gateway/cache.lua
```diff
@@ -0,0 +1,455 @@
...
+    if not ok then
+        mark_backend_failure(conf, err)
+        return local_set(key, value, conf.fallback_ttl)
+    end
+
+    return true
+end
+
+local function has_directive(value, directive)
+    return value and ngx.re.find(lower(value), "(?:^|,)\\s*" .. directive .. "(?:\\s|,|=|$)", "jo")
```
lilien1010, line 257: **Blocker — these directive checks never match.** `ngx.re.find` uses PCRE, but the callers pass Lua-pattern escapes (`"no%-store"`, `"no%-cache"`, `"max%-age=0"`, `"s%-maxage=0"`). In PCRE, `%` is a literal character, so the regex searches for the literal string `no%-store` and never matches `no-store`.

Practical impact:
- a request with `Cache-Control: no-store` / `no-cache` does **not** bypass the cache;
- a response with `Cache-Control: no-store` / `no-cache` (without `max-age`) **is still cached** — contradicting the "Cache Safety" section of the doc. (`max-age=0` / `s-maxage=0` only happen to be caught by the separate — correctly written — `ngx.re.match` TTL clamp.)

`-` needs no escaping in PCRE, so passing plain `"no-store"` etc. is enough. Please also add test cases for `Cache-Control: no-store`/`no-cache` on both the request and response side — the current suite has no Cache-Control coverage at all, which is why this slipped through.

apisix/plugins/query-gateway/cache.lua
```diff
@@ -0,0 +1,455 @@
...
+    local content_length = tonumber(headers["content-length"])
+    if content_length and content_length > conf.max_request_body_size then
+        return nil, "request body exceeds cache limit"
+    end
+
+    return cookie_key
+end
+
+function _M.fetch(conf, ctx)
+    local headers = core.request.headers(ctx)
```
lilien1010, line 345: `core.request.headers()` (and `ngx.resp.get_headers()` in `header_filter`) return a **table** for repeated headers. A client sending two `Cookie` headers makes `cookie:gmatch(...)` throw; a duplicated `Content-Type` breaks the `concat` building `key_material`; a duplicated upstream `Cache-Control` makes `lower(value)` throw in `header_filter`. That is an externally triggerable 500.

Please normalize table values (e.g. join with `", "`) or simply bypass the cache when any relevant header arrives as a table.

apisix/plugins/query-gateway/cache.lua
```diff
@@ -0,0 +1,455 @@
...
+}
+
+local function sha256_hex(value)
+    local sha256 = resty_sha256:new()
+    sha256:update(value)
+    return to_hex(sha256:final())
+end
+
+local function shared_dict()
+    return ngx.shared["query-gateway-cache"]
```
lilien1010: `ngx.shared["query-gateway-cache"]` has no nil guard. With a user-maintained `config.yaml` that doesn't pick up the new default `lua_shared_dict`, every request on a cache-enabled route fails with `attempt to index a nil value`. Notably, this PR's own integration profile has to declare the dict via `http_configuration_snippet` in its `conf/config.yaml`.

Please degrade gracefully when the dict is missing (log once and treat the cache as disabled).

apisix/plugins/query-gateway.lua
```diff
@@ -0,0 +1,186 @@
...
+    local post_cache_enabled = conf.post and conf.post.cache_enabled and conf.post.read_only
+    if cache_conf and cache_conf.enabled and (method == "QUERY" or post_cache_enabled) then
+        local entry, status = cache.fetch(cache_conf, ctx)
+        if entry then
+            ctx.query_gateway_cache_hit = true
+            return cache.serve(entry)
+        end
+
+        if method == "QUERY" and status == "missing content-type" then
+            return 400
```
lilien1010: Enabling the cache changes request semantics here: a QUERY without `Content-Type` gets a 400, while the same request passes through untouched when `cache.enabled` is false. Turning on a cache shouldn't make previously working requests fail — suggest bypassing the cache for this case instead of rejecting the request.

apisix/plugins/query-gateway/cache.lua
```diff
@@ -0,0 +1,455 @@
...
+    local entry, err = core.json.decode(value)
+    if not entry then
+        core.log.warn("invalid query cache entry: ", err)
+        return nil, "invalid entry"
+    end
+
+    return entry, "hit"
+end
+
+function _M.serve(entry)
```
lilien1010, line 398: Cache hits replay the stored `Date` header and never emit `Age`, so clients computing freshness per RFC 9111 §4.2.3 will over-estimate how fresh the response is. Consider storing the entry creation time and emitting an `Age` header on hits (or refreshing `Date`).

apisix/plugins/query-gateway/cache.lua
```diff
@@ -0,0 +1,455 @@
...
+    local key, ttl = ctx.query_gateway_cache_key, entry.ttl
+    local cache_conf = conf
+    ctx.query_gateway_cache_entry = nil
+
+    if cache_conf.backend == "local" then
+        backend_set(cache_conf, key, payload, ttl)
+        return
+    end
+
+    local ok, err = ngx.timer.at(0, function(premature)
```
lilien1010: One `ngx.timer.at(0, ...)` per cache miss can exhaust `lua_max_pending_timers` (default 1024) under load, after which cache stores silently stop with only a warn log. At minimum this deserves a note in the doc; batching writes or reusing a single background worker would be more robust.

apisix/cli/config.lua
```diff
@@ -254,6 +256,7 @@ local _M = {
     "proxy-mirror",
     "graphql-proxy-cache",
     "proxy-rewrite",
+    "query-gateway",
```
lilien1010: Nit: this list is kept in plugin-priority order (`proxy-mirror` 1010 → `graphql-proxy-cache` 1009 → `proxy-rewrite` 1008 → `workflow` 1006). With `priority = -1001`, `query-gateway` belongs after `example-plugin` (0), not here. (The placement in `t/admin/plugins.t` is already correct.)

lilien1010 (commented):
Thanks for the contribution — the overall engineering here is solid: strict schema with `encrypt_fields`, correct reuse of `apisix/utils/redis.lua`, a thoughtful cache key (consumer identity, route scope, body SHA-256, negotiation headers, cookie allowlist), the circuit-breaker fallback, and the trace-correlated integration profile are all appreciated.

That said, a few things need attention before this can move forward. The inline comments cover the correctness issues (one of them is a real cache-safety bug). In addition:

- **`ui/.gitkeep` is unrelated to this PR — please remove it.**
- **Relationship with `proxy-cache`**: APISIX already ships `proxy-cache` and `graphql-proxy-cache`. Please add a short section to the doc explaining why body-aware QUERY caching is a separate plugin rather than an extension of `proxy-cache`'s cache key — this question will come up in review anyway.
- **CI cost**: the dedicated workflow builds the full APISIX docker image and pulls an Elasticsearch image on every PR touching these paths. Please consider folding it into the existing CI, and pin `actions/upload-artifact` by commit SHA like the checkout step.
- Minor: no zh doc yet (fine to follow up); new plugins conventionally start at `version = 0.1`; `parse_cookie` captures trailing spaces in cookie names, so `a = 1` is treated as an unallowlisted cookie and bypasses the cache.

fatihbm:
> Thanks for the contribution — the overall engineering here is solid: strict schema with `encrypt_fields`, correct reuse of `apisix/utils/redis.lua`, a thoughtful cache key (consumer identity, route scope, body SHA-256, negotiation headers, cookie allowlist), the circuit-breaker fallback, and the trace-correlated integration profile are all appreciated.
> 
> That said, a few things need attention before this can move forward. The inline comments cover the correctness issues (one of them is a real cache-safety bug). In addition:
> 
> * **`ui/.gitkeep` is unrelated to this PR — please remove it.**
> * **Relationship with `proxy-cache`**: APISIX already ships `proxy-cache` and `graphql-proxy-cache`. Please add a short section to the doc explaining why body-aware QUERY caching is a separate plugin rather than an extension of `proxy-cache`'s cache key — this question will come up in review anyway.
> * **CI cost**: the dedicated workflow builds the full APISIX docker image and pulls an Elasticsearch image on every PR touching these paths. Please consider folding it into the existing CI, and pin `actions/upload-artifact` by commit SHA like the checkout step.
> * Minor: no zh doc yet (fine to follow up); new plugins conventionally start at `version = 0.1`; `parse_cookie` captures trailing spaces in cookie names, so `a = 1` is treated as an unallowlisted cookie and bypasses the cache.

Thank you for the thorough review. I have pushed follow-up commits addressing the unrelated file, proxy-cache documentation, CI cost, version, and cookie parsing points. The Elasticsearch-specific workflow has been removed in favor of mock-upstream coverage in the existing test suite. I will follow up on the inline comments separately

apisix/plugins/query-gateway/cache.lua
fatihbm: Fixed in the follow-up commit. The directive checks now use valid PCRE patterns, and regression coverage was added for request and response Cache-Control: no-store / no-cache.

apisix/plugins/query-gateway/cache.lua
fatihbm: Fixed. The cache now bypasses when a relevant request or response header is repeated, avoiding ambiguous normalization and preventing the reported errors. Regression coverage was added.

apisix/plugins/query-gateway/cache.lua
fatihbm: Fixed. A missing shared dict is now logged once and treated as a cache bypass, so cache-enabled routes continue to proxy normally.

apisix/plugins/query-gateway.lua
fatihbm: I kept the 400 behavior intentionally. RFC 10008 requires a server to fail a QUERY request without Content-Type; bypassing the cache would accept an invalid QUERY request. This check is therefore independent of cache.enabled.

apisix/plugins/query-gateway/cache.lua
fatihbm: Fixed. Cache entries now retain their storage time and hits return a computed Age value.

apisix/plugins/query-gateway/cache.lua
fatihbm: Fixed. Remote writes now use a bounded shared-dict queue with one drain timer per cache backend, rather than scheduling one timer per miss.

apisix/cli/config.lua
fatihbm: Fixed. query-gateway is now placed after example-plugin, consistent with its priority.

fatihbm:
Closing this PR based on the maintainer feedback

Issue #8952, opened 2023-02-27
bug: failed to find proto by id in grpc-transcode plugin


Baoyuantop:
### Current Behavior

When I was troubleshooting the issue https://github.com/apache/apisix-dashboard/issues/2662, I found that the `proto_id` field of the `grpc-transcode` plugin is of type `string/integer`. but the strange thing is that when configured as string, everything is fine, and when configured as integer, APISIX shows the following error log:

```
2023/02/27 07:47:13 [error] 51#51: *377378 [lua] grpc-transcode.lua:146: phase_func(): proto load error: failed to find proto by id: 1, client: 172.19.0.1, server: _, request: "GET /grpctest HTTP/1.1", host: "127.0.0.1:9080"
```

### Expected Behavior

The configuration should work correctly regardless of whether string or integer types are used. Otherwise, the documentation needs to be updated.

### Error Logs

```
2023/02/27 07:47:13 [error] 51#51: *377378 [lua] grpc-transcode.lua:146: phase_func(): proto load error: failed to find proto by id: 1, client: 172.19.0.1, server: _, request: "GET /grpctest HTTP/1.1", host: "127.0.0.1:9080"
```

### Steps to Reproduce

Follow the steps in https://apisix.apache.org/docs/apisix/plugins/grpc-transcode/#enabling-the-plugin
1. Create proto resource
2. Create Route with
```
curl http://127.0.0.1:9180/apisix/admin/routes/111 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -d '
{
    "methods": ["GET"],
    "uri": "/grpctest",
    "plugins": {
        "grpc-transcode": {
            "proto_id": 1,
            "service": "helloworld.Greeter",
            "method": "SayHello"
        }
    },
    "upstream": {
        "scheme": "grpc",
        "type": "roundrobin",
        "nodes": {
            "127.0.0.1:50051": 1
        }
    }
}'
```

### Environment

- APISIX version (run `apisix version`): 3.1.0
- Operating system (run `uname -a`): Linux cedc51dc68a5 5.15.49-linuxkit #1 SMP Tue Sep 13 07:51:46 UTC 2022 x86_64 GNU/Linux
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant: 3.0.0
[...]

Issue #3406, opened 2021-01-25
feat: support gRPC HTTP mapping annotation for grpc trancode plugin


hnlq715:
We can use `gRPC HTTP mapping annotation` to specifies the HTTP mapping for methods in protobuf as well, like this:

```protobuf
// Returns a specific bookstore shelf.
rpc GetShelf(GetShelfRequest) returns (Shelf) {
  // Client example - returns the first shelf:
  //   curl http://DOMAIN_NAME/v1/shelves/1
  option (google.api.http) = { get: "/v1/shelves/{shelf}" };
}

...
// Request message for GetShelf method.
message GetShelfRequest {
  // The ID of the shelf resource to retrieve.
  int64 shelf = 1;
}
```
* option (google.api.http) specifies that this method is a gRPC HTTP mapping annotation.
* get specifies that this method is mapped to an HTTP GET request.
* "/v1/shelves" is the URL path template (appended to your service's domain) that the GET request uses to call this method. The URL path is also known as the resource path because it typically specifies the "thing" or resource you want to use. In this case, all our Bookstore's shelf resources.

details can be found in https://cloud.google.com/endpoints/docs/grpc/transcoding

tokers:
Yes we could. But in which case we can use it? AFAIK the use case is for bulding a grpc gateway like what etcd did.

Another problem, does it support Lua ecosystem?

hnlq715:
This is a standard protocol which has no relations to specific languages.

The situation is indeed to build a grpc gateway, but we do not need to define any api bussiness related config in the gateway(apisix), but in the protobuf directly. And all we need to do is just to open/close the `grpc trancode` plugin switch.

Just like what envoy did.

tokers:
If i understood correctly, we need to refine these APIs from the protobuf  and build a route radixtree and match routes further when the grpc-transcode plugin executing but still use HTTP protocol.

It may has some benefits although i can't find any reasons why we need to use HTTP protocol in grpc transcode here. Also, I don't know whether the ecosystem in Lua is complete.

hnlq715:
You're right.

From the user side, like douyu, we strongly need this for better grpc api maintainance :-)

1. do not store protobuf in etcd.
2. do not consider too much api bussiness logic in gateway.
3. just link services to the specific protobuf set in gateway, and everything goes simple.

Baoyuantop:
Closed due to prolonged lack of progress. If you're still interested in this feature, please let us know.

PR #13811, opened 2026-08-12
fix(grpc-transcode): resolve proto_id given as an integer

alirezashamsabad:
### Description

`id_schema` allows a resource id to be either a string or an integer, and the
`grpc-transcode` docs type `proto_id` as `string/integer`. But `create_proto_obj`
compares the configured `proto_id` against the stored id with `==`, and in Lua
`1 == "1"` is false. The admin API stores the id as a string, so any Route
configuring `"proto_id": 1` fails to resolve its proto:

```
proto load error: failed to find proto by id: 1
```

Reproducer against master: two Routes over the same proto, one with the id as a
string and one as an integer.

```shell
curl "http://127.0.0.1:9180/apisix/admin/routes/1" -H "X-API-KEY: $admin_key" -X PUT -d '
{
  "uri": "/str",
  "plugins": { "grpc-transcode": { "proto_id": "1", "service": "helloworld.Greeter", "method": "SayHello" } },
  "upstream": { "scheme": "grpc", "type": "roundrobin", "nodes": { "127.0.0.1:10051": 1 } }
}'

curl "http://127.0.0.1:9180/apisix/admin/routes/2" -H "X-API-KEY: $admin_key" -X PUT -d '
{
  "uri": "/int",
  "plugins": { "grpc-transcode": { "proto_id": 1, "service": "helloworld.Greeter", "method": "SayHello" } },
  "upstream": { "scheme": "grpc", "type": "roundrobin", "nodes": { "127.0.0.1:10051": 1 } }
}'
```

`/str` transcodes, `/int` returns an error and logs `failed to find proto by id: 1`.

This compares both sides as strings, matching how `graphql-proxy-cache` and
`upstream.lua` already normalize ids.

#### Which issue(s) this PR fixes:

Fixes [#8952](https://github.com/apache/apisix/issues/8952), which was closed as
stale without a fix. The report is still reproducible on master.

### Tests

Two blocks added to `t/plugin/grpc-transcode.t`: a Route configured with an
integer `proto_id`, and a request through it. Both fail without the change and
pass with it.

Developed with AI assistance; design decisions, review and testing are my own.

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change (not needed: the docs already type `proto_id` as `string/integer`; this makes the code match them)
[...]

PR #13812, opened 2026-08-12
feat(grpc-transcode): resolve service and method from google.api.http annotations

alirezashamsabad:
### Description

Today a Route using `grpc-transcode` is bound to exactly one gRPC method through
`service` and `method`, so a service with ten methods needs ten Routes. That duplicates
configuration the proto usually already declares via `google.api.http` annotations.

This adds an optional `use_http_annotations` flag. When enabled, the plugin reads those
annotations and selects the method whose path template and HTTP method match the request.
One Route can then serve a whole annotated service:

```json
{
  "uri": "/api/v1/*",
  "plugins": {
    "grpc-transcode": { "proto_id": "1", "use_http_annotations": true }
  },
  "upstream": { "scheme": "grpc", "type": "roundrobin", "nodes": { "127.0.0.1:50051": 1 } }
}
```

`GET /api/v1/items/42` reaches `ItemService/GetItem` with `id` bound to `42`, and
`POST /api/v1/items` reaches `ItemService/CreateItem`, without per-method Routes.

Relates to [#3406](https://github.com/apache/apisix/issues/3406), which was closed as
not planned before any implementation existed. This PR provides one, with tests and
documentation, so the trade-offs can be judged concretely.

### What is supported

- Path templates per the `google.api.http` grammar: literals, `{var}`, `{var=*}`,
  `{var=a/*/b}`, `**`, nested field paths such as `{user.id}`, and a trailing `:verb`
- `additional_bindings`
- `body` semantics: omitted means the payload is not read; `"*"` means the whole payload
  is the message; `"<field>"` maps the payload to that field and leaves siblings to the
  query string
- Path-captured values override query or body values for the same field
- Match precedence: more literal segments, then fewer variables, then service/method name.
  `:verb` is matched separately and is not part of the ranking
- While the flag is on, configured `service` and `method` are ignored

### Backwards compatibility

The flag defaults to `false`. When it is off, `service` and `method` stay mandatory, so
existing configs and validation errors are unchanged.

Behavior change in both modes: `access` returns `503` when the proto cannot be loaded.
Previously the request was forwarded untranscoded and often surfaced as a confusing
upstream `405`. The gateway now fails closed and logs the reason.

### Notes for reviewers

- **Binary descriptor sets only.** Annotations are preserved only in a descriptor set from
  `protoc --include_imports --descriptor_set_out=...`. A plain `.proto` uploaded to
  `/apisix/admin/protos` cannot resolve `google/api/annotations.proto`.
- **No new dependency.** Pinned lua-protobuf 0.5.3 exposes the extension as
  `method.options.http`.
- **Vendored test protos.** `t/grpc_server_example/proto/google/api/{annotations,http}.proto`
  are checked in for the fixture descriptor set (Apache-2.0 under Google's copyright;
  listed in `LICENSE` and ignored in `.licenserc.yaml`).
- **Limitation:** `body` supports a top-level field name only (not `body: "item.nested"`).
- **Out of scope:** an intermittent `failed to find proto by id` can still occur while
  the etcd config sync is in flight. It reproduces on legacy `service`/`method` routes
  as well, and its cause is distinct from the type mismatch in
  [#8952](https://github.com/apache/apisix/issues/8952). I have not root-caused it, so
  it is not addressed here.

### Tests

`t/plugin/grpc-transcode-http-annotations.t` covers path-template parsing and rejection,
end-to-end routing, `body` modes, `:verb`, path-over-query/body, `404`, `405` with
`Allow`, `400`, `503`, `proxy-rewrite`, and rebuilding the table after proto replacement.
`t/plugin/grpc-transcode*.t` passes (297 assertions).

Developed with AI assistance; design decisions, review and testing are my own.

### Checklist

- [x] I have verified that this change is backward compatible (If it is not backward compatible, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)


alirezashamsabad:
For anyone triaging the size:XXL label: of the 3,196 added lines, 1,160 are
generated .pb.go, 401 are the vendored google/api protos, and 799 are tests and
fixtures.The hand-written plugin change is 567 lines across four files.

Related: #13811 fixes a separate pre-existing bug in the same plugin
(grpc-transcode/proto.lua). The two branches are independent and do not conflict.

PR #13813, opened 2026-08-12
Fix for oidc temporarily_unavailable callback

### LINKED ISSUE #13776: bug: openid-connect error response results in 500 error from apisix

rhysyngsun:
### Description

This adds handling for the `temporarily_unavailable` [OAuth2 error response](https://www.rfc-editor.org/info/rfc6749/#section-4.1.2.1) by reattempting OIDC authentication. Prior to this fix all error callbacks resulted in a user-facing 500 error. It's probably not appropriate to retry on other error codes as those typically indicate either configuration issues or a more prolonged outage of the authorization service, neither of which are resolved by another attempt, hence the narrow scope of the handling added here. 

Fixes #13776

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -1322,6 +1322,27 @@ function _M.rewrite(plugin_conf, ctx)
                 return 302
             end
 
+            -- Transient authorization error from the identity provider: the ID
+            -- provider redirected to the redirect_uri with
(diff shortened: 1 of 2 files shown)


janiussyafiq:
Could you please fix CI?

juzhiyuan:
closed by https://github.com/apache/apisix/pull/13825

PR #13825, opened 2026-08-14
fix(openid-connect): handle temporarily_unavailable error redirects from the ID provider

### LINKED ISSUE #13776: bug: openid-connect error response results in 500 error from apisix

janiussyafiq:
### Description

When the ID provider redirects back to the redirect_uri with a standard OAuth2 error response (RFC 6749 section 4.1.2.1) such as `?error=temporarily_unavailable&error_description=authentication_expired` (e.g. Keycloak after its login session expired), the plugin returned a dead-end 500 to the browser.

The plugin now restarts the authentication flow for this transient error by redirecting the browser back to the original URL, where the retry typically completes without user interaction via the provider's SSO session. To keep this safe:

- the retry only applies when resty.openidc reported an unhandled request to the redirect_uri, so other failures on a request carrying that query argument are still surfaced, and a session-less request cannot be redirected to itself
- consecutive restarts of the authentication flow (shared with the existing state-mismatch restart) are capped at 3 per session, and the budget resets once a request authenticates

Other OAuth2 error codes (`access_denied`, `login_required`, ...) reflect a deliberate outcome and are not retried.

Fixes #13776

### Checklist

### DIFF
--- a/apisix/plugins/openid-connect.lua
+++ b/apisix/plugins/openid-connect.lua
@@ -1322,19 +1322,38 @@ function _M.rewrite(plugin_conf, ctx)
             -- Other OAuth2 error codes (access_denied, login_required, ...)
             -- reflect a deliberate outcome and are not retried.
             local restart_reason
+            local restart_url = target_url
             if err == STATE_MISMATCH_ERR then
+                -- state already matched by resty.openidc; no in-flight flow
+                -- for it, so its session-level original_url is all we have
                 restart_reason = "state mismatch (replayed or pruned callback)"
-            elseif core.string.has_prefix(err, UNHANDLED_REDIRECT_URI_ERR) then
+            elseif session and core.string.has_prefix(err, UNHANDLED_REDIRECT_URI_ERR) then
                 local uri_args = ngx.req.get_uri_args()
-                if uri_args.error == "temporarily_unavailable" then
+                -- resty.openidc bails on this path before validating state, so
+                -- match it here as it would: rejects a forged callback, and
+                -- recovers the original_url of the flow it belongs to (each
+                -- in-flight flow keeps its own since 1.9.0)
+                local authorization_state
+                if uri_args.error == "temporarily_unavailable" and uri_args.state then
+                    local states = session:get("authorization_states")
+                    authorization_state = states and states[uri_args.state]
+                    if not authorization_state
+                       and uri_args.state == session:get("state") then
+                        authorization_state = {
+                            original_url = session:get("original_url")
+                        }
+                    end
+                end
+                if authorization_state then
                     restart_reason = "authorization callback reported a " ..
                         "temporarily unavailable identity provider" ..
                         (type(uri_args.error_description) == "string" and
                             (" (" .. uri_args.error_description .. ")") or "")
+                    restart_url = authorization_state.original_url or target_url
                 end
             end
 
-            if restart_reason and target_url and session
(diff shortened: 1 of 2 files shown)


apisix/plugins/openid-connect.lua
```diff
@@ -1294,34 +1303,65 @@ function _M.rewrite(plugin_conf, ctx)
...
+            -- mid-navigation, so restart the authentication flow by sending
+            -- it back to the original URL instead of dead-ending with a 500.
+            -- Other OAuth2 error codes (access_denied, login_required, ...)
+            -- reflect a deliberate outcome and are not retried.
+            local restart_reason
+            if err == STATE_MISMATCH_ERR then
+                restart_reason = "state mismatch (replayed or pruned callback)"
+            elseif core.string.has_prefix(err, UNHANDLED_REDIRECT_URI_ERR) then
+                local uri_args = ngx.req.get_uri_args()
+                if uri_args.error == "temporarily_unavailable" then
```
nic-6443: This branch restarts the flow without ever validating `state`, which I think is worth fixing before merge.

`resty.openidc` bails out at `if not args.code or not args.state` in `openidc_authorization_response()`, so the state check right below it (`openidc_authorization_state()`) never runs on this path. That means any third party can trigger the 302 plus the session write with something like ``, and three of those burn the whole restart budget — the user's next genuine `authentication_expired` callback then dead-ends at 500, since the budget only resets on a successful authentication. mod_auth_openidc matches state *before* it looks at the `error` parameter, for exactly this reason.

Keycloak does send `state` back on the error redirect (`buildErrorRedirectUri()` adds it whenever the client sent one, and it is visible in the log in #13776), and the tests here already pass it, so the check is cheap — roughly what `openidc_authorization_state()` itself does:

```lua
local authorization_states = session:get("authorization_states") or {}
local authorization_state = authorization_states[uri_args.state]
    or (uri_args.state and uri_args.state == session:get("state") and {
        original_url = session:get("original_url")
    })
```

That also fixes a second thing. Since 1.9.0 each in-flight flow keeps its own `original_url` inside `authorization_states`, but this path falls back to the session-level `session:get("original_url")`, i.e. whichever flow started last. With two tabs mid-login, an error in tab A currently sends the browser to tab B's target; looking the state up gives you the right URL for free.

The final request in TEST 7 succeeding with `state=deadbeef` is the tell that nothing is checking it today.

janiussyafiq: fixed

nic-6443 (dismissed):

membphis (commented):
[P2] Do not redirect when persisting the retry counter fails

`auth_flow_restarts` only bounds the loop if its update reaches the session store. On `session:save()` failure, the code logs the error but still returns `302`; a Redis write failure or a cookie serialization failure leaves the old session state intact, so the browser can repeatedly restart the flow beyond the three-retry limit. The reset path has the symmetric problem: successful authentication can return normally while the old exhausted counter remains persisted.

Please make counter persistence part of the state transition: do not issue the retry redirect if saving the increment fails, and define a fail-closed or safe session-recovery path when clearing the counter fails. Please add a regression that forces `session:save()` to fail.

membphis (approved):

AlinsRan (approved):

nic-6443 (approved):

PR #13839, opened 2026-08-17
fix(data-mask): keep request header masking effective in the log phase

shreemaan-abhishek:
### Description

`data-mask` masks configured request headers from its log-phase handler by calling `core.request.set_header()`, which calls `ngx.req.set_header()`. That function returns early without touching the request header list whenever `r->headers_out.status` is already `400` ([`ngx_http_lua_headers_in.c`](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_headers_in.c#L766)):

```c
if (r->headers_out.status == 400 || r->headers_in.headers.last == NULL) {
    /* must be a 400 Bad Request */
    return NGX_OK;
}
```

The check was added in 2016 to avoid a segfault when writing headers on a request nginx itself rejected as malformed. It reads `headers_out.status`, which is unset in the rewrite/access phases where `set_header()` is normally called, but in the log phase holds the final response status. A well-formed request answered with `400` is therefore indistinguishable from a rejected one, and the masking silently does nothing while returning success.

The result: with `data-mask` and a logger plugin on the same route, the configured header was recorded unmasked whenever the final status was `400`. Every other status masked correctly, `401` and `404` included, since the check is an equality test.

The status only has to be `400` by the time the log phase runs; where it came from does not matter. An upstream `400`, a `response-rewrite` setting `status_code: 400`, and a `fault-injection` abort that never reaches the upstream all reach the same early return. Tests cover all three.

Both routes a logger takes to the request headers were affected:

- the default log entry, built from `ngx.req.get_headers()` in `log-util`
- a custom `log_format` referencing `$http_*`, since `modify_header()` invalidates the `ctx.var` cache and the read falls through to the native header

**How:** the plugin now keeps the masked headers on `ctx` and `log-util` reads them when present. The `core.request.set_header()` call is kept, so nothing changes for any other status and other plugins still observe the masked value through `core.request.header()`.

N/A

### Checklist

- [x] I have updated the documentation to reflect this change (no doc change needed: no configuration or documented behavior changes, the documented behavior now holds for all response statuses)

### DIFF
--- a/apisix/plugins/data-mask.lua
+++ b/apisix/plugins/data-mask.lua
@@ -20,6 +20,8 @@ local next      = next
 local type      = type
 local t_remove  = table.remove
 local re_sub    = ngx.re.sub
+local str_lower = string.lower
+local str_gsub  = string.gsub
 local core      = require("apisix.core")
 local jp        = require("jsonpath")
 
@@ -153,6 +155,17 @@ local function mask_table(tab, conf)
 end
 
 
+-- ngx.req.set_header() is a no-op once a response status exists, so the masked
+-- value is also kept on ctx for the loggers to read
+local function mask_header(ctx, headers, name, value)
+    core.request.set_header(ctx, name, value)
+
+    local key = str_lower(name)
+    headers[key] = value
+    ctx.var["http_" .. str_gsub(key, "-", "_")] = value or ""
+end
+
+
 -- jsonpath index of array starts from 0, lua table index starts from 1
 local function table_index(idx)
     if type(idx) == "number" then
@@ -217,6 +230,7 @@ function _M.log(conf, ctx)
     local body = ngx.req.get_body_data()
     local json_body
     local body_masked = false
+    local masked_headers
 
     if conf.request then
         for _, item in ipairs(conf.request) do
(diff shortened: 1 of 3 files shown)


membphis (commented):
**P2: `skywalking-logger` still bypasses masked request headers for 400 responses**

This patch makes `log-util` read `ctx.data_mask_headers`, but `skywalking-logger.log()` still reads `ngx.req.get_headers()["sw8"]` directly. When the final status is 400, `ngx.req.set_header()` leaves the native request header unchanged, so a configured `data-mask` rule for `sw8` is bypassed. An invalid original `sw8` can also be written to the APISIX error log by the existing parse warning, while a valid value is still used to build `traceContext`.

Please read `sw8` from the same masked header view (or a shared accessor) and add a 400-status regression test covering `remove`, `replace`, or `regex`. Since this is a localized P2 edge case, I am approving and leaving the final merge decision to the maintainers.

membphis (approved):
LGTM

AlinsRan (approved):

nic-6443 (approved):

PR #13841, opened 2026-08-18
test(ai-rate-limiting): wait for token usage to reach redis between requests

nic-6443:
### Description

`t/plugin/ai-rate-limiting.t` TEST 37 pipelines four requests against a redis-policy route and expects the fourth to be rejected. It is racy: `ai-rate-limiting` commits token usage from a log phase timer (`limit-count-redis.log_phase_incoming` schedules `timer_at(0)`), so a request can be evaluated against a budget that still looks unspent and the rejection lands one request late.

```
TEST 37: redis policy shares counter and rejects the 4th request - status code ok
         got: '200'
    expected: '503'
```

Pipelined requests leave no point at which to wait, so the block now sends the four requests sequentially and waits for each one's usage to reach redis before sending the next. Waiting for the commit beats sleeping a fixed amount: it takes only as long as the commit actually needs, and a sleep sized for a busy machine would eat into the rule's own time window.

The helpers go in `t/lib/test_redis.lua`, next to the `flush_all()` this suite already uses — `connect` and `release` are factored out of `flush_single` so both share them, and `connect` grew a `database` option because the route under test keeps its counters in redis db 1.

Verified by delaying the commit in `limit-count-redis.lua` (`timer_at(0)` -> `timer_at(0.3)`), which turns the race into a deterministic failure: TEST 37 fails on current `master` with exactly the mismatch above, and passes with this change. The whole file passes locally either way once the delay is removed (218 subtests).

### Checklist

### DIFF
--- a/t/lib/test_redis.lua
+++ b/t/lib/test_redis.lua
@@ -70,7 +70,26 @@ local function auth_if_needed(red, opts)
     return nil, err
 end
 
-local function flush_single(host, port, opts)
+local function release(red, host, port, opts)
+    local keepalive_pool = opts.keepalive_pool
+    if keepalive_pool == nil then
+        keepalive_pool = 0
+    end
+    if keepalive_pool == 0 then
+        local ok_close, close_err = red:close()
+        if not ok_close then
+            log_warn("failed to close redis connection ", host, ":", port, ": ", close_err)
+        end
+    else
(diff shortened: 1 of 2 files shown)


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13844, opened 2026-08-19
fix(jwe-decrypt): reject malformed tokens instead of returning 500

AlinsRan:
### Description

A crafted JWE token makes `jwe-decrypt` throw a Lua error, so the request ends with a 500 instead of the 400 the plugin already returns for invalid tokens:

- the header segment is decoded with `core.json.decode()` and only checked for truthiness. A header of `null` decodes to the truthy `cjson.null` userdata and a JSON scalar decodes to a number, so `jwe_obj.header_obj.kid` throws:
  ```
  attempt to index field 'header_obj' (a userdata value)
  ```
- an `iv`, `ciphertext` or `tag` that is not valid base64url decodes to nil. `aes:new()` then returns `nil, "iv is needed"`, and calling `decrypt()` on it throws:
  ```
  attempt to index local 'aes_default' (a nil value)
  ```

Both are reachable with a single request against any route using the plugin, e.g.

```
Authorization: Bearer bnVsbA..MTIzNDU2Nzg5MDEy._0DrWD0.vl-ydutnNuMpkYskwNqu-Q
Authorization: Bearer eyJhbGciOiJkaXIiLCJraWQiOiJ1c2VyLWtleSIsImVuYyI6IkEyNTZHQ00ifQ..!!!.!!!.!!!
```

The consumer secret has the same shape of problem. `check_schema()` skips the length check when data encryption is enabled — which is the default — so a `secret` that is not valid base64url is accepted, and `get_secret()` returns nil at request time, again ending in a 500. When the check does run (`data_encryption.enable_encrypt_fields: false`), `#base64.decode_base64url(conf.secret)` throws on such a secret instead of reporting a schema error.

This PR validates the decoded header, the token segments and the secret, and checks the `aes:new()` return value, so all of these paths end in the 400 the plugin already returns.

### Checklist

Tests: 5 new blocks in `t/plugin/jwe-decrypt.t` cover the `null` header, a scalar header, invalid base64url token segments, a consumer secret that cannot be decoded at request time, and the schema error for a non-base64url secret. All five fail on master and pass with this change. No documentation change is needed: the plugin already documents 400 for an invalid token, and this only stops the malformed cases from escaping as 500.

### DIFF
--- a/apisix/plugins/jwe-decrypt.lua
+++ b/apisix/plugins/jwe-decrypt.lua
@@ -19,6 +19,7 @@ local consumer_mod    = require("apisix.consumer")
 local base64          = require("ngx.base64")
 local aes             = require("resty.aes")
 local sub_str         = string.sub
+local type            = type
 local cipher          = aes.cipher(256, "gcm")
 
 local plugin_name     = "jwe-decrypt"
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13854, opened 2026-08-19
test(cli): wait for the pid file before reading it

nic-6443:
`t/cli/test_cmd.sh` fails intermittently in CI:

```
./t/cli/test_cmd.sh: line 39: logs/nginx.pid: No such file or directory
```

It reads `logs/nginx.pid` on the line right after `make run`, but `apisix start` returns once nginx has daemonized and the master writes its pid file after that — so the read races the write and the script dies under `set -e`.

Poll for the file first. The helper follows the shape of the existing `wait_for_tcp` in `t/cli/common.sh`. The other places that read the pid file (`test_load_full_data_init_worker.sh`, `test_limit_conn_redis_ttl.sh`) already wait on a request or guard with `-f`, so this is the only racy read.

Verified by running `./t/cli/test_cmd.sh` locally end to end.

### DIFF
--- a/t/cli/common.sh
+++ b/t/cli/common.sh
@@ -39,6 +39,27 @@ exit_if_not_customed_nginx() {
     openresty -V 2>&1 | grep apisix-nginx-module || exit 0
 }
 
+# wait_for_pidfile <path> [timeout_secs]
+# Poll until the pid file exists and is non-empty; default timeout 10s.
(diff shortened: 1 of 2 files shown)


AlinsRan (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13856, opened 2026-08-20
docs: sign Digest header in hmac-auth body validation example

### LINKED ISSUE #13395: docs: hmac-auth body validation example should sign the Digest header

CChongz1:
### Description

Updates the `hmac-auth` body-validation example so that the generated `Digest` header is included in the HTTP signature.

Fixes #13395

### Checklist

- [x] I have explained the changes added
- [ ] I have added tests corresponding to this change (documentation-only change)

### DIFF
--- a/docs/en/latest/plugins/hmac-auth.md
+++ b/docs/en/latest/plugins/hmac-auth.md
@@ -894,27 +894,29 @@ gmt_time = datetime.now(timezone.utc).strftime('%a, %d %b %Y %H:%M:%S GMT')
 # the date and any subsequent custom headers should be lowercased and separated by a
 # single space character, i.e. `<key>:<space><value>`
 # https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12#section-2.1.6
+# create the SHA-256 digest of the request body and base64 encode it
+body_digest = hashlib.sha256(body.encode('utf-8')).digest()
+body_digest_base64 = base64.b64encode(body_digest).decode('utf-8')
+digest_header = f"SHA-256={body_digest_base64}"
+
 signing_string = (
     f"{key_id}\n"
     f"{request_method} {request_path}\n"
     f"date: {gmt_time}\n"
+    f"digest: {digest_header}\n"
 )
 
 # create signature
 signature = hmac.new(secret_key, signing_string.encode('utf-8'), hashlib.sha256).digest()
 signature_base64 = base64.b64encode(signature).decode('utf-8')
 
-# create the SHA-256 digest of the request body and base64 encode it
-body_digest = hashlib.sha256(body.encode('utf-8')).digest()
-body_digest_base64 = base64.b64encode(body_digest).decode('utf-8')
-
 # construct the request headers
 headers = {
     "Date": gmt_time,
-    "Digest": f"SHA-256={body_digest_base64}",
+    "Digest": digest_header,
     "Authorization": (
         f'Signature keyId="{key_id}",algorithm="hmac-sha256",'
-        f'headers="@request-target date",'
+        f'headers="@request-target date digest",'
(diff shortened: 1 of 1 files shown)


docs/en/latest/plugins/hmac-auth.md
```diff
@@ -894,27 +894,29 @@ gmt_time = datetime.now(timezone.utc).strftime('%a, %d %b %Y %H:%M:%S GMT')
...
-
 # construct the request headers
 headers = {
     "Date": gmt_time,
-    "Digest": f"SHA-256={body_digest_base64}",
+    "Digest": digest_header,
     "Authorization": (
         f'Signature keyId="{key_id}",algorithm="hmac-sha256",'
-        f'headers="@request-target date",'
+        f'headers="@request-target date digest",'
```
kayx23, line 919: This now produces `headers="@request-target date digest"`, but the frozen sample output, success `curl`, and response body below still use `headers="@request-target date"` and `signature="rjS6NxOBKmzS8CZL05uLiAfE16hXdIpMD/L/HukOTYE="`.

For the documented timestamp `Fri, 06 Sep 2024 09:16:16 GMT`, the script prints:

```text
headers="@request-target date digest"
signature="LGBTz7bVQQWlkijeyDpEwJWo+ppwX735uRZk5F8KhmU="
```

Please refresh those samples (issue item 4). A short note that `validate_request_body` only checks `Digest` against the body — and that signing `digest` is what binds the body to the HMAC — would also match the issue.

kayx23 (changes requested):
The Python generator now signs `Digest`, which is the right fix for #13395. The rest of the example still shows the old unsigned request, and the same example in the Chinese page is unchanged.

Please also update `docs/zh/latest/plugins/hmac-auth.md` (same Python snippet, printed headers, curl, and sample response). APISIX does not require en+zh in one PR, but this example is duplicated there and would otherwise keep the incomplete body-validation flow.

PR #13862, opened 2026-08-21
fix(workflow): reject invalid case expressions and missing action conf

### LINKED ISSUE #13910: bug: workflow check_schema accepts invalid case expressions and 500s on missing action conf
### Current Behavior

workflow plugin check_schema has two validation bugs.

1、after expr.new(rule.case), the code inspects the wrong variable:

```lua
local expr, err = expr.new(rule.case)
-- still the schema-check result, which is already true
if not ok then
    return false, "failed to validate the 'case' expression: " .. err
end
```

2、action conf is optional in the JSON schema (actions items are arrays with minItems = 1, no requirement that index 2 exists). check_schema then does:

```lua
action[2]["_vid"] = idx
```

when the action is just ["return"] with no conf table, this indexes nil and the Admin API returns HTTP 500 instead of a validation error.

### Expected Behavior

1、Invalid case expressions (expr.new failure) must be rejected at schema-check time with failed to validate the 'case' expression: ...

2、Missing action configuration must be rejected with failed to validate the '<action>' action: configuration is required and must not 500 the Admin API.

### Error Logs

```txt
2026/09/02 08:20:14 [error] 49#49: *57321127 lua entry thread aborted: runtime error: /usr/local/apisix/apisix/plugins/workflow.lua:144: attempt to index a nil value
stack traceback:
coroutine 0:
        /usr/local/apisix/apisix/plugins/workflow.lua: in function 'check_schema'
        /usr/local/apisix/apisix/plugin.lua:992: in function 'check_single_plugin_schema'
        /usr/local/apisix/apisix/plugin.lua:1198: in function 'schema_plugin'
        /usr/local/apisix/apisix/admin/routes.lua:130: in function 'checker'
        /usr/local/apisix/apisix/admin/resource.lua:126: in function 'check_conf'
        /usr/local/apisix/apisix/admin/resource.lua:256: in function </usr/local/apisix/apisix/admin/resource.lua:243>
        /usr/local/apisix/apisix/admin/init.lua:235: in function 'handler'
        /usr/local/apisix//deps/share/lua/5.1/resty/radixtree.lua:1013: in function 'dispatch'
        /usr/local/apisix/apisix/init.lua:1203: in function 'http_admin'
```

### Steps to Reproduce

```curl
curl --location --request PUT 'http://172.18.0.21:9180/apisix/admin/routes/test' \
--header 'Content-Type: application/json' \
--header 'X-API-KEY: ••••••' \
--data '{
    "uri": "/fake",
    "name": "xxx",
    "plugins": {
        "workflow": {
            "rules": [
                {
                    "case": [
                        [
                            "uri",
                            "bad_case_????!!!!",
                            "/hello"
                        ]
                    ],
                    "actions": [
                        [
                            "return"
                        ]
                    ]
                }
            ]
        }
    },
    "upstream": {
        "type": "roundrobin",
        "scheme": "https",
        "pass_host": "node",
        "nodes": {
            "xxx:6443": 1
        }
    }
}'
```

### Environment

- APISIX version (run `apisix version`): 3.18

Arjen10:
### Description

check_schema inspected the wrong variable after expr.new, so illegal case values passed validation and matched every request at runtime. Missing action conf also indexed a nil value and 500'd the Admin API.

Fixes https://github.com/apache/apisix/issues/13910

### Checklist


janiussyafiq:
Thanks for fixing the CI lint, however a few points need to be addressed:
1. CI lint. `apisix/plugins/workflow.lua:144` uses the global `type` without `local type = type`, so `lj-releng` errors. You can run CI locally first to check if it passes.
2. Empty `actions: []` still crashes at runtime. Schema has `minItems` on each action but not on actions itself, so the validation loop is a no-op. Admin API returns passed, then every request on that rule gets 500 at `workflow.lua:174` (`rule.actions[1]` is nil).
3. Non-string action name 500s. `workflow.lua:141` concatenates `action[1]`, so `[true], [{}], [null]` crash with attempt to concatenate.
4. Test gap. New tests only unit-call check_schema. No Admin-API-level error_code: 400 assertion for the issue's repro, no empty-actions case. `t/plugin/workflow2.t:316` has the pattern to reuse.
  
Design:
One schema change beats the guard by declaring actions.items as [string, object] with minItems: 2. WDYT?

Arjen10:
> Thanks for fixing the CI lint, however a few points need to be addressed:
> 
> 1. CI lint. `apisix/plugins/workflow.lua:144` uses the global `type` without `local type = type`, so `lj-releng` errors. You can run CI locally first to check if it passes.
> 2. Empty `actions: []` still crashes at runtime. Schema has `minItems` on each action but not on actions itself, so the validation loop is a no-op. Admin API returns passed, then every request on that rule gets 500 at `workflow.lua:174` (`rule.actions[1]` is nil).
> 3. Non-string action name 500s. `workflow.lua:141` concatenates `action[1]`, so `[true], [{}], [null]` crash with attempt to concatenate.
> 4. Test gap. New tests only unit-call check_schema. No Admin-API-level error_code: 400 assertion for the issue's repro, no empty-actions case. `t/plugin/workflow2.t:316` has the pattern to reuse.
> 
> Design: One schema change beats the guard by declaring actions.items as [string, object] with minItems: 2. WDYT?

Thank you for the code review. i went throuth the [workflow plugin doc](https://apisix.apache.org/zh/docs/apisix/plugins/workflow/) again. i think your solution is better. i will fix it

Arjen10:
CI failed on pnpm setup under sudo, not on the tests

Arjen10:
merge https://github.com/apache/apisix/pull/13921

apisix/plugins/workflow.lua
```diff
@@ -45,9 +45,14 @@ local schema = {
                     },
                     actions = {
                         type = "array",
+                        minItems = 1,
```
nic-6443, line 48: I've checked the code. Currently it only supports one action, and the array length inside each action can only be 2. Therefore, a maxItem limit should also be added.

Arjen10: Done. added TEST 13 and TEST 14 to workflow2.t

nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13866, opened 2026-08-21
chore: bump APISIX-Runtime to 1.3.17

nic-6443:
### Description

Bumps `APISIX_RUNTIME` to 1.3.17, along with the version guard and the two `apisix-runtime-debug` checksums in `ci/linux-install-openresty.sh`.

1.3.17 changes two components relative to 1.3.16:

- **apisix-nginx-module 1.19.9 → 1.19.10** (api7/apisix-nginx-module#126). `ngx_http_apisix_set_upstream_ssl()` used to apply the caller-supplied trusted store only inside its `ctx->upstream_cert != NULL` branch, so `upstream.set_ssl_trusted_store()` was silently dropped without mTLS. nginx loads `proxy_ssl_trusted_certificate` into the SSL_CTX only when `proxy_ssl_verify` is on at configuration time, which APISIX never sets, so such a caller ended up with an empty trust store. This is what `upstream.tls.ca_certs` in #13863 needs in order to work on its own.
- **ngx_http_ffi_client v0.1.1 → v0.1.3**: request bodies containing CR or LF are no longer rejected, connection errors log at info level, the trust store falls back to `lua_ssl_trusted_certificate`, and names resolve through a resolver installed by the host application.

Checksums were taken from the published release artifacts:

```
d60067ba7a89cab6fca8e70994e4158fa8c414a569406e1692998be2567832a6  apisix-runtime-debug_1.3.17-0.debianbookworm-slim_amd64.deb
2db6619c6fa31128e7ea45b2cdcc56dd26d2b11d7b4a68600f4f04746cb34766  apisix-runtime-debug_1.3.17-0.debianbookworm-slim_arm64.deb
```

N/A

### Checklist

The whole test suite runs against the new runtime, which is the coverage a version bump can have; there is nothing to document.


nic-6443:
Folding this into #13863 instead. `upstream.tls.ca_certs` only works on 1.3.17, so keeping the bump separate meant #13863 could merge onto a runtime where its main path — verifying an upstream without mTLS — silently does nothing, and its tests had to pair every `ca_certs` case with a client certificate to pass. The bump and a test for the no-mTLS path now travel with the feature.

PR #13870, opened 2026-08-24
fix(ai-proxy): return 502 when a streaming upstream produces no output

AlinsRan:
### Description

A streaming AI request whose upstream answers with an SSE content type and then closes without writing a single body byte crashes the gateway in `balancer_by_lua`.

`body_reader()` returns EOF on its first call, the loop body never runs, nothing is written downstream, and `parse_streaming_response()` returns `nil`. `run_plugin()` only calls `core.response.exit()` when a plugin returns a status code, so returning `nil` leaves the request neither answered nor terminated. It falls through to `proxy_pass` and enters `balancer_by_lua`, where `upstream_conf` is always `nil` because `handle_upstream()` skips `set_upstream()` for `bypass_nginx_upstream` plugins:

```
failed to run balancer_by_lua*: apisix/balancer.lua:
    attempt to index local 'up_conf' (a nil value)
    in function 'pick_server'
    in function 'run'
    in function 'http_balancer_phase'
```

`header_filter` then fails as well, so the client gets no HTTP response at all and the connection is closed.

The `output_sent` guard added in the EOF branch already covers this, but only when a converter is active:

```lua
if converter and not output_sent then
    ...
    return 502, msg
end
return
```

An OpenAI-format client against an OpenAI-compatible provider matches `caps[client_protocol]`, so `converter` is `nil`, the guard is skipped, and the bare `return` below hands `nil` back up — the passthrough case (no protocol conversion) was never covered.

This drops the `converter` condition and words the log message by whether a converter was involved. 502 rather than 504: nothing timed out, the upstream closed cleanly and the request's own timeout never fired; 502 is also what this same guard already returns for the converter case.

`t/plugin/ai-proxy-empty-stream.t` mocks an upstream that answers `200` + SSE headers and closes with a zero-byte body, and asserts a 502 with no `up_conf` error.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/ai-providers/base.lua
+++ b/apisix/plugins/ai-providers/base.lua
@@ -576,13 +576,15 @@ function _M.parse_streaming_response(self, ctx, res, target_proto, converter, co
             res._upstream_bytes = bytes_read
             ctx.var.apisix_upstream_response_time = math.floor(
                 (ngx_now() - ctx.llm_request_start_time) * 1000)
-            if converter and not output_sent then
+            if not output_sent then
                 if flush_thread then
                     ngx.thread.kill(flush_thread)
                 end
-                local msg = "streaming response completed without producing "
-                            .. "any output; the upstream likely returned a "
-                            .. "different stream format than the converter expects"
+                local msg = converter
+                    and "streaming response completed without producing "
+                        .. "any output; the upstream likely returned a "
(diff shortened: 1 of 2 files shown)


membphis (approved):
LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13876, opened 2026-08-25
fix(ai-proxy): do not turn a streaming read error after partial output into a 5xx

nic-6443:
### Description

When an LLM upstream returns `200` with a streaming content type and at least one SSE event has already been forwarded downstream, a later `body_reader()` failure (`closed`, `timeout`) was still converted unconditionally into `500`/`504`.

The downstream response is already committed as `200` at that point, so the status cannot be changed. Two things go wrong:

- `ai-proxy` returns the 5xx to the plugin runner, which calls `core.response.exit()` on the committed `200` and logs `attempt to set ngx.status after sending out response headers`.
- `ai-proxy-multi` treats the 5xx as retryable and sends another billable LLM request after the client has already received partial output. `retry_on_error()` documents the assumption that "the failed attempt's body never reaches the client", which does not hold on this path.

The broken upstream connection was also handed to `set_keepalive()` instead of being closed.

`parse_streaming_response()` already tracks `output_sent` for the EOF branch. This uses it on the error branch as well, mirroring what the `max_stream_duration_ms` / `max_response_bytes` path already does once output has been committed: stop reading, close the upstream connection, run one last `body_filter` pass with `llm_request_done` set so plugins that buffer the whole stream flush what they hold, and return no status. No terminator is synthesized, so the client detects the truncation from the missing `[DONE]` / `message_stop` / `response.completed`.

A read error before the first downstream byte is unchanged: still `504` for a timeout and `500` otherwise, and still eligible for fallback. A read error that arrives after the protocol's completion event is treated as a complete response rather than an aborted one.

Setting `llm_request_done` on that last pass is also what would make `ai-aws-content-moderation` and `ai-aliyun-content-moderation` append a completion event to a stream that was cut short, reporting a truncated response as complete. That synthesis is now gated on `ctx.ai_stream_aborted`, which already marks every cut-short stream, so it also stops firing on the pre-existing stream-limit path.

### Scoping the completion flag to one attempt

`ai-proxy-multi` retries inside a single request context and only `ctx.ai_stream_aborted` was cleared between attempts. `ctx.var.llm_request_done` survived, so an attempt that set it while producing no downstream output left the next attempt looking finished before parsing any completion event of its own. That state is reachable: the Anthropic converter emits nothing for a `done` event while `state.is_first` is still true, so a `[DONE]`-only upstream stream sets `llm_request_done` with `output_sent` false, returns `502`, and is therefore retryable.

The inherited flag also makes the moderation plugins treat the retry as finished from its first chunk, appending a terminator to a stream that is still running. The flag is now reset at the start of every attempt, and the read-error path decides on a new attempt-local `protocol_completed` set only where the current attempt parses a `done` / `usage_and_done` event.

### Tests

`t/plugin/ai-proxy-stream-truncated.t` drives a raw upstream that commits `200` with `Content-Type: text/event-stream`, flushes one valid SSE event, then closes before the terminating chunk and before the completion event. It asserts that the client keeps `200` and the delivered event, that no status change is attempted after the headers are sent, that the broken connection is not offered to keepalive, that no fallback request is sent after partial output, that no `[DONE]` is synthesized, and that a read error before the first downstream byte still falls back.

`TEST 10` covers the cross-attempt case: attempt 1 serves a `[DONE]`-only stream through the Anthropic converter and produces no downstream output, the retry emits one content event plus a usage event and then truncates. It asserts the content is delivered and that no `message_stop` is synthesized.

### Checklist

### DIFF
--- a/apisix/plugins/ai-aliyun-content-moderation.lua
+++ b/apisix/plugins/ai-aliyun-content-moderation.lua
@@ -530,7 +530,12 @@ function _M.lua_body_filter(conf, ctx, headers, body)
             end
             table.insert(raw_events, sse.encode(event))
         end
-        if not contains_done_event and proto and ctx.var.llm_request_done then
+        -- llm_request_done only means "no more content is coming", which is also
+        -- set when a stream is cut short (upstream read error, stream limit).
+        -- ctx.ai_stream_aborted marks those cases: synthesizing a terminator there
+        -- would tell the client a truncated response completed successfully.
+        if not contains_done_event and proto and ctx.var.llm_request_done
+           and not ctx.ai_stream_aborted then
             table.insert(raw_events, proto.build_done_event())
         end
         return nil, table.concat(raw_events, "\n")
--- a/apisix/plugins/ai-aws-content-moderation.lua
+++ b/apisix/plugins/ai-aws-content-moderation.lua
@@ -446,7 +446,12 @@ local function annotate_stream(ctx, body)
         table.insert(raw_events, sse.encode(event))
     end
 
-    if not contains_done_event and proto.build_done_event and ctx.var.llm_request_done then
+    -- llm_request_done only means "no more content is coming", which is also
+    -- set when a stream is cut short (upstream read error, stream limit).
+    -- ctx.ai_stream_aborted marks those cases: synthesizing a terminator there
+    -- would tell the client a truncated response completed successfully.
+    if not contains_done_event and proto.build_done_event
+       and ctx.var.llm_request_done and not ctx.ai_stream_aborted then
         table.insert(raw_events, proto.build_done_event())
     end
     return table.concat(raw_events)
--- a/apisix/plugins/ai-providers/base.lua
+++ b/apisix/plugins/ai-providers/base.lua
@@ -555,7 +555,13 @@ function _M.parse_streaming_response(self, ctx, res, target_proto, converter, co
 
         local chunk, err = body_reader()
         if err then
-            ctx.ai_stream_aborted = "read_error"
+            -- A read error that arrives after the protocol's completion event
+            -- means the stream itself finished and only the transport died late,
+            -- so the response is complete rather than aborted.
+            local completed = ctx.var.llm_request_done
+            if not completed then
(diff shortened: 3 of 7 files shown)


AlinsRan (dismissed):

membphis (commented):
**[P2] A retry attempt inherits `llm_request_done` from the previous attempt**

`ai-proxy-multi` performs fallback retries within the same request context, but a new upstream attempt only clears `ai_stream_aborted`; it does not clear `ctx.var.llm_request_done`.

This is not a retry-after-partial-output issue: once `output_sent` is true, the current code does not trigger a fallback.

The issue occurs when the first attempt produces no downstream output and is therefore legitimately retryable, but has already set `llm_request_done=true`. For example, during protocol conversion, an upstream stream containing only `[DONE]` sets `llm_request_done`, while the converter emits no downstream event. EOF then returns `502`, allowing an `http_5xx` fallback to start another attempt.

The second attempt therefore inherits stale completion state. If it later encounters a read error after emitting partial output, the new `completed = ctx.var.llm_request_done` check treats the truncated stream as if the current attempt had completed normally, skipping both the `ai_stream_aborted` marker and abort finalization.

Please reset attempt-local completion and abort state at the start of every upstream attempt. Preferably, use a local `protocol_completed` flag that is set only when the current attempt parses a `done` or `usage_and_done` event. Please also add a regression where the first attempt emits only `[DONE]`, produces no downstream output, and triggers a fallback, while the second attempt emits partial output and then encounters a read error. The test should verify that the second attempt does not inherit completion state and correctly follows the aborted path.

AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

nic-6443:
Fair point — the AWS branch was changed without a test of its own, and the existing AWS suite only covers synthesis, never with `ctx.ai_stream_aborted` set.

This merged before I got the case written, so it is in apache/apisix#13880: the AWS counterpart of the aliyun case, asserting the content is annotated and released while no `[DONE]` is appended to the truncated stream. Removing the gate makes it fail.

PR #13879, opened 2026-08-26
docs(openid-connect): recommend setting redirect_uri to an absolute URL

shreemaan-abhishek:
### Description

Documents a recommendation for the `openid-connect` plugin: set `redirect_uri` to an absolute URL (with scheme and host).

When `redirect_uri` is unset or a relative path, the plugin assembles the host from request headers (`Host` / `Forwarded`). Setting an absolute `redirect_uri` makes the value fixed and bypasses header-based host assembly entirely, so a forged header cannot influence the redirect URI sent to the identity provider.

Updated in two places in `docs/en/latest/plugins/openid-connect.md`:
- the `redirect_uri` attribute description
- the "Misconfigured Redirection URI" troubleshooting section

Fixes #

### Checklist


kayx23 (approved):
Verified against APISIX and `lua-resty-openidc`: a fully qualified `redirect_uri` bypasses request-derived scheme and host assembly, so the recommendation is sound.

Non-blocking wording nuance: header-based assembly applies to an unset value or a root-relative value beginning with `/`; current APISIX also honors forwarded origin headers only from peers configured in `apisix.trusted_addresses`.

nic-6443 (approved):

membphis (approved):
LGTM

PR #13880, opened 2026-08-26
test(ai-proxy): cover the AWS moderation path in the truncated-stream suite

nic-6443:
### Description

Follow-up to the truncated-stream fix: it gated the terminator synthesis in both `ai-aws-content-moderation` and `ai-aliyun-content-moderation` on `ctx.ai_stream_aborted`, but the regression suite that came with it only configured the aliyun plugin. The AWS branch was changed without a test that fails before the gate and passes after — the existing AWS suite exercises synthesis, never with `ctx.ai_stream_aborted` set.

This adds the AWS counterpart of the aliyun case to `t/plugin/ai-proxy-stream-truncated.t`: an upstream that delivers a content event and a usage event and then loses the transport, with `ai-aws-content-moderation` in `final_packet` mode. It asserts the content is annotated and released to the client and that no `[DONE]` is appended to the truncated stream. Removing the `not ctx.ai_stream_aborted` gate makes it fail.

The Comprehend stand-in answers one clean verdict per submitted segment, which is all the terminator branch needs; the verdict itself is not what is under test, so it does not reuse the fixture-driven mock from the AWS suite.

Test-only change; no plugin code is touched.

### Checklist

### DIFF
--- a/t/lib/truncated_sse.lua
+++ b/t/lib/truncated_sse.lua
@@ -111,4 +111,32 @@ function _M.serve_abort_once()
 end
 
 
+-- Minimal AWS Comprehend detectToxicContent stand-in: answers one clean result
+-- per submitted segment. The truncation tests only need the moderation plugin to
+-- reach its terminator-synthesis branch, not a particular verdict.
+function _M.serve_clean_comprehend()
+    local json = require("cjson.safe")
+    ngx.req.read_body()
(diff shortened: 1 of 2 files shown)


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13836, opened 2026-08-17
fix(basic-auth): split credentials on the first colon only

Arjen10:
### Description

RFC 7617 treats everything after the first ':' as the password. ngx.re.split on all colons truncated passwords that contained ':'.

Align with the implementation of ldap-auth-advanced

#### Which issue(s) this PR fixes:

Fixes #13835

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [x] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #13883, opened 2026-08-26
fix(basic-auth): reject empty consumer password and fail closed

### LINKED ISSUE #13881: bug: basic-auth accepts an empty consumer password, which allows passwordless access
### Current Behavior

The `basic-auth` consumer schema declares `password = { type = "string" }` with no `minLength`, so the Admin API accepts a consumer whose password is `""`. Such a consumer then authenticates, so it looks protected while it effectively has no secret:

- on `master`, `Authorization: Basic base64("alice: ")` returns 200: the whitespace is stripped by the `\s+` gsub and the plugin compares `"" == ""`.
- `base64("alice:")` currently returns 401 only by accident (`ngx.re.split` drops the trailing empty field, so the credential is treated as malformed). Once #13836 lands (split on the first colon, which is the correct RFC 7617 parsing), `alice:` returns 200 as well.

RFC 7617 lets `user:` appear on the wire, but the OpaqueString password profile it points to (RFC 8265, section 4.1) says a password MUST NOT be zero-length.

### Expected Behavior

- The Admin API rejects `password: ""` on consumer / credential create and update (`minLength = 1` on `consumer_schema.password`).
- The plugin fails closed with the usual generic 401 when the resolved password is empty, which also covers existing data and `$secret://` / `$ENV://` references that resolve to `""`.
- Regression tests: Admin API rejection of `password: ""`, `user:` returns 401, an empty secret reference returns 401, and a password containing `:` keeps authenticating (from #13836).

### Error Logs

None, the request succeeds.

### Steps to Reproduce

```shell
# 1. consumer with an empty password: accepted
curl -s http://127.0.0.1:9180/apisix/admin/consumers -H "X-API-KEY: $KEY" -X PUT \
  -d '{"username":"alice","plugins":{"basic-auth":{"username":"alice","password":""}}}'

# 2. route protected by basic-auth
curl -s http://127.0.0.1:9180/apisix/admin/routes/1 -H "X-API-KEY: $KEY" -X PUT \
  -d '{"uri":"/hello","plugins":{"basic-auth":{}},"upstream":{"type":"roundrobin","nodes":{"127.0.0.1:1980":1}}}'

# 3. whitespace-only password: 200 on master, expected 401
curl -si http://127.0.0.1:9080/hello -H "Authorization: Basic $(printf 'alice: ' | base64)"

# 4. empty password: 401 on master by accident, 200 with #13836 applied, expected 401
curl -si http://127.0.0.1:9080/hello -H "Authorization: Basic $(printf 'alice:' | base64)"
```

### Environment

- APISIX version (run `apisix version`): master (3.18.0), step 4 also verified on the #13836 branch
- Operating system (run `uname -a`): Linux 7.0.14 aarch64
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 3.5.11
- APISIX Dashboard version, if relevant: n/a
- Plugin runner version, for issues related to plugin runners: n/a
- LuaRocks version, for installation issues (run `luarocks --version`): n/a

waterWang:
### Summary

Fixes #13881 — `basic-auth` currently accepts a consumer with an empty password, allowing passwordless access.

RFC 8265 section 4.1 (the OpaqueString password profile referenced by RFC 7617) requires a password to be non-zero-length. Today the Admin API accepts `password: ""` because `consumer_schema.password` is declared as plain `{ type = "string" }` with no `minLength`. Such a consumer authenticates, so it looks protected while it effectively has no secret. Once #13836 lands (split on the first colon per RFC 7617), `user:` would also return 200 for such a consumer.

### Changes

- **`apisix/plugins/basic-auth.lua`**
  - Add `minLength = 1` to `consumer_schema.password` so the Admin API rejects empty passwords on consumer create/update.
  - Fail closed in `find_consumer` when either the presented or the resolved password is empty. This also covers existing empty-password consumers and `$secret://` / `$ENV://` references that resolve to `""`.

- **`t/plugin/basic-auth.t`** — regression tests:
  - consumer schema rejects an empty password
  - Admin API rejects a consumer with an empty password (400)
  - `foo:` (empty password) returns 401
  - whitespace-only password (`foo: `) returns 401

### Compatibility with #13836

This change is orthogonal to #13836 (which fixes parsing to split on the first colon). With or without #13836, an empty/whitespace-only password now fails closed with a 401.

### DIFF
--- a/apisix/plugins/basic-auth.lua
+++ b/apisix/plugins/basic-auth.lua
@@ -43,7 +43,10 @@ local consumer_schema = {
     title = "work with consumer object",
     properties = {
         username = { type = "string" },
-        password = { type = "string" },
+        password = {
+            type = "string",
+            minLength = 1,
+        },
     },
     encrypt_fields = {"password"},
     required = {"username", "password"},
@@ -145,6 +148,13 @@ local function find_consumer(ctx)
             return nil, nil, err
         end
         core.log.warn(err)
+        return nil, nil, "Invalid user authorization"
+    end
+
+    -- RFC 8265 section 4.1: a password MUST NOT be zero-length. Fail closed
+    -- when the resolved password is empty, which also covers existing data
+    -- and $secret:// / $ENV:// references that resolve to "".
+    if password == "" or cur_consumer.auth_conf.password == "" then
         return nil, nil, "Invalid user authorization"
     end
 
--- a/t/plugin/basic-auth.t
+++ b/t/plugin/basic-auth.t
@@ -708,3 +708,81 @@ Authorization: bASiC Zm9vOmJhcg==
 hello world
 --- error_log
 find consumer foo
+
+
+=== TEST 40: consumer schema rejects empty password
+--- config
+    location /t {
+        content_by_lua_block {
+            local core = require("apisix.core")
+            local plugin = require("apisix.plugins.basic-auth")
+            local ok, err = plugin.check_schema({username = 'foo', password = ''}, core.schema.TYPE_CONSUMER)
+            if ok then
+                ngx.say("unexpected: schema accepted empty password")
+                return
+            end
+            ngx.say(err or "rejected")
+        }
+    }
+--- request
+GET /t
+--- response_body_like
+property "password" validation failed
+--- no_error_log
+unexpected
+
+
+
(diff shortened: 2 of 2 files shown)


juzhiyuan:
Thanks but this issue has been fixed by another PR :)

PR #13884, opened 2026-08-26
fix(basic-auth): reject an empty consumer password

### LINKED ISSUE #13881: bug: basic-auth accepts an empty consumer password, which allows passwordless access

janiussyafiq:
### Description

The `basic-auth` consumer schema accepted `password: ""`, and such a consumer could authenticate with `user: ` on the wire (and with `user:` once #13836 splits credentials on the first colon).

- `consumer_schema.password` now requires `minLength = 1`, so the Admin API rejects an empty password on consumers and credentials.
- The plugin fails closed with the usual generic 401 when the resolved password is empty, which also covers data written before this change and `$secret://` / `$env://` references that resolve to `""`.
- Docs: the `password` attribute is documented as a non-empty string.

Fixes #13881

### Checklist


apisix/plugins/basic-auth.lua
```diff
@@ -148,7 +148,19 @@ local function find_consumer(ctx)
...
+    -- password, or a secret reference that resolves to "", must never authenticate
+    local expected = cur_consumer.auth_conf.password
+    if expected == "" then
+        err = "empty password configured for consumer: " .. cur_consumer.consumer_name
+        if auth_utils.is_running_under_multi_auth(ctx) then
+            return nil, nil, err
+        end
+        core.log.warn(err)
+        return nil, nil, "Invalid user authorization"
+    end
```
nic-6443: I believe setting constraints in the JSON Schema is sufficient, and no additional coding is required.

janiussyafiq: I believe this is needed for the `$secret://` or `$env://` reference, or else TEST 35 & 36 will likely fail. After dropping this codeblock, apparently TEST 35 still passed due to #13836 not yet getting merged.

also investigated root cause of flaky CI, would be fix in the latest version of apisix-build-tools with https://github.com/api7/apisix-build-tools/pull/490 included

nic-6443: I think this is excessive defensive programming.

nic-6443 (dismissed):

shreemaan-abhishek (dismissed):

membphis:
[P1] Reject passwords that resolve to an empty secret or environment value

minLength = 1 only validates the non-empty $secret://... or $env://... reference string. consumer.lua resolves that reference before building the consumer cache, and a resolved empty string is not rejected. basic-auth.lua then compares it directly with the request password, so a Basic header containing the matching username and an empty password can still authenticate when the reference resolves to "".

Please reject an empty password after secret/environment resolution, for example in the basic-auth runtime check or by revalidating the resolved consumer config, and add regressions showing $env:// and $secret:// values resolving to "" return the generic 401 response.

membphis (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13887, opened 2026-08-26
fix(ai-cache): key the passthrough protocol on the client method, path and query

shreemaan-abhishek:
### Description

For the `passthrough` protocol, `ai-proxy` forwards the client's request method, path and query string verbatim (`ai-proxy/base.lua`), so on a wildcard route they select which upstream endpoint answers. The `ai-cache` fingerprint covered only the client body and the instance config, so the same body sent to `/v1/images/generations` and `/v1/chat/completions` (or to one path with different query parameters) collided on one cache entry and the second request was served the first one's response.

This folds the client method, `uri` and `args` into `build_repr` under a passthrough gate. Other protocols build a fixed upstream request from the body alone, so their fingerprints are unchanged; an existing entry keeps its key.

Tests: a `key.lua` unit block (path, query and method each flip the passthrough fingerprint; the `openai-chat` fingerprint still ignores the client path) plus an end-to-end sequence on a `/v1/*` route: same body to two upstream paths is MISS/MISS, a query-string variant is a MISS, and an exact repeat is a HIT. Without the fix TESTs 58, 61 and 62 fail.

Docs: the `cache_key` note in `ai-cache.md` (en/zh) now states that passthrough keys include the method, path and query.

Fixes #

### Checklist

### DIFF
--- a/apisix/plugins/ai-cache/key.lua
+++ b/apisix/plugins/ai-cache/key.lua
@@ -69,12 +69,22 @@ local function build_repr(ctx, body, messages)
     local proto = ctx.ai_client_protocol and protocols.get(ctx.ai_client_protocol)
     params.stream = (proto and proto.is_streaming(body)) == true
 
+    local client = {
+        protocol = ctx.ai_client_protocol or "",
+        messages = messages,
+        params   = params,
+    }
+    -- passthrough proxies the client's method, path and query string verbatim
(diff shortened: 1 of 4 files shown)


nic-6443 (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13891, opened 2026-08-27
fix(control): always report healthcheck nodes as a JSON array

AlinsRan:
### Description

`GET /v1/healthcheck` returns `nodes` as `[...]` once the health checker has
registered its targets, but as `{}` before that: the field comes straight from
`healthcheck.get_target_list()`, which is a plain Lua table, and cjson encodes an
empty table as an object.

The window is reachable in normal operation. An upstream is listed as soon as it is
configured with `checks`, but the health checker itself is only created the first
time that upstream serves a request, so every configured-but-not-yet-used upstream
reports `"nodes": {}`:

```json
[{"name":"/routes/1","nodes":{},"type":"http"}]
```

The top-level result has the same problem — `GET /v1/healthcheck` returns `{}`
instead of `[]` when nothing is being health checked (this was pinned by TEST 4 in
`t/control/healthcheck.t`).

Both tables are now tagged with `core.json.array_mt`, so they keep their array type
regardless of content. `/v1/healthcheck/{src_type}/{src_id}` goes through the same
`extra_checker_info()` and is fixed with it.

**Behavior change**: `nodes` and the top-level list are `[]` instead of `{}` when
empty. Any client that already decoded the response into a list keeps working; a
client that special-cased the object form does not.

While here, the docs claimed an upstream shows up in the result only after it has
served a request in some worker process. That is not what the code does — it is
listed as soon as a health checker is configured, and it is `nodes` that stays empty
until then. Corrected in both `docs/en` and `docs/zh`.

Fixes #

### Checklist

### DIFF
--- a/apisix/control/v1.lua
+++ b/apisix/control/v1.lua
@@ -27,6 +27,7 @@ local get_upstreams = upstream_mod.upstreams
 local collectgarbage = collectgarbage
 local ipairs = ipairs
 local pcall = pcall
+local setmetatable = setmetatable
 local str_format = string.format
 local ngx = ngx
 local ngx_var = ngx.var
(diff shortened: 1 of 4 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

Issue #13775, opened 2026-08-04
bug: APISIX 3.17.0 is unable to start containers built from the base image


wangchao732:
### Current Behavior

baseimgae: apache/apisix:3.17.0-ubuntu

Since the Apisix user lacks apt permissions, OpenJDK 21 is installed using docker --privileged --user root.

```dockerfile
FROM xxxapisix:3.17.0-ubuntu-2026080415

RUN apt-get update && apt-get install -y openjdk-21-jdk-headless

COPY target/*.jar  /usr/local/apisix/

RUN chown -R apisix:root  /usr/local/apisix/

USER apisix
```

ext-plugin It fails to start regardless of whether it's configured or not.

I tried using sleep 3600 to keep the container running. When I manually executed /usr/local/openresty/bin/openresty -p /usr/local/apisix -g 'daemon off;' inside the container, it worked fine. However, running bash /docker-entrypoint.sh produces no output, and ps -aux shows no openresty process.

### Error Logs

null

### Steps to Reproduce

1、pull image apache/apisix:3.17.0-ubuntu;
2、docker run --user root apache/apisix:3.17.0-ubuntu && docker commit xxx   xxx/apisix:3.17.0-ubuntu-2026080415;
3、build images above Dockefile;
4、deployment;

### Environment

- APISIX version: 3.17.0
- Operating system (run `uname -a`):  Ubuntu 24.04.4 LTS
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`): 1.29.2.4

janiussyafiq:
Hi @wangchao732, I reproduced your exact chain (run `--user root`, `docker commit`, your Dockerfile with openjdk-21-jdk-headless, `chown -R apisix:root`, `USER apisix`) and the resulting image starts APISIX normally, so the commit-based image itself is fine.

What you observed is expected behavior of the entrypoint. `/docker-entrypoint.sh` only starts APISIX when its first argument is `docker-start`:

```bash
if [[ "$1" == "docker-start" ]]; then
    ...
    exec /usr/local/openresty/bin/openresty -p /usr/local/apisix -g 'daemon off;'
fi

exec "$@"
```

Running `bash /docker-entrypoint.sh` with no argument falls through to `exec "$@"`, which does nothing: it exits 0 with no output and no openresty process, exactly what you saw. The same command with the argument works: `bash /docker-entrypoint.sh docker-start`.

[...]

PR #13898, opened 2026-08-30
docs: warn on missing docker-start and document custom-image CMD pitfall

mehrdadbn9:
### Description

A custom image built from `docker commit` of a shell session bakes the shell into `CMD`, replacing the official `["docker-start"]`. The entrypoint then falls through to `exec "$@"` and the container exits 0 with no logs and no APISIX process (apache/apisix#13775).

This PR:
- emits a stderr warning in `/docker-entrypoint.sh` when the first argument is not `docker-start`, so the silent failure is obvious instead of a mysterious exit 0.
- adds an FAQ entry with the correct Dockerfile pattern (switch `USER` inside the Dockerfile instead of committing a shell session), plus the Kubernetes `command`/`args` override.

### Checklist

- [x] I have explained my changes
- [x] Doc/entrypoint change, no behavior change for the normal `docker-start` path

Refs apache/apisix#13775

### DIFF
--- a/docker/debian-dev/docker-entrypoint.sh
+++ b/docker/debian-dev/docker-entrypoint.sh
@@ -61,4 +61,17 @@ _EOC_
     exec /usr/local/openresty/bin/openresty -p /usr/local/apisix -g 'daemon off;'
 fi
 
+# If we get here, the first argument was not "docker-start". The container will
+# run "$@" and exit. That is expected when the user overrides the command on
+# purpose (e.g. to run a shell), but it is the classic symptom of a broken image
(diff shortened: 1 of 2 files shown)


kayx23 (commented):
Thanks for investigating and documenting the root cause. I suggest closing this PR because APISIX and the official image behave as designed; the failure comes from `docker commit` preserving an overridden `CMD`. The entrypoint warning would also affect legitimate command overrides and could be misleading. The resolution in #13775 should be sufficient unless this problem becomes common enough to justify a short troubleshooting note in the Docker documentation.

PR #13901, opened 2026-08-31
fix(traffic-label): cache the compiled match expressions outside the plugin config

nic-6443:
### Description

`traffic-label`'s `access()` compiles each rule's `match` expression on the first request and stores the result in `conf.rules_arr` — inside the plugin configuration itself.

For the `ipmatch` operator, `resty.expr` compiles the right-hand side into an `ipmatcher` (`resty/expr/v1.lua`), and an ipmatcher's `ipv4` / `ipv4_mask` lookup tables are keyed by integers (the 32-bit address and the prefix length). cjson refuses to serialise such a table:

```
failed to encode: Cannot serialise table: excessively sparse array force: true
```

So from the first request that reaches the plugin onwards, the route configuration can no longer be JSON encoded. With the error log at `info` level, every line that dumps the route comes out empty and each one emits two warnings (`log_wrapper` evaluates a `__tostring` argument twice):

```
[warn]  json.lua:179: failed to encode: Cannot serialise table: excessively sparse array force: true
[warn]  json.lua:179: failed to encode: Cannot serialise table: excessively sparse array force: true
[info]  init.lua:862: http_access_phase(): matched route:
```

`matched route`, `insert uri route` and `route conf` are all affected, i.e. the debugging output you turned `info` on for is exactly what is lost.

Keep the compiled expression on the rule itself but behind a metatable, so it stays invisible to serialisation while its lifetime still matches the configuration exactly — compiled once, no TTL, no eviction cap, no recompilation. This is the same technique #11770 introduced for `body-transformer` and that `set_plugins_meta_parent` uses for the plugin `_meta.parent`.

N/A

### Checklist

- [x] I have updated the documentation to reflect this change (none needed: no user-facing behaviour change)


shreemaan-abhishek (approved):

AlinsRan (approved):

membphis (approved):
LGTM

PR #13904, opened 2026-09-01
feat: improve API-driven standalone update reliability

bzp2010:
### Description

This PR aims to improve the reliability of configuration updates in API-driven standalone mode by implementing the following measures:

- A new "standalone-status" reporting mechanism has been added. Now, each subsystem on every worker will report the digest of its loaded configuration, down to the level of each entity type (`worker:<id>:<subsystem>:<entity_type> = <digest>`). This information will indicate whether the new configuration has taken effect on each worker (e.g., the router tree will be reset; the consumer lrucache will be flushed), and so on.
  - Technically, a worker reporting a digest simply indicates that the configuration has been loaded but does not necessarily reflect that the router has actually been rebuilt. The router is passively rebuilt based on differences in the `conf_version` value for each request; therefore, as long as `conf_version` has been updated, we can assume that the router has been rebuilt.

- The `PUT /apisix/admin/configs` request now supports a `wait` parameter. It allows you to specify a value in milliseconds; APISIX will collect if configuration applied status during this wait period and report it to the client. It is based on the report on "standalone-status" mentioned above.
  - This API now returns a `200` or `202`. A `200` status code indicates that the configuration has been accepted and loaded on each worker, while a `202` status code indicates that the configuration has been accepted but its loading status is not guaranteed.
  - Configuration loading is expected to complete within 1 second, as each worker has a timer-driven loading mechanism. However, this may take longer under certain special circumstances (such as extreme CPU load).

- Refactor the configuration loading from `shdict` to improve the configuration loading latency window for new workers (which may be restarted via `reload`). Configuration is now always loaded synchronously during `init_worker` and consumed immediately during `core.config.new`.

- Refactor the storage format of configurations in shdict, they are now always encoded as `<digest_len>\n<digest><raw_json>`, combining the digest and JSON using a fixed-length prefix. This ensures concurrency safety while allowing timer-based `poll` operations to avoid having to parse large JSON data every time.

  - Previously, the digest was stored within the JSON, requiring the JSON to be fully parsed each time—which could be as large as 10 MB—representing a significant overhead. Now that the digest is encoded outside the JSON, there is no longer a need to parse the JSON in advance.

All of the above changes are backward-compatible, and existing clients can work with them. Older clients do not send the `wait` parameter, so they will not trigger the new waiting behavior. The other changes are internal refactoring improvements that do not break compatibility.

### Checklist


apisix/admin/standalone.lua
```diff
@@ -64,19 +84,82 @@ local function update_and_broadcast_config(apisix_yaml)
...
+local is_stream_enabled
+local function stream_enabled()
+    if is_stream_enabled ~= nil then
+        return is_stream_enabled
+    end
+    local local_conf = core.config.local_conf()
+    local proxy_mode = local_conf and local_conf.apisix and local_conf.apisix.proxy_mode
+    is_stream_enabled = proxy_mode ~= nil and str_find(proxy_mode, "stream", 1, true) ~= nil
+    return is_stream_enabled
+end
```
nic-6443: How about putting this piece of code into the `core.config.local_conf` module? I've noticed this code is duplicated in several places.

bzp2010: Agree, I will do this.

apisix/core/config_yaml.lua
```diff
@@ -62,6 +69,37 @@ local _M = {
...
+
+    local digest_len = tonumber(sub_str(stored, 1, idx - 1))
+    if not digest_len then
+        return nil, nil, "invalid digest length prefix"
+    end
+
+    local digest_start = idx + 1
+    local digest_end = digest_start + digest_len - 1
+    return sub_str(stored, digest_start, digest_end), sub_str(stored, digest_end + 1)
+end
```
nic-6443: I feel it's somewhat inappropriate to put this piece of code in `config_yaml.lua`. When I saw `config_yaml.decode_config` and `config_yaml.encode_config` at the call site, I thought they were performing YAML decode and encode operations, but they actually handle a custom data structure.

nic-6443: I see. It's not that the code is placed incorrectly, but that the file name is too restrictive. It now has capabilities far beyond YAML. It is also a fundamental module for the standalone mode, so it can be renamed to `config_standalone.lua`.

nic-6443 (dismissed):
Minor suggestions have been submitted, and most of the code LGTM

apisix/core/config_yaml.lua
bzp2010: This was intentional. I can't have a `core` module have a dependency on certain code under `admin/`. A functional module should depend on the core's std lib, not the other way around.

Maybe it really isn't the right place for it; I'll think about it some more.

bzp2010: I moved them to `admin/standalone.lua`; now it's cohesive and doesn't export anything.

At the same time, the API-driven standalone-specific capability of restoring configuration from shdict at startup has also been moved to`admin/standalone.lua`, which seems reasonable.

Perhaps in a future refactoring we can create a `config_standalone.lua` or something similar as a new configuration source, but I won’t do that now.

apisix/admin/standalone.lua
```diff
@@ -40,22 +58,60 @@ local METADATA_DIGEST = "X-Digest"
...
+local function encode_config(digest, raw)
+    digest = digest or ""
+    return #digest .. CONFIG_DIGEST_LENGTH_SEPARATOR .. digest .. raw
+end
+
+
+local function decode_config(stored)
+    local idx = str_find(stored, CONFIG_DIGEST_LENGTH_SEPARATOR, 1, true)
+    if not idx then
+        return nil, nil, "missing digest length prefix"
```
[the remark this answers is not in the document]
bzp2010: That's complete bullshit.

There is no proper workflow that allows you to perform a hot update for Lua. This update will only be available in the next minor version; if you want to upgrade, you must perform a full restart, because there is no viable incremental path.

Therefore, there is absolutely no possibility of retaining shdict. Nor is there any possibility of running both old and new versions of Lua scripts simultaneously.

What you're assuming will never happen; you have absolutely no understanding of how this project works or its programming model. Stop that nonsense.

nic-6443 (dismissed):

apisix/admin/standalone.lua
```diff
@@ -64,19 +120,105 @@ local function update_and_broadcast_config(apisix_yaml)
...
 local validate_configuration = config_validate.validate_configuration
 
+
+local MAX_WAIT_MS = 60000
+local POLL_INTERVAL = 0.05
+
+
+local function parse_wait_ms(ctx)
+    local args = core.request.get_uri_args(ctx)
+    local wait = args and tonumber(args.wait)
```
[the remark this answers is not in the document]
bzp2010: ...Do you really understand Lua, or are you just speaking based on that ridiculous LLM hallucinations?

```lua
print(tonumber(nil)) -- nil
print(tonumber(true)) -- nil
print(tonumber({a=1,b=2}))  -- nil
print(tonumber("1"))  -- 1
```

AlinsRan (dismissed):

apisix/admin/standalone.lua
```diff
@@ -64,19 +120,105 @@ local function update_and_broadcast_config(apisix_yaml)
...
+local function all_workers_applied(target_digest)
+    if not status_shared_dict then
+        return false
+    end
+
+    local n = worker_count()
+    local check_stream = config_local.is_stream_enabled()
+    for key in pairs(APPLIED_CHECK_KEYS) do
+        for id = 0, n - 1 do
+            if HTTP_RESOURCE_KEYS[key] then
```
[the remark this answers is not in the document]
bzp2010: A fundamental refactoring has been implemented.

apisix/admin/standalone.lua
```diff
@@ -313,6 +476,14 @@ end
 
 
 function _M.init_worker()
+    local ok, err = try_restore_from_shared_dict()
```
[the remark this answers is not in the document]
bzp2010: Fixed

AlinsRan (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #13908, opened 2026-09-02
fix(aws-lambda): request "/" for a path-less function_uri and log function error responses

### LINKED ISSUE #11176: `aws-lambda` plugin returns `400 bad request` for wrong IAM credentials or wrong URL
### Current State

The `aws-lambda` plugin returns `400 Bad Request` for wrong IAM credentials (wrong access key or secret key), or wrong URL.

For example, **_if the URL is missing the ending slash_**, APISIX will return `400 Bad Request` without further information.

The error log is the following and it is not very informative:

```text
2024/04/22 02:10:01 [warn] 277#277: *11342 [lua] plugin.lua:1160: run_plugin(): aws-lambda exits with http status code 400, client: 172.25.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:9080"
```

### Desired State

As an enhancement, APISIX could return more telling error message to help users troubleshooting.

For example, with AWS CLI `aws sts get-caller-identity`, wrong secret key would lead to the following error msg:

```
The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
```

Wrong access key would lead to the following error msg:

```
The security token included in the request is invalid.
```

### Environment

APISIX version (run `apisix version`): 3.9.0

janiussyafiq:
### Description

`aws-lambda` (and the other plugins built on `serverless/generic-upstream.lua`: `azure-functions`, `openfunction`) answered `400 Bad Request` with no useful detail when `function_uri` had no path, e.g. `https://xxx.lambda-url.us-east-1.on.aws` without the trailing slash. `net.url` parses that into an empty path string, and `url_decoded.path or "/"` never kicked in because `""` is truthy in Lua. The plugin then sent `GET  HTTP/1.1` with an empty request-target, which the function host rejects before it reaches the function. This is the "wrong URL" case from the issue.

The issue also asks for better troubleshooting output on wrong IAM credentials. The function's error body (for example AWS's `The security token included in the request is invalid.`) was already forwarded to the client, but `error.log` only carried `aws-lambda exits with http status code 403`, so an operator looking at the gateway alone could not tell why.

Changes:

- Treat an empty parsed path as `/`, so the request-target matches the SigV4 canonical URI and is accepted by the function host.
- When the function answers with a 4xx/5xx, log the status and body at `warn` level (same shape as `authz-keycloak`), in addition to forwarding them to the client unchanged.

Tests: `t/plugin/aws-lambda.t` TEST 10 (a path-less `function_uri` reaches the mock as `/`) and TEST 11 (a 403 JSON body is forwarded to the client and present in `error.log`). Both fail without the fix.

Fixes #11176

### Checklist


nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13909, opened 2026-09-02
feat: label WebSocket sessions with request_type=websocket

### LINKED ISSUE #12661: feat: APISIX Prometheus plugins latency metrics should exclude websocket requests
### Description

When a route with WebSocket enabled receives a WebSocket request, the upstream_latency metric from the Prometheus plugin reports artificially high values. 

Due to their long-lived connections, WebSocket requests inflate the upstream_latency metric, making it not representative of true latency. We should exclude them from this measurement

janiussyafiq:
#### Motivation

On a route that serves WebSocket traffic, the `request` and `upstream` types of `apisix_http_latency` are observed in the log phase, which for an upgraded connection only runs when the session closes. Both values therefore equal the lifetime of the session (`$upstream_response_time` also stops only when the upstream connection closes), not a request/response round trip. A handful of long-lived sessions is enough to push the histogram's buckets and quantiles far beyond any real latency, and nothing in the metrics tells a session apart from an ordinary request.

#### What this PR does

- Sets `$request_type` to `websocket` in the header filter phase when the response is `101 Switching Protocols`. `request_type` already labels `apisix_http_status`, `apisix_http_latency` and `apisix_bandwidth`, so a session lands in its own series and can be left out of latency queries with `request_type!="websocket"`. No data is dropped at the source.
- Ordinary requests keep their existing series and label values. Only requests answered with 101 move from `traditional_http` to `websocket`.
- The classification lives in core rather than the `prometheus` plugin, so any logger that emits `$request_type` sees the same value.
- The trigger is the 101 response, not the route's `enable_websocket` flag or the client's `Upgrade` header, so a refused handshake on the same route stays `traditional_http`.

This replaces the earlier opt-in `exclude_websocket_latency` metadata from the first revision of this PR, following the review suggestion to distinguish WebSocket requests with a label instead.

Tests cover the `$request_type` value seen in the log phase, the label on `http_status`, `http_latency` and `bandwidth`, and a refused handshake on the same route staying in its own `traditional_http` series. English and Chinese docs list the new value.

Fixes #12661

### Checklist


apisix/plugins/prometheus/exporter.lua
```diff
@@ -771,18 +778,22 @@ function _M.http_log(conf, ctx)
 
     local latency_extra_label_values = extra_labels("http_latency", ctx)
 
-    metrics.latency:observe(latency,
-        get_enabled_label_values_for_metric("http_latency", disabled_label_metric_map,
-            "request", route_id, service_id, consumer_name, balancer_ip,
-            vars.request_type, request_llm_model_label, llm_model_label,
-            unpack(latency_extra_label_values)))
+    local skip_session_latency = tonumber(vars.status) == 101 and exclude_websocket_latency()
```
nic-6443: How about adding a protocol label in metrics to distinguish websocket requests?

janiussyafiq: I think it's a better and cleaner approach.
Similar to your suggestion but instead of adding a new protocol label, could we simply add websocket to the existing `request_type` which currently holds `traditional_http, ai_chat, ai_stream`?

nic-6443: LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis:
### [P2] Keep the request-variable cache in sync when classifying WebSocket sessions

The new header-filter assignment updates `ngx.var.request_type` directly, but `ctx.var` caches values on first read. If an earlier plugin has read `ctx.var.request_type`, the Prometheus exporter still receives `traditional_http` after a successful `101` upgrade.

A supported trigger is a WebSocket route whose `proxy-rewrite.headers.set` contains `"X-Request-Type": "$request_type"`. The rewrite phase caches the old value, and the header filter then updates only the NGINX variable. As a result, `apisix_http_status`, `apisix_http_latency`, and `apisix_bandwidth` can still classify the session as ordinary HTTP. Filtering with `request_type!="websocket"` therefore leaves the long session in the latency statistics. This should be fixed before merge.

This is deterministic when the earlier read occurs. WebSocket requests without that earlier read are unaffected, and forwarding itself is unchanged.

Please write through `api_ctx.var.request_type = "websocket"` when the request context is available, preserving the no-context fallback. Add a regression that reads the variable in `rewrite` before a successful upgrade, then checks that both variable views and all three metric families use `websocket`.

Validation: a focused LuaJIT reproduction using the repository's actual variable metatable and this PR's added statements, with a simulated NGINX variable interface, produced `ngx.var.request_type=websocket` and `ctx.var.request_type=traditional_http`. Writing through the existing context setter made both views agree.

Evidence at `e7a399130955948d57ea3cb48a964a27372e695a`:

- [Classification assignment](https://github.com/apache/apisix/blob/e7a399130955948d57ea3cb48a964a27372e695a/apisix/init.lua#L1027)
- [Variable cache and setter](https://github.com/apache/apisix/blob/e7a399130955948d57ea3cb48a964a27372e695a/apisix/core/ctx.lua#L392)
- [Earlier read in proxy-rewrite](https://github.com/apache/apisix/blob/e7a399130955948d57ea3cb48a964a27372e695a/apisix/plugins/proxy-rewrite.lua#L302)
- [Prometheus label collection](https://github.com/apache/apisix/blob/e7a399130955948d57ea3cb48a964a27372e695a/apisix/plugins/prometheus/exporter.lua#L762)

membphis (approved):
LGTM

PR #13911, opened 2026-09-02
feat(stream): match a stream route by several SNIs

AlinsRan:
### Description

A stream route can only carry one `sni`, so serving several names from one backend means duplicating the route once per name — and with it the plugins and any traffic-split state attached to it.

This adds `snis`, the plural form of the same field. A route matches when the SNI in the ClientHello equals any entry:

```shell
curl http://127.0.0.1:9180/apisix/admin/stream_routes/1 -H "X-API-KEY: $admin_key" -X PUT -d '
{
    "snis": ["a.test.com", "b.test.com"],
    "upstream": {"nodes": {"127.0.0.1:5991": 1}, "type": "roundrobin"}
}'
```

Wildcards keep the existing suffix semantics (`*.test.com` also matches `a.b.test.com`), and a bare `*` puts no restriction on the SNI at all — the same as a route carrying neither field.

### Why the two are mutually exclusive

The schema carries `not: required[sni, snis]`, so a route with both is rejected at write time rather than resolved by a precedence rule nobody can guess. `ssl` makes the same choice for its own `sni`/`snis`.

It is expressed as `not` rather than `oneOf` on purpose: `oneOf` would mean *exactly one*, which would forbid a stream route carrying neither — that is a supported case, matching purely on `remote_addr`/`server_addr`/`server_port` (see `t/stream-node/sni.t` TEST 5 and TEST 11).

### Motivation: Gateway API TLSRoute

[TLSRoute](https://gateway-api.sigs.k8s.io/reference/api-types/tlsroute/) matches on

> the Server Name Indication (SNI) hostname provided during the TLS handshake

and its `hostnames` is a list. With `snis`, one TLSRoute rule translates to one stream route instead of one per hostname.

### Changes

- `apisix/schema_def.lua`: `stream_route.snis`, and `not: required[sni, snis]`.
- `apisix/stream/router/ip_port.lua`: `create_router()` registers one radixtree path per SNI instead of one per route. The `sni_to_items` dedup is preserved, so routes sharing an SNI still share one entry.
- `apisix/router.lua`: the shared route filter lowercases `snis` the way it already does for `hosts`. It now also lowercases `sni`, which fixes a pre-existing bug — a stream route with a mixed-case `sni` never matched, because `apisix/ssl.lua`'s `server_name()` always returns a lowercased SNI.

### Tests

`t/stream-node/stream-route-snis.t`: every SNI of a route matched and a third name not; the singular `sni` still working on its own; `sni` + `snis` rejected with 400; wildcard suffix matching; a bare `*` falling through to address matching; and a mixed-case entry matched by a lowercase SNI.

54 assertions, all passing. **10 of them fail without this change**, so the file tests the change rather than the framework.


membphis (approved):
LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):
we could eventually deprecate `sni`, but it doesn't pose a tech debt (only a tiny one) so it's fine either way. LGTM.

PR #13914, opened 2026-09-02
fix: preserve servlet upstream URI boundaries

shreemaan-abhishek:
### Description

URI-encode the original path before proxying when servlet-style normalization is enabled. This preserves servlet path parameters while keeping encoded delimiters as path data throughout routing and forwarding.

Add a focused regression test with overlapping exact and wildcard routes. The test verifies that the exact route still requires its API key and that an encoded question mark remains part of the forwarded path.

N/A

### Checklist


membphis (approved):
LGTM

AlinsRan (approved):

nic-6443 (approved):

PR #13915, opened 2026-09-03
fix: write request_type=websocket through ctx.var

janiussyafiq:
#### Motivation

Follow-up to #13909, addressing https://github.com/apache/apisix/pull/13909#issuecomment-5520928764.

`ctx.var` caches a variable on its first read. The header filter added in #13909 sets `ngx.var.request_type` directly, so a plugin that already resolved `$request_type` before the `101` upgrade completes leaves the cached value at `traditional_http`. A supported trigger is a WebSocket route whose `proxy-rewrite` sets a header to `$request_type`. The prometheus exporter reads `ctx.var` in the log phase, so `apisix_http_status`, `apisix_http_latency` and `apisix_bandwidth` label the whole session as ordinary HTTP and `request_type!="websocket"` no longer excludes it.

#### What this PR does

- Writes `request_type = "websocket"` through `api_ctx.var`, whose setter updates both the NGINX variable and the cached view, keeping the bare `ngx.var` write as the fallback when no request context exists.
- Adds a regression to `t/plugin/prometheus-websocket.t`: a route whose `proxy-rewrite` sets `X-Request-Type: $request_type` reads the variable before the upgrade. The log phase asserts that the upstream header carried `traditional_http` (the early read happened) and that `ngx.var` and `ctx.var` both report `websocket`, and the three metric families are checked for that route.

No behaviour change for requests that never read `$request_type` before the upgrade.

Fixes #

### Checklist

### DIFF
--- a/apisix/init.lua
+++ b/apisix/init.lua
@@ -1024,13 +1024,17 @@ function _M.http_header_filter_phase()
         set_resp_upstream_status(up_status)
     end
 
+    local api_ctx = ngx_ctx.api_ctx
     if ngx.status == 101 then
-        ngx_var.request_type = "websocket"
+        if api_ctx then
+            api_ctx.var.request_type = "websocket"
+        else
+            ngx_var.request_type = "websocket"
+        end
     end
 
     common_phase("header_filter")
 
-    local api_ctx = ngx.ctx.api_ctx
     if not api_ctx then
         return
     end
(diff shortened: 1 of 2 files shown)


membphis (dismissed):
LGTM

shreemaan-abhishek (dismissed):

nic-6443 (dismissed):

nic-6443 (approved):

membphis (approved):

AlinsRan (approved):

PR #13916, opened 2026-09-03
fix(openid-connect): validate introspection issuer

shreemaan-abhishek:
### Description

When remote introspection was used, `claim_validator.issuer.valid_issuers` was not applied to the introspection result. This change validates the `iss` field of a successful introspection response when an issuer allowlist is explicitly configured. Missing and non-matching issuer values return 401 with a Bearer `invalid_token` challenge.

Configurations without an explicit issuer allowlist keep the existing behavior.

N/A

### Breaking changes

For remote-introspection routes with an explicit issuer allowlist, the introspection response must now contain a string `iss` value that exactly matches one of the configured issuers. To retain the previous provider behavior, omit the issuer allowlist; otherwise configure the provider to include an allowed issuer in its response.

### Checklist


membphis (approved):

membphis (approved):
LGTM

AlinsRan (approved):

nic-6443 (approved):

PR #13917, opened 2026-09-04
fix(etcd): watch from the revision the configuration was read at

nic-6443:
### Description

A write made to etcd while APISIX is starting can be lost, permanently and without a log line.

The configuration is read once in `init_by_lua` and the snapshot is kept in `loaded_configuration`. Each resource type takes its own entry when it registers, and `core.config.new()` removes the entry as it does so. When the first worker starts watching, `do_run_watch()` took the start revision from `next(loaded_configuration)` — whatever entry was still there.

Once every preloaded type has registered, there is nothing there. `rev` stays 0, the code falls through to re-reading the current revision with `get(prefix)`, and the watch starts from *now* rather than from the snapshot. Anything written in between is never delivered to that worker, and nothing says so.

What keeps this from being permanently broken today is a coincidence: the http worker registers every type in `HTTP_ETCD_DIRECTORY` except `/stream_routes`, and the stream worker registers every type in `STREAM_ETCD_DIRECTORY` except `/ssls`, so one entry always survives. That balance is not a design — it shifts with the subsystem, with `enable_admin` (which decides whether `/plugins` is registered), and with which plugins are loaded (`/protos` comes from `grpc-transcode`). Registering one more type, or running against an etcd where one of those directories was never created, is enough to turn the loss on.

### The fix

Record the revision where it is produced. `init_loaded_configuration()` performs a single `readdir`, and `create_formatter()` stamps every entry it stores with that one response's headers, so there is exactly one revision to keep and no ambiguity about which entry to read it from. The test for "was anything preloaded" is made there, before anything has been consumed, which is precisely what makes it correct — the same test at watch time is the bug.

**An empty read is deliberately left alone**, and finding that boundary is most of what this PR is. When nothing is preloaded the start revision decides nothing: every resource type has no entry to take, so it reads for itself at first sync, and that read already returns whatever was written since. An empty-prefix version of the test below passes unpatched, 2/2.

Meanwhile that branch does carry something. `rev == 0` runs a `get(prefix)` retry loop, and the loop doubles as an undocumented readiness gate: while the endpoint is unhealthy the watch never starts, and resource types keep polling `readdir`, which still converges. Recording a revision unconditionally skipped the loop, started the watch against an endpoint that could not serve it, and left those types waiting on a watch instead of polling. `t/core/etcd-mtls.t` TEST 3 caught it — 2 failures in 4 runs against a freshly emptied etcd, 0 in 4 without the change, and green at the same sample size once the recording was narrowed. So the `rev == 0` path keeps both of its jobs, untouched.

### On the X-Etcd-Index handling

The check that #13364 added at the watch site is not carried over to the new capture site, because it can no longer be reached there.

This is a consequence of #13361, not a judgement about #13364. When #13364 landed, `get_format()` assigned `res.headers["X-Etcd-Index"]` from `res.body.header.revision` without checking it, so a response missing that field produced an entry with no usable header and the branch was live — its test was guarding something real. #13361 then made `get_format()` call `get_header_revision()` and fail the read when the revision is absent. From that point on a `readdir` can only return successfully with a usable header, and the branch became dead.

So the degradation path for a backend that omits `header.revision` is now the one #13361 established, and it is strictly louder than before: `get_format()` returns `nil, "etcd response missing header.revision"`, `readdir()` propagates it, `init_loaded_configuration()` returns it, and `init.lua` logs `failed to load the configuration:` and continues without rethrowing. `loaded_configuration_rev` is then nil, `rev` is 0, and the `get(prefix)` retry loop takes over — which itself re-checks `body.header.revision` and keeps retrying with a log line until the backend answers properly.

`t/core/config_etcd.t` TEST 15 keeps its slot rather than being deleted. Its injection wrote an entry with empty headers directly into `loaded_configuration`, which no longer reaches anything; it now reaches the same fallback through `disable_sync_configuration_during_start`, a real configuration that needs no stubbing, and asserts that the watch still starts. The numbering of the cases after it is unchanged.

### Tests

`t/core/etcd-watch-start-revision.t` drives the preload table empty using only configuration — an etcd prefix holding a single directory, so `/routes` is the one type with anything to consume — then writes a route between the snapshot and the first watch, from `extra_init_by_lua`. The route is not in the snapshot, so it can only arrive through the watch. It fails on `master` with `route written during startup was lost` and passes with the fix, deterministically in both directions.

The test checks the etcd responses rather than trusting `io.popen`, which reports nothing about how `curl` fared, and it clears its prefix at both ends of the run.

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -48,6 +48,7 @@ local debug        = debug
 local string       = string
 local error        = error
 local pairs        = pairs
+local next         = next
 local rand         = math.random
 local constants    = require("apisix.constants")
 local health_check = require("resty.etcd.health_check")
@@ -1216,11 +1217,15 @@ local function init_loaded_configuration()
     end
 
     -- One readdir backs every entry create_formatter() stored, so there is a
-    -- single revision to record, and it is recorded even when the range came
-    -- back empty: the snapshot is still a point in time to watch from.
-    -- get_format() fails the read when the response carries no revision, so
-    -- there is nothing to validate here.
-    loaded_configuration_rev = tonumber(res.headers["X-Etcd-Index"])
+    -- single revision to record. Nothing is consumed yet at this point, so an
+    -- empty table here means the read stored nothing -- unlike the same test
+    -- made at watch time, which is the bug being fixed. Leave that case alone:
+    -- with no preloaded data every resource type reads for itself at first
+    -- sync, which already picks up whatever was written since, and the
+    -- rev == 0 path below keeps handling it exactly as before.
+    if next(loaded_configuration) then
+        loaded_configuration_rev = tonumber(res.headers["X-Etcd-Index"])
+    end
 
     configuration_loaded_time = ngx_time()
 end
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -1218,14 +1218,9 @@ local function init_loaded_configuration()
     -- One readdir backs every entry create_formatter() stored, so there is a
     -- single revision to record, and it is recorded even when the range came
     -- back empty: the snapshot is still a point in time to watch from.
-    local rev = tonumber(res.headers["X-Etcd-Index"])
-    if not rev or rev <= 0 then
-        log.warn("invalid or missing X-Etcd-Index header, the watch will ",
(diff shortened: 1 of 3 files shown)


AlinsRan (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13921, opened 2026-09-07
fix(ci): unbreak CI under pnpm 12

nic-6443:
CI is red everywhere, and it is not a test failure. Two separate breakages, one root cause: pnpm 12 promoted two long-standing warnings into hard errors, and nothing pins the pnpm used outside `t/`.

**1. Environment setup dies before anything runs.** `install_nodejs()` calls `pnpm setup` under sudo:

```
Error: ERR_PNPM_SUDO_NOT_SUPPORTED
  × Running "pnpm setup" with sudo is not supported
```

pnpm 11.25.0 only warned here ("will fail in pnpm v12"); 12 made it fatal. The call is also redundant — `pnpm setup` exists to write `PNPM_HOME` and its `PATH` entry into a shell profile, and `install_nodejs()` already exports both in the same shell. The pnpm binary comes from the corepack shim on `PATH`, not from `setup`. So this one is removed rather than worked around, which also stops us running a configuration pnpm has said it does not support.

**2. `t/plugin/grpc-web.t` then fails.** With setup fixed, CI got far enough to reveal the second one — `pnpm dlx` now treats an unapproved build script as fatal, and that test runs its client via `pnpx tsx`:

```
Error: ERR_PNPM_IGNORED_BUILDS
  × adding a new package
  ╰─▶ Ignored build scripts: esbuild@0.28.2
```

This is why removing `pnpm setup` alone is not enough, and why the pin is needed: the last green run resolved pnpm to **11.25.0** at the repository root, where nothing pins it, so `corepack prepare pnpm@11.25.0 --activate` puts that back and stops the version moving on its own. Commands under `t/` are unaffected — its `package.json` already pins `packageManager` to 10.14.0.

The pin has a stated expiry in the comment: it can be dropped once `grpc-web.t` no longer needs pnpm to run esbuild's build script unprompted.

Verified in a clean `ubuntu:24.04` container reproducing the CI setup: on 11.25.0 `pnpx tsx` is clean, on 12.3.2 and 12.3.4 it fails as above; `pnpm install`, `pnpm exec jest` and the global `pnpm install -g tsx` in `t/plugin/grpc-web/setup.sh` all work without `pnpm setup`, with `tsx` landing in `PNPM_HOME` on `PATH`.

### DIFF
--- a/ci/common.sh
+++ b/ci/common.sh
@@ -108,7 +108,6 @@ install_nodejs () {
     export PNPM_HOME="/pnpm"
     export PATH="$PNPM_HOME:$PATH"
     corepack enable pnpm
-    pnpm setup
 }
 
 install_brotli () {
--- a/ci/common.sh
+++ b/ci/common.sh
@@ -108,6 +108,15 @@ install_nodejs () {
     export PNPM_HOME="/pnpm"
     export PATH="$PNPM_HOME:$PATH"
     corepack enable pnpm
+    # Pin the pnpm used outside t/, which has its own packageManager pin.
+    # pnpm 12 turned two long-standing warnings into hard errors that both
+    # break CI: running "pnpm setup" under sudo, and "pnpm dlx" installing a
+    # package whose dependency has an unapproved build script -- the latter
+    # hits t/plugin/grpc-web.t, which runs the client through "pnpx tsx"
+    # (esbuild). 11.25.0 is the last version CI was green on.
+    # Expiry: drop this pin once grpc-web.t no longer needs pnpm to run
+    # esbuild's build script unprompted.
+    corepack prepare pnpm@11.25.0 --activate
 }
 
 install_brotli () {


nic-6443:
Flagging this one for attention since master is currently red and every PR is affected — `CI`, `CI Kubernetes`, `CI Tars` and `CLI Test` all fail on `f3623982e`, and they fail in `install_nodejs` before anything is built, so no test actually runs.

Nothing in the codebase changed to cause it; pnpm 12.3.4 shipped on 2026-09-04 and started rejecting `pnpm setup` under sudo. The commit boundary is misleading here — master simply had no CI run between the pnpm release and `f3623982e`. A PR branch based before that commit (`cac900ba3`, run on 09-05) fails identically, which rules the code out.

The change is one line, and CI on this PR is getting past the point that was failing.

shreemaan-abhishek (dismissed):

membphis (dismissed):

AlinsRan (dismissed):

membphis (approved):

nic-6443:
Pushed a second commit, so the approval above predates it — worth a re-look.

Removing `pnpm setup` did get CI past environment setup, and most shards went green, but it exposed a second pnpm 12 breakage underneath: `pnpm dlx` now treats an unapproved build script as fatal, so `t/plugin/grpc-web.t` fails on `pnpx tsx` with `ERR_PNPM_IGNORED_BUILDS` (esbuild). Same root cause, different symptom.

The last green run resolved pnpm to 11.25.0 at the repository root, where nothing pins it, so the new commit pins that explicitly. `t/` is unaffected — it already pins 10.14.0 via `packageManager`. I left an expiry note on the pin. Description updated.

AlinsRan (approved):

shreemaan-abhishek (approved):

PR #13922, opened 2026-09-07
fix(ai-aliyun-content-moderation): report final results without usage

nic-6443:
Streaming responses without `usage` could skip response moderation or end without a client-visible result.

Record whether the provider supplied usage and finalize assembled text at protocol completion or clean EOF. In `final_packet` mode, retain the existing data-event loop and add `deny_message` alongside `risk_level`, preserving content and usage. The usage flag only controls whether to insert an additional Chat or Anthropic result before the terminator; Responses retains its existing completed-event carrier. Allowed results use an empty denial message. Injected events include zero usage. Chat uses an empty usage chunk (`choices: []`): denial text is returned only in the top-level `deny_message` field, without adding model output or changing finish reasons.

The final-packet filter returns before decoding until assembled text is available; error detection reuses the provider parser's existing result. Native SSE dispatch reuses the parser's complete-frame byte boundaries, preserving comments and blank lines without a second plugin buffer. Incomplete tails and errors must not synthesize successful termination. A client completion marker prevents converted streams from receiving a second terminator during their final empty flush.

Results are informational and cannot retract streamed content. Existing usage remains unchanged; injected events report zero usage. Responses SDK wrappers may discard unknown top-level fields, so read the extensions from raw events. English and Chinese documentation describes the behavior.

Integration coverage includes usage and no-usage streams, split/coalesced frames and terminators, in-place and injected fields, safe/rejected/failed scans, multiple choices, protocol conversion, and truncated streams. Lakera tests verify scanning without usage while retaining fail-open/fail-closed coverage for incomplete streams. Provider conversion semantics and plugin schemas are unchanged.


AlinsRan (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #13934, opened 2026-09-11
fix(etcd): check availability before starting watcher

shreemaan-abhishek:
### Description

The shared etcd watcher publishes `watch_ctx.started` before calling `watchdir`. Config objects interpret that state as permission to wait only for events from the watcher. When a preload revision was available, watcher initialization skipped the range request that otherwise verified etcd was reachable, so config objects could wait on a watcher that had not connected.

This change extracts that range request into an explicit availability check and runs it during every initial watcher setup. The preload revision remains the watch cursor when present, preventing writes made after the snapshot from being skipped; the revision returned by the availability check is used only when no preload revision exists.

The startup-revision regression test now also verifies that the availability check occurs when the preload supplies the revision.

N/A

### Checklist

### DIFF
--- a/apisix/core/config_etcd.lua
+++ b/apisix/core/config_etcd.lua
@@ -128,6 +128,29 @@ local function produce_res(res, err)
 end
 
 
+local function wait_for_etcd_available(etcd_cli, prefix)
+    while true do
+        local res, err = etcd_cli:get(prefix)
+        if not res then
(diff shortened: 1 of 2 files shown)


nic-6443 (approved):

membphis (approved):
LGTM

AlinsRan (approved):

PR #13052, opened 2026-02-28
docs: add stream proxy example for limit-conn plugin (#4933)

ashruthg:
Add a stream proxy usage example to the limit-conn plugin documentation.

PR #13936, opened 2026-09-14
docs: add stream proxy example for limit-conn plugin

### LINKED ISSUE #4933: docs: limit-conn plugin in stream proxy
### Improved or typo fixed.

Currently, the limit-conn is supported in the stream proxy, but the documentation about limit-conn doesn't give a good example to show how to use it in stream proxy.

Documentation link: http://apisix.apache.org/docs/apisix/plugins/limit-conn.

### Solution

We may add an example for how to use limit-conn in the stream proxy.

bhuvan-somisetty:
### Description

The `limit-conn` plugin doc has no example showing how to use it in stream (L4/TCP) proxy mode, even though the plugin has supported stream proxy since it was written (`apisix/stream/plugins/limit-conn.lua`).

**Root cause:** the doc gap, plus a prior attempt (#13052) that tried to close this issue documented an incorrect restriction: only `remote_addr`/`server_addr` can be used as `key`. That's not accurate: in `apisix/plugins/limit-conn/init.lua`, when `key_type` is `var` (the default), the key is resolved via `ctx.var[conf_key]`, i.e. plain NGINX stream-variable lookup. Any variable available in the stream context works (e.g. `server_port`), not just `remote_addr`/`server_addr`.

**Solution:**
- Added an "Apply Rate Limiting in Stream Proxy" section to `docs/en/latest/plugins/limit-conn.md`, following the doc's existing Admin API / ADC / Ingress Controller tabbed example format, with a full, copy-pasteable `stream_routes` example (including `upstream`).
- Documented the `key` resolution behavior accurately instead of the narrower (incorrect) restriction from the earlier attempt.
- Added a regression test in `t/stream-plugin/limit-conn.t` (`key: "server_port"` with default `key_type: var`) so the corrected documentation claim is backed by an actual test, guarding against future regressions in the key-resolution logic.

Testing: ran the modified `t/stream-plugin/limit-conn.t` (all 12 test blocks, including the 2 new ones) locally against a real OpenResty + etcd environment; all pass. Also ran `markdownlint-cli` against the modified doc file with the repo's `.markdownlint.yml` config; no findings.

Fixes #4933

### Checklist


kayx23 (commented):
I found a few issues that affect the accuracy and reproducibility of the new stream example.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,153 @@ Response: 429
...
+        "burst": 1,
+        "default_conn_delay": 0.1,
+        "key_type": "var",
+        "key": "remote_addr"
+      }
+    },
+    "upstream": {
+      "type": "roundrobin",
+      "nodes": {
+        "127.0.0.1:1995": 1
```
kayx23, line 1640: The example is not runnable as written because it does not start a TCP service on `127.0.0.1:1995`. It also depends on stream mode and a TCP listener on port `9100`, but that prerequisite is only discoverable by leaving this section. Could you state the listener prerequisite here and provide a reachable upstream or the commands needed to create one? The Ingress example has the same issue because `tcp-echo.aic.svc` is not defined by the manifest.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,153 @@ Response: 429
...
+
+</TabItem>
+<TabItem value="ingress" label="Ingress Controller">
+
+<Tabs groupId="k8s-api">
+<TabItem value="gateway-api" label="Gateway API">
+
+:::info
+
+Attaching L4 Plugins is currently not supported with Gateway API. This example cannot be completed with Gateway API.
```
kayx23: This is no longer accurate for the current released controller. APISIX Ingress Controller 2.2.0 introduced `apisix.apache.org/v1alpha1` `L4RoutePolicy`, which attaches stream plugins such as `limit-conn` to `TCPRoute`, `UDPRoute`, and `TLSRoute`. Please replace this callout with a `TCPRoute` plus `L4RoutePolicy` example, or qualify the limitation to controller versions before 2.2.0.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,153 @@ Response: 429
...
+</Tabs>
+
+❶ `key_type`: set to `var` to interpret `key` as a variable.
+
+❷ `key`: calculate rate limiting count by the connection's `remote_addr`.
+
+Open three concurrent TCP connections to the stream Route:
+
+```shell
+seq 1 3 | xargs -n1 -P3 -I{} bash -c 'echo "hello" | nc -w 1 127.0.0.1 9100'
```
kayx23: With `conn: 2` and `burst: 1`, APISIX permits three concurrent connections; rejection starts with the fourth. This command opens only three, and the short `echo | nc -w 1` sessions may not overlap reliably, so it cannot verify the following rejection claim. Please use a deterministic test, such as `conn: 1`, `burst: 0`, holding one connection open while a second is reset, then retrying after the first closes. The documentation should also show the actual observed output.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,153 @@ Response: 429
...
+
+### Apply Rate Limiting in Stream Proxy
+
+The `limit-conn` Plugin can also be used on a [stream Route](../stream-proxy.md) to limit the number of concurrent TCP connections.
+
+The following example demonstrates how to rate limit TCP connections by `remote_addr`, with example connection and burst thresholds.
+
+:::note
+
+When `key_type` is `var` (the default), `key` is resolved as an [NGINX stream module variable](https://nginx.org/en/docs/stream/ngx_stream_core_module.html). It is not limited to `remote_addr` or `server_addr`: any variable available in the stream context, such as `server_port`, can be used.
```
kayx23, line 1579: The stream plugin has a distinct, smaller schema: `conn`, `burst`, `default_conn_delay`, `only_use_default_delay`, `key`, and `key_type`; `conn` and `burst` are integers. The page’s existing Attributes table describes the HTTP plugin and includes HTTP-only fields such as Redis policies, `rules`, degradation, and custom HTTP rejection fields. Adding a stream example without separating or explicitly scoping those attributes can lead readers to apply unsupported fields to stream routes. Could this PR add a separate stream attribute table or clearly qualify the existing table?

t/stream-plugin/limit-conn.t
```diff
@@ -339,3 +339,51 @@ GET /test_concurrency
...
+
+
+
+=== TEST 12: exceeding the burst when keyed by a non-address var
+--- request
+GET /test_concurrency
+--- response_body
+200
+200
+200
```
kayx23, line 384: This result does not prove that `server_port` was resolved and used as the limiter key. If lookup failed, the implementation would fall back to `remote_addr` and produce the same three-success/two-rejection pattern. Please assert that the logged limit key contains the expected listener port (the existing remote-address test uses this pattern) and assert that `The value of the configured key is empty` is absent.

bhuvan-somisetty:
Thanks for the detailed review! Fixed all of these - added the missing listener/upstream prerequisites, scoped the attribute table to the stream schema, swapped the outdated Gateway API note for a TCPRoute + L4RoutePolicy example, made the rejection test deterministic with conn=1/burst=0, and tightened the stream test to actually assert the resolved key. PTAL

docs/en/latest/plugins/limit-conn.md
kayx23: The listener and upstream setup are now documented, but one portability issue remains: when APISIX runs in Docker, `127.0.0.1:1995` refers to the APISIX container, not the host running `socat`. Please scope this workflow to a host installation or add Docker-specific networking instructions.

docs/en/latest/plugins/limit-conn.md
kayx23: The original unsupported-Gateway-API statement has been removed and `L4RoutePolicy` coverage was added. A separate inline comment tracks the API version used by the new `TCPRoute` manifest.

docs/en/latest/plugins/limit-conn.md
kayx23: The original threshold mismatch is fixed by using `conn: 1` and `burst: 0`. A separate inline comment tracks the remaining portability issue in the replacement `nc` procedure.

docs/en/latest/plugins/limit-conn.md
kayx23: The stream attributes are now separated, but the `only_use_default_delay` description still reflects HTTP behavior. The stream subsystem does not calculate request latency, so this setting does not change stream behavior. Also, connections at `conn + burst` are accepted; use “at or below,” not “below.”

t/stream-plugin/limit-conn.t
kayx23: Fixed. The test now verifies that the resolved key contains `1985` and that fallback does not occur.

kayx23 (commented):
Three details remain in the revised documentation.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,296 @@ Response: 429
...
+  name: tcp-echo-service
+spec:
+  selector:
+    app: tcp-echo
+  ports:
+    - name: tcp-echo
+      port: 1995
+      targetPort: 1995
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
kayx23: APISIX Ingress Controller 2.2.0 uses Gateway API 1.6.0. Its standard `TCPRoute` CRD serves `gateway.networking.k8s.io/v1`; `v1alpha2` is not served. Please use `apiVersion: gateway.networking.k8s.io/v1`.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,296 @@ Response: 429
...
+</Tabs>
+
+❶ `key_type`: set to `var` to interpret `key` as a variable.
+
+❷ `key`: calculate rate limiting count by the connection's `remote_addr`.
+
+With `conn` set to `1` and `burst` set to `0`, only one concurrent connection from the same `remote_addr` is allowed, and a second concurrent connection is rejected immediately. Open one long-lived connection in the background, then attempt a second connection while the first is still open:
+
+```shell
+nc 127.0.0.1 9100 &
```
kayx23: `nc ... &` does not reliably keep the first connection open across shells because its standard input may be suspended or reach EOF. Use a deterministic source such as `sleep 30 | nc 127.0.0.1 9100 &`, or hold the connection open in another terminal. The following paragraph also needs adjustment because this block already runs `kill %1`.

docs/en/latest/plugins/limit-conn.md
```diff
@@ -1567,3 +1567,296 @@ Response: 429
...
+kubectl apply -f limit-conn-stream-ic.yaml
+```
+
+</TabItem>
+</Tabs>
+
+</TabItem>
+</Tabs>
+
+❶ `key_type`: set to `var` to interpret `key` as a variable.
```
kayx23: The `❶` and `❷` explanations do not correspond to annotation markers in any of the configuration examples. Please add matching markers to the relevant fields or replace these labels with ordinary prose.

bhuvan-somisetty:
Good catches, all fixed - docker networking note added, only_use_default_delay description corrected for stream, TCPRoute bumped to v1, the nc test now uses sleep+$! and doesn't contradict itself, and dropped the dangling markers. PTAL

docs/en/latest/plugins/limit-conn.md
kayx23: Fixed. The example now states the host assumption and explains how to reach the upstream when APISIX runs in Docker.

docs/en/latest/plugins/limit-conn.md
kayx23: One factual detail remains: “excess connections are always delayed by `default_conn_delay`” is not correct. `resty.limit.conn` calculates the delay as `default_conn_delay × floor((current connections - 1) / conn)`, so it can be a multiple of `default_conn_delay`. Please keep the first two sentences and replace the final clause with that calculation.

docs/en/latest/plugins/limit-conn.md
kayx23: Fixed. `TCPRoute` now uses `gateway.networking.k8s.io/v1`.

docs/en/latest/plugins/limit-conn.md
kayx23: Fixed. The procedure now uses `sleep 30 | nc ... &`, saves the process ID, and closes the first connection in a separate step.

docs/en/latest/plugins/limit-conn.md
kayx23: Fixed. The dangling numbered annotations were replaced with ordinary prose.

docs/en/latest/plugins/limit-conn.md
bhuvan-somisetty: good catch, fixed the formula

kayx23 (approved):

PR #13937, opened 2026-09-14
test(xds): wait for the privileged agent's startup sync before flush_all in TEST 4

janiussyafiq:
### Description

`t/xds-library/config_xds_2.t` TEST 4 fails intermittently on loaded runners with:

```
[error] [lua] config_xds.lua:174: decode the conf of [/routes/2] failed, err: bad argument #1 to '?' (string expected, got nil), conf_str: nil, context: init_worker_by_lua*
[error] [lua] config_xds.lua:180: invalid conf of [/routes/2], conf: nil, it should be an object, context: init_worker_by_lua*
```

APISIX runs one worker plus the privileged agent, and both execute the blocking startup sync in `config_xds._M.new` during `init_worker`. The worker begins serving before the agent finishes that sync. TEST 4 calls `ngx.shared["xds-config"]:flush_all()` on the very first line of the request, so when it lands during the agent's `get_keys()` / `get()` loop, the keys it already listed are expired under it, `get()` returns nil, and the `[error]` lines above trip the block's `no_error_log` assertion. The `init_worker_by_lua*` context in the log confirms the failing sync is the startup one, not the periodic timer.

This PR adds a `ngx.sleep(1.5)` before the flush, the same guard TEST 3 already uses for its own `flush_all()`.

Verified locally by running the file repeatedly with 24 CPU hogs on an 11-core box:

| | failures | which test |
|---|---|---|
| before | 7/30, then 10/24 | all TEST 4 |
| after | 0/24 | none |

N/A

### Checklist


nic-6443 (approved):

nic-6443 (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

PR #13018, opened 2026-02-17
fix: Set upstream and upstream_latency vars in ext-plugin-post-resp for logging

MariaLapovska:
### Description

When using ext-plugin-post-resp and http-logger, some values are missing from the logs, specifically the ones related to upstream - upstream and upstream_latency.
This is happening because upstream_response_time and upstream_addr nginx vars are unavailable, so when log-util.lua is trying to inject them into logs, no value is present.
This PR adds these two vars to ext-plugin-post-resp, they are now set explicitly after the call to upstream.

#### Which issue(s) this PR fixes:

Fixes https://github.com/apache/apisix/issues/13016

### Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [ ] I have updated the documentation to reflect this change
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [APISIX mailing list](https://github.com/apache/apisix/tree/master#community) first)

PR #13940, opened 2026-09-14
fix(ext-plugin-post-resp): set upstream_addr and upstream_response_time for loggers

### LINKED ISSUE #13016: bug: Some nginx vars related to upstream are missing when using ext-plugin-post-resp
### Current Behavior

When I'm using ext-plugin-post-resp and http-logger, some values are missing from the logs, specifically the ones related to upstream - upstream and upstream_latency (same as described in this issue - https://github.com/apache/apisix/issues/10369).

### Expected Behavior

I believe this isis happening because `upstream_response_time` and `upstream_addr` nginx vars are unavailable, so when log-util.lua is trying to inject them into logs, no value is present. I understand that when using ext-plugin-post-resp, the mechanics of calling upstream is different, but I still expect to have these values in the logs, no matter what route configuration I have.

For now, the only way for me to fix it, was to change _M.before_proxy method in ext-plugin-post-resp.lua and explicitly set these vars:
```
function _M.before_proxy(conf, ctx)
    local start_time = ngx.now()  -- EDIT: Start timer here
    local http_obj = http.new()
    local res, err = get_response(ctx, http_obj)

    -- EDIT: Set upstream vars, right after get_response returns
    ctx.var.upstream_response_time = math.floor((ngx.now() - start_time) * 1000 + 0.5) / 1000
    ctx.var.upstream_addr = ctx.picked_server.host .. ":" .. ctx.picked_server.port
```

### Error Logs

_No response_

### Steps to Reproduce

1. Run APISIX
2. Create a route that would have two plugins - http-logger and ext-plugin-post-resp
3. Try to make call to upstream
4. Check the logs sent by http-logger and see that upstream and upstream_latency are missing

### Environment

- APISIX version (run `apisix version`): 3.13.0
- Operating system (run `uname -a`): Darwin 25.2.0
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`): 9.7.3
- APISIX Dashboard version, if relevant: 0.8.2
- Plugin runner version, for issues related to plugin runners: ext-java-plugin-runner 0.6.0

janiussyafiq:
### Description

`ext-plugin-post-resp` sends the upstream request with lua-resty-http, so the nginx proxy module never sets `$upstream_addr` and `$upstream_response_time`. Loggers built on `log-util` then report no `upstream` and no `upstream_latency`, and the whole upstream round trip is attributed to `apisix_latency`.

This PR fills `ctx.var.upstream_addr` and `ctx.var.upstream_response_time` right after the upstream call, and adds a test that checks both fields reach `http-logger`.

Supersedes #13018 (same approach, credit to @MariaLapovska; that PR went stale waiting for tests).

Fixes #13016

### Checklist


nic-6443 (dismissed):

membphis (commented):
### [P2] Finish upstream timing after the response body has been read

`get_response()` calls `http_obj:request()`, which returns the response headers and a body reader. Setting `ctx.var.upstream_response_time` immediately afterward stops the timer before the body is consumed. The body is read later by `send_response()` or by the runner's `RespBody` handling in `ext-plugin/init.lua`.

If an upstream sends headers promptly and then delays the body, the new field underreports upstream latency and `log-util` still attributes the body wait to `apisix_latency`. The added test only checks that the field is numeric, so it also passes with a zero or incomplete duration.

Please finalize the timing when upstream body reading completes or terminates, cover both the direct forwarding and runner-read paths, and exclude runner processing time. Add a regression upstream that sends headers immediately and delays its body, then assert that the reported upstream latency includes that delay.

Evidence: `apisix/plugins/ext-plugin-post-resp.lua:154`, the subsequent `send_response()` call, and the `RespBody` branch in `apisix/plugins/ext-plugin/init.lua`.

membphis (dismissed):
LGTM

shreemaan-abhishek (approved):

membphis (approved):
LGTM

nic-6443 (approved):

PR #13947, opened 2026-09-15
fix(ai-proxy): avoid aborting streams on empty flushes

nic-6443:
Streaming could be aborted as a client disconnect when an SSE fragment, skipped converter event, or buffering response filter left no downstream output for a periodic flush.

Arm periodic flushing after successful native or converted dispatch, including the final empty dispatch that can release buffered output, and treat `nothing to flush` as nonfatal. Continue the flush thread silently for this condition; other flush errors retain the existing abort behavior.

Add integration coverage for fragmented frames, skipped converter events, synchronous flushing, and a buffering response filter that exercises the real empty-flush return and verifies complete delivery without empty-flush warnings. Converted completion coverage verifies that buffered output reaches the client before the upstream closes the connection.

Document the streaming flush behavior in English and Chinese.


nic-6443:
The failed CLI job stops in `test_etcd_sync_event_handle.sh`: the first authenticated etcd write after restarting the test services times out. The test writes immediately after container startup, without waiting for etcd readiness.

#13948 adds an authenticated health check before those writes and retains the synchronization assertions. It also fixes the plugin-checker test race. This PR can pick up those test fixes after #13948 lands.

AlinsRan (approved):

membphis (approved):
LGTM

shreemaan-abhishek (approved):

PR #12462, opened 2025-07-25
fix: skipped failing bailedout tests in CI

Revolyssup:
The test here https://github.com/apache/apisix/actions/runs/16511586539/job/46694453675
failed due to invalid configuration but in case of Bailout, the tests is considered successful. This change asserts and finds for the bailed out test and fails CI if any test was bailed out.

The following tests fail after the changes in script in this pr https://github.com/apache/apisix/actions/runs/16519080359/job/46716551717
https://github.com/apache/apisix/actions/runs/16519816286/job/46718888226?pr=12462#step:16:236

I have modified the above tests to finally pass.
## Further enquiries
The doc says
"By default the test scaffold treats NGINX server startup failures as fatal errors in running the tests. The must_die section, however, turns such a failure into a normal test checkup."

If you have just one test then it holds true. But when even one test passes and subsequent test has invalid config then TEST IS CONSIDERED PASS.

WHen invalid configuration was passed in Test 1 (As expected the Bailout is considered total test failure)

```bash
Test run interrupted!
Test Summary Report
-------------------
t/core/ctx.t (Wstat: 65280 (exited 255) Tests: 0 Failed: 0)
  Non-zero exit status: 255
Files=1, Tests=0, 31 wallclock secs ( 0.01 usr  0.00 sys +  0.19 cusr  0.17 csys =  0.37 CPU)
Result: FAIL
```

When invalid configuration was passed in Test 2(or any other test after one passed test): (The test pass even after bailout)

```bash
Bailout called.  Further testing stopped:  t/core/ctx.t TEST 2: http header - Cannot start nginx using command "/usr/bin/openresty -p /home/ashish/dev/apisix/t/servroot/ -c /home/ashish/dev/apisix/t/servroot/conf/nginx.conf > /dev/null" (status code 256).
t/core/ctx.t .. ok
Test run interrupted!
All tests successful.
Files=1, Tests=6, 31 wallclock secs ( 0.02 usr  0.01 sys +  0.27 cusr  0.22 csys =  0.52 CPU)
Result: PASS
All tests successful.
Files=1, Tests=6, 31 wallclock secs ( 0.02 usr  0.01 sys +  0.27 cusr  0.22 csys =  0.52 CPU)
Result: PASS
[...]

PR #13948, opened 2026-09-15
test: stabilize plugin tests and cover CI bailouts

nic-6443:
The plugin checker test replaces `etcd.get_format` globally. While it waits for the configuration watcher, a background timer can receive the fake revision and fail its compare-and-swap update. Restrict the replacement to the test response and delegate other responses to the original formatter. The existing assertions are preserved; the race depends on startup scheduling.

Add regression coverage for the existing CI bailout guard from #12462: a completed passing suite succeeds, and a bailout fails even when the final summary says `Result: PASS`. Document how function replacements interact with background timers.

Wait for an authenticated etcd health check after restarting the test services, before writing the queued route updates. Container startup can finish before etcd accepts requests; all existing synchronization and log assertions remain in place.

Clear the Sentinel test Redis before checking database isolation, so unexpired counters from earlier cases cannot inflate the key counts. Keep the alternating requests and per-database counter assertions.

Poll Kafka for the current request’s log entry within a bounded deadline before checking its timestamp. Run the sliding-window burst assertions within one window and check that the burst does not cross its boundary; all three request statuses remain asserted.

Wait for the Loggly metadata format to reach the worker cache before sending the request that asserts that format. Fail explicitly if it does not propagate within five seconds; preserve the response and complete syslog assertions.

Check the Elasticsearch metadata DELETE response so the queue-limit test fails when cleanup fails, instead of leaving its metadata behind without reporting the error.

Add coverage for Kafka topic initialization: retry failed creation until the broker is ready, and stop the setup process when retries are exhausted so a later successful command cannot hide the failure. Clarify that service-backed tests need the services and topic layout selected by their CI job.

Checklist:

- [x] Explain the problem and why the change is needed.
- [x] Describe the changes.
- [x] Add or update corresponding tests.
- [x] Update the testing documentation.
- [x] Verify backward compatibility: only tests and test documentation change.


AlinsRan (approved):

shreemaan-abhishek (approved):

membphis (approved):
LGTM

